PR #2728, opened 2019-07-30
Do not use Count when Any can be used

paulomorgado:
This PR adds diagnostic CA1828 as counterpart for CA1827 for asynchronous methods consolidating the analyzer, fixer and tests implementation.


src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.sarif
```diff
@@ -2144,6 +2128,38 @@
...
+              "Visual Basic"
+            ]
+          }
+        },
+        "CA1828": {
+          "id": "CA1828",
+          "shortDescription": "Do not use CountAsync() when AnyAsync() can be used",
+          "fullDescription": "For non-empty collections, CountAsync() enumerates the entire sequence, while AnyAsync() stops at the first item or the first item that satisfies a condition.",
+          "defaultLevel": "warning",
+          "helpUri": "https://docs.microsoft.com/visualstudio/code-quality/ca1828",
```
mavasani: Kindly file a tracking docs issue to author this page.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpDoNotUseCountAsyncWhenAnyAsyncCanBeUsed.Fixer.cs
```diff
@@ -160,14 +174,34 @@ private static void GetFixerForGreaterThanOrEqualExpression(BinaryExpressionSynt
...
         {
-            GetExpressionAndInvocationArguments(invocationExpression, out expression, out arguments);
+            GetExpressionAndInvocationArguments(sourceExpression, isAsync, out expression, out arguments);
             negate = (int)literalExpression.Token.Value == value;
         }
 
-        private static void GetExpressionAndInvocationArguments(InvocationExpressionSyntax invocationExpression, out SyntaxNode expression, out IEnumerable<SyntaxNode> arguments)
+        private static void GetExpressionAndInvocationArguments(ExpressionSyntax sourceExpression, bool isAsync, out SyntaxNode expression, out IEnumerable<SyntaxNode> arguments)
         {
+            while (sourceExpression is ParenthesizedExpressionSyntax parenthesizedExpression)
```
mavasani: We can move this loop into an extension method, say `WalkUpParentheses`, if one does not already exist.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpDoNotUseCountAsyncWhenAnyAsyncCanBeUsed.Fixer.cs
```diff
@@ -160,14 +174,34 @@ private static void GetFixerForGreaterThanOrEqualExpression(BinaryExpressionSynt
...
+            {
+                sourceExpression = parenthesizedExpression.Expression;
+            }
+
+            if (isAsync)
+            {
+                if (sourceExpression is AwaitExpressionSyntax awaitExpressionSyntax)
+                {
+                    sourceExpression = awaitExpressionSyntax.Expression;
```
mavasani: Can the source expression be something other then an `InvocationExpressionSyntax` here, say for error cases? If so, should we be more defensive and check `if (sourceExpression is AwaitExpressionSyntax awaitExpressionSyntax && awaitExpressionSyntax.Expression is InvocationExpressionSyntax)`?

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.Fixer.cs
```diff
@@ -16,18 +16,24 @@
 using Microsoft.CodeAnalysis.Editing;
 using Microsoft.CodeAnalysis.Formatting;
 
-namespace Microsoft.CodeQuality.Analyzers.Performance
+namespace Microsoft.NetCore.Analyzers.Performance
 {
     /// <summary>
-    /// CA1827: Do not use Count() when Any() can be used.
```
mavasani: Did you mean to retain both `CA1827` and `CA1828` in your doc comment?

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.Fixer.cs
```diff
@@ -52,57 +58,72 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+        /// <param name="isAsync"><see langword="true" /> if it's an asynchronous method; <see langword="false"/> otherwise.</param>
+        /// <param name="expression">If this method returns <see langword="true" />, contains the expression to be used to invoke <c>Any</c>.</param>
+        /// <param name="arguments">If this method returns <see langword="true" />, contains the arguments from <c>Any</c> to be used on <c>Count</c>.</param>
+        /// <param name="negate">If this method returns <see langword="true" />, indicates whether to negate the expression.</param>
         /// <returns><see langword="true" /> if a fixer was found., <see langword="false" /> otherwise.</returns>
-        protected abstract bool TryGetFixer(SyntaxNode node, out SyntaxNode expression, out IEnumerable<SyntaxNode> arguments, out bool negate);
+        protected abstract bool TryGetFixer(SyntaxNode node, bool isAsync, out SyntaxNode expression, out IEnumerable<SyntaxNode> arguments, out bool negate);
 
-        private class DoNotUseCountWhenAnyCanBeUsedCodeAction : CodeAction
+        private class DoNotUseCountAsyncWhenAnyAsyncCanBeUsedCodeAction : CodeAction
```
mavasani: Given this code action is used for fixing both CA1827 and CA1828, we should probably just use a terminology that may imply it is used only for one of them. Even something like `MyCodeAction` used across the repo would be fine, or any other better neutral term.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -0,0 +1,465 @@
...
+
+                context.RegisterOperationAction(
+                    operationActionsHandler.AnalyzeBinaryOperation,
+                    OperationKind.BinaryOperator);
+            }
+
+            if (WellKnownTypes.Queryable(context.Compilation) is INamedTypeSymbol queriableType)
+            {
+                var operationActionsHandler = new OperationActionsHandler(
+                    targetType: queriableType,
```
mavasani: Typo: should be `queryableType`

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -0,0 +1,465 @@
...
+                    OperationKind.Invocation);
+
+                context.RegisterOperationAction(
+                    operationActionsHandler.AnalyzeBinaryOperation,
+                    OperationKind.BinaryOperator);
+            }
+
+            const string AsyncMethodName = "CountAsync";
+
+            if (context.Compilation.GetTypeByMetadataName("Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions") is INamedTypeSymbol entityFrameworkQueryableExtensionsType)
```
mavasani: Should the type be added to `WellKnownTypes`?

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -0,0 +1,465 @@
...
+                context.RegisterOperationAction(
+                    operationActionsHandler.AnalyzeInvocationOperation,
+                    OperationKind.Invocation);
+
+                context.RegisterOperationAction(
+                    operationActionsHandler.AnalyzeBinaryOperation,
+                    OperationKind.BinaryOperator);
+            }
+
+            if (context.Compilation.GetTypeByMetadataName("System.Data.Entity.QueryableExtensions") is INamedTypeSymbol queryableExtensionsType)
```
mavasani: Should the type be added to `WellKnownTypes`?

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -0,0 +1,465 @@
...
+                return true;
+            }
+
+            /// <summary>
+            /// Checks the <paramref name="operation"/> is an invocation of the method <see cref="targetMethod" /> in the <see cref="targetType" />.
+            /// </summary>
+            /// <param name="operation">The operation.</param>
+            /// <returns><see langword="true" /> if the <paramref name="operation"/> is an invocation of the method <see cref="targetMethod" /> in the <see cref="targetType" />; 
+            /// <see langword="false" /> otherwise.</returns>
+            private bool IsCountAsyncMethodInvocationAwaited(IOperation operation)
```
mavasani: This should be probably renamed to avoid implying it only deals with `CountAsync`, as it actually seems to deal with all possible target methods.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
paulomorgado: I thought about it, but it seemed so specific.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
paulomorgado: I thought about it, but it seemed so specific.

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.sarif
paulomorgado: I was waiting for all this to be on master to point those issues to the source code.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpDoNotUseCountAsyncWhenAnyAsyncCanBeUsed.Fixer.cs
paulomorgado: Where would that extensions go?

paulomorgado:
Added support for `LongCount` and `LongCountAsync`.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpDoNotUseCountAsyncWhenAnyAsyncCanBeUsed.Fixer.cs
mavasani: Possibly on SyntaxNodeExtensions, but you don't need to do it as part of this PR..

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -236,14 +235,14 @@ private static void OnCompilationStart(CompilationStartAnalysisContext context)
         private sealed class OperationActionsHandler
         {
             private readonly INamedTypeSymbol targetType;
```
mavasani: Please use `_` as the prefix for instance field names. I am planning to bump up the naming violations to warning level to aid here.

mavasani: This can be done in your other open PR. I will merge this one.

PR #2729, opened 2019-07-31
CA3147 handle async ASP.NET MVC controller action methods

### LINKED ISSUE #2721: CA3147 does not account for async
#### Analyzer package

Microsoft.NetFramework.Analysers

#### Package Version

v2.9.4 (Latest)

#### Diagnostic ID
CA3147

#### Repro steps
the following method should give a warning for CA3147 but it doesn't since its async. 
```
[HttpPost]
public async Task<ActionResult> Post(PostModel model) 
{
// do stuff
}
```

#### Expected behavior
warning should be displayed for async methods

#### Actual behavior
warning only works for sync methods.

dotpaul:
Fixes #2721


dotpaul:
FYI @DSIE63 @matt-desai @timhermann

dotpaul:
Thanks @nathanstocking !

chinwobble:
@dotpaul will this solution work if the method is not marked with async?

```
        [HttpPost]       
        public Task<ActionResult> DoSomethingAsync(string input)
        {
            return Task.Completed;
        }
```

dotpaul:
> @dotpaul will this solution work if the method is not marked with async?

Oops, no.  I'll make CA3147 work for that case as well.  Thanks for pointing that out!

Issue #2708, opened 2019-07-26
Assert in TaintedDataAnalysis.TaintedDataSymbolMap<>.GetInfosForType


mavasani:
1. Attempt to build the roslyn-analyzers repo with the latest analyzer debug bits built from master

**Got:**
Assert fires: https://github.com/dotnet/roslyn-analyzers/blob/c844bfb163342a5babef14e0b5bf79b8736c492a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMap.cs#L86

```
>	Microsoft.NetCore.Analyzers.dll!Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis.TaintedDataSymbolMap<Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis.SourceInfo>.GetInfosForType(Microsoft.CodeAnalysis.INamedTypeSymbol namedTypeSymbol) Line 86	C#
 	Microsoft.NetCore.Analyzers.dll!Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis.TaintedDataSymbolMapExtensions.IsSourceConstantArrayOfType(Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis.TaintedDataSymbolMap<Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis.SourceInfo> sourceSymbolMap, Microsoft.CodeAnalysis.IArrayTypeSymbol arrayTypeSymbol) Line 85	C#
 	Microsoft.NetCore.Analyzers.dll!Microsoft.NetCore.Analyzers.Security.SourceTriggeredTaintedDataAnalyzerBase.Initialize.AnonymousMethod__4(Microsoft.CodeAnalysis.Diagnostics.OperationAnalysisContext operationAnalysisContext) Line 190	C#
```

Assert seems benign as we do handle the null value in the very next statement.

dotpaul:
@LLLXXXCCC can you look at this one as well?  Also seems related to the TaintedDataAnalysis changes for constant arrays.

mavasani:
@LLLXXXCCC This should be fixed now?

LLLXXXCCC:
It is supposed to be caused by the same reason with #2707 . This problem was solved. But I'll extend it to handle hard code in jagged array. @dotpaul

[...]

PR #2730, opened 2019-07-31
Fix the bug occurring when element type of array is not INamedType.

LLLXXXCCC:


LLLXXXCCC:
#2708

dotpaul:
I thought the assert was firing for params arrays.  Were you able to repro against the roslyn-analysis repo?

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotHardCodeEncryptionKeyTests.cs
```diff
@@ -312,6 +312,72 @@ public void TestMethod(byte[] someOtherBytesForIV)
...
+    {
+    }
+
+    public void TestMethod(T1 t)
+    {
+        MethodWithArrayParameter(t);
+    }
+}");
+        }
+
```
dotpaul: Not that I'm super worried about someone putting a hardcoded key in a params byte[] array, but would we treat the params array as a tainted source if someone called `MethodWithArrayParameters<byte>(1, 2, 3)`?  If it's easy to handle, I think we should since that would count as a constant byte array.  If it's not easy to handle, we should at least put a comment somewhere that SourceInfo.TaintConstantArray doesn't handle params arrays.

LLLXXXCCC: This case can be handled. At the point of the invocation of MethodWithArrayParameters<byte>(1, 2, 3), it will trigger the callback of OperationKind.ArrayInitializer. A test case will be added.

Issue #2132, opened 2019-02-15
Consider replacing WellKnownTypes with WellKnownTypeProvider everywhere


dotpaul:
As it is now, `WellKnownTypes` just wraps `Compilation.GetTypeByMetadataName()` calls.  It doesn't seem like a terribly necessary layer of abstraction.

We could replace usages of `WellKnownTypes` with `WellKnownTypeProvider`.  Then at least we can have our own cache of `INamedTypeSymbols`.

Evangelink:
@dotpaul Is there anything left to do to consider this issue as closed?

sharwell:
I believe this was started by #2871 and finally fixed by #2932.

dotpaul:
I think we can close this particular issue, and open a new issue if needed for considering banning `Compilation.GetTypeByMetadataName`. Thanks again @Evangelink !

PR #2733, opened 2019-08-01
Adding analyzer to check SerializationBinders for properly throwing exceptions on unrecognized types

nathanstocking:
This includes two commits, affecting three files. An analyzer, DoNotOverloadSerializationBinderWithoutThrowingAnException, is added. Unit tests for this analyzer (total of four) are also included.

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
@@ -1074,4 +1074,10 @@
   <data name="JsonNetMaybeInsecureSerializerTitle" xml:space="preserve">
     <value>Ensure that JsonSerializer has a secure configuration when deserializing</value>
   </data>
+  <data name="DoNotOverloadSerializationBinderWithoutThrowingAnExceptionTitle" xml:space="preserve">
+    <value>Do not overload SerializationBinder without throwing an exception for unexpected types.</value>
+  </data>
+  <data name="DoNotOverloadSerializationBinderWithoutThrowingAnExceptionMessage" xml:space="preserve">
+    <value>The {0}'s BindToType() method might not be throwing exceptions for unexpected types. If you're using a SerializationBinder to restrict types as a mitigation against deserialization attacks, then you need to throw exceptions for unexpected types. Returning 'null' in C# or 'Nothing' in Visual Basic allows deserialization of the type using the default serialization functions, which opens the code to injection of arbitrary information, including code, by providing compromised data to the deserializer. Throwing an exception will prevent deserialization of unrecognized types.</value>
+  </data>
 </root>
--- a/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
@@ -1075,7 +1075,7 @@
     <value>Ensure that JsonSerializer has a secure configuration when deserializing</value>
   </data>
   <data name="DoNotOverloadSerializationBinderWithoutThrowingAnExceptionTitle" xml:space="preserve">
-    <value>Do not overload SerializationBinder without throwing an exception for unexpected types.</value>
+    <value>Do not overload SerializationBinder without throwing an exception for unexpected types</value>
   </data>
   <data name="DoNotOverloadSerializationBinderWithoutThrowingAnExceptionMessage" xml:space="preserve">
     <value>The {0}'s BindToType() method might not be throwing exceptions for unexpected types. If you're using a SerializationBinder to restrict types as a mitigation against deserialization attacks, then you need to throw exceptions for unexpected types. Returning 'null' in C# or 'Nothing' in Visual Basic allows deserialization of the type using the default serialization functions, which opens the code to injection of arbitrary information, including code, by providing compromised data to the deserializer. Throwing an exception will prevent deserialization of unrecognized types.</value>
--- a/src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
-using System.IO;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using Analyzer.Utilities;
+using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
@@ -21,11 +21,11 @@ namespace Microsoft.NetCore.Analyzers.Security
     /// <remarks>
     /// 
     /// </remarks>
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
     public class DoNotOverloadSerializationBinderWithoutThrowingAnException : DiagnosticAnalyzer
     {
         protected string OverloadingClassMetadataName { get; }
-        enum cacheState
+        enum CacheState
         {
             Absent,
             InProgress,
@@ -39,7 +39,7 @@ namespace Microsoft.NetCore.Analyzers.Security
             nameof(MicrosoftNetCoreAnalyzersResources.DoNotOverloadSerializationBinderWithoutThrowingAnExceptionMessage),
             DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
             helpLinkUri: null,
-            customTags: WellKnownDiagnosticTagsExtensions.Telemetry);
+            customTags: WellKnownDiagnosticTags.Telemetry);
 
         public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
             ImmutableArray.Create<DiagnosticDescriptor>(
@@ -47,33 +47,39 @@ namespace Microsoft.NetCore.Analyzers.Security
 
         public sealed override void Initialize(AnalysisContext context)
         {
-
             context.EnableConcurrentExecution();
-
             // Security analyzer - analyze and report diagnostics on generated code.
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
 
             context.RegisterCompilationStartAction(
                 (CompilationStartAnalysisContext compilationStartAnalysisContext) =>
                 {
-                    Dictionary<IMethodSymbol, cacheState> methodCache = new Dictionary<IMethodSymbol, cacheState>();
+                    Dictionary<IMethodSymbol, CacheState> methodCache = new Dictionary<IMethodSymbol, CacheState>();
+                    WellKnownTypeProvider typeProvider = WellKnownTypeProvider.GetOrCreate(compilationStartAnalysisContext.Compilation);
+                    INamedTypeSymbol SerializationBinder = null;
+                    if (!typeProvider.TryGetTypeByMetadataName("System.Runtime.Serialization.SerializationBinder", out SerializationBinder))
+                    {
+                        System.Diagnostics.Debug.Fail("Unable to find System.Runtime.Serialization.SerializationBinder threw WellKnownTypeProvider");
+                        return;
+                    }
+
                     compilationStartAnalysisContext.RegisterSymbolAction(
                         (SymbolAnalysisContext symbolAnalysisContext) =>
                         {
                             IMethodSymbol msym = (IMethodSymbol)symbolAnalysisContext.Symbol;
                             if (msym.Name == "BindToType")
                             {
-                                if (isStandardSerBinder(msym.ContainingType))
+                                if (ITypeSymbolExtensions.Inherits(msym.ContainingType, SerializationBinder))
                                 {
                                     IBlockOperation msymOps = msym.GetTopmostOperationBlock(compilationStartAnalysisContext.Compilation);
-                                    methodCache[msym] = cacheState.InProgress;
+                                    methodCache[msym] = CacheState.InProgress;
                                     if (!hasThrow(msymOps.Operations))
                                     {
                                         symbolAnalysisContext.ReportDiagnostic(
                                             Diagnostic.Create(
                                                 DoNotOverloadSerializationBinderWithoutThrowingAnExceptionRule,
                                                 msym.Locations[0],
-                                                msym.ToDisplayString(
+                                                msym.ContainingType.ToDisplayString(
                                                     SymbolDisplayFormat.MinimallyQualifiedFormat)));
                                     }
                                 }
@@ -85,50 +91,58 @@ namespace Microsoft.NetCore.Analyzers.Security
                     {
                         if (ops == null)
                         {
-                            return false; //never happens, left for comprehensiveness
+                            //doesn't happen, but just in case
+                            System.Diagnostics.Debug.Fail("hasThrow called on null value. It should be called on an IEnumerable of IOperations.");
+                            return false;
                         }
+
                         foreach (IOperation op in ops)
                         {
-                            if (op == null) { continue; }
+                            if (op == null)
+                            {
+                                continue;
+                            }
                             if (op.Kind == OperationKind.Invocation)
                             {
                                 IMethodSymbol func = ((IInvocationOperation)op).TargetMethod;
-                                cacheState methodState = cacheState.Absent;
+                                CacheState methodState = CacheState.Absent;
                                 if (!methodCache.TryGetValue(func, out methodState))
                                 {
-                                    methodState = cacheState.Absent;
+                                    methodState = CacheState.Absent;
                                 }
-                                if (methodState == cacheState.PresentTrue)
+                                if (methodState == CacheState.PresentTrue)
                                 {
                                     return true; //we've done this before
                                 }
-                                if (methodState == cacheState.PresentFalse)
+                                if (methodState == CacheState.PresentFalse)
                                 {
                                     continue; //no need to go here again
                                 }
-                                if (methodState == cacheState.InProgress)
+                                if (methodState == CacheState.InProgress)
                                 {
                                     continue; //avoid recursive functions causing a loop
                                 }
-                                methodCache[func] = cacheState.InProgress;
+                                methodCache[func] = CacheState.InProgress;
                                 IBlockOperation funcOps = func.GetTopmostOperationBlock(compilationStartAnalysisContext.Compilation);
                                 if (funcOps == null)
                                 {
                                     //function's source is in another assembly, and we cannot scan it
                                     //we will assume in this case that, even if this function throws an exception, it's probably not one we care about
-                                    methodCache[func] = cacheState.PresentFalse;
+                                    methodCache[func] = CacheState.PresentFalse;
                                     continue;
                                 }
                                 if (hasThrow(funcOps.Operations))
                                 {
-                                    methodCache[func] = cacheState.PresentTrue;
+                                    methodCache[func] = CacheState.PresentTrue;
                                     return true;
                                 }
                             }
+
                             if (op.Kind == OperationKind.Throw)
                             {
                                 return true;
                             }
+
                             if (hasThrow(op.Descendants()))
                             {
                                 return true;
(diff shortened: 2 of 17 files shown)


src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1074,4 +1074,10 @@
   <data name="JsonNetMaybeInsecureSerializerTitle" xml:space="preserve">
     <value>Ensure that JsonSerializer has a secure configuration when deserializing</value>
   </data>
+  <data name="DoNotOverloadSerializationBinderWithoutThrowingAnExceptionTitle" xml:space="preserve">
+    <value>Do not overload SerializationBinder without throwing an exception for unexpected types.</value>
```
dotpaul: >. [](start = 97, length = 1)

No punctuation in titles

Also, did files under src\Microsoft.NetCore.Analyzers\Core\xlf change?  Should push those as well.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+namespace Microsoft.NetCore.Analyzers.Security
+{
+    /// <summary>
+    /// Checks overloads of the serialization binder's BindToType method
+    /// to check whether they properly throw an exception
+    /// </summary>
+    /// <remarks>
+    /// 
+    /// </remarks>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```
dotpaul: Might as well delete empty <remarks>

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+    public class DoNotOverloadSerializationBinderWithoutThrowingAnException : DiagnosticAnalyzer
+    {
+        protected string OverloadingClassMetadataName { get; }
```
dotpaul: >OverloadingClassMetadataName [](start = 25, length = 28)

Unused?

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+            // Security analyzer - analyze and report diagnostics on generated code.
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterCompilationStartAction(
+                (CompilationStartAnalysisContext compilationStartAnalysisContext) =>
+                {
+                    Dictionary<IMethodSymbol, CacheState> methodCache = new Dictionary<IMethodSymbol, CacheState>();
+                    WellKnownTypeProvider typeProvider = WellKnownTypeProvider.GetOrCreate(compilationStartAnalysisContext.Compilation);
+                    INamedTypeSymbol SerializationBinder = null;
+                    if (!typeProvider.TryGetTypeByMetadataName("System.Runtime.Serialization.SerializationBinder", out SerializationBinder))
```
dotpaul: >"System.Runtime.Serialization.SerializationBinder" [](start = 63, length = 50)

Put this constant in WellKnownTypeNames.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                    {
+                        System.Diagnostics.Debug.Fail("Unable to find System.Runtime.Serialization.SerializationBinder threw WellKnownTypeProvider");
+                        return;
```
dotpaul: I'm not sure if this a "never happens" case.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                    }
+
+                    compilationStartAnalysisContext.RegisterSymbolAction(
+                        (SymbolAnalysisContext symbolAnalysisContext) =>
+                        {
+                            IMethodSymbol msym = (IMethodSymbol)symbolAnalysisContext.Symbol;
+                            if (msym.Name == "BindToType")
+                            {
+                                if (ITypeSymbolExtensions.Inherits(msym.ContainingType, SerializationBinder))
```
dotpaul: >ITypeSymbolExtensions.Inherits(msym.ContainingType, SerializationBinder) [](start = 36, length = 72)

Could write this invocation of an extension method as:

```csharp
msym.ContainingType.Inherits(SerializationBinder)
```

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+            context.EnableConcurrentExecution();
+            // Security analyzer - analyze and report diagnostics on generated code.
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterCompilationStartAction(
+                (CompilationStartAnalysisContext compilationStartAnalysisContext) =>
+                {
+                    Dictionary<IMethodSymbol, CacheState> methodCache = new Dictionary<IMethodSymbol, CacheState>();
+                    WellKnownTypeProvider typeProvider = WellKnownTypeProvider.GetOrCreate(compilationStartAnalysisContext.Compilation);
+                    INamedTypeSymbol SerializationBinder = null;
```
dotpaul: >SerializationBinder [](start = 37, length = 19)

Nit: start local variable names with lowercase

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                            }
+                        }
+                    , SymbolKind.Method);
+
+                    bool hasThrow(IEnumerable<Microsoft.CodeAnalysis.IOperation> ops)
+                    {
+                        if (ops == null)
+                        {
+                            //doesn't happen, but just in case
+                            System.Diagnostics.Debug.Fail("hasThrow called on null value. It should be called on an IEnumerable of IOperations.");
```
dotpaul: >System.Diagnostics. [](start = 28, length = 19)

Nit: Add a using for System.Diagnostics so you don't have to specify the full namespace here.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                        }
+
+                        foreach (IOperation op in ops)
+                        {
+                            if (op == null)
+                            {
+                                continue;
+                            }
+                            if (op.Kind == OperationKind.Invocation)
+                            {
```
dotpaul: Nit: Add blank lines to separate code blocks that aren't logically related.

All the if statements below as well.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                                IMethodSymbol func = ((IInvocationOperation)op).TargetMethod;
+                                CacheState methodState = CacheState.Absent;
```
dotpaul: >CacheState.Absent [](start = 57, length = 17)

The TryGetValue() in the next line will set the value of methodState, so you don't need to provide one here.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                                    continue;
+                                }
+                                if (hasThrow(funcOps.Operations))
+                                {
+                                    methodCache[func] = CacheState.PresentTrue;
+                                    return true;
+                                }
+                            }
+
+                            if (op.Kind == OperationKind.Throw)
```
dotpaul: This could be an else if.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+                                    methodCache[msym] = CacheState.InProgress;
+                                    if (!hasThrow(msymOps.Operations))
+                                    {
+                                        symbolAnalysisContext.ReportDiagnostic(
+                                            Diagnostic.Create(
+                                                DoNotOverloadSerializationBinderWithoutThrowingAnExceptionRule,
+                                                msym.Locations[0],
+                                                msym.ContainingType.ToDisplayString(
+                                                    SymbolDisplayFormat.MinimallyQualifiedFormat)));
+                                    }
```
dotpaul: >} [](start = 36, length = 1)

Just for completeness's sake, could have:

```csharp
else
{
    methodCache[msym] = CacheState.PresentFalse;
}
```

dotpaul:
Add some unit tests for:
- A class derived from another class named "SerializationBinder", but not the System.Runtime.Serialization one.
- Invocations like MethodA(MethodB(MethodC())).

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            // Security analyzer - analyze and report diagnostics on generated code.
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterCompilationStartAction(
+                (CompilationStartAnalysisContext compilationStartAnalysisContext) =>
+                {
+                    Dictionary<IMethodSymbol, CacheState> methodCache = new Dictionary<IMethodSymbol, CacheState>();
```
dotpaul: >Dictionary [](start = 20, length = 10)

@mavasani I think we need to worry about thread safety, right?

Could use a ConcurrentDictionary instead.

mavasani: Yes, certainly needs a Concurrent Dictionary

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
mavasani: Yes, and you probably don’t need to create a WellKnownTypeProvider here. Look at APIs in WellKnownTypes, you can probably just add a static extension method there.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
mavasani: Correct, assert should be removed.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
mavasani: +1

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
```diff
@@ -0,0 +1,158 @@
...
+
+                    compilationStartAnalysisContext.RegisterSymbolAction(
+                        (SymbolAnalysisContext symbolAnalysisContext) =>
+                        {
+                            IMethodSymbol msym = (IMethodSymbol)symbolAnalysisContext.Symbol;
+                            if (msym.Name == "BindToType")
+                            {
+                                if (ITypeSymbolExtensions.Inherits(msym.ContainingType, SerializationBinder))
+                                {
+                                    IBlockOperation msymOps = msym.GetTopmostOperationBlock(compilationStartAnalysisContext.Compilation);
```
mavasani: I looked at your logic and I believe you are trying to flag all methods named BindToType in types deriving from SerializationBinder that don’t throw an exception in its body. Is that correct? If so, a much simpler implementation would be to RegisterOperationBlockStartAction which validates method name and containing type, and if that succeeds registers (1) a nested RegisterOperationAction for OperationKind.Throw, which sets a boolean flag scoped to the OperationBlockStartAction and (2) a nested OperationBlockEndAction that reports a diagnostic if the boolean flag is not set. Let me know if you have any questions on this.

mavasani: This might be a good example: https://github.com/dotnet/roslyn-analyzers/blob/c9d6579f31fd4ad45f3d23187189657343b395a4/src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs#L54-L116

nathanstocking: That is broadly the original logic, but it has been extended after some research into typical use to also search any function called from BindToType() for throws as well. If any of them throw an exception and are in the same assembly, the test passes. This has to continue functioning for multiple layers of called functions and/or recursive functions. Would there be value in modifying the logic with extra callbacks?

mavasani: Current logic is not performant. You are going from symbol to operations, which may lead to duplicate binding. Instead any analyzer which is interested in semantic analysis of executable code blocks should try to restrict analysis to operations. Additionally, I believe this will simplify the implementation significantly to just a handful of lines - you won't need the dictionary or your recursive local function.

mavasani: To clarify - having additional analyer callbacks is not expensive, as the compiler makes the callback at the exact time it binds/compiles the code unit. However, going from symbols to operations (method bodies) will be expensive as that will create a separate semantic model for the analyzer and lead to duplicate binding for the compiler and analyzer.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
dotpaul: I prefer WellKnownTypeProvider, because that at least has its own cache.  I still think we should replace WellKnownTypes (#2132).

mavasani: Sounds reasonable.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
dotpaul: Wait, who said this should be PresentFalse?  Isn't that backwards?

---
In reply to: [309484313](https://github.com/dotnet/roslyn-analyzers/pull/2733#discussion_r309484313) [](ancestors = 309484313)

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotOverloadDeserializationBinderWithoutThrowingAnException.cs
nathanstocking: The problem there is that I have to follow calls to other functions. If I only look at operations, I can see those calls occurring by catching IInvocationOperations, but those don't automatically allow me to see their operations without using their symbols. Since some of those functions may already have been compiled, I can't even identify their symbols and start registering a bunch more callbacks unless doing this will cause the compiler to retroactively call the function that didn't exist at the original time of compilation.

dotpaul: @mavasani `DisposeMethodsShouldCallBaseClass` doesn't consider intermediate methods calling `base.Dispose()`, does it?

I guess another approach is to build a call graph for every method invocation, and dictionary of whether a given method throws an exception, and at then at the end, walk through the call graphs of all `BindToType` methods we care about and see if they throw / call another method that throws.

mavasani: > I guess another approach is to build a call graph for every method invocation, and dictionary of whether a given method throws an exception, and at then at the end, walk through the call graphs of all BindToType methods we care about and see if they throw / call another method that throws.

Yes, if you care about following the interprocedural call chain, then you should follow the approach taken by @LLLXXXCCC in her analyzers, like: https://github.com/dotnet/roslyn-analyzers/blob/c9d6579f31fd4ad45f3d23187189657343b395a4/src/Microsoft.NetCore.Analyzers/Core/Security/DoNotCallDangerousMethodsInDeserialization.cs

It would also be good to share the code and move the call graph generator and call graph walker at compilation end from that analyzer into a shared utility.

nathanstocking: I'm new to this, so perhaps I'm not understanding the performance implications properly, but as I understand it, creating such a graph requires an analysis of every function at every point in any code that is being compiled. This would seem to be very intensive from a compute and memory perspective, when there's no guarantee that they even use SerializationBinders. The current approach only starts the analysis with functions that are eligible for the diagnostic, and builds the graph retroactively. Because of the included cache, the required subset of the graph is created for future reference, but the unvisited nodes are not computed. Wouldn't the current approach be faster?

mavasani: The performance problem from your current implementation comes when you register a symbol action callback and then try to get the top level block operation for that method symbol. At that moment, the block operation that you get is using a different semantic model then what is used by the compiler for binding method bodies, which means that your analyzer will bind/compile all the method bodies one extra time. On the other hand all OperationBlock/Operation callbacks already provide you the entire operation tree with all semantic details, and the operations are generated from the semantic model that was used by the compiler during the actual compile, so it is essentially re-using the work done during regular compilation for semantic analysis.

The rule of thumb is that is if you are planning to do any semantic analysis of executable code blocks/method bodies and/or require to work on operations, you should do so by registering explicit operation callbacks with the compiler, not using our internal API that gets you operation blocks for symbol. That internal API is only meant for interprocedural analysis scenario, where we don't have any option but to do the extra binding work for the invoked method analysis.

We can discuss in person if there is still some confusion.

nathanstocking: That part makes sense. However, since there is some interprocedural work, how does the impact of rechecking those functions called by BindToType compare to registering a callback for every function in the codebase and then compiling a dictionary of every function? I'd imagine that the simple loop through operations on each function would take some processing power, and the graph would take a lot of memory depending on the size of the code. Meanwhile, this comparatively longer way only checks a few functions. I don't have good assumptions for the ratio between processing times of the two methods of getting operations.

dotpaul:
When you have time to work on this again, create a new PR to merge into the 2.9.x branch (not master).

mavasani:

mavasani:
@nathanstocking @dotpaul is this PR still meant to be kept active?

mavasani:
Closing this PR - feel free to re-open if this is still being worked upon.

PR #2734, opened 2019-08-01
Make all the analyzer utility types and CFG related extension methods…

mavasani:
… internal

Cleans up our public API surface and also avoids accidental usage of our pooled objects by FlowAnalysis utility clients (see https://github.com/dotnet/roslyn/pull/37640 which fixes a related RPS regression)


mavasani:
@genlu @jasonmalinowski Hoping to pull in the Flow utilities package with this change into Roslyn soon after this PR.

PR #2736, opened 2019-08-01
Use property instead of count method when available

paulomorgado:
Analyzer/fixer for usage of `Enumerable.Count()` on types that properties with the same semantics (`Length`, `Count`).


sharwell:
📝 The review here will be simpler once #2735 is merged.

mavasani:
> The review here will be simpler once #2735 is merged.

Done. @paulomorgado - can you please merge latest?

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1074,4 +1083,22 @@
   <data name="JsonNetMaybeInsecureSerializerTitle" xml:space="preserve">
     <value>Ensure that JsonSerializer has a secure configuration when deserializing</value>
   </data>
+  <data name="DoNotUseCountAsyncWhenAnyAsyncCanBeUsedDescription" xml:space="preserve">
+    <value>For non-empty collections, CountAsync() enumerates the entire sequence, while Any() stops at the first item or the first item that satisfies a condition.</value>
```
sharwell:
```suggestion
    <value>For non-empty collections, CountAsync() enumerates the entire sequence, while AnyAsync() stops at the first item or the first item that satisfies a condition.</value>
```

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1074,4 +1083,22 @@
...
+    <value>CountAsync() is used where AnyAsync() could be used instead to improve performance.</value>
+  </data>
+  <data name="DoNotUseCountAsyncWhenAnyAsyncCanBeUsedTitle" xml:space="preserve">
+    <value>Do not use CountAsync() when AnyAsync() can be used</value>
+  </data>
+  <data name="UsePropertyInsteadOfCountMethodWhenAvailableDescription" xml:space="preserve">
+    <value>Enumerable.Count() potentially enumerates the sequence while a Length/Count property is a direct access.</value>
+  </data>
+  <data name="UsePropertyInsteadOfCountMethodWhenAvailableMessage" xml:space="preserve">
+    <value>Use the {0} property instead of Enumerable.Count().</value>
```
sharwell:
```suggestion
    <value>Use the '{0}' property instead of Enumerable.Count().</value>
```

paulomorgado:
I think the analyzer is not quite right.

Should we explicitly go just for arrays, `ImmutableArray` and `IColection<T>` (and interfaces that extend it)?

OR should we go for any type that implicitly implements `ICollection<T>` and any type that has or implements an interface that has either a `Length` or `Count` property?

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1119,6 +1119,15 @@
...
     <value>Do not use CountAsync() or LongCountAsync() when AnyAsync() can be used</value>
   </data>
+  <data name="UsePropertyInsteadOfCountMethodWhenAvailableDescription" xml:space="preserve">
+    <value>Enumerable.Count() potentially enumerates the sequence while a Length/Count property is a direct access.</value>
+  </data>
+  <data name="UsePropertyInsteadOfCountMethodWhenAvailableMessage" xml:space="preserve">
+    <value>Use the "{0}" property instead of Enumerable.Count().</value>
+  </data>
+  <data name="UsePropertyInsteadOfCountMethodWhenAvailableTitle" xml:space="preserve">
+    <value>Use property instead of Count() when available</value>
```
mavasani: Should the title mention the names of the properties? `Use Length/Count property instead of Count() when available`

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// </summary>
+        /// <param name="context">A <see cref="CodeFixContext" /> containing context information about the diagnostics to fix.
+        /// The context must only contain diagnostics with a <see cref="Diagnostic.Id" /> included in the <see cref="CodeFixProvider.FixableDiagnosticIds" /> 
+        /// for the current provider.</param>
+        /// <returns>A <see cref="Task" /> that represents the asynchronous operation.</returns>
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var node = root.FindNode(context.Span);
+            var propertyName = context.Diagnostics[0].Properties["PropertyName"];
```
mavasani: `context.Diagnostics[0].Properties.TryGetValue(...`

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,282 @@
...
+
+            internal void AnalyzeInvocationOperation(OperationAnalysisContext context)
+            {
+                var invocationOperation = (IInvocationOperation)context.Operation;
+
+                if (GetEnumerableCountInvocationTargetType(invocationOperation) is ITypeSymbol invocationTarget &&
+                    GetReplacementProperty(invocationTarget) is string propertyName)
+                {
+                    var propertiesBuilder = ImmutableDictionary.CreateBuilder<string, string>();
+                    propertiesBuilder.Add("PropertyName", propertyName);
```
mavasani: > "PropertyName"

Define a constant field in the analyzer that is shared by the fixer.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,282 @@
...
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            if (WellKnownTypes.Enumerable(context.Compilation) is INamedTypeSymbol enumerableType)
+            {
+                var operationActionsContext = new UsePropertyInsteadOfCountMethodWhenAvailableAnalyzer.OperationActionsContext(
+                    context.Compilation,
+                    enumerableType);
+
+                var operationActionsHandler = context.Compilation.Language == LanguageNames.CSharp
+                    ? (OperationActionsHandler)new CSharpOperationActionsHandler(operationActionsContext)
```
mavasani: We do not recommend language specific handlers in the shared layer. Would it be possible to define an abstract handler which have language specific implementations in language specific projects?

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
paulomorgado: Aren't the diagnostic arguments only for the message?

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
paulomorgado: That's what it is. I was just too lazy to put it on the language-specific project. I'll change it.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
paulomorgado: Forgot that one.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
paulomorgado: just in case....

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpUsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        /// <seealso cref="Microsoft.NetCore.Analyzers.Performance.UsePropertyInsteadOfCountMethodWhenAvailableAnalyzer.OperationActionsHandler" />
+        private sealed class CSharpOperationActionsHandler : OperationActionsHandler
+        {
+            internal CSharpOperationActionsHandler(OperationActionsContext context)
+                : base(context)
+            {
+            }
+
+            protected override ITypeSymbol GetEnumerableCountInvocationTargetType(IInvocationOperation invocationOperation)
```
mavasani: I presume the language specific code is needed due to https://github.com/dotnet/roslyn/issues/23625 :(

paulomorgado: Yes!

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpDoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+            }
+
+            /// <summary>
+            /// Checks the <paramref name="operation" /> is an invocation of one of the <see cref="DoNotUseCountWhenAnyCanBeUsedAnalyzer.OperationActionsHandler.TargetMethodNames" /> in the <see cref="DoNotUseCountWhenAnyCanBeUsedAnalyzer.OperationActionsHandler.TargetType" />.
+            /// </summary>
+            /// <param name="operation">The operation.</param>
+            /// <returns>The name of the invoked method if the value of the invocation of one of the <see cref="DoNotUseCountWhenAnyCanBeUsedAnalyzer.OperationActionsHandler.TargetMethodNames" /> in the <see cref="DoNotUseCountWhenAnyCanBeUsedAnalyzer.OperationActionsHandler.TargetType" />
+            /// is being compared with 0 or 1 using <see cref="int" /> comparison operators; otherwise, <see langword="null" />.</returns>
+            protected sealed override string IsCountMethodInvocation(IOperation operation)
```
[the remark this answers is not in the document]
paulomorgado: Reverted!

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -27,20 +27,6 @@ namespace Microsoft.NetCore.Analyzers.Performance
     /// </para>
     /// <list type="table">
     /// <listheader><term>detected</term><term>fix</term></listheader>
-    /// <item><term><c> enumerable.Count() == 0               </c></term><description><c> !enumerable.Any() </c></description></item>
```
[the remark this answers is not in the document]
paulomorgado: Reverted!

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpUsePropertyInsteadOfCountMethodWhenAvailable.cs
mavasani: I think the logic here can be significantly signified as follows:

1. Base analyzer defines an abstract method `IOperation GetReceiverForExtensionMethodInvocation(IInvocationOperation`.
2. All the logic that you have in `GetEnumerableCountInvocationTargetType` overrides can be moved to the base type which uses the above helper to just find the IOperation on which to operation.
3. C# and VB language implementations should ideally then be just a single method override with a trivial one line implementation to address the core difference from https://github.com/dotnet/roslyn/issues/23625

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Formatting;
+
+namespace Microsoft.NetCore.Analyzers.Performance
+
```
mavasani: Nit: extra new line.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// Implements the <see cref="Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer" />
+    /// </summary>
+    public abstract class UsePropertyInsteadOfCountMethodWhenAvailableFixer : CodeFixProvider
+    {
+        /// <summary>
+        /// A list of diagnostic IDs that this provider can provider fixes for.
+        /// </summary>
+        /// <value>The fixable diagnostic ids.</value>
+        public override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(UsePropertyInsteadOfCountMethodWhenAvailableAnalyzer.RuleId);
+
```
mavasani: Nit: extra new line.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// The context must only contain diagnostics with a <see cref="Diagnostic.Id" /> included in the <see cref="CodeFixProvider.FixableDiagnosticIds" /> 
+        /// for the current provider.</param>
+        /// <returns>A <see cref="Task" /> that represents the asynchronous operation.</returns>
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var node = root.FindNode(context.Span);
+
+            if (node is object &&
+                context.Diagnostics[0].Properties.TryGetValue(UsePropertyInsteadOfCountMethodWhenAvailableAnalyzer.PropertyNameKey, out var propertyName) &&
```
mavasani: `Properties` can be null, so you can go a step further and use `Diagnostics[0]?.Properties`

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <see cref="Enumerable.Count{TSource}(System.Collections.Generic.IEnumerable{TSource})" /> method with a property invocation.
+        /// </summary>
+        /// <param name="invocationNode">The invocation node to get a fixer for.</param>
+        /// <param name="memberAccessNode">The member access node for the invocation node.</param>
+        /// <param name="nameNode">The name node for the invocation node.</param>
+        /// <returns><see langword="true"/> if a <paramref name="memberAccessNode" /> and <paramref name="nameNode"/> were found;
+        /// <see langword="false" /> otherwise.</returns>
+        protected abstract bool TryGetExpression(SyntaxNode invocationNode, out SyntaxNode memberAccessNode, out SyntaxNode nameNode);
+
+        private class UsePropertyInsteadOfCountMethodWhenAvailableCodeAction : CodeAction
```
mavasani: Nit: sealed

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            private readonly Document document;
```
mavasani: Please fix the naming violations to use `_` prefix for fields.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,226 @@
...
+        private static readonly DiagnosticDescriptor s_rule = new DiagnosticDescriptor(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Performance,
+            DiagnosticHelpers.DefaultDiagnosticSeverity,
+            isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultForVsixAndNuget,
+            description: s_localizableDescription,
+#pragma warning disable CA1308 // Normalize strings to uppercase
+            helpLinkUri: "https://docs.microsoft.com/visualstudio/code-quality/" + RuleId.ToLowerInvariant());
```
mavasani: Given that we are relying on this being the future URI for the help documentation, we should make this requirement explicit in the doc bugs that we open once this PR is merged.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,226 @@
...
+            private readonly Lazy<INamedTypeSymbol> _immutableArrayType;
+            private readonly Lazy<IPropertySymbol> _iCollectionCountProperty;
+            private readonly Lazy<INamedTypeSymbol> _iCollectionOfType;
+
+            public OperationActionsContext(Compilation compilation, INamedTypeSymbol enumerableType)
+            {
+                Compilation = compilation;
+                EnumerableType = enumerableType;
+                _immutableArrayType = new Lazy<INamedTypeSymbol>(() => WellKnownTypes.ImmutableArray(Compilation), true);
+                _iCollectionCountProperty = new Lazy<IPropertySymbol>(() => WellKnownTypes.ICollection(Compilation)?.GetMembers(CountPropertyName).OfType<IPropertySymbol>().Single(), true);
```
mavasani: Please switch from `Single` to `FirstOrDefault`. We have had numerous bugs from broken code or references where expected well known types/members are missing or have duplicates.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,226 @@
...
+                        propertiesBuilder.ToImmutable(),
+                        propertyName);
+
+                    context.ReportDiagnostic(diagnostic);
+                }
+            }
+
+            protected abstract ITypeSymbol GetEnumerableCountInvocationTargetType(IInvocationOperation invocationOperation);
+
+            private string GetReplacementProperty(ITypeSymbol invocationTarget)
```
mavasani: Your current approach to determine replacement property is fine. I would even be okay with us being very loose about the requirement here and have all the below checks in this method replaced with a single check that walks the entire type chain of the invocation target and return true if any type has a public int property named either `Length` or `Count`.

src/Microsoft.NetCore.Analyzers/VisualBasic/Performance/BasicUsePropertyInsteadOfCountMethodWhenAvailable.Fixer.vb
```diff
@@ -0,0 +1,54 @@
...
+        ''' <param name="memberAccessNode">The member access node for the invocation node.</param>
+        ''' <param name="nameNode">The name node for the invocation node.</param>
+        ''' <returns><see langword="true" /> if a <paramref name="memberAccessNode" /> and <paramref name="nameNode" /> were found;
+        ''' <see langword="false" /> otherwise.</returns>
+        Protected Overrides Function TryGetExpression(invocationNode As SyntaxNode, ByRef memberAccessNode As SyntaxNode, ByRef nameNode As SyntaxNode) As Boolean
+
+            Dim invocationExpression = TryCast(invocationNode, InvocationExpressionSyntax)
+
+            If invocationExpression Is Nothing Then
+
```
mavasani: Nit: some extra new lines in this file

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -0,0 +1,452 @@
...
+using System.Linq;
+public static class C
+{{
+    public static {type} GetData() => default;
+    public static int M() => GetData().Count();
+}}
+");
+
+        [Theory]
+        [InlineData("System.Collections.Generic.List(Of Integer)")]
```
mavasani: Shouldn't we be flagging this case? and the IList and ICollection cases below?

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -0,0 +1,452 @@
...
+");
+
+        [Fact]
+        public static Task CSharp_ClassShadowingICollectionOfT_NoDiagnostic()
+            => VerifyCS.VerifyAnalyzerAsync(
+                $@"using System;
+using System.Linq;
+public class T : global::System.Collections.Generic.ICollection<string>
+{{
+    public int Count => throw new NotImplementedException();
```
mavasani: Shouldn't we be flagging this case? Looking at the added unit tests here, I wouldn't mind us loosening the `GetReplacementProperty` in the analyzer to just check for the presence of a public int property named `Length` or `Count`. Personally, I don't think that will lead to any _real_ false positives, and would benefit cases such as this one being flagged.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -0,0 +1,226 @@
...
+{
+    /// <summary>
+    /// CA1829: Use property instead of <see cref="Enumerable.Count{TSource}(System.Collections.Generic.IEnumerable{TSource})"/>, when available.
+    /// Implements the <see cref="Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer" />
+    /// </summary>
+    /// <remarks>
+    /// Flags the use of <see cref="Enumerable.Count{TSource}(System.Collections.Generic.IEnumerable{TSource})"/> on types that are know to have a property with the same semantics:
+    /// <c>Length</c>, <c>Count</c>.
+    /// </remarks>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```
mavasani: This attribute is no longer needed on the abstract base type.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
paulomorgado: As per our chat, `.Properties` is never null, so this is OK!

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
paulomorgado: Will do once it's on master in order to reference the sources.

src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
paulomorgado: Replaced LINQ with `foreach`.

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
paulomorgado: Simplified the logic but, do to the substancial differences between C# and Visual Basic, I think it's better to repeat some code and avoid more virtual method invocations.

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
paulomorgado: Maybe I'm too restrictive, but I'm making sure the only the interface implementations are flagged.

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
paulomorgado: Maybe I'm too restrictive, but I'm making sure the only the interface implementations are flagged.

paulomorgado:
https://github.com/MicrosoftDocs/visualstudio-docs/issues/3923

PR #2738, opened 2019-08-01
Fix CA2000 false positive for array initializers with multiple elements

mavasani:
Escaping logic in PointsToAnalysis was not using unique escaping key for each array element initializer.

Fixes https://github.com/dotnet/roslyn/issues/37528


src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
```diff
@@ -2017,6 +2017,46 @@ End Sub
 End Class");
         }
 
+        [Fact, WorkItem(37528, "https://github.com/dotnet/roslyn/issues/37528")]
+        public void ArrayInitializer_MultipleElementsWithDisposableAssignment_NoDiagnostic()
```
mavasani: Verified the test fails before the fix.

PR #2741, opened 2019-08-02
CA5364 and CA5386 not diagnosing on &= assignments

dotpaul:


src/Microsoft.NetCore.Analyzers/Core/Security/DoNotUseDeprecatedSecurityProtocols.cs
```diff
@@ -61,47 +61,21 @@ public override void Initialize(AnalysisContext context)
...
-                            return false;
-                        }
-                    }
-
                     compilationStartAnalysisContext.RegisterOperationAction(
                         (OperationAnalysisContext operationAnalysisContext) =>
                         {
                             var fieldReferenceOperation = (IFieldReferenceOperation)operationAnalysisContext.Operation;
+
+                            // Make sure we're not inside an &= assignment like:
```
mavasani: Long term, do we want to consider doing value content dataflow analysis (assuming it handles all the involved operators, otherwise we can add support)?

dotpaul: Unless value content DFA could track bits being enabled / disabled (for people using |= assignments), I don't think it'd be worth it for these particular rules.

But now I'm thinking I should make sure I'm excluding &= assignments for ServicePointManager.SecurityProtocol.

mavasani: I am not sure we have tests, but the code seems to be there to handle binary OR and binary AND operations: https://github.com/dotnet/roslyn-analyzers/blob/558ca610c8f6d67ae091ecf3b57951a08b2b2d61/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ValueContentAnalysis/ValueContentAbstractValue.cs#L430-L442

https://github.com/dotnet/roslyn-analyzers/blob/558ca610c8f6d67ae091ecf3b57951a08b2b2d61/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ValueContentAnalysis/ValueContentAnalysis.ValueContentDataFlowOperationVisitor.cs#L196-L204

dotpaul: That only works if DFA knows the initial value, right?  If someone just has a method that only contains `ServicePointManager.SecurityProtocol |= SecurityProtocolType.Ssl3`, we'd still end up with Unknown, wouldn't we?

mavasani: Ah, I get your point...

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotUseDeprecatedSecurityProtocols.cs
```diff
@@ -193,6 +179,51 @@ public override void Initialize(AnalysisContext context)
                         },
                         OperationKind.SimpleAssignment,
                         OperationKind.CompoundAssignment);
+
+                    return;
```
LLLXXXCCC: What is it used for?

dotpaul: Just to show that lines afterwards aren't executing as part of the code block, since they're local functions.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotUseDeprecatedSecurityProtocols.cs
```diff
@@ -54,54 +55,32 @@ public override void Initialize(AnalysisContext context)
             context.RegisterCompilationStartAction(
                 (CompilationStartAnalysisContext compilationStartAnalysisContext) =>
                 {
-                    var securityProtocolTypeTypeSymbol = compilationStartAnalysisContext.Compilation.GetTypeByMetadataName(WellKnownTypeNames.SystemNetSecurityProtocolType);
```
nathanstocking: I think the plan was to either use a type provider for this or to wrap the call to Compilation.GetTypeByMetadataName in a method in WellKnownTypes. Should we do one of those?

dotpaul: That's what I'm doing, using WellKnownTypeProvider.  😁

dotpaul:
Thanks @mavasani and @nathanstocking !

PR #2754, opened 2019-08-09
Fixed style violations, added tests, fixed for different types of const values

paulomorgado:
* Fixed style violations
* Added extensions for walking up and dow parenthesis
* Added extensions for walking up and down conversions
* Added extensions for creating diagnostics with properties
* Added more tests
* Fixed for different types of const values

/cc @mavasani


src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.Fixer.cs
```diff
@@ -53,27 +53,50 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
         }
 
+
+        /// <summary>
+        /// Determines whether the specified <paramref name="value"/> is a match for <paramref name="expected"/>.
+        /// </summary>
+        /// <param name="value">The value to match.</param>
+        /// <param name="expected">The expected value.</param>
+        /// <returns><see langword="true" /> if the specified <paramref name="value"/> is a match for <paramref name="expected"/>; otherwise, <see langword="false" />.</returns>
+        protected static bool IsMatch(object value, int expected)
```
mavasani: We have some `TryConvertToUInt64` helper methods in DiagnosticHelpers. Can those be used to simplify this?

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -251,18 +258,37 @@ public OperationActionsHandler(INamedTypeSymbol targetType, ImmutableHashSet<str
...
+                var method = invocationOperation.TargetMethod;
 
-                if (invocationOperation.Arguments.Length == 1)
+                if (invocationOperation.Arguments.Length == 1 &&
+                    IsEqualsMethod(method))
                 {
-                    var method = invocationOperation.TargetMethod;
-                    if (IsInt32EqualsMethod(method) &&
-                        (IsCountEqualsZero(invocationOperation, out var methodName) || IsZeroEqualsCount(invocationOperation, out methodName)))
+                    string operation;
```
mavasani: Nit: rename local to include `String` or `Str` suffix for clarity.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -101,6 +101,13 @@ public sealed class DoNotUseCountWhenAnyCanBeUsedAnalyzer : DiagnosticAnalyzer
     {
         internal const string AsyncRuleId = "CA1828";
         internal const string SyncRuleId = "CA1827";
+        internal const string OperationKey = "Operation";
```
mavasani: Nit: Consider using nameof like `internal const string OperationKey = nameof(OperationKey);` as the actual value is not really important.

mavasani: Same for fields below, if you decide to change.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
```diff
@@ -274,13 +300,34 @@ public void AnalyzeBinaryOperation(OperationAnalysisContext context)
...
+                    else if (IsRightCountComparison(binaryOperation, out methodName, out shouldNegate))
+                    {
+                        operation = OperationBinaryRight;
+                    }
+                    else
+                    {
+                        return;
+                    }
+
+                    var propertiesBuilder = ImmutableDictionary.CreateBuilder<string, string>(StringComparer.Ordinal);
```
mavasani: We have a PooledDictionary that you might be able to use for the builder.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
paulomorgado: Would you settle for `operationName`? I don't really like to suffix identifiers other than the expected ones (`Exception`,  `Attribute`,  `EventArgs`).

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.Fixer.cs
paulomorgado: This is note used. Forgot to remove it.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
paulomorgado: What would be the benefit of that?

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpDoNotUseCountWhenAnyCanBeUsed.Fixer.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// <param name="expression">If this method returns <see langword="true" />, contains the expression to be used to invoke <c>Any</c>.</param>
+        /// <param name="arguments">If this method returns <see langword="true" />, contains the arguments from <c>Any</c> to be used on <c>Count</c>.</param>
+        /// <returns><see langword="true" /> if a fixer was found., <see langword="false" /> otherwise.</returns>
+        /// <exception cref="System.NotImplementedException"></exception>
+        protected override bool TryGetFixer(SyntaxNode node, string operation, bool isAsync, out SyntaxNode expression, out IEnumerable<SyntaxNode> arguments)
+        {
+            switch (operation)
+            {
+                case DoNotUseCountWhenAnyCanBeUsedAnalyzer.OperationEqualsInstance:
+
```
mavasani: Nit: avoid extra newline before braces for the switch cases.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
mavasani: It is pooled, but it is fine not to use it given this is not a hot code path.

src/Microsoft.NetCore.Analyzers/Core/Performance/DoNotUseCountWhenAnyCanBeUsed.cs
paulomorgado: I went with `operationKey`.

PR #2759, opened 2019-08-12
Analyzer for deprecated/hardcoded SslProtocols enum values

dotpaul:
Similar to the existing DoNotUseDeprecatedSecurityProtocols analyzer.

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
@@ -1132,16 +1132,16 @@
     <value>Older protocol versions of Transport Layer Security (TLS) are less secure than TLS 1.2 and TLS 1.3, and are more likely to have new vulnerabilities. Avoid older protocol versions to minimize risk.</value>
   </data>
   <data name="DeprecatedSslProtocolsMessage" xml:space="preserve">
-    <value>Hardcoded use of deprecated Transport Layer Security protocol version '{0}'</value>
+    <value>Transport Layer Security protocol version '{0}' is deprecated.  Use 'None' to let the Operation System choose a version.</value>
   </data>
   <data name="DeprecatedSslProtocolsTitle" xml:space="preserve">
     <value>Do not use deprecated SslProtocols values</value>
   </data>
   <data name="HardcodedSslProtocolsDescription" xml:space="preserve">
-    <value>Existing Transport Layer Security protocol verions may become deprecated if vulnerabilities are found. Avoid hardcoding SslProtocols values to keep your application secure.</value>
+    <value>Current Transport Layer Security protocol versions may become deprecated if vulnerabilities are found. Avoid hardcoding SslProtocols values to keep your application secure. Use 'None' to let the Operating System choose a version.</value>
   </data>
   <data name="HardcodedSslProtocolsMessage" xml:space="preserve">
-    <value>Avoid hardcoding SslProtocols '{0}' to ensure your application remains secure in the future</value>
+    <value>Avoid hardcoding SslProtocols '{0}' to ensure your application remains secure in the future. Use 'None' to let the Operation System choose a version.</value>
   </data>
   <data name="HardcodedSslProtocolsTitle" xml:space="preserve">
     <value>Avoid hardcoded SslProtocols values</value>
--- a/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
@@ -1128,4 +1128,22 @@
   <data name="SetHttpOnlyForHttpCookieMessage" xml:space="preserve">
     <value>HttpCookie.HttpOnly is set to false or not set at all when using an HttpCookie. Ensure security sensitive cookies are marked as HttpOnly to prevent malicious scripts from stealing the cookies</value>
   </data>
+  <data name="DeprecatedSslProtocolsDescription" xml:space="preserve">
+    <value>Older protocol versions of Transport Layer Security (TLS) are less secure than TLS 1.2 and TLS 1.3, and are more likely to have new vulnerabilities. Avoid older protocol versions to minimize risk.</value>
+  </data>
+  <data name="DeprecatedSslProtocolsMessage" xml:space="preserve">
+    <value>Hardcoded use of deprecated Transport Layer Security protocol version '{0}'</value>
+  </data>
+  <data name="DeprecatedSslProtocolsTitle" xml:space="preserve">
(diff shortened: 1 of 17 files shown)


src/Microsoft.NetCore.Analyzers/Core/Security/SslProtocolsAnalyzer.cs
```diff
@@ -0,0 +1,248 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class SslProtocolsAnalyzer : DiagnosticAnalyzer
+    {
+        internal static DiagnosticDescriptor DeprecatedRule = SecurityHelpers.CreateDiagnosticDescriptor(
+            "CA5397",
+            nameof(MicrosoftNetCoreAnalyzersResources.DeprecatedSslProtocolsTitle),
+            nameof(MicrosoftNetCoreAnalyzersResources.DeprecatedSslProtocolsMessage),
+            descriptionResourceStringName: nameof(MicrosoftNetCoreAnalyzersResources.DeprecatedSslProtocolsDescription),
+            isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
+            helpLinkUri: "https://docs.microsoft.com/visualstudio/code-quality/ca5397",
```
dotpaul: Hopefully I'll have docs up before we ship.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1128,4 +1128,22 @@
   <data name="SetHttpOnlyForHttpCookieMessage" xml:space="preserve">
     <value>HttpCookie.HttpOnly is set to false or not set at all when using an HttpCookie. Ensure security sensitive cookies are marked as HttpOnly to prevent malicious scripts from stealing the cookies</value>
   </data>
+  <data name="DeprecatedSslProtocolsDescription" xml:space="preserve">
+    <value>Older protocol versions of Transport Layer Security (TLS) are less secure than TLS 1.2 and TLS 1.3, and are more likely to have new vulnerabilities. Avoid older protocol versions to minimize risk.</value>
+  </data>
+  <data name="DeprecatedSslProtocolsMessage" xml:space="preserve">
+    <value>Hardcoded use of deprecated Transport Layer Security protocol version '{0}'</value>
```
nathanstocking: This seems quite terse. I see that there's a description, but maybe a phrasing like <value>Transport Layer Security protocol version '{0}' is hardcoded. This version is deprecated as it is insecure. Using a more modern version and refraining from hardcoding a version are preferable.</value> would be better.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1128,4 +1128,22 @@
...
+  </data>
+  <data name="DeprecatedSslProtocolsTitle" xml:space="preserve">
+    <value>Do not use deprecated SslProtocols values</value>
+  </data>
+  <data name="HardcodedSslProtocolsDescription" xml:space="preserve">
+    <value>Existing Transport Layer Security protocol verions may become deprecated if vulnerabilities are found. Avoid hardcoding SslProtocols values to keep your application secure.</value>
```
nathanstocking: Typo: replace "verions" with "versions".

src/Microsoft.NetCore.Analyzers/Core/Security/SslProtocolsAnalyzer.cs
```diff
@@ -0,0 +1,248 @@
...
+            isEnabledByDefault: false,
+            helpLinkUri: "https://docs.microsoft.com/visualstudio/code-quality/ca5398",
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        private readonly ImmutableHashSet<string> HardcodedSslProtocolsMetadataNames = ImmutableHashSet.Create(
+            StringComparer.Ordinal,
+            "Tls12",
+            "Tls13");
+
+        private const int UnsafeBits = 12 | 48 | 192 | 768;    // SslProtocols Ssl2 Ssl3 Tls10 Tls11
```
nathanstocking: Probably should be ignored, but I prefer to see binary addition in hex (0x4 | 0x8 | 0x30 | 0xc0 | 0x300) because the math is easier.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
dotpaul: How abouts:
Transport Layer Security protocol version '{0}' is deprecated.  Use 'None' to let the Operation System choose a version.

src/Microsoft.NetCore.Analyzers/Core/Security/SslProtocolsAnalyzer.cs
```diff
@@ -0,0 +1,248 @@
...
+
+        private const int HardcodedBits = 3072 | 12288;    // SslProtocols Tls12 Tls13
```
nathanstocking: Likewise: 0x400 | 0x800 | 0x3000.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
```diff
@@ -0,0 +1,548 @@
...
+        {
+            VerifyCSharp(@"
+using System;
+using System.Security.Authentication;
+
+public class ExampleClass
+{
+    public void ExampleMethod()
+    {
+        // CA5395 violation for using Tls11
```
nathanstocking: Isn't this CA5397, not CA5395?

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
```diff
@@ -0,0 +1,548 @@
...
+        {
+            VerifyCSharp(@"
+using System;
+using System.Security.Authentication;
+
+public class ExampleClass
+{
+    public void ExampleMethod()
+    {
+        // CA5396 violation
```
nathanstocking: Similar to last one, this file's comments and the analyzers file disagree on the numbering:
Analyzers: 5397 and 5398,
test file: 5395 and 5396

src/Microsoft.NetCore.Analyzers/Core/Security/SslProtocolsAnalyzer.cs
dotpaul: If they were flags that we defined, I would agree, but in this case, I'd rather just copy the values that the documentation says, which is in decimal.

dotpaul:
Thanks @nathanstocking !

PR #2763, opened 2019-08-14
Implement Semmle rule "HttpClientCheckCRLDisabled.ql".

LLLXXXCCC:


src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,218 @@
...
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.NetCore.Analyzers.Security.Helpers;
+
+namespace Microsoft.NetCore.Analyzers.Security
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    class DoNotDisableHttpClientCRLCheck : DiagnosticAnalyzer
+    {
+        internal static DiagnosticDescriptor DefinitelyDisableHttpClientCRLCheckRule = SecurityHelpers.CreateDiagnosticDescriptor(
+            "CA5397",
```
dotpaul: >CA5397 [](start = 13, length = 6)

These IDs are already used.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,218 @@
...
+            typeof(MicrosoftNetCoreAnalyzersResources),
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyDisableHttpClientCRLCheck),
```
dotpaul: Since we're down to one resource, we should probably just get rid of this overload that specifies a resource.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,218 @@
...
+                    if (!wellKnownTypeProvider.TryGetTypeByMetadataName(WellKnownTypeNames.SystemNetHttpHttpClient, out INamedTypeSymbol httpClientTypeSymbol))
+                    {
+                        return;
+                    }
+
+                    if (typeToTrackMetadataNames.Any(s => !wellKnownTypeProvider.TryGetTypeByMetadataName(s, out _)))
+                    {
+                        return;
+                    }
+
```
dotpaul: Given the handlers are (maybe?) in different assemblies, I don't think it makes sense to skip analysis if not all of them are available.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,218 @@
...
+                "CheckCertificateRevocationList",
+                (ValueContentAbstractValue valueContentAbstractValue) =>
+                    PropertySetCallbacks.EvaluateLiteralValues(
+                        valueContentAbstractValue,
+                        (object o) => o is bool booleanValue && booleanValue == false),
+                CheckCertificateRevocationListIndex));
+
+        private static readonly HazardousUsageEvaluatorCollection HazardousUsageEvaluators = new HazardousUsageEvaluatorCollection(
+            new HazardousUsageEvaluator(
+                WellKnownTypeNames.SystemNetHttpHttpClient,
```
dotpaul: >SystemNetHttpHttpClient [](start = 35, length = 23)

Not a straightforward thing to handle, but should we worry about derived classes?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysis.PropertySetDataFlowOperationVisitor.cs
```diff
@@ -149,62 +156,72 @@ protected override void SetAbstractValueForTupleElementAssignment(AnalysisEntity
...
                     }
+
+                    PointsToAbstractValue pointsToAbstractValue = this.GetPointsToAbstractValue(operation);
+                    this.SetAbstractValue(pointsToAbstractValue, abstractValue);
                 }
                 else
                 {
-                    Debug.Fail("Unhandled ConstructorMapper");
-                    return abstractValue;
+                    if (TryFindNonTrackedTypeHazardousUsageEvaluator(operation.Constructor, operation, operation.Arguments, out var hazardousUsageEvaluator, out var propertySetInstance))
```
dotpaul: >var [](start = 128, length = 3)

Nit: explicit types in this file

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysis.PropertySetDataFlowOperationVisitor.cs
```diff
@@ -489,37 +508,46 @@ public override PropertySetAbstractValue VisitInvocation_NonLambdaOrDelegateOrLo
...
 
                 return baseValue;
+            }
 
-                // Local functions.
-                bool TryFindNonTrackedTypeHazardousUsageEvaluator(out HazardousUsageEvaluator evaluator, out IOperation instance)
+            /// <summary>
+            /// Find the hazardous usage evaluator for the non tracked type.
+            /// </summary>
+            /// <param name="method">The petential hazardous method.</param>
```
dotpaul: >petential [](start = 41, length = 9)

Nit: potential

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysis.cs
```diff
@@ -154,6 +154,42 @@ private PropertySetAnalysis(PropertySetAnalysisDomain analysisDomain, PropertySe
...
+        Compilation compilation,
+        IEnumerable<(IOperation Operation, ISymbol ContainingSymbol)> rootOperationsNeedingAnalysis,
+        AnalyzerOptions analyzerOptions,
+        ImmutableHashSet<string> typeToTrackMetadataNames,
+        ConstructorMapper constructorMapper,
+        PropertyMapperCollection propertyMappers,
+        HazardousUsageEvaluatorCollection hazardousUsageEvaluators,
+        InterproceduralAnalysisConfiguration interproceduralAnalysisConfig,
+        bool pessimisticAnalysis = false)
         {
```
dotpaul: Is this indented for you?

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,218 @@
...
+                                        (IObjectCreationOperation)operationAnalysisContext.Operation;
+
+                                    if (httpClientTypeSymbol.Equals(objectCreationOperation.Type))
+                                    {
+                                        if (objectCreationOperation.Arguments.Length == 0)
+                                        {
+                                            operationAnalysisContext.ReportDiagnostic(
+                                                objectCreationOperation.CreateDiagnostic(
+                                                    DefinitelyDisableHttpClientCRLCheckRule));
+                                        }
```
dotpaul: In this case at least, setting ServicePointManager.CheckCertificateRevocationList seems to check the CRL.

Would be good to find out for other cases, where a handler is provided.

LLLXXXCCC: Are we targeting .Net Core? Refer to [this](https://github.com/dotnet/corefx/issues/29452), ServicePointManager won't have impact on HttpClient.

dotpaul: Sometimes?  🙂  There's a helper method somewhere for determining whether the code is targeting .NET Framework or .NET Core.  I was wondering if it'd be worth making this particular case a separate rule or have an .editorconfig option to not flag it.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
LLLXXXCCC: It's hard to say. Is it possible that people only want to track this class not its derived classes? Need to handle all these cases?

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
dotpaul: Do you already have another commit that needs to be pushed?

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,213 @@
...
+        internal static DiagnosticDescriptor DefinitelyDisableHttpClientCRLCheckRule = SecurityHelpers.CreateDiagnosticDescriptor(
+            "CA5399",
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyDisableHttpClientCRLCheck),
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyDisableHttpClientCRLCheckMessage),
+            DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
+            helpLinkUri: null,
+            descriptionResourceStringName: nameof(MicrosoftNetCoreAnalyzersResources.DoNotDisableHttpClientCRLCheckDescription),
+            customTags: WellKnownDiagnosticTagsExtensions.DataflowAndTelemetry);
+        internal static DiagnosticDescriptor MaybeDisableHttpClientCRLCheckRule = SecurityHelpers.CreateDiagnosticDescriptor(
+            "CA5400",
```
dotpaul: >CA5400 [](start = 13, length = 6)

Huh, you didn't have to update `src\Utilities\Compiler\DiagnosticCategoryAndIdRanges.txt` for this?

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
```diff
@@ -7,7 +7,7 @@ Design: CA1000-CA1099
 Globalization: CA1300-CA1399, CA2101
 Mobility: CA1600-CA1699
 Performance: CA1800-CA1899, HA
-Security: CA2100-CA2199, CA2300-CA2399, CA3000-CA3147, CA5300-CA5399
+Security: CA2100-CA2199, CA2300-CA2399, CA3000-CA3147, CA5300-CA5420
```
dotpaul: >CA5420 [](start = 62, length = 6)

Let's take them all!  5499

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,213 @@
...
+                    PropertySetCallbacks.EvaluateLiteralValues(
+                        valueContentAbstractValue,
+                        (object o) => o is bool booleanValue && booleanValue == false),
+                CheckCertificateRevocationListIndex));
+
+        private static readonly HazardousUsageEvaluatorCollection HazardousUsageEvaluators = new HazardousUsageEvaluatorCollection(
+            new HazardousUsageEvaluator(
+                WellKnownTypeNames.SystemNetHttpHttpClient,
+                ".ctor",
+                "handler",
```
dotpaul: >handler [](start = 17, length = 7)

If we do handle derived classes, the parameter name might not be the same as HttpClient's constructor's.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -0,0 +1,213 @@
...
+namespace Microsoft.NetCore.Analyzers.Security
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    class DoNotDisableHttpClientCRLCheck : DiagnosticAnalyzer
+    {
+        internal static DiagnosticDescriptor DefinitelyDisableHttpClientCRLCheckRule = SecurityHelpers.CreateDiagnosticDescriptor(
+            "CA5399",
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyDisableHttpClientCRLCheck),
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyDisableHttpClientCRLCheckMessage),
+            DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
```
dotpaul: >EnabledByDefaultIfNotBuildingVSIX [](start = 30, length = 33)

Let's first ship with these rules disabled by default.

dotpaul:
Looking at https://github.com/dotnet/corefx/blob/30decee8aa72ce93eec0394e94f05a86cec7ab9d/src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpHandler.cs#L1150 and https://github.com/dotnet/corefx/blob/d3911035f2ba3eb5c44310342cc1d654e42aa316/src/System.Net.Http/src/System/Net/Http/CurlHandler/CurlHandler.SslProvider.Linux.cs#L357, I'm wondering if having a [HttpClientHandler.ServerCertificateCustomValidationCallback](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=netframework-4.8) essentially ignores CheckCertificateRevocationList (which would make sense to me).  In that case, we'd also want to track ServerCertificateCustomValidationCallback and not warn if that's set.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysis.cs
dotpaul: The parameters still don't look indented to me.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotDisableHttpClientCRLCheck.cs
LLLXXXCCC: Yes. For these specific cases, people can add a new HazardousUsageEvaluator for that derived classes.

PR #2764, opened 2019-08-14
Enforce that symbol comparison should use an equality comparer

ryzngard:
With the introduction of nullable reference types, comparison of symbols has become more complex. The recommended way is to use an EqualityComparer. This adds a check for the `Equals` method where all parameters are `ISymbol` types. If the method invocation is from an instance, the instance is also checked to be `ISymbol`. 

Also updated the text to reflect new recommended comparison and resolution.


sharwell:
❓ Have you verified that this diagnostic does not suggest using an EqualityComparer if the user is building against Roslyn 1.x or 2.x? For those releases, `Equals` is the expected approach.

ryzngard:
> ❓ Have you verified that this diagnostic does not suggest using an EqualityComparer if the user is building against Roslyn 1.x or 2.x? For those releases, `Equals` is the expected approach.

I have not yet. I also may have misinterpreted the recommendation from @jasonmalinowski . 

Is there a good way to version lock? Can we check for the existence of the symbol wrapped type?

mavasani:
> Is there a good way to version lock? Can we check for the existence of the wrapped type?

Is there a specific equality comparer type in Roslyn that was added in 3.x? If so, you can fetch this named type symbol at compilation start and keep your logic guarded around the presence of that type symbol.

ryzngard:
```C#
private const string s_symbolEqualityComparerName = typeof(SymbolEqualityComparer).FullName;
...

comparerType = compilation.GetTypeByMetadataName(s_symbolEqualityComparerName);

```

Is that the correct way? 

Currently it doesn't look like the version of Roslyn being referenced by the analyzers is up to date enough to have the comparer.

mavasani:
@ryzngard Master is currently targeting 2.9.0 version of MS.CA so that the analyzers can execute on Dev15 toolset. You may want to consider retargeting your PR to 3.x branch and move the version of referenced MS.CA in that branch to your required Roslyn version. You would need to change the following: https://github.com/dotnet/roslyn-analyzers/blob/ddb105b54eff74f7c328cc45670f3e10a8ea20db/eng/Versions.props#L21-L22 and also update the VersionPrefix so the major and minor versions for the analyzer package match that of the consumed Roslyn version: https://github.com/dotnet/roslyn-analyzers/blob/ddb105b54eff74f7c328cc45670f3e10a8ea20db/eng/Versions.props#L5

mavasani:
Actually, stepping back. If you don't need to invoke any new Roslyn API in 3.x version and are just looking for a type symbol for `SymbolEqualityComparer`, then you don't need to move to newer Roslyn bits. You can just hardcode the fully qualified type name for `SymbolEqualityComparer` in the call to `GetTypeByMetadataName`

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -339,7 +339,7 @@
     <value>Upgrade MSBuildWorkspace</value>
   </data>
   <data name="CompareSymbolsCorrectlyDescription" xml:space="preserve">
-    <value>Symbols should be compared for equality, not identity.</value>
+    <value>Symbols should be compared for equality with an EqualityComparer.</value>
```
mavasani: I think we should revert the message given `EqualityComparer` is only available on most recent Roslyn bits. Instead, we should just update the code fix provider associated with this diagnostic to use the EqualityComparer when the comparer is available.

ryzngard:
@sharwell with the updated code I'm having trouble with tests.  Is there an assembly reference I should add? I would have thought `Microsoft.CodeAnalysis` was already referenced

```
Diagnostics:
    // Test0.cs(5,29): error CS0234: The type or namespace name 'SymbolEqualityComparer' does not exist in the namespace 'Microsoft.CodeAnalysis' (are you missing an assembly reference?)
```

mavasani:
> I would have thought Microsoft.CodeAnalysis was already referenced

I believe the tests would also reference the older (2.9.x) version of code analysis. You should extend your test code to define the required equality comparer in test source code, we do so in many tests. For example, see https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/UpgradeMSBuildWorkspaceAnalyzerTests.cs#L20

ryzngard:
@mavasani  ah, thanks. For some reason I thought the public comparer was older but was added at the same time. I'll update the test

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -90,19 +113,27 @@ private void HandleBinaryOperator(in OperationAnalysisContext context, INamedTyp
...
-                }
+                return;
+            }
 
-                if (operation.Type.AllInterfaces.Contains(symbolType))
-                {
-                    return true;
-                }
+            var parameters = methodReference.Method.Parameters;
+            if (methodReference.Method.Name == s_symbolEqualsName && parameters.All(p => IsSymbolType(p.Type, symbolType)))
```
mavasani: Consider moving the name compare check to the start of the method:

```csharp
if (methodReference.Method.Name != s_symbolEqualsName)
{
   return;
}
```

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
```diff
@@ -228,6 +306,35 @@ class TestClass {{
...
+        [Theory]
+        [InlineData(nameof(ISymbol))]
+        [InlineData(nameof(INamedTypeSymbol))]
+        public async Task CompareSymbolImplementationWithInterface_NoComparer_CSharp(string symbolType)
+        {
+            var source = $@"
+using Microsoft.CodeAnalysis;
+class TestClass {{
+    bool Method(Symbol x, {symbolType} y) {{
+        return [|x == y|];
```
mavasani: Do we have a unit test where input source has `Equals(x, y);`, symbol equality comparer is available and we offer a diagnostic/fix to change it to `SymbolEqualityComparer.Default.Equals(x, y);`?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -90,19 +113,27 @@ private void HandleBinaryOperator(in OperationAnalysisContext context, INamedTyp
             context.ReportDiagnostic(binary.Syntax.GetLocation().CreateDiagnostic(Rule));
         }
 
-        private static bool IsSymbolType(IOperation operation, INamedTypeSymbol symbolType)
+        private static void HandleMethodReferenceOperation(OperationAnalysisContext context, INamedTypeSymbol symbolType)
```
mavasani: Are you sure this callback is being made in presence of invocations? We have https://github.com/dotnet/roslyn/issues/26206 which tracks that `IInvocationOperation` does not expose the underlying `IMethodReferenceOperation` in the operation tree. If you are getting this callback for invocations, then that bug might be fixed, otherwise you would need to registered for `OperationKind.Invocation` instead.

ryzngard: I thought I was, but it's been a bit since I ran this. Testing again now...

ryzngard: Changed to use IInvocationOperation

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -90,19 +112,32 @@ private void HandleBinaryOperator(in OperationAnalysisContext context, INamedTyp
...
-                if (operation.Type.AllInterfaces.Contains(symbolType))
-                {
-                    return true;
-                }
+            if (invocationOperation.Instance != null && !IsSymbolType(invocationOperation.Instance, symbolType))
+            {
+                return;
+            }
+
+            var parameters = invocationOperation.Arguments;
```
mavasani: Nit: rename to `arguments`

PR #2766, opened 2019-08-15
Update analyzer documentation files in the repo

mavasani:


src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.sarif
```diff
@@ -3645,6 +3645,63 @@
...
+          "defaultLevel": "warning",
+          "properties": {
+            "category": "Security",
+            "isEnabledByDefault": false,
+            "typeName": "SetHttpOnlyForHttpCookie",
+            "languages": [
+              "C#",
+              "Visual Basic"
+            ],
+            "tags": [
```
mavasani: @dotpaul @LLLXXXCCC - can you confirm the tags are correct for each of the added rules? Also do we respect https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#excluded-symbol-names for the DFA rule? If so, we should update that list of configurable rules.

mavasani:
Closing in favor of https://github.com/dotnet/roslyn-analyzers/pull/2771

PR #2775, opened 2019-08-21
ReleaseNotesUtil: Handling changing enabled by default rules, slightly less hardcoding

dotpaul:


src/ReleaseNotesUtil/Program.cs
```diff
@@ -88,32 +86,55 @@ private static void GetRulesJson(string nugetInstalledPackagesPath, string versi
             IEnumerable<RuleInfo> removedRules =
                 oldContent.Rules
                     .Where(r => !newRulesById.ContainsKey(r.Id));
+            IEnumerable<RuleInfo> changedRules =
+                newContent.Rules
+                    .Where(r => oldRulesById.TryGetValue(r.Id, out RuleInfo oldRule)
+                                && r.IsEnabledByDefault != oldRule.IsEnabledByDefault);
```
mavasani: Should we compare any other rule metadata? DiagnosticDescriptor itself seems to equatable: http://source.roslyn.io/#Microsoft.CodeAnalysis/Diagnostic/DiagnosticDescriptor.cs,14

dotpaul: I figured enabled-by-default would be the only important thing people would care about, unless you think people want to see fixers or something?

I didn't think it'd be all that important to note changes to descriptions or tags or what not.

mavasani: Sounds reasonable!

dotpaul:
Thanks @mavasani !

PR #2779, opened 2019-08-22
Handle IOperation tree shape change for IThrowOperation

### LINKED ISSUE #2761: CA1065 raised in property getters
#### Analyzer package

#### Package Version

Example: v2.9.4 (Latest)

#### Diagnostic ID

[CA1065](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1065-do-not-raise-exceptions-in-unexpected-locations)

#### Repro steps

With .Net Core 3 Preview 8 and Visual Studio 16.3 Preview 2 Community installed I get errors whenever a NotSupportedException or InvalidOperationException is raised in a property getter (the same code did not produce warning with Preview 7, 16.3 Preview 1), e.g.:
````
    public class ProxyStream : Stream
    {
        ...
        public override bool CanSeek => throw new NotSupportedException();
        ...
    }
````

#### Expected behavior

No warning.

#### Actual behavior

````
warning CA1065: get_CanSeek creates an exception of type Exception, an exception type that should not be raised in a property. If this exception instance might be raised, use a different exception type, convert this property into a method, or change this property's logic so that it no longer raises an exception.
````

mavasani:
Fixes #2761
IOperation tree for IThrowOperation now has an implicit IConversionOperation wrapping the underlying thrown object. Details at https://github.com/dotnet/roslyn/blame/a5418f7398495db4ea858f0c9efd9b13541ff478/src/Compilers/CSharp/Portable/Binder/Binder_Statements.cs#L395

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotRaiseExceptionsInUnexpectedLocations.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotRaiseExceptionsInUnexpectedLocations.cs
@@ -96,7 +96,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     operationBlockContext.RegisterOperationAction(operationContext =>
                     {
                         // Get ThrowOperation's ExceptionType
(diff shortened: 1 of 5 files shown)


mavasani:
Thanks @dotpaul

PR #2800, opened 2019-08-31
Fix an indeterministic test failure

mavasani:
AbstractDataFlowAnalysisContext hash code computation did not account for the whole CFG, just the original operation of the CFG. This bug was masked by our prior CWT based caching/leak that ensured that the static cache on IOperationExtensions from IOperation to CFG always returned the same CFG. With the new BoundedCache approach, it is more likely for the cache entries to get evicted under high memory pressure and multiple simultaneous compilations, which makes this failure lot more prominent, even though non-deterministic.

Unit test `Test_PassTaintedSourceInfoAsParameter_SinkProperties_Interprocedual_Diagnostic` fails intermittently prior to this fix. I updated the test to execute multiple instances in parallel and it fails consistently prior to the fix. It passed consistently multiple times after this fix.

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotHardCodeEncryptionKeyTests.cs
+++ b/src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotHardCodeEncryptionKeyTests.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Test.Utilities;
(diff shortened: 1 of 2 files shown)


src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotHardCodeEncryptionKeyTests.cs
```diff
@@ -272,7 +273,9 @@ public void CreateEncryptor(byte[] rgbKey, byte[] rgbIV)
         [Fact]
         public void Test_PassTaintedSourceInfoAsParameter_SinkProperties_Interprocedual_Diagnostic()
         {
-            VerifyCSharpWithDependencies(@"
+            Parallel.For(1, 100, _ =>
```
mavasani: I can revert the unit test changes if we think that is more appropriate, especially given the newly added asserts will fire if the underlying issue crops up again. Though having it does help us ensure we have a unit test validating our flow analysis framework under large number of simultaneous compilations.

mavasani:
Thanks @dotpaul

PR #2808, opened 2019-09-04
Add few requested configuration options

### LINKED ISSUE #1375: CA1801 (Review unused parameters) needs an option to ignore effectively public methods
#### Analyzer package

#### Analyzer

ReviewUnusedParametersAnalyzer

#### Suggestion

We need some way to configure the analyzer to ignore effectively public methods - lot of clients would not want their shipped public API surface to change/be flagged (for example, Roslyn itself). This is not going to be trivial until we implement analyzer options API. So, meanwhile we can consider having 2 different code fixes/actions for this diagnostic for a user friendly FixAll experience:

1. Works on entire API surface
2. Works just on APIs which are not effectively public

I think we need this to ensure the code fix for rule is useful.

### LINKED ISSUE #2691: CA1062: Validate arguments of public methods - when using custom extension method
Visual Studio 2019 (16.1.6) with Microsoft.CodeAnalysis.FxCopAnalyzers 2.9.3.

We use an extension method that helps keep our methods clean and simple but this results in a "CA1062: Validate arguments of public methods" warning. This isn't ideal as we still want the warning if we don't call our extension method (`CheckArgumentNull`).

Extension method:
```
        public static void CheckArgumentNull(this object value, string paramName)
        {
            if (value == null)
        }
```
Allows us to check parameters like this:
          `  paramName.CheckArgumentNull(nameof(paramName));`

...rather than the more wordy:
```
            if (paramName == null)
                throw new ArgumentNullException(nameof(paramName));
```

Some thoughts on this:
1. Perhaps the code analysis can look into extension method for a check.
2. Perhaps we can decorate the extension method with something like AttributeUsage to indicate the CA1062 is handled and can be ignored.

### LINKED ISSUE #2713: New rule: Do not catch NullReferenceException type
NullReferenceException should be fixed and not caught. In most situations it should be possible to convert/replace the try-catch-block into a null check. Any code from the catch block could be moved to the "is null" branch of an if-statement.

### LINKED ISSUE #2772: CA1802 could also flag some instance fields
**Version Used**: 
3.3.0-beta3-19409-05+ed92d532473db83c1db47b313ee1c1bd7520aa08

**Steps to Reproduce**:
```C#
using System;

class Program
{
    static void Main()
    {
        foreach (string s in new Splitter().Split("hello,world"))
            Console.WriteLine(s);
    }
}

class Splitter
{
    private readonly char _separator = ',';

    public string[] Split(string s) => s.Split(_separator);
}
```

**Expected Behavior**:
CA1802 would suggest refactoring `_separator` to be `const`.

**Actual Behavior**:
CA1802 ignores `_separator` because it's an instance rather than static field.

### LINKED ISSUE #2799: CA2241 - Possibility to support custom string formatting methods?
#### Analyzer package

Microsoft.NetCore.Analyzers

#### Package Version

v2.9.4

#### Diagnostic ID

[CA2241](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2241-provide-correct-arguments-to-formatting-methods)

#### Repro steps

Have a custom method that accepts the typical format string + params args.

IE:
public static string Log(string format, params object[] args)
{
   ...
}

#### Expected behavior

Log("{1}", 2) triggers CA2241

#### Actual behavior

CA2241 is hardcoded to look at Console.Write(Line) and string.Format, so no error.

I assume adding this would be a breaking change WRT backwards compatibility, but could this be made configurable?

mavasani:
Recommended to review commit by commit.

1. https://github.com/dotnet/roslyn-analyzers/commit/40cd30186631a3e5d33492b37f5c69b0aa59876c: Add a configurable option `disallowed_symbol_names` to CA1031 to allow configuration of disallowed exception types. Fixes #2713
2. https://github.com/dotnet/roslyn-analyzers/commit/7807234e18b3c7797506b26bbd4596955f6dcc09: Add a new configurable option 'additional_string_formatting_methods' for CA2241. [CA2241](https://docs.microsoft.com/visualstudio/code-quality/ca2241-provide-correct-arguments-to-formatting-methods) considers String.Format and Console.WriteLine as special formatting methods to validate the callsites for formatting arguments. This change allows end users to supply additional formatting methods that should be validated. Fixes #2799
3. https://github.com/dotnet/roslyn-analyzers/commit/af63bad9c0da2c43c2814e4f2ec54351e378bde4: Add configuration option 'required_modifiers' for CA1802. [CA1802](https://docs.microsoft.com/visualstudio/code-quality/ca1802-use-literals-where-appropriate) only flags static readonly flags by default. This matches the legacy FxCop implementation for compat reasons. With this change, users can make the rule more aggressive by also flagging instance fields. Fixes #2772
4. https://github.com/dotnet/roslyn-analyzers/commit/a557ba71bc079bb9f1c60ec3ccb1ccbbc9ca22fa: Add configurable option `exclude_extension_method_this_parameter` for [CA1062](https://docs.microsoft.com/visualstudio/code-quality/ca1062-validate-arguments-of-public-methods) to skip analysis of extension method 'this' parameter. Fixes #2691
5. https://github.com/dotnet/roslyn-analyzers/pull/2808/commits/b6d9409c77d0b0616c9c13e776b63ec48b14b46e: Respect existing option `api_surface` for CA1802 (Review unused parameters). Fixes #1375


docs/Analyzer Configuration.md
```diff
@@ -66,6 +66,31 @@ Default Value: All output kinds
 
 Example: `dotnet_code_quality.CA2007.output_kind = ConsoleApplication, DynamicallyLinkedLibrary`
 
+### Required modifiers for analyzed APIs
```
mavasani: @gewarren Would you be able to do a quick review of the new option names? I will soon file a tracking bug in the docs repo for all the new options since https://docs.microsoft.com/en-us/visualstudio/code-quality/fxcop-analyzer-options?view=vs-2019 was last updated. Thanks!

gewarren: @mavasani They look good to me!

mavasani: Thanks!

mavasani:
@dotpaul @genlu can you do a quick review? We would like to release a 2.9.5-beta1 given the recent performance fixes and I would like to get this in if possible.

mavasani:
Thank you @dotpaul

PR #2819, opened 2019-09-05
Update checked in analyzer documentation files in the repo

mavasani:


src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md
```diff
@@ -3,50 +3,50 @@ Sr. No. | Rule ID | Title | Category | Enabled | CodeFix | Description |
...
-3 | [CA1003](https://docs.microsoft.com/visualstudio/code-quality/ca1003-use-generic-event-handler-instances) | Use generic event handler instances | Design | False | True | A type contains an event that declares an EventHandler delegate that returns void, whose signature contains two parameters (the first an object and the second a type that is assignable to EventArgs), and the containing assembly targets Microsoft .NET Framework?2.0. |
+3 | [CA1003](https://docs.microsoft.com/visualstudio/code-quality/ca1003-use-generic-event-handler-instances) | Use generic event handler instances | Design | False | False | A type contains an event that declares an EventHandler delegate that returns void, whose signature contains two parameters (the first an object and the second a type that is assignable to EventArgs), and the containing assembly targets Microsoft .NET Framework?2.0. |
```
sharwell: :bulb: This would be easier to glance over if False was changed to :x: and True was changed to ✔

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidEmptyInterfaces.Fixer.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
     /// </summary>
     public abstract class AvoidEmptyInterfacesFixer : CodeFixProvider
     {
-        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(AvoidEmptyInterfacesAnalyzer.RuleId);
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray<string>.Empty;
```
sharwell: :\ These are going to be easy to mess up if/when the fixers are implemented. I would _strongly_ prefer to update the analysis to recognize the stub implementation in `RegisterCodeFixesAsync` (returning a completed task) and use that as the trigger for True/False.

sharwell:
Also it would be nice to change this to be a build artifact with client-side rendering like the pages here. There are a few bugs in the analysis code, but the strategy as a whole is quite robust with no need to ever check in updated files.

https://dotnetanalyzers.github.io/StyleCopAnalyzers/
https://dotnetanalyzers.github.io/DocumentationAnalyzers/

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidEmptyInterfaces.Fixer.cs
mavasani: The [current approach](https://github.com/dotnet/roslyn-analyzers/blob/master/src/GenerateAnalyzerRulesets/Program.cs#L448) to generate the documentation files does not to source analysis. It loads the analyzer assemblies via our AnalyzerFileReference API and then uses the runtime object to query supported diagnostics (analyzer) and fixable diagnostics (fixer). It is unlikely that this will regress because the changed files were all auto-generated stubs by a tool, and I don't think we will change these again unless we are actually implementing a fixer.

mavasani:
> Also it would be nice to change this to be a build artifact with client-side rendering like the pages here. There are a few bugs in the analysis code, but the strategy as a whole is quite robust with no need to ever check in updated files.

This documentation files are auto-generated build artifacts by just executing CIBuild.cmd, I did not update these manually. Once https://github.com/dotnet/roslyn-tools/issues/529 is implemented, these PRs would get auto-generated just like merge PRs. Additionally, the MD and SARIF format documentation files are also embedded in the shipped NuGet packages. Theoretically, no one needs to review the changes to these documentation files, unless you end up finding that the documentation page has wrong information. The checked in versions in the repo are just for quick reference to know the latest state of rules in the repo, and it seems much more reliable and discoverable to have them in the repo and not rely on (manual?) publishing to external sites. We used to have a legacy tool that use to generate a similar page as StyleCop's, and it ran into quite a few bugs and needed regular fixes and also kept forgetting running the manual publishing step.

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md
mavasani: These are auto-generated/updated files, we don't expect reviewing changes to these files.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidEmptyInterfaces.Fixer.cs
dotpaul: Would be neat to come up with something, but this seems fine to get the docs updated before shipping 2.9.5-beta1.

mavasani: Yep, we can have a separate discussion on whether we need a better approach here. I am going to retain the current approach for this PR.

mavasani:
Confirmed offline with Sam that we can discuss alternate approaches to the concerns raised here in a separate issue. Merging this one..

PR #2820, opened 2019-09-05
Fixing an InvalidCastException

dotpaul:
error AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.DoNotUseInsecureSettingsForJsonNet' threw an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Symbols.SourceMemberFieldSymbolFromDeclarator' to type 'Microsoft.CodeAnalysis.IMethodSymbol'.'.

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
@@ -1969,7 +1969,7 @@ namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow
             }
 
             // Check if we are already at the maximum allowed interprocedural call chain length.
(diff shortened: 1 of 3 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -1969,7 +1969,7 @@ protected bool TryGetInterproceduralAnalysisResult(IOperation operation, out TAn
             }
 
             // Check if we are already at the maximum allowed interprocedural call chain length.
-            int currentMethodCallCount = currentMethodsBeingAnalyzed.Where(m => !((IMethodSymbol)m.OwningSymbol).IsLambdaOrLocalFunctionOrDelegate()).Count();
+            int currentMethodCallCount = currentMethodsBeingAnalyzed.Where(m => !(m.OwningSymbol is IMethodSymbol ms && ms.IsLambdaOrLocalFunctionOrDelegate())).Count();
```
dotpaul: OwningSymbol is no longer just an IMethodSymbol, since DFA is also looking at field / property initializers.  Since we're just excluding lambdas / locals / delegates from the count, I'm hoping this is okay.

PR #2822, opened 2019-09-06
Implement Semmle rule "SymmetricEncryptionNonDefaultInitializationVector.ql".

LLLXXXCCC:


dotpaul:
Tagging @gewarren for .resx file changes (no rush)

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotUseCreateEncryptorWithNonDefaultIV.cs
```diff
@@ -0,0 +1,205 @@
...
+namespace Microsoft.NetCore.Analyzers.Security
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotUseCreateEncryptorWithNonDefaultIV : DiagnosticAnalyzer
+    {
+        internal static DiagnosticDescriptor DefinitelyUseCreateEncryptorWithNonDefaultIVRule = SecurityHelpers.CreateDiagnosticDescriptor(
+            "CA5401",
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyUseCreateEncryptorWithNonDefaultIV),
+            nameof(MicrosoftNetCoreAnalyzersResources.DefinitelyUseCreateEncryptorWithNonDefaultIVMessage),
+            DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
```
dotpaul: >EnabledByDefaultIfNotBuildingVSIX [](start = 30, length = 33)

Let's disable DFA-based rules by default.

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotUseCreateEncryptorWithNonDefaultIV.cs
```diff
@@ -0,0 +1,205 @@
...
+                    return PropertySetAbstractValueKind.Flagged;
+                }));
+
+        private static readonly HazardousUsageEvaluatorCollection HazardousUsageEvaluators = new HazardousUsageEvaluatorCollection(
+            new HazardousUsageEvaluator(
+                "CreateEncryptor",
+                (IMethodSymbol methodSymbol, PropertySetAbstractValue abstractValue) =>
+                {
+                    if (methodSymbol.Parameters.Length != 0)
+                    {
```
dotpaul: Nit: Add a comment why this case isn't evaluated here.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1179,4 +1179,19 @@
   <data name="MaybeDisableHttpClientCRLCheckMessage" xml:space="preserve">
     <value>HttpClient may be created without enabling CheckCertificateRevocationList</value>
   </data>
+  <data name="DefinitelyUseCreateEncryptorWithNonDefaultIV" xml:space="preserve">
+    <value>Do not use CreateEncryptor with non default IV</value>
```
gewarren:
```suggestion
    <value>Do not use CreateEncryptor with non-default IV</value>
```

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1179,4 +1179,19 @@
...
+  </data>
+  <data name="DefinitelyUseCreateEncryptorWithNonDefaultIVMessage" xml:space="preserve">
+    <value>Symmetric encryption uses potentially repeatable initialization vector</value>
+  </data>
+  <data name="MaybeUseCreateEncryptorWithNonDefaultIV" xml:space="preserve">
+    <value>Ensure use CreateEncryptor with the default IV </value>
+  </data>
+  <data name="MaybeUseCreateEncryptorWithNonDefaultIVMessage" xml:space="preserve">
+    <value>Non default initialization vector, which can be potentially repeatable, is used in the encrypion. Ensure use the default one.</value>
```
gewarren:
```suggestion
    <value>The non-default initialization vector, which can be potentially repeatable, is used in the encryption. Use the default one instead.</value>
```

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1179,4 +1179,19 @@
...
     <value>HttpClient may be created without enabling CheckCertificateRevocationList</value>
   </data>
+  <data name="DefinitelyUseCreateEncryptorWithNonDefaultIV" xml:space="preserve">
+    <value>Do not use CreateEncryptor with non default IV</value>
+  </data>
+  <data name="DefinitelyUseCreateEncryptorWithNonDefaultIVMessage" xml:space="preserve">
+    <value>Symmetric encryption uses potentially repeatable initialization vector</value>
+  </data>
+  <data name="MaybeUseCreateEncryptorWithNonDefaultIV" xml:space="preserve">
+    <value>Ensure use CreateEncryptor with the default IV </value>
```
gewarren:
```suggestion
    <value>Use CreateEncryptor with the default IV</value>
```

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1179,4 +1179,19 @@
   <data name="MaybeDisableHttpClientCRLCheckMessage" xml:space="preserve">
     <value>HttpClient may be created without enabling CheckCertificateRevocationList</value>
   </data>
+  <data name="DefinitelyUseCreateEncryptorWithNonDefaultIV" xml:space="preserve">
```
gewarren: This name confuses me. It contradicts what the message is saying. Should it be `DontUseCreateEncryptorWithNonDefaultIV`?

src/Microsoft.NetCore.Analyzers/Core/Security/DoNotUseCreateEncryptorWithNonDefaultIV.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        private static readonly PropertyMapperCollection PropertyMappers = new PropertyMapperCollection(
+            new PropertyMapper(
+                "IV",
+                (PointsToAbstractValue pointsToAbstractValue) =>
+                {
+                    return PropertySetAbstractValueKind.Flagged;
+                }));
+
+        private static readonly HazardousUsageEvaluatorCollection HazardousUsageEvaluators = new HazardousUsageEvaluatorCollection(
```
dotpaul: Did you decide not to take GenerateIV() into account?

dotpaul: If not taking GenerateIV() into account, could probably skip PropertySetAnalysis and just produce a diagnostic on assigning to the .IV property.

---
In reply to: [323019864](https://github.com/dotnet/roslyn-analyzers/pull/2822#discussion_r323019864) [](ancestors = 323019864)

LLLXXXCCC: I see what you mean now. But PropertyMapper can't tell if the property is set with a value from GenerateIV(). And if we just detect it's assignment, we will produce a diagnostic when there's no invocation of CreateEncryptor(). Is it ok?

dotpaul: I'm sorry, I forgot about decryption!  We do need to care about if CreateEncryptor() is invoked.  Can you revert back to using PropertySetAnalysis?

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
LLLXXXCCC: There're two cases. One is that we are 100% sure that CreateEncryptor uses non-default IV. The other is that non-default IV can be used but we are not sure. For example in the case below, non-default IV will be used when it fails to meets the requirement r.Next(6) == 4:
        var aesCng  = new AesCng();
        Random r = new Random();

        if (r.Next(6) == 4)
        {
            aesCng.IV = rgbIV;
        }

        aesCng.CreateEncryptor();

So, we created these two rules (Definitely and Maybe) for different scenarios.

src/Microsoft.NetCore.Analyzers/Core/xlf/MicrosoftNetCoreAnalyzersResources.cs.xlf
```diff
@@ -68,13 +68,28 @@
         <note />
       </trans-unit>
       <trans-unit id="DoNotUseInsecureRandomnessDescription">
+        <source>Using a cryptographically weak pseudo-random number generator may allow an attacker to predict what security sensitive value will be generated. Use a cryptographically strong random number generator if an unpredictable value is required, or ensure that weak pseudo-random numbers aren't used in a security sensitive manner.</source>
+        <target state="new">Using a cryptographically weak pseudo-random number generator may allow an attacker to predict what security sensitive value will be generated. Use a cryptographically strong random number generator if an unpredictable value is required, or ensure that weak pseudo-random numbers aren't used in a security sensitive manner.</target>
```
dotpaul: ><target state="new"> [](start = 8, length = 20)

Should probably manually swap DoNotUseInsecureRandomnessDescription / Message so that localizers don't have to retranslate the strings (keep state="translated")

dotpaul:
Create a new PR to merge into the 2.9.x branch (not master) when you're ready.

src/Microsoft.NetCore.Analyzers/Core/xlf/MicrosoftNetCoreAnalyzersResources.cs.xlf
dotpaul: You also have to keep the translated text from before your change.

---
In reply to: [329818334](https://github.com/dotnet/roslyn-analyzers/pull/2822#discussion_r329818334) [](ancestors = 329818334)

src/Microsoft.NetCore.Analyzers/Core/xlf/MicrosoftNetCoreAnalyzersResources.de.xlf
```diff
@@ -68,13 +68,28 @@
         <note />
       </trans-unit>
       <trans-unit id="DoNotUseInsecureRandomnessDescription">
+        <source>Using a cryptographically weak pseudo-random number generator may allow an attacker to predict what security sensitive value will be generated. Use a cryptographically strong random number generator if an unpredictable value is required, or ensure that weak pseudo-random numbers aren't used in a security sensitive manner.</source>
+        <target state="new">Using a cryptographically weak pseudo-random number generator may allow an attacker to predict what security sensitive value will be generated. Use a cryptographically strong random number generator if an unpredictable value is required, or ensure that weak pseudo-random numbers aren't used in a security sensitive manner.</target>
+        <note />
```
dotpaul: Should keep all the the already translated text in all the XLF files.  Sorry that wasn't clear.

LLLXXXCCC: You've make yourself very clear. I replaced it all but didn't check if it's committed right. My bad.

PR #2825, opened 2019-09-06
Removing unmaintained SDL rulesets

dotpaul:
Since these aren't being maintained and we're not using them internally.


dotpaul:
Thanks @mavasani !

PR #2826, opened 2019-09-06
Bail out from CA2235 for NetStandard assemblies

### LINKED ISSUE #1775: CA2235 warning on primitive data types
#### Analyzer package

#### Package Version

Example: v2.6.2-beta2

#### Diagnostic ID

[CA2235](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2235-mark-all-non-serializable-fields)

#### Repro steps

1. Create a class decorated with the Serializable attribute.
2. Add an auto-property of type string.

#### Expected behavior

No warnings shown since all .NET primitive types are serializable by defualt.

#### Actual behavior

A CA2235 warning is shown:
`Warning	CA2235	Field Name is a member of type MyClass which is serializable but is of type string which is not serializable`

mavasani:
Fixes #1775


mavasani:
Thanks @dotpaul

PR #2828, opened 2019-09-06
ReleaseNoteUtil: Ignore CodeFixProviders with no-op RegisterCodeFixesAsync() methods

dotpaul:
In another PR, we can consolidate FixerExtensions.cs implementations, and make sure GenerateAnalyzerRulesets works (I seem to be getting different results between ReleaseNotesUtil and GenerateAnalyzerRulesets even before this change).


dotpaul:
Thanks @mavasani !

PR #2829, opened 2019-09-06
Disable couple of DFA security analyzers by default

mavasani:
They were showing up in the performance traces for building Roslyn.sln with latest bits and slowing down the build considerably. I will file a separate issue to investigate performance improvements for the analyses.


mavasani:
Thanks @dotpaul

PR #2831, opened 2019-09-08
[WIP] Add rule for detecting certain readonly mutable structs

### LINKED ISSUE #2811: New rule: using mutable structs in readonly fields
A split off issue from [corefx/issues/40739](https://github.com/dotnet/corefx/issues/40739) to track the suggested rule

`readonly SpinLock fields. SpinLock is a mutable struct, meant only for advanced scenarios. Accidentally making a SpinLock field readonly can result in silent but significant problems, as any mutations to the instance (e.g. Enter, Exit) will be done on a compiler-generated copy and thus be ignored, making the lock an expensive nop. (It might make sense to extend this analyzer to additional mutable struct types where storing them in a readonly field is likely a bug, e.g. GCHandle.)`

----

Detecting this should be easy - for each class/struct find all fields that are `readonly` and of a type that is designed to be mutable (for starters, the `GCHandle` and `SpinLock`).

Questions:
1. Can we think of more types that should be included in this list?
2. An automated fix should just remove the readonly modified from the field, right?
3. What would be the default severity for the rule? As there is a rather large chance this situation will lead to a bug, is a warning sufficient?
4. Are there any edge-cases where having a readonly mutable-by-design struct might actually be desired?

mareklinka:
As proposed in [corefx/issues/40739](https://github.com/dotnet/corefx/issues/40739) and #2811, this PR adds a new rule for detecting readonly fields of type `System.Threading.SpinLock` and `System.Runtime.InteropServices.GCHandle`.

Both C# and VB versions are implemented but VB is not my strong suite so it might not be entirely idiomatic VB.NET.

Tests for both analyzers and fixers are there but if I managed to omit some cases, let me know and I'll add them.

I'm also not sure I'm defensive enough (it's been a while since I worked with Roslyn so I might have missed some recommended practices).

ID of the new rule is Performance/CA1829.

Fixes #2811


src/Microsoft.NetCore.Analyzers/Core/Performance/MutableStructsShouldNotBeUsedForReadonlyFields.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Performance
+{
+    public abstract class MutableStructsShouldNotBeUsedForReadonlyFieldsAnalyzer : DiagnosticAnalyzer
```
mavasani: This analyzer can be written as a language agnostic symbol analyzer, which just registers a SymbolAction with `SymbolKind.Field`. For example, look at https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotDeclareVisibleInstanceFields.cs.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.cs
```diff
@@ -0,0 +1,42 @@
...
+            var fieldDeclaration = (FieldDeclarationSyntax)context.Node;
+
+            var fieldTypeInfo = context.SemanticModel.GetTypeInfo(fieldDeclaration.Declaration.Type);
+
+            if (!(fieldTypeInfo.Type is INamedTypeSymbol fieldType))
+            {
+                return;
+            }
+
+            var typesToCheck = MutableValueTypesOfInterest.Select(typeName => context.Compilation.GetTypeByMetadataName(typeName)).Where(type => type != null).ToList();
```
mavasani: `GetTypeByMetadataName` should only be invoked during CompilationStartAction callback. You should check for null values of the required well known type(s) and bail out if the analyzer is not applicable for the current compilation, and register nested symbol action if it is applicable. For example: https://github.com/dotnet/roslyn-analyzers/blob/3e4ef017b0a83b1330527b35aa2781bbca927e0d/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCreateTasksWithoutPassingATaskScheduler.cs#L42-L54

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpMutableStructsShouldNotBeUsedForReadonlyFieldsAnalyzer : MutableStructsShouldNotBeUsedForReadonlyFieldsAnalyzer
+    {
+        protected override void RegisterDiagnosticAction(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(NodeAction, SyntaxKind.FieldDeclaration);
+        }
+
+        private static void NodeAction(SyntaxNodeAnalysisContext context)
```
mavasani: In general, we strongly recommend not using SyntaxNodeAction in this repo. Majority of analyzers should fall under the below 2 buckets:

1. Analyze the attributes/properties of a symbols: Prefer RegisterSymbolAction
2. Analyze executable code: Prefer RegisterOperationAction

SyntaxNodeActions are only required for analyses related to syntax nodes that are not part of any symbol's declaration OR executable code, which is extremely rare for an analyzers in this repo.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.Fixer.cs
```diff
@@ -0,0 +1,55 @@
...
+                EquivalencyKey);
+
+            context.RegisterCodeFix(removeReadonlyAction, context.Diagnostics);
+        }
+
+        private static async Task<Document> RemoveReadonlyKeyword(CodeFixContext context,
+            FieldDeclarationSyntax fieldDeclarationSyntax)
+        {
+            var editor = await DocumentEditor.CreateAsync(context.Document, context.CancellationToken).ConfigureAwait(false);
+            var withoutReadonly = fieldDeclarationSyntax.WithModifiers(
```
mavasani: A slighly simpler implementation would be: http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Features/CodeFixes/HideBase/HideBaseCodeFixProvider.AddNewKeywordAction.cs,37

This also allows you to make the fixer language agnostic (you don't need language specific `SyntaxKind` anymore)

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/MutableStructsShouldNotBeUsedForReadonlyFieldsTests.Fixer.cs
```diff
@@ -0,0 +1,211 @@
...
+public class C
+{
+    private SpinLock _sl = new SpinLock();
+    private GCHandle _gch = new GCHandle();
+
+    private SpinLock _sl_noinit;
+    private GCHandle _gch_noinit;
+}
+";
+            await CSharpCodeFixVerifier<CSharpMutableStructsShouldNotBeUsedForReadonlyFieldsAnalyzer, CSharpMutableStructsShouldNotBeUsedForReadonlyFieldsFixer>
```
mavasani: Kindly follow the pattern used by all the existing unit tests in the repo and define usings to avoid clutter in individial unit tests such as https://github.com/dotnet/roslyn-analyzers/blob/c4e547383037fc989178ad61974b3de5547ffb76/src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/TestForNaNCorrectlyTests.Fixer.cs#L8-L52

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/MutableStructsShouldNotBeUsedForReadonlyFieldsTests.Fixer.cs
```diff
@@ -0,0 +1,211 @@
...
+        public async Task CSharpReadonlyFieldsOfKnownMutableTypes_RemovesReadonlyModifier()
+        {
+            var initial = @"
+using System;
+using System.Threading;
+using System.Runtime.InteropServices;
+
+public class C
+{
+    private [|readonly|] SpinLock _sl = new SpinLock();
```
mavasani: Can you add C# and VB unit tests where you have a field declaration with multiple fields declared, such as `private readonly SpinLock _sl1 = new SpinLock(), sl2 = new SpinLock();`

src/Microsoft.NetCore.Analyzers/UnitTests/Performance/MutableStructsShouldNotBeUsedForReadonlyFieldsTests.Fixer.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        [Fact]
+        public async Task BasicReadonlyFieldsOfKnownMutableTypes_RemovesReadonlyModifier()
+        {
+            var initial = @"
+Imports System.Threading
+Imports System.Runtime.InteropServices
+
+Public Class Class1
+    Public [|ReadOnly|] _sl As SpinLock = New SpinLock()
```
mavasani: Can you add VB unit test with field declaration that using AsNew clause. For example, `Private Readonly _sl As New SpinLock()`

src/Microsoft.NetCore.Analyzers/Core/Performance/MutableStructsShouldNotBeUsedForReadonlyFields.Fixer.cs
```diff
@@ -0,0 +1,41 @@
...
+            var root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            var node = root.FindNode(context.Span);
+
+            AnalyzeCodeFix(context, node);
+        }
+
+        protected abstract void AnalyzeCodeFix(CodeFixContext context, SyntaxNode targetNode);
+
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(MutableStructsShouldNotBeUsedForReadonlyFieldsAnalyzer.RuleId);
```
mavasani: Nit: Move the property overrides before the method overrides.

src/Microsoft.NetCore.Analyzers/Core/Performance/MutableStructsShouldNotBeUsedForReadonlyFields.Fixer.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public override FixAllProvider GetFixAllProvider()
```
mavasani: Nit: Move this override ahead of `RegisterCodeFixesAsync` override.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1179,4 +1179,13 @@
...
     <value>HttpClient may be created without enabling CheckCertificateRevocationList</value>
   </data>
+  <data name="MutableStructsShouldNotBeUsedForReadonlyFieldsTitle" xml:space="preserve">
+    <value>Certain readonly structs should not be used as readonly fields</value>
+  </data>
+  <data name="MutableStructsShouldNotBeUsedForReadonlyFieldsDescription" xml:space="preserve">
+    <value>Certain value types (e.g. SpinLock, GCHandle) are explicitly designed to be mutable. Declaring a field of such type as readonly might cause unintentional shadow copying by the compiler, leading to performance degradation or bugs.</value>
+  </data>
+  <data name="MutableStructsShouldNotBeUsedForReadonlyFieldsMessage" xml:space="preserve">
+    <value>Field {0} of type {1} should not be readonly.</value>
```
mavasani: `Field `{0}` of type `{1}` should not be readonly.`

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1179,4 +1179,13 @@
   <data name="MaybeDisableHttpClientCRLCheckMessage" xml:space="preserve">
     <value>HttpClient may be created without enabling CheckCertificateRevocationList</value>
   </data>
+  <data name="MutableStructsShouldNotBeUsedForReadonlyFieldsTitle" xml:space="preserve">
+    <value>Certain readonly structs should not be used as readonly fields</value>
+  </data>
+  <data name="MutableStructsShouldNotBeUsedForReadonlyFieldsDescription" xml:space="preserve">
+    <value>Certain value types (e.g. SpinLock, GCHandle) are explicitly designed to be mutable. Declaring a field of such type as readonly might cause unintentional shadow copying by the compiler, leading to performance degradation or bugs.</value>
```
mavasani: Given that `SpinLock` and `GCHandle` are just couple of potentially many such types for which this rule can be applicable, I think we should make this configurable for end users to allow specifying the set of disallowed types such that mutable readonly fields of all of those types whould be flagged. If we agree, then I would suggest the following:

1. Use the [existing option](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#disallowed-symbol-names) `disallowed_symbol_names`
2. See how this option is used to make CA1031 configurable with this commit: https://github.com/dotnet/roslyn-analyzers/commit/40cd30186631a3e5d33492b37f5c69b0aa59876c#diff-078582649f9598152821c2fd110c7219. I would suggest a similar implementation and unit test for it.

If you feel this would overload your PR, I can do this as a followup PR as well. In this case, please file a separate tracking issue to make this rule configurable.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.Fixer.cs
mareklinka: I tried approaching it as you suggested but I think I'm missing something. The situation:

1. The analyzer now operates on symbols and reports the location of the diagnostic as `fieldSymbol.Locations.First()`. This means the location of the symbol name is reported.
2. That means the the fix will get a span that points to a `VariableDeclaratorSyntax`, not the actual `FieldDeclarationSyntax`. That works fine for the `SyntaxGenerator`, which ascends the tree correctly and returns a new `FieldDeclarationSyntax` node, but then I can't simply call `root.ReplaceNode` because the nodes are of different types.
3. Furthermore, this complicates executing the fix on statement `private readonly SpinLock _sl1 = new SpinLock(), sl2 = new SpinLock();`. Here the generator doesn't work because it creates a new `FieldDeclarationSyntax` but *without* the second variable declaration.

Either I'm missing the correct API/usage of it or it looks like it will be easier to make the fix language-specific and simply use:

``` csharp
root.FindNode(context.Span).FirstAncestorOrSelf<FieldDeclarationSyntax>() // this is language-specific
...
var generator = SyntaxGenerator.GetGenerator(context.Document);
var nodeWithoutReadonly = generator.WithModifiers(targetNode, generator.GetModifiers(targetNode).WithIsReadOnly(false));
var newRoot = root.ReplaceNode(originalFieldDeclaration, nodeWithoutReadonly);
return Task.FromResult(context.Document.WithSyntaxRoot(newRoot));
```

Testing this, it leads to correct behavior even in case of multi-declaration.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
mareklinka: I agree that configuring this could be useful but I'd opt into enabling this configuration in a separate PR. I'll file it once this one is merged, if that's ok.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.Fixer.cs
mavasani: Thanks, I meant to recommend the exact same implementation as your comment, but keep it in the shared layer with a type parameter TFieldDeclarationSyntax, where TFieldDeclarationSyntax: SyntaxNode. Language specific subtypes will just need to provide the language specific type argument FieldDeclarationSyntax and have no code within.

src/Microsoft.NetCore.Analyzers/Core/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Sounds reasonable!

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.cs
sharwell: 📝 Note that syntax node actions are not inherently bad. However, the `GetSymbolInfo` method is very expensive, and the cost can be avoided by using one of the other approaches. If you need to operate on symbols (as opposed to just syntax), symbol and operation actions are more efficient. In some cases, operation actions also simplify the work required to support both C# and VB.

Evangelink:
@sharwell, @mavasani have you been able to meet and discuss about the direction this PR should go to?

Evangelink:
@mareklinka Is there something I can do to help moving forward with this new analyzer?

mareklinka:
@Evangelink Well, as this was originally waiting for a clarification meeting and there was no update for a long time, I stopped really paying attention after a while.

As @sharwell mentioned, if the analyzer is supposed to be detecting copying of structs, it will be much more complicated. In that case, I think I will leave the implementation to someone with more time for it.

If the readonly semantics are fine, I think I should be able to revisit the PR and solve the merge conflicts so that it could be reviewed and merged.

Evangelink:
@mavasani, @sharwell have you been able to discuss https://github.com/dotnet/roslyn-analyzers/pull/2831#pullrequestreview-285234353? The original ticket is talking about non-readonly IMO, the non-copyable part would also make sense but could be either an update to this rule or a separate rule.

src/Microsoft.NetCore.Analyzers/CSharp/Performance/CSharpMutableStructsShouldNotBeUsedForReadonlyFields.cs
Evangelink: @mareklinka regarding the "rewrite" of this analyzer, you can have a look at this PR I started: https://github.com/dotnet/roslyn-analyzers/pull/3405/files

sharwell:
@mareklinka I implemented the initial non-copyable analysis in #3420

sharwell:
@mavasani My recommendation for this is to use it as more evidence in support of copy value callbacks integrated directly into the analyzer driver as part of preparing for dotnet/runtime#50389.

PR #2833, opened 2019-09-09
Add assert for required parameter 'valueDomain'

nschuessler:
The parameter 'valueDomain' is optional in the constructor of `AbstractDataFlowAnalysisContext` and must not be null in `ComputeHashCodeParts` which looks like a NR exception waiting to happen.


src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/AbstractDataFlowAnalysisContext.cs
```diff
@@ -40,6 +40,7 @@ public abstract class AbstractDataFlowAnalysisContext<TAnalysisData, TAnalysisCo
             InterproceduralAnalysisData<TAnalysisData, TAnalysisContext, TAbstractAnalysisValue> interproceduralAnalysisDataOpt,
             InterproceduralAnalysisPredicate interproceduralAnalysisPredicateOpt)
         {
+            Debug.Assert(valueDomain != null, "valueDomain must not be null for use in ComputeHashCodeParts");
```
mavasani: Nit: not sure the message is useful here. ValueDomain is one of the core pieces of merging DFA values, so the assert seems self contained.

nschuessler: Ok. I figured you could figure out which parameter w/o attaching a debugger by giving a message.

nschuessler:
Didn't mean to toss away the PR.

PR #2842, opened 2019-09-11
Improve CA2200 Rethrow analyzer to allow reassigned exception

goyzhang:
If the caught exception is reassigned (new or set to another reference), it's then not actually "rethrowing the caught exception" such that it should not be reported as a diagnostic.
For example, this should be allowed:
```
            try
            {
                //.....
            }
            catch(Exception ex)
            {
                //.....
                ex = new SomeCustomException("some msg");
                throw ex;
            }
```

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/CSharp/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/CSharp/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
@@ -38,7 +38,9 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
                                 catchClause.Declaration != null &&
                                 catchClause.Declaration.Identifier.RawKind != 0)
                             {
-                                if (!(context.SemanticModel.GetSymbolInfo(expr).Symbol is ILocalSymbol local) || local.Locations.Length == 0)
+                                if (!(context.SemanticModel.GetSymbolInfo(expr).Symbol is ILocalSymbol local)
+                                    || local.Locations.Length == 0
+                                    || context.SemanticModel.AnalyzeDataFlow(catchClause.Block).WrittenInside.Contains(local))
                                 {
                                     return;
                                 }
--- a/src/Microsoft.CodeQuality.Analyzers/VisualBasic/QualityGuidelines/BasicRethrowToPreserveStackDetails.vb
+++ b/src/Microsoft.CodeQuality.Analyzers/VisualBasic/QualityGuidelines/BasicRethrowToPreserveStackDetails.vb
@@ -55,10 +55,12 @@ Namespace Microsoft.CodeQuality.VisualBasic.Analyzers.QualityGuidelines
 
         Private Shared Function IsCaughtLocalThrown(semanticModel As SemanticModel, catchStatement As CatchStatementSyntax, throwExpression As ExpressionSyntax) As Boolean
             Dim local = TryCast(semanticModel.GetSymbolInfo(throwExpression).Symbol, ILocalSymbol)
-            Dim catchBlock = TryCast(catchStatement.Parent, CatchBlockSyntax)
+            Dim catchBlock = DirectCast(catchStatement.Parent, CatchBlockSyntax)
             If local Is Nothing _
                     OrElse local.Locations.Length = 0 _
-                    OrElse semanticModel.AnalyzeDataFlow(catchBlock?.Statements.First, catchBlock?.Statements.Last).WrittenInside.Contains(local) Then
+                    OrElse catchBlock Is Nothing _
+                    OrElse catchBlock.Statements.Count = 0 _
+                    OrElse semanticModel.AnalyzeDataFlow(catchBlock.Statements.First, catchBlock.Statements.Last).WrittenInside.Contains(local) Then
                 Return False
             End If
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.CodeQuality.Analyzers/CSharp/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
```diff
@@ -38,7 +38,9 @@ private static void AnalyzeNode(SyntaxNodeAnalysisContext context)
                                 catchClause.Declaration != null &&
                                 catchClause.Declaration.Identifier.RawKind != 0)
                             {
-                                if (!(context.SemanticModel.GetSymbolInfo(expr).Symbol is ILocalSymbol local) || local.Locations.Length == 0)
+                                if (!(context.SemanticModel.GetSymbolInfo(expr).Symbol is ILocalSymbol local)
+                                    || local.Locations.Length == 0
+                                    || context.SemanticModel.AnalyzeDataFlow(((CatchClauseSyntax)syntax).Block).WrittenInside.Contains(local))
```
mavasani: Kindly also update the VB implementation of the analyzer and add a unit test: https://github.com/dotnet/roslyn-analyzers/blob/d578f2239d4ab331e71b0d5e27f6ffb537176176/src/Microsoft.CodeQuality.Analyzers/VisualBasic/QualityGuidelines/BasicRethrowToPreserveStackDetails.vb#L56-L68

Thanks!

goyzhang: I'd like to. But I'm not familiar with vb. And for some reason I can't get the `DataFlowAnalysis` from CatchBlockSyntax. Can someone else help?

mavasani: Tagging @AlekseyTs and @gafter for guidance on `AnalyzeDataFlow` API for VB catch block.
Seems like `AnalyzerDataFlow` method in VB expects a node deriving from `ExecutableStatementSyntax`: http://source.roslyn.io/#Microsoft.CodeAnalysis.VisualBasic/Compilation/SemanticModel.vb,3363, but [CatchBlockSyntax](http://source.roslyn.io/#q=CatchBlockSyntax) derives from `VisualBasicSyntaxNode`

mavasani: Ah, I think you can re-use this overload of the API: http://source.roslyn.io/#Microsoft.CodeAnalysis/Compilation/Extensions.cs,152 pass in the first and last statement of `CatchBlockSyntax.Statements`

goyzhang: @mavasani Thanks for the tip! I have updated the commit.

src/Microsoft.CodeQuality.Analyzers/VisualBasic/QualityGuidelines/BasicRethrowToPreserveStackDetails.vb
```diff
@@ -55,7 +55,10 @@ Namespace Microsoft.CodeQuality.VisualBasic.Analyzers.QualityGuidelines
 
         Private Shared Function IsCaughtLocalThrown(semanticModel As SemanticModel, catchStatement As CatchStatementSyntax, throwExpression As ExpressionSyntax) As Boolean
             Dim local = TryCast(semanticModel.GetSymbolInfo(throwExpression).Symbol, ILocalSymbol)
-            If local Is Nothing OrElse local.Locations.Length = 0 Then
+            Dim catchBlock = TryCast(catchStatement.Parent, CatchBlockSyntax)
+            If local Is Nothing _
+                    OrElse local.Locations.Length = 0 _
+                    OrElse semanticModel.AnalyzeDataFlow(catchBlock?.Statements.First, catchBlock?.Statements.Last).WrittenInside.Contains(local) Then
```
mavasani: API throws an ArgumentNullException for a null value passed in for first or last statement: http://source.roslyn.io/#Microsoft.CodeAnalysis.VisualBasic/Compilation/SemanticModel.vb,c793e537f42cae32.

goyzhang: uh...So it is possible that a `catchBlock` is not null but its` statements` are null?

mavasani: The added code has a try cast, so the catchBlock itself can be null. I am also not sure what is the syntax tree structure for an empty block, does it have a null first and/or last statement? It would be good to add a test.

goyzhang: Alright, from the caller the `catchStatement` is passed in from `Dim catchStatement = DirectCast(node, CatchBlockSyntax).CatchStatement` such that it would throw NullReference before `IsCaughtLocalThrown` is called. It doesn't make sense to me that the callback is triggered but we have no catchblock. I can add a test for this. But if it is null, shall I just continue to check local?

mavasani: 1. Can you change `Dim catchBlock = TryCast(catchStatement.Parent, CatchBlockSyntax)` to a `DirectCast`?
2. Please add a unit test for empty catch block

goyzhang: @mavasani  I have updated it to `TryCast` with null check. But for the unit test, this analyzer would not even be invoked for cases like:
```
        Try
            Throw New Exception()
        Catch e As Exception

        End Try
```
Is it still necessary for this unit test?

mavasani: No, it should be fine. Thanks for the contribution.

PR #2857, opened 2019-09-18
Fix IndexOutOfRangeException bug of CA5391.

### LINKED ISSUE #2844: CA5391 fires on [ApiController] annotated MVC controllers
#### Analyzer package

Example: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

Example: 2.9.5-beta1.final

#### Diagnostic ID

Example: CA5391 (not yet documented)

#### Repro steps

1. Create an API controller using the instructions here: https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-2.2
2. Create a `PUT` or `DELETE` controller action that requires authorization: https://docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-2.2
3. Specify the version `2.9.5-beta1.final` for the `Microsoft.CodeAnalysis.FxCopAnalyzers` package.

#### Expected behavior

I would expect that any controllers annotated with `[ApiController]` would be exempt to the anti-forgery validation rule. That being said, no alternative solutions have been provided in the form of documentation.

#### Actual behavior

Compiling will give you warnings like so:
```
Controllers/TestController.cs(73,41): warning CA5391: Method DeleteAsync handles a HttpDelete request without performing antiforgery token validation. You also need to ensure that your HTML form sends an antiforgery token. [/Users/anthonyiacono/Git/TestProject/src/Web/Web.csproj]
CSC : warning AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.IndexOutOfRangeException' with message 'Index was outside the bounds of the array.'. [/Users/anthonyiacono/Git/TestProject/src/Web/Web.csproj]
CSC : warning AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.IndexOutOfRangeException' with message 'Index was outside the bounds of the array.'. [/Users/anthonyiacono/Git/TestProject/src/Web/Web.csproj]
CSC : warning AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.IndexOutOfRangeException' with message 'Index was outside the bounds of the array.'. [/Users/anthonyiacono/Git/TestProject/src/Web/Web.csproj]
```

LLLXXXCCC:
Fixes #2844


LLLXXXCCC:
#2844

mavasani:
@LLLXXXCCC minor suggestion - if your PR closes a specific issue, then you can add a `Fixes #%issue_id%` comment in the PR description and this will ensure that when you merge the PR, the mentioned issue will automatically get closed.

docs/Analyzer Configuration.md
```diff
@@ -298,3 +298,12 @@ Option Values: Integer values of System.Runtime.InteropServices.DllImportSearchP
 Default Value: Specific to each configurable rule ('770', which is AssemblyDirectory | UseDllDirectoryForDependencies | ApplicationDirectory, by default for most rules)
 
 Example: `dotnet_code_quality.CA5392.unsafe_DllImportSearchPath_bits = 770`
+
+#### Configure if only look at Controller-derived classes when considering CSRF
```
mavasani: Is this an option specific to Dataflow analysis? I think we should move this option, and even `unsafe_DllImportSearchPath_bits` and `sufficient_IterationCount_for_weak_KDF_algorithm` above the Dataflow analysis section: https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#dataflow-analysis

mavasani: Can you also add "Configurable Rules" entry like https://github.com/dotnet/roslyn-analyzers/pull/2857/files#diff-1b5cdc4937a8cdfc79545945ad6c0783R251 so we are explicit about which CA rule specifies which option? It would be good to specify it even for prior to options.

docs/Analyzer Configuration.md
```diff
@@ -298,3 +298,12 @@ Option Values: Integer values of System.Runtime.InteropServices.DllImportSearchP
...
+Option Name: `only_look_at_derived_classes_of_Controller`
+
+Option Values: Boolean values
+
+Default Value: Specific to each configurable rule ('true' by default for most rules)
```
mavasani: Is this correct? Seems like this option is currently only respected by one rule and the default value is always true.

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
```diff
@@ -215,9 +224,10 @@ public override void Initialize(AnalysisContext context)
                     var derivedControllerTypeSymbol = (INamedTypeSymbol)symbolAnalysisContext.Symbol;
                     var baseTypes = derivedControllerTypeSymbol.GetBaseTypes();
 
-                    // An subtype of `Microsoft.AspNetCore.Mvc.Controller` or `Microsoft.AspNetCore.Mvc.ControllerBase`).
+                    // An subtype of `Microsoft.AspNetCore.Mvc.Controller`, which indicates that cookie-based authentication is used and thus CSRF is a concern.
                     if (baseTypes.Contains(controllerTypeSymbol) ||
-                        baseTypes.Contains(controllerBaseTypeSymbol))
+                            (!onlyLookAtDerivedClassesOfController &&
+                            baseTypes.Contains(controllerBaseTypeSymbol)))
```
mavasani: Looking at the implementation, I think a better option name here might be `exclude_mvc_controller_base`? @dotpaul?

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
```diff
@@ -190,16 +197,18 @@ public override void Initialize(AnalysisContext context)
...
                                     .GetMembers()
                                     .OfType<IMethodSymbol>()
                                     .FirstOrDefault(
-                                        s => (s.Name == "OnAuthorizationAsync" ||
-                                            s.Name == "OnAuthorization") &&
-                                            s.ReturnType.Equals(taskTypeSymbol) &&
+                                        s => (s.Name == "OnAuthorizationAsync" &&
+                                            s.ReturnType.Equals(taskTypeSymbol) ||
+                                            s.Name == "OnAuthorization" &&
+                                            s.ReturnType.SpecialType == SpecialType.System_Void) &&
```
mavasani: `s.ReturnsVoid`?

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
```diff
@@ -340,7 +350,8 @@ void FindAllTheSpecifiedCalleeMethods(ISymbol methodSymbol, HashSet<ISymbol> vis
 
                         foreach (var child in callingMethods.Keys)
                         {
-                            if (onAuthorizationAsyncMethodSymbols.Contains(child))
+                            if (child is IMethodSymbol childMethodSymbol &&
+                                onAuthorizationAsyncMethodSymbols.ContainsKey(childMethodSymbol))
```
mavasani: Do we know where the IndexOutOfRangeException was being thrown? I presume it is in the indexer access below `results[methodSymbol]`? If so, I am not sure how this PR would fix that exception.

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
```diff
@@ -100,7 +107,7 @@ public override void Initialize(AnalysisContext context)
                 // Verify that validate anti forgery token attributes are used somewhere within this project,
                 // to avoid reporting false positives on projects that use an alternative approach to mitigate CSRF issues.
                 var usingValidateAntiForgeryAttribute = false;
-                var onAuthorizationAsyncMethodSymbols = new HashSet<IMethodSymbol>();
+                var onAuthorizationAsyncMethodSymbols = new ConcurrentDictionary<IMethodSymbol, bool>();
                 var actionMethodSymbols = new HashSet<(IMethodSymbol, string)>();
                 var actionMethodNeedAddingHttpVerbAttributeSymbols = new HashSet<IMethodSymbol>();
```
dotpaul: Don't you want to worry about these HashSets as well?

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
dotpaul: Whatever the name is, I think "aspnet_core" ought to appear in the name somewhere.

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
dotpaul: @LLLXXXCCC based on your guess of the exception is a thread safety issue, I've come up with an unreliable repro (get the rule to trigger when there are multiple controllers in the project).

https://github.com/dotpaul/repro2844

Replace the Microsoft.NetCore.Analyzers.dll with your own build (maybe with Debug.Break() somewhere so you can catch the exceptions), and try building / rebuilding multiple times.

dotpaul: Also, don't remember if I got IndexOutOfRangeExceptions, but definitely ran into some others like:

warning AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.ArgumentException' with message 'Destination array was not long enough. Check destIndex and length, and the array's lower bounds.'.

warning AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.

mavasani: I think these might all be likely due to use of non-concurrent data structure in this analyzer.

docs/Analyzer Configuration.md
dotpaul: Instead of just "Controller", I think we should have the fully qualified class name so it's not ambiguous with the older ASP.NET.

Also, tagging @gewarren for wording suggestions

docs/Analyzer Configuration.md
```diff
@@ -155,7 +155,29 @@ Examples:
...
+
+Configurable Rules: CA5393
+
+Option Values: Integer values of System.Runtime.InteropServices.DllImportSearchPath
+
+Default Value: '770', which is AssemblyDirectory | UseDllDirectoryForDependencies | ApplicationDirectory
+
+Example: `dotnet_code_quality.CA5393.unsafe_DllImportSearchPath_bits = 770`
+
+### Configure if exclude aspnet core mvc ControllerBase when considering CSRF
```
gewarren:
```suggestion
### Exclude ASP.NET Core MVC ControllerBase when considering CSRF
```

docs/Analyzer Configuration.md
```diff
@@ -155,7 +155,29 @@ Examples:
 |`dotnet_code_quality.excluded_type_names_with_derived_types = MyType1\|MyType2` | Matches all types named either 'MyType1' or 'MyType2' and all of their derived types in the compilation
 |`dotnet_code_quality.excluded_type_names_with_derived_types = M:NS.MyType` | Matches specific type 'MyType' with given fully qualified name and all of its derived types
 |`dotnet_code_quality.excluded_type_names_with_derived_types = M:NS1.MyType1\|M:NS2.MyType2` | Matches specific types 'MyType1' and 'MyType2' with respective fully qualified names and all of their derived types
- 
+
+### Configure unsafe DllImportSearchPath bits when using DefaultDllImportSearchPaths attribute
```
gewarren:
```suggestion
### Unsafe DllImportSearchPath bits when using DefaultDllImportSearchPaths attribute
```

(Just to be consistent with existing headings.)

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -73,5 +73,10 @@ internal static partial class EditorConfigOptionNames
         /// Do not use the OR operator to represent the bitwise combination of its member values, use the integeral value directly.
         /// </summary>
         public const string UnsafeDllImportSearchPathBits = "unsafe_DllImportSearchPath_bits";
+
+        /// <summary>
+        /// Boolean option to configure if exclude aspnet core mvc ControllerBase when considering CSRF.
```
gewarren:
```suggestion
        /// Boolean option to configure whether to exclude aspnet core mvc ControllerBase when considering CSRF.
```

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
dotpaul: @LLLXXXCCC were you able to repro some exceptions being thrown, and verify your changes don't have exceptions?

Would probably be a good idea to see if a unit test with multiple controllers can repro (and should have one anyway).

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
```diff
@@ -215,9 +224,10 @@ public override void Initialize(AnalysisContext context)
                     var derivedControllerTypeSymbol = (INamedTypeSymbol)symbolAnalysisContext.Symbol;
                     var baseTypes = derivedControllerTypeSymbol.GetBaseTypes();
 
-                    // An subtype of `Microsoft.AspNetCore.Mvc.Controller` or `Microsoft.AspNetCore.Mvc.ControllerBase`).
+                    // An subtype of `Microsoft.AspNetCore.Mvc.Controller`, which indicates that cookie-based authentication is used and thus CSRF is a concern.
```
dotpaul: >which indicates that cookie-based authentication is used and thus CSRF is a concern. [](start = 76, length = 84)

which probably indicates views are used and maybe cookie-based authentication is used and thus CSRF is a concern.

(Can't be sure.  🙂)

src/Microsoft.NetCore.Analyzers/Core/Security/UseAutoValidateAntiforgeryToken.cs
LLLXXXCCC: Yes, I can reproduce the two warnings you mentioned and the IndexOutOfRangeException. For now, the fix works well.

dotpaul: Please add a unit test that contains multiple controllers and produces diagnostics, so that we have a test committed that reproduces the original problem.

docs/Analyzer Configuration.md
```diff
@@ -213,6 +213,28 @@ Examples:
...
+Configurable Rules: CA5393
+
+Option Values: Integer values of System.Runtime.InteropServices.DllImportSearchPath
+
+Default Value: '770', which is AssemblyDirectory | UseDllDirectoryForDependencies | ApplicationDirectory
+
+Example: `dotnet_code_quality.CA5393.unsafe_DllImportSearchPath_bits = 770`
+
+### Exclude ASP.NET Core MVC ControllerBase when considering CSRF
+Option Name: `exclude_aspnet_core_mvc_controller_base`
```
dotpaul: >controller_base [](start = 38, length = 15)

Nit: I think this should be "controllerbase" rather than have an underscore in it, to reflect the class name.

PR #2861, opened 2019-09-23
Disabling CA5392/CA5393 by default

### LINKED ISSUE #2855: CA5392/CA5393 on cross-platform code
#### Analyzer package

#### Package Version

v2.9.5-beta1.final (Latest)

#### Diagnostic ID

* CA5392
* CA5393

#### Repro steps

CA5392:

``` csharp
[DllImport("System.Security.Cryptography.Native.OpenSsl")]
[return: MarshalAs(UnmanagedType.Bool)]
private static extern unsafe bool CryptoNative_GetRandomBytes(byte* buffer, int length);
```

CA5393:

``` csharp
[DllImport("System.Security.Cryptography.Native.OpenSsl")]
[DefaultDllImportSearchPaths(DllImportSearchPath.LegacyBehavior)]
[return: MarshalAs(UnmanagedType.Bool)]
private static extern unsafe bool CryptoNative_GetRandomBytes(byte* buffer, int length);
```

#### Expected behavior

`DefaultDllImportSearchPathsAttribute` appears to be ignored on platforms other than Windows, and `LegacyBehavior` is the only available option.  Warnings that are unavoidable on platforms other than Windows should be disabled by default.

#### Actual behavior

CA5392 and CA5393 are enabled by default.

dotpaul:
We can reevaluate if these rules become more universally applicable in the future.

Fixes #2855


src/Microsoft.NetCore.Analyzers/Core/Security/UseDefaultDllImportSearchPathsAttribute.cs
```diff
@@ -21,15 +21,15 @@ public sealed class UseDefaultDllImportSearchPathsAttribute : DiagnosticAnalyzer
             typeof(MicrosoftNetCoreAnalyzersResources),
             nameof(MicrosoftNetCoreAnalyzersResources.UseDefaultDllImportSearchPathsAttribute),
             nameof(MicrosoftNetCoreAnalyzersResources.UseDefaultDllImportSearchPathsAttributeMessage),
-            DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
+            false,
```
mavasani: Would be good to even update the documentation files, though that can come in a separate PR just prior to the release.

dotpaul: One day I'll remember.  Today is not that day.  🙂

mavasani: We should just fix the tooling to auto-generate these PRs instead!

PR #2866, opened 2019-09-24
Add workaround for no location diagnostics

mavasani:
First item listed in https://github.com/dotnet/roslyn/issues/38042: `Diagnostics with Location.None cannot be configured with .editorconfig` leads to the Location.None diagnostics being non-configurable in editorconfig. This is also currently blocking https://github.com/dotnet/roslyn/pull/37795, as FxCop analyzers is the primary analyzer package with Location.None diagnostics.
This change should be able to unblock https://github.com/dotnet/roslyn/pull/37795.

### DIFF
--- a/src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license 
+
+using System;
+
+namespace Microsoft.CodeAnalysis
+{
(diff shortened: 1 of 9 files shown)


src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+                case ReportDiagnostic.Suppress:
+                case ReportDiagnostic.Default:
+                    return null;
+
+                default:
+                    throw new NotImplementedException();
+            }
+        }
+
+        public static bool IsLessSevereThen(this ReportDiagnostic current, ReportDiagnostic other)
```
dotpaul: >Then [](start = 39, length = 4)

Nit: Than

PR #2869, opened 2019-09-25
Update CA2002 to detect usage of lock (this) and SyncLock Me

### LINKED ISSUE #294: Feature request: "lock(this)" should also be reported as bad practice
Maybe it could be added to CA2002 as discussed long time ago (see attached link)?
I think the reasons why this is bad are obvious - see also http://blogs.msdn.com/b/springboard/archive/2008/07/30/lock-this-bad-locking-patterns.aspx.

https://social.msdn.microsoft.com/Forums/en-US/42d85262-0611-4a81-be2d-e76b66734e92/lockthissynclock-me-and-fxcopca?forum=vstscode

Evangelink:
I have updated the `DoNotLockOnObjectsWithWeakIdentity` to also report when locking on `this` or `Me`.

Fix #294

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotLockOnObjectsWithWeakIdentity.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotLockOnObjectsWithWeakIdentity.cs
@@ -4,6 +4,7 @@ using System.Collections.Immutable;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
 
@@ -11,14 +12,14 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 {
     /// <summary>
     /// CA2002: Do not lock on objects with weak identities
-    /// 
+    ///
     /// Cause:
     /// A thread that attempts to acquire a lock on an object that has a weak identity could cause hangs.
-    /// 
+    ///
     /// Description:
-    /// An object is said to have a weak identity when it can be directly accessed across application domain boundaries. 
-    /// A thread that tries to acquire a lock on an object that has a weak identity can be blocked by a second thread in 
-    /// a different application domain that has a lock on the same object. 
+    /// An object is said to have a weak identity when it can be directly accessed across application domain boundaries.
+    /// A thread that tries to acquire a lock on an object that has a weak identity can be blocked by a second thread in
+    /// a different application domain that has a lock on the same object.
     /// </summary>
     [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
     public sealed class DoNotLockOnObjectsWithWeakIdentityAnalyzer : DiagnosticAnalyzer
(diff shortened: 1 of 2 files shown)


src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotLockOnObjectsWithWeakIdentity.cs
```diff
@@ -51,10 +52,18 @@ public override void Initialize(AnalysisContext analysisContext)
                 compilationStartContext.RegisterOperationAction(context =>
                 {
                     var lockStatement = (ILockOperation)context.Operation;
-                    ITypeSymbol type = lockStatement.LockedValue?.Type;
-                    if (type != null && TypeHasWeakIdentity(type, compilation))
+
+                    if (lockStatement.LockedValue.Kind == OperationKind.InstanceReference)
+                    {
+                        context.ReportDiagnostic(lockStatement.LockedValue.Syntax.CreateDiagnostic(Rule, lockStatement.LockedValue.Syntax.ToString()));
```
Evangelink: There was nothing to help me know whether the argument specified is used for something in particular so I have decided to pass the `this` or `Me` instead of the type in this case. This can be changed to also pass the type if needed.

Evangelink:
Note that the links provided in the issue also mention not to lock on `typeof(X)` when X is publicly available. I can cover this in another commit or PR if you want me to.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotLockOnObjectsWithWeakIdentity.cs
```diff
@@ -51,10 +51,18 @@ public override void Initialize(AnalysisContext analysisContext)
                 compilationStartContext.RegisterOperationAction(context =>
                 {
                     var lockStatement = (ILockOperation)context.Operation;
-                    ITypeSymbol type = lockStatement.LockedValue?.Type;
-                    if (type != null && TypeHasWeakIdentity(type, compilation))
+
+                    if (lockStatement.LockedValue.Kind == OperationKind.InstanceReference)
```
Evangelink: I decided not to go for a separate registered action (even if I know syntax access is faster) mainly because this project doesn't reference Visual Basic syntax kinds so I won't be able to also test for the `Me` syntax.

Evangelink:
Out of curiosity, is it just a lack of time (no pressure) or have I missed something preventing this PR to be reviewed?

mavasani:
I will review the PRs later today.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotLockOnObjectsWithWeakIdentity.cs
mavasani: Can you make the check even more explicit/strict here by validating the `ReferenceKind` of the instance reference?

```csharp
if (lockStatement.LockedValue is IInstanceReferenceOperation instanceReference &&
    instanceReference.ReferenceKind == InstanceReferenceKind.ContainingTypeInstance)
```

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotLockOnObjectsWithWeakIdentity.cs
mavasani: Seems fine to me.

mavasani:
@Evangelink Overall LGTM. If you address the suggestion, I can approve and merge the PR. Kindly file an issue on https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2002-do-not-lock-on-objects-with-weak-identity?view=vs-2019 to update the documentation.

Evangelink:
Done for both.

https://github.com/MicrosoftDocs/visualstudio-docs/issues/4062

PR #2871, opened 2019-09-26
Replace WellKnownTypes with WellKnownTypeProvider

### LINKED ISSUE #2132: Consider replacing WellKnownTypes with WellKnownTypeProvider everywhere

Evangelink:
**Description of the changes**

The idea is to replace calls to the simple wrapper `WellKnownTypes` to a cached provider, reusing the one already in place called `WellKnownTypeProvider`.

**Open questions**

~~I am having a couple of questions I'd like to have some feedback upon before moving on with the finalization of this PR.~~

- ~~I think that all type names should not be cached by default in the ctor but be created in a lazy way so that we avoid processing and loading lots of types in memory if no analyzer using them is turned on. WDYT?~~

- ~~Before I finish adding all the comments and so on, I am actually wondering whether it's worth manually adding the properties every time we need it or whether it would be better to have some auto-generated partial class from the constants defined in `WellKnownTypeNames`.~~

- ~~I have changed the namespace of `WellKnownTypeProvider` from `Microsoft.CodeAnalysis.FlowAnalysis.DataFlow` to `Analyzer.Utilities` to mimic what's done in `WellKnownTypes` and also because this class is no longer specific to C# nor to the dataflow analysis. Is that ok?~~

- ~~I have kept all local variables but I am not sure what's your preferred coding style so if you want me to remove the local variables and to always call directly the property from the provider I can do!~~

**Closed Issues**

Fix #2132


dotpaul:
Thanks for the PR, @Evangelink !

> * I think that all type names should not be cached by default in the ctor but be created in a lazy way so that we avoid processing and loading lots of types in memory if no analyzer using them is turned on. WDYT?

Agree, we should be lazy.

> * Before I finish adding all the comments and so on, I am actually wondering whether it's worth manually adding the properties every time we need it or whether it would be better to have some auto-generated partial class from the constants defined in `WellKnownTypeNames`.

I've just been using the `WellKnownTypeProvider.TryGetTypeByMetadataName()` method myself, and skipping having properties for each type.  Manually adding a property for each type is kinda annoying.  I don't think we need properties at all.

What do you think @mavasani ?  I think you had some other ideas for more convenient methods as well.

> * I have changed the namespace of `WellKnownTypeProvider` from `Microsoft.CodeAnalysis.FlowAnalysis.DataFlow` to `Analyzer.Utilities` to mimic what's done in `WellKnownTypes` and also because this class is no longer specific to C# nor to the dataflow analysis. Is that ok?

@mavasani, have an opinion on namespaces?

> * I have kept all local variables but I am not sure what's your preferred coding style so if you want me to remove the local variables and to always call directly the property from the provider I can do!

Local variables are fine in my opinion.

mavasani:
> I've just been using the WellKnownTypeProvider.TryGetTypeByMetadataName() method myself, and skipping having properties for each type. Manually adding a property for each type is kinda annoying. I don't think we need properties at all.

Lets keep the properties only for types that are used extremely often, so the list should be minimal. I would stick with only the ones that are already there and not add any new ones. I added them for types that were accessed a bunch in our Dataflow analysis components

mavasani:
> I have changed the namespace of WellKnownTypeProvider from Microsoft.CodeAnalysis.FlowAnalysis.DataFlow to Analyzer.Utilities to mimic what's done in WellKnownTypes and also because this class is no longer specific to C# nor to the dataflow analysis. Is that ok?

Seems fine to me as well

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerAttributeAnalyzer.cs
```diff
@@ -54,7 +54,8 @@ public override void Initialize(AnalysisContext context)
         [SuppressMessage("AnalyzerPerformance", "RS1012:Start action has no registered actions.", Justification = "Method returns an analyzer that is registered by the caller.")]
         protected override DiagnosticAnalyzerSymbolAnalyzer GetDiagnosticAnalyzerSymbolAnalyzer(CompilationStartAnalysisContext compilationContext, INamedTypeSymbol diagnosticAnalyzer, INamedTypeSymbol diagnosticAnalyzerAttribute)
         {
-            var attributeUsageAttribute = WellKnownTypes.AttributeUsageAttribute(compilationContext.Compilation);
+            var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation);
+            var attributeUsageAttribute = wellKnownTypeProvider.AttributeUsageAttribute;
```
mavasani: Can we avoid requiring each analyzer to have to get an instance of the provider? We can define an extension method `GetOrCreateWellKnownType` in [CompilationExtensions.cs](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/Extensions/CompilationExtensions.cs) that internally uses the WellKnownProvider so the analyzer can write this as:

```csharp
compilationContext.Compilation.GetOrCreateWellKnownType(WellKnownTypeNames.AttributeUsageAttribute);
```

Analyzers that need to pass around the well known type provider can explicitly invoke `WellKnownTypeProvider.GetOrCreate` to get the instance.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -30,7 +30,11 @@ private WellKnownTypeProvider(Compilation compilation)
             GenericTask = GetTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksGenericTask);
             CollectionTypes = GetWellKnownCollectionTypes();
             SerializationInfo = GetTypeByMetadataName(WellKnownTypeNames.SystemRuntimeSerializationSerializationInfo);
-            GenericIEquatable = GetTypeByMetadataName(WellKnownTypeNames.SystemIEquatable1);
+            Array = GetTypeBySpecialType(SpecialType.System_Array);
```
Evangelink: I have decided to add the special types as property too. If you prefer them to not be properties here I can create another commit to change it.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -108,12 +137,42 @@ public bool TryGetTypeByMetadataName(string fullTypeName, out INamedTypeSymbol n
...
+        /// <summary>
+        /// Attempts to get the type by its special type.
+        /// </summary>
+        /// <param name="specialType">ID of the special runtime type.</param>
+        /// <param name="namedTypeSymbol">Named type symbol, if any.</param>
+        /// <returns>True if found in the compilation, false otherwise.</returns>
+        private bool TryGetTypeBySpecialType(SpecialType specialType, out INamedTypeSymbol namedTypeSymbol)
+        {
+            namedTypeSymbol = _fullNameToTypeMap.GetOrAdd(
+                specialType.ToString(),
```
Evangelink: Use the name of the enum as unique name for the type.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -135,5 +135,25 @@ private static bool Inherits(ITypeSymbol symbol, ITypeSymbol baseType)
             Debug.Assert(baseType.Equals(baseType.OriginalDefinition));
             return symbol?.OriginalDefinition.Inherits(baseType) ?? false;
         }
+
+        private static ImmutableHashSet<INamedTypeSymbol> GetIImmutableInterfaces(Compilation compilation)
```
Evangelink: This is used in only one place so I decided to move the code here. If you prefer, I can add it to the provider.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotRaiseExceptionsInUnexpectedLocations.cs
```diff
@@ -167,21 +167,25 @@ private static List<MethodCategory> GetMethodCategories(Compilation compilation)
             var methodCategories = new List<MethodCategory> {
                 new MethodCategory(IsPropertyGetter, true,
                     PropertyGetterRule,
-                    WellKnownTypes.InvalidOperationException(compilation), WellKnownTypes.NotSupportedException(compilation)),
+                    compilation.GetTypeByMetadataName(typeof(System.InvalidOperationException).FullName),
```
Evangelink: Some types are used multiple time so I could create a local instance instead of doing the call to end up with the same instance. I am not sure whether this is a useless micro-optimization.

src/Utilities/Compiler/Analyzer.Utilities.projitems
```diff
@@ -75,8 +74,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)FxCopWellKnownDiagnosticTags.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)UnusedValue.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)WellKnownTypeNames.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)WellKnownTypeProvider.cs" />
```
Evangelink: Make sure the type is always compiled otherwise we have some errors.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
```diff
@@ -62,5 +62,14 @@ internal static INamedTypeSymbol GetVisibleTypeByMetadataName(this Compilation c
...
             return publicType ?? anyType;
         }
+
+        /// <summary>
+        /// Gets a type by its full type name and cache it at the compilation level.
+        /// </summary>
+        /// <param name="compilation">The compilation.</param>
+        /// <param name="fullTypeName">Namespace + type name, e.g. "System.Exception".</param>
+        /// <returns>The <see cref="INamedTypeSymbol"/> if found, null otherwise.</returns>
+        internal static INamedTypeSymbol GetOrCreateWellKnownType(this Compilation compilation, string fullTypeName) =>
```
Evangelink: Extension method as suggested by @mavasani

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerAttributeAnalyzer.cs
```diff
@@ -54,7 +54,7 @@ public override void Initialize(AnalysisContext context)
         [SuppressMessage("AnalyzerPerformance", "RS1012:Start action has no registered actions.", Justification = "Method returns an analyzer that is registered by the caller.")]
         protected override DiagnosticAnalyzerSymbolAnalyzer GetDiagnosticAnalyzerSymbolAnalyzer(CompilationStartAnalysisContext compilationContext, INamedTypeSymbol diagnosticAnalyzer, INamedTypeSymbol diagnosticAnalyzerAttribute)
         {
-            var attributeUsageAttribute = WellKnownTypes.AttributeUsageAttribute(compilationContext.Compilation);
+            var attributeUsageAttribute = compilationContext.Compilation.GetTypeByMetadataName(typeof(System.AttributeUsageAttribute).FullName);
```
mavasani: I think we should prioritize on readability and consistency for this as almost every analyzer would need be fetching well known types.
This change seems like the readability has gone down here as each analyzer needs to specify the name of the type. It also introduces the inconsistency that certain well known type names are defined in `WellKnownTypeNames`, but not all. I would instead prefer the following:

1. Consistency: Every queried well known type must have a corresponding constant/static string in `WellKnownTypeNames`. So in this case there will be a public field named `AttributeUsageAttribute` in `WellKnownTypeNames`.
2. Analyzer implementation will be one of the following buckets:
    1. `compilationContext.Compilation.GetTypeByMetadataName(WellKnownTypeNames.AttributeUsageAttribute);`: Used when analyzer wants to fetch a well known type without caching.
   2. `compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.AttributeUsageAttribute, out var wellKnownType);`: Used when analyzer wants to fetch well known type with caching and bail out on null value.
   3. `compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.AttributeUsageAttribute);`: Used when analyzer wants to fetch well known type with caching and continue even on null value.
   4. `var provider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation); 
 provider.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.AttributeUsageAttribute, out var wellKnownType);`: Used when analyzer needs to query large number of well known types  with caching and/or pass around the provider to different components (such as DFA APIs), but bail out on null return value of type.
   5. `var provider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation); 
 provider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.AttributeUsageAttribute);`: Used when analyzer needs to query large number of well known types  with caching and/or pass around the provider to different components (such as DFA APIs), but not bail out on null return value of type.

I am fine if we don't introduce 3. and 5. from above as well.

src/Microsoft.CodeQuality.Analyzers/CSharp/QualityGuidelines/CSharpRemoveEmptyFinalizersAnalyzer.cs
```diff
@@ -81,7 +81,7 @@ private static bool IsConditionalInvocation(InvocationExpressionSyntax invocatio
                 return false;
             }
 
-            var conditionalAttributeSymbol = WellKnownTypes.ConditionalAttribute(semanticModel.Compilation);
+            var conditionalAttributeSymbol = semanticModel.Compilation.GetTypeByMetadataName(typeof(System.Diagnostics.ConditionalAttribute).FullName);
```
mavasani: Yep, more changes I look at I don't like this. Its lot more verbose and error prone.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -57,11 +57,11 @@ public override void Initialize(AnalysisContext analysisContext)
...
                 (context) =>
                 {
-                    INamedTypeSymbol iCollectionType = WellKnownTypes.ICollection(context.Compilation);
-                    INamedTypeSymbol genericICollectionType = WellKnownTypes.GenericICollection(context.Compilation);
-                    INamedTypeSymbol arrayType = WellKnownTypes.Array(context.Compilation);
-                    INamedTypeSymbol dataMemberAttribute = WellKnownTypes.DataMemberAttribute(context.Compilation);
-                    ImmutableHashSet<INamedTypeSymbol> immutableInterfaces = WellKnownTypes.IImmutableInterfaces(context.Compilation);
+                    INamedTypeSymbol iCollectionType = context.Compilation.GetTypeByMetadataName(WellKnownTypeNames.SystemCollectionsICollection);
+                    INamedTypeSymbol genericICollectionType = context.Compilation.GetTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericICollection1);
+                    INamedTypeSymbol arrayType = WellKnownTypeProvider.GetOrCreate(context.Compilation).SystemArray;
```
mavasani: This seems to have introduced inconsistency now. It is unclear why you would use `compilation.GetTypeByMetadataName` for one well known type and `WellKnownTypeProvider.GetOrCreate` for another. Previous implementation was much more readable and consistent.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotRaiseExceptionsInUnexpectedLocations.cs
```diff
@@ -282,14 +286,14 @@ private static bool IsGetHashCodeInterfaceImplementation(IMethodSymbol method, C
             {
                 // Substitute the type of the first parameter of Equals in the generic interface and then check if that
                 // interface method is implemented by the given method.
-                INamedTypeSymbol iEqualityComparer = WellKnownTypes.GenericIEqualityComparer(compilation);
+                INamedTypeSymbol iEqualityComparer = compilation.GetTypeByMetadataName(typeof(IEqualityComparer<>).FullName);
```
mavasani: Lot of inconsistency in the way well known types are accessed in this file, and now the analyzer also needs to know how to access generic types.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerAttributeAnalyzer.cs
Evangelink: +1 for always using `WellKnownTypeNames`.

Regarding point 2i, I am not sure why you would want to access a type without wanting it be to cached. Indeed, when writing an analyzer you don't really know whether some other will need to access this type. And if you go to the extreme and only look at your analyzer then you might miss some of the benefits of the caching.

Evangelink:
I see two ways to go forward.

1/ I am getting back to the initial suggestion about having all types from `WellKnownTypeNames` automatically available as properties on a partial instance of `WellKnownTypeProvider` + some hard-coded properties for things like `SpecialType`. This would mean that usage would always be `WellKnownTypeProvider.GetOrCreate(compilation).XXX` where `XXX` is the type name to access.

2/ We get rid of the properties on `WellKnownTypeProvider` and instead of going for a functional name we have one caching method with different overloads (string, `SpecialType`...)

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerAttributeAnalyzer.cs
mavasani: Sounds reasonable - we could start with just `TryGetOrCreateTypeByMetadataName` methods as our preferred API.

Note that the core `Compilation.GetTypeByMetadataName` API exposed by the compiler is also [caching](http://source.roslyn.io/#Microsoft.CodeAnalysis/Compilation/Compilation.cs,947) well known types, but it is a bounded cache (size 50). IMO, we have not seen any perf traces showing that this cache is not adequate. In general, caching here is just a micro-optimization (whether it is done by the compiler or our well known type provider), which is secondary benefit. The most important aspect here is keeping the analyzer code most readable and have a consistent way to do so.

mavasani:
I think 2. is probably preferable, with the additional requirement that all well known type names be also added to `WellKnownTypeNames`.

Evangelink:
@mavasani Out of curiosity do you know what's the difference between accessing to a named type from its full name and from the special type? I can see that the special type contains the collections but we are getting them from their string name. If the difference is minor maybe we could access to array, string, intptr... from their string name and not special type so that we can have only one "unified" way to access to the type.

For now, I will try to move their usage directly to the calling spot instead of being on the provider.

mavasani:
> Out of curiosity do you know what's the difference between accessing to a named type from its full name and from the special type? I can see that the special type contains the collections but we are getting them from their string name. If the difference is minor maybe we could access to array, string, intptr

Compiler has baked in knowledge of special types, cache the named types for these in separate caches and also exposes a special enum for these: http://source.roslyn.io/#Microsoft.CodeAnalysis/SpecialType.cs,5b11a29d644330dc. These special types should always be queried with [GetSpecialType API](http://source.roslyn.io/#Microsoft.CodeAnalysis/Compilation/Compilation.cs,804).

Evangelink:
@mavasani Ok so ideally we should not use `compilation.GetTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericICollection1)` but instead use `compilation.GetSpecialType(SpecialType.System_Collections_Generic_ICollection_T)`. Is that right?

mavasani:
> Ok so ideally we should not use compilation.GetTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericICollection1) but instead use compilation.GetSpecialType(SpecialType.System_Collections_Generic_ICollection_T). Is that right?

Yes, correct.

Evangelink:
@mavasani I have applied all changes we discussed about. Could you have a look when you have some time?

I am still having some issues with `RS0016` and `RS0017` complaining about some members not declared in the API although they were already present when I created this PR. I am not sure if I should add them into `PublicAPI.Unshipped.txt`.

I have noticed that there are quite a lot of places where the type name is hard-coded instead of being declared in the `WellKnownTypeNames`. If you are pleased with the current state of the PR, do you want me to create a separate commit (or PR) to move them to the type names class?

Thank you for your time and explanations!

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotRaiseExceptionsInUnexpectedLocations.cs
```diff
@@ -167,21 +167,25 @@ private static List<MethodCategory> GetMethodCategories(Compilation compilation)
             var methodCategories = new List<MethodCategory> {
                 new MethodCategory(IsPropertyGetter, true,
                     PropertyGetterRule,
-                    WellKnownTypes.InvalidOperationException(compilation), WellKnownTypes.NotSupportedException(compilation)),
+                    compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemInvalidOperationException),
```
mavasani: I see that we use `GetOrCreateTypeByMetadataName` at some places and `GetTypeByMetadataName` at others. Is there a reasoning behind how to chose one over another at these places?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
```diff
@@ -175,7 +176,8 @@ protected override void EscapeValueForParameterPointsToLocationOnExit(IParameter
             {
                 Debug.Assert(!escapedLocations.IsEmpty);
                 Debug.Assert(parameter.RefKind != RefKind.None);
-                var escapedDisposableLocations = escapedLocations.Where(l => l.LocationTypeOpt?.IsDisposable(WellKnownTypeProvider.IDisposable) == true);
+                var escapedDisposableLocations =
+                    escapedLocations.Where(l => l.LocationTypeOpt?.IsDisposable(WellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemIDisposable)) == true);
```
mavasani: Can you extract out `WellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemIDisposable)` into a local?

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -44,51 +35,6 @@ static WellKnownTypeProvider CreateWellKnownTypeProvider(Compilation compilation
 
         public Compilation Compilation { get; }
 
-        /// <summary>
```
mavasani: All the properties here are the ones that are very commonly accessed in dataflow analysis. Given that this well known type provider is not specific to flow analysis anymore, it makes sense to remove them from here. However, I think we should still move all of these to [DataflowOperationVisitor](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs). They could be implemented in this file as following:

```csharp
protected INamedTypeSymbol Exception { get; } = WellKnownTypeProvider.GetOrCreateTypeByMetadataName(SystemException);
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
mavasani: You won't need to extract to a local if you implemented the suggestion https://github.com/dotnet/roslyn-analyzers/pull/2871/files#r331985326

src/Utilities/Compiler/WellKnownTypeProvider.cs
mavasani: I can also do this as a follow-up change if you want to avoid another iteration.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotRaiseExceptionsInUnexpectedLocations.cs
Evangelink: All calls to the `WellKnownTypes` and `WellKnownTypeProvider` have been converted to `GetOrCreateTypeByMetadataName` while all previous calls to `GetTypeByMetadataName` were kept as they were. 

I personally don't see any "good" reason to keep calls to `GetTypeByMetadataName` and I am fine to add one more commit taking care of that if that's aligned with your way of seeing the code after this PR.

src/Utilities/Compiler/WellKnownTypeProvider.cs
Evangelink: That's fine, I am taking care to add one more commit for this change. Thank you for the feedback.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -185,6 +185,19 @@ protected DataFlowOperationVisitor(TAnalysisContext analysisContext)
         {
             DataFlowAnalysisContext = analysisContext;
 
+            // All of these named type are very commonly accessed in the dataflow analysis so we want to ensure
```
mavasani: Thank you!

dotpaul:
:ship:

Thanks again, @Evangelink !  I've retargetted this PR to 2.9.x, and changes will also eventually be merged into the master branch.

Issue #692, opened 2015-12-02
Port FxCop rule CA2200: RethrowToPreserveStackDetails


lgolding:
**Title:** Rethrow to preserve stack details

**Description:**

An exception is rethrown and the exception is explicitly specified in the throw statement. If an exception is rethrown by specifying the exception in the throw statement, the list of method calls between the original method that threw the exception and the current method is lost.

**Dependency:** None

**Notes:**

mavasani:
This needs a fixer.

mavasani:
Fixer added with https://github.com/dotnet/roslyn-analyzers/pull/2873

PR #2873, opened 2019-09-29
Add a fixer for CA2200

goyzhang:
Replace `throw e;` to `throw;`
Reference #692


src/Microsoft.CodeQuality.Analyzers/CSharp/QualityGuidelines/CSharpRethrowToPreserveStackDetails.Fixer.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
+{
+    /// <summary>
+    /// CA2200： Rethrow to preserve stack details
+    /// </summary>
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public sealed class CSharpRethrowToPreserveStackDetailsFixer : RethrowToPreserveStackDetailsFixer
+    {
+        protected async override Task<Document> MakeThrowAsync(Document document, SyntaxNode nodeToReplace, CancellationToken cancellationToken)
```
mavasani: The entire code can be moved to the language agnostic layer by using [SyntaxGenerator.ThrowStatement](http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Editing/SyntaxGenerator.cs,1430) API to generate the throw statement node - it should work for both C# and VB

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/RethrowToPreserveStackDetails.Fixer.cs
```diff
@@ -0,0 +1,41 @@
...
+            // See https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md for more information on Fix All Providers'
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+        public async sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var diagnostic = context.Diagnostics.First();
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+
+            var nodeToReplace = root.FindNode(diagnosticSpan);
```
mavasani: Couple of comments:

1. You don't need to get the first diagnostic or use it's span
2. Please add a null check for nodeToReplace

```csharp
            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
            var nodeToReplace = root.FindNode(context.Span);
            if (nodeToReplace == null)
            {
                return;
            }

            context.RegisterCodeFix(
                CodeAction.Create(...),
                context.Diagnostics);
```

goyzhang: Uh, Ok, I just got it from the boilerplate https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/tutorials/how-to-write-csharp-analyzer-code-fix  and it does the job as well. I will modify it.

PR #2883, opened 2019-10-01
Fix NRE in recently added AssigningSymbolAndItsMemberInSameStatement …

mavasani:
…analyzer

Handle null operation type, which is possible for OperationKind.None
Found while dogfooding the latest package


src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.cs
```diff
@@ -50,29 +50,21 @@ private void AnalyzeAssignment(OperationAnalysisContext context)
             }
 
             // This analyzer makes sense only for reference type objects
-            if (operationTarget.Instance?.Type.IsValueType == true)
+            if (operationTarget.Instance?.Type?.IsReferenceType != true)
```
mavasani: This is the core fix for the NRE. Other change below is just applying an IDE suggestion to use switch expression.

mavasani:
Thanks @dotpaul

PR #2896, opened 2019-10-02
Merge branch 3.x into master

mavasani:
We have accumulated quite a few issues in the repo that are blocked on us moving to VS2019 as the minimum required build (Microsoft.CodeAnalysis 3.0.0 APIs). After an offline chat with @dotpaul, we decided to do the following changes to our branching structure:

1. Merge 3.x branch back into master now, so master branch will target 3.0.0 release next. 
2. Fork 2.9.x branch out of master sources just prior to the 3.x merge back to master.
3. Security team (@dotpaul and @LLLXXXCCC) will continue with their bug fixes and adding/updating security rules in 2.9.x branch. They will drive any required back porting of commits from master to 2.9.x, as appropriate. Additionally, they will own the release timelines and quality of 2.9.x releases going forward, as they should primarily be security rule releases.


azure-pipelines.yml
```diff
@@ -1,10 +1,12 @@
 # Branches that trigger a build on commit
 trigger:
 - master
+- 3.x
```
mavasani: Let me push a separate commit to change this to 2.9.x

mavasani:
Tagging @sharwell @genlu as FYI

PR #2916, opened 2019-10-08
Update CA1001 to stop reporting on types implementing IAsyncDisposable

### LINKED ISSUE #2325: CA1001 isn't aware of IAsyncDisposable
#### Analyzer package

Microsoft.CodeQuality.Analyzers from [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

2.9.1

#### Diagnostic ID

[CA1001](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1001-types-that-own-disposable-fields-should-be-disposable)

#### Repro steps

``` cs
namespace Example
{
    public class Test : IAsyncDisposable
    {
        private Stream _disposableMember;

        public ValueTask DisposeAsync() => _disposableMember.DisposeAsync();
    }
}
```

#### Expected behavior

No code analysis warning

#### Actual behavior

Received CA1001

Evangelink:
Fix #2325 

Note that the documentation will need to be updated.

I will need some help with the unit test, I don't know how to prevent the error about the unknown type `IAsyncDisposable`. Note that I have tried to use parse options with C#8 and Preview but both are failing.

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/TypesThatOwnDisposableFieldsShouldBeDisposable.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/TypesThatOwnDisposableFieldsShouldBeDisposable.cs
@@ -68,10 +68,15 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             public void AnalyzeSymbol(SymbolAnalysisContext symbolContext)
             {
                 INamedTypeSymbol namedType = (INamedTypeSymbol)symbolContext.Symbol;
-                if (namedType.AllInterfaces.Contains(_disposableTypeSymbol) ||
-                    (_asyncDisposableTypeSymbol != null && namedType.AllInterfaces.Contains(_asyncDisposableTypeSymbol)))
+
+                for (int i = 0; i < namedType.AllInterfaces.Length; i++)
                 {
-                    return;
+                    var currentInterface = namedType.AllInterfaces[i];
+                    if (currentInterface != null &&
+                        (currentInterface.Equals(_disposableTypeSymbol) || currentInterface.Equals(_asyncDisposableTypeSymbol)))
+                    {
+                        return;
+                    }
                 }
 
                 IEnumerable<IFieldSymbol> disposableFields = from member in namedType.GetMembers()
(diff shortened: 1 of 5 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/TypesThatOwnDisposableFieldsShouldBeDisposable.cs
```diff
@@ -41,58 +41,64 @@ public override void Initialize(AnalysisContext analysisContext)
...
-                _disposableTypeSymbol = disposableTypeSymbol;
+                _disposableTypeSymbol = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemIDisposable);
+                _asyncDisposableTypeSymbol = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemIAsyncDisposable);
             }
 
             public void AnalyzeSymbol(SymbolAnalysisContext symbolContext)
             {
                 INamedTypeSymbol namedType = (INamedTypeSymbol)symbolContext.Symbol;
-                if (!namedType.AllInterfaces.Contains(_disposableTypeSymbol))
+                if (namedType.AllInterfaces.Contains(_disposableTypeSymbol) ||
```
Evangelink: I have added the second part of the `if` and reverted the condition to avoid deep nesting. The rest of the code of this method didn't changed except from the indentation.

mavasani: More efficient to write a single loop that walks through all interfaces and performs null check and then symbol equality check with both disposable type symbols. Note that `AllInterfaces` is an array, not a hashset.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/TypesThatOwnDisposableFieldsShouldBeDisposableTests.cs
```diff
@@ -215,6 +217,26 @@ public class Class1
...
+        public void CA1001CSharpTestWithIAsyncDisposable()
+        {
+            VerifyCSharp(@"
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace ClassLibrary1
+{
+    public class Class1 : IAsyncDisposable
```
mavasani: It should be fine to add the following definition to test code itself:

```csharp
namespace System
{
   interface IAsyncDisposable
   {
      ValueTask DisposeAsync();
   }
}
```

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/TypesThatOwnDisposableFieldsShouldBeDisposable.cs
Evangelink: You are right I went for the easier to right not the optimal for the situation.

PR #2918, opened 2019-10-09
Fix CA1721 - Do not report on overriden members

### LINKED ISSUE #2914: CA1721 should not raise on overrides
#### Analyzer package

#### Package Version

v2.9.6 (Latest)

#### Diagnostic ID

Example: [CA1721](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1721-property-names-should-not-match-get-methods)

#### Repro steps

```csharp
public class MyBase
{
    public object Content { get; }
    public virtual object GetContent() => null;
}

public class Foo
{
	// Issue is raised here but as I cannot change anything in here
    // there should be no issue raised.
    public override object GetContent() => this;
}
```

#### Expected behavior

No CA1721 is raised on `Foo.GetContent()` as there is no way to solve the issue except editing the base class which might not be under our control (cf `RadOpenFolderDialog` from Telerik).

#### Actual behavior

The issue is raised on `Foo.GetContent()`

I will try and have a PR fixing this issue if you are fine with it.

Evangelink:
When a base class with virtual members is not complying with the rule (i.e. has a property and a get method with the same name) and your derived class override one or two of the member you will get the issue raised upon your override. Sadly there is nothing much you can do if you don't own the base class so the analyzer should not raise an issue here.

I have tried to add test for C# and VB.NET for all the cases I could think of.

Fix #2914

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
@@ -94,7 +94,9 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                         continue;
                     }
 
-                    // Ignore declared type that overrides a base member
+                    // We only want to report an issue when the user is free to update the member.
+                    // So if the method or/and the property is an override and base clase defines both members we bail out
+                    // but if base defines only one of the member and derived defines the other we still want to raise an issue
                     if (symbol.IsOverride)
                     {
                         continue;
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
@@ -50,6 +50,16 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             string identifier;
             var symbol = context.Symbol;
 
+            // We only want to report an issue when the user is free to update the member.
+            // This method will be called for both the property and the method so we can bail out
+            // when the member (symbol) is an override.
+            // Note that in the case of an override + a local declaration, the issue will be raised from
+            // the local declaration.
+            if (symbol.IsOverride)
+            {
+                return;
+            }
+
             // Bail out if the method/property is not exposed (public, protected, or protected internal) by default
             var configuredVisibilities = context.Options.GetSymbolVisibilityGroupOption(Rule, SymbolVisibilityGroup.Public, context.CancellationToken);
             if (!configuredVisibilities.Contains(symbol.GetResultantVisibility()))
@@ -81,14 +91,6 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                 var exposedMembers = type.GetMembers(identifier).Where(member => configuredVisibilities.Contains(member.GetResultantVisibility()));
                 foreach (var member in exposedMembers)
                 {
-                    // We only want to report an issue when the user is free to update the member.
-                    // So we can bail out when the property or method (symbol) is override and the other member (member)
-                    // is not local (i.e. symbol and member are declared
-                    if (symbol.IsOverride && member.ContainingType.Equals(type))
-                    {
-                        continue;
-                    }
-
                     // Ignore Object.GetType, as it's commonly seen and Type is a commonly-used property name.
                     if (member.ContainingType.SpecialType == SpecialType.System_Object &&
                         member.Name == nameof(GetType))
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
@@ -41,7 +41,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             analysisContext.EnableConcurrentExecution();
             analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
-            // Analyze properties, methods 
+            // Analyze properties, methods
             analysisContext.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.Property, SymbolKind.Method);
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
```diff
@@ -94,6 +94,12 @@ private static void AnalyzeSymbol(SymbolAnalysisContext context)
                         continue;
                     }
 
+                    // Ignore declared type that overrides a base member
+                    if (symbol.IsOverride)
```
mavasani: I am not sure if we want to ignore all overrides. For example, if base type defines a virtual/abstract method and derive type defines an override for it and also defines a new property with a name that violates it, we still want to flag the derive type's property. I think we only want to not flag when both the property and method symbol are override.

Evangelink: Looks like the comment is not good enough. Basically if you look at the test you'll see that if you override the property or the method or both no issue is raised. On the other hand if you override one of the 2 and add the complementary member you will have an issue raised.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethodsTests.cs
```diff
@@ -376,6 +376,151 @@ End Function
...
+        GetCA1721BasicResultAt(line: 3, column: 42, identifierName: "Value", otherIdentifierName: "GetValue"));
+        }
+
+        [Fact, WorkItem(2914, "https://github.com/dotnet/roslyn-analyzers/issues/2914")]
+        public void CA1721_OverrideWithLocalMemberDiagnostic()
+        {
+            VerifyCSharp(@"
+public class BaseClass1
+{
+    public virtual int Value { get; }
```
mavasani: Can you add a test where the virtual in base type is a method instead of a property?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
mavasani: I understand why this works, just that it wasn’t clear from a quick glance. Is there a reason to have your newly added check inside the loop instead of bailing out right at the start of the method?

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethodsTests.cs
Evangelink: See tests from line 466+. `BaseClass2` has a virtual method and `C2` adds the property.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
Evangelink: > It’s not the comment but the implementation that doesn’t seem tight enough. I would have expected the newly added if statement condition to check if both symbol.IsOverride and member.IsOverride is true, not just one.

Well if I have both `symbol.IsOverride` and `member.IsOverride` then I will keep reporting on cases where the base class defines both the property and the method and I override only one of them.
I will make some more tests to see if I can somehow find some more restrictive condition to add.

> I understand why your current implementation works, just that it wasn’t clear from a quick glance. Is there a reason to have your newly added check inside the loop instead of bailing out right at the start of the method?

Nope no good reason. I will fix it.

Evangelink: Alright so according to this "truth table":

```
local P + nothing => no issue
local P + local M => issue
local P + base M => issue
local P + override M => issue
==> issue when local P + whatever M

override P + nothing => no issue
override P + local M => issue
override P + base M => no issue
override P + override M => no issue
==> issue when override P + local M

local M + nothing => no issue
local M + local P => issue
local M + base P => issue
local M + override P => issue
==> issue when local M + whatever P

override M + nothing => no issue
override M + local P => issue
override M + base P => no issue
override M + override P => no issue
==> issue when override M + local M
```

and based on the fact that the analyzer will kick-in for both the method and the property (when they are present), we could either keep the if located here update it to `symbol.IsOverride && member.ContainingType.Equals(type)` or we can move it to the beginning of the `AnalyzerSymbol` method. Indeed, when the member is an override, the only case where we want to report is when the other member is declared locally AND so the second check will take care of it.

Let me know what's your preference.

mavasani: Thanks for the detailed explanation. I am fine with either approach, with a weaker preference towards just bailing out up front for symbols that are overrides.

Evangelink: I will update to add the second part to the if then :D
Not trying to convince you of going for the other solution but do you think the overhead of the various visibility, loop and member checks are neglectable?

mavasani: Yes, I think performance is not really a concern here. Best to come up with a clean, maintainable solution. I am fine with either of your proposed routes. Thanks!

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/PropertyNamesShouldNotMatchGetMethods.cs
```diff
@@ -81,6 +81,14 @@ private static void AnalyzeSymbol(SymbolAnalysisContext context)
                 var exposedMembers = type.GetMembers(identifier).Where(member => configuredVisibilities.Contains(member.GetResultantVisibility()));
                 foreach (var member in exposedMembers)
                 {
+                    // We only want to report an issue when the user is free to update the member.
+                    // So we can bail out when the property or method (symbol) is override and the other member (member)
+                    // is not local (i.e. symbol and member are declared
+                    if (symbol.IsOverride && member.ContainingType.Equals(type))
```
mavasani: The comment says "is not local" but the check seems to validate it is local?

Evangelink: You are absolutely right! Sadly I cannot simply negate the equal condition as we are looping through the different base types... And for the same reason this check is not enough `symbol.IsOverride && (!member.ContainingType.Equals(symbol.ContainingType) || member.IsOverride)` as we end up reporting multiple times for the same member when the member is overriden and a local member is declared. The only "workaround" I can think of is to access the original declaration to do the comparison but that's adding quite a lot of complexity IMO.

I am reverting to the simple `symbol.IsOverride` check and moving it to the top of the class with an updated comment.

PR #2924, opened 2019-10-10
Somewhat cache DFA results

### LINKED ISSUE #2900: Build hangs after upgrading to 2.9.5
#### Analyzer package

#### Package Version

v2.9.5 (Latest)

#### Diagnostic ID

N/A

#### Repro steps

After upgrading from 2.9.4 to 2.9.5 one of my builds hangs. How do I diagnose the issue?

#### Expected behavior
Build completes.

#### Actual behavior
Build hangs and never completes.

dotpaul:
Partially fixes #2900 

From my observations with slow analysis with DoNotHardCodeEncryptionKey, the root cause is a combination of ValueContent analysis can be slow and ValueContentAnalysisResult cache misses, due to the limited DFA result cache. That causes ValueContent analysis to be invoked multiple times.

This somewhat mitigates the problem, by caching analysis on a per-analyzer-operationblock basis, in TaintedData analyzers that use ValueContentAnalysis.

Longer term we should see if we can come up with a way of caching DFA results on a per compilation basis without memory leaks, and see what we can do to optimize ValueContent analysis.

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/Security/SourceTriggeredTaintedDataAnalyzerBase.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Security/SourceTriggeredTaintedDataAnalyzerBase.cs
@@ -73,6 +73,11 @@ namespace Microsoft.NetCore.Analyzers.Security
                                 return;
                             }
 
+                            PointsToAnalysisResult pointsToResult = null;
+                            bool pointsToComputed = false;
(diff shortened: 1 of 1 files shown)


mavasani:
> Longer term we should see if we can come up with a way of caching DFA results on a per compilation basis without memory leaks, and see what we can do to optimize ValueContent analysis.

I am still concerned about significant memory usage that comes along with holding all DFA results for a compilation in memory. IMO, we need some hybrid solution here.

PR #2932, opened 2019-10-12
Improve WellKnownTypeProvider handling of duplicate types

sharwell:


src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -46,7 +48,47 @@ public bool TryGetOrCreateTypeByMetadataName(string fullTypeName, out INamedType
...
+                fullyQualifiedMetadataName =>
+                {
+                    // Caching null results in our cache is intended.
+                    var type = Compilation.GetTypeByMetadataName(fullyQualifiedMetadataName)
+                        ?? Compilation.Assembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
+                    if (type is null)
+                    {
+                        foreach (var module in Compilation.Assembly.Modules)
+                        {
+                            foreach (var referencedAssembly in module.ReferencedAssemblySymbols)
```
Evangelink: I am not sure whether there is a real impact or anything but I thought that it was better to use `for` loop instead of `foreach` loop for `ImmutableArray`?

sharwell: ImmutableArray has a struct enumerator, so it should be identical.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -46,7 +48,47 @@ public bool TryGetOrCreateTypeByMetadataName(string fullTypeName, out INamedType
...
+                            {
+                                var currentType = referencedAssembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
```
mavasani: Observation (not a change request): Any API based analyzer that will bail out for a compilation due to the API not being present in the compilation with end up executing `GetTypeByMetadataName` for each module's assembly references. Hopefully, the API invocation is cheap when the assembly has no such API. If not, we may have to tune this in future.

PR #2938, opened 2019-10-15
CA1305: Do not raise for bool.ToString() and char.ToString()

### LINKED ISSUE #2394: Disable CA1305 for Boolean.ToString
#### Analyzer package

#### Package Version
v2.9.2

#### Diagnostic ID
[CA1305: Specify IFormatProvider](https://docs.microsoft.com/de-de/visualstudio/code-quality/ca1305-specify-iformatprovider?view=vs-2019)

#### Repro steps
Try the following code snippet:

```
Sub Test
    Dim TestBool As Boolean
    Dim TestString As String

    TestString = TestBool.ToString
End Sub
```

#### Expected behavior
`TestBool.ToString` is not underlined by code analyzer

#### Actual behavior
`TestBool.ToString` is underlined by code analyzer

Citing [https://docs.microsoft.com/de-de/dotnet/api/system.boolean.tostring?view=netframework-4.8#System_Boolean_ToString_System_IFormatProvider_](url):
**"The provider parameter is reserved. It does not participate in the execution of this method. This means that the Boolean.ToString(IFormatProvider) method, unlike most methods with a provider parameter, does not reflect culture-specific"**
So, this analyzer warning seems to be useless and confusing. 

Edit: This issue applies to the method `Char.ToString` as well.

Evangelink:
Add `bool` and `char` to the exclusion list as the `IFormatProvider` parameter is ignored.

Fix #2394

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
@@ -83,6 +83,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 }
 
                 var objectType = csaContext.Compilation.GetSpecialType(SpecialType.System_Object);
+                var charType = csaContext.Compilation.GetSpecialType(SpecialType.System_Char);
+                var boolType = csaContext.Compilation.GetSpecialType(SpecialType.System_Boolean);
                 var stringType = csaContext.Compilation.GetSpecialType(SpecialType.System_String);
                 var stringFormatMembers = stringType?.GetMembers("Format").OfType<IMethodSymbol>();
 
@@ -132,7 +134,9 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                     if (targetMethod.IsGenericMethod || targetMethod.ContainingType == null || targetMethod.ContainingType.IsErrorType() ||
                         (targetMethod.ContainingType != null &&
                          (activatorType != null && activatorType.Equals(targetMethod.ContainingType)) ||
-                         (resourceManagerType != null && resourceManagerType.Equals(targetMethod.ContainingType))))
+                         (resourceManagerType != null && resourceManagerType.Equals(targetMethod.ContainingType)) ||
+                         (charType != null && charType.Equals(targetMethod.ContainingType)) ||
+                         (boolType != null && boolType.Equals(targetMethod.ContainingType))))
                     {
                         return;
                     }
--- a/src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
@@ -131,12 +131,15 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                     var targetMethod = invocationExpression.TargetMethod;
 
                     #region "Exceptions"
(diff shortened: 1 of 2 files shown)


src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -132,7 +134,9 @@ public override void Initialize(AnalysisContext analysisContext)
                     if (targetMethod.IsGenericMethod || targetMethod.ContainingType == null || targetMethod.ContainingType.IsErrorType() ||
                         (targetMethod.ContainingType != null &&
                          (activatorType != null && activatorType.Equals(targetMethod.ContainingType)) ||
-                         (resourceManagerType != null && resourceManagerType.Equals(targetMethod.ContainingType))))
+                         (resourceManagerType != null && resourceManagerType.Equals(targetMethod.ContainingType)) ||
+                         (charType != null && charType.Equals(targetMethod.ContainingType)) ||
```
Evangelink: The if condition starts to be a bit ugly IMO maybe that would be worth creating some method, WDYT?

sharwell: That might be nice, but I don't think we need to make that change now. Keeping the current form isn't beautiful but I can certainly review it quickly. 😄

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/SpecifyIFormatProviderTests.cs
```diff
@@ -859,6 +859,50 @@ End Sub
 End Class");
         }
 
+        [Fact]
+        [WorkItem(2394, "https://github.com/dotnet/roslyn-analyzers/issues/2394")]
+        public void CA1305_BoolToString_NoDiagnostics()
```
sharwell: 💡 Can you add a test for `string` inputs as well?

Evangelink: Good catch! This does raises an issue. I will also add a check for `string`.

src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -129,10 +131,15 @@ public override void Initialize(AnalysisContext analysisContext)
                     var targetMethod = invocationExpression.TargetMethod;
 
                     #region "Exceptions"
+                    const string ToStringMethodName = "ToString";
                     if (targetMethod.IsGenericMethod || targetMethod.ContainingType == null || targetMethod.ContainingType.IsErrorType() ||
                         (targetMethod.ContainingType != null &&
                          (activatorType != null && activatorType.Equals(targetMethod.ContainingType)) ||
-                         (resourceManagerType != null && resourceManagerType.Equals(targetMethod.ContainingType))))
+                         (resourceManagerType != null && resourceManagerType.Equals(targetMethod.ContainingType)) ||
+                         (targetMethod.Name == ToStringMethodName &&
```
Evangelink: I needed to add the method name check for the `string` type so I thought that it was better to also use it for `bool` and `char`.

Bouke:
If and when might this PR make it into a release? The PR is merged into master, but releases seem to be cut from the `2.9.x` tree. If there is some branching layout and release building documentation somewhere, I'd be happy to read it there.

PR #2941, opened 2019-10-16
Fixed formatting of numbered lists in the markdown table.

reduckted:
Fixed the formatting of the numbered lists for RS1012 and RS1013. Because they were multi-line, they were causing the table to end, which messed up the formatting for everything after it. ~~I've changed them to be a single line and use HTML to create the numbered list.~~ I've changed the markdown generator to replace line breaks in the description with `<br>`.


mavasani:
@reduckted this is a tool generated markdown file. Manual edits will not be sufficient. We need to fix the underlying diagnostic’s resource string.

reduckted:
> 
> 
> @reduckted this is a tool generated markdown file. Manual edits will not be sufficient. We need to fix the underlying diagnostic’s resource string.

Oops! No worries, I'll fix the resource string them. Would using HTML in the resource string have any unwanted side effects?

reduckted:
> Would using HTML in the resource string have any unwanted side effects?

Looks like it will. This is from the ruleset editor with an analyzer I used for testing.

image

Any idea what generates that markdown file? I changed one of the descriptions in the resource file and rebuilt the solution, but the markdown file didn't change. I guess it has to be manually run from somewhere.

I think the simplest solution is to have the markdown generator replace `\r\n` with `<br>` It fixes the table formatting and, even though it's not formatted as a true numbered list, it's still readable.

reduckted:
> Any idea what generates that markdown file?

Ah, never mind, I found it. I was searching for the wrong thing. 😄 

I've pushed up the new changes.

mavasani:
> Any idea what generates that markdown file?

Yep, you have updated the correct tool. To regenerate all the markdown files for the repo, you can just re-run `eng\common\cibuild.cmd` from the root of the repo.

PR #2942, opened 2019-10-16
Tool to convert ruleset file to equivalent editorconfig file

mavasani:
Generated NuGet package contains a netcoreapp2.0 executable that takes the input ruleset and an output editorconfig as command line arguments. The package also contains a netstandard2.0 library with a public API to perform the conversion. We will likely use the latter in some of our migration tooling inside VS.

Primary complexity in the tool is from attempting to port existing comments in the ruleset file to the editorconfig file, so the port does not lead to loss of context of configuration entries.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,6 +9,11 @@
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
   <PropertyGroup>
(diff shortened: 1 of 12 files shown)


mavasani:
Recommended reviewing with https://github.com/dotnet/roslyn-analyzers/pull/2942/files?w=1

mavasani:
Ping for reviews.

mavasani:
Thank you @dotpaul

PR #2944, opened 2019-10-17
PropertySetAnalysis: Handling assignments to flow captures when tracking properties

dotpaul:
And also a doc sample and demonstrating some behavior that needs to be investigated.


dotpaul:
Thanks @mavasani !

PR #2959, opened 2019-10-21
Update CA1822 code fix to preserve modifiers

### LINKED ISSUE #2888: CA1822 Code fix strips the async keyword.
#### Analyzer package

FxCopAnalyzers

#### Package Version

2.9.4 & 2.9.5

#### Diagnostic ID

[CA1822](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1822-mark-members-as-static?view=vs-2019)

#### Repro steps

Enter the following code, wait for the rule to identify that the BarAsync method could be static and then apply the code fix.

```csharp
public async Task<int> BarAsync()
{
    await Task.Delay(20).ConfigureAwait(false);
    return 20;
}
```

#### Expected behavior

I would expect it to simply add the static keyword,

```csharp
public static async Task<int> BarAsync()
{
    await Task.Delay(20).ConfigureAwait(false);
    return 20;
}
```

#### Actual behavior

The current behaviour is that static keyword replaces the async keyword, resulting in compiler errors.

```csharp
public static Task<int> BarAsync()
{
    await Task.Delay(20).ConfigureAwait(false);
    return 20;
}
```

Evangelink:
This PR was made to fix issue #2888, `async` modifier is lost during refactoring, but turned out as a more global problem where previous modifiers are always lost.

Fix #2888

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.Fixer.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.Fixer.cs
@@ -59,7 +59,8 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
 
             // Update definition to add static modifier.
             var syntaxGenerator = SyntaxGenerator.GetGenerator(document);
-            var madeStatic = syntaxGenerator.WithModifiers(node, DeclarationModifiers.Static).WithAdditionalAnnotations(s_annotationForFixedDeclaration);
+            var oldModifiersAndStatic = syntaxGenerator.GetModifiers(node).WithIsStatic(true);
+            var madeStatic = syntaxGenerator.WithModifiers(node, oldModifiersAndStatic).WithAdditionalAnnotations(s_annotationForFixedDeclaration);
             document = document.WithSyntaxRoot(root.ReplaceNode(node, madeStatic));
             var solution = document.Project.Solution;
 
--- a/src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/MarkMembersAsStaticTests.Fixer.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/MarkMembersAsStaticTests.Fixer.cs
@@ -1371,28 +1371,5 @@ Public Class C
     End Function
 End Class");
         }
-
-        [Fact, WorkItem(2888, "https://github.com/dotnet/roslyn-analyzers/issues/2888")]
-        public async Task CA1822_CSharp_UnsafeModifier()
-        {
-            await VerifyCS.VerifyCodeFixAsync(@"
-public class C
-{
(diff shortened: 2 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.Fixer.cs
```diff
@@ -59,7 +59,8 @@ private async Task<Solution> MakeStaticAsync(Document document, SyntaxNode root,
 
             // Update definition to add static modifier.
             var syntaxGenerator = SyntaxGenerator.GetGenerator(document);
-            var madeStatic = syntaxGenerator.WithModifiers(node, DeclarationModifiers.Static).WithAdditionalAnnotations(s_annotationForFixedDeclaration);
+            var oldModifiersAndStatic = syntaxGenerator.GetModifiers(node).WithIsStatic(true);
```
Evangelink: For the 2nd test I have added (with `unsafe` modifier), the variable `oldMofifiersAndStatic` correctly contains both `unsafe` and `static` modifiers but when applied through `syntaxGenerator.WithModifiers(node, oldModifiersAndStatic)` the `unsafe` modifier is lost and I cannot understand why.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/MarkMembersAsStaticTests.Fixer.cs
```diff
@@ -1325,5 +1326,73 @@ Return fieldC
...
+    public unsafe void [|M1|]()
+    {
+        int var = 20;
+        int* p = &var;
+    }
+}",
+@"
+public class C
+{
+    public static unsafe void M1()
```
Evangelink: This test is currently failing as the `unsafe` keyword is lost.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.Fixer.cs
mavasani: This might be a bug in Rosyln's syntax generator. Would you be able to create a unit test in http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Workspaces.UnitTests/CodeGeneration/SyntaxGeneratorTests.cs,17 to verify and then file an issue on Roslyn repo: https://github.com/dotnet/roslyn/issues/new

Evangelink: I have added the following test 

```csharp
VerifySyntax<MethodDeclarationSyntax>(
    Generator.MethodDeclaration("m", modifiers: DeclarationModifiers.Unsafe.WithIsStatic(true)),
    "static unsafe void m()\r\n{\r\n}");
```

which was successful so I guess there was only a problem on my usage.

PR #2962, opened 2019-10-22
Ban Compilation.GetTypeByMetadataName

### LINKED ISSUE #2953: Ban the use of Compilation.GetTypeByMetadataName
We need to enable BannedApiAnalyzers on this repository, and ban the use of Compilation.GetTypeByMetadataName. Existing uses have already been updated, but we don't want more to slip in with future pull requests.

drewnoakes:
Fixes #2953

This is using the analyzer via NuGet.

### DIFF
--- a/src/BannedSymbols.txt
+++ b/src/BannedSymbols.txt
@@ -1,0 +1,1 @@
+M:Microsoft.CodeAnalysis.Compilation.GetTypeByMetadataName(System.String); Use WellKnownTypeProvider instead
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -21,7 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="2.9.6" />
+    <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="$(MicrosoftCodeAnalysisBannedApiAnalyzersVersion)" />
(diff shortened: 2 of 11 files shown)


src/Directory.Build.targets
```diff
@@ -20,6 +20,11 @@
     <UpToDateCheckInput Include="$(MSBuildThisFileDirectory)..\eng\Analyzers_ShippingRules.ruleset" Condition="'$(CodeAnalysisRuleSet)' == '$(MSBuildThisFileDirectory)..\build\Analyzers_NonShippingRules.ruleset'" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="2.9.6" />
```
mavasani: Please move the package reference to eng\Versions.props

mavasani: Sorry, I meant version of the referenced package must be defined in Versions.props

sharwell: @mavasani I'll go ahead and fix this and the other issues that arose during build

mavasani: Thanks!

drewnoakes: Thanks @sharwell for picking this up while I slept :)

src/BannedSymbols.txt
```diff
@@ -0,0 +1 @@
+M:Microsoft.CodeAnalysis.Compilation.GetTypeByMetadataName(System.String); Use WellKnownTypeProvider instead
```
drewnoakes: After seeing @sharwell's fixes, I think the diagnostic message here should probably have been

> Use GetOrCreateTypeByMetadataName with WellKnownTypeProvider instead

mavasani: Agreed. I will fix it. Thanks @drewnoakes

PR #2964, opened 2019-10-22
Update CA2242: do not report in generated code

### LINKED ISSUE #2293: CA2242 (Test for NaN correctly) is reported in generated code
`TestForNaNCorrectlyAnalyzer` should explicitly configure its generated code behavior. If the current behavior is kept, no change is needed to the tests. If the behavior is changed so the warning is _not_ reported in generated code, the tests should be updated to use `CSharpCodeFixVerifier` instead of `CSharpSecurityCodeFixVerifier` (and similar for VB).

Evangelink:
Fix #2293

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/TestForNaNCorrectlyTests.Fixer.cs
+++ b/src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/TestForNaNCorrectlyTests.Fixer.cs
@@ -4,7 +4,6 @@
 
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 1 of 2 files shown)


src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/TestForNaNCorrectlyTests.Fixer.cs
```diff
@@ -5,10 +5,10 @@
 using System.Threading.Tasks;
 using Xunit;
 // Use the security verifiers since this diagnostic is reported in generated code
```
sharwell: 💡  This comment is no longer correct, so it would be good to remove it

Issue #1932, opened 2018-12-17
Code fix for CA1826 "Do not use Enumerable methods on indexable collections. Instead use the collection directly"


Neme12:
It would be nice to have a code fix at least for the case of `First()` and `Last()`.

TKharaishvili:
Is this issue labeled correctly?

I just installed `Microsoft.CodeQuality.Analyzers` in the .NET 4.7 console app and couldn't find `CA1826` anywhere in the analyzer list.

After searching around some more I found it in the `Microsoft.NetCore.Analyzers` package:

![image](https://user-images.githubusercontent.com/4363170/53642249-c4e37c00-3c4a-11e9-87d8-2c9911058bdc.png)

but writing this piece of code gives me no warnings:
```C#
var list = new List<string> { "a", "b", "c" };
var f = list.First();
var l = list.Last();
var count = **list.Count();
```
since it says `NetCore.Analyzers` I then tried it on a .net core console app but still got no warnings.

Needless to say I'm quite confused.

TKharaishvili:
Just looked at the source code and found out that the warning is not issued for `IList<T>`-s. Probably because those enumerable methods are optimized for them.

But the question of this issue not being labeled correctly still remains.

TKharaishvili:
Hi @mavasani I just created a PR for this - https://github.com/dotnet/roslyn-analyzers/pull/2173

It's not ideal but I'm hoping you can guide me to the right direction with some code reviews. Cheers.

Evangelink:
A code fix now exists for `First`, `Last` and `Count` so I think the intent of this issue is actually solved.

The remaining work is:

- Handle `FirstOrDefault` and `LastOrDefault` 

- Handle C# 8 index expression for the codefix of `Last` and `LastOrDefault`. This is optional as some other codefix could help the migration to this new syntax

- Handle the creation on an intermediate variable for `Last` and `LastOrDefault`. This case can be a bit tricky and lead to various modifications in the case of an arrow expression or an invocation.

paul1956:
This warning should not apply to VB or at least there should be a Fix to locally suppress it and I don't see what the fix would be for the OrDefault versions without the code getting much more complex.

Evangelink:
[...]

PR #2968, opened 2019-10-23
CA1826: Add codefix for .Last() and .Count()

Evangelink:
Add 2 more code fixes for CA1826.

Relates to #1932

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
@@ -13,7 +13,6 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.Editing;
 using Microsoft.CodeAnalysis.Operations;
-using Microsoft.CodeAnalysis.Text;
 
 namespace Microsoft.NetCore.Analyzers.Runtime
 {
@@ -23,6 +22,10 @@ namespace Microsoft.NetCore.Analyzers.Runtime
     [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic), Shared]
     public class DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyFixer : CodeFixProvider
     {
+        private const string FirstPropertyName = "First";
+        private const string LastPropertyName = "Last";
+        private const string CountPropertyName = "Count";
+
         public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyAnalyzer.RuleId);
 
         public sealed override FixAllProvider GetFixAllProvider()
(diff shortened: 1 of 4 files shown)


src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
```diff
@@ -77,14 +79,48 @@ private async Task<Document> UseCollectionDirectly(Document document, TextSpan s
...
+
+            if (methodName == "First")
+            {
+                var zeroLiteral = generator.LiteralExpression(0);
+                return generator.ElementAccessExpression(collectionSyntaxNoTrailingTrivia, zeroLiteral);
+            }
+
+            if (methodName == "Last")
+            {
+                // TODO: Handle C# 8 index expression (and vb.net equivalent if any)
```
Evangelink: I don't know how to detect that the target code supports the index expression syntax. I have seen that I could access to `semanticModel.Compilation.LanguageVersion` if I cast the compilation to the C# or VB one but that seems wrong.

src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static string ToAnalyzerConfigString(this ReportDiagnostic reportDiagnost
                 ReportDiagnostic.Warn => DiagnosticSeverity.Warning,
                 ReportDiagnostic.Info => DiagnosticSeverity.Info,
                 ReportDiagnostic.Hidden => DiagnosticSeverity.Hidden,
-                ReportDiagnostic.Suppress => null,
+                ReportDiagnostic.Suppress => default(DiagnosticSeverity?),
```
mavasani: Why was this change needed?

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
```diff
@@ -77,14 +79,48 @@ private async Task<Document> UseCollectionDirectly(Document document, TextSpan s
...
+                // TODO: Handle cases were `collectionSyntax` is an invocation. We would need to create some intermediate variable.
```
Evangelink: I haven't handled this case yet because it adds quite a lot of complexity. I will need to find the previous statement to insert a new variable, find an available variable name and in the case of arrow expression I could need to change the arrow expression into a block.

I am fine to work on this but I might need some help.

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyTests.Fixer.cs
```diff
@@ -13,8 +14,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
 {
     public class DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyFixerTests
     {
-        [Fact]
-        public async Task CA1826FixEnumerableFirstExtensionCallCSharp()
+        [Fact, WorkItem(1932, "https://github.com/dotnet/roslyn-analyzers/issues/1932")]
+        public async Task CA1826_CSharp_EnumerableFirstExtensionCall()
```
Evangelink: I have renamed all tests to follow the same pattern and I have added a work item attribute to link to the ticket.

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyTests.Fixer.cs
```diff
@@ -113,8 +114,160 @@ End Class
...
+
+    Function GetList() As IReadOnlyList(Of Integer)
+        Return New List(Of Integer) From {1, 2, 3}
+    End Function
+End Class
+");
+        }
+
+        [Fact, WorkItem(1932, "https://github.com/dotnet/roslyn-analyzers/issues/1932")]
+        public async Task CA1826_CSharp_EnumerableCountExtensionCall()
```
Evangelink: The VB.Net pendant of this test doesn't trigger any issue.

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyTests.Fixer.cs
```diff
@@ -291,6 +663,8 @@ class C
     void M()
     {
         {|CS0201:GetList()[0]|};
+        {|CS0201:GetList()[GetList().Count - 1]|};
+        {|CS0201:GetList().Count|};
```
Evangelink: I am surprised that this line does raise `CS0201`.

src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
Evangelink: I will revert this change. I don't know why but I have to do this change to have the solution compiling.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
```diff
@@ -77,14 +79,48 @@ private async Task<Document> UseCollectionDirectly(Document document, TextSpan s
...
+                var countMemberAccess = generator.MemberAccessExpression(collectionSyntaxNoTrailingTrivia, "Count");
```
mavasani: You should add a defensive check that the type of the receiver has a property named `Count`. If not, you should bail out.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
```diff
@@ -77,14 +79,48 @@ private async Task<Document> UseCollectionDirectly(Document document, TextSpan s
...
+                var oneLiteral = generator.LiteralExpression(1);
+
+                // The SubstractExpression method will wrap left and right in parenthesis but those will be automatically removed later on
+                var substraction = generator.SubtractExpression(countMemberAccess, oneLiteral);
+                return generator.ElementAccessExpression(collectionSyntaxNoTrailingTrivia, substraction);
+            }
+
+            if (methodName == "Count")
+            {
+                return generator.MemberAccessExpression(collectionSyntaxNoTrailingTrivia, "Count");
```
mavasani: You should add a defensive check that the type of the receiver has a property named `Count`. If not, you should bail out.

src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
mavasani: I think you probably need to upgrade to a later VS2019 build. Target-typed switch expressions was not in the original implementation of C# 8 switch expressions, it was done as a bug fix, and likely your version of VS still does not have that fix.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
mavasani: I think it is fine to not do it for now.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
mavasani: I think it is fine to start with your current implementation. VS2019 should offer a subsequent refactoring to convert it to an index expression.

Evangelink: Shall I remove the todo comment? Shall I create some separate ticket?

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
Evangelink: Shall I remove the todo comment? Shall I create some separate ticket?

Evangelink: Obviously `FirstOrDefault` and `LastOrDefault` will fall under the same need.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
```diff
@@ -77,14 +79,99 @@ private async Task<Document> UseCollectionDirectly(Document document, TextSpan s
...
+                return HasCountProperty(collectionSyntax, semanticModel)
+                    ? generator.MemberAccessExpression(collectionSyntaxNoTrailingTrivia, "Count")
+                    : null;
+            }
+
+            Debug.Fail($"Unexpected method name '{methodName}' for {DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyAnalyzer.RuleId} code fix.");
+            return null;
+        }
+
+        private static bool HasCountProperty(SyntaxNode node, SemanticModel model)
```
Evangelink: @mavasani I have crafted a method to find the `Count` property but you might be able to redirect me to some helper method that would be better than what I did.

mavasani: We have an analyzer that detects invocations of Linq method `Count()` and flags them if they can be replaced with property reference `.Length` or `.Count`: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs#L301. You can possibly move out the logic to a shared helper?

Evangelink: @mavasani it seems that the other analyzer you are referring to doesn't handle `IReadOnlyCollection.Count`. Shall I create a separate PR for this and create the shared helper there?

mavasani: A separate PR sounds reasonable. Once that PR is merged in, you can update this PR to use the helper.

Evangelink:
@mavasani ready for review when you have some time.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
mavasani: @Evangelink Can you remove this private method and use the new helper method for presence of any Count property that you added in a separate PR?

Evangelink: @mavasani done

src/Microsoft.NetCore.Analyzers/Core/Runtime/DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectly.Fixer.cs
```diff
@@ -77,14 +79,58 @@ private async Task<Document> UseCollectionDirectly(Document document, TextSpan s
...
+            }
+
+            if (methodName == "Last")
+            {
+                // TODO: Handle C# 8 index expression (and vb.net equivalent if any)
+
+                var typeSymbol = semanticModel.GetTypeInfo(collectionSyntax).Type;
+                if (!typeSymbol.HasAnyCollectionCountProperty(WellKnownTypeProvider.GetOrCreate(semanticModel.Compilation)))
+                {
+                    return null;
```
mavasani: Would you be able to refactor the code so the code fix is only registered when we are sure the replacement property exists?

Evangelink: Let me know if my attempt looks ok or if it needs some change.

PR #2969, opened 2019-10-23
CA1068: Do not report when last argument is of type IProgress<T>

### LINKED ISSUE #2281: Let CA1068 not report on a method taking IProgress as the last parameter
#### Analyzer package

#### Package Version

v2.9.1

#### Diagnostic ID

CA1068

#### Repro steps
```cs
using System;
using System.Threading;
using System.Threading.Tasks;

public class C
{
    public Task SomeAsync(CancellationToken cancellationToken, IProgress<int> progress)
    {
    }
}
```

#### Expected behavior
CA1068 is not reported.
Because some Microsoft's documentations take IProgress as the last parameter:
1. https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/implementing-the-task-based-asynchronous-pattern#io-bound-tasks
1. https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap#choosing-the-overloads-to-provide

#### Actual behavior
CA1068 is reported:
`Method 'C.SomeAsync(System.Threading.CancellationToken, System.IProgress<int>)' should take CancellationToken as the last parameter`

Evangelink:
Fix #2281

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -37,76 +37,88 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
 
             context.RegisterCompilationStartAction(compilationContext =>
             {
-                INamedTypeSymbol cancellationTokenType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingCancellationToken);
-                if (cancellationTokenType != null)
+                var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation);
+                INamedTypeSymbol cancellationTokenType = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingCancellationToken);
+                INamedTypeSymbol iprogressType = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemIProgress1);
+                if (cancellationTokenType == null)
                 {
-                    compilationContext.RegisterSymbolAction(symbolContext =>
-                    {
-                        var methodSymbol = (IMethodSymbol)symbolContext.Symbol;
-                        if (methodSymbol.IsOverride ||
-                            methodSymbol.IsImplementationOfAnyInterfaceMember())
-                        {
-                            return;
-                        }
-
-                        int last = methodSymbol.Parameters.Length - 1;
-                        if (last >= 0 && methodSymbol.Parameters[last].IsParams)
-                        {
-                            last--;
-                        }
-
-                        // Skip optional parameters, UNLESS one of them is a CancellationToken
-                        // AND it's not the last one.
-                        if (last >= 0 && methodSymbol.Parameters[last].IsOptional
-                            && !methodSymbol.Parameters[last].Type.Equals(cancellationTokenType))
-                        {
-                            last--;
-
-                            while (last >= 0 && methodSymbol.Parameters[last].IsOptional)
-                            {
-                                if (methodSymbol.Parameters[last].Type.Equals(cancellationTokenType))
-                                {
-                                    symbolContext.ReportDiagnostic(Diagnostic.Create(
-                                        Rule, methodSymbol.Locations.First(), methodSymbol.ToDisplayString()));
-                                }
-
-                                last--;
-                            }
-                        }
-
-                        // Ignore multiple cancellation token parameters at the end of the parameter list.
-                        while (last >= 0 && methodSymbol.Parameters[last].Type.Equals(cancellationTokenType))
-                        {
-                            last--;
-                        }
-
-                        // Ignore parameters passed by reference when they appear at the end of the parameter list.
-                        while (last >= 0 && methodSymbol.Parameters[last].RefKind != RefKind.None)
-                        {
-                            last--;
-                        }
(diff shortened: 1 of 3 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -37,76 +37,88 @@ public override void Initialize(AnalysisContext context)
 
             context.RegisterCompilationStartAction(compilationContext =>
             {
-                INamedTypeSymbol cancellationTokenType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingCancellationToken);
-                if (cancellationTokenType != null)
+                var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation);
+                INamedTypeSymbol cancellationTokenType = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingCancellationToken);
+                INamedTypeSymbol iprogressType = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemIProgress1);
+                if (cancellationTokenType == null)
```
Evangelink: I am not exiting here when `iprogressType` to keep the previous behavior.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -37,76 +37,88 @@ public override void Initialize(AnalysisContext context)
 
             context.RegisterCompilationStartAction(compilationContext =>
             {
-                INamedTypeSymbol cancellationTokenType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingCancellationToken);
-                if (cancellationTokenType != null)
```
Evangelink: I reverted the if statement to reduce nesting.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -37,76 +37,88 @@ public override void Initialize(AnalysisContext context)
...
+                        last--;
+                    }
+
+                    // Ignore parameters passed by reference when they appear at the end of the parameter list.
+                    while (last >= 0 && methodSymbol.Parameters[last].RefKind != RefKind.None)
+                    {
+                        last--;
+                    }
+
+                    // Ignore IProgress<T> when last
```
Evangelink: Shall I handle multiple `IProgress` arguments?

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -238,6 +238,66 @@ public static void M1(this object p1, CancellationToken p2, object p3)
...
+Imports System.Threading
+Imports System.Threading.Tasks
+
+Public Class C
+    Public Function SomeAsync(ByVal cancellationToken As CancellationToken, ByVal o As Object, ByVal progress As IProgress(Of Integer)) As Task
+        Throw New NotImplementedException()
+    End Function
+End Class",
+            new DiagnosticResult(CancellationTokenParametersMustComeLastAnalyzer.Rule).WithLocation(7, 21)
+                .WithArguments("Public Function SomeAsync(cancellationToken As System.Threading.CancellationToken, o As Object, progress As System.IProgress(Of Integer)) As System.Threading.Tasks.Task"));
```
Evangelink: The message is quite weird compared to C#, was it expected? I have seen that there isn't any test for VB.NET while the analyzer actually targets VB.NET.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
sharwell: I've never seen guidelines allowing for this, so I would say only allow one for now. A test would be good so show the current behavior.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
sharwell: I'd say it's fine unless either:

1. We have specific guidelines for how to format signatures in messages like this, and this isn't following it
2. Someone is complaining about it and wants to propose a solution

sharwell: Either way, since it didn't _change_ as part of this pull request I would not expect it to need a fix here. 😄

sharwell:
@Evangelink In the future can you avoid force-pushes that change the base commit after the review starts unless absolutely necessary? It caused the diff to pick up a large number of unrelated changes: https://github.com/dotnet/roslyn-analyzers/compare/1ae96c3808c79ddbac1ee07b1b0030b4ebe4ea8b..f3ca827359bd681a2c152895ae59b83938d57012

sharwell:

Issue #2807, opened 2019-09-03
New analyzer: Do not call BeginInvoke on a delegate for .NET Core


sharwell:
This would apply to projects targeting .NET Standard or .NET Core.

https://github.com/dotnet/corefx/issues/5940
https://github.com/dotnet/winforms/issues/1548

danmoseley:
Or EndInvoke, right?

sharwell:
@danmoseley I wouldn't expect users to continue calling EndInvoke after the calls to BeginInvoke were removed. Isn't the return value from BeginInvoke one of the arguments to EndInvoke?

LunarWhisper:
> @danmoseley I wouldn't expect users to continue calling EndInvoke after the calls to BeginInvoke were removed. Isn't the return value from BeginInvoke one of the arguments to EndInvoke?

Task implements IAsyncResult so we need to check `EndInvoke` too.

PR #2972, opened 2019-10-24
New analyzer: Do not call BeginInvoke on a delegate for .NET Core #2807

duracellko:
Implements #2807

New analyzer: CA2069:DoNotCallBeginInvokeOnDelegate

Analyzer reports warning on a call of `BeginInvoke` method on Delegate. It reports only on compilation targeting .NET Standard or .NET Core, because BeginInvoke is not implemented in .NET Core.

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Tasks
+{
+    /// <summary>
+    /// CA1069: Do not call BeginInvoke on a delegate in .NET Core and .NET Standard.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotCallBeginInvokeOnDelegate : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1069";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(RuleId,
+                                                                             s_localizableTitle,
+                                                                             s_localizableMessage,
+                                                                             DiagnosticCategory.Design,
+                                                                             DiagnosticHelpers.DefaultDiagnosticSeverity,
+                                                                             isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
+                                                                             description: s_localizableDescription,
+                                                                             helpLinkUri: null,
+                                                                             customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    var invocation = (IInvocationOperation)operationContext.Operation;
+                    var containingType = invocation.TargetMethod?.ContainingType;
+
+                    if (containingType != null &&
+                        containingType.TypeKind == TypeKind.Delegate &&
+                        invocation.TargetMethod.Name == "BeginInvoke")
+                    {
+                        operationContext.ReportDiagnostic(Diagnostic.Create(Rule, invocation.Syntax.GetLocation()));
+                    }
+                }, OperationKind.Invocation);
+            });
+        }
+    }
+}
--- a/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
@@ -1,9 +1,8 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Immutable;
-using System.Linq;
 using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
@@ -38,28 +37,28 @@ namespace Microsoft.NetCore.Analyzers.Tasks
         public sealed override void Initialize(AnalysisContext context)
         {
             context.EnableConcurrentExecution();
-            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
 
             context.RegisterCompilationStartAction(compilationContext =>
             {
-                var isNetStandardAssembly = compilationContext.Compilation.ReferencedAssemblyNames.Any(identity => string.Equals(identity.Name, "netstandard", StringComparison.OrdinalIgnoreCase));
-
-                // Run diagnostic only on .NET Standard and .NET Core assemblies, because BeginInvoke is not implemented in .NET Core
-                if (isNetStandardAssembly)
+                if (compilationContext.Compilation.DoesTargetDotNetFramework())
                 {
-                    compilationContext.RegisterOperationAction(operationContext =>
-                    {
-                        var invocation = (IInvocationOperation)operationContext.Operation;
-                        var containingType = invocation.TargetMethod?.ContainingType;
-
-                        if (containingType != null &&
-                            containingType.TypeKind == TypeKind.Delegate &&
-                            invocation.TargetMethod.Name == "BeginInvoke")
-                        {
-                            operationContext.ReportDiagnostic(Diagnostic.Create(Rule, invocation.Syntax.GetLocation()));
-                        }
-                    }, OperationKind.Invocation);
+                    // Run diagnostic only on .NET Standard and .NET Core assemblies
+                    return;
                 }
+
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    var invocation = (IInvocationOperation)operationContext.Operation;
+                    var containingType = invocation.TargetMethod?.ContainingType;
+
+                    if (containingType != null &&
+                        containingType.TypeKind == TypeKind.Delegate &&
+                        invocation.TargetMethod.Name == "BeginInvoke")
+                    {
+                        operationContext.ReportDiagnostic(Diagnostic.Create(Rule, invocation.Syntax.GetLocation()));
+                    }
+                }, OperationKind.Invocation);
             });
         }
     }
--- a/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Immutable;
+using System.Linq;
 using Analyzer.Utilities;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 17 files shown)


mavasani:
@sharwell for review

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
```diff
@@ -0,0 +1,66 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                var isNetStandardAssembly = compilationContext.Compilation.ReferencedAssemblyNames.Any(identity => string.Equals(identity.Name, "netstandard", StringComparison.OrdinalIgnoreCase));
```
sharwell: ❕ There are two problems with this:

1. It covers .NET Standard 2.x, but misses .NET Standard 1.x
2. It fails to account for .NET Framework 4.7.2 projects that pick up the netstandard facade by referencing a .NET Standard assembly

I believe a better approach here is to exclude the known .NET Framework cases, which are builds that reference one of the following:

* mscorlib, Version=4.0.0.0, PublicKeyToken=b77a5c561934e089
* mscorlib, Version=2.0.0.0, PublicKeyToken=b77a5c561934e089

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                // Run diagnostic only on .NET Standard and .NET Core assemblies, because BeginInvoke is not implemented in .NET Core
+                if (isNetStandardAssembly)
```
sharwell: 💡 It would be easier to read the method if the code flow here is inverted

```csharp
if (!isNetStandardAssembly)
{
  // Run diagnostic only on .NET Standard and .NET Core assemblies
  return;
}
```

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
```diff
@@ -0,0 +1,66 @@
...
+                                                                             description: s_localizableDescription,
+                                                                             helpLinkUri: null,
+                                                                             customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
```
sharwell: This analyzer also needs to report diagnostics in generated code:

```suggestion
            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
```diff
@@ -0,0 +1,656 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing;
+using Test.Utilities;
+using Xunit;
+
+namespace Microsoft.NetCore.Analyzers.Tasks.UnitTests
+{
+    public class DoNotCallBeginInvokeOnDelegateTests : DiagnosticAnalyzerTestBase
```
sharwell: 📝 `DiagnosticAnalyzerTestBase` is the legacy test framework. New tests should be written using the new `VerifyCS` and `VerifyVB` patterns. If you don't implement this change as part of this pull request, it's fine to open a new issue to convert this.

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
mavasani: Please add a shared helper method for whatever functionality is implemented. I believe we added similar logic in another analyzer, that would also need to be fixed.

Also adding @stephentoub for suggestions on how to detect this in the best possible way.

stephentoub:
How are we thinking about analyzers like this vs https://github.com/microsoft/dotnet-apiport/tree/dev/docs/VSExtension and https://docs.microsoft.com/en-us/dotnet/standard/analyzers/api-analyzer ?

There are a bunch of cases of APIs that aren't 100% portable across either version of .NET or OS platform, and I'm not sure we want to be adding a new rule for each.  Shouldn't we instead be investing in the aforementioned existing analyzers?

cc: @terrajobst

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
duracellko: I extracted detection of compilation target to `CompilationExtensions.DoesTargetDotNetFramework`. However, implementation is 100% valid. The problem is that tests compile targeting .NET Framework, so there would be no way to verify this analyzer. Unless, tests are modified to compile to .NET Standard or .NET Core.

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
duracellko: Resolved

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
duracellko: Update, but I had to setup tests to skip verification of auto-generated code. I don't know, if it is possible to setup tests to expect diagnostics even with auto-generated code.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
duracellko: Resolved

duracellko:
I forgot to add my open questions:
- Is Rule ID correct?
- Is Role category and analyzer class namespace correct?
- Is there need to add the Rule into any other file in repository for documentation?

mavasani:
@duracellko Thanks for your PR. However, as per @stephentoub's comment, I think we need to first revisit if this is the correct repo for this analyzer and whether this is the space we want to invest in right now as there are large number of such APIs.

src/Test.Utilities/VisualBasicCodeFixVerifier`2.cs
```diff
@@ -23,14 +24,18 @@ public static DiagnosticResult Diagnostic(string diagnosticId)
         public static DiagnosticResult Diagnostic(DiagnosticDescriptor descriptor)
             => VisualBasicCodeFixVerifier<TAnalyzer, TCodeFix, XUnitVerifier>.Diagnostic(descriptor);
 
-        public static async Task VerifyAnalyzerAsync(string source, params DiagnosticResult[] expected)
+        public static Task VerifyAnalyzerAsync(string source, params DiagnosticResult[] expected)
+            => VerifyAnalyzerAsync(source, default(Action<Test>), expected);
+
+        public static async Task VerifyAnalyzerAsync(string source, Action<Test> setup, params DiagnosticResult[] expected)
```
sharwell: ❕ Do not add this overload. It is already directly supported by the test framework:

```csharp
await new VerifyVB.Test
{
  TestCode = code,
  TestBehaviors = TestBehaviors.SkipGeneratedCodeCheck,
}.RunAsync();
```

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
```diff
@@ -0,0 +1,663 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+using Xunit;
+
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
```
sharwell: 📝 Use `CSharpSecurityCodeFixVerifier` for analyzers that report diagnostics in generated code. It applies `TestBehaviors.SkipGeneratedCodeCheck` as part of the default test setup.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
```diff
@@ -26,5 +31,41 @@ internal static class CompilationExtensions
...
+        /// <param name="compilation">The compilation</param>
+        /// <returns><c>True</c> if the compilation targets .NET Framework; otherwise <c>false</c>.</returns>
+        internal static bool DoesTargetDotNetFramework(this Compilation compilation)
+        {
+            // This condition should be removed in future.
+            // It only ensures that tests are successful, because tests compilate to .NET Framework.
+            if (compilation.ReferencedAssemblyNames.Any(identity => string.Equals(identity.Name, "netstandard", StringComparison.OrdinalIgnoreCase)))
+            {
+                return false;
+            }
```
sharwell: 💭 I think we should remove this from the start. With a small amount of work to complete dotnet/roslyn-sdk#329 we should be able to get this working properly.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
duracellko: Thank you. I didn't notice that class.

sharwell:
Marking as **Blocked** pending decision above

mavasani:
Is this PR still active? If not, can we close out?

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(RuleId,
+                                                                             s_localizableTitle,
+                                                                             s_localizableMessage,
+                                                                             DiagnosticCategory.Design,
```
sharwell, line 28: ❓ Should this be in `Interoperability` instead?

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Microsoft.NetCore.Analyzers.Tasks
+{
+    /// <summary>
+    /// CA1069: Do not call BeginInvoke on a delegate in .NET Core and .NET Standard.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotCallBeginInvokeOnDelegate : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1069";
```
sharwell, line 18: ❓ Why would this not be an integral part of `PlatformCompatibilityAnalyzer`?

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
```diff
@@ -0,0 +1,656 @@
...
+                .WithMessage(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateMessage)
+                .WithLocation(line, column);
+        }
+
+        private static DiagnosticResult GetVBResultAt(int line, int column)
+        {
+            return VerifyVB.Diagnostic(DoNotCallBeginInvokeOnDelegate.RuleId)
+                .WithMessage(MicrosoftNetCoreAnalyzersResources.DoNotCallBeginInvokeOnDelegateMessage)
+                .WithLocation(line, column);
+        }
```
sharwell, line 654: 💡 These helpers can now be replaced with markup syntax (`[|...|]`)

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
```diff
@@ -0,0 +1,656 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Testing;
+using Xunit;
+
+using VerifyCS = Test.Utilities.CSharpSecurityCodeFixVerifier<
```
sharwell, line 6: ❕ These are not security analyzers

```suggestion
using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
```

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
```diff
@@ -0,0 +1,656 @@
...
+    Microsoft.NetCore.Analyzers.Tasks.DoNotCallBeginInvokeOnDelegate,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+using VerifyVB = Test.Utilities.VisualBasicSecurityCodeFixVerifier<
```
sharwell, line 10:
```suggestion
using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
```

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
```diff
@@ -0,0 +1,656 @@
...
+    Private Sub D()
+        Console.WriteLine(""Test"")
+    End Sub
+
+    Private Sub Callback(ByVal ar As IAsyncResult)
+        Console.WriteLine(ar.ToString())
+    End Sub
+End Class
+";
+            await VerifyVB.VerifyAnalyzerAsync(code, GetVBResultAt(7, 9));
```
sharwell, line 41: After the conversion to markup, it would be good to expand these tests:

```cs
await new VerifyVB.Test
{
  ReferenceAssemblies = ReferenceAssemblies.NetStandard.NetStandard10,
  TestCode = code,
}.RunAsync();

await new VerifyVB.Test
{
  ReferenceAssemblies = ReferenceAssemblies.NetStandard.NetStandard20,
  TestCode = code,
}.RunAsync();

await new VerifyVB.Test
{
  ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp31,
  TestCode = code,
}.RunAsync();

await new VerifyVB.Test
{
  TestCode = code,
}.RunAsync();

await new VerifyVB.Test
{
  ReferenceAssemblies = ReferenceAssemblies.NetFramework.NetFramework472.Default,
  TestState =
  {
    Sources = { code },

    // Ignore markup; no diagnostics will be reported for .NET Framework
    MarkupMode = MarkupMode.Ignore,
  },
}.RunAsync();
```

sharwell (changes requested):

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
sharwell: ➡️ Detection has been updated. Reference assemblies used in testing is a separate consideration moved below.

src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCallBeginInvokeOnDelegate.cs
sharwell: ➡️ Test library now automatically handles this case.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/DoNotCallBeginInvokeOnDelegateTests.cs
sharwell: 📝 Previously, `CSharpSecurityCodeFixVerifier` was used due to a limitation in the test library. This limitation no longer applies and `CSharpCodeFixVerifier` will now work without the need for other workarounds.

stephentoub:
> Is this PR still active? If not, can we close out?

Seems like there hasn't been any progress or response in a year and a half. I'll close it for now.

Thanks for the efforts.

PR #2973, opened 2019-10-24
CA1024: Do not report on implicit interface implementation

### LINKED ISSUE #1551: CA1024 False-positive when class implements IXmlSerializable
#### Analyzer package

Microsoft.CodeQuality.Analyzers, version [2.6.0](https://www.nuget.org/packages/Microsoft.CodeQuality.Analyzers/2.6.0)

#### Analyzer

CA1024: Use properties where appropriate

#### Repro steps

Create a class that implements IXmlSerializable:

    public class Example : IXmlSerializable
    {
        public XmlSchema GetSchema()
        {
            return null;
        }

        public void ReadXml(XmlReader reader) {...}

        public void WriteXml(XmlWriter writer) {...}
    }

#### Expected behavior

No warning is shown for GetSchema(). 

#### Actual behavior

A warning is shown for GetSchema().

When the method declaration comes from an interface, maybe it would make sense to show this warning only in the interface and never in the implementing classes. This would solve this issue and potentially similar issues. If user has a reason to use a method instead of a property, he would just need to suppress the warning once in the interface and not in all implementations.

Evangelink:
- Add a test to show that no issue is reported on explicit interface.

- Update the rule (and add a test) to stop reporting on implicit interface implementation.

- Do nothing regarding `override` as it is already handled.

Fix #1551

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
@@ -4,6 +4,12 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Testing;
 using Test.Utilities;
 using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.UsePropertiesWhereAppropriateAnalyzer,
+    Microsoft.CodeQuality.CSharp.Analyzers.ApiDesignGuidelines.CSharpUsePropertiesWhereAppropriateFixer>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.UsePropertiesWhereAppropriateAnalyzer,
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
```diff
@@ -4,12 +4,6 @@
 using Microsoft.CodeAnalysis.Testing;
 using Test.Utilities;
 using Xunit;
-using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
```
Evangelink: Unused

sharwell: These should not be removed. They are only unused because the tests are written using the legacy test framework and still needed to be updated to use these. I added them to all the tests so it's easier for users to aid in this conversion (as separate pull requests).

Evangelink:
My bad! I will revert the remove

Evangelink:
@sharwell anything left to do on my side here?

sharwell:

RobSwDev:
When will this fix be released? I have Microsoft.CodeQuality.Analyzers 2.9.8, and it still shows the warning. 
Looking at the il code, I see

`if (val != null && ... && !(val.get_Name() == "GetEnumerator"))`

i.e. the last check is against GetEnumerator, and there's no check for `methodSymbol.IsImplementationOfAnyImplicitInterfaceMember()`

Evangelink:
@RobSwDev AFAIK there will be a new release coming up soon. Tagging @mavasani

PR #2975, opened 2019-10-24
CA1829: Extend the rule to cover IReadOnlyCollection<T>.Count

### LINKED ISSUE #2974: CA1829: Extend the rule to cover IReadOnlyCollection<T>.Count
#### Analyzer package

#### Package Version

v2.9.5 (Latest)

#### Diagnostic ID

CA1829

#### Repro steps

```csharp
using System.Linq;

public class Foo
{
    public IReadOnlyCollection<int> GetData() => null;
    public int M() => GetData().Count();
}
```

#### Expected behavior

I expect to have a diagnostic raised against `GetData().Count()`.

#### Actual behavior

No diagnostic.

Evangelink:
Extend the rule to cover the `IReadOnlyCollection.Count` property. 
Extract an helper to find whether the type has any collection `Count` property so that we can reuse it in #2968 

Fix #2974

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
@@ -88,9 +88,8 @@ namespace Microsoft.NetCore.Analyzers.Performance
         /// </summary>
         protected sealed class OperationActionsContext
         {
+            private readonly Lazy<WellKnownTypeProvider> _wellKnownTypeProvider;
             private readonly Lazy<INamedTypeSymbol> _immutableArrayType;
-            private readonly Lazy<IPropertySymbol> _iCollectionCountProperty;
-            private readonly Lazy<INamedTypeSymbol> _iCollectionOfType;
 
             /// <summary>
             /// Initializes a new instance of the <see cref="OperationActionsContext"/> class.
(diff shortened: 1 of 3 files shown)


src/Microsoft.NetCore.Analyzers/Core/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.cs
```diff
@@ -88,9 +88,8 @@ private void OnCompilationStart(CompilationStartAnalysisContext context)
         /// </summary>
         protected sealed class OperationActionsContext
```
Evangelink: I think that the overall class might need a bit of cleaning (for simplification).

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -287,6 +287,57 @@ public static Accessibility DetermineMinimalAccessibility(this ITypeSymbol typeS
             return typeSymbol.Accept(MinimalAccessibilityVisitor.Instance);
         }
 
+        public static bool HasAnyCollectionCountProperty(this ITypeSymbol invocationTarget, WellKnownTypeProvider wellKnownTypeProvider)
+        {
+            const string countPropertyName = "Count";
```
Evangelink: I kept it as a local constant for now but it might be better to create some kind of `WellKnownMethodNames` like we have `WellKnownTypeNames` to avoid string duplication.

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -287,6 +287,57 @@ public static Accessibility DetermineMinimalAccessibility(this ITypeSymbol typeS
             return typeSymbol.Accept(MinimalAccessibilityVisitor.Instance);
         }
 
+        public static bool HasAnyCollectionCountProperty(this ITypeSymbol invocationTarget, WellKnownTypeProvider wellKnownTypeProvider)
```
Evangelink: Please feel free to suggest a better name, I am not entirely satisfied with this one but couldn't find any better (with a reasonable length xD).

Evangelink: I am not sure whether this should take:

1/ A `Compilation` object.

2/ The `WellKnownTypeProvider`.

3/ The 3 `INamedTypeSymbol`.

Evangelink:
@mavasani You are too fast for me :D

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
mavasani: I think this is fine.

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
mavasani: You can consider adding a doc comment on this method in your follow-up PR that uses this method.

mavasani: What you have currently is fine.

Issue #2324, opened 2019-04-12
CA1815, CA1066, and CA2231 on public ref struct


Zenexer:
#### Analyzer package

Microsoft.CodeQuality.Analyzers from [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

2.9.1

#### Diagnostic ID

* [CA1815](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1815-override-equals-and-operator-equals-on-value-types)
* [CA1066](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeQuality.Analyzers/Documentation/CA1066_ImplementIEquatableOfTWhenOverridingObjectEquals.md)
* [CA2231](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2231-overload-operator-equals-on-overriding-valuetype-equals)

#### Repro steps

CA1815 and CA2231:

``` cs
namespace Example
{
    public ref struct Test
    {
        byte _data;
    }
}
```

CA1066:

``` cs
namespace Example
{
    public ref struct Test
    {
        byte _data;

        public override bool Equals(object obj) => false;
    }
}
```

#### The problems

1. CA1815 doesn't make sense because Equals must always return false, since ref structs can't be cast to `object`.
2. CA1066 doesn't make sense because ref structs can't implement interfaces.
3. CA2231 doesn't make sense without also exposing an Equals method.  It's possible to implement operator == and operator !=, but it's probably not a great idea to promote that as the only equality checking method (which is what would happen if CA1815 and CA1066 were removed).

#### Proposed solution

1. Exclude ref structs from CA1815.
2. Exclude ref structs from CA1066 and anything else that recommends implementing an interface.
3. Exclude ref structs from CA2231, as the wording and proposed fix don't make sense for ref structs.
4. Add an analyzer that recommends adding either `public bool Equals(RefType other)` or `public static bool Equals(RefType left, RefType right)`
5. Add an analyzer that recommends overriding == and != if either `public bool Equals(RefType other)` or `public static bool Equals(RefType left, RefType right)` are present.

Evangelink:
@mavasani For points 4 and 5, shall we create 2 separate rules or add those checks to some existing rules?

PR #2978, opened 2019-10-25
CA1815: Do not report on ref struct

Evangelink:
Contributes to #2324

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/OverrideEqualsAndOperatorEqualsOnValueTypes.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/OverrideEqualsAndOperatorEqualsOnValueTypes.cs
@@ -66,10 +66,12 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     //  2. Do not fire for enumerators.
                     //  3. Do not fire for value types without members.
                     //  4. Externally visible types by default.
+                    //  5. Do not fire for ref struct.
                     // Note all the descriptors/rules for this analyzer have the same ID and category and hence
                     // will always have identical configured visibility.
                     if (!namedType.IsValueType ||
                         namedType.TypeKind == TypeKind.Enum ||
+                        (namedType.TypeKind == TypeKind.Struct && namedType.IsRefLikeType) ||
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/OverrideEqualsAndOperatorEqualsOnValueTypes.cs
```diff
@@ -66,10 +66,12 @@ public override void Initialize(AnalysisContext analysisContext)
                     //  2. Do not fire for enumerators.
                     //  3. Do not fire for value types without members.
                     //  4. Externally visible types by default.
+                    //  5. Do not fire for ref struct.
                     // Note all the descriptors/rules for this analyzer have the same ID and category and hence
                     // will always have identical configured visibility.
                     if (!namedType.IsValueType ||
                         namedType.TypeKind == TypeKind.Enum ||
+                        (namedType.TypeKind == TypeKind.Struct && namedType.IsRefLikeType) ||
```
Evangelink: I don't think there is any other type can be `RefLikeType` but I went for the belt and braces option here.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/OverrideEqualsAndOperatorEqualsOnValueTypesTests.cs
```diff
@@ -387,6 +387,16 @@ End Structure
...
         }
 
+        [Fact, WorkItem(2324, "https://github.com/dotnet/roslyn-analyzers/issues/2324")]
+        public void CSharp_RefStruct_NoDiagnostic()
+        {
+            VerifyCSharp(@"
+public ref struct A
+{
+    public int X;
+}", parseOptions: new CSharpParseOptions(LanguageVersion.CSharp8));
```
Evangelink: I don't know what's best between `CSharp8`, `Latest` or some other option that would automatically take care of testing with each and every new version starting at 8.

sharwell: `CSharp8` is good. `Latest` should never be used anywhere (it's not a version).

PR #2979, opened 2019-10-25
CA1066: Do not report on ref struct

Evangelink:
Contributes to #2324


src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EquatableAnalyzerTests.cs
```diff
@@ -370,6 +371,21 @@ public class D : C<D>
...
+            VerifyCSharp(@"
+public ref struct S
+{
+    public override bool Equals(object other)
+    {
+        return false;
+    }
+}
+",
+            parseOptions: new CSharpParseOptions(LanguageVersion.CSharp8));
```
Evangelink: I don't know what's best between `CSharp8`, `Latest` or some other option that would automatically take care of testing with each and every new version starting at 8.

sharwell: `CSharp8` is good. `Latest` should never be used anywhere (it's not a version).

PR #2980, opened 2019-10-25
CA2231: Do not report on ref struct

Evangelink:
Contributes to #2324


src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/OverloadOperatorEqualsOnOverridingValueTypeEquals.cs
```diff
@@ -39,6 +39,7 @@ public override void Initialize(AnalysisContext analysisContext)
             {
                 var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
                 if (namedTypeSymbol.IsValueType &&
+                    !(namedTypeSymbol.IsRefLikeType && namedTypeSymbol.TypeKind == TypeKind.Struct) &&
```
sharwell: Since we already have the `IsValueType` check, can we just do this?

```suggestion
                    !namedTypeSymbol.IsRefLikeType &&
```

Evangelink: Indeed I went for the stupid copy-paste... Sorry

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/OverloadOperatorEqualsOnOverridingValueTypeEqualsTests.cs
```diff
@@ -166,6 +166,20 @@ public override bool Equals(Object obj)
...
+        {
+            VerifyCSharp(@"
+public ref struct S
+{
+    public override bool Equals(object other)
+    {
+        return false;
+    }
+}
+");
```
Evangelink: I am even surprised this test passed without passing the `CSharpParseOptions`, I will retry locally.

sharwell: Does it fail locally? I know the new `VerifyCS` helper would not allow the test to pass in that state, but I'm not sure about the old test helpers that this test class is still using.

Evangelink: Yeah the test is green locally but I am still gonna add the parse option, it will be less of a problem for later refactoring.

Evangelink:
@sharwell anything left to do on my side here?

sharwell:

PR #3001, opened 2019-11-03
Implement UseValueTasksCorrectly rule

stephentoub:
ValueTasks are more specialized than Tasks, and are constrained in how they can be used. This rule tries to check for common misuse.

Putting this up as a draft PR to get initial feedback while I address a few TODOs, including adding VB tests and rounding out the test suite further.  This currently triggers 4 false positives in corefx (after catching a few straggling issues in corefx I've fixed), which is more than I'd like, but those cases are also advanced and niche, and while I'll continue to explore whether 1 or 2 of those can be addressed, 4 false positives out of 377 analyzed call sites to methods that return ValueTask in a code base that's advanced in this regard is probably acceptable. If it's not acceptable and it can't reasonably be fixed, we can disable the rule by default.

Contributes to https://github.com/dotnet/corefx/issues/40739
cc: @sharwell, @mavasani, @danmosemsft


sharwell:
@stephentoub as you iterated on this, were you able to add tests for each of the non-trivial cases you found in corefx? I'm referring both to cases that no longer report false positives, and to cases that do. The latter would have a note in the test that it's documenting a known false positive.

stephentoub:
> and to cases that do. The latter would have a note in the test that it's documenting a known false positive

Not for the "1 or 2" I mentioned I'd still like to try to address. I'll either address then and add valid test cases accordingly, or follow your suggestion and add test cases documenting the false positive.

The other 2 involve storing a ValueTask into a field, at which point my analysis gives up, and I don't plan to address those. There is a test for that.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+                    {
+                        switch (parentIo.TargetMethod.Name)
+                        {
+                            case "AsTask":
+                            case "Preserve":
+                                // Using AsTask to convert to a Task is acceptable consumption.
+                                // Use Preserve to enable subsequent unlimited consumption is acceptable.
+                                return;
+
+                            case "GetAwaiter" when parentIo.Parent is IInvocationOperation ga && ga.TargetMethod != null && ga.TargetMethod.Name == "GetResult":
```
huoyaoyuan: `when parentIo.Parent is IInvocationOperation { TargetMethod: { Name: "GetResult"} }`
I have never seen nested pattern matching actually used in production code, and here is an excellent case.

stephentoub:
@mavasani, @sharwell, any feedback on this, or should I just finish it up and promote it from a draft and assume it's on track?

333fred:
Does this need to be a CompilationStart analyzer? That's going to mean it doesn't run as part of the inner loop, only on full builds.

sharwell:
@stephentoub I'm trying to get code coverage working on this repository before moving forward with the other open PRs. I'm down to one final step there...

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+                    var invocation = (IInvocationOperation)operation;
+                    if (invocation.TargetMethod is null ||
+                        invocation.TargetMethod.MethodKind == MethodKind.Constructor) // ignore 'new ValueTask(...)'
+                    {
+                        return;
+                    }
+
+                    // Does the method return ValueTask?  If not, we're done.
+                    if (!valueTaskType.Equals(invocation.TargetMethod.ReturnType) &&
+                        !valueTaskOfTType.Equals(invocation.TargetMethod.ReturnType.OriginalDefinition))
```
333fred: Do we care about Nullable<ValueTask>?

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+                            operation = operation.Parent.Parent;
+                            continue;
+                        }
+                        else if (operation.Parent is IExpressionStatementOperation)
+                        {
+                            // Warn! This is a statement.  The result should have been used.
+                            operationContext.ReportDiagnostic(Diagnostic.Create(UnconsumedDescriptor, invocation.Syntax.GetLocation()));
+                            return;
+                        }
+                        else if (TryGetLocalSymbolAssigned(operation.Parent, out var valueTypeSymbol, out var cfg, out var startingBlock))
```
333fred: This case could end up being hit multiple times in the same method. We should try and only analyze the cfg once.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
333fred: Lots of opportunities for this.

mavasani:
> Does this need to be a CompilationStart analyzer? That's going to mean it doesn't run as part of the inner loop, only on full builds.

@333fred CompilationStart actions are executed during live analysis. Only CompilationEnd actions are not executed during live analysis, unless user has enabled Full Solution Analysis, at which point even compilation end actions are executed.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+                compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsDebug, out var debugType);
+
+                // Process all invocations.  This analyzer works by finding all method invocations that return ValueTasks,
+                // and then analyzing those invocations and their surroundings.
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    // Is this a method invocation?  If not, we're done.
+                    var operation = operationContext.Operation;
+                    var invocation = (IInvocationOperation)operation;
+                    if (invocation.TargetMethod is null ||
```
mavasani: This if statement can be removed. `TargetMethod` is guaranteed to be non-null and not a constructor. Latter generates an `IObjectCreationOperation`.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+                    if (!valueTaskType.Equals(invocation.TargetMethod.ReturnType) &&
+                        !valueTaskOfTType.Equals(invocation.TargetMethod.ReturnType.OriginalDefinition))
+                    {
+                        return;
+                    }
+
+                    // The only method that returns a ValueTask on which we allow unlimited later consumption
+                    // is ValueTask.Preserve.  Use is rare, but special-case it.  If this is Preserve, we're done.
+                    if (invocation.TargetMethod.Name == "Preserve" &&
+                        (valueTaskType.Equals(invocation.TargetMethod.ContainingType) ||
```
mavasani: Isn't the check after `&&` always true given the prior if statement?

mavasani: Ah sorry, I see the difference.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+                         valueTaskOfTType.Equals(invocation.TargetMethod.ContainingType.OriginalDefinition)))
+                    {
+                        return;
+                    }
+
+                    // If this is a method call off of the ValueTask, then check to see if it's special.
+                    if (invocation.Parent is IInvocationOperation parentIo && parentIo.TargetMethod != null)
```
mavasani:
```suggestion
                    if (invocation.Parent is IInvocationOperation parentIo)
```

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
mavasani:
```suggestion
                            case "GetAwaiter" when parentIo.Parent is IInvocationOperation ga && ga.TargetMethod.Name == "GetResult":
```

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,547 @@
...
+
+                            case "ConfigureAwait":
+                                // ConfigureAwait returns another awaitable.  Use that one instead for subsequent analysis.
+                                operation = invocation = parentIo;
+                                break;
+                        }
+                    }
+                    else if (invocation.Parent is IPropertyReferenceOperation prop && prop.Property.Name == "Result")
+                    {
+                        operationContext.ReportDiagnostic(Diagnostic.Create(AccessingIncompleteResultDescriptor, invocation.Syntax.GetLocation()));
```
mavasani:
```suggestion
                        operationContext.ReportDiagnostic(invocation.CreateDiagnostic(AccessingIncompleteResultDescriptor));
```

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
stephentoub: Thanks, @mavasani.  I thought that was probably the case, but I saw other analyzers doing this check, e.g.
https://github.com/dotnet/roslyn-analyzers/blob/2c46f6fe2e53b13b21066e1753cba95a8d6de3bb/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCreateTasksWithoutPassingATaskScheduler.cs#L57
and so wasn't sure.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
mavasani: Yes, that is correct. The best way to implement this would be to create a separate flow analysis, similar to ones in https://github.com/dotnet/roslyn-analyzers/tree/master/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis, and run it once for the entire operation block, which computes the state at individual operations (program points) and/or basic block start for all value task creations within the method body. I will write up a one pager today to help folks start writing such flow analyses.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
stephentoub: > Do we care about Nullable?

I don't think so.  If you're returning `ValueTask?`, you're off the rails :)  But I hesitate to warn about such a case, as I have no idea why you'd be doing it; might be something we add later.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
stephentoub: This is hitting a cache in IOperationExtensions.cs.  Doesn't that address the concern?

stephentoub:
I addressed the initial feedback (thanks) and my own TODOs.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,537 @@
...
+    /// <summary>CA2012: Use ValueTasks correctly.</summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class UseValueTasksCorrectlyAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2012";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseValueTasksCorrectlyTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseValueTasksCorrectlyDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static readonly DiagnosticDescriptor GeneralDescriptor = new DiagnosticDescriptor(
```
Evangelink: [cosmetic] Most of the other analyzers defined in the project use the `Rule` suffix and not `Descriptor` but feel free to ignore my comment as I am jumping on a PR I'm not implied onto.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,537 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(GeneralDescriptor, UnconsumedDescriptor, DoubleConsumptionDescriptor, AccessingIncompleteResultDescriptor);
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                // Get the target ValueTask / ValueTask<T> types.  If they don't exist, nothing more to do.
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksValueTask, out var valueTaskType) ||
```
Evangelink: [minor] The performance impact is low but you could create a variable for the known type provider as it is used 3 times. `var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation);` and then replace all calls to `compilationContext.Compilation.TryGetOrCreateTypeByMetadataName` with `wellKnownTypeProvider.TryGetOrCreateTypeByMetadataName`.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,537 @@
...
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksGenericValueTask, out var valueTaskOfTType))
+                {
+                    return;
+                }
+
+                // Get the type for System.Diagnostics.Debug.  If we can't find it, that's ok, we just won't use it.
+                compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsDebug, out var debugType);
```
Evangelink: [minor] You are not doing anything with the result of the `TryGet` call so you could use the `GetOrCreateTypeByMetadataName` method.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
stephentoub: Renamed.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
stephentoub: Done.

src/Microsoft.NetCore.Analyzers/Core/Tasks/UseValueTasksCorrectly.cs
stephentoub: Done.

stephentoub:
@mavasani, what's the PR process at this point?

mavasani:
> what's the PR process at this point?

We need at least one approval from a member on @dotnet/roslyn-analysis. I will try to get to reviewing this PR a bit later today.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -0,0 +1,1274 @@
...
+    {
+        protected override DiagnosticAnalyzer GetBasicDiagnosticAnalyzer() => new UseValueTasksCorrectlyAnalyzer();
+        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer() => new UseValueTasksCorrectlyAnalyzer();
+
+        private static readonly ParseOptions s_csharpPreviewParseOptions = new CSharpParseOptions(CodeAnalysis.CSharp.LanguageVersion.Preview);
+        private static readonly ParseOptions s_basicPreviewParseOptions = new VisualBasicParseOptions(CodeAnalysis.VisualBasic.LanguageVersion.Latest);
+
+        private static string CSharpBoilerplate(string s) => s +
+            @"
+            internal static class Helpers
```
mavasani: Nit: Can consider moving these to https://github.com/dotnet/roslyn-analyzers/tree/master/src/Test.Utilities/MinimalImplementations so can be reused by other tests. It seems these will potentially be used by multiple analyzers in future.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -0,0 +1,1274 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.VisualBasic;
+using Test.Utilities;
+using Xunit;
+
+namespace Microsoft.NetCore.Analyzers.Tasks.UnitTests
+{
+    public class UseValueTasksCorrectlyTests : DiagnosticAnalyzerTestBase
```
mavasani: @sharwell and @Evangelink are working towards removal of `DiagnosticAnalyzerTestBase` in favor of new Microsoft.CodeAnalysis.Testing library, see https://github.com/dotnet/roslyn-analyzers/pull/3052. We may want to ensure new tests are directly written with the new library, so there are no merge conflicts with that PR.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -0,0 +1,1274 @@
...
+        [Fact]
+        public void NoDiagnostics_WorkBetweenStoreAndAwait()
+        {
+            VerifyCSharp(CSharpBoilerplate(@"
+                using System;
+                using System.Threading.Tasks;
+
+                class C
+                {
+                    public async Task WorkBetweenStoreAndAwait()
```
mavasani: Can you add unit tests where task is created in a `try` block and awaited in `finally` and/or `catch` blocks?

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -0,0 +1,1274 @@
...
+                        ValueTask vt1 = Helpers.ReturnsValueTask();
+                        await vt1;
+
+                        ValueTask<string> vt2 = Helpers.ReturnsValueTaskOfT<string>();
+                        await vt2;
+
+                        ValueTask<int> vt3 = Helpers.ReturnsValueTaskOfInt();
+                        await vt3;
+
+                        await vt1;
```
mavasani: Can you add a test where awaits are across blocks with differing conditions, like say:

```csharp
if (condition1)
{
   await vt1;
}
else
{
   await vt2;
}

await vt1; // Should be flagged as there is at least one code path with duplicate await
await vt3;

if (condition2)
{
   await vt2; // Should not be flagged as we don't know the relation between condition1 and condition2, so don't know for sure if this a duplicate await

   await vt3;  // Should be flagged as there is at least one code path with duplicate await
}
```

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -0,0 +1,1274 @@
...
+                            T result = vt.Result;
+                        }
+                        Debug.Assert(vt." + isProp + @");
+                    }
+                }"),
+                    parseOptions: s_csharpPreviewParseOptions,
+                    expected: new DiagnosticResult[]
+                    {
+                        // Test0.cs(10,40): warning CA2012: Use ValueTasks correctly.
+                        GetCSharpResultAt(10, 40, UseValueTasksCorrectlyAnalyzer.AccessingIncompleteResultRule),
```
mavasani: With the new test framework, you can actually add annotations to original source code for the diagnostic location, which makes it easier to understand the squiggle span. I am presuming this diagnostic is flagging the value task creation and not the `.GetResult()` or `.Result` operations. Is that correct? If so, should the diagnostic instead be flagging the incorrect use site. i.e. should we have 6 diagnostics here, 2 in each method?

mavasani:
FYI: You will need to move the added analyzer files to a different folder and project, `Microsoft.CodeAnalysis.NetAnalyzers`, once https://github.com/dotnet/roslyn-analyzers/pull/3165 is merged. See https://github.com/dotnet/roslyn-analyzers/pull/3139#issue-358842977 for details.

stephentoub:
Thanks, @mavasani.  I'm getting back to this.  Regarding:

> I am fine if we feel that need not come in the first implementation - I would also be willing to volunteer to experiment converting your flow analysis to our repo's flow analysis framework as a follow-up PR.

It'd be great if we could go ahead with it as-is, and then I'd love to take you up on your offer to try converting it over.  Would be a good learning experience for me to see what it is you do.

I'll address the rest of the feedback in the meantime so we can get this in.

src/Microsoft.NetCore.Analyzers/UnitTests/Tasks/UseValueTaskCorrectlyTests.cs
stephentoub: > If so, should the diagnostic instead be flagging the incorrect use site. i.e. should we have 6 diagnostics here, 2 in each method?

It's flagging the first usage failure for each creation site; as soon as it hits a failure, it stops looking for others.

stephentoub:
@mavasani, @sharwell, I finally got back to this, rebased on master, cleaned up the differences from the testing overhaul, and addressed the remaining feedback.  Mind taking another look?

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.sarif
```diff
@@ -2340,6 +2340,25 @@
             ]
           }
         },
+        "CA2012": {
+          "id": "CA2012",
+          "shortDescription": "Use ValueTasks correctly",
+          "fullDescription": "ValueTasks must only be consumed once, generally by directly awaiting them.  Attempts to consume a ValueTask multiple times may result in an exception or corruption.",
```
halter73: Should the full description for CA2012 include anything about the AccessingIncompleteResultRule case? Maybe something like the following?

```suggestion
          "fullDescription": "ValueTasks must only be consumed once and should not have their result accessed until the instance has already completed. ValueTasks should generally be awaited. Attempts to consume a ValueTask multiple times may result in an exception or corruption.",
```

src/Microsoft.NetCore.Analyzers/Microsoft.NetCore.Analyzers.sarif
```diff
@@ -325,6 +325,25 @@
             ]
           }
         },
+        "CA2012": {
+          "id": "CA2012",
+          "shortDescription": "Use ValueTasks correctly",
+          "fullDescription": "ValueTasks must only be consumed once, generally by directly awaiting them.  Attempts to consume a ValueTask multiple times may result in an exception or corruption.",
+          "defaultLevel": "warning",
```
mavasani: Was this file generated prior to changing the `RuleLevel` values in the analyzer? `"defaultLevel": "warning"` indicates `RuleLevel.BuildWarning` was used when creating the analyzer descriptors.

I am wondering if this rule should be a `RuleLevel.BuildWarningCandidate`, that should be triaged to be enabled as a warning by default?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/UseValueTasksCorrectly.cs
```diff
@@ -0,0 +1,542 @@
...
+                }, OperationKind.Invocation);
+            });
+        }
+
+        /// <summary>Handles more complicated analysis to warn on more complicated erroneous usage patterns.</summary>
+        private static void HandleAtypicalValueTaskUsage(OperationAnalysisContext operationContext, INamedTypeSymbol? debugType, IOperation operation, IInvocationOperation invocation)
+        {
+            if (TryGetLocalSymbolAssigned(operation.Parent, out var valueTypeSymbol, out var startingBlock))
+            {
+                // At this point, it's very likely misuse and we could warn. However, there are a few advanced
```
mavasani: Can you file a tracking issue to attempt converting this to a dataflow analysis based on the DFA framework in the repo and assign it to me?

src/Microsoft.NetCore.Analyzers/Microsoft.NetCore.Analyzers.sarif
mavasani: Actually, I take my comment back. I did not see that this was `Microsoft.NetCore.Analyzers.sarif`, which is the analyzer assembly that ships with FxCopAnalyzers package, where we do want all analyzers to be warnings by default. I do see further down that `Microsoft.CodeAnalyzers.NetAnalyzers.sarif` does have this rule listed as a hidden rule by default.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -0,0 +1,1319 @@
...
+    Microsoft.NetCore.Analyzers.Tasks.UseValueTasksCorrectlyAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.NetCore.Analyzers.Tasks.UnitTests
+{
+    public class UseValueTasksCorrectlyTests
+    {
+        private static string CSBoilerplate(string s) => s +
+            @"
+            internal static class Helpers
```
mavasani: @sharwell Can we now avoid this boilerplate code in tests by directly referencing a NuGet package?

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.sarif
stephentoub: Thanks.  Addressed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/UseValueTasksCorrectly.cs
stephentoub: Done: https://github.com/dotnet/roslyn-analyzers/issues/3332. Thanks.

PR #3005, opened 2019-11-04
Prefer expected diagnostics over CompilerDiagnostics.None

### LINKED ISSUE #2995: Review uses of CompilerDiagnostics.None
Following the PR #2988 the various use of `CompilerDiagnostics.None` should be reviewed and replaced with a proper expected diagnostics.

This will avoid hiding some issues with a test change or rule change.

cc @sharwell

Evangelink:
Review use of `CompilerDiagnostics.None` to favor expected diagnostics.

Fix #2995


src/Microsoft.CodeAnalysis.Analyzers/UnitTests/FixAnalyzers/FixerWithFixAllAnalyzerTests.cs
```diff
@@ -90,11 +90,13 @@ public class MyDerivedCodeActionWithEquivalenceKey : MyAbstractCodeActionWithEqu
...
 
             // Verify expected diagnostics for fixer that supports FixAllProvider.
-            await VerifyCS.VerifyAnalyzerAsync(source + fixAllProviderString + sourceSuffix, compilerDiagnostics, expected);
+            await VerifyCS.VerifyAnalyzerAsync(source + fixAllProviderString + sourceSuffix, expected);
 
             // Verify RS1016 (OverrideGetFixAllProviderRule) diagnostic for fixer that does not support FixAllProvider.
-            expected = new DiagnosticResult[] { missingGetFixAllProviderOverrideDiagnostic };
-            await VerifyCS.VerifyAnalyzerAsync(source + sourceSuffix, compilerDiagnostics, expected);
+            var allExpected = new[] { missingGetFixAllProviderOverrideDiagnostic }
+                .Concat(expected.Where(x => x.Severity == DiagnosticSeverity.Error))
```
Evangelink: Not sure whether that's best to check the severity or something else. We could check ID starts with `CS` (or `BC` for vbnet).

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
```diff
@@ -331,7 +332,7 @@ public int GetSomethingWithUnboundInvocation()
 ");
         }
 
-        [Fact(Skip = "https://github.com/dotnet/roslyn/issues/7222")]
+        [Fact]
```
Evangelink: Shall I keep the `Skip`?

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
```diff
@@ -341,7 +342,11 @@ public async Task VisualBasic_CA1024NoDiagnosticOnUnboundMethodCaller()
         Return 0
     End Function
 End Class
-", CompilerDiagnostics.None);
+",
+                // Test0.vb(3,21): warning CA1024: Use properties where appropriate
+                GetCA1024BasicResultAt(3, 21, "GetSomethingWithUnboundInvocation"),
+                // Test0.vb(4) : error BC30451: 'Console' is not declared. It may be inaccessible due to its protection level.
```
Evangelink: Adding the `Imports System` will prevent the unbound method call.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -421,7 +421,9 @@ void CatchAndRethrowExplicitly()
         }
     }
 }
-", CompilerDiagnostics.None);
+",
+            new DiagnosticResult("CS0160", CodeAnalysis.DiagnosticSeverity.Error).WithLocation(18, 20),
+            new DiagnosticResult("CS0103", CodeAnalysis.DiagnosticSeverity.Error).WithLocation(26, 27));
```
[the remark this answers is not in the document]
Evangelink: That's a way better syntax

Evangelink:
@sharwell Let me know if you prefer a merge or a rebase to fix the conflict.

sharwell:
@Evangelink rebase would be fine. I only skimmed the review so far.

sharwell:
@Evangelink Let me know when the conflicts are resolved and I'll review again. I also merged #3011 which provides the first version of the reference assemblies (and PackageReference) support for analyzer testing.

Evangelink:
All good @sharwell

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
```diff
@@ -3768,7 +3768,9 @@ private void M2(C c)
         var x = c._field;
     }
 }
-", CompilerDiagnostics.None,
+",
+            // Test0.cs(8,9): error CS0165: Use of unassigned local variable 'c2'
+            DiagnosticResult.CompilerError("CS0165").WithLocation(8, 9),
```
sharwell: 💭 I'd almost prefer to see `WithMessage` used here _instead of the comment.

```suggestion
            DiagnosticResult.CompilerError("CS0165").WithLocation(8, 9).WithMessage("Use of unassigned local variable 'c2'"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -297,18 +299,24 @@ public async Task CSharpReassignLocalValueTypeVariableAndReferToItsProperty()
...
 
 public class Test
 {
     public void Method()
     {
         S a, b;
-        a.Property = c = a = b;
+        a.Property = c = a = b;     // error CS0165: Use of unassigned local variable 'a'
+                                    // error CS0103: The name 'c' does not exist in the current context
+                                    // error CS0165: Use of unassigned local variable 'b'
```
sharwell: 📝 Let's move all the new comments into the `WithMessage` calls instead. There are two groups to consider:

1. Comments added _inside the test code strings_: These comments (technically) change the test inputs. We should move these to `WithMessage` instead.
2. Comments added as normal comments preceding the `CompilerError` calls: These comments match the old pattern of adding comments to the diagnostic location lines so it's easy to read them. I find them easier to maintain long-term by placing them in `WithMessage` calls (where they are never out of date because the test will fail if the message changes), but this is an optional change for this pull request.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/UseNameOfInPlaceOfStringTests.cs
```diff
@@ -139,9 +141,9 @@ class C
 {
     void M(int x)
     {
-        throw new ArgumentNullException(""test"", ""test2"", ""test3"");
+        throw new ArgumentNullException(""test"", ""test2"");
```
sharwell: 📝 We should make sure this early return line of the analyzer is still covered by tests:

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUnusedPrivateFieldsTests.cs
```diff
@@ -416,13 +439,16 @@ <System.ComponentModel.Composition.ExportAttribute> _
...
                 {
+                    // Test0.vb(3) : error BC30002: Type 'System.Composition.ExportAttribute' is not defined.
+                    DiagnosticResult.CompilerError("BC30002").WithLocation(3, 6),
                     // Test0.vb(4,13): warning CA1823: Unused field 'fieldWithMefV1ExportAttribute'.
                     GetCA1823BasicResultAt(4, 13, "fieldWithMefV1ExportAttribute"),
+                    // Test0.vb(6) : error BC30002: Type 'System.ComponentModel.Composition.ExportAttribute' is not defined.
+                    DiagnosticResult.CompilerError("BC30002").WithLocation(6, 6),
                     // Test0.vb(7,13): warning CA1823: Unused field 'fieldWithMefV2ExportAttribute'.
-                    GetCA1823BasicResultAt(7, 13, "fieldWithMefV2ExportAttribute"),
+                    GetCA1823BasicResultAt(7, 13, "fieldWithMefV2ExportAttribute")
```
sharwell: 📝 Trailing commas on the last item of a multi-line collection initializer should not be removed.

```suggestion
                    GetCA1823BasicResultAt(7, 13, "fieldWithMefV2ExportAttribute"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
sharwell: Yes, though this test is supposed to be testing this case.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
sharwell: @mavasani ?

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
sharwell: See https://github.com/dotnet/roslyn-analyzers/pull/3005#discussion_r343703984 for the main discussion on this change.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
Evangelink: I was actually thinking about that at first but I was afraid it would make tests to fail on the CI because it tests with other languages hence will change the expected message. But maybe this was a wrong assumption on my side.

sharwell: If it does impact tests, we can use an xunit extension [like `UseCultureAttribute`](https://github.com/dotnet/roslyn/blob/master/src/Test/Utilities/Portable/Assert/UseCultureAttribute.cs) to set the language for a specific test.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/UseNameOfInPlaceOfStringTests.cs
Evangelink: No test is reaching this point. But keeping the invalid syntax and adding an expected compiler error still doesn't reach this point. I am happy to work on adding coverage for this condition but it might be better to have a separate PR, don't you think?

Evangelink:
@sharwell Could you do a new review when you have some time? Or at least let me know if there is something I still need to do to move forward.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
Evangelink: Ping @mavasani

Evangelink:
Hey folks, sorry to bother with this but would it be possible to have a review or to let me know if something is blocking the PR?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/FixAnalyzers/FixerWithFixAllAnalyzerTests.cs
Evangelink: @sharwell Is it ok for you?

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
Evangelink: @mavasani Could you please let me know if that's ok to remove the `skip`?

Evangelink:
@sharwell, @mavasani we need to have this PR merged before the other test framework conversion PR. Could you please have a look at this PR soon?

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/UseNameOfInPlaceOfStringTests.cs
Evangelink: Ticket #3115

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
mavasani: If the test is passing it is fine to remove the skip attribute - thanks!

mavasani:
It seems like some of these have already made it into the repo. I filed https://github.com/dotnet/roslyn-analyzers/issues/3116 to even revert these.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
Evangelink: @mavasani Isn't it better to keep the change for the time being to enforce a stricter check on the message, as explained by @sharwell, and in the event of a new localized build to update the strategy either as suggested or with something different?

I really can see pros and cons in the 2 options but having the message tested kind of provide a second level of safety for `Roslyn` itself so I tend to think that's better.

mavasani: I don't think I agree with that. Validating the messages for compiler diagnostics is not the intent of any tests in this repo. We are regressing ability to do localization runs for an unwanted validation. Note that at one point we were close to enabling spanish runs in this repo. Tagging @JoeRobich - it would be good if we can enable the spanish run for this repo sooner rather than later. Joey I can do this work if you point me to the steps.

Evangelink:
Fix #3116

mavasani:
@Evangelink We would also want to cherry-pick https://github.com/dotnet/roslyn-analyzers/pull/3005/commits/ee3bb59d716f011c37b1ac728b4d9d974514e26e to 2.9.x branch. Would you be able to create a separate PR for that? Otherwise, I can do it once this PR is merged in. Thanks!

mavasani:
https://github.com/dotnet/roslyn-analyzers/pull/3118 enables the Spanish leg for this repo, and as expected it fails due to the compiler error message validation. I am going to do the following:

1. Accept and merge this PR.
2. Cherry-pick https://github.com/dotnet/roslyn-analyzers/commit/ee3bb59d716f011c37b1ac728b4d9d974514e26e to https://github.com/dotnet/roslyn-analyzers/pull/3118
3. File a separate issue to discuss if validating compiler messages in our tests is useful, and the tests should be special cased to just run on en-US locale. I am opposed to this as it unnecessary disables our analyzer validation tests on our localized runs for a redundant validation for compiler messages. @sharwell supports this change. We are going to discuss this at next design meeting.

mavasani:
Had an offline chat with @sharwell. We have decided to force merge https://github.com/dotnet/roslyn-analyzers/pull/3118 to enable spanish leg on 2.9.x branch, and eventually on master through normal code flow. Sam is going to create a separate PR to fix the spanish failures by switch compiler diagnostic validation to use markup.

@Evangelink You can revert https://github.com/dotnet/roslyn-analyzers/commit/ee3bb59d716f011c37b1ac728b4d9d974514e26e on this PR for now.

Evangelink:
@mavasani This PR starts to be a bit of a mess, do you mind if I open a new one superseding this one? Also, is the consensus to use markup syntax for compiler errors? If so, I can recreate it with that instead of simply cherry-picking commits.

mavasani:
@Evangelink new PR with markup seems good.

PR #3011, opened 2019-11-05
Update to Microsoft.CodeAnalysis.Testing 1.0.1-beta1.19554.4

sharwell:


src/Microsoft.NetCore.Analyzers/Core/Runtime/SerializationRulesDiagnosticAnalyzer.cs
```diff
@@ -128,7 +128,8 @@ public override void Initialize(AnalysisContext analysisContext)
                         return;
                     }
 
-                    var isNetStandardAssembly = context.Compilation.ReferencedAssemblyNames.Any(identity => string.Equals(identity.Name, "netstandard", StringComparison.OrdinalIgnoreCase));
+                    var systemObjectSymbol = context.Compilation.GetSpecialType(SpecialType.System_Object);
+                    var isNetStandardAssembly = systemObjectSymbol.ContainingAssembly.Name != "mscorlib";
```
sharwell: 📝 This recently came up in the context of another rule.

mavasani: Can we make this an extension method on Compilation? Can be done in a separate PR...

sharwell: ➡️ #3017

sharwell: Heh, I commented on this right before you did 👍

PR #2228, opened 2019-03-20
WIP: TypeConversionAllocationAnalyzer to use IOperation

edespong:
Change of the TypeConversionAllocationAnalyzer to use IOperation. It is nice to note that the TypeConversionAllocationAnalyzer.cs went from around 350 lines to 100. This PR uses the same AbstractAllocationAnalyzer code as #2214 but is not based off that branch as the diff added a lot of noise.

The code is still in the `Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers` project as moving it creates massive amount of noise in the diff. If the current changes are ok, I can move the resources etc as well.

Changes in logic:
* In the original code, the analyzer made difference between
       ```Func<object, string> func2 = fooObjCall;```
and 
        ```Func<object, string> func1 = new Func<object, string>(fooObjCall);```
where it warned for the former but not fot the latter. While this might make sense from an educational perspective, there is no difference in allocations between them, so both now generates a diagnostic.
* Removed ReadonlyMethodGroupAllocationRule and let it fall under MethodGroupAllocationRule. It was reported by @sharwell here https://github.com/Microsoft/RoslynClrHeapAllocationAnalyzer/issues/32 and might still be relevant to have, but I think it is quite unlikely that code where the scenario happens would be marked as performance sensitive.
* The conversion also fixed issue https://github.com/Microsoft/RoslynClrHeapAllocationAnalyzer/issues/66 which was reported after the analyzers were pulled from the RoslynClrHeapAllocationAnalyzer project. Added unit test to verify
* Made some of the larger unit tests more granular

PR #3016, opened 2019-11-06
Convert the CallSiteImplicitAllocationAnalyzer to use IOperation

edespong:
This PR converts CallSiteImplicitAllocationAnalyzer to use IOperation instead of syntax.

One thing I did not like with the old code and still do not like with the new is that the code:

`FunctionCall(new [] { 1, 2 });`

reports (if FunctionCall has params object[])

> This call site is calling into a function with a 'params' parameter. This results in an array allocation even if no parameter is passed in for the params parameter

While technically true, the message is confusing. It is not obvious that the `int[]` is converted into a `object[] `. I have tried to differentiate between this case and the others, but have not been able to find a correct way of doing it. 

Could possibly mitigate by altering the message to something like:

> Function call to a function with a 'params {0}' parameter. This will result in an array allocation.

Where {0} is the params type.

### DIFF
--- a/src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
+++ b/src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
@@ -2,17 +2,15 @@
 
 using System;
 using System.Collections.Immutable;
-using System.Runtime.CompilerServices;
-using System.Threading;
 using Analyzer.Utilities;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
 using Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers;
 
 namespace Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers
 {
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAnalyzer<SyntaxKind>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAnalyzer
     {
         public const string ParamsParameterRuleId = "HAA0101";
         public const string ValueTypeNonOverridenCallRuleId = "HAA0102";
@@ -42,65 +40,33 @@ namespace Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(ParamsParameterRule, ValueTypeNonOverridenCallRule);
 
-        protected override ImmutableArray<SyntaxKind> Expressions => ImmutableArray.Create(SyntaxKind.InvocationExpression);
+        protected override ImmutableArray<OperationKind> Operations => ImmutableArray.Create(OperationKind.Invocation);
 
         private static readonly object[] EmptyMessageArgs = Array.Empty<object>();
 
-        protected override void AnalyzeNode(SyntaxNodeAnalysisContext context, in PerformanceSensitiveInfo info)
+        protected override void AnalyzeNode(OperationAnalysisContext context, in PerformanceSensitiveInfo info)
         {
-            var node = context.Node;
-            var semanticModel = context.SemanticModel;
-            Action<Diagnostic> reportDiagnostic = context.ReportDiagnostic;
-            var cancellationToken = context.CancellationToken;
-
-            var invocationExpression = node as InvocationExpressionSyntax;
-
-            if (semanticModel.GetSymbolInfo(invocationExpression, cancellationToken).Symbol is IMethodSymbol methodInfo)
+            if (!(context.Operation is IInvocationOperation invocation))
             {
-                if (methodInfo.IsOverride)
-                {
-                    CheckNonOverridenMethodOnStruct(methodInfo, reportDiagnostic, invocationExpression);
-                }
+                return;
+            }
 
-                if (methodInfo.Parameters.Length > 0 && invocationExpression.ArgumentList != null)
+            if (invocation.TargetMethod.IsOverride)
+            {
+                var type = invocation.TargetMethod.ContainingType;
+                if (string.Equals(type.Name, "ValueType", StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(type.Name, "Enum", StringComparison.OrdinalIgnoreCase))
                 {
-                    var lastParam = methodInfo.Parameters[methodInfo.Parameters.Length - 1];
-                    if (lastParam.IsParams)
-                    {
-                        CheckParam(invocationExpression, methodInfo, semanticModel, reportDiagnostic, cancellationToken);
-                    }
+                    context.ReportDiagnostic(Diagnostic.Create(ValueTypeNonOverridenCallRule, invocation.Syntax.GetLocation(), EmptyMessageArgs));
                 }
             }
-        }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private static void CheckParam(InvocationExpressionSyntax invocationExpression, IMethodSymbol methodInfo, SemanticModel semanticModel, Action<Diagnostic> reportDiagnostic, CancellationToken cancellationToken)
-        {
-            var arguments = invocationExpression.ArgumentList.Arguments;
-            if (arguments.Count != methodInfo.Parameters.Length)
+            if (invocation.Arguments.Length > 0)
             {
-                reportDiagnostic(Diagnostic.Create(ParamsParameterRule, invocationExpression.GetLocation(), EmptyMessageArgs));
-            }
-            else
-            {
-                var lastIndex = arguments.Count - 1;
-                var lastArgumentTypeInfo = semanticModel.GetTypeInfo(arguments[lastIndex].Expression, cancellationToken);
-                if (lastArgumentTypeInfo.Type != null && !lastArgumentTypeInfo.Type.Equals(methodInfo.Parameters[lastIndex].Type))
+                var lastArgument = invocation.Arguments[invocation.Arguments.Length - 1];
+                if (lastArgument.ArgumentKind == ArgumentKind.ParamArray && lastArgument.IsImplicit)
                 {
-                    reportDiagnostic(Diagnostic.Create(ParamsParameterRule, invocationExpression.GetLocation(), EmptyMessageArgs));
-                }
-            }
-        }
-
-        private static void CheckNonOverridenMethodOnStruct(IMethodSymbol methodInfo, Action<Diagnostic> reportDiagnostic, SyntaxNode node)
-        {
-            if (methodInfo.ContainingType != null)
-            {
-                // hack? Hmmm.
-                var containingType = methodInfo.ContainingType.ToString();
-                if (string.Equals(containingType, "System.ValueType", StringComparison.OrdinalIgnoreCase) || string.Equals(containingType, "System.Enum", StringComparison.OrdinalIgnoreCase))
-                {
-                    reportDiagnostic(Diagnostic.Create(ValueTypeNonOverridenCallRule, node.GetLocation(), EmptyMessageArgs));
+                    context.ReportDiagnostic(Diagnostic.Create(ParamsParameterRule, invocation.Syntax.GetLocation(), EmptyMessageArgs));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


sharwell:
Marking as blocked on #2228 just so I don't get confused

edespong:
Running against Roslyn, I got 193 differing diagnostics, all of them HAA0101.

New true positives:
- [Calling a constructor via this](https://github.com/dotnet/roslyn/blob/de140f9fd3f5dcb82f17916c0ca79de4cf094b61/src/Compilers/Core/CodeAnalysisTest/ObjectSerializationTests.cs#L1245)
- [Calling a method with a shorthand value-tuple params array](https://github.com/dotnet/roslyn/blob/0b3b6d5a28c7b5e10dd312d339b210b1b5c9cef0/src/ExpressionEvaluator/CSharp/Test/ExpressionCompiler/ReferencedModulesTests.cs#L68)

New false positive:
- [Calling a method with the params parameter as a named parameter](https://github.com/dotnet/roslyn/blob/1f0ce4646b5d4471ea6d1bdac79495b22653e462/src/CodeStyle/Core/Analyzers/AbstractCodeStyleDiagnosticAnalyzer.cs#L73)

Removed false positives:
- Calling method with object[] params with a array of reference type objects, e.g. calling `string.Format(string, object[])` with a `string[]`

Evangelink:
@edespong @sharwell What is missing on this PR to get it merged?

edespong:
If I recall correctly, this PR was ready to be merged after review.

Evangelink:
@edespong Would you mind fixing the conflict?

@sharwell @mavasani could you review?

Evangelink:
Ping @mavasani

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -2,17 +2,15 @@
 
 using System;
```
Youssef1313, line 3: This file should be moved to the "Core" layer now as it's language-agnostic.

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -42,65 +40,33 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
...
         {
-            var node = context.Node;
-            var semanticModel = context.SemanticModel;
-            Action<Diagnostic> reportDiagnostic = context.ReportDiagnostic;
-            var cancellationToken = context.CancellationToken;
-
-            var invocationExpression = node as InvocationExpressionSyntax;
-
-            if (semanticModel.GetSymbolInfo(invocationExpression, cancellationToken).Symbol is IMethodSymbol methodInfo)
+            if (!(context.Operation is IInvocationOperation invocation))
```
Youssef1313, line 49: I believe you can cast the operation directly. The registered operation above are only OperationKind.Invocation, so it seems impossible for context.Operation not to be IInvocationOperation.

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -42,65 +40,33 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(ParamsParameterRule, ValueTypeNonOverridenCallRule);
 
-        protected override ImmutableArray<SyntaxKind> Expressions => ImmutableArray.Create(SyntaxKind.InvocationExpression);
+        protected override ImmutableArray<OperationKind> Operations => ImmutableArray.Create(OperationKind.Invocation);
 
         private static readonly object[] EmptyMessageArgs = Array.Empty<object>();
 
-        protected override void AnalyzeNode(SyntaxNodeAnalysisContext context, in PerformanceSensitiveInfo info)
+        protected override void AnalyzeNode(OperationAnalysisContext context, in PerformanceSensitiveInfo info)
```
Youssef1313, line 47: The naming of this method doesn't seem to be correct.

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -42,65 +40,33 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
...
-            }
-            else
+            if (invocation.TargetMethod.IsOverride)
             {
-                var lastIndex = arguments.Count - 1;
-                var lastArgumentTypeInfo = semanticModel.GetTypeInfo(arguments[lastIndex].Expression, cancellationToken);
-                if (lastArgumentTypeInfo.Type != null && !lastArgumentTypeInfo.Type.Equals(methodInfo.Parameters[lastIndex].Type))
+                var type = invocation.TargetMethod.ContainingType;
+                if (string.Equals(type.Name, "ValueType", StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(type.Name, "Enum", StringComparison.OrdinalIgnoreCase))
```
Youssef1313, line 58: I think comparing actual symbols is much better than string comparison here. But let someone else confirm.

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -42,65 +40,33 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(ParamsParameterRule, ValueTypeNonOverridenCallRule);
 
-        protected override ImmutableArray<SyntaxKind> Expressions => ImmutableArray.Create(SyntaxKind.InvocationExpression);
+        protected override ImmutableArray<OperationKind> Operations => ImmutableArray.Create(OperationKind.Invocation);
 
         private static readonly object[] EmptyMessageArgs = Array.Empty<object>();
```
Youssef1313, line 45: Not sure if this is needed

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -149,5 +149,97 @@ public void Foo()
...
+        A((""1"", 1), (""2"", 2));
+    }
+
+    public void A(params (string x, int y)[] args)
+    {
+    }
+}";
+
+            await VerifyCS.VerifyAnalyzerAsync(sampleProgram,
+                VerifyCS.Diagnostic(CallSiteImplicitAllocationAnalyzer.ParamsParameterRule).WithLocation(10, 9));
```
Youssef1313, line 242: The two-argument form of WithLocation isn't preferred.

You can put `[|` and `|]` around the span where the diagnostic is expected in the source itself, and pass only the source to VerifyAnalyzerAsync.

Youssef1313 (commented):
Consider adding tests for VB as well.

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
Evangelink: +1 for the symbol comparison.

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -42,65 +40,33 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
...
             {
-                var lastIndex = arguments.Count - 1;
-                var lastArgumentTypeInfo = semanticModel.GetTypeInfo(arguments[lastIndex].Expression, cancellationToken);
-                if (lastArgumentTypeInfo.Type != null && !lastArgumentTypeInfo.Type.Equals(methodInfo.Parameters[lastIndex].Type))
+                var type = invocation.TargetMethod.ContainingType;
+                if (string.Equals(type.Name, "ValueType", StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(type.Name, "Enum", StringComparison.OrdinalIgnoreCase))
                 {
-                    reportDiagnostic(Diagnostic.Create(ParamsParameterRule, invocationExpression.GetLocation(), EmptyMessageArgs));
+                    context.ReportDiagnostic(Diagnostic.Create(ValueTypeNonOverridenCallRule, invocation.Syntax.GetLocation(), EmptyMessageArgs));
```
Evangelink, line 60: You don't need the empty args, also please use the `CreateDiagnostic` helper method instead.

```suggestion
                    context.ReportDiagnostic(invocation.CreateDiagnostic(ValueTypeNonOverridenCallRule));
```

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -42,65 +40,33 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
...
-            if (methodInfo.ContainingType != null)
+            for (int i = invocation.Arguments.Length - 1; i >= 0; i--)
             {
-                // hack? Hmmm.
-                var containingType = methodInfo.ContainingType.ToString();
-                if (string.Equals(containingType, "System.ValueType", StringComparison.OrdinalIgnoreCase) || string.Equals(containingType, "System.Enum", StringComparison.OrdinalIgnoreCase))
+                if (invocation.Arguments[i].ArgumentKind == ArgumentKind.ParamArray && invocation.Arguments[i].IsImplicit)
                 {
-                    reportDiagnostic(Diagnostic.Create(ValueTypeNonOverridenCallRule, node.GetLocation(), EmptyMessageArgs));
+                    context.ReportDiagnostic(Diagnostic.Create(ParamsParameterRule, invocation.Syntax.GetLocation(), EmptyMessageArgs));
```
Evangelink, line 68: Same remark here

```suggestion
                    context.ReportDiagnostic(invocation.CreateDiagnostic(ParamsParameterRule));
```

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -149,5 +149,97 @@ public void Foo()
...
+            var sampleProgram = @"
+using System.IO;
+using Roslyn.Utilities;
+
+public class MyClass
+{
+    [PerformanceSensitive(""uri"")]
+    public void Testing()
+    {
+        string[] args = new [] { ""foo"", ""bar""};
```
Evangelink, line 165: Please rewrite the tests to avoid the use of "foo", "bar" words. They are disallowed by Microsoft.

@mavasani Might be worth updating the contribution doc and maybe create an analyzer for these, WDYT?

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -149,5 +149,97 @@ public void Foo()
...
+    {
+    }
+
+    public MyClass(params object[] args)
+    {
+    }
+}";
+
+            await VerifyCS.VerifyAnalyzerAsync(sampleProgram,
+                VerifyCS.Diagnostic(CallSiteImplicitAllocationAnalyzer.ParamsParameterRule).WithLocation(9, 9));
```
Evangelink, line 194: Please do not use this `WithLocation` overload instead use one of the following options:
- (preferred option when no arguments) the markup syntax (e.g. `[|MyClass|]` and no need for diagnostic)
- (preferred option when arguments) the markup syntax (e.g. `{|#0:MyClass|}` and `WithLocation(0)`)
- `WithSpan`

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -149,5 +149,97 @@ public void Foo()
...
+        A(args: """");
+    }
+
+    public void A(string x = """", params object[] args)
+    {
+    }
+}";
+
+            await VerifyCS.VerifyAnalyzerAsync(sampleProgram,
+                VerifyCS.Diagnostic(CallSiteImplicitAllocationAnalyzer.ParamsParameterRule).WithLocation(10, 9));
```
Evangelink, line 218: Same remark here.

Evangelink:
Hi @edespong, was it closed by error? Are you willing to continue the work on this?

edespong:
No,  I lost interest due to the lack of interest. The code is still available if someone finds it valuable and wants to pick it up.

PR #3035, opened 2019-11-11
[WIP] Convert remanining tests to the new test framework

Evangelink:
### Description

I am trying to convert the remaining tests to the new test framework.

This PR is currently based upon #3005 to avoid duplicated work but as soon as the other PR is merge I will adjust the base.

### Problems

I am facing some issues with the migration and would be grateful for some help/input from your side (cc @sharwell).

~~1. Some tests are failing with a weird error about issues not being listed as expected while they are. Because of the error message I suspect this is linked to generated code but I haven't yet found how the test needs to be tweaked to work.~~

```
Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessageException : Context: Diagnostics of test state
    Context: Verifying exclusions in <auto-generated> code
    Mismatch between number of diagnostics returned, expected "0" actual "1"
```

2. Some of the tests need a reference toward the ASP.NET dlls. I have seen that I need to use the package reference helper but I am not sure which version I shall be using for the tests.

3. ~~I am facing some `InvalidOperationException` issue with the new test framework when not specifying the `DiagnosticDescriptor` because the `SupportedDiagnostics` property is empty for `DisposeMethodsShouldCallBaseClassDispose` (see https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs#L39). I am not sure whether I can safely change this empty array to an array with the rule.~~

4. ~~I don't know how to translate the tests using `VerifyCSharpAcrossTwoAssemblies`. The `Sources` property of the new test framework allows to provide multiple files not assemblies.~~

~~5. I am not sure how to translate the `VerifyUnsafe` calls.~~

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/FixAnalyzers/FixerWithFixAllAnalyzerTests.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/FixAnalyzers/FixerWithFixAllAnalyzerTests.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Analyzers.FixAnalyzers;
 using Microsoft.CodeAnalysis.CodeActions;
@@ -73,8 +74,7 @@ public class MyDerivedCodeActionWithEquivalenceKey : MyAbstractCodeActionWithEqu
 }
 ";
         private async Task TestCSharpCoreAsync(string source, DiagnosticResult missingGetFixAllProviderOverrideDiagnostic,
-            bool withCustomCodeActions = false, CompilerDiagnostics compilerDiagnostics = CompilerDiagnostics.Errors,
-            params DiagnosticResult[] expected)
+            bool withCustomCodeActions = false, params DiagnosticResult[] expected)
         {
             var fixAllProviderString = @"public override FixAllProvider GetFixAllProvider()
     {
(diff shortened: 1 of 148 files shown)


Evangelink:
Some of the tests are relying on markup syntax but this syntax doesn't support analyzers with multiple rules and the syntax `{|CAXXX:|}` doesn't support analyzers with multiple rules using the same ID. Shall I convert these tests to a no scope syntax?

Evangelink:
@mavasani @sharwell no rush but if you have some ideas to unlock some of the "blocking" points I have described that would be awesome to help me move forward with the changes.

Evangelink:
For all the issues related to (1) it seems that the problem is inside `Microsoft.CodeAnalysis.Testing.AnalyzerTest<T>` lines 217 to 222:

```csharp
// Diagnostics reported by the compiler and analyzer diagnostics which don't have a location will
// still be reported. We also insert a new line at the beginning so we have to move all diagnostic
// locations which have a specific position down by one line.
var expectedResults = expected
	.Where(x => !IsSubjectToExclusion(x, sources))
	.Select(x => IsInSourceFile(x, sources) ? x.WithLineOffset(1) : x)
	.ToArray();
```

The `expectedResults` is empty while it should contain for this case the expected diagnostics with the line shift. Sadly the method handling the exclusion doesn't show any hint toward how I could make sure those issues aren't excluded:

```csharp
private static bool IsSubjectToExclusion(DiagnosticResult result, (string filename, SourceText content)[] sources)
{
	if (result.Id.StartsWith("CS", StringComparison.Ordinal)
		|| result.Id.StartsWith("BC", StringComparison.Ordinal))
	{
		// This is a compiler diagnostic
		return false;
	}

	if (result.Id.StartsWith("AD", StringComparison.Ordinal))
	{
		// This diagnostic is reported by the analyzer infrastructure
		return false;
	}

	if (result.Spans.IsEmpty)
	{
		return false;
	}

	if (!IsInSourceFile(result, sources))
	{
		// This diagnostic is not reported in a source file
		return false;
	}

	return true;
}
```

sharwell:
> Some tests are failing with a weird error about issues not being listed as expected while they are. 

This occurs when an analyzer is configured to report diagnostics in generated code, but the test is not configured the same way. In terms of this repository, it means that `VerifyCS` is set to `CSharpCodeFixVerifier` instead of `CSharpSecurityCodeFixVerifier`, or similar for `VerifyVB`. If it's not obvious that diagnostics in generated code are desired, consider opening an issue to make a decision regarding the behavior.

sharwell:
> Some of the tests need a reference toward the ASP.NET dlls. I have seen that I need to use the package reference helper but I am not sure which version I shall be using for the tests.

The test should previously have a package reference as part of the test project, which is then used in the tests. You could use the same version of in the PackageIdentity when referencing it using the new NuGet package support.

> Some of the tests are relying on markup syntax but this syntax doesn't support analyzers with multiple rules and the syntax `{|CAXXX:|}` doesn't support analyzers with multiple rules using the same ID. Shall I convert these tests to a no scope syntax?

You can either convert them, or use [`MarkupOptions.UseFirstDescriptor`](https://github.com/dotnet/roslyn-sdk/blob/66b2eb24a0e1e8fe31ff777a8255ceecbcbddb51/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/MarkupOptions.cs#L18-L23) if it applies to the scenario.

PR #3041, opened 2019-11-14
Add one pager for writing custom dataflow analysis and dataflow based…

mavasani:
… analyzers in our repo


docs/Writing dataflow analysis based analyzers.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Well-known flow analyses
+
+We have some common analyses that you may likely want to consume for your custom dataflow analysis implementation or use directly in dataflow analyzers:
+
+1. [PointsToAnalysis](https://github.com/dotnet/roslyn-analyzers/blob/v2.9.7/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.cs): Dataflow analysis to track locations pointed to by AnalysisEntity and IOperation instances. This is the most commonly used dataflow analysis in all our flow based analyzers/analyses.
+2. [ValueContentAnalysis](https://github.com/dotnet/roslyn-analyzers/blob/v2.9.7/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ValueContentAnalysis/ValueContentAnalysis.cs): Dataflow analysis to track possible constant values that might be stored in an AnalysisEntity and IOperation instances. This is identical to constant propagation for constant values stored in non-constant symbols.
+3. [TaintedDataAnalysis](https://github.com/dotnet/roslyn-analyzers/blob/v2.9.7/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.cs): Dataflow analysis to track tainted state of AnalysisEntity and IOperation instances.
+
```
dotpaul: Is this documentation intended for Microsoft.CodeAnalysis.FlowAnalysis.Utilities? Cuz TaintedData isn't in there, right? (Unless you're signing up for supporting it as a consumable API 🙂)

mavasani: Ah, no. This is primarily intended for writing flow based analyzers within our repo.

dotpaul: Oh okay. Could mention PropertySetAnalysis, intended to find incorrect usage of an object based on a set of its properties.

nschuessler:
So this may not be related to this doc, but explaining the `PointsToAnalysisResult` was really important for me to get my analyzer working. Basically, knowing it resolves all the aliases to identify which are common and how to inspect them could be important.

```
string x = "my string";
string y = x;

a = x.ToLower()
b = y.ToLower();
```

(this is a redundant operation which you can determine because if I remember correctly the PointsToAnalysis will return the same object)

mavasani:
@nschuessler Thanks, that is a very good point. I'll expand a bit more on points to analysis and also add the example you provided.

docs/Writing dataflow analysis based analyzers.md
```diff
@@ -0,0 +1,139 @@
...
+int c3 = c1 + c2;
+int c4 = flag ? c3 : c3 + param;     // assume 'param' is a parameter for this method block with unknown value content from callers.
+```
+ValueContentAnalysis will compute that variables `c1`, `c2` and `c3` have identical `ValueContentAbstractValue` with a single literal value `0` and `ValueContainsNonLiteralState.No` to indicate it cannot contain a non-literal value. It will compute that `c4` has a different `ValueContentAbstractValue` with a single literal value `0` and `ValueContainsNonLiteralState.Maybe` to indicate that it may contain some non-literal value(s) in some code path(s).
+
+4. [TaintedDataAnalysis](https://github.com/dotnet/roslyn-analyzers/blob/v2.9.7/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.cs): Dataflow analysis to track tainted state of AnalysisEntity and IOperation instances.
+5. [PropertySetAnalysis](https://github.com/dotnet/roslyn-analyzers/blob/v2.9.7/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysis.cs): Dataflow analysis to track values assigned to more then one property of an object to identify and flag incorrect/insecure object state.
```
dotpaul: >more then one property [](start = 220, length = 22)

How abouts, "one or more properties"?

mavasani:
```suggestion
5. [PropertySetAnalysis](https://github.com/dotnet/roslyn-analyzers/blob/v2.9.7/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysis.cs): Dataflow analysis to track values assigned to one or more properties of an object to identify and flag incorrect/insecure object state.
```

PR #3050, opened 2019-11-18
Fix null cfg bug.

LLLXXXCCC:


src/Microsoft.NetCore.Analyzers/Core/Security/SourceTriggeredTaintedDataAnalyzerBase.cs
```diff
@@ -77,6 +77,11 @@ public override void Initialize(AnalysisContext context)
                             }
 
                             ControlFlowGraph cfg = operationBlockStartContext.OperationBlocks.GetControlFlowGraph();
+                            if (cfg == null)
```
mavasani: Can we create a unit test for this?

dotpaul: Looks like `Test_NullCfg_NoDiagnostic` cover this.

Any performance concerns with `GetControlFlowGraph()` off the top of your head?

mavasani: Oh are we creating a CFG for every operation block? We should certainly avoid that. CFG should only be computed for methods where we know will be perfoming DFA.

mavasani: @dotpaul Can we create a work item/PR to ensure we create CFG lazily at the point where we know we will be executing DFA?

dotpaul: I was hoping @LLLXXXCCC could fix this instance in this PR. I think this is the only place where the CFG is can be created without attempting DFA.

dotpaul: I'll fix a new PR.

PR #3052, opened 2019-11-18
Convert remaining tests to the new test framework

Evangelink:
## Description

Convert remaining tests to the new test framework. This needs to be rebased on master after #2995 is closed.

## Remaining work

~~- All the tests with a reference to `System.Web.UI` are failing. I couldn't find how to make them work. I have tried to add a reference to the package ASP NET Mvc but it still fails.~~
Fixed by adding a simple implementation as it is done for other tests.

~~- All the tests where I have tried to add an additional project (to mimic the `VerifyAcrossAssemblies` method) are missing references to everything. Trying to add `AdditionalMetadataRefernce.Default` didn't solve the issue.~~


src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/DoNotDirectlyAwaitATaskTests.cs
```diff
@@ -380,25 +382,23 @@ public async Task M()
...
-            editorConfigText = "dotnet_code_quality.CA2007.output_kind = ConsoleApplication, WindowsApplication";
-            VerifyCSharp(code, GetEditorConfigAdditionalFile(editorConfigText), compilationOptions, parseOptions: null);
+"
+                    },
+                    AdditionalFiles = { (".editorconfig", editorConfigText) }
+                },
+                SolutionTransforms =
+                {
+                    (solution, projectId) =>
+                        solution.WithProjectCompilationOptions(projectId, new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary))
```
sharwell: 📝 This is already the default, so there is no need to specify a solution transform for it

src/Microsoft.NetFramework.Analyzers/UnitTests/TypesShouldNotExtendCertainBaseTypesTests.cs
```diff
@@ -347,97 +347,97 @@ End Class
         private static DiagnosticResult GetCSharpCollectionBaseResultAt(int line, int column, string declaredTypeName, string badBaseTypeName)
         {
             string message = string.Format(CultureInfo.CurrentCulture, MicrosoftNetFrameworkAnalyzersResources.TypesShouldNotExtendCertainBaseTypesMessageSystemCollectionsCollectionBase, declaredTypeName, badBaseTypeName);
-            return new DiagnosticResult(TypesShouldNotExtendCertainBaseTypesAnalyzer.Rule).WithLocation(line, column).WithArguments(message);
+            return VerifyCS.Diagnostic(TypesShouldNotExtendCertainBaseTypesAnalyzer.Rule).WithLocation(line, column).WithArguments(message);
```
sharwell: 💡 Seems like this can be simplified based on lines below

```suggestion
            return VerifyCS.Diagnostic().WithLocation(line, column).WithArguments(message);
```

src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs
```diff
@@ -36,8 +36,7 @@ public sealed class DisposeMethodsShouldCallBaseClassDispose : DiagnosticAnalyze
                                                                              helpLinkUri: "https://docs.microsoft.com/visualstudio/code-quality/ca2215-dispose-methods-should-call-base-class-dispose",
                                                                              customTags: FxCopWellKnownDiagnosticTags.PortedFxCopRule);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray<DiagnosticDescriptor>.Empty;
-        //ImmutableArray.Create(Rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
```
sharwell: @mavasani Is there a reason this analyzer was disabled by removing diagnostic IDs?

src/Microsoft.NetCore.Analyzers/UnitTests/Data/ReviewSQLQueriesForSecurityVulnerabilitiesTests.cs
```diff
@@ -1580,20 +1586,20 @@ End Sub
         }
 
         [Fact]
-        public void MissingWellKnownTypes_NoDiagnostic()
+        public async Task MissingWellKnownTypes_NoDiagnostic()
```
sharwell: 💭 I'm not sure this test is still correct. You should make sure the test fails if assembly is provided, by adding a using alias directive that references a type defined in System.Data (ideally one of the types used by the analyzer).

src/Microsoft.NetCore.Analyzers/UnitTests/Data/ReviewSQLQueriesForSecurityVulnerabilitiesTests.cs
```diff
@@ -1580,20 +1586,20 @@ End Sub
...
-class C { }", ReferenceFlags.RemoveSystemData);
+            await VerifyCS.VerifyAnalyzerAsync(@"
+class C { }");
 
-            VerifyBasic(@"
-Class C
-End Class", ReferenceFlags.RemoveSystemData);
+            await VerifyVB.VerifyAnalyzerAsync(@"
+            Class C
+            End Class");
```
sharwell: 📝 These lines are part of a verbatim string literal, so the indentation should not change

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1100,29 +1099,17 @@ public class C
...
-            var message = string.Format(CultureInfo.CurrentCulture,
-                isSpecial ?
-                    MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectSuffixMessageSpecialCollection :
-                    MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectSuffixMessageDefault,
-                symbolName,
-                replacementName);
-            return new DiagnosticResult(IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
+            => VerifyCS.Diagnostic(isSpecial
+                    ? IdentifiersShouldHaveCorrectSuffixAnalyzer.SpecialCollectionRule
+                : IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
```
sharwell:
```suggestion
                    : IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1100,29 +1099,17 @@ public class C
...
-            var message = string.Format(CultureInfo.CurrentCulture,
-               isSpecial ?
-                   MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectSuffixMessageSpecialCollection :
-                   MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectSuffixMessageDefault,
-               symbolName,
-               replacementName);
-            return new DiagnosticResult(IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
+            => VerifyVB.Diagnostic(isSpecial
+                    ? IdentifiersShouldHaveCorrectSuffixAnalyzer.SpecialCollectionRule
+                : IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
```
sharwell:
```suggestion
                    : IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
```

src/Microsoft.NetCore.Analyzers/UnitTests/Security/UseXmlReaderForValidatingReaderTests.cs
```diff
@@ -64,14 +62,22 @@ public void TestMethod(XmlReader xmlReader)
 }");
         }
 
-        protected override DiagnosticAnalyzer GetBasicDiagnosticAnalyzer()
+        private static async Task VerifyCSharp(string source, params DiagnosticResult[] expected)
```
sharwell: 💡 Use the naming convention from the new test framework.

```suggestion
        private static async Task VerifyCSharpAnalyzerAsync(string source, params DiagnosticResult[] expected)
```

src/Microsoft.NetCore.Analyzers/UnitTests/Security/UseXmlReaderForSchemaReadTests.cs
```diff
@@ -90,14 +88,22 @@ public XmlSchema Test
 }");
         }
 
-        protected override DiagnosticAnalyzer GetBasicDiagnosticAnalyzer()
+        private static async Task VerifyCSharp(string source, params DiagnosticResult[] expected)
```
sharwell: 💡 Use the naming convention from the new test framework

```suggestion
        private static async Task VerifyCSharpAnalyzerAsync(string source, params DiagnosticResult[] expected)
```

src/Microsoft.NetCore.Analyzers/UnitTests/Security/UseXmlReaderForSchemaReadTests.cs
```diff
@@ -90,14 +88,22 @@ public XmlSchema Test
...
         {
-            return new UseXmlReaderForSchemaRead();
-        }
+            var csharpTest = new VerifyCS.Test
+            {
+                TestCode = source,
+                TestBehaviors = TestBehaviors.SkipGeneratedCodeCheck
+            };
```
sharwell: 📝 This is just `CSharpSecurityCodeFixVerifier`. This helper method is unnecessary if you fix the definition of `VerifyCS` above.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/UseXmlReaderForXPathDocumentTests.cs
```diff
@@ -113,14 +116,22 @@ public void TestMethod(XmlReader reader, XmlSpace space)
...
 
-        protected override DiagnosticAnalyzer GetBasicDiagnosticAnalyzer()
+        private static async Task VerifyCSharp(string source, params DiagnosticResult[] expected)
         {
-            return new UseXmlReaderForXPathDocument();
-        }
+            var csharpTest = new VerifyCS.Test
+            {
+                TestCode = source,
+                TestBehaviors = TestBehaviors.SkipGeneratedCodeCheck
```
sharwell: 📝 All the `VerifyCSharp` methods that only exist to set `TestBehaviors` like this can be converted to use `CSharpSecurityCodeFixVerifier` and remove the intermediate helper method.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
```diff
@@ -465,9 +466,9 @@ public void Method()
         }
 
         [Fact]
-        public void UseTls12Or192_Diagnostic()
+        public async Task UseTls12Or192_Diagnostic()
         {
-            VerifyCSharp(@"
+            await VerifyCS.VerifyAnalyzerAsync(@"
```
Evangelink: This test is failing with an inconsistent order between the 2 diagnostics. If I invert the order or the 2 diagnostics it keeps failing.

sharwell: This should only occur when the two diagnostics have the same diagnostic ID, which is not the case here. Can you try specifying the span for these diagnostics instead of the location (i.e. use `WithSpan` instead of `WithLocation`) and see if the problem goes away?

Evangelink: Works perfectly fine! Thank you

Evangelink: Sorry! It worked for another test not for this one. The 2 issues are reported on the exact same span.

sharwell: And reordering them did not fix it, even when you specified `WithSpan`?

Evangelink: No it didn't work even after reordering. The span are the same for the 2 issues `WithSpan(9, 34, 9, 52)`.

mavasani: Tagging @dotpaul @LLLXXXCCC 

@Evangelink The security team are working off a separate `2.9.x` branch, and merge frequently into master. Your PR changing the test framework used by security analyzer unit tests is going to create merge issues between master and their branch. I think you have couple of options here:
1. Revert all the changes made to security analyzer unit tests in this PR. Then make these changes in 2.9.x branch and let it merge back into master with the automated merge PRs
2. Retarget your PR to 2.9.x branch. This might mean additional work for you to revert changes to files/tests only in master.

I think I would recommend the first approach, but will let @dotpaul make a call here.

Evangelink: I will wait for @dotpaul call then :)

dotpaul: Hey @Evangelink, I'd prefer the former approach that @mavasani suggested (revert changes in this PR, make a new PR tests under security/ for 2.9.x).

Evangelink:
@mavasani I have removed the changes related to `Security`.

@sharwell I think that I have found an issue with `Microsoft.CodeAnalysis.Testing.AnalyzerTest<TVerifier>`:

```csharp
private static bool IsInSourceFile(DiagnosticResult result, (string filename, SourceText content)[] sources)
{
	return sources.Any(source => source.filename.Equals(result.Spans[0].Span.Path));
}
```

When the expected issues are mixing locations an no location the code `result.Spans[0]` will fail.

Shall I create a ticket on Roslyn's side?

FYI when this problem is solved the PR shall be green.

sharwell:
> When the expected issues are mixing locations an no location the code result.Spans[0] will fail.

I think I fixed that already with dotnet/roslyn-sdk#408.

Evangelink:
@sharwell Do you know which version of the testing framework will contain this fix? I'd like to be able to update to see if I have other issues to fix.

Evangelink:
It worked! The PR is ready for review.

Evangelink:
I don't get the following CI issue `Maintainability\AvoidDeadConditionalCode_ValueContentAnalysis.cs(2722,13): error CS0103: The name 'VerifyCSharp' does not exist in the current context `.

The line doesn't exist and there is no `VerifyCSharp`, is it doing some kind of auto-merge that would be broken?

mavasani:
@Evangelink Can you please resolve the merge conflicts?

@sharwell Anything else blocking this PR?

src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs
Evangelink: ping @mavasani

Evangelink: @mavasani Do you have any info regarding why the rule was disabled?

Evangelink:
Also fixes #3116

src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs
mavasani: I can't recall as of now, but I would rather we not do this change in this PR. Can we file a separate issue to track this?

Evangelink: https://github.com/dotnet/roslyn-analyzers/issues/3141

src/Test.Utilities/CSharpProjectState.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Testing;
+
+namespace Microsoft.NetCore.Analyzers.Security.UnitTests
+{
+    /// <summary>
+    /// This class is a copy-paste from
+    /// https://github.com/dotnet/roslyn-sdk/blob/master/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.Analyzer.Testing/CSharpProjectState.cs
+    /// </summary>
+    public class CSharpProjectState : ProjectState
```
sharwell: 😕 Why are we adding this?

src/Test.Utilities/VisualBasicProjectState.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Testing;
+
+namespace Microsoft.NetCore.Analyzers.Security.UnitTests
+{
+    /// <summary>
+    /// This class is a copy-paste from
+    /// https://github.com/dotnet/roslyn-sdk/blob/master/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing/VisualBasicProjectState.vb
+    /// </summary>
+    public class VisualBasicProjectState : ProjectState
```
sharwell: 😕 Why are we adding this?

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -35,12 +35,12 @@ public static class AdditionalMetadataReferences
         internal static MetadataReference VisualBasicSymbolsReference { get; } = MetadataReference.CreateFromFile(typeof(VisualBasicCompilation).Assembly.Location);
         internal static MetadataReference WorkspacesReference { get; } = MetadataReference.CreateFromFile(typeof(Workspace).Assembly.Location);
         internal static MetadataReference SystemWebReference { get; } = MetadataReference.CreateFromFile(typeof(System.Web.HttpRequest).Assembly.Location);
-        internal static MetadataReference SystemRuntimeSerialization { get; } = MetadataReference.CreateFromFile(typeof(System.Runtime.Serialization.NetDataContractSerializer).Assembly.Location);
+        public static MetadataReference SystemRuntimeSerialization { get; } = MetadataReference.CreateFromFile(typeof(System.Runtime.Serialization.NetDataContractSerializer).Assembly.Location);
```
sharwell: 📝 This should not be necessary with the latest test framework support for package and framework references

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -35,12 +35,12 @@ public static class AdditionalMetadataReferences
...
         internal static MetadataReference TestReferenceAssembly { get; } = MetadataReference.CreateFromFile(typeof(OtherDll.OtherDllStaticMethods).Assembly.Location);
         internal static MetadataReference SystemDirectoryServices { get; } = MetadataReference.CreateFromFile(typeof(System.DirectoryServices.DirectoryEntry).Assembly.Location);
         internal static MetadataReference SystemXaml { get; } = MetadataReference.CreateFromFile(typeof(System.Xaml.XamlReader).Assembly.Location);
         internal static MetadataReference PresentationFramework { get; } = MetadataReference.CreateFromFile(typeof(System.Windows.Markup.XamlReader).Assembly.Location);
         public static MetadataReference SystemWeb { get; } = MetadataReference.CreateFromFile(typeof(System.Web.HttpRequest).Assembly.Location);
-        internal static MetadataReference SystemWebExtensions { get; } = MetadataReference.CreateFromFile(typeof(System.Web.Script.Serialization.JavaScriptSerializer).Assembly.Location);
+        public static MetadataReference SystemWebExtensions { get; } = MetadataReference.CreateFromFile(typeof(System.Web.Script.Serialization.JavaScriptSerializer).Assembly.Location);
```
sharwell: 📝 This should not be necessary with the latest test framework support for package and framework references

src/Test.Utilities/CSharpProjectState.cs
Evangelink: No longer required.

src/Test.Utilities/VisualBasicProjectState.cs
Evangelink: No longer required.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/DoNotCallOverridableMethodsInConstructorsTests.cs
```diff
@@ -337,12 +340,15 @@ Inherits System.ComponentModel.Component
...
-            Project project = document.Project.AddMetadataReference(MetadataReference.CreateFromFile(typeof(System.Web.UI.Control).Assembly.Location));
-            project = project.AddMetadataReference(MetadataReference.CreateFromFile(typeof(System.Windows.Forms.Control).Assembly.Location));
-            DiagnosticAnalyzer analyzer = GetBasicDiagnosticAnalyzer();
-            GetSortedDiagnostics(analyzer, project.Documents.Single()).Verify(analyzer, GetDefaultPath(LanguageNames.VisualBasic));
+"
+                    },
+                    AdditionalReferences =
+                    {
+                        MetadataReference.CreateFromFile(typeof(System.Web.UI.Control).Assembly.Location),
+                        MetadataReference.CreateFromFile(typeof(System.Windows.Forms.Control).Assembly.Location),
```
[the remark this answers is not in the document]
Evangelink: Is there a package for WinForms?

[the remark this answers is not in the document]
Evangelink: Same question for WebForms.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumWithFlagsAttributeTests.cs
```diff
@@ -117,18 +106,18 @@ public enum SimpleFlagsEnumClass
 }}
 
 {0}
-[|public enum HexFlagsEnumClass
+public enum {{|CA1027:HexFlagsEnumClass|}}
```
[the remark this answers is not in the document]
Evangelink: Ok thank you!

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/DoNotCallOverridableMethodsInConstructorsTests.cs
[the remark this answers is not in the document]
Evangelink: Oh sorry, I just understood what you meant.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs
Evangelink: @mavasani @sharwell I cannot undo this change otherwise the tests are failing because the issue isn't raised.

mavasani: IMO that seems like a drawback/regression from the testing library that should be fixed. It shouldn't be requiring non-empty supported diagnostics just to enable testing of the analyzer.

sharwell: @mavasani The old test framework doesn't support this case either. It just silently passes all the tests by returning immediately here:

https://github.com/dotnet/roslyn-analyzers/blob/8e719bd21d008691303e385a76f6ec1a372ed51b/src/Test.Utilities/DiagnosticAnalyzerTests.Extensions.cs#L32-L36

Evangelink:
@sharwell Sorry to be a pain with this PR but could you have a look at the remaining files so that we can get a merge on?!

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -261,14 +266,18 @@ public new void Dispose()
     protected virtual void Dispose(bool disposing)
     {
     }
-}|]
-");
+}
+",
+                // Test0.cs(9,14): warning CA1063: Provide an overridable implementation of Dispose(bool) on 'B' or mark the type as sealed. A call to Dispose(false) should only clean up native resources. A call to Dispose(true) should clean up both managed and native resources.
+                GetCA1063BasicProvideDisposeBoolResultAt(9, 14, "B"),
```
sharwell:
```suggestion
                GetCA1063CSharpProvideDisposeBoolResultAt(9, 14, "B"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -470,20 +488,23 @@ protected virtual void Dispose(bool disposing)
...
 {
     public override void Dispose()
     {
     }
-}|]
+}
 ",
-            GetCA1063CSharpDisposeOverrideResultAt(24, 26, "C", "Dispose"));
+                // Test0.cs(6,25): warning CA1063: Ensure that 'B.Dispose' is declared as public and sealed.
+                GetCA1063BasicDisposeSignatureResultAt(6, 25, "B", "Dispose"),
```
sharwell:
```suggestion
                GetCA1063CSharpDisposeSignatureResultAt(6, 25, "B", "Dispose"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -511,25 +532,30 @@ public override void Dispose()
...
     public override void Dispose()
     {
         Dispose(true);
     }
-}|]
+}
 ",
-            GetCA1063CSharpDisposeOverrideResultAt(31, 26, "C", "Dispose"));
+                // Test0.cs(6,25): warning CA1063: Ensure that 'A.Dispose' is declared as public and sealed.
+                GetCA1063BasicDisposeSignatureResultAt(6, 25, "A", "Dispose"),
```
sharwell:
```suggestion
                GetCA1063CSharpDisposeSignatureResultAt(6, 25, "A", "Dispose"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -1704,14 +1754,18 @@ End Sub
     Protected Overridable Overloads Sub Dispose(disposing As Boolean)
     End Sub
 
-End Class|]
-");
+End Class
+",
+                // Test0.vb(10,14): warning CA1063: Provide an overridable implementation of Dispose(bool) on 'B' or mark the type as sealed. A call to Dispose(false) should only clean up native resources. A call to Dispose(true) should clean up both managed and native resources.
+                GetCA1063CSharpProvideDisposeBoolResultAt(10, 14, "B"),
```
sharwell:
```suggestion
                GetCA1063BasicProvideDisposeBoolResultAt(10, 14, "B"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -1758,24 +1817,28 @@ Implements IDisposable
...
     Implements ITest
 
     Public Property Test As Integer Implements ITest.Test
 
-End Class|]
-");
+End Class
+",
+                // Test0.vb(10,14): warning CA1063: Provide an overridable implementation of Dispose(bool) on 'B' or mark the type as sealed. A call to Dispose(false) should only clean up native resources. A call to Dispose(true) should clean up both managed and native resources.
+                GetCA1063CSharpProvideDisposeBoolResultAt(10, 14, "B"),
```
sharwell:
```suggestion
                GetCA1063BasicProvideDisposeBoolResultAt(10, 14, "B"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -1977,20 +2040,23 @@ End Sub
...
     Inherits B
 
     Public Overrides Sub Dispose()
     End Sub
-End Class|]
+End Class
 ",
-            GetCA1063BasicDisposeOverrideResultAt(25, 26, "C", "Dispose"));
+                // Test0.vb(7,28): warning CA1063: Ensure that 'B.Dispose' is declared as public and sealed.
+                GetCA1063CSharpDisposeSignatureResultAt(7, 28, "B", "Dispose"),
```
sharwell:
```suggestion
                GetCA1063BasicDisposeSignatureResultAt(7, 28, "B", "Dispose"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -1977,20 +2040,23 @@ End Sub
...
+                // Test0.vb(25,26): warning CA1063: Remove 'C.Dispose', override Dispose(bool disposing), and put the dispose logic in the code path where 'disposing' is true.
+                GetCA1063CSharpDisposeOverrideResultAt(25, 26, "C", "Dispose"));
```
sharwell:
```suggestion
                GetCA1063BasicDisposeOverrideResultAt(25, 26, "C", "Dispose"));
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -2018,21 +2084,26 @@ Inherits A
...
 
     Public Overrides Sub Dispose()
         Dispose(True)
     End Sub
-End Class|]
+End Class
 ",
-            GetCA1063BasicDisposeOverrideResultAt(32, 26, "C", "Dispose"));
+                // Test0.vb(7,28): warning CA1063: Ensure that 'A.Dispose' is declared as public and sealed.
+                GetCA1063CSharpDisposeSignatureResultAt(7, 28, "A", "Dispose"),
```
sharwell:
```suggestion
                GetCA1063BasicDisposeSignatureResultAt(7, 28, "A", "Dispose"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -2018,21 +2084,26 @@ Inherits A
...
+                // Test0.vb(25,26): warning CA1063: Remove 'B.Dispose', override Dispose(bool disposing), and put the dispose logic in the code path where 'disposing' is true.
+                GetCA1063CSharpDisposeOverrideResultAt(25, 26, "B", "Dispose"),
```
sharwell:
```suggestion
                GetCA1063BasicDisposeOverrideResultAt(25, 26, "B", "Dispose"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -2018,21 +2084,26 @@ Inherits A
...
+                // Test0.vb(32,26): warning CA1063: Remove 'C.Dispose', override Dispose(bool disposing), and put the dispose logic in the code path where 'disposing' is true.
+                GetCA1063CSharpDisposeOverrideResultAt(32, 26, "C", "Dispose"));
```
sharwell:
```suggestion
                GetCA1063BasicDisposeOverrideResultAt(32, 26, "C", "Dispose"));
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -2283,24 +2374,28 @@ Implements IDisposable
...
 
     Protected Overrides Sub Finalize()
         MyBase.Finalize()
     End Sub
-End Class|]
-");
+End Class
+",
+                // Test0.vb(4,14): warning CA1063: Provide an overridable implementation of Dispose(bool) on 'B' or mark the type as sealed. A call to Dispose(false) should only clean up native resources. A call to Dispose(true) should clean up both managed and native resources.
+                GetCA1063CSharpProvideDisposeBoolResultAt(4, 14, "B"),
```
sharwell:
```suggestion
                GetCA1063BasicProvideDisposeBoolResultAt(4, 14, "B"),
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs
```diff
@@ -2283,24 +2374,28 @@ Implements IDisposable
...
+                // Test0.vb(7,16): warning CA1063: Modify 'B.Dispose' so that it calls Dispose(true), then calls GC.SuppressFinalize on the current object instance ('this' or 'Me' in Visual Basic), and then returns.
+                GetCA1063CSharpDisposeImplementationResultAt(7, 16, "B", "Dispose"));
```
sharwell:
```suggestion
                GetCA1063BasicDisposeImplementationResultAt(7, 16, "B", "Dispose"));
```

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -26,6 +26,9 @@ public static class AdditionalMetadataReferences
         public static ReferenceAssemblies DefaultWithNewtonsoftJson { get; } = Default
             .AddPackages(ImmutableArray.Create(new PackageIdentity("Newtonsoft.Json", "10.0.1")));
 
+        public static ReferenceAssemblies DefaultWithSystemWebAndWinForms { get; } = Default
+            .AddAssemblies(ImmutableArray.Create("System.Windows.Forms", "System.Web"));
```
sharwell: 💡 Can we try restricting the new one to `DefaultWithWindowsForms`, and using the existing `DefaultWithSystemWeb` for the others? If one test requires both then I guess it's weird but fine.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
```diff
@@ -20,18 +19,15 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines.UnitTests
     [Trait(Traits.DataflowAnalysis, Traits.Dataflow.NullAnalysis)]
     [Trait(Traits.DataflowAnalysis, Traits.Dataflow.PointsToAnalysis)]
     [Trait(Traits.DataflowAnalysis, Traits.Dataflow.PredicateAnalysis)]
-    public partial class ValidateArgumentsOfPublicMethodsTests : DiagnosticAnalyzerTestBase
+    public partial class ValidateArgumentsOfPublicMethodsTests
```
sharwell:
```suggestion
    public class ValidateArgumentsOfPublicMethodsTests
```

src/Microsoft.NetCore.Analyzers/UnitTests/Resources/MarkAssembliesWithNeutralResourcesLanguageTests.cs
```diff
@@ -35,79 +28,85 @@ End Class
...
         [Fact]
-        public void TestBasicvalidAttribute()
+        public async Task TestBasicvalidAttribute()
         {
-            VerifyBasic(GetSources(@"<Assembly: System.Resources.NeutralResourcesLanguage(""en"")>", LanguageNames.VisualBasic));
+            await VerifyBasicWithDependencies(@"<Assembly: System.Resources.NeutralResourcesLanguage(""en"")>");
         }
 
-        private FileAndSource[] GetSources(string code, string language)
+        private async Task VerifyCSharpWithDependencies(string source, params DiagnosticResult[] expected)
```
sharwell:
```suggestion
        private async Task VerifyCSharpWithDependenciesAsync(string source, params DiagnosticResult[] expected)
```

src/Microsoft.NetCore.Analyzers/UnitTests/Resources/MarkAssembliesWithNeutralResourcesLanguageTests.cs
```diff
@@ -35,79 +28,85 @@ End Class
...
+            var csharpTest = new VerifyCS.Test { TestCode = source };
+
+            csharpTest.TestState.Sources.Add(("Test.Designer.cs", CSharpDesignerFile));
+            csharpTest.ExpectedDiagnostics.AddRange(expected);
+
+            await csharpTest.RunAsync();
         }
 
-        private FileAndSource GetDesignerFile(string language)
+        private async Task VerifyBasicWithDependencies(string source, params DiagnosticResult[] expected)
```
sharwell:
```suggestion
        private async Task VerifyBasicWithDependenciesAsync(string source, params DiagnosticResult[] expected)
```

src/Microsoft.NetCore.Analyzers/UnitTests/Resources/MarkAssembliesWithNeutralResourcesLanguageTests.cs
```diff
@@ -35,79 +28,85 @@ End Class
...
         {
-            return new FileAndSource
-            {
-                FilePath = "Test.Designer" + (language == LanguageNames.CSharp ? ".cs" : ".vb"),
-                Source = language == LanguageNames.CSharp ? CSharpDesignerFile : BasicDesignerFile
-            };
+            var vbTest = new VerifyVB.Test { TestCode = source };
+
+            vbTest.TestState.Sources.Add(("Test.Designer.vb", BasicDesignerFile));
```
sharwell:
```
var vbTest = new VerifyVB.Test
{
  TestState =
  {
    Sources =
    {
      source,
      ("Test.Designer.vb", BasicDesignerFile),
    },
  },
};
```

src/Microsoft.NetCore.Analyzers/UnitTests/Resources/MarkAssembliesWithNeutralResourcesLanguageTests.cs
```diff
@@ -35,79 +28,85 @@ End Class
...
+            await VerifyBasicWithDependencies(@"<Assembly: System.Resources.NeutralResourcesLanguage(""en"")>");
         }
 
-        private FileAndSource[] GetSources(string code, string language)
+        private async Task VerifyCSharpWithDependencies(string source, params DiagnosticResult[] expected)
         {
-            return new[] { GetDesignerFile(language), new FileAndSource { FilePath = null, Source = code } };
+            var csharpTest = new VerifyCS.Test { TestCode = source };
+
+            csharpTest.TestState.Sources.Add(("Test.Designer.cs", CSharpDesignerFile));
```
sharwell:
```
var csharpTest = new VerifyCS.Test
{
  TestState =
  {
    Sources =
    {
      source,
      ("Test.Designer.cs", CSharpDesignerFile),
    },
  },
};
```

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/DisposeMethodsShouldCallBaseClassDisposeTests.cs
```diff
@@ -775,10 +776,10 @@ End Sub
 End Class");
         }
 
-        [Fact, WorkItem(1654, "https://github.com/dotnet/roslyn-analyzers/issues/1654")]
-        public void BaseDisposeCalls_SomePaths_Diagnostic()
+        [Fact(Skip = "Analyzer isn't yet flow based."), WorkItem(1654, "https://github.com/dotnet/roslyn-analyzers/issues/1654")]
```
sharwell: ❔ I'm guessing this never worked, but we just never knew it?

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -16,18 +13,8 @@
 
 namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines.UnitTests
 {
-    public partial class RethrowToPreserveStackDetailsTests : DiagnosticAnalyzerTestBase
+    public partial class RethrowToPreserveStackDetailsTests
```
sharwell:
```suggestion
    public class RethrowToPreserveStackDetailsTests
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/RemoveEmptyFinalizersTests.cs
```diff
@@ -16,18 +13,8 @@
 
 namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines.UnitTests
 {
-    public partial class RemoveEmptyFinalizersTests : DiagnosticAnalyzerTestBase
+    public partial class RemoveEmptyFinalizersTests
```
sharwell:
```suggestion
    public class RemoveEmptyFinalizersTests
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/DoNotCallOverridableMethodsInConstructorsTests.cs
```diff
@@ -286,18 +280,23 @@ abstract class F : System.ComponentModel.Component
...
-            GetSortedDiagnostics(analyzer, project.Documents.Single()).Verify(analyzer, GetDefaultPath(LanguageNames.CSharp));
+"
+                    },
+                }
+            }.RunAsync();
         }
 
         [Fact]
-        public void CA2214SpecialInheritanceBasic()
+        public async Task CA2214SpecialInheritanceBasic()
```
sharwell: 💡 Can we break this test up into one test for System.Web and a second test for System.Windows.Forms?

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/DoNotCallOverridableMethodsInConstructorsTests.cs
```diff
@@ -237,9 +224,16 @@ End Class
         }
 
         [Fact]
-        public void CA2214SpecialInheritanceCSharp()
+        public async Task CA2214SpecialInheritanceCSharp()
```
sharwell: 💡 Can we break this test up into one test for System.Web and a second test for System.Windows.Forms?

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidDeadConditionalCode_NullAnalysis.cs
```diff
@@ -4757,16 +4788,16 @@ void M1(T t)
             return;
         }
 
-        var d = (D)t;   // Compiler error CS0030: Cannot convert type 'T' to 'D'
+        var d = {|CS0030:(D)t|};
```
sharwell: 💡 It would be good to keep the comment since the method name doesn't clearly indicate that the test covers compiler error cases.

```suggestion
        var d = {|CS0030:(D)t|};   // Compiler error CS0030: Cannot convert type 'T' to 'D'
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidDeadConditionalCode_NullAnalysis.cs
```diff
@@ -4780,20 +4811,20 @@ Class Test(Of T As C)
             Return
         End If
 
-        Dim d = DirectCast(t, D)    '  Compiler error BC30311: Value of type 'T' cannot be converted to 'D'
+        Dim d = DirectCast({|BC30311:t|}, D)
```
sharwell: 💡 It would be good to keep the comment since the method name doesn't clearly indicate that the test covers compiler error cases.

```suggestion
        Dim d = DirectCast({|BC30311:t|}, D)    '  Compiler error BC30311: Value of type 'T' cannot be converted to 'D'
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -629,109 +633,90 @@ class FifthDerivedClass : FourthDerivedClass { }";
...
-            return new DiagnosticResult(CodeMetricsAnalyzer.CA1505RuleId, DiagnosticHelpers.DefaultDiagnosticSeverity)
-                .WithLocation(line, column)
-                .WithMessageFormat(MicrosoftCodeQualityAnalyzersResources.AvoidUnmantainableCodeMessage)
-                .WithArguments(symbolName, metricValue, threshold);
-        }
+        private const string AdditionalFileName = "CodeMetricsConfig.txt";
 
-        // '{0}' is coupled with '{1}' different types from '{2}' different namespaces. Rewrite or refactor the code to decrease its class coupling below '{3}'.
-        private static DiagnosticResult GetCA1506ExpectedDiagnostic(int line, int column, string symbolName, int coupledTypesCount, int namespaceCount, int threshold)
+        private async Task VerifyCSharp(string source, string codeMetricsConfigSource, params DiagnosticResult[] expected)
```
sharwell:
```suggestion
        private async Task VerifyCSharpAsync(string source, string codeMetricsConfigSource, params DiagnosticResult[] expected)
```

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -629,109 +633,90 @@ class FifthDerivedClass : FourthDerivedClass { }";
...
+                }
+            };
+
+            csharpTest.ExpectedDiagnostics.AddRange(expected);
+
+            await csharpTest.RunAsync();
         }
 
-        private static DiagnosticResult GetCA1509ExpectedDiagnostic(int line, int column, string entry, string additionalFile)
+        private async Task VerifyBasic(string source, string codeMetricsConfigSource, params DiagnosticResult[] expected)
```
sharwell:
```suggestion
        private async Task VerifyBasicAsync(string source, string codeMetricsConfigSource, params DiagnosticResult[] expected)
```

Evangelink:
The merge is going to be painful v.v

sharwell:
@Evangelink We will take care of the merge

Evangelink:
@sharwell is it ok? I have been flooding you with reviews and questions lately...

sharwell:
@Evangelink Yep, Manish is familiar with the new layout so we can get the conflicts addressed more efficiently.

mavasani:
Note that for the test project, there is only a folder move that was done - all the test file names, namespaces and contents were unchanged. So, ideally just moving the test files should suffice. I will attempt the merge now.

mavasani:
@Evangelink The merge is complete - there are some test failures that need to be investigated.

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/DisposeMethodsShouldCallBaseClassDisposeTests.cs
Evangelink: I am not responsible for the change, it seems to be coming from some merge.

src/Microsoft.NetCore.Analyzers/Core/Runtime/DisposeMethodsShouldCallBaseClassDispose.cs
mavasani: Oh, that's even worse :-). I am fine with this change and we see any fallouts, we can handle it subsequently.

Evangelink:
That was a BIG chunk of work! Thank you both for your time!

sharwell:
🎉 Thank you @Evangelink !

mavasani:
Yes, a big thank you to @Evangelink and @sharwell to push this through!

PR #3055, opened 2019-11-19
Rule CA2011: Avoid infinite recursion

### LINKED ISSUE #777: C#: infinite recursive setter should be warned at compile time
Ported from https://github.com/dotnet/roslyn/issues/7695 - reported by @zhangtairan

---

For example:

```
    class Test
    {
         public int Abc
         {
             set
             {
                 this.Abc = value;
             }
         }
    }
    Test test = new Test();
    test.Abc = 1;
```

This behavior will cause StackOverflowException, why not disable infinite recursive setter by the compiler?
Thanks!

Evangelink:
Add a new rule to detect infinite recursive setter.

Fix #777

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
@@ -1343,10 +1343,13 @@
   <data name="AssigningSymbolAndItsMemberInSameStatementTitle" xml:space="preserve">
     <value>Assigning symbol and its member in the same statement.</value>
   </data>
-  <data name="AvoidInfiniteRecursionMessage" xml:space="preserve">
-    <value>This call is leading to an infinite recursion.</value>
+  <data name="AvoidInfiniteRecursionMessageSure" xml:space="preserve">
+    <value>Do not assign the property within its setter. This call will result in an infinite recursion.</value>
   </data>
   <data name="AvoidInfiniteRecursionTitle" xml:space="preserve">
     <value>Avoid infinite recursion</value>
+  </data>
+  <data name="AvoidInfiniteRecursionMessageMaybe" xml:space="preserve">
+    <value>Do not assign the property within its setter. This call might result in an infinite recursion.</value>
   </data>
 </root>
--- a/src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
@@ -1343,4 +1343,10 @@
   <data name="AssigningSymbolAndItsMemberInSameStatementTitle" xml:space="preserve">
     <value>Assigning symbol and its member in the same statement.</value>
   </data>
+  <data name="AvoidInfiniteRecursionMessage" xml:space="preserve">
+    <value>This call is leading to an infinite recursion.</value>
+  </data>
+  <data name="AvoidInfiniteRecursionTitle" xml:space="preserve">
+    <value>Avoid infinite recursion</value>
+  </data>
 </root>
--- a/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Immutable;
 using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
@@ -65,10 +66,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                         ancestor.Kind != OperationKind.LocalFunction &&
                         ancestor.Kind != OperationKind.Conditional);
 
-                    operationContext.ReportDiagnostic(
-                        Diagnostic.Create(
-                            ancestor != null ? MaybeRule : Rule,
-                            assignmentOperation.Syntax.GetLocation(), operationTarget.Property.Name));
+                    assignmentOperation.CreateDiagnostic(ancestor != null ? MaybeRule : Rule, operationTarget.Property.Name);
                 }, OperationKind.SimpleAssignment);
             });
         }
--- a/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Immutable;
 using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
@@ -38,6 +39,10 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                         var assignmentOperation = (IAssignmentOperation)operationContext.Operation;
 
                         if (assignmentOperation.Target is IPropertyReferenceOperation operationTarget &&
+                            operationTarget.Instance is IInstanceReferenceOperation targetInstanceReference &&
+                            targetInstanceReference.ReferenceKind == InstanceReferenceKind.ContainingTypeInstance &&
+                            assignmentOperation.GetAncestor<ILocalFunctionOperation>(OperationKind.LocalFunction) == null &&
+                            assignmentOperation.GetAncestor<IAnonymousFunctionOperation>(OperationKind.AnonymousFunction) == null &&
                             operationTarget.Member.Equals(methodSymbol.AssociatedSymbol))
                         {
                             operationContext.ReportDiagnostic(Diagnostic.Create(Rule, assignmentOperation.Syntax.GetLocation(), operationTarget.Property.Name));
--- a/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AvoidInfiniteRecursion : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2008";
+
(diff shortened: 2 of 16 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1343,4 +1343,10 @@
   <data name="AssigningSymbolAndItsMemberInSameStatementTitle" xml:space="preserve">
     <value>Assigning symbol and its member in the same statement.</value>
   </data>
+  <data name="AvoidInfiniteRecursionMessage" xml:space="preserve">
+    <value>This call is leading to an infinite recursion.</value>
```
Evangelink: Shall I keep a "generic" message in prevision for the next implementation or go for a specific message for the time being?

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AvoidInfiniteRecursion : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2008";
```
Evangelink: This seems to be the first `Reliability` available ID. Please review.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
```
Evangelink: I am not sure whether this new analyzer should be part of the `QualityGuidelines` folder but I decided to put it next to `CA2245`.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,50 @@
...
+                    methodSymbol.MethodKind == MethodKind.PropertySet)
+                {
+                    operationBlockStartContext.RegisterOperationAction(operationContext =>
+                    {
+                        var assignmentOperation = (IAssignmentOperation)operationContext.Operation;
+
+                        if (assignmentOperation.Target is IPropertyReferenceOperation operationTarget &&
+                            operationTarget.Member.Equals(methodSymbol.AssociatedSymbol))
+                        {
+                            operationContext.ReportDiagnostic(Diagnostic.Create(Rule, assignmentOperation.Syntax.GetLocation(), operationTarget.Property.Name));
```
mavasani: Implementation looks good to me, but on second thought I am wondering if you need to handle cases where the assignment happens inside a lambda or local function within the setter. That does seem like a contrived case, but we can choose to be more defensive by making sure that we have an additional check here that none of the ancestors of this assignment operation are `IAnonymousFunctionOperation` or `ILocalFunctionOperation`. I would let you make a call on this.

src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: I think we need a better message here, something like `Do not set the property within the property setter`? @dotnet/roslyn-analysis for any suggestions.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
Evangelink: Good point! I tend to think that's better to exclude the cases inside lambdas as we won't precisely know how it will be used.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                if (operationBlockStartContext.OwningSymbol is IMethodSymbol methodSymbol &&
+                    methodSymbol.MethodKind == MethodKind.PropertySet)
+                {
+                    operationBlockStartContext.RegisterOperationAction(operationContext =>
+                    {
+                        var assignmentOperation = (IAssignmentOperation)operationContext.Operation;
+
+                        if (assignmentOperation.Target is IPropertyReferenceOperation operationTarget &&
+                            operationTarget.Member.Equals(methodSymbol.AssociatedSymbol))
```
mavasani: I believe you need to also check that the instance receiver is 'this' or 'Me'. Otherwise, you will generate a false positive for below case:

```csharp
public class A
{
    private A _field;

    public int Abc
    {
        set
        {
            _field.Abc = value;
        }
    }
}
```

mavasani: Please add the above unit test as well.

src/Microsoft.CodeQuality.Analyzers/UnitTests/QualityGuidelines/AvoidInfiniteRecursionTests.cs
```diff
@@ -0,0 +1,153 @@
...
+
+namespace Microsoft.CodeQuality.Analyzers.UnitTests.QualityGuidelines
+{
+    public class AvoidInfiniteRecursionTests
+    {
+        [Fact]
+        public async Task PropertySetterRecursion_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public class A
```
mavasani:
```csharp
public class A
{
    private bool _flag;

    public int Abc
    {
        set
        {
            if (_flag)
            {
                 _flag = false;
                 this.Abc = value;
            }
        }
    }
}
```

This looks like possible false positive as user is guarding the infinite recursion explicitly with additional logic. Not flagging this case could lead to bunch of false negatives, so I am in two minds here. I am wondering if we can generate a separate and less assertive diagnostic message when the setter is parented by some conditional code OR always use a less assertive message for this diagnostic as we never know what crazy kind of code people write and guard infinite recursion. @sharwell thoughts?

Evangelink: I like the idea of 2 "levels" of message. Although I am wondering if the conditional cases would be less important even if possibly less frequent.

Evangelink: @mavasani In case we go for the 2 level of messages, I think it could be nice to start raising on lambda and local functions with this "uncertain" message, WDYT?

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,76 @@
...
+                    do
+                    {
+                        ancestor = ancestor.Parent;
+                    } while (ancestor != null &&
+                        ancestor.Kind != OperationKind.AnonymousFunction &&
+                        ancestor.Kind != OperationKind.LocalFunction &&
+                        ancestor.Kind != OperationKind.Conditional);
+
+                    operationContext.ReportDiagnostic(
+                        Diagnostic.Create(
```
mavasani: `assignmentOperation.CreateDiagnostic(`

mavasani:
Will wait for additional review as this is a new rule - @sharwell or @genlu?

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
mavasani: CA2008 is already in use: https://github.com/dotnet/roslyn-analyzers/blob/7ad9f725f22c7f5a9e8f70a5ebe6e5fd4c14d8d6/src/Microsoft.NetCore.Analyzers/Core/Tasks/DoNotCreateTasksWithoutPassingATaskScheduler.cs#L19

mavasani: You can look at https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md for IDs already in use. Seems `CA2011` is the next available one.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        internal const string RuleId = "CA2011";
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(RuleId,
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidInfiniteRecursionTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidInfiniteRecursionMessageSure), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+            DiagnosticCategory.Reliability,
+            DiagnosticHelpers.DefaultDiagnosticSeverity,
+            isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
+            helpLinkUri: null);
```
Evangelink: @mavasani I'm not sure whether I should have added any custom tag?

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        internal static DiagnosticDescriptor MaybeRule = new DiagnosticDescriptor(RuleId,
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidInfiniteRecursionTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidInfiniteRecursionMessageMaybe), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+            DiagnosticCategory.Reliability,
+            DiagnosticHelpers.DefaultDiagnosticSeverity,
+            isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
+            helpLinkUri: null);
```
Evangelink: Same question about custom tag here.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/AvoidInfiniteRecursion.cs
mavasani: Please use telemetry tag like other analyzers in the repo.

PR #3057, opened 2019-11-20
Convert NetCore Security tests to the new test framework

Evangelink:
This PR is a split from #3052 as security related code should not be merged into master for the time being.


src/Test.Utilities/CSharpProjectState.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Testing;
+
+namespace Microsoft.NetCore.Analyzers.Security.UnitTests
+{
+    /// <summary>
+    /// This class is a copy-paste from
+    /// https://github.com/dotnet/roslyn-sdk/blob/master/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.Analyzer.Testing/CSharpProjectState.cs
+    /// </summary>
+    public class CSharpProjectState : ProjectState
```
Evangelink: The `ProjectState` class is not available. It seems that the dll does contain the class and that the class is `public` so I am not really sure what's missing. Could it be a different analyzer version compared to master?

src/Test.Utilities/VisualBasicProjectState.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Testing;
+
+namespace Microsoft.NetCore.Analyzers.Security.UnitTests
+{
+    /// <summary>
+    /// This class is a copy-paste from
+    /// https://github.com/dotnet/roslyn-sdk/blob/master/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing/VisualBasicProjectState.vb
+    /// </summary>
+    public class VisualBasicProjectState : ProjectState
```
Evangelink: Same issues as for `CSharpProjectState`.

src/Test.Utilities/CSharpProjectState.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Testing;
+
+namespace Microsoft.NetCore.Analyzers.Security.UnitTests
+{
+    /// <summary>
+    /// This class is a copy-paste from
+    /// https://github.com/dotnet/roslyn-sdk/blob/master/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.Analyzer.Testing/CSharpProjectState.cs
```
Evangelink: As explained in this comment this class should not really exists as it is purely a copy-paste from `Roslyn`. I guess that we need to update the testing dependency but I am not sure about the update politic for the branch. Could you please let me know if I can update the version? And also up to which version? Feel free to push a commit on this PR if you want.

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -20,14 +20,14 @@ public static class AdditionalMetadataReferences
         internal static readonly MetadataReference WorkspacesReference = MetadataReference.CreateFromFile(typeof(Workspace).Assembly.Location);
         internal static readonly MetadataReference SystemDiagnosticsDebugReference = MetadataReference.CreateFromFile(typeof(Debug).Assembly.Location);
         internal static readonly MetadataReference SystemDataReference = MetadataReference.CreateFromFile(typeof(System.Data.DataSet).Assembly.Location);
-        internal static readonly MetadataReference SystemWebReference = MetadataReference.CreateFromFile(typeof(System.Web.HttpRequest).Assembly.Location);
-        internal static readonly MetadataReference SystemRuntimeSerialization = MetadataReference.CreateFromFile(typeof(System.Runtime.Serialization.NetDataContractSerializer).Assembly.Location);
+        public static readonly MetadataReference SystemWebReference = MetadataReference.CreateFromFile(typeof(System.Web.HttpRequest).Assembly.Location);
```
Evangelink: Shall I reorder the items to have the `public` ones first?

src/Test.Utilities/VisualBasicProjectState.cs
Evangelink: No longer valid issue.

src/Test.Utilities/CSharpProjectState.cs
Evangelink: No longer valid issue.

Evangelink:
The failing tests have the following error message:
`error AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.DoNotSetSwitch' threw an exception of type 'System.InvalidOperationException' with message 'Feature 'Flow-Analysis' is disabled.'.`

I am not sure whether I should always add the info in the additional file or whether something shall be fixed in the code.

mavasani:
2.9.x branch works against Microsoft.CodeAnalysis 2.9.0, which needs the project to have the `flow-analysis` feature flag: https://github.com/dotnet/roslyn-analyzers/blob/115ded2bd8c61a3b51de5c2f657e1ece26e865d1/src/Test.Utilities/DiagnosticAnalyzerTestBase.cs#L453-L457

dotpaul:
Looking at https://github.com/dotnet/roslyn-sdk/blob/66b2eb24a0e1e8fe31ff777a8255ceecbcbddb51/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest%601.cs#L722, maybe you could modify https://github.com/dotnet/roslyn-analyzers/blob/2.9.x/src/Test.Utilities/CSharpCodeFixVerifier%602%2BTest.cs to override the `ApplyCompilationOptions` method, which could then call the base method and then apply `flow-analysis` feature flag in the same way that @mavasani pointed out. Same with the VB version.

Evangelink:
Thank you for the input! I have pushed a change that is not working locally to see if it passes on the CI. For some reason, I am now having lots of test failing because of a debug assertion related to objects being free twice.

dotpaul:
Dunno if this is it, but try bumping up MicrosoftCodeAnalysisTestingVersion in eng/Versions.prop to be the same as it is in the master branch.

Evangelink:
@dotpaul That's working but now I have to do quite a lot of changes in the various test utilities classes (copy-paste from master) to solve the issue. I will keep you posted with the changes.

BTW is it ok to simply do some copy-pasting or shall I cherry-pick multiple commits?

dotpaul:
I think copy and pasting is fine.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseDeprecatedSecurityProtocolsTests.cs
```diff
@@ -435,14 +435,14 @@ public void TestMethod()
...
 }",
-                GetCSharpResultAt(9, 48, DoNotUseDeprecatedSecurityProtocols.HardCodedRule, "Tls12"),
-                GetCSharpResultAt(9, 48, DoNotUseDeprecatedSecurityProtocols.DeprecatedRule, "3264"));
+                VerifyCS.Diagnostic(DoNotUseDeprecatedSecurityProtocols.DeprecatedRule).WithSpan(9, 48, 9, 102).WithArguments("3264"),
+                VerifyCS.Diagnostic(DoNotUseDeprecatedSecurityProtocols.HardCodedRule).WithSpan(9, 48, 9, 74).WithArguments("Tls12"));
         }
 
         [Fact]
-        public void TestUse768DeconstructionAssignmentNoDiagnostic()
+        public async Task TestUse768DeconstructionAssignmentNoDiagnostic()
```
Evangelink: Test is failing with the following error

```// Test0.cs(10,53): error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
```

I am pretty sure that I screwed up something in the `Test` class definition mixing netfwk and netstandard or something like that but I cannot really spot where is the mistake.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
```diff
@@ -478,14 +479,14 @@ public void Method()
...
 }",
-                GetCSharpResultAt(9, 34, SslProtocolsAnalyzer.HardcodedRule, "Tls12"),
-                GetCSharpResultAt(9, 34, SslProtocolsAnalyzer.DeprecatedRule, "3264"));
+                VerifyCS.Diagnostic(SslProtocolsAnalyzer.DeprecatedRule).WithSpan(9, 34, 9, 52).WithArguments("3264"),
+                VerifyCS.Diagnostic(SslProtocolsAnalyzer.HardcodedRule).WithSpan(9, 34, 9, 52).WithArguments("Tls12"));
         }
 
         [Fact]
-        public void Use768DeconstructionAssignment_NoDiagnostic()
+        public async Task Use768DeconstructionAssignment_NoDiagnostic()
```
Evangelink: Same error with the `ValueTuple`

```
// Test0.cs(12,34): error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
```

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
```diff
@@ -478,14 +479,14 @@ public void Method()
         SslProtocols protocols = SslProtocols.Tls12 | (SslProtocols)192;
     }
 }",
-                GetCSharpResultAt(9, 34, SslProtocolsAnalyzer.HardcodedRule, "Tls12"),
-                GetCSharpResultAt(9, 34, SslProtocolsAnalyzer.DeprecatedRule, "3264"));
+                VerifyCS.Diagnostic(SslProtocolsAnalyzer.DeprecatedRule).WithSpan(9, 34, 9, 52).WithArguments("3264"),
```
Evangelink: Test is failing saying the other expected diagnostic appears first. When I change the order, it's then the other test appearing first.

src/MetaCompilation.Analyzers/UnitTests/MetaCompilation.Analyzers.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
```
Evangelink: Version is copied from master.

src/Test.Utilities/VisualBasicCodeFixVerifier`2+Test.cs
```diff
@@ -17,32 +16,10 @@ public class Test : VisualBasicCodeFixTest<TAnalyzer, TCodeFix, XUnitVerifier>
         {
             public Test()
             {
+                ReferenceAssemblies = AdditionalMetadataReferences.Default;
```
Evangelink: Copy-pasted from master

src/Test.Utilities/MinimalImplementations/NewtonsoftJsonNetApis.cs
```diff
@@ -5,11 +5,36 @@ namespace Test.Utilities.MinimalImplementations
     public static class NewtonsoftJsonNetApis
     {
         public const string CSharp = @"
+using System;
```
Evangelink: The minimal code wasn't valid for C# nor VB.NET, I fixed it.

src/Test.Utilities/CSharpSecurityCodeFixVerifier`2+Test.cs
```diff
@@ -17,6 +20,17 @@ public Test()
                 // These analyzers run on generated code by default.
                 TestBehaviors |= TestBehaviors.SkipGeneratedCodeCheck;
             }
+
+            protected override Project ApplyCompilationOptions(Project project)
+            {
+                var newProject = base.ApplyCompilationOptions(project);
+
+                var parseOptions = newProject.ParseOptions.WithFeatures(
```
Evangelink: As discussed, I have applied this change for all security tests through this helper.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseDeprecatedSecurityProtocolsTests.cs
dotpaul: My wild guess is `AdditionalMetadataReferences.Default` including the Microsoft.CodeAnalysis package. One work around is having a separate set of ReferenceAssemblies for the Security*CodeFixVerifiers that doesn't include the Microsoft.CodeAnalysis package.

Maybe someone has a better idea? 🙂

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
dotpaul: @sharwell ?

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseDeprecatedSecurityProtocolsTests.cs
Evangelink: It solves the issue but now all the tests using `Microsoft.CodeAnalysis` fail. I will go this way and update locally waiting for a better option.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
Evangelink: Same issue/fix as the previous comment.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseDeprecatedSecurityProtocolsTests.cs
Evangelink: On the second though that doesn't seem right to go this way. I'll wait for someone input here. @sharwell, @mavasani any idea?

src/Test.Utilities/CSharpSecurityCodeFixVerifier`2+Test.cs
mavasani: This method override should be moved to the base type `CSharpCodeFixVerifier<TAnalyzer, TCodeFix>.Test`. The feature flag is needed for all tests in 2.9.x branch, not just security tests.

src/Test.Utilities/DiagnosticAnalyzerTestBase.cs
```diff
@@ -430,21 +433,17 @@ protected static Project CreateProject(string[] sources, string language = Langu
...
                 .AddProject(projectId, projectName, projectName, language)
-                .AddMetadataReference(projectId, MetadataReferences.CorlibReference)
-                .AddMetadataReference(projectId, MetadataReferences.SystemCoreReference)
-                .AddMetadataReference(projectId, AdditionalMetadataReferences.SystemXmlReference)
-                .AddMetadataReference(projectId, MetadataReferences.CodeAnalysisReference)
-                .AddMetadataReference(projectId, AdditionalMetadataReferences.SystemRuntimeFacadeRef)
-                .AddMetadataReference(projectId, AdditionalMetadataReferences.SystemThreadingFacadeRef)
-                .AddMetadataReference(projectId, AdditionalMetadataReferences.SystemThreadingTaskFacadeRef)
+                .AddMetadataReferences(projectId, references)
+                .AddMetadataReference(projectId, AdditionalMetadataReferences.CodeAnalysisReference)
```
mavasani: Why is `AdditionalMetadataReferences.CodeAnalysisReference` added by default here? We already have `ReferenceFlags.RemoveCodeAnalysis` check below and this Add should potentially move inside that if statement. It is also not clear to me what is `AdditionalMetadataReferences.CSharpSymbolsReference`? Can you clarify which assembly reference that is pulling in?

src/Test.Utilities/VisualBasicSecurityCodeFixVerifier`2+Test.cs
```diff
@@ -17,6 +20,17 @@ public Test()
                 // These analyzers run on generated code by default.
                 TestBehaviors |= TestBehaviors.SkipGeneratedCodeCheck;
             }
+
+            protected override Project ApplyCompilationOptions(Project project)
+            {
+                var newProject = base.ApplyCompilationOptions(project);
+
+                var parseOptions = newProject.ParseOptions.WithFeatures(
+                    newProject.ParseOptions.Features.Concat(
```
mavasani: Same comment as C# case - this feature flag addition should move to the base type.

src/Microsoft.NetCore.Analyzers/UnitTests/Data/ReviewSQLQueriesForSecurityVulnerabilitiesTests_FlowAnalysis.cs
```diff
@@ -1,17 +1,35 @@
...
+using VerifyVB = Test.Utilities.VisualBasicSecurityCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Data.ReviewSqlQueriesForSecurityVulnerabilities,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
 
 namespace Microsoft.NetCore.Analyzers.Data.UnitTests
 {
     [Trait(Traits.DataflowAnalysis, Traits.Dataflow.ValueContentAnalysis)]
     public class ReviewSQLQueriesForSecurityVulnerabilitiesTests_FlowAnalysis : ReviewSQLQueriesForSecurityVulnerabilitiesTests
     {
+        private static DiagnosticResult GetCSharpResultAt(int line, int column, string invokedSymbol, string containingMethod)
```
mavasani: I am wondering if these should be helper methods which take line, column and a params string array so each test file doesn't have to define its own version. Not sure how often you had to add these to test files.

Evangelink: I think @sharwell's idea is to get rid of all those helper methods and to use inline `VerifyCS.Diagnostic().WithLocation()...` but I am not sure.

To answer your question all analyzer tests have something similar, for some I have to specify the actual `DiagnosticDescriptor`, for some there is no arguments.

src/Test.Utilities/VisualBasicSecurityCodeFixVerifier`2+Test.cs
Evangelink: Will do.

src/Test.Utilities/CSharpSecurityCodeFixVerifier`2+Test.cs
Evangelink: Will do.

src/Test.Utilities/DiagnosticAnalyzerTestBase.cs
Evangelink: I'll have a look but if I recall correctly I simply copy-pasted the content of the class from master so I am afraid I won't be able to explain much.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/ApprovedCipherModeTests.cs
```diff
@@ -148,14 +154,14 @@ End Sub
             );
         }
 
-        protected override DiagnosticAnalyzer GetBasicDiagnosticAnalyzer()
-        {
-            return new ApprovedCipherModeAnalyzer();
-        }
+        private static DiagnosticResult GetCSharpResultAt(int line, int column, params string[] arguments)
```
mavasani: Yeah seems like you are defining this in almost all the test files. Might make sense to have these as helper methods, unless they are causing a lot of churn on your side - if so it is fine to leave it as a follow-up cleanup item.

mavasani: Probably good to have these as internal static helper methods on VerifyCS and VerifyVB?

src/Test.Utilities/DiagnosticAnalyzerTestBase.cs
Evangelink: @mavasani I confirm that I simply copy-pasted the full method from the class in master when the code has started to fail after the bump of version for the test framework.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseDeprecatedSecurityProtocolsTests.cs
mavasani: Yeah, seems like a bit of a hack to me. Given that this is a large PR and also blocks the corresponding PR in master, I think it is reasonable to create tracking issues for any workarounds that can be put to unblock this PR. Lets get this PR merged soon and try to tackle the correct fix with a follow-up PR.

src/Test.Utilities/DiagnosticAnalyzerTestBase.cs
mavasani: Sounds good. Might be good to create a separate tracking issue where you can just point to all the workarounds and unresolved comments in this PR so we can move forward with this PR, while not losing track of these issues.

src/Microsoft.NetCore.Analyzers/UnitTests/Data/ReviewSQLQueriesForSecurityVulnerabilitiesTests_FlowAnalysis.cs
mavasani: Sounds good - you don't need to address this comment in this PR. We can discuss in the separate tracking issue if it is worth just having a single helper in VerifyCS/VerifyVB versus polluting every test file.

Evangelink: https://github.com/dotnet/roslyn-analyzers/issues/3088

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
Evangelink: @mavasani I could keep the base class and the old test verification for this specific test and create a ticket for this issue. Otherwise I will need some help for this case.

mavasani: Sounds good. Anything that helps you move forward with the PR works for me. Thanks!

src/Microsoft.NetCore.Analyzers/UnitTests/Security/ApprovedCipherModeTests.cs
Evangelink: See #3088

src/Test.Utilities/DiagnosticAnalyzerTestBase.cs
Evangelink: See #3089

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
Evangelink: See #3091

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseDeprecatedSecurityProtocolsTests.cs
Evangelink: See #3092

src/Microsoft.NetCore.Analyzers/UnitTests/Security/SslProtocolsAnalyzerTests.cs
Evangelink: See #3092

sharwell:
@Evangelink I've been trying to reach you on Gitter. Are you on there? 😄

eng/Versions.props
```diff
@@ -28,7 +28,7 @@
     <MicrosoftCodeAnalysisTestResourcesProprietaryVersion>2.0.0-pre-20160714</MicrosoftCodeAnalysisTestResourcesProprietaryVersion>
 
     <!-- Roslyn Testing -->
-    <MicrosoftCodeAnalysisTestingVersion>1.0.0-beta1.19264.1</MicrosoftCodeAnalysisTestingVersion>
+    <MicrosoftCodeAnalysisTestingVersion>1.0.1-beta1.19557.2</MicrosoftCodeAnalysisTestingVersion>
```
sharwell: 📝 I'm going to port this change to the 2.9.x branch separately.

sharwell: ➡️ #3093

src/MetaCompilation.Analyzers/UnitTests/MetaCompilation.Analyzers.UnitTests.csproj
sharwell: After #3093 is merged, you can revert all the target framework changes in this pull request (they are not necessary since I updated the Microsoft.CodeAnalysis reference to target Roslyn 2.10.0 in this branch).

Evangelink: I did the merge and when I revert the projects to `net46` I have a bunch of tests failing because they rely on APIs available only starting from `net47`. I am not sure whether that's best to stick with `net46` and update the failing tests to reference `net47` assemblies or to change/keep `net47` for test projects.

PR #3058, opened 2019-11-21
Performance fix for AnalysisEntityMapAbstractDomain.Merge algorithm

mavasani:
We now ensure that we only create new entries in the result map if the entity key is a valid tracked entity


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/TrackedEntitiesBuilder.cs
```diff
@@ -10,16 +12,29 @@ namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PointsToAnalysis
...
         public void Dispose()
         {
-            AllEntities.Free();
+            _allEntities.Free();
         }
+
+        public void AddEntity(AnalysisEntity analysisEntity)
+            => _allEntities.Add(analysisEntity);
+
+        public IEnumerable<AnalysisEntity> EnumerateEntities()
```
dotpaul: >EnumerateEntities [](start = 43, length = 17)

Nit: Could make TrackedEntitiesBuilder implement IEnumerable<AnalysisEntity>.

mavasani: This is no longer applicable as now this type tracks both entities and points to values.

mavasani:
@dotpaul I was able to validate no performance regression in building Roslyn after this fix, let me know once your validations are done. Thanks!

dotpaul:
Other than the one case @LLLXXXCCC emailed about, the previous repros looks good to me! I'm trying a pass on a larger code base; will approve after that finishes.

mavasani:
Thank you @dotpaul!

PR #3074, opened 2019-11-27
Update CA1305 to report on TryParse methods

### LINKED ISSUE #1848: Let CA1305 warn on TryParse
#### Analyzer package

#### Package Version

v2.6.2

#### Diagnostic ID

[CA1305](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1305-specify-iformatprovider)

#### Repro steps

Introduce a call to p.e. [`float.TryParse(string, out float)`](https://docs.microsoft.com/en-us/dotnet/api/system.single.tryparse)

#### Expected behavior
I see a warning:
The behavior of 'float.TryParse(string,float)' could vary based on the current user's locale settings. Replace this call in 'Program.Main(string[])' with a call to 'float.TryParse(string,NumberStyles,IFormatProvider,float)'.

#### Actual behavior

I don't see a warning.

#### Rationale

`float.Parse(string)` and `float.ToString()` already generate CA1305. The rule should help the developer to not always think about that issue. However, if a developer is relying upon the analyzer, she might be bugged as soon as she uses `TryParse` method.

Evangelink:
Update the behavior of CA1305 to allow reporting on all `TryParse` methods when the related option is activated.

Fix #1848

### DIFF
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -409,3 +409,14 @@ Option Values:
 Default Value: `Heuristic`.
 
 Example: `dotnet_code_quality.CA1712.enum_values_prefix_trigger = AnyEnumValue`
+
+### Specify IFormatProvider
+Option Name: `exclude_tryparse_methods`
+
+Configurable Rules: [CA1305](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1305)
+
+Option Values: `true` or `false`
+
+Default Value: `true`
+
+Example: `dotnet_code_quality.CA1305.exclude_tryparse_methods = false`
--- a/src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -123,6 +124,9 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 var installedUICulturePropertyOfComputerInfoType = computerInfoType?.GetMembers("InstalledUICulture").OfType<IPropertySymbol>().FirstOrDefault();
 
                 var obsoleteAttributeType = csaContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemObsoleteAttribute);
+
+                var excludeTryParseMethods = csaContext.Options.GetBoolOptionValue(EditorConfigOptionNames.ExcludeTryParseMethods, IFormatProviderAlternateRule,
+                    defaultValue: true, cancellationToken: csaContext.CancellationToken);
                 #endregion
 
                 csaContext.RegisterOperationAction(oaContext =>
@@ -175,6 +179,13 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                         // If there are two matching overloads, one with CultureInfo as the first parameter and one with CultureInfo as the last parameter,
                         // report the diagnostic on the overload with CultureInfo as the last parameter, to match the behavior of FxCop.
                         var correctOverload = correctOverloads.FirstOrDefault(overload => overload.Parameters.Last().Type.Equals(iformatProviderType)) ?? correctOverloads.FirstOrDefault();
+
+                        // Use a different mechanism for TryParse methods as they usually have more than 1 extra parameter and
+                        // the IFormatProvider parameter is neither first nor last.
+                        if (correctOverload == null && targetMethod.Name.Equals("TryParse", StringComparison.Ordinal) && !excludeTryParseMethods)
+                        {
+                            correctOverload = methodsWithSameNameAsTargetMethod.FirstOrDefault(x => HasCorrectTryMethodParameters(x, targetMethod));
+                        }
 
                         // Sample message for IFormatProviderAlternateRule: Because the behavior of Convert.ToInt64(string) could vary based on the current user's locale settings,
                         // replace this call in IFormatProviderStringTest.TestMethod() with a call to Convert.ToInt64(string, IFormatProvider).
@@ -245,6 +256,35 @@ namespace Microsoft.NetCore.Analyzers.Runtime
         private static ParameterInfo GetParameterInfo(INamedTypeSymbol type, bool isArray = false, int arrayRank = 0, bool isParams = false)
         {
             return ParameterInfo.GetParameterInfo(type, isArray, arrayRank, isParams);
+        }
+
+        private static bool HasCorrectTryMethodParameters(IMethodSymbol candidateMethod, IMethodSymbol referenceMethod)
+        {
+            if (candidateMethod.Parameters.Length <= referenceMethod.Parameters.Length)
+            {
+                return false;
+            }
+
+            var remainingCandidateMethodParameters = candidateMethod.Parameters.ToList();
+            for (int i = 0; i < referenceMethod.Parameters.Length; i++)
+            {
+                for (int j = 0; j < remainingCandidateMethodParameters.Count; j++)
+                {
+                    if (remainingCandidateMethodParameters[j].Type.Equals(referenceMethod.Parameters[i].Type))
+                    {
+                        remainingCandidateMethodParameters.RemoveAt(i);
+                        continue;
+                    }
+
+                    if (j == remainingCandidateMethodParameters.Count)
+                    {
+                        // Cannot find any parameter of the expected type.
+                        return false;
+                    }
+                }
+            }
+
+            return true;
         }
     }
 }
--- a/src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/SpecifyIFormatProviderTests.cs
+++ b/src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/SpecifyIFormatProviderTests.cs
@@ -915,6 +915,240 @@ End Class
 ");
         }
 
+        [Fact]
+        public async Task CA1305_Parse_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public class Foo
+{
+    public Foo()
+    {
+        sbyte.Parse("""");
+        byte.Parse("""");
+        short.Parse("""");
+        ushort.Parse("""");
+        int.Parse("""");
+        uint.Parse("""");
+        long.Parse("""");
+        ulong.Parse("""");
+
+        char.Parse(""""); // no issue because no overload with format provider
+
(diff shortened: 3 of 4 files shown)


docs/Analyzer Configuration.md
```diff
@@ -408,4 +408,15 @@ Option Values:
 
 Default Value: `Heuristic`.
 
-Example: `dotnet_code_quality.CA1712.enum_values_prefix_trigger = AnyEnumValue`
\ No newline at end of file
+Example: `dotnet_code_quality.CA1712.enum_values_prefix_trigger = AnyEnumValue`
+
+### Specify IFormatProvider
+Option Name: `exclude_tryparse_methods`
```
Evangelink: All the other boolean options seem to be working on a negative way so I have mimic this behavior. If you prefer another mechanism, please let me know.

src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -176,6 +180,13 @@ public override void Initialize(AnalysisContext analysisContext)
                         // report the diagnostic on the overload with CultureInfo as the last parameter, to match the behavior of FxCop.
                         var correctOverload = correctOverloads.FirstOrDefault(overload => overload.Parameters.Last().Type.Equals(iformatProviderType)) ?? correctOverloads.FirstOrDefault();
 
+                        // Use a different mechanism for TryParse methods as they usually have more than 1 extra parameter and
+                        // the IFormatProvider parameter is neither first nor last.
+                        if (correctOverload == null && targetMethod.Name.Equals("TryParse", StringComparison.Ordinal) && !excludeTryParseMethods)
```
Evangelink: I have tried to always favor the old mechanism way.

src/Microsoft.NetCore.Analyzers/UnitTests/Runtime/SpecifyIFormatProviderTests.cs
```diff
@@ -915,6 +915,240 @@ End Class
 ");
         }
 
+        [Fact]
+        public async Task CA1305_Parse_Diagnostic()
```
Evangelink: Not really related to this change but I couldn't find any test with the parse methods and I wanted to have some comparison to see how should work the `TryParse` variants.

Evangelink:
I'll wait for approval (or simply a go from your side) before creating a ticket to update the doc.

src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -176,6 +180,13 @@ public override void Initialize(AnalysisContext analysisContext)
                         // report the diagnostic on the overload with CultureInfo as the last parameter, to match the behavior of FxCop.
                         var correctOverload = correctOverloads.FirstOrDefault(overload => overload.Parameters.Last().Type.Equals(iformatProviderType)) ?? correctOverloads.FirstOrDefault();
 
+                        // Use a different mechanism for TryParse methods as they usually have more than 1 extra parameter and
```
mavasani: I don't believe `TryParse` methods should be special cased this way, neither in the code, nor in the option. I think instead we should have 2 modes:

1. Current behavior: Flags when invoked method has an overload with one additional `IFormatProvider` parameter as the last parameter and the remainder of parameters are same.
2. New behavior: Flags when invoked method has an overload with one additional `IFormatProvider` parameter as the last non-ref and non-out parameter and the remainder of parameters (including the following ref and out parameters) are the same.

Evangelink: I started my implementation with what you describe in point 2 but I have started to face some issues with some types like numbers, dates... `DateTime.TryParse(string, IFormatProvider, DateTimeStyles, out DateTime)` where there is an extra parameter `XXXStyles`. I was not sure if it would be better to ignore `out` as last and an enum whose name ends in `Styles` as before last or go with my approach. Let me know your preference.

Regarding the option, are you suggesting an enum or to keep a boolean but to have a different name (something like `extended`)?

mavasani: Having put more thought in this, I think the TryParse case above falls into the more generic analysis:

- I have `n` different overloads for a method for similar functionality which were added over a time period.
- I would prefer to have an analyzer that allows me to flag use of non-preferred overloads
- Analyzer would have some default set of BCL APIs where we already know the preferred and non-preferred overloads. In our case, this will be the `TryParse` methods.
- Analyzer would have an option to allow me to specify fully qualified signatures of preferred (and/or non-preferred) overloads for custom APIs.

The above seems to very closely match the [BannedApiAnalyzers](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/BannedApiAnalyzers.Help.md), though that analyzer does not special case banned APIs from BCL libraries and is a much more stricter version of _banned_ as opposed to _non-preferred_ APIs. I am wondering if we should add a new rule to that analyzer with a different ID and much less stricter message, that only deals with preferred and non-preferred symbol pairs, special cases the APIs that we already know about (TryParse group) and allows users to specify an additional customized preferences in `BannedSymbols.txt`. Building on the example from [here](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/BannedApiAnalyzers.Help.md):

```csharp
class MyType
{
   void Method(T1 t);
   void Method(T2 t);  // preferred overload
}
```

Potential "BannedSymbols.txt":
```
MyType.Method(T1);prefer:MyType.Method(T2)
```

@drewnoakes @sharwell for additional comments.

Evangelink: @mavasani I think your suggestion is very nice and would definitely be a valuable new rule. Yet I don't think that the `TryParse` method should differ from `Parse` methods, meaning that if we have the other rule then this one shall be made fully obsolete. Methods with a culture should always be favored against non-culture overloads.

mavasani: That is a reasonable argument from an end user experience perspective. However, I don't like us special casing the method in the code and in the option name. What if some other API tomorrow adds a similar overload with some different additional parameter (not styles) and we run into this same problem again? I do believe we need a more generic solution here, even if it is implemented in this analyzer itself.

Evangelink: I guess my comment wasn't really clear. My point was to say that I am fully aligned with you and I am just thinking that if/when we have the new rule (assuming this is a different one) then this rule no longer really have value and it would be better to make it obsolete. The new rule with preferred API will most likely favor `Parse` overloads with the culture.

Evangelink: @mavasani Shall I create a new issue for the new rule proposal and try to come up with a first implementation?

Evangelink: See https://github.com/dotnet/roslyn-analyzers/issues/3188

I still think that it might be nice to have this PR merged and to deprecate this specific analyzer when the related ticket (the new analyzer) is implemented.

WDYT @mavasani ?

mavasani: I don't believe we should have an option targeting just specific set of methods. Can you update the implementation to not have this option at all to start with? There is a risk of this being a breaking change, but it seems a reasonable one.

src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -246,5 +257,34 @@ private static ParameterInfo GetParameterInfo(INamedTypeSymbol type, bool isArra
...
+        {
+            if (candidateMethod.Parameters.Length <= referenceMethod.Parameters.Length)
+            {
+                return false;
+            }
+
+            var remainingCandidateMethodParameters = candidateMethod.Parameters.ToList();
+            for (int i = 0; i < referenceMethod.Parameters.Length; i++)
+            {
+                for (int j = 0; j < remainingCandidateMethodParameters.Count; j++)
```
mavasani: I think the implementation here needs to be hardened. We need to ensure that `candidateMethod.Parameters.Length == referenceMethod.Parameters.Length + 1`, and the extra parameter of candidateMethod has `IFormatProvider` type and parameter types of both methods, excluding that parameter from candidateMethod, exactly match both in their indices and type.

Evangelink: There is a problem with this implementation for date and times as they have overload with more than 1 parameter being valid.

src/Microsoft.NetCore.Analyzers/Core/Runtime/SpecifyIFormatProvider.cs
Evangelink: Just to be sure, you want me to remove the following option https://github.com/dotnet/roslyn-analyzers/pull/3074/files#diff-1b5cdc4937a8cdfc79545945ad6c0783R414-R422 and to have the tryparse methods allowed by default?

Evangelink:
Hey @mavasani what do you want me to do with this PR? Shall we close it or shall we take some time to define the scope of what we want to do and I make the required changes?

docs/Analyzer Configuration.md
```diff
@@ -408,4 +408,15 @@ Option Values:
 
 Default Value: `Heuristic`.
 
-Example: `dotnet_code_quality.CA1712.enum_values_prefix_trigger = AnyEnumValue`
\ No newline at end of file
+Example: `dotnet_code_quality.CA1712.enum_values_prefix_trigger = AnyEnumValue`
+
+### Specify IFormatProvider
```
Youssef1313, line 413: nit

```suggestion
### Specify IFormatProvider

```

docs/Analyzer Configuration.md
```diff
@@ -408,4 +408,15 @@ Option Values:
...
+Option Name: `exclude_tryparse_methods`
+
+Configurable Rules: [CA1305](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1305)
```
Youssef1313, line 416:
```suggestion
Configurable Rules: [CA1305](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1305)
```

Evangelink:
@mavasani What shall we do with this PR? Drop it or align on a scope and finish up the work.

rjpowers10:
Just commenting to say that I am interested in `TryParse` being included in the analyzer coverage. `decimal.TryParse` was my motivation for enabling CA1305 so I was surprised that `decimal.Parse` was included but `decimal.TryParse` was not.

PR #3075, opened 2019-11-27
Use fully qualified name for types in generated xml metrics

### LINKED ISSUE #2804: Metrics.exe : generated Xml does not distinguish nested types.
Microsoft.CodeAnalysis.Metrics

https://www.nuget.org/packages/Microsoft.CodeAnalysis.Metrics/

v2.9.4 (Latest)

When reviewing the generated XML for my project, for which several classes have a nested private structure(all with the same name), the nested types are not qualified in any way.

This makes it difficult to determine which nested class belongs to which parent. 

Further, if there is a name clash between outer/inner class names, this becomes much more difficult to resolve.

Would it be possible to make the Name attribute include the parent name e.g. ParentName.InnerName?

Evangelink:
The class doesn't have any test so there is no test to update.

Fix #2804

### DIFF
--- a/src/Tools/Metrics/MetricsOutputWriter.cs
+++ b/src/Tools/Metrics/MetricsOutputWriter.cs
@@ -4,6 +4,7 @@ using System.Collections.Immutable;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Xml;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeMetrics;
@@ -61,7 +62,16 @@ namespace Metrics
                 switch (data.Symbol.Kind)
                 {
                     case SymbolKind.NamedType:
-                        writer.WriteAttributeString("Name", data.Symbol.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat));
+                        var minimalTypeName = new StringBuilder(data.Symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat));
+
+                        var containingType = data.Symbol.ContainingType;
+                        while (containingType != null)
+                        {
+                            minimalTypeName.Insert(0, ".");
+                            minimalTypeName.Insert(0, containingType.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat));
+                        }
+
+                        writer.WriteAttributeString("Name", minimalTypeName.ToString());
                         break;
 
                     case SymbolKind.Method:
--- a/src/Tools/Metrics/MetricsOutputWriter.cs
+++ b/src/Tools/Metrics/MetricsOutputWriter.cs
@@ -61,7 +61,7 @@ namespace Metrics
                 switch (data.Symbol.Kind)
                 {
                     case SymbolKind.NamedType:
-                        writer.WriteAttributeString("Name", data.Symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat));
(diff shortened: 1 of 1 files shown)


src/Tools/Metrics/MetricsOutputWriter.cs
```diff
@@ -61,7 +61,7 @@ void writeHeader()
                 switch (data.Symbol.Kind)
                 {
                     case SymbolKind.NamedType:
-                        writer.WriteAttributeString("Name", data.Symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat));
```
mavasani: I don’t think we want to use fully qualified names for all type names. Also this format adds global:: prefix. Can we just qualify with parent type name for nested types?

Evangelink: Ok! Yet I am not sure if it's not better to use the fully qualified name for nested types rather than concatenating with parent name. In case of multiple level of nesting, is it fine to simply display current level + parent? It seems to still be confusing so we will need to repeat the concat for all containing types.

WDYT?

src/Tools/Metrics/MetricsOutputWriter.cs
```diff
@@ -61,7 +62,16 @@ void writeHeader()
...
+                        var minimalTypeName = new StringBuilder(data.Symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat));
```
Evangelink: I am not sure whether that's best to use a `StringBuilder` and `Insert` or to use a collection then `string.Join`, any suggestion?

Evangelink:
@mavasani Sorry for the loop error, I got interrupted and forgot that the code wasn't finished when I pushed. Regarding the manual test, here is the output:

```
<NamedType Name="OuterClass">
  <Metrics>
    <Metric Name="MaintainabilityIndex" Value="100" />
    <Metric Name="CyclomaticComplexity" Value="1" />
    <Metric Name="ClassCoupling" Value="0" />
    <Metric Name="DepthOfInheritance" Value="1" />
    <Metric Name="SourceLines" Value="14" />
    <Metric Name="ExecutableLines" Value="0" />
  </Metrics>
</NamedType>
<NamedType Name="OuterClass.InnerClass">
  <Metrics>
    <Metric Name="MaintainabilityIndex" Value="100" />
    <Metric Name="CyclomaticComplexity" Value="1" />
    <Metric Name="ClassCoupling" Value="0" />
    <Metric Name="DepthOfInheritance" Value="1" />
    <Metric Name="SourceLines" Value="10" />
    <Metric Name="ExecutableLines" Value="0" />
  </Metrics>
</NamedType>
<NamedType Name="OuterClass.InnerClass.InnerInnerClass">
  <Metrics>
    <Metric Name="MaintainabilityIndex" Value="100" />
    <Metric Name="CyclomaticComplexity" Value="1" />
    <Metric Name="ClassCoupling" Value="0" />
    <Metric Name="DepthOfInheritance" Value="1" />
    <Metric Name="SourceLines" Value="7" />
    <Metric Name="ExecutableLines" Value="0" />
  </Metrics>
</NamedType>
<NamedType Name="OuterClass.InnerClass.InnerInnerClass.InnerInnerInnerClass">
  <Metrics>
    <Metric Name="MaintainabilityIndex" Value="100" />
    <Metric Name="CyclomaticComplexity" Value="1" />
    <Metric Name="ClassCoupling" Value="0" />
    <Metric Name="DepthOfInheritance" Value="1" />
    <Metric Name="SourceLines" Value="4" />
    <Metric Name="ExecutableLines" Value="0" />
  </Metrics>
</NamedType>
```

PR #3078, opened 2019-11-28
Update CA1812 to handle typeof(), is operator and is pattern

### LINKED ISSUE #1708: CA1812 Shouldn't Fire for DebuggerTypeProxy
To change how a class is displayed in the VS debugger, one defines a type proxy and then adds the DebuggerTypeProxy attribute to the original class, so that the proxy class determines what is displayed by VS instead of the original class. It is natural to make the proxy class internal, since it is not intended to be part of the API surface. (VS finds and instantiates it via reflection.) When one does this, the proxy class triggers a CA1812 warning "XXX is an internal class that is apparently never instantiated."

I propose that if an internal class is referenced in any DebuggerTypeProxy attribute in an assembly, it should not trigger CA1812.

Evangelink:
Fix #1708

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
@@ -1342,8 +1342,108 @@ internal interface IFoo4 {}
 
 internal class CFoo {}
 ",
-                CompilerDiagnostics.None, // Invalid syntaxes around CoClass
-                GetCSharpResultAt(16, 16, AvoidUninstantiatedInternalClassesAnalyzer.Rule, "CFoo"));
+                CompilerDiagnostics.None); // Invalid syntaxes around CoClass
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_TypeOf_NoDiagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+internal class Foo
+{
+}
+
+public class C
+{
+    public C()
+    {
+        var res = typeof(Foo);
+    }
+}");
+
+            await VerifyVB.VerifyAnalyzerAsync(@"
+Friend Class Foo
+End Class
+
+Public Class C
+    Public Sub New()
+        Dim res = GetType(Foo)
+    End Sub
+End Class");
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_DebuggerTypeProxyTypeOf_NoDiagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+internal class FooDebugView
+{
+}
+
+[System.Diagnostics.DebuggerTypeProxy(typeof(FooDebugView))]
+public class Foo
+{
+}
+");
+
+            await VerifyVB.VerifyAnalyzerAsync(@"
+Friend Class FooDebugView
+End Class
+
+<System.Diagnostics.DebuggerTypeProxy(GetType(FooDebugView))>
+Public Class Foo
+End Class");
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_IsOperator_NoDiagnostic()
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1342,8 +1342,108 @@ internal interface IFoo4 {}
 
 internal class CFoo {}
 ",
-                CompilerDiagnostics.None, // Invalid syntaxes around CoClass
-                GetCSharpResultAt(16, 16, AvoidUninstantiatedInternalClassesAnalyzer.Rule, "CFoo"));
```
Evangelink: Issue is no longer reported because `CFoo` is declared through some `typeof()`.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1342,8 +1342,108 @@ internal interface IFoo4 {}
...
+                CompilerDiagnostics.None); // Invalid syntaxes around CoClass
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_TypeOf_NoDiagnostic()
```
sharwell: 📝 I'm not convinced this is desirable. Tagging @mavasani for discussion.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1342,8 +1342,108 @@ internal interface IFoo4 {}
...
+
+Public Class C
+    Public Sub New()
+        Dim res = GetType(Foo)
+    End Sub
+End Class");
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_DebuggerTypeProxyTypeOf_NoDiagnostic()
```
sharwell: 👍 This one I agree with

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1342,8 +1342,108 @@ internal interface IFoo4 {}
...
+Friend Class FooDebugView
+End Class
+
+<System.Diagnostics.DebuggerTypeProxy(GetType(FooDebugView))>
+Public Class Foo
+End Class");
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_IsOperator_NoDiagnostic()
```
sharwell: :grey_exclamation: The `is` operator definitely should not suppress this diagnostic, as it defeats one of the primary use cases.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1342,8 +1342,108 @@ internal interface IFoo4 {}
...
+Public Class C
+    Public Sub New(ByVal o As Object)
+        If TypeOf o Is Foo Then
+        End If
+    End Sub
+End Class");
+        }
+
+        [Fact, WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        public async Task CA1812_IsPattern_NoDiagnostic()
```
sharwell: :grey_exclamation: The `is` (pattern) operator definitely should not suppress this diagnostic, as it defeats one of the primary use cases.

Evangelink:
@sharwell I have to admit I didn't think how much this implementation was changing the purpose of the original rule. I did the implementation thinking more about a rule that would report on unused types rather than something that cares about un-instantiated types. I'll let you discuss but I am more than happy to close this PR and only handle `typeof` for the `DebuggerDisplayAttribute` in #3080

mavasani:
I agree with @sharwell

Evangelink:
Replaced by #3080

PR #3079, opened 2019-11-28
Update CA1707 to not report on discard symbol

### LINKED ISSUE #2440: CA1707 reported for throwaway underscore identifier
#### Analyzer package

#### Package Version

2.9.2

#### Diagnostic ID

[CA1707](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1707-identifiers-should-not-contain-underscores?view=vs-2019)

#### Repro steps

``` csharp
using Microsoft.AspNetCore.Html;
using Microsoft.AspNetCore.Mvc.Rendering;

namespace Test
{
    public static class HtmlHelperExtension
    {
        public static IHtmlContent Example(this IHtmlHelper _)  // CA1707 here
            => new HtmlString("Hello, World!");
    }
}
```

#### Expected behavior

No warning: `_` is a throwaway identifier that suppresses IDE0060.  If I used something else, I'd get IDE0060.

#### Actual behavior

CA1707 reported for parameter `_`.

Evangelink:
Fix #2440

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
@@ -2,6 +2,7 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Globalization;
+using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Testing;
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -439,6 +440,73 @@ public Span(string text)
 ");
         }
 
+        [Fact]
+        public async Task CA1707_CSharp_DiscardSymbolParameter_NoDiagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public static class MyHelper
+{
+    public static int GetBar(this string _) => 42;
```
sharwell: 💭 I'm not sure this is a good thing. If this was my code, I would still expect a proper parameter name here. Tagging @mavasani for discussion.

Evangelink: I see your point. Maybe we could allow discard parameters only for interface implementation or overrides, WDYT?

Evangelink: ping @sharwell

mavasani: Proper parameter name here would generate CA1801 (https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1801?view=vs-2019)

```
If you're using FxCop analyzers, this rule does not flag parameters that are named with the discard symbol, for example, _, _1, and _2. This reduces warning noise on parameters that are needed for signature requirements, for example, a method used as a delegate, a parameter with special attributes, or a parameter whose value is implicitly accessed at run time by a framework but is not referenced in code.
```

PR #3080, opened 2019-11-29
Update CA1812: Handled types declared in DesignerAttribute

### LINKED ISSUE #2957: CA1812 Should ignore DesignerAttribute usage.
Using [Microsoft.CodeAnalysis.FxCopAnalyzers]

#### Package Version

3.0.0.1925803 (8th May 2019)

#### Diagnostic ID

CA1812 

#### Repro steps

1. Create new derived Control e.g. MyTextBox : TextBox
2. Create derived ControlDesigner  MyTextBoxDesginer
3.Mark new Control with [Designer(MyTextBoxDesigner)]

#### Expected behavior

No diagnostic

#### Actual behavior

CA1812 MyTextBoxDesigner is an internal class that is apparently never instantiated.If so, remove the code from the assembly.If this class is intended
to contain only static members, make it static (Shared in Visual Basic)

Evangelink:
Fix #2957

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
@@ -129,14 +129,22 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
                             }
 
                             if (designerAttributeSymbol != null &&
-                                attribute.AttributeClass.Equals(designerAttributeSymbol) &&
-                                attribute.ConstructorArguments[0].Value is string designerTypeName)
+                                attribute.AttributeClass.Equals(designerAttributeSymbol))
                             {
-                                var nameParts = designerTypeName.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
-                                if (nameParts.Length >= 2 &&
-                                    nameParts[1].Trim().Equals(context.Compilation.AssemblyName, StringComparison.Ordinal))
+                                switch (attribute.ConstructorArguments[0].Value)
                                 {
-                                    instantiatedTypeNames.TryAdd(nameParts[0].Trim(), null);
+                                    case string designerTypeName:
+                                        var nameParts = designerTypeName.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                                        if (nameParts.Length >= 2 &&
+                                            nameParts[1].Trim().Equals(context.Compilation.AssemblyName, StringComparison.Ordinal))
+                                        {
+                                            instantiatedTypeNames.TryAdd(nameParts[0].Trim(), null);
+                                        }
+                                        break;
+
+                                    case INamedTypeSymbol namedType:
+                                        instantiatedTypes.TryAdd(namedType, null);
+                                        break;
                                 }
                             }
                         }
--- a/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -45,6 +46,7 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
             analysisContext.RegisterCompilationStartAction(startContext =>
             {
                 var instantiatedTypes = new ConcurrentDictionary<INamedTypeSymbol, object?>();
+                var instantiatedTypeNames = new ConcurrentDictionary<string, object?>();
                 var internalTypes = new ConcurrentDictionary<INamedTypeSymbol, object?>();
 
                 var compilation = startContext.Compilation;
@@ -72,6 +74,7 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
                 var mef1ExportAttributeSymbol = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemComponentModelCompositionExportAttribute);
                 var mef2ExportAttributeSymbol = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCompositionExportAttribute);
                 var coClassAttributeSymbol = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesCoClassAttribute);
+                var designerAttributeSymbol = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemComponentModelDesignerAttribute);
 
                 startContext.RegisterOperationAction(context =>
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,39 @@ public override void Initialize(AnalysisContext analysisContext)
...
-                    // Consider class type declared in the CoClass attribute as instantiated
-                    if (coClassAttributeSymbol != null &&
-                        type.TypeKind == TypeKind.Interface &&
-                        type.GetAttributes().FirstOrDefault(x => x.AttributeClass.Equals(coClassAttributeSymbol)) is AttributeData coClassAttribute &&
-                        coClassAttribute.ConstructorArguments.Length == 1 &&
-                        coClassAttribute.ConstructorArguments[0].Kind == TypedConstantKind.Type &&
-                        coClassAttribute.ConstructorArguments[0].Value is INamedTypeSymbol typeSymbol &&
-                        typeSymbol.TypeKind == TypeKind.Class)
+                    // Consider class types declared in the CoClassAttribute or DesignerAttribute as instantiated
+                    if ((coClassAttributeSymbol != null || designerAttributeSymbol != null) &&
```
Evangelink: We already had to filter attributes for the `CoClassAttribute` case so I have decided to merge the 2 cases into on single loop. This might be a bit less readable as the loop now mixes 2 behaviors. I couldn't really decided which side was best so please let me know your preference.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,39 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                        (type.TypeKind == TypeKind.Class || type.TypeKind == TypeKind.Interface))
                     {
-                        instantiatedTypes.TryAdd(typeSymbol, null);
+                        bool isCoClassHandled = false;
+                        foreach (var attribute in type.GetAttributes())
+                        {
+                            if (coClassAttributeSymbol != null &&
+                                !isCoClassHandled &&
```
Evangelink: We were using `FirstOrDefault` so I kept the same behavior introducing a boolean flag.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,39 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                                {
+                                    instantiatedTypes.TryAdd(typeSymbol, null);
+                                }
+                            }
+
+                            if (designerAttributeSymbol != null &&
+                                attribute.AttributeClass.Equals(designerAttributeSymbol) &&
+                                attribute.ConstructorArguments[0].Value is string designerTypeName)
+                            {
+                                var nameParts = designerTypeName.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
```
Evangelink: Let me know if you prefer that I use substring + indexof instead of split.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,39 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                                if (nameParts.Length >= 2 &&
+                                    nameParts[1].Trim().Equals(context.Compilation.AssemblyName, StringComparison.Ordinal))
```
Evangelink: `Trim` is required because when splitting `Foo, MyAssembly` the second string starts with a whitespace. Even though trimming the start would be sufficient it seems safer to handle extra trailing space.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,39 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                                {
+                                    instantiatedTypeNames.TryAdd(nameParts[0].Trim(), null);
```
Evangelink: `Trim` is there to play on the safe side in case there is any leading/trailing whitespace.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -144,21 +144,6 @@ public async Task CA1812_Basic_NoDiagnostic_InternalModule()
...
-        }
-
-        [Fact]
-        public async Task CA1812_Basic_NoDiagnostic_InternalAbstractClass()
-        {
-            await VerifyVB.VerifyAnalyzerAsync(
-@"Friend MustInherit Class A
-End Class");
-        }
-
```
Evangelink: Oops I will revert this change!

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,47 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                                    attribute.ConstructorArguments[0].Kind == TypedConstantKind.Type &&
+                                    attribute.ConstructorArguments[0].Value is INamedTypeSymbol typeSymbol &&
+                                    typeSymbol.TypeKind == TypeKind.Class)
+                                {
+                                    instantiatedTypes.TryAdd(typeSymbol, null);
+                                }
+                            }
+
+                            if (designerAttributeSymbol != null &&
+                                attribute.AttributeClass.Equals(designerAttributeSymbol))
```
mavasani: Needs additional check for `attribute.ConstructorArguments.Length` being non-zero.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,47 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                            {
+                                switch (attribute.ConstructorArguments[0].Value)
+                                {
+                                    case string designerTypeName:
```
mavasani: Can this be null?

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -204,7 +238,8 @@ static IEnumerable<INamedTypeSymbol> GetAllNamedTypeConstraints(ITypeParameterSy
                     var uninstantiatedInternalTypes = internalTypes
                         .Select(it => it.Key.OriginalDefinition)
                         .Except(instantiatedTypes.Select(it => it.Key.OriginalDefinition))
-                        .Where(type => !HasInstantiatedNestedType(type, instantiatedTypes.Keys));
+                        .Where(type => !HasInstantiatedNestedType(type, instantiatedTypes.Keys))
+                        .Where(type => !instantiatedTypeNames.ContainsKey(type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat).Substring("global::".Length)));
```
mavasani: This doesn't look correct. We shouldn't be relying on fully qualified names and should not be invoking ToDisplayString.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -104,16 +107,47 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                                attribute.AttributeClass.Equals(designerAttributeSymbol))
+                            {
+                                switch (attribute.ConstructorArguments[0].Value)
+                                {
+                                    case string designerTypeName:
+                                        var nameParts = designerTypeName.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                                        if (nameParts.Length >= 2 &&
+                                            nameParts[1].Trim().Equals(context.Compilation.AssemblyName, StringComparison.Ordinal))
+                                        {
+                                            instantiatedTypeNames.TryAdd(nameParts[0].Trim(), null);
```
mavasani: I would avoid tracking names. Can we instead try to invoke compilation.GetTypeByMetadataName for the name and see if it is a non-null symbol whose containing assembly matches compilation's assembly.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
Evangelink: I guess it can, thank you.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/AvoidUninstantiatedInternalClasses.cs
Evangelink: Indeed that's a way better (faster and more robust) solution... I don't know why I didn't think of it.

src/Microsoft.CodeQuality.Analyzers/UnitTests/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1346,6 +1346,154 @@ internal class CFoo {}
...
+
+namespace Foo
+{
+    [Designer(""Foo.MyTextBox.MyTextBoxDesigner, TestProject"")]
+    public class MyTextBox
+    {
+        internal class MyTextBoxDesigner { }
+    }
+}",
+                // False-Positive: when evaluating the string of the DesignerAttribute the type symbol doesn't exist yet
```
Evangelink: @mavasani FYI we no longer properly support the case of a nested-type referenced by string from an outer designer attribute.

mavasani: I think this is fine.

PR #3081, opened 2019-11-29
New analyzer for classes that are not DiagnosticAnalyzer

### LINKED ISSUE #1690: Feature request: warn if DiagnosticAnalyzerAttribute is used on a type that is not a DiagnosticAnalyzer
#### Analyzer package

Microsoft.CodeAnalysis.Analyzers

#### Repro steps

```cs
namespace RoslynSandbox
{
    using Microsoft.CodeAnalysis;
    using Microsoft.CodeAnalysis.Diagnostics;

    internal static class Foo
    {
    }
}
```

Evangelink:
I did the implementation as a new rule but if you prefer me to update the behavior of `RS0001` instead I could do so.

Fix #1690

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
@@ -374,4 +374,13 @@
   <data name="EnableConcurrentExecutionFix" xml:space="preserve">
     <value>Enable concurrent execution</value>
   </data>
+  <data name="ClassIsNotDiagnosticAnalyzerDescription" xml:space="preserve">
+    <value>Types marked with DiagnosticAnalyzerAttribute(s) should inherit from DiagnosticAnalyzer.</value>
+  </data>
+  <data name="ClassIsNotDiagnosticAnalyzerMessage" xml:space="preserve">
+    <value>Inherit from DiagnosticAnalyzer or remove the DiagnosticAnalyzerAttribute(s).</value>
+  </data>
+  <data name="ClassIsNotDiagnosticAnalyzerTitle" xml:space="preserve">
+    <value>Class is not a DiagnosticAnalyzer.</value>
+  </data>
 </root>
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
@@ -374,13 +374,10 @@
   <data name="EnableConcurrentExecutionFix" xml:space="preserve">
     <value>Enable concurrent execution</value>
   </data>
-  <data name="ClassIsNotDiagnosticAnalyzerDescription" xml:space="preserve">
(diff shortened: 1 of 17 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/ClassIsNotDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+                    // We don't need to check assemblies unless they're referencing Microsoft.CodeAnalysis which defines DiagnosticAnalyzer.
+                    return;
+                }
+
+                csac.RegisterSymbolAction(sac =>
+                {
+                    var namedType = (INamedTypeSymbol)sac.Symbol;
+
+                    if (namedType.TypeKind == TypeKind.Class &&
+                        namedType.GetAttributes().Any(a => a.AttributeClass.Equals(diagnosticAnalyzerAttribute)) &&
```
mavasani:
```suggestion
                        namedType.HasAttribute(diagnosticAnalyzerAttribute) &&
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/ClassIsNotDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+                        !namedType.GetBaseTypes().Any(type => type.Equals(diagnosticAnalyzer)))
```
mavasani:
```suggestion
                        !namedType.DerivesFrom(diagnosticAnalyzer, baseTypesOnly: true))
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -374,4 +374,13 @@
   <data name="EnableConcurrentExecutionFix" xml:space="preserve">
     <value>Enable concurrent execution</value>
   </data>
+  <data name="ClassIsNotDiagnosticAnalyzerDescription" xml:space="preserve">
+    <value>Types marked with DiagnosticAnalyzerAttribute(s) should inherit from DiagnosticAnalyzer.</value>
+  </data>
+  <data name="ClassIsNotDiagnosticAnalyzerMessage" xml:space="preserve">
+    <value>Inherit from DiagnosticAnalyzer or remove the DiagnosticAnalyzerAttribute(s).</value>
```
mavasani:
```suggestion
    <value>Inherit type '{0}' from DiagnosticAnalyzer or remove the DiagnosticAnalyzerAttribute(s).</value>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -374,4 +374,13 @@
...
+  </data>
+  <data name="ClassIsNotDiagnosticAnalyzerTitle" xml:space="preserve">
+    <value>Class is not a DiagnosticAnalyzer.</value>
```
mavasani: I would probably just use `Types marked with DiagnosticAnalyzerAttribute(s) should inherit from DiagnosticAnalyzer.` as the Title and avoid having a description at all.

PR #3083, opened 2019-12-02
Update CA1806 to report on ignored IDisposable returns

### LINKED ISSUE #1057: CA1806 DoNotIgnoreMethodResults should generate warning for IDisposable return values
#### Analyzer package

`Microsoft.Maintainability.Analyzers`
#### Analyzer

CA1806: `DoNotIgnoreMethodResults`
#### Repro steps

Consider the following line:

```
cancellationToken.Register(() => socket.Dispose());
```

where [`CancellationToken.Register`](https://msdn.microsoft.com/nl-be/library/dd321635%28v=vs.110%29.aspx) returns a `CancellationTokenRegistration`, which implements `IDisposable`.
#### Expected behavior

CA1806 does not generate a warning, apparently it only triggers warnings for very specific methods.

However, I believe that CA1806 could benefit from generating a warning if:
- The method returns a value
- That value implements `IDisposable`

This seems to be a clear indication that the return value should be disposed of by the caller; hence the caller should use the method result.
#### Actual behavior

CA1806 does not generate a warning.

Evangelink:
Fix #1057


src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -130,9 +141,11 @@ public override void Initialize(AnalysisContext analysisContext)
 
             analysisContext.RegisterCompilationStartAction(compilationContext =>
             {
-                INamedTypeSymbol? expectedExceptionType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingExpectedExceptionAttribute);
-                INamedTypeSymbol? nunitAssertType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.NUnitFrameworkAssert);
-                INamedTypeSymbol? xunitAssertType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitAssert);
+                WellKnownTypeProvider wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation);
```
Evangelink: Extract the well known type provider as it is used multiple times in a row.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -181,6 +194,14 @@ public override void Initialize(AnalysisContext analysisContext)
                                 {
                                     rule = PureMethodRule;
                                 }
+                                else if (disposableType != null &&
+                                    targetMethod.ReturnType.ImplementsIDisposable(disposableType) &&
+                                    // We want to ignore some of the async methods from assertion frameworks
```
Evangelink: I had to add those exclusions as one of the condition of the `ShouldSkipAnalyzing` isn't true (see later comment).

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -197,6 +218,33 @@ public override void Initialize(AnalysisContext analysisContext)
                             opContext.ReportDiagnostic(diagnostic);
                         }
                     }, OperationKind.ExpressionStatement);
+
+                    osContext.RegisterOperationAction(opContext =>
```
Evangelink: In this operation we want to focus on inner invocations as the other operation only focuses on outer ones.

src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1343,4 +1343,7 @@
   <data name="AssigningSymbolAndItsMemberInSameStatementTitle" xml:space="preserve">
     <value>Assigning symbol and its member in the same statement.</value>
   </data>
+  <data name="DoNotIgnoreMethodResultsMessageDisposable" xml:space="preserve">
+    <value>{0} calls {1} which is never disposed. Ensure this is the intended use.</value>
```
Evangelink: Please feel free to suggest a better message. On the ticket the message was more assertive but IMO we cannot guarantee that a method returning some `IDisposable` expect the caller to dispose it. The `IDisposable` interface mixes the notion of owned resource and used/shared resource.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -116,6 +116,15 @@ public sealed class DoNotIgnoreMethodResultsAnalyzer : DiagnosticAnalyzer
                                                                              isPortedFxCopRule: true,
                                                                              isDataflowRule: false);
 
+        internal static DiagnosticDescriptor DisposableRule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                             s_localizableTitle,
+                                                                             s_localizableMessageDisposable,
+                                                                             DiagnosticCategory.Performance,
+                                                                             RuleLevel.IdeSuggestion,
+                                                                             description: s_localizableDescription,
+                                                                             isPortedFxCopRule: false,
```
Evangelink: This wasn't part of fxcop so I have put `false`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1370,4 +1370,7 @@
   <data name="EnumShouldNotHaveDuplicatedValuesTitle" xml:space="preserve">
     <value>Enums values should not be duplicated</value>
   </data>
+  <data name="DoNotIgnoreMethodResultsMessageDisposable" xml:space="preserve">
+    <value>{0} calls {1} which is never disposed. Ensure this is the intended use.</value>
```
Evangelink: `MyMethod calls GetSomethingDisposable which is never disposed. Ensure this is the intended use.`

maybe it would be better with

`'MyMethod' calls 'GetSomethingDisposable' whose value is never disposed. Ensure this is the intended use.`

Evangelink:
Tagging @sharwell or @mavasani

Evangelink:
Ping @mavasani

Evangelink:
Ping @mavasani (sorry)

PR #3097, opened 2019-12-05
Rewrite the Analyzer Implementation to use IOpertion API

### LINKED ISSUE #1214: Rewrite the Analyzer Implementation to use IOpertion API
- [ ] [CA1821](https://github.com/dotnet/roslyn-analyzers/issues/1176#issuecomment-304150350) : blocked by dotnet/roslyn#17861

Evangelink:
Fix #1214


src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/RemoveEmptyFinalizers.cs
```diff
@@ -33,31 +35,45 @@ public override void Initialize(AnalysisContext analysisContext)
...
             analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
-            analysisContext.RegisterCodeBlockAction(codeBlockContext =>
+            analysisContext.RegisterOperationBlockStartAction(obsac =>
             {
-                if (codeBlockContext.OwningSymbol.Kind != SymbolKind.Method)
+                if (!obsac.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsConditionalAttribute, out var conditionalAttributeType) ||
+                    obsac.OperationBlocks.Length != 1 ||
+                    !(obsac.OperationBlocks[0] is IBlockOperation blockOperation) ||
+                    !(obsac.OwningSymbol is IMethodSymbol methodSymbol) ||
```
Evangelink: Is there any real benefit for me to check the `Kind` before doing the cast or is the performance gain negligible? (e.g. instead of `!(obsac.OwningSymbol is IMethodSymbol methodSymbol)` I could write `obsac.OwningSymbol.Kind != SymbolKind.MethodSymbol` and do the cast afterward).

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/RemoveEmptyFinalizers.cs
```diff
@@ -33,31 +35,45 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                    !methodSymbol.IsDestructor())
                 {
                     return;
                 }
 
-                var methodSymbol = (IMethodSymbol)codeBlockContext.OwningSymbol;
-                if (!methodSymbol.IsDestructor())
+                const int IS_EMPTY = 0;
+                const int IS_NOT_EMPTY = 1;
```
Evangelink: Do you prefer fields for these constants?

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/RemoveEmptyFinalizers.cs
```diff
@@ -33,31 +35,45 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                var state = IS_EMPTY;
+
+                obsac.RegisterOperationAction(oac =>
                 {
-                    return;
-                }
+                    if (Interlocked.CompareExchange(ref state, IS_EMPTY, IS_EMPTY) == IS_EMPTY &&
```
Evangelink: I am pretty beginner with `Interlocked` use so there might be better solutions to have this flag done in a thread-safe manner. Note that I have tried to avoid running the lambda if we have already found any valid statement.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/RemoveEmptyFinalizers.cs
```diff
@@ -33,31 +35,45 @@ public override void Initialize(AnalysisContext analysisContext)
...
                     return;
                 }
 
-                var methodSymbol = (IMethodSymbol)codeBlockContext.OwningSymbol;
-                if (!methodSymbol.IsDestructor())
+                const int IS_EMPTY = 0;
+                const int IS_NOT_EMPTY = 1;
+                var state = IS_EMPTY;
+
+                obsac.RegisterOperationAction(oac =>
```
mavasani: @Evangelink Sorry for not replying earlier on this. I think we want to avoid searching for specific operation kinds within the block. Instead we want to identify if the operation block has any _explicit_ IOperations which were not compiler generated. You should be able to use `IOperationExtensions.WithoutFullyImplicitOperations` API to check this and report a diagnostic.

Evangelink: Thank you for the pointer. I still need to do some extra processing after this method call because the current implementation excludes invocations to methods marked with the conditional attribute and also `throw` expressions.

mavasani: That should be fine, and the strategy we have used for other similar analyzers.

Evangelink: @mavasani I was trying to rewrite the analyzer using `WithoutFullyImplicitOperations` but I can't manage to have the same results as with the current implementation. I do need to know the specific children operation to make a decision. For example, if the block only contains an invocation it can be wrapped in an expression so I need to access to the invocation child to detect whether or not this is a conditional call. At the same time, I don't want to get the children of the invocation because I don't care about the arguments... Could you have a quick look and let me know if you see how to untangle this please?!

Evangelink: @mavasani no rush but when you have some time to look at this, I'd be glad to have some feedback.

mavasani: Can you extract out the code to a helper/extension methods that operates on the entire Operation block, i.e. the root? I presume the same functionality is likely to be needed in different analyzer(s).

Evangelink: 1/ Does it mean you are fine with the current algo?

2/ I am not sure which part you would like to extract.

mavasani: I was thinking:
- Helper/extension method: `IsEmptyOperationBlock(IOperation root, bool ignoreConditionalCalls)`
- Logic:
   - Use `GetTopmostExplicitDescendants` to get all non-implicit topmost operations
   - if (ignoreConditionalCalls) then verify all the non-implicit operations are invocations to conditional methods
   - otherwise, verify there are no non-implicit operations

Evangelink: I'm sorry but there is something I don't get with the logic. I think that we want to get to the block operation of the given operation (so using `GetTopmostParentBlock()` and not `GetTopmostExplicitDescendants()`) then loop through the descendants (call to `Descendants()`) and check for the presence of some specific operations (`SimpleAssignment`, `VariableDeclaration`, `Loop`, `Conditional` and `Invalid`) and filter out or not `Invocation` operations depending on `ignoreConditionalCalls`. Am I right?

mavasani: > check for the presence of some specific operations

That is my primary concern with your approach - deciding specific operations is not a scalable or reliable approach. I was seeing this as look at the operations inside method body, if there are no explicit operations, the block is empty. Otherwise, let the caller decide if they want to ignore conditional invocations or not.

Evangelink: @mavasani I have made quite some tests with the directions you have been describing and I don't manage to have the tests passing. I am not sure how to move on with this, maybe it's best if we close this PR?

Evangelink:
@mavasani I think this new implementation shall be closed to what you were expecting 😄

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -638,6 +638,38 @@ public static IOperation WalkDownConversion(this IOperation operation)
...
+                {
+                    var current = enumerator.Current;
+                    stack.Add(enumerator);
+
+                    if (current != null &&
+                        (descendIntoOperation == null || descendIntoOperation(current)))
+                    {
+                        if (!current.IsImplicit &&
+                            // This prevents non explicit operations like expression to be considered as ok
+                            (current.ConstantValue.HasValue || current.Type != null))
```
Evangelink: This second part of the condition is copied from the `WithoutFullyImplicitOperations` helper but it is a bit more global than the other helper was suggesting (i.e. `or are explicit with no constant value or type, indicating it is not user written code`). The conditional operation linked to a `if` statement for example will be considered as not user written code _BUT_ the content of the condition will be properly analyzed.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -638,6 +638,38 @@ public static IOperation WalkDownConversion(this IOperation operation)
...
+                if (enumerator.MoveNext())
+                {
+                    var current = enumerator.Current;
+                    stack.Add(enumerator);
+
+                    if (current != null &&
+                        (descendIntoOperation == null || descendIntoOperation(current)))
+                    {
+                        if (!current.IsImplicit &&
+                            // This prevents non explicit operations like expression to be considered as ok
```
mavasani: Nit: consider moving the comment outside the if expression. Can do it as a follow-up PR or part of any other PR in future.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RemoveEmptyFinalizers.cs
```diff
@@ -32,31 +33,29 @@ public override void Initialize(AnalysisContext analysisContext)
...
-                    return;
-                }
-
-                var methodBody = methodSymbol.DeclaringSyntaxReferences.First().GetSyntax();
-
-                if (IsEmptyFinalizer(methodBody, codeBlockContext))
-                {
-                    codeBlockContext.ReportDiagnostic(methodSymbol.CreateDiagnostic(Rule));
-                }
+                    if (context.OperationBlocks.Length == 1 &&
```
mavasani: Great, this looks much better.

PR #3098, opened 2019-12-05
Add support for IAsyncDisposable in dispose analysis

### LINKED ISSUE #3042: CA2000 emitted when await using (false positive)
#### Analyzer package

#### Package Version

v2.9.7 (Latest)

#### Diagnostic ID

Example: [CA2000](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2000?view=vs-2019)

#### Repro steps

1. .Net Core 3.1 (latest preview).  I happen to be using Blazor Server.
2. Create a Service class that accepts a DbContextOptions
3. Instantate a DbContext in a method.  Use the "await using..." C#8 syntax

```
public class MyService
{
	public DocTypesService(DbContextOptions<MyDbContext> options)
	{
		this.options = options;
	}

	private readonly DbContextOptions<MyDbContext> options;

	public async Task<Something> GetSomethingAsync(string id, CancellationToken token)
	{
		await using var context = new DocTypesContext(this.options);
			....
	}
}
```

#### Expected behavior

> await using var context = new DocTypesContext(this.options);

The context is created with a using statement, and will be disposed of when it goes out of scope and so no CA2000 error should **not** be emitted.

A CA2000 error is emitted.  

However, were we to drop the `await` command `using var context = new DocTypesContext(this.options);` then the error is not emitted, but we really do need to put the `await` in.

mavasani:
Fixes #3042

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysisHelper.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysisHelper.cs
@@ -41,14 +41,18 @@ namespace Analyzer.Utilities
         private readonly ImmutableHashSet<INamedTypeSymbol> _disposeOwnershipTransferLikelyTypes;
         private ConcurrentDictionary<INamedTypeSymbol, ImmutableHashSet<IFieldSymbol>>? _lazyDisposableFieldsMap;
         public INamedTypeSymbol? IDisposable { get; }
+        public INamedTypeSymbol? IAsyncDisposable { get; }
         public INamedTypeSymbol? Task { get; }
+        public INamedTypeSymbol? ValueTask { get; }
 
         private DisposeAnalysisHelper(Compilation compilation)
         {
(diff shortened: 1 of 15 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysisHelper.cs
```diff
@@ -180,5 +184,11 @@ public bool IsDisposableCreationOrDisposeOwnershipTransfer(AbstractLocation loca
 
             return IsDisposableCreation(location.CreationOpt);
         }
+
+        public bool IsDisposable([NotNullWhen(returnValue: true)] ITypeSymbol? type)
+            => type != null && type.IsDisposable(IDisposable, IAsyncDisposable);
+
+        public DisposeMethodKind GetDisposeMethodKind(IMethodSymbol method)
+            => method.GetDisposeMethodKind(IDisposable, IAsyncDisposable, Task, ValueTask);
     }
```
dotpaul: Nit: Worth consolidating with DataFlowOperationVisitor's implementation?

mavasani:
Thank you @dotpaul

Issue #3092, opened 2019-12-04
Error with new test framework - System.ValueTuple`2


Evangelink:
When migrating the tests to the new test framework the following 2 tests:

- `TestUse768DeconstructionAssignmentNoDiagnostic`
- `Use768DeconstructionAssignment_NoDiagnostic`

are failing with the error: 

```
error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
```

Issue #3091, opened 2019-12-04
Migrate test UseTls12Or192_Diagnostic to new test framework


Evangelink:
On branch `2.9.x`, there is an issue when migrating the test `UseTls12Or192_Diagnostic`, the 2 issues are on the exact same span and keep failing with an incorrect order.

PR #3099, opened 2019-12-05
Convert NetCore Security tests to the new test framework

sharwell:
Supersedes #3057

Primary [changes](https://github.com/Evangelink/roslyn-analyzers/compare/new-tests-security..sharwell:new-tests-security) from the original:

* Avoid updates to target frameworks
* Avoid changing default behavior except for security tests
* Use reference assemblies instead of System.Web, System.Web.Extensions, and System.Runtime.Serialization
* Use package references for Newtonsoft.Json and WindowsAzure.Storage


mavasani:
Tagging @dotpaul for review

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotDisableHttpClientCRLCheckTests.cs
```diff
@@ -1,31 +1,42 @@
...
+        private async Task VerifyCSharpWithDependenciesAsync(string source, params DiagnosticResult[] expected)
         {
-            this.VerifyCSharp(
-                new[] { source, SystemNetHttpApis.CSharp }.ToFileAndSource(),
-                expected);
+            var csharpTest = new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources = { source, SystemNetHttpApis.CSharp },
```
Evangelink: Is it still useful to stick with minimal implementations rather than adding a reference to the actual code?

src/Microsoft.NetCore.Analyzers/UnitTests/Data/ReviewSQLQueriesForSecurityVulnerabilitiesTests.cs
```diff
@@ -1,23 +1,30 @@
...
-using System;
-using Microsoft.CodeAnalysis.Diagnostics;
+using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Testing;
 using Test.Utilities;
+using Test.Utilities.MinimalImplementations;
 using Xunit;
+using VerifyCS = Test.Utilities.CSharpSecurityCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Data.ReviewSqlQueriesForSecurityVulnerabilities,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
```
Evangelink: We have been using a lot this syntax with the `EmptyCodeFixProvider`, would it be worth having a sub-class to `CSharpSecurityCodeFixVerifier` with only the analyzer as generic type argument? On one side that would make it easier for new rule but on the other side that means that there are 2 ways to declare this static using. WDYT?

Evangelink:
I have just seen that we forgot to convert the `TaintedDataAnalyzerTestBase` class to the new "format". Shall we add an extra commit or shall we create a separate ticket for it and move on with this PR?

Evangelink:
@sharwell Could you try and revert the commit `Revert some changes to avoid failing tests - 83ead98c45396910d0c84706f53491e71cae3d3b` from #3057 and see if you are still experiencing the following issues: #3092, #3091?

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotDisableHttpClientCRLCheckTests.cs
sharwell: I planning to remove most of the other minimal implementations in the master branch.

src/Microsoft.NetCore.Analyzers/UnitTests/Data/ReviewSQLQueriesForSecurityVulnerabilitiesTests.cs
sharwell: I think it's fine like this. It helps with two things:

1. Encourage creating a code fix with an analyzer as the recommended practice
2. Discourage separating analyzer tests from code fix tests in separate files

sharwell:
@Evangelink yep it worked, I resolved those two issues

sharwell:
> I have just seen that we forgot to convert the TaintedDataAnalyzerTestBase class to the new "format". Shall we add an extra commit or shall we create a separate ticket for it and move on with this PR?

I'll fix this in a separate pull request.

src/Microsoft.NetCore.Analyzers/UnitTests/Security/DoNotUseInsecureDeserializerBinaryFormatterWithoutBinderTests.cs
```diff
@@ -940,14 +943,13 @@ public object DeserializeWithDelegate(byte[] bytes)
     }
 }",
                 GetCSharpResultAt(21, 20, BinderNotSetRule, "object BinaryFormatter.Deserialize(Stream serializationStream, HeaderHandler handler)"),
-                GetCSharpResultAt(21, 20, BinderNotSetRule, "object BinaryFormatter.UnsafeDeserialize(Stream serializationStream, HeaderHandler handler)")
-                );
+                GetCSharpResultAt(21, 20, BinderNotSetRule, "object BinaryFormatter.UnsafeDeserialize(Stream serializationStream, HeaderHandler handler)"));
         }
```
dotpaul: Other than this test not reliably passing (diagnostic ordering?), all the changes LGTM.

sharwell: Thanks. I'll stabilize this one before merging.

mavasani: If fixing the test/test framework will take time, can we skip the test and merge the PR?

sharwell: I'll get it fixed today

sharwell: Fixed by simply removing the last commit from this PR

mavasani:
@sharwell Is this good to merge?

sharwell:
@Evangelink @dotpaul thank you both for all the work here

dotpaul:
Yes, thank you @Evangelink

PR #3102, opened 2019-12-06
Rule CA1069: Enums values should not be duplicated

### LINKED ISSUE #1646: Warn on duplicate explicit enum values
If you have an enum with explicit values, we should warn if two of those values are identical. These can often be introduced when you're merging in work done in different branches, and can cause failures that seem impossible and are not readily apparent.

Evangelink:
Fix #1646


src/Microsoft.CodeQuality.Analyzers/CSharp/ApiDesignGuidelines/CSharpEnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,75 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpEnumShouldNotHaveDuplicatedValues : EnumShouldNotHaveDuplicatedValues
+    {
+        protected override void AnalyzeEnumMemberValues(Dictionary<object, List<SyntaxNode>> membersByValue, SymbolAnalysisContext context)
+        {
+            foreach (var kvp in membersByValue)
+            {
+                for (int i = 0; i < kvp.Value.Count; i++)
+                {
+                    var enumMember = (EnumMemberDeclarationSyntax)kvp.Value[i];
```
Evangelink: Shall I go defensive here?

src/Microsoft.CodeQuality.Analyzers/CSharp/ApiDesignGuidelines/CSharpEnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+        }
+
+        private static void AnalyzeBinaryExpression(BinaryExpressionSyntax binaryExpression, SymbolAnalysisContext context, bool isDuplicatedValue)
+        {
+            var hasBitwiseIssue = false;
+            var seen = new HashSet<object>();
+
+            foreach (var identifier in binaryExpression.DescendantNodes().OfType<IdentifierNameSyntax>())
```
Evangelink: Is it better to pass the descending predicate to the `DescendantNodes` call?

Evangelink: Is it better if I avoid the `OfType` call and do a `as` cast inside the loop?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,67 @@
...
+                                                                                                   customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(RuleDuplicatedValue, RuleDuplicatedBitwiseValuePart);
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSymbolAction(symbolContext =>
```
Evangelink: I couldn't find any good way to achieve something similar using operations. I hope that's fine like this.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                var typeSymbol = (INamedTypeSymbol)symbolContext.Symbol;
+                if (typeSymbol.TypeKind != TypeKind.Enum)
+                {
+                    return;
+                }
+
+                var enumMemberValues = typeSymbol.GetMembers()
+                    .Where(m => m.Kind == SymbolKind.Field)
```
Evangelink: I don't know whether that's best for the analyzer to do the kind check before the cast or to do `OfType` directly.

src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1352,4 +1352,13 @@
   <data name="AvoidInfiniteRecursionMessageMaybe" xml:space="preserve">
     <value>Do not assign the property within its setter. This call might result in an infinite recursion.</value>
   </data>
+  <data name="EnumShouldNotHaveDuplicatedValuesMessageDuplicatedBitwiseValuePart" xml:space="preserve">
+    <value>This enum value bitwise part duplicates another part.</value>
```
Evangelink: I don't like the message but I don't have much inspiration today.

src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1352,4 +1352,13 @@
...
+  </data>
+  <data name="EnumShouldNotHaveDuplicatedValuesMessageDuplicatedValue" xml:space="preserve">
+    <value>This enum value duplicates another value.</value>
```
Evangelink: I don't like the message but I don't have much inspiration today.

src/Microsoft.CodeQuality.Analyzers/VisualBasic/ApiDesignGuidelines/BasicEnumShouldNotHaveDuplicatedValues.vb
```diff
@@ -0,0 +1,63 @@
...
+
+Imports Analyzer.Utilities.Extensions
+Imports Microsoft.CodeAnalysis
+Imports Microsoft.CodeAnalysis.Diagnostics
+Imports Microsoft.CodeAnalysis.VisualBasic.Syntax
+Imports Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
+
+Namespace Microsoft.CodeQuality.VisualBasic.Analyzers.ApiDesignGuidelines
+    <DiagnosticAnalyzer(LanguageNames.VisualBasic)>
+    Public Class BasicEnumShouldNotHaveDuplicatedValues
```
Evangelink: I am not really good with VB.NET so feel free to suggest better syntaxes for the whole class.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
mavasani: I think we can write this as a much simpler language agnostic analyzer, completely based on operations as follows:

1. Register SymbolStartAction for NamedTypeSymbols.
2. Bail out from callback if type is not enum
3. Register nested operation callback for `OperationKind.FieldInitializer`.
4. Bail out in operation callback if:
   1. InitializedFields.Count != 1
   2. Operation does not have constant value
   3. Has any child IFieldReferenceOperation whose underlying field has same containing type
5. Otherwise, add to a dictionary of constant value to list of fields with that value
6. In SymbolEndAction, report diagnostics for all dictionary entries that have more than one field in the value - one diagnostic for each field excluding the one that is lexically the first field.
3.

Evangelink: Yay that does work really well! I'll push the changes tomorrow.

Evangelink: Regarding point 4. iii, I think the condition shall be more complex otherwise we won't report for duplicated bitwise parts nor for binary expression using field ref like `Value3 = Value2 * 42` (no idea if someone writes that but that's allowed by the syntax).

Evangelink: Also we need to cover the bitwise duplication `Value1 | Value1` but we need to be careful about other binary expressions `Value1 | Value2 + Value1` shouldn't be marked with the bitwise error.

Evangelink: There is also an issue with uninitialized values 
```csharp
public enum MyEnum
{
    Value1 = 1,
    Value2,
    Value3 = 2,
}
```
I can't handle this case with the new technique.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            context.RegisterSymbolStartAction(ssac =>
+            {
+                var enumSymbol = (INamedTypeSymbol)ssac.Symbol;
+                if (enumSymbol.TypeKind != TypeKind.Enum)
+                {
+                    return;
+                }
+
+                var invalidMembersByValue = new ConcurrentDictionary<object, List<SyntaxNode>>();
```
Evangelink: Is there another concurrent dictionary that you recommend to use?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+                ssac.RegisterOperationAction(oc =>
+                {
+                    var fieldInitializerOperation = (IFieldInitializerOperation)oc.Operation;
+                    if (fieldInitializerOperation.InitializedFields.Length != 1 ||
+                        !fieldInitializerOperation.Value.ConstantValue.HasValue)
+                    {
+                        return;
+                    }
+
+                    var onlyReferencesOneField = GetFilteredDescendants(fieldInitializerOperation, op => op.Kind != OperationKind.Binary)
```
Evangelink: Filters referenced fields that are not part of a binary expression. If any there shouldn't be more than 1 found, do you want me to use a debug specific code with an assertion for this?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+                        .OfType<IFieldReferenceOperation>()
+                        .Any();
+                    if (onlyReferencesOneField)
+                    {
+                        return;
+                    }
+
+                    invalidMembersByValue.AddOrUpdate(fieldInitializerOperation.Value.ConstantValue.Value,
+                        new List<SyntaxNode> { fieldInitializerOperation.Syntax.Parent },
```
Evangelink: Get parent syntax to get to the field itself as we are currently on the initializer.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+                        (key, value) =>
+                        {
+                            value.Add(fieldInitializerOperation.Syntax.Parent);
```
Evangelink: Get parent syntax to get to the field itself as we are currently on the initializer.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+                            return value;
+                        });
+                }, OperationKind.FieldInitializer);
+
+                // ...and report at the end of the enum declaration
+                ssac.RegisterSymbolEndAction(sac =>
+                {
+                    foreach (var kvp in invalidMembersByValue)
+                    {
+                        foreach (var item in kvp.Value.OrderBy(x => x.GetLocation().SourceSpan).Skip(1))
```
Evangelink: Is there anything better than location for the sorting to avoid reproducibility issues?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+                        {
+                            oc.ReportDiagnostic(item.CreateDiagnostic(RuleDuplicatedBitwiseValuePart));
+                        }
+                    }
+                }, OperationKind.Binary);
+            }, SymbolKind.NamedType);
+        }
+
+        private static IEnumerable<IOperation> GetFilteredDescendants(IOperation operation, Func<IOperation, bool> descendIntoOperation)
```
Evangelink: This is a copy of the `Descendant` method where I have replaced `Push`/`Pop` calls with `Add`/`Get`/`Remove` calls as the `ArrayBuilder` is different from the one in Roslyn.

Is it worth suggesting this predicate overload on Roslyn?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            while (stack.Any())
+            {
+                var enumerator = stack.Last();
+                stack.RemoveLast();
+                if (enumerator.MoveNext())
+                {
+                    var current = enumerator.Current;
+                    stack.Add(enumerator);
+                    if (current != null && descendIntoOperation(current))
```
Evangelink: I have added the predicate check.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        [Fact]
+        public async Task EnumImplicitDuplication_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public enum MyEnum
+{
+    Value1 = 1,
+    Value2,
```
Evangelink: This case is no longer handled as there is no field initializer here.

I am actually wondering if it might be useful to have a change on Roslyn to have a callback for the default initialization of a field with some flag (maybe implicit) to detect user written from auto-generated ones. WDYT?

The only solution I found to handle this case isn't right IMO as it consists of a list initialized with the enum fields, for which I'll remove each initialized field in the operation block. In the end action, I will add to the dictionary all uninitialized fields with their values.

mavasani:
FYI: You will need to move the added analyzer files to a different folder and project, Microsoft.CodeAnalysis.NetAnalyzers`, once https://github.com/dotnet/roslyn-analyzers/pull/3165 is merged. See https://github.com/dotnet/roslyn-analyzers/pull/3139#issue-358842977 for details.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
Evangelink: ping @mavasani

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
mavasani: What if the referenced field belongs to a different type? Do you need to check the containing type of referenced fields?

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        public async Task EnumDuplicatedValueMaths_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public enum MyEnum
+{
+    Value1 = 1,
+    Value2 = 2 - 1,
+    Value3 = Value1 * Value2,
+}",
+                GetCSharpResultAt(5, 5),
```
mavasani: I think it feels worthwhile to mention the duplicated value and the field names with the duplicated value in the diagnostic message. Would that be easy to implement?

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
mavasani: > I am actually wondering if it might be useful to have a change on Roslyn to have a callback for the default initialization of a field with some flag (maybe implicit) to detect user written from auto-generated ones.

That is a good idea, can you open a Roslyn issue and tag me on it?

> The only solution I found to handle this case isn't right IMO as it consists of a list initialized with the enum fields, for which I'll remove each initialized field in the operation block. In the end action, I will add to the dictionary all uninitialized fields with their values.

I don't have a better idea either. I am fine with your proposed approach as a workaround until your previous suggestion about callback with implicit default initialization is implemented in Roslyn.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
mavasani: Is this callback guaranteed to be atomic? If not, you may want to use a ConcurrentBag or something instead of a List.

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
Evangelink: https://github.com/dotnet/roslyn/issues/40811

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
Evangelink: I am not sure when reading the documentation so I'll go for a concurrent collection. Thanks for the remark.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
Evangelink: ping @mavasani

src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
Evangelink: I am not entirely sure to get what you would like.

Taking this code as an example:

```csharp
public enum MyEnum
{
    Value1 = 1,
    Value2 = 2 - 1,
    Value3 = Value1 * Value2,
}
```

We currently report 1 issue on `Value2` and 1 on `Value3`.

Are you suggesting to

1. Tweak the message to say that `Value2` duplicates `Value1`?

2. Report only once on `Value1` saying that `Value2` and `Value3` duplicates this field?

mavasani: Reporting once on each field seems reasonable. May be the diagnostic on `Value1` can say that Field `Value1` has the same constant value `x` as field `Value2`? Not required to implement, just a suggestion.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -0,0 +1,186 @@
...
+                    var onlyReferencesOneField = GetFilteredDescendants(fieldInitializerOperation, op => op.Kind != OperationKind.Binary)
+                        .OfType<IFieldReferenceOperation>()
+                        .Where(fro => fro.Field.ContainingType.Equals(enumSymbol))
+                        .Any();
+                    if (onlyReferencesOneField)
+                    {
+                        return;
+                    }
+
+                    invalidMembersByValue.AddOrUpdate(fieldInitializerOperation.Value.ConstantValue.Value,
```
mavasani: I wonder if `invalidMembersByValue` is the correct name as you seem to be adding all fields which either have no RHS or whose RHS references 0 or greater then 1 fields. Should be called just `membersByValue` instead?

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
mavasani: I think we don't want to encourage people to use the Children or Descendants API unless they really need to. You can possibly open an issue for it in Roslyn but it may not get traction.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
mavasani: I think location based sorting sounds the best approach.

mavasani:
@Evangelink You also have merge conflicts to resolve.

Evangelink:
@mavasani It might be nice to add `*.resx merge=union` in the `.gitattributes` file to help with merge issues on resource files.

Evangelink:
@mavasani let me know if there is anything else to do on my side.

PR #3118, opened 2019-12-18
Enable Spanish test leg

mavasani:


mavasani:
Spanish leg is failing as expected due to recently added compiler error message validation.

@sharwell feel free to fix these with a separate commit on this PR or opening a separate PR as convenient.

mavasani:
Had a chat with @sharwell offline and we are going to force merge this PR, so @sharwell can fix these in a follow-up PR.

PR #3119, opened 2019-12-18
Fix test failures revealed by Spanish run

sharwell:


mavasani:
@sharwell Seems like the tests are now all green!

PR #3125, opened 2019-12-19
Add nullability to shipped API files

jcouv:
This PR adds `?` annotations to public APIs.
Before we can add `!` annotations, we need the compiler to expose an internal flag. I'll get that going.

I didn't manage to update the reference to `CodeAnalysis.Common`, so couldn't use `SymbolDisplayMiscellaneousOptions)IncludeNullableReferenceTypeModifier` enum value.

Relates to https://github.com/dotnet/roslyn/issues/40502 (Shipped APIs files should include nullable annotations)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -21,7 +21,7 @@
 
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
-    <MicrosoftCodeAnalysisForShippedApisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisForShippedApisVersion>
+    <MicrosoftCodeAnalysisForShippedApisVersion>3.5.0-beta2-20056-01</MicrosoftCodeAnalysisForShippedApisVersion>
     <MicrosoftNetCompilersVersion>3.4.0-beta2-final</MicrosoftNetCompilersVersion>
     <DogfoodAnalyzersVersion>3.0.0-beta2.19529.2+e119d9cf</DogfoodAnalyzersVersion>
     <MicrosoftCodeAnalysisBannedApiAnalyzersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisBannedApiAnalyzersVersion>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -20,8 +20,8 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
 
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisVersion>
-    <MicrosoftNetCompilersVersion>3.5.0-beta2-19619-02</MicrosoftNetCompilersVersion>
+    <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
+    <MicrosoftNetCompilersVersion>3.4.0-beta2-final</MicrosoftNetCompilersVersion>
     <DogfoodAnalyzersVersion>3.0.0-beta2.19529.2+e119d9cf</DogfoodAnalyzersVersion>
     <MicrosoftCodeAnalysisBannedApiAnalyzersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisBannedApiAnalyzersVersion>
     <MicrosoftCodeAnalysisFXCopAnalyersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisFXCopAnalyersVersion>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -20,7 +20,8 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
 
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisForShippedApisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisForShippedApisVersion>
     <MicrosoftNetCompilersVersion>3.4.0-beta2-final</MicrosoftNetCompilersVersion>
     <DogfoodAnalyzersVersion>3.0.0-beta2.19529.2+e119d9cf</DogfoodAnalyzersVersion>
     <MicrosoftCodeAnalysisBannedApiAnalyzersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisBannedApiAnalyzersVersion>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -20,8 +20,8 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
 
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
-    <MicrosoftNetCompilersVersion>3.4.0-beta2-final</MicrosoftNetCompilersVersion>
+    <MicrosoftCodeAnalysisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisVersion>
+    <MicrosoftNetCompilersVersion>3.5.0-beta2-19619-02</MicrosoftNetCompilersVersion>
     <DogfoodAnalyzersVersion>3.0.0-beta2.19529.2+e119d9cf</DogfoodAnalyzersVersion>
     <MicrosoftCodeAnalysisBannedApiAnalyzersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisBannedApiAnalyzersVersion>
     <MicrosoftCodeAnalysisFXCopAnalyersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisFXCopAnalyersVersion>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -20,7 +20,7 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
 
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisVersion>
     <MicrosoftNetCompilersVersion>3.4.0-beta2-final</MicrosoftNetCompilersVersion>
     <DogfoodAnalyzersVersion>3.0.0-beta2.19529.2+e119d9cf</DogfoodAnalyzersVersion>
     <MicrosoftCodeAnalysisBannedApiAnalyzersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisBannedApiAnalyzersVersion>
--- a/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
+++ b/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
@@ -10,6 +10,7 @@ using System.Linq;
 using System.Threading;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Text;
 
@@ -50,16 +51,44 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
         }
 
 #pragma warning disable CA1815 // Override equals and operator equals on value types
+        private readonly struct NullableLine
+#pragma warning restore CA1815 // Override equals and operator equals on value types
+        {
+            public int Rank { get; }
+
+            internal NullableLine(int rank)
+            {
+                Rank = rank;
+            }
+        }
+
+#pragma warning disable CA1815 // Override equals and operator equals on value types
+        private struct ApiName
+#pragma warning restore CA1815 // Override equals and operator equals on value types
+        {
+            public string Name { get; }
+            public string NameWithNullability { get; }
+
+            public ApiName(string name, string nameWithNullability)
+            {
+                Name = name;
+                NameWithNullability = nameWithNullability;
+            }
+        }
+
+#pragma warning disable CA1815 // Override equals and operator equals on value types
         private readonly struct ApiData
 #pragma warning restore CA1815 // Override equals and operator equals on value types
         {
             public ImmutableArray<ApiLine> ApiList { get; }
             public ImmutableArray<RemovedApiLine> RemovedApiList { get; }
+            public ImmutableArray<NullableLine> NullableLines { get; }
 
-            internal ApiData(ImmutableArray<ApiLine> apiList, ImmutableArray<RemovedApiLine> removedApiList)
+            internal ApiData(ImmutableArray<ApiLine> apiList, ImmutableArray<RemovedApiLine> removedApiList, ImmutableArray<NullableLine> nullableLines)
             {
                 ApiList = apiList;
                 RemovedApiList = removedApiList;
+                NullableLines = nullableLines;
             }
         }
 
@@ -70,6 +99,7 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
 
             private readonly Compilation _compilation;
             private readonly ApiData _unshippedData;
+            private readonly bool _useNullability;
             private readonly ConcurrentDictionary<ITypeSymbol, bool> _typeCanBeExtendedCache = new ConcurrentDictionary<ITypeSymbol, bool>();
             private readonly ConcurrentDictionary<string, UnusedValue> _visitedApiList = new ConcurrentDictionary<string, UnusedValue>(StringComparer.Ordinal);
             private readonly IReadOnlyDictionary<string, ApiLine> _publicApiMap;
@@ -77,6 +107,7 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
             internal Impl(Compilation compilation, ApiData shippedData, ApiData unshippedData)
             {
                 _compilation = compilation;
+                _useNullability = shippedData.NullableLines.Length > 0;
                 _unshippedData = unshippedData;
 
                 var publicApiMap = new Dictionary<string, ApiLine>(StringComparer.Ordinal);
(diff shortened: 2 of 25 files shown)


eng/Versions.props
```diff
@@ -20,8 +20,8 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
 
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
-    <MicrosoftNetCompilersVersion>3.4.0-beta2-final</MicrosoftNetCompilersVersion>
+    <MicrosoftCodeAnalysisVersion>3.5.0-beta2-19619-02</MicrosoftCodeAnalysisVersion>
```
mavasani: I don't think we want to move all of the repo to MS.CA 3.5. Can we just move PublicApiAnalyzers to the newer version: https://github.com/dotnet/roslyn-analyzers/blob/master/src/PublicApiAnalyzers/Core/Analyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.csproj#L10?

Note this is required to allow us to release analyzer packages from the repo that work on VS2019 16+ from this branch.

mavasani: Also tagging @sharwell since he has made conscious efforts to keep the PublicApiAnalyzers on MS.CA 1.2.1.

mavasani: Do you need the new version just for the tests? Seems like the analyzer does not use any APIs? Also should nullable annotations be opt-in?

jcouv: I'll look at narrowing down the dependency on new package. It should indeed only be used for testing.

I think that nullability annotations should always be part of public API files because changing annotations can cause compat breaks. But I'm open to discussing this. 
What do you think?

jcouv:
As soon as we add the new display option (to display `!`) to the compiler API (PR https://github.com/dotnet/roslyn/pull/40519) and get a new package, I'll incorporate into this PR.

eng/Versions.props
sharwell: > Also tagging @sharwell since he has made conscious efforts to keep the PublicApiAnalyzers on MS.CA 1.2.1.

We should only update the dependency if both of the following hold:

1. We need to use an API from the newer version
2. We cannot use the newer API via reflection

sharwell: > Also should nullable annotations in public API files be opt-in?

I don't see any need to have this be an option. The API files won't change if NRT is disabled, and if NRT is enabled then it only makes sense to update the API files too.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
```diff
@@ -130,7 +132,8 @@ public sealed partial class DeclarePublicApiAnalyzer : DiagnosticAnalyzer
                     SymbolDisplayParameterOptions.IncludeName |
                     SymbolDisplayParameterOptions.IncludeDefaultValue,
                 miscellaneousOptions:
-                    SymbolDisplayMiscellaneousOptions.UseSpecialTypes);
+                    SymbolDisplayMiscellaneousOptions.UseSpecialTypes |
+                    (SymbolDisplayMiscellaneousOptions)IncludeNullableReferenceTypeModifier);
```
sharwell: ❓ What happens if the analyzer passes this flag to an older version of the compiler? If the input is validated to not contain extraneous flags, then we should conditionally include this option similar to the following:

```csharp
private static readonly SymbolDisplayMiscellaneousOptions IncludeNullableReferenceTypeModifier
    = Enum.TryParse(nameof(IncludeNullableReferenceTypeModifier), out SymbolDisplayMiscellaneousOptions value) ? value : SymbolDisplayMiscellaneousOptions.None;
```

src/PublicApiAnalyzers/UnitTests/Microsoft.CodeAnalysis.PublicApiAnalyzers.UnitTests.csproj
```diff
@@ -4,7 +4,7 @@
     <TargetFramework>net472</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersion)" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.5.0-beta2-19619-02" />
```
sharwell: ❕ We should still have this version number in the Versions.props file somehow.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
jcouv: I reviewed the usage of `MiscellaneousOptions` in the compiler code and I don't think there is any problem. We have a method to add a flag, another to remove a flag, a method to replace the whole set, and one to test the presence of a flag. All of these are safe in the presence of unanticipated flags.

eng/Versions.props
mavasani: > if NRT is enabled then it only makes sense to update the API files too.

This will mean an enforced breaking change on consumers of this analyzer. I am fine with going that route as long as it is a conscious decision.

jcouv: Yes, this addition to shipped API files would break folks who start using nullability and then upgrade to newer analyzer. If we're ever going to do this, it should be earlier rather than later (most projects probably not using nullability yet).

jcouv: @mavasani @tmat @sharwell Jared convinced me that making this change conditional would be preferable (as Manish suggested earlier), based on our own experience annotating compiler and IDE code.

I'm thinking that we can just use a marker `#nullable enable` in shipped API files to configure that. Sounds okay?

jcouv: I just want to record a good idea from @jasonmalinowski: we could later on add a `#nullable enable force` (or maybe `#nullable enable strict`) option, which would give an error if any oblivious public API is encountered.

jcouv:
I'll update this PR to also handle `!` shortly (once new compiler bits are available supporting the new display option).

eng/Versions.props
```diff
@@ -21,6 +21,7 @@
 
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisForShippedApisVersion>3.5.0-beta2-20056-01</MicrosoftCodeAnalysisForShippedApisVersion>
```
jcouv: @mavasani @sharwell I've update this PR to use the latest compiler package for testing (that package supports the new display option for `!`), but now I can't run tests any more (I tried the tests for the new analyzer, as well as an existing one).

Let me know if you have any clue what might be going on. I don't see the possible link between using a new compiler package and hitting a roslyn-SDK nuspec issue with packages for reference assemblies.

The error  is:
```
 Message: 
    NuGet.Packaging.Core.PackagingException : The package is missing the required nuspec file. Path: C:\Users\jcouv\AppData\Local\Temp\test-packages\Microsoft.NETFramework.ReferenceAssemblies.net472.1.0.0-preview.2
  Stack Trace: 
    PackageFolderReader.GetNuspecFile()
    PackageReaderBase.GetNuspec()
    PackageReaderBase.get_NuspecReader()
    PackageReaderBase.GetFrameworkItems()
    PackageReaderBase.GetFrameworkItemsAsync(CancellationToken cancellationToken)
    <ResolveCoreAsync>d__39.MoveNext() line 316
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter`1.GetResult()
    <ResolveAsync>d__38.MoveNext() line 168
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter`1.GetResult()
    <CreateSolutionAsync>d__71.MoveNext() line 692
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter`1.GetResult()
    <CreateProjectImplAsync>d__70.MoveNext() line 625
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter`1.GetResult()
    <CreateProjectAsync>d__69.MoveNext() line 603
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter`1.GetResult()
    <GetSolutionAsync>d__68.MoveNext() line 577
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter`1.GetResult()
    <GetSortedDiagnosticsAsync>d__65.MoveNext() line 491
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    ConfiguredTaskAwaiter.GetResult()
    <VerifyDiagnosticsAsync>d__55.MoveNext() line 197
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    ConfiguredTaskAwaiter.GetResult()
    <RunAsync>d__38.MoveNext() line 189
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter.GetResult()
    <VerifyAdditionalFileFixAsync>d__2.MoveNext() line 59
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter.GetResult()
    <VerifyCSharpAdditionalFileFixAsync>d__1.MoveNext() line 43
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter.GetResult()
    <AnnotatedMemberInAnnotatedShippedAPI>d__5.MoveNext() line 135
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    TaskAwaiter.ThrowForNonSuccess(Task task)
    TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

jcouv: Never mind. I deleted the local temp folder and retried. Tests are working again now :-)

src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
```diff
@@ -126,6 +126,15 @@
   <data name="DeclarePublicApiDescription" xml:space="preserve">
     <value>All public types and members should be declared in PublicAPI.txt. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
   </data>
+  <data name="AnnotatePublicApiTitle" xml:space="preserve">
+    <value>Annotate public types and members in the declared API</value>
+  </data>
+  <data name="AnnotatePublicApiMessage" xml:space="preserve">
+    <value>Symbol '{0}' is not annotated in the declared API.</value>
```
mavasani: Nit: Does the message need to have the term "nullable" or "nullability" somewhere for clarity?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
```diff
@@ -263,6 +298,16 @@ private static bool ValidateApiFiles(ApiData shippedData, ApiData unshippedData,
                 errors.Add(Diagnostic.Create(PublicApiFilesInvalid, Location.None, InvalidReasonShippedCantHaveRemoved));
             }
 
+            if (shippedData.NullableLines.Any(nl => nl.Rank != 0))
```
mavasani: Term `NullableLines` confused me into thinking these are lines which have nullable annotations. Given that you are using this only to determine if nullability annotations are enabled or not, can we instead just store a single int, say `NullabilityRank`, instead of all the lines enabling nullable? This should also allow you to report the `InvalidReasonMisplacedNullableEnable` diagnostics below.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -70,13 +99,15 @@ private sealed class Impl
...
             private readonly ApiData _unshippedData;
+            private readonly bool _useNullability;
             private readonly ConcurrentDictionary<ITypeSymbol, bool> _typeCanBeExtendedCache = new ConcurrentDictionary<ITypeSymbol, bool>();
             private readonly ConcurrentDictionary<string, UnusedValue> _visitedApiList = new ConcurrentDictionary<string, UnusedValue>(StringComparer.Ordinal);
             private readonly IReadOnlyDictionary<string, ApiLine> _publicApiMap;
 
             internal Impl(Compilation compilation, ApiData shippedData, ApiData unshippedData)
             {
                 _compilation = compilation;
+                _useNullability = shippedData.NullableLines.Length > 0;
```
mavasani: Instead of enforcing that nullable enable is allowed only in shipped files, would it better to allow it in either of the shipped/unshipped files?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -193,25 +225,36 @@ private void OnSymbolActionCore(ISymbol symbol, Action<Diagnostic> reportDiagnos
...
+                    if (!hasPublicApiEntryWithNullability)
+                    {
+                        var hasPublicApiEntryWithoutNullability = _publicApiMap.TryGetValue(publicApiName.Name, out foundApiLine);
+                        if (!hasPublicApiEntryWithoutNullability)
+                        {
+                            reportDeclareNewApi(symbol, isImplicitlyDeclaredConstructor, publicApiName.NameWithNullability);
+                        }
+                        else
+                        {
+                            reportAnnotateApi(symbol, isImplicitlyDeclaredConstructor, publicApiName, foundApiLine.IsShippedApi);
```
mavasani: I presume you have tested that FixAll works for this diagnostic? I think that would be very important functionality when enabling nullable annotations in public API files.

jcouv: It looks like we don't have test framework in place to test FixAll. I'll do some manual validation.
Please advise if there is a way to automate (I couldn't find any example in this solution nor in the roslyn-SDK that I could follow).

mavasani: Hmm I thought the below annotations did a FixAll: https://github.com/dotnet/roslyn-analyzers/blob/4578f20ea22fae2c340abca5bfad102afcf01a7d/src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs#L980-L1001

Does it instead just fix each diagnostic separately?

mavasani: I also wonder if for discoverability of the new functionality it would be good to report a new RS diagnostic on public APIs with nullable annotations in source but nullability not enabled in public API files, say "Enable nullable annotations in public API files", that adds nullable enable to shipped/unshipped file and then adds annotations to these files. If someone doesn't care about it, they can just disable the new RS diagnostic ID for their project.

jcouv: Hum, I'm not sure how to deploy this project into a live VS instance...

mavasani: Can you try setting `Microsoft.CodeAnalysis.PublicApiAnalyzers.Setup.csproj` as the startup project and F5?

jcouv: Regarding FixAll: then my test `AnnotatedMemberInAnnotatedShippedAPI` already covers this, and I may be lacking a non-FixAll test scenario instead ;-)

jcouv: Regarding adding `#nullable enable`. I love the idea. I'll add that shortly.

jcouv: Regarding FixAll, I found out that the test framework used here automatically tests FixAll whenever testing a single Fix too. I broke the FixAll implementation and it was detected.

jcouv:
@mavasani I've added a new diagnostic and corresponding fixer to add `#nullable enable` as suggested.
It only triggers when the public surface area includes some APIs that aren't oblivious. The new diagnostic is reported on those APIs.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: Regarding manual validation: the setup project didn't seem to work as far as I could tell.

jcouv:
> Should the fix add #nullable enable to both the shipped and unshipped files?

I don't think so. Only one is needed, so I think it should be as unobtrusive as possible. We could adjust later.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
mavasani: > I found out that the test framework used here automatically tests FixAll whenever testing a single Fix too

Ah great, I forgot we did enable that auto FixAll validation in unit tests.

mavasani: > Regarding manual validation: the setup project didn't seem to work as far as I could tell.

Interesting, guess we need to investigate why it did not work. Can you file a separate tracking issue?

mavasani:
Thanks @jcouv - changes LGTM.

I'd let @sharwell review your PR again for his signoff.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: Filed https://github.com/dotnet/roslyn-analyzers/issues/3153 for the setup issue.

jcouv:
@sharwell I'll wait for your sign-off. Let me know if otherwise. Thanks

jcouv:
@sharwell for review. Thanks

jcouv:
From offline chat with @sharwell the concern with version change was addressed, so good to merge.

PR #3147, opened 2020-01-06
Update CA1822: Do not report for web specific methods

### LINKED ISSUE #3123: Configure Severity for Asp.netcore Controller Action methods [CA1822: Mark members as static]
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/858561/configure-severity-for-aspnetcore-controller-actio.html)._

---
Hi,

If you install the FX Cop analyzer on an asp.net core project which contains controllers (e.g. web api template), you get this warning on Controller Actions. E.g.

[HttpGet]
public IEnumerable<WeatherForecast> Get() //member Get does not access instance data and can be marked as static

Is there a straightforward way to have the analizer ignore this rule JUST on controllers for instnance? E.g. via the editorconfig file.

Or better yet, have the analyzer itself ignore this rule on asp.net core controllers?

---
#### Visual Studio Feedback System on 12/18/2019, 01:32 AM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

#### Visual Studio Feedback System on 12/18/2019, 03:43 PM: 

<p>This issue is currently being investigated. Our team will get back to you if either more information is needed, a workaround is available, or the issue is resolved.</p>

---
### Original Solutions
(no solutions)

Evangelink:
Fix #3123 and also handles `Application_Start` and `Application_End` special methods.

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
@@ -6,6 +6,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
+using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
@@ -156,7 +157,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                 return false;
             }
 
-            if (IsSkippedMethod(methodSymbol))
+            if (IsSkippedMethod(methodSymbol, wellKnownTypeProvider))
             {
                 return false;
             }
@@ -284,7 +285,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
             return builder?.ToImmutable() ?? ImmutableArray<INamedTypeSymbol>.Empty;
         }
 
-        private static bool IsSkippedMethod(IMethodSymbol methodSymbol)
+        private static bool IsSkippedMethod(IMethodSymbol methodSymbol, WellKnownTypeProvider wellKnownTypeProvider)
         {
             if (methodSymbol.IsConstructor() || methodSymbol.IsFinalizer())
             {
@@ -293,7 +294,8 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
 
             if (methodSymbol.ReturnsVoid &&
                 methodSymbol.Parameters.IsEmpty &&
-                (methodSymbol.Name == "Application_Start" || methodSymbol.Name == "Application_End"))
+                (methodSymbol.Name == "Application_Start" || methodSymbol.Name == "Application_End") &&
+                methodSymbol.ContainingType.Inherits(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebHttpApplication)))
             {
                 return true;
             }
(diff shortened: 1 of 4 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -240,28 +242,63 @@ void Add(INamedTypeSymbol? symbol)
...
         }
+
+        private static bool IsSkippedMethod(IMethodSymbol methodSymbol)
+        {
+            if (methodSymbol.IsConstructor() || methodSymbol.IsFinalizer())
+            {
+                return true;
+            }
+
+            if (methodSymbol.ReturnsVoid &&
```
Evangelink: I could also check that the enclosing type is a controller.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -240,28 +242,63 @@ void Add(INamedTypeSymbol? symbol)
                 }
             }
 
-            Add(compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebServicesWebMethodAttribute));
+            // Web attributes
+            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebServicesWebMethodAttribute));
```
Evangelink: I could also check that the enclosing type is a controller.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
mavasani: I think it would be good to check the enclosing type is a controller for this case.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
mavasani: I think it is fine to not check enclosing type for the attribute case.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -227,7 +229,7 @@ private static bool IsExplicitlyVisibleFromCom(IMethodSymbol methodSymbol, Compi
             return false;
         }
 
-        private static ImmutableArray<INamedTypeSymbol> GetSkippedAttributes(Compilation compilation)
+        private static ImmutableArray<INamedTypeSymbol> GetSkippedAttributes(WellKnownTypeProvider wellKnownTypeProvider)
         {
             ImmutableArray<INamedTypeSymbol>.Builder? builder = null;
```
mavasani: Nit: can simplify this method by using the pooled `ArrayBuilder` instead.

Evangelink: I am not entirely sure about the usage, could you review the commit?

Evangelink:
@mavasani the extension methods already checks for null.

src/Microsoft.CodeQuality.Analyzers/Core/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: I didn't managed to use the pooled `ArrayBuilder` correctly. Am I supposed to replace 

```csharp
builder ??= ImmutableArray.CreateBuilder<INamedTypeSymbol>();
builder.Add(symbol);
```

with

```csharp
ArrayBuilder<INamedTypeSymbol>.GetInstance().Add(symbol);
```

Because when doing so the builder doesn't contain all the types.

Please let me know the proper use of this builder and I'll update this PR.

mavasani: Its fine to keep the current code, we can do a separate PR to audit all uses of ImmutableArray and ImmutableDictionary builders.

PR #3148, opened 2020-01-06
Update CA1710 to allow user configuration

### LINKED ISSUE #3065: Change the behavior of CA1710
Currently the rule [CA1710](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1710) will look at all base types and interfaces and might suggest a suffix based on an hard-coded dictionary mapping some base types to some suffix.

I would like to suggest a new behavior for this rule. I think the rule should look only at the base type or the interfaces declared (if no base type) and not look at a higher level. IMO there is no point going up in the hierarchy because each level is doing a specialization and is the only one that shall be respected. For example, in this case `public class MyDbReader : IDataReader`, it seems rather bad to suggest to the user to suffix the class with `Collection` (because `IDataReader` implements `IEnumerable`) or in this case `public class MyReadOnlyDictionary<TKey, TValue> : IReadOnlyDictionary<TKey, TValue>` to suggest `Collection`. This change won't prevent having some hard-coded mapping for the suffix but would solve some inconsistencies induced by the current implementation.

If we need to keep the compatibility with the current behavior for some users, we could add some configuration to let the user select the mode he wants.

In addition to the previous change, I'd like to suggest (maybe again through some configuration) an extension to this rule to not only suggest a prefix based on the hard-coded list but on the base type or first interface to enforce some specialization. For example, `public class MyReader : DbDataReader` would raise an issue suggesting the name to end with `DbDataReader`, `public class C : IDictionary, IEnumerable, IFoo` would suggest the class name to end with `Dictionary` (first interface).

WDYT?

Relates to/overrules issues #2955 and #1818.

Evangelink:
Allow user specific suffixes
Allow exclusion of indirect inheritance

Fix #3065


docs/Analyzer Configuration.md
```diff
@@ -408,4 +408,47 @@ Option Values:
...
+   // code omitted for simplicity
+}
+```
+
+### Additional required suffixes
+Option Name: `additional_required_suffixes`
+
+Configurable Rules: [CA1710](https://docs.microsoft.com/visualstudio/code-quality/ca1710-identifiers-should-have-correct-suffix)
+
+Option Values: List (separated by '|') of fully qualified type name followed by colon and the required suffix.
```
mavasani: I would prefer we keep re-using the existing supported formats for type name specification using documentation comment ID. See https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#excluded-type-names-with-derived-types.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -217,5 +218,42 @@ CategorizedAnalyzerConfigOptions ComputeCategorizedAnalyzerConfigOptions()
                 return CategorizedAnalyzerConfigOptions.Empty;
             }
         }
+
+        public static ImmutableArray<(INamedTypeSymbol type, string suffix)> GetAdditionalRequiredSuffixesOption(
```
mavasani: You should be able to replace this method with following once you switch over to using the existing supported type name specification format:

```csharp
public static SymbolNamesOption GetAdditionalRequiredSuffixesOption(
            this AnalyzerOptions options,
            DiagnosticDescriptor rule,
            Compilation compilation,
            CancellationToken cancellationToken)
            => options.GetSymbolNamesOption(EditorConfigOptionNames.AdditionalRequiredSuffixes, namePrefixOpt: "T:", rule, compilation, cancellationToken);

```

docs/Analyzer Configuration.md
Evangelink: I am not sure to get you... Are you suggesting to use this syntax for the type name but keep the `:SUFFIX` part?

mavasani: Yes, sorry I had forgotten that you need to have a suffix for each type name. Please use the same format for type name specification, though you would need to write your own custom parse method for each type name + suffix pair.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
mavasani: My suggestion will not work as you need to parse the suffix part specially.

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -25,33 +27,41 @@ private SymbolNamesOption(ImmutableHashSet<string> names, ImmutableHashSet<ISymb
...
+            foreach (var symbolName in symbolNames)
             {
-                if (name.Equals(".ctor", StringComparison.Ordinal) ||
-                    name.Equals(".cctor", StringComparison.Ordinal) ||
-                    !name.Contains(".") && !name.Contains(":"))
+                var parts = getSymbolNamePartsFunc != null
+                    ? getSymbolNamePartsFunc(symbolName)
+                    : new NameParts(symbolName);
+
+                if (parts.TypeName.Equals(".ctor", StringComparison.Ordinal) ||
```
mavasani: Should this be called `SymbolName` instead of `TypeName`?

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -85,13 +103,16 @@ public static SymbolNamesOption Create(ImmutableArray<string> symbolNames, Compi
...
+            return new SymbolNamesOption(namesBuilder.ToImmutableDictionaryAndFree(), symbolsBuilder.ToImmutableDictionaryAndFree());
         }
 
         public bool IsEmpty => ReferenceEquals(this, Empty);
 
         public bool Contains(ISymbol symbol)
-            => _symbols.Contains(symbol) || _names.Contains(symbol.Name);
+            => _symbols.ContainsKey(symbol) || _names.ContainsKey(symbol.Name);
+
+        public bool TryGetSuffix(ISymbol symbol, [NotNullWhen(true)] out string? suffix) =>
```
mavasani:
```suggestion
        public bool TryGetValue(ISymbol symbol, [NotNullWhen(true)] out string? value) =>
```

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -85,13 +103,16 @@ public static SymbolNamesOption Create(ImmutableArray<string> symbolNames, Compi
...
+            _symbols.TryGetValue(symbol, out suffix) || _names.TryGetValue(symbol.Name, out suffix);
```
mavasani:
```suggestion
            _symbols.TryGetValue(symbol, out value) || _names.TryGetValue(symbol.Name, out value);
```

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -101,13 +122,27 @@ public override bool Equals(object obj)
...
+
+        public sealed class NameParts
+        {
+            public NameParts(string typeName, string? suffix = null)
+            {
+                TypeName = typeName;
+                Suffix = suffix;
+            }
+
+            public string TypeName { get; }
```
mavasani:
```suggestion
            public string SymbolName { get; }
```

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -101,13 +122,27 @@ public override bool Equals(object obj)
         public bool Equals(SymbolNamesOption? other)
         {
             return other != null &&
-                _names.SetEquals(other._names) &&
-                _symbols.SetEquals(other._symbols);
+                _names.Count == other._names.Count &&
+                _symbols.Count == other._symbols.Count &&
+                _names.Keys.All(key => other._names.ContainsKey(key) && string.Equals(_names[key], other._names[key], StringComparison.Ordinal)) &&
```
mavasani: I thought we had some helper/extension method for dictionary equality. Can you check? If not, can you extract this into one?

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -101,13 +122,27 @@ public override bool Equals(object obj)
...
+                _symbols.Keys.All(key => other._symbols.ContainsKey(key) && string.Equals(_symbols[key], other._symbols[key], StringComparison.Ordinal));
         }
 
         public override int GetHashCode()
         {
             return HashUtilities.Combine(HashUtilities.Combine(_names), HashUtilities.Combine(_symbols));
         }
+
+        public sealed class NameParts
```
mavasani: Can you add a doc comment with an example of what this is used for?

src/Utilities/Compiler/Options/SymbolNamesOption.cs
mavasani: Please also add a doc comment that this method will fetch the string value associated with the given symbol in the option specification.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -163,13 +163,33 @@ static bool TryParseValue(string value, out ImmutableHashSet<TEnum> result)
...
+            this AnalyzerOptions options,
+            DiagnosticDescriptor rule,
+            Compilation compilation,
+            CancellationToken cancellationToken)
+        {
+            return options.GetSymbolNamesOption(EditorConfigOptionNames.AdditionalRequiredSuffixes, namePrefixOpt: "T:", rule, compilation, cancellationToken, GetParts);
+
+            static SymbolNamesOption.NameParts GetParts(string name)
+            {
+                var lastIndexOfColon = name.LastIndexOf(':');
```
mavasani: Looking at the examples under "Methods" section in https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format, it seems `:` can be present inside the doc comment ID for few symbol kinds. Can we choose a different separator character? Maybe `::` or `->`?

mavasani: You can probably also check that there is a single separator in the value, instead of using LastIndex, for a valid suffix.

mavasani: I think it would also be good to invoke Trim() on each of the name parts, probably in the constructor?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1142,18 +1142,386 @@ public class C : IReadOnlyCollection<int>
...
-            => VerifyCS.Diagnostic(isSpecial
-                    ? IdentifiersShouldHaveCorrectSuffixAnalyzer.SpecialCollectionRule
-                    : IdentifiersShouldHaveCorrectSuffixAnalyzer.DefaultRule)
+        [Theory, WorkItem(3065, "https://github.com/dotnet/roslyn-analyzers/issues/3065")]
+        [InlineData("")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = MyNamespace.Bar:Bar")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:MyNamespace.Bar:Bar")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = Buzz:ABC")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:Buzz:ABC")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = MyNamespace.Bar:Bar|MyNamespace.IFoo:Foo")]
```
mavasani: Would be good to avoid `Foo` and `Bar`

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1142,18 +1142,386 @@ public class C : IReadOnlyCollection<int>
...
+    End Class
+End Namespace"
+                    },
+                    AdditionalFiles = { (".editorconfig", editorConfigText)  },
+                }
+            }.RunAsync();
+        }
+
+        [Fact, WorkItem(3065, "https://github.com/dotnet/roslyn-analyzers/issues/3065")]
+        public async Task CA1710_HardcodedMappingWinsOverAdditionals()
```
mavasani: I think user option should win over the hard code mapping in the analyzer.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1142,18 +1142,386 @@ public class C : IReadOnlyCollection<int>
...
+                    }
+                }
+            }.RunAsync();
+        }
+
+        [Theory, WorkItem(3065, "https://github.com/dotnet/roslyn-analyzers/issues/3065")]
+        [InlineData("")]
+        [InlineData("dotnet_code_quality.CA1710.exclude_indirect_base_types = false")]
+        [InlineData("dotnet_code_quality.CA1710.exclude_indirect_base_types = true")]
+        [InlineData("dotnet_code_quality.CA1710.exclude_indirect_base_types = junk")]
```
mavasani:
```suggestion
        [InlineData("dotnet_code_quality.CA1710.exclude_indirect_base_types = invalid")]
```

docs/Analyzer Configuration.md
mavasani: Do we need to add documentation about the type name format before the separator character? Similar to https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#excluded-type-names-with-derived-types:

```
Allowed type name formats:

- Type name only (includes all types with the name, regardless of the containing type or namespace)
- Fully qualified names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format with an optional "T:" prefix.
```

docs/Analyzer Configuration.md
```diff
@@ -409,3 +409,46 @@ Option Values:
...
+
+Default Value: None
+
+Examples:
+
+| Option Value | Summary |
+| --- | --- |
+|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass:Class` | All types inheriting from MyClass are expected to have the 'Class' suffix |
+|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass:Class|MyNamespace.IFoo:Foo` | All types inheriting from MyClass are expected to have the 'Class' suffix AND all types implementing MyNamespace.IFoo are expected to have the 'Foo' suffix. |
```
mavasani: Please avoid `Foo`

docs/Analyzer Configuration.md
```diff
@@ -409,3 +409,48 @@ Option Values:
...
+   // code omitted for simplicity
+}
+```
+
+### Additional required suffixes
+Option Name: `additional_required_suffixes`
+
+Configurable Rules: [CA1710](https://docs.microsoft.com/visualstudio/code-quality/ca1710-identifiers-should-have-correct-suffix)
+
+Option Values: List (separated by '|') of type names with their required suffix (separated by '->').<br/>Allowed type name formats:
```
Evangelink: I had to use `<br/>` to force the line return to happen correctly in the markdown.

Evangelink: Is it clear enough? I couldn't really found a nice way to express the format.

src/Utilities/Compiler/Extensions/IDictionaryExtensions.cs
```diff
@@ -28,5 +29,9 @@ internal static class IDictionaryExtensions
                 dictionary.Add(item);
             }
         }
+
+        public static bool IsEqualTo<TKey, TValue>(this IReadOnlyDictionary<TKey, TValue> dictionary, IReadOnlyDictionary<TKey, TValue> other)
```
Evangelink: I have seen that for `ImmutableHashSet` the method was called `SetEquals` to I was wondering whether to call this `DictionaryEquals`. Let me know your preference.

Evangelink: I did something generic for the value so I have lost the ability to pass the string comparison. Is it fine like this or do you prefer an extra comparer argument?

mavasani: I think it is fine to add it when required.

docs/Analyzer Configuration.md
mavasani: Looks good to me.

Issue #2937, opened 2019-10-14
Proposed Analyzer: Leap Year Readiness


mj1856:
Per step 1 of the [guidelines](https://github.com/dotnet/roslyn-analyzers/blob/master/GuidelinesForNewRules.md), I'm filing an issue in advance of contributing a new analyzer.

My team at Microsoft is focused on Leap Year Readiness, as 2020 is a leap year.  (MSFT employees can visit https://aka.ms/leapyear to learn more.)  As part of this effort, we have built an analyzer for detecting anti-patterns for improper usage of `DateTime` and `DateTimeOffset` that can lead to reliability problems for services running on leap day.  

As an example, our analyzer can detect the following:

```csharp
// problematic code
DateTime result = new DateTime(dt.Year + 1, dt.Month, dt.Day);
```

This matters because when `dt1` is `2020-02-29`, trying to create a date of `2021-02-29` will throw an `ArgumentOutOfRangeException`.

The corrected code should be:

```csharp
// correct code
DateTime result = dt.AddYears(1).Date;
```

Another pattern our analyzer can detect is this:

```csharp
// problematic code
DateTime result = new DateTime(foo.Year, bar.Month, bar.Day);
```

This time, there's no arithmetic going on, but there's still the possibility of `bar` being a Feb 29th, and `foo` being in a year that doesn't have one.  Again, an `ArgumentOutOfRangeException` would be thrown.

There are a few ways to correct for this, but arguably the simplest is:

```csharp
// correct code
DateTime result = bar.AddYears(foo.Year - bar.Year).Date;
```

In all of the above examples, I show `.Date` being called to truncate time, as the original code did not carry through the hours, minutes, seconds, etc.  Also, the same patterns apply to `DateTimeOffset`.

At the moment, we don't have suggested code fixes implemented in the analyzer, and we are still improving our detection logic.  However, we would like to contribute our analyzers to the Roslyn Analyzers repo in their current state while we continue working on improving them.

Since they apply to specific APIs, per the guidelines it seems we shouold be contributing to the `Microsoft.NetCore.Analzyers` project.  I believe we are either in the "Reliability Warnings" or "Usage Warnings" category, but I'd appreciate advice as to which is more appropriate.

Thanks!

Evangelink:
@mj1856 Looks like the leapyear link is not publicly available. Would it be possible to update it? I am genuinely interested about this issue.

mj1856:
[...]

PR #3139, opened 2020-01-03
Refactor the repo into RoslynAnalyzers.sln and FxCopAnalyzers.sln

mavasani:
RoslynAnalyzers.sln has all the existing non-FxCop analyzers projects and a new group of `Microsoft.CodeAnalysis.NetAnalyzers` projects/package. Latter has all the merged CA analyzers from individual packages in FxCop analyzers package and will be used for the new .NET5 analyzers packaged into a single analyzer package. A follow-up change will adjust the default severities and enabled by default values for CA rules shipped as part of this package to be much more conservative then the current FxCopAnalyzers package.

FxCopAnalyzers.sln is the new solution for the existing FxCopAnalyzers package and all its subpackages - Microsoft.CodeQuality.Analyzers, Microsoft.NetCore.Analyzers and Microsoft.NetFramework.Analyzers. There are no changes to the contained CA rules in these packages, and even the default severities and enabled by default values will be preserved. This allows us to continue shipping the existing analyzer packages in the current structure until we are ready to completely migrate from FxCopAnalyzers package to Microsoft.CodeAnalysis.NetAnalyzers package.

**NOTE:** FxCopAnalyzers.sln is just present to generate current packages and VSIXes. RoslynAnalyzers.sln will be continue to be the primary solution for all the development in the repo. You should not need to make any changes/updates to FxCopAnalyzers.sln as part of your development workflow.

PR #3152, opened 2020-01-06
Add analyzers for leap year readiness CA2260&CA2261

kwilkins:
Changes associated with Proposed Analyzer: Leap Year Readiness #2937

Leap year date analyzers added
- DoNotIncrementYearValueInDateConstructor : CA2260
- DoNotUserYearValueFromDifferentDateInDateConstructor : CA2261

### DIFF
--- a/src/Microsoft.NetCore.Analyzers/Core/LeapYear/DoNotIncrementYearValueInDateConstructor.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/LeapYear/DoNotIncrementYearValueInDateConstructor.cs
@@ -63,7 +63,7 @@ namespace Microsoft.NetCore.Analyzers.LeapYear
                         // the date won't be a possible leap year.
                         if (!dateKindContext.AreMonthOrDayValuesSafe())
                         {
-                            if (this.YearIncrementIssueExists(dateKindContext))
+                            if (YearIncrementIssueExists(dateKindContext))
                             {
                                 if (dateKindContext.YearArgumentBinaryExpression != null)
                                 {
@@ -94,10 +94,10 @@ namespace Microsoft.NetCore.Analyzers.LeapYear
         /// Examines stored code analysis to see if there is a year increment issue.
         /// </summary>
         /// <returns>True if a year increment pattern was detected.</returns>
-        private bool YearIncrementIssueExists(DateKindContext dateKindContext)
+        private static bool YearIncrementIssueExists(DateKindContext dateKindContext)
         {
-            return this.IsBinaryExpressionAYearIncrementTriggeringPattern(dateKindContext.YearArgumentBinaryExpression)
-                || this.IsBinaryExpressionAYearIncrementTriggeringPattern(dateKindContext.YearArgumentIdentifierBinaryExpression);
+            return IsBinaryExpressionAYearIncrementTriggeringPattern(dateKindContext.YearArgumentBinaryExpression)
+                || IsBinaryExpressionAYearIncrementTriggeringPattern(dateKindContext.YearArgumentIdentifierBinaryExpression);
         }
 
         /// <summary>
@@ -105,14 +105,15 @@ namespace Microsoft.NetCore.Analyzers.LeapYear
         /// or subtraction(-) and neither of the operands are numeric literals greater
         /// than 1000.
         /// </summary>
-        private bool IsBinaryExpressionAYearIncrementTriggeringPattern(BinaryExpressionSyntax node)
+        private static bool IsBinaryExpressionAYearIncrementTriggeringPattern(BinaryExpressionSyntax? node)
         {
-            return (node.IsKind(SyntaxKind.AddExpression) || node.IsKind(SyntaxKind.SubtractExpression))
-                && !this.IsExpressionLiteralIntValueGreaterThanOrEqual(node.Left, 1000)
-                && !this.IsExpressionLiteralIntValueGreaterThanOrEqual(node.Right, 1000);
+            return node != null
+                && (node.IsKind(SyntaxKind.AddExpression) || node.IsKind(SyntaxKind.SubtractExpression))
+                && !IsExpressionLiteralIntValueGreaterThanOrEqual(node.Left, 1000)
+                && !IsExpressionLiteralIntValueGreaterThanOrEqual(node.Right, 1000);
         }
 
-        private bool IsExpressionLiteralIntValueGreaterThanOrEqual(ExpressionSyntax expressionSyntax, int limit)
+        private static bool IsExpressionLiteralIntValueGreaterThanOrEqual(ExpressionSyntax expressionSyntax, int limit)
         {
             return expressionSyntax.IsKind(SyntaxKind.NumericLiteralExpression)
                 && expressionSyntax is LiteralExpressionSyntax literal
--- a/src/Microsoft.NetCore.Analyzers/Core/LeapYear/DoNotIncrementYearValueInDateConstructor.cs
+++ b/src/Microsoft.NetCore.Analyzers/Core/LeapYear/DoNotIncrementYearValueInDateConstructor.cs
@@ -11,12 +11,12 @@ using Microsoft.CodeAnalysis.Diagnostics;
 namespace Microsoft.NetCore.Analyzers.LeapYear
 {
     /// <summary>
-    /// CA2300: Do not increment or decrement year parameter in DateTime constructor.
+    /// CA2260: Do not increment or decrement year parameter in DateTime constructor.
     /// </summary>
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     public sealed class DoNotIncrementYearValueInDateConstructor : DiagnosticAnalyzer
     {
-        internal const string RuleId = "CA2270";
+        internal const string RuleId = "CA2260";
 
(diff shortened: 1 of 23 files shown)


mavasani:
FYI: You will need to move the added analyzer files to a different folder and project, Microsoft.CodeAnalysis.NetAnalyzers`, once https://github.com/dotnet/roslyn-analyzers/pull/3165 is merged. See https://github.com/dotnet/roslyn-analyzers/pull/3139#issue-358842977 for details.

I will try to review this PR tomorrow. Thanks!

mj1856:
Thanks for the great work on this @kwilkins !

@mavasani - Would it be possible to somehow mark the rest of CA2262 - CA2269 range (or larger) as "reserved for future use"?  I can see that we might want to submit additional analyzers in this space in the future.  It would be good to keep them grouped sequentially if possible.  Thanks.

mavasani:
> Would it be possible to somehow mark the rest of CA2262 - CA2269 range (or larger) as "reserved for future use"?

@mj1856 I would recommend updating the ranges in https://github.com/dotnet/roslyn-analyzers/blob/cf0425bf67200496ba75d3ce7eac46a52ffc8e7c/src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt#L5-L15 to explicitly introduce a gap for the IDs you want to reserve, say `Usage: CA1801, CA1806, CA1816, CA2200-CA2209, CA2211-CA2262, CA2269-CA2299`. Please also add a comment about the rules that need that reserved range. Note that this text file follows a specific format and we have an analyzer that enforces the rule category to rule ID range mapping specified in this file. Let me know if you need help in updating this file appropriately to get your desired reservation.

src/Microsoft.NetCore.Analyzers/Core/LeapYear/DoNotIncrementYearValueInDateConstructor.cs
```diff
@@ -0,0 +1,123 @@
...
+            isEnabledByDefault: false,
+            helpLinkUri: null);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule, IdentifierRule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+            context.EnableConcurrentExecution();
+            context.RegisterSemanticModelAction(semanticModelAnalysisContext =>
```
mavasani: Please don't register this action - there are performance issues associated with it. I think you need to move to using IOperation API and RegisterOperationAction for all semantic analysis of executable code.

src/Microsoft.NetCore.Analyzers/UnitTests/LeapYear/DoNotIncrementYearValueInDateConstructorTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            GetCSharpResultAt(23, 37, "a + 908"),
+            GetCSharpResultAt(29, 34, "24 + i"),
+            GetCSharpResultAt(30, 27, "orig.Year + 31"),
+            GetCSharpResultAt(31, 27, "orig.Year + 32"),
+            GetCSharpResultAt(34, 32, "orig.Year + 101"),
+            GetCSharpResultAt(35, 57, "orig.Year - 13"));
+        }
+
+        [Fact]
+        public async Task YearIncrementVariableCases()
```
mavasani: Tagging @dotpaul 

Looking at this test case, it seems like you essentially want to do a version of TaintedDataFlowAnalysis, where an expression is considered tainted if it is a binary operation with one of the operands as a year/month/day of a DateTime variable and other operand is a literal. We already have large number of Taint analysis based rules in the repo, and I think you should build on top of that framework instead of attempting to write your pseudo-dataflow analysis (done in DateKindSemanticModel.cs and DateKindContext.cs at start of this PR).

@dotpaul Your thoughts? If you agree, might be good for you guys to sync in person.

mavasani: Note that if you don't want to handle the DFA cases in this test, and just want the cases in prior tests to work, which might be a reasonable place to start, then you can write an extremely simple IOperation based analyzer (possible 20-30 lines) that can handle those cases. I'd let you guys decide if you want the simpler analyzer that just analyzes arguments or a dataflow analysis based analyzer.

mavasani: As an example, consider the below SQL injection test based on TaintedDataFlowAnalysis: https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/ReviewCodeForSqlInjectionVulnerabilitiesTests.cs#L1924-L1962

It has a variable holding tainted data below as the string has a concatenation from an unknown source: https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/ReviewCodeForSqlInjectionVulnerabilitiesTests.cs#L1946

This tainted value gets passed to a method call as argument: https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/ReviewCodeForSqlInjectionVulnerabilitiesTests.cs#L1948

Then it gets passed to the SqlCommand constructor in the invoked method: https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/ReviewCodeForSqlInjectionVulnerabilitiesTests.cs#L1953-L1957

The constructor gets flagged by this unit test.

IMO, this test shows how the analysis required for this analyzer closely matches the one done for SQL injection analyzer and can re-use almost all of the taint analysis dataflow framework.

dotpaul: TaintedDataAnalysis would need a bit of modification to treat integer values that get modified as a source.

I think the larger question is, are there other cases would we want to handle that dataflow analysis would help with it? E.g. dunno if something like this would be a case worth trying to find:

```csharp
int y = originalDateTime.Year + 1;
int year = y;
DateTime newDateTime = new DateTime(year, originalDateTime.Month, originalDateTime.Day);
```

(There's also tracking the month and day values for determining if it's February 29, so that's another place where dataflow analysis might be helpful [but even more work].)

But all that being said, dataflow analysis isn't easy, so unless there are harder cases you really want to catch and you have time to work on it, I'd avoid it for now.

mavasani: @dotpaul - are you suggesting that the analyzer be currently written as a simple IOperation analyzer that just flags violations where arithmetic is done in the callsite argument and avoid any DFA based cases? I think that is indeed a reasonable starting point and I am fine with that approach. I am lot more concerned with us attempting to write another pseudo-dataflow analysis solution, which should be avoided.

dotpaul: I'm saying I'll defer to @kwilkins as what to do for DFA. :) I dunno if tainted data analysis will meet all the needs. Agree it'd be better to avoid pseudo-DFA. I'd probably consider starting an entirely new analysis myself, given I had the time, but I haven't looked at all the cases.

mj1856:
@mavasani - Thanks for the feedback.  While we agree, unfortunately we do not have resources to rewrite this to use tainted dataflow analysis in time for it to be released and usable before leap day.

We will still consider working on this to prevent leap year bugs over the long term, as we will have another leap year in 2024, 2028, etc.

mavasani:
@mj1856 That sounds reasonable. Would you be willing to simplify the analyzer as per https://github.com/dotnet/roslyn-analyzers/pull/3152/files#r365386559, so you don't perform any dataflow analysis at all, i.e. the first version of analyzer only supports analyzing values at arguments to call sites, but not flowing data values across statements? If so, you can delete large portion of the code being added in this PR, and then when you are ready to enhance the analyzer with dataflow analysis, you can work with @dotpaul to come with an appropriate DFA implementation.

mavasani:
Note about the merge conflicts: You will need to move your code to under `https://github.com/dotnet/roslyn-analyzers/tree/master/src/NetAnalyzers/Core`. The repo has been refactored since this PR was created.

mavasani:

mavasani:
Is this PR still actively being worked upon? If not, can we close it and reopen once ready?

mattjohnsonpint:
It is not being worked on presently.   The need still exists though. 

You can close this.  I hope to (at some point) make another one that does dataflow analysis.

PR #3154, opened 2020-01-07
Update CA1725 fixer to really fix all instances of the diagnostic

### LINKED ISSUE #1402: CA1725 FixAll is not helpful
#### Analyzer package

#### Analyzer

ParameterNamesShouldMatchBaseDeclarationFixer

#### Repro steps

```
interface I
{
    void M(int x1);
    void M2(int x2);
}

class C : I
{
    public void M(int y1) // CA1725 here
    {
    }

    public void M2(int y2) // CA1725 here
    {
    }
}
```

#### Expected Behavior
FixAll for CA1725 fixes all CA1725 instances

#### Actual Behavior
FixAll only fixes instances where parameter name matches the instance on which FixAll was triggered. This means one has to manually fix almost all the violations.

[This line](https://github.com/dotnet/roslyn-analyzers/blob/6ff9c17d8fc05fae8b1c48b7a4d25b1e080f0b97/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/ParameterNamesShouldMatchBaseDeclaration.Fixer.cs#L50) needs to be changed to not include `newName` in the equivalence key.

Evangelink:
Fix #1402

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ParameterNamesShouldMatchBaseDeclarationTests.Fixer.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ParameterNamesShouldMatchBaseDeclarationTests.Fixer.cs
@@ -49,7 +49,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.UnitTests
                               }",
                     },
                 },
-                NumberOfFixAllIterations = 2,
+                NumberOfFixAllIterations = 1,
             }.RunAsync();
 
             await new VerifyCS.Test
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/ParameterNamesShouldMatchBaseDeclarationTests.Fixer.cs
```diff
@@ -49,7 +49,7 @@ public class TestClass : BaseClass
                               }",
                     },
                 },
-                NumberOfFixAllIterations = 2,
+                NumberOfFixAllIterations = 1,
```
sharwell: 📝 This is the default value for this property. Now that the analyzer fixes everything in one pass, this property can be left at the default value and the test can potentially switch back to the simple form.

PR #3155, opened 2020-01-07
Update CA1801 to losen event handler recognition for UWP

### LINKED ISSUE #3106: CA1801 is flagged for methods used as event handlers
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

[CA1801](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1801?view=vs-2019)

#### Repro steps

1. Create a default XAML Win10 app.
2. Add a Page_Loaded event handler to the Page derived class like this:

```
private void Page_Loaded(object sender, RoutedEventArgs e)
{
    // ... do stuff without referencing sender or e
}

```
#### Expected behavior
Event handlers with mandatory signature are not flagged for unused variables.

#### Actual behavior

Seems like same issue as [False CA1801 reports (remove unused parameter) for methods used as event handlers #1218](https://github.com/dotnet/roslyn-analyzers/issues/1218) but it is still being flagged for event handlers...

Evangelink:
Fix #3106

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/ReviewUnusedParameters.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/ReviewUnusedParameters.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/ReviewUnusedParameters.cs
```diff
@@ -150,7 +151,8 @@ public override void Initialize(AnalysisContext context)
             if (eventsArgSymbol != null &&
                 method.Parameters.Length == 2 &&
                 method.Parameters[0].Type.SpecialType == SpecialType.System_Object &&
-                method.Parameters[1].Type.Inherits(eventsArgSymbol))
+                // UWP has specific EventArgs not inheriting from System.EventArgs. It was decided to go for a suffix match rather than a whitelist.
+                (method.Parameters[1].Type.Inherits(eventsArgSymbol) || method.Parameters[1].Type.Name.EndsWith("EventArgs", StringComparison.Ordinal)))
```
mavasani: I think you need to move the `eventsArgSymbol != null` check here:
```suggestion
                (eventsArgSymbol != null && method.Parameters[1].Type.Inherits(eventsArgSymbol) || method.Parameters[1].Type.Name.EndsWith("EventArgs", StringComparison.Ordinal)))
```

mavasani: Or you can just remove `eventsArgSymbol != null` completely given that Inherits extension method already does a null check.

Evangelink: Good catch!

PR #3156, opened 2020-01-07
Update CA1707 to ignore global.asax special methods

### LINKED ISSUE #3121: CA1707 should not trigger on MSFT required naming conventions
...like all the methods that would be in a Global.asax. Changing the names would break functionality, and we shouldn't have to suppress them all the time. I don't have a computer handy at the moment, but I imagine win/web form control event handlers like `btnOk_Click` would be triggering this as well.

Analyzers that complain about these default or required conventions feel unfinished and are more likely to be uninstalled or deactivated as fast as they were installed.

Evangelink:
Fix #3121 

Following comment is still open https://github.com/dotnet/roslyn-analyzers/issues/3121#issuecomment-571537124


mavasani:
@Evangelink Can you please resolve the merge conflicts?

Evangelink:
@mavasani Let me know if you want me to handle the test and event exclusions.

mavasani:
@Evangelink Lets wait for more feedback for test and event handler methods.

PR #3157, opened 2020-01-07
CA2231: Add test for fix-all

### LINKED ISSUE #1381: Fixer for CA2231 has no FixAll support
#### Analyzer package

#### Analyzer

OverloadOperatorEqualsOnOverridingValueTypeEquals.Fixer.cs

Evangelink:
Fix #1381 

Please let me know if this test isn't the proper way to test the fix-all.

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/OverloadOperatorEqualsOnOverridingValueTypeEqualsTests.Fixer.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/OverloadOperatorEqualsOnOverridingValueTypeEqualsTests.Fixer.cs
@@ -82,5 +82,123 @@ Public Structure A
 End Structure
 ");
         }
+
+        [Fact]
+        public async Task CA2231_CSharp_FixAll()
+        {
+            await new VerifyCS.Test
+            {
+                TestCode = @"
+using System;
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeQuality.Analyzers/UnitTests/ApiDesignGuidelines/OverloadOperatorEqualsOnOverridingValueTypeEqualsTests.Fixer.cs
```diff
@@ -82,5 +82,123 @@ End Operator
...
+    {
+        return left.Equals(right);
+    }
+
+    public static bool operator !=(B left, B right)
+    {
+        return !(left == right);
+    }
+}",
+                NumberOfFixAllIterations = 1
```
sharwell: 📝 Fix All is always tested, and `1` is the default value for this property. Therefore:

* The name of this test is a bit redundant. It would be better to call it `CA2231_CSharp_MultipleViolations`
* The method can use the simpler form `await VerifyCS.VerifyCodeFixAsync`

Evangelink: Thank you for the clarification! I'll create a cleanup PR.

PR #3158, opened 2020-01-07
Increase timeouts for test runs

mavasani:


Evangelink:
This will be nice, thank you!

mavasani:
@sharwell Can you please do a quick review? We have been seeing this very frequently in the PRs now. I will merge to master as soon as this goes into 2.9.x

sharwell:
This is likely caused (or exacerbated) by the new `ReferenceAssemblies`/NuGet support in the analyzer test library. It could potentially be improved by a two-pass package resolution approach, where offline packages are used to attempt to resolve a package, and online packages are only used if the offline resolution fails.

Evangelink:
We had this exact issue when we implemented our own nuget package downloader for tests and went for the two-pass approach.

sharwell:
@Evangelink if you have some ideas for how to implement it I'd love to hear it 😄 It took a long time to get the current form working and I'm not sure where the changes would need to be made.

Evangelink:
Well we _hacked_ something would be a more accurate version xD

[NugetMetadataFactory](https://github.com/SonarSource/sonar-dotnet/blob/1c6600cf2d9f94c83c69ae9136c7732f18aaf1e9/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestFramework/NugetMetadataFactory.cs) is responsible for the check and download. [NuGetMetadataReference](https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/NuGetMetadataReference.cs) is used to list the packages we want to access.

It worked but IMO that's far from being optimal or even a good solution.

But I'd be happy to have a look and see if I can think of something better, and if needed I am available for some chat/skype.

PR #3163, opened 2020-01-07
RS1030: Do not use Compilation.GetSemanticModel

### LINKED ISSUE #3114: Meta analyzer to flag use of Compilation.GetSemanticModel inside diagnostic analyzers
We have seen tons of performance issues in sub-optimally written analyzers that end up creating completely new semantic models by invoking `Compilation.GetSemanticModel`. This leads the compiler to rebind the entire file, and doing it multiple times (say with SyntaxNode callbacks) can be a performance hog. For example, https://github.com/dotnet/roslyn-analyzers/issues/2576 identified this can lead a 4x-5x slowdown. With nullability flow analysis in place inside the compiler, this can be even worse.

We should write an analyzer that flags use of this API inside a diagnostic analyzer. Implementation of this analyzer should potentially be a sub-type of [DiagnosticAnalyzerCorrectnessAnalyzer](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.cs).

Evangelink:
Fix #3114


src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -398,4 +398,13 @@
...
   </data>
+  <data name="DoNotUseCompilationGetSemanticModelDescription" xml:space="preserve">
+    <value>This call leads the compiler to rebind the entire file, and doing it multiple times (say with SyntaxNode callbacks) can be a performance hog.</value>
+  </data>
+  <data name="DoNotUseCompilationGetSemanticModelMessage" xml:space="preserve">
+    <value>This call can lead to performance hog and should be used carefully.</value>
+  </data>
+  <data name="DoNotUseCompilationGetSemanticModelTitle" xml:space="preserve">
+    <value>Do not invoke Compilation.GetSemanticModel() method.</value>
+  </data>
```
Evangelink: I am not happy with those messages.

Also, I am not sure whether the rule is intended to be a strict "do not use" or more of a "review use of".

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.DoNotUseCompilationGetSemanticModelTitle), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.DoNotUseCompilationGetSemanticModelMessage), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.DoNotUseCompilationGetSemanticModelDescription), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources));
+
+        public static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticIds.DoNotUseCompilationGetSemanticModelRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.MicrosoftCodeAnalysisCorrectness,
+            DiagnosticHelpers.DefaultDiagnosticSeverity,
```
Evangelink: Shall it be a `Message` severity by default?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+            isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
```
Evangelink: Shall it be deactivated by default?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            base.Initialize(context);
+        }
+
+        protected override DiagnosticAnalyzerSymbolAnalyzer? GetDiagnosticAnalyzerSymbolAnalyzer(CompilationStartAnalysisContext compilationContext, INamedTypeSymbol diagnosticAnalyzer, INamedTypeSymbol diagnosticAnalyzerAttribute)
+        {
+            if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisCompilation, out var compilationType))
```
Evangelink: I decided to force having a reference to at least `Microsoft.CodeAnalysis`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                var invocation = (IInvocationOperation)oac.Operation;
+
+                if (invocation.TargetMethod.Name.Equals("GetSemanticModel", StringComparison.Ordinal) &&
+                    invocation.TargetMethod.Parameters.Length == 2 &&
+                    (
+                        invocation.TargetMethod.ContainingType.Equals(compilationType) ||
+                        invocation.TargetMethod.ContainingType.Equals(csharpCompilation) ||
+                        invocation.TargetMethod.ContainingType.Equals(visualBasicCompilation)
+                    ))
```
Evangelink: Is it enough or shall I also test the arguments type?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -398,4 +398,13 @@
   <data name="DoNotUseReservedDiagnosticIdTitle" xml:space="preserve">
     <value>Do not use reserved diagnostic IDs.</value>
   </data>
+  <data name="DoNotUseCompilationGetSemanticModelDescription" xml:space="preserve">
+    <value>This call leads the compiler to rebind the entire file, and doing it multiple times (say with SyntaxNode callbacks) can be a performance hog.</value>
```
mavasani: I think we need a more refined comment here. @333fred @sharwell do you have a suggestion on how to word this correctly?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -398,4 +398,13 @@
...
+  </data>
+  <data name="DoNotUseCompilationGetSemanticModelMessage" xml:space="preserve">
+    <value>This call can lead to performance hog and should be used carefully.</value>
```
mavasani: Same here, needs a more refined message.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
mavasani: No worries, we will get the wording figured out. Thanks for the contribution!

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
mavasani: I think this is fine - we do want to warn users to avoid using this API.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            var csharpCompilation = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisCSharpCSharpCompilation);
+            var visualBasicCompilation = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisVisualBasicVisualBasicCompilation);
+
+            compilationContext.RegisterOperationAction(oac =>
+            {
+                var invocation = (IInvocationOperation)oac.Operation;
+
+                if (invocation.TargetMethod.Name.Equals("GetSemanticModel", StringComparison.Ordinal) &&
+                    invocation.TargetMethod.Parameters.Length == 2 &&
```
mavasani: I think we can avoid the parameter length check here. In case we add new overloads to this API, we still want to flag their usage.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
mavasani: I think just the containing type check is sufficient.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            base.Initialize(context);
+        }
+
+        protected override DiagnosticAnalyzerSymbolAnalyzer? GetDiagnosticAnalyzerSymbolAnalyzer(CompilationStartAnalysisContext compilationContext, INamedTypeSymbol diagnosticAnalyzer, INamedTypeSymbol diagnosticAnalyzerAttribute)
```
mavasani: Your current implementation will flag invocation of this API even outside a DiagnosticAnalyzer type. Code in the symbol action callback of DiagnosticAnalyzerSymbolAnalyzer is the one that reduces the analysis to just inside diagnostic analyzer type: https://github.com/dotnet/roslyn-analyzers/blob/09f25eda649579b875952673cbdd3c8997b07ab0/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.DiagnosticAnalyzerSymbolAnalyzer.cs#L27-L34

I think a correct approach here is to create a new type `DiagnosticAnalyzerSymbolStartAnalyzer` with methods `internal void AnalyzeSymbolStart(SymbolStartAnalysisContext symbolStartContext)` and `protected abstract void AnalyzeDiagnosticAnalyzer(SymbolStartAnalysisContext symbolStartContext);` analogous to the ones in `DiagnosticAnalyzerSymbolAnalyzer.cs`. You probably want to share the rest of the code between `DiagnosticAnalyzerSymbolAnalyzer` and `DiagnosticAnalyzerSymbolStartAnalyzer`. Then you can register the operation action on the symbol start context inside the override `AnalyzeDiagnosticAnalyzer(SymbolStartAnalysisContext symbolStartContext)`.

I will also be fine if you think it is better to get rid of `DiagnosticAnalyzerCorrectnessAnalyzer` type completely and extract some of its contents to just helper methods.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DoNotUseCompilationGetSemanticModelTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+class Analyzer : DiagnosticAnalyzer
+{
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => throw null;
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.RegisterCompilationStartAction(csac =>
+        {
+            csac.Compilation.GetSemanticModel(null);
```
mavasani: Can you add a test where this API is invoked outside a DiagnosticAnalyzer and ensure we don't flag it?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.cs
```diff
@@ -1,48 +1,13 @@
...
-        internal static readonly string SymbolAnalysisContextFullName = typeof(SymbolAnalysisContext).FullName;
-        internal static readonly string SyntaxNodeAnalysisContextFullName = typeof(SyntaxNodeAnalysisContext).FullName;
-        internal static readonly string SyntaxTreeAnalysisContextFullName = typeof(SyntaxTreeAnalysisContext).FullName;
-        internal static readonly string CodeBlockStartAnalysisContextFullName = typeof(CodeBlockStartAnalysisContext<>).FullName;
-        internal static readonly string CodeBlockAnalysisContextFullName = typeof(CodeBlockAnalysisContext).FullName;
-        internal static readonly string OperationBlockStartAnalysisContextFullName = typeof(OperationBlockStartAnalysisContext).FullName;
-        internal static readonly string OperationBlockAnalysisContextFullName = typeof(OperationBlockAnalysisContext).FullName;
-        internal static readonly string OperationAnalysisContextFullName = typeof(OperationAnalysisContext).FullName;
-        internal static readonly string SymbolKindFullName = typeof(SymbolKind).FullName;
```
Evangelink: Moved to `WellKnownTypeNames`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.cs
```diff
@@ -1,48 +1,13 @@
...
-        internal const string RegisterOperationBlockStartActionName = nameof(AnalysisContext.RegisterOperationBlockStartAction);
-        internal const string RegisterOperationBlockEndActionName = nameof(OperationBlockStartAnalysisContext.RegisterOperationBlockEndAction);
-        internal const string RegisterOperationBlockActionName = nameof(AnalysisContext.RegisterOperationBlockAction);
-        internal const string RegisterOperationActionName = nameof(AnalysisContext.RegisterOperationAction);
-        internal const string RegisterCompilationStartActionName = nameof(AnalysisContext.RegisterCompilationStartAction);
-        internal const string RegisterCompilationEndActionName = nameof(CompilationStartAnalysisContext.RegisterCompilationEndAction);
-        internal const string RegisterCompilationActionName = nameof(AnalysisContext.RegisterCompilationAction);
-        internal const string ReportDiagnosticName = nameof(CompilationAnalysisContext.ReportDiagnostic);
-        internal const string SupportedDiagnosticsName = nameof(SupportedDiagnostics);
-        internal const string TLanguageKindEnumName = @"TLanguageKindEnum";
```
Evangelink: Extracted into a helper class.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,73 @@
...
+                    {
+                        obsac.RegisterOperationAction(oac =>
+                        {
+                            var invocation = (IInvocationOperation)oac.Operation;
+
+                            if (invocation.TargetMethod.Name.Equals("GetSemanticModel", StringComparison.Ordinal) &&
+                                (
+                                    invocation.TargetMethod.ContainingType.Equals(compilationType) ||
+                                    invocation.TargetMethod.ContainingType.Equals(csharpCompilation) ||
+                                    invocation.TargetMethod.ContainingType.Equals(visualBasicCompilation)
```
Evangelink: Do you have any preference between these 3 OR conditions and a local method?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Helpers/DiagnosticWellKnownNames.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers.Helpers
+{
+    internal static class DiagnosticWellKnownNames
```
Evangelink: Extracted out from `DiagnosticAnalyzerCorrectnessAnalyzer`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
mavasani: I think this is fine.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
mavasani: How about
- Title/Message: `Do not invoke Compilation.GetSemanticModel() method within a diagnostic analyzer`
- Description: `'GetSemanticModel' is an expensive method to invoke within a diagnostic analyzer because it creates a completely new semantic model, which does not share compilation data with the compiler or other analyzers. This incurs an additional performance cost during semantic analysis. Instead, consider registering a different analyzer action which allows used of a shared 'SemanticModel', such as 'RegisterOperationAction', 'RegisterSyntaxNodeAction', or 'RegisterSemanticModelAction'`.

mavasani:
@Evangelink there are still merge conflicts.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,73 @@
...
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(csac =>
```
sharwell:
```suggestion
            context.RegisterCompilationStartAction(context =>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (!wellKnownTypeProvider.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisDiagnosticsDiagnosticAnalyzer, out var diagnosticAnalyzerType) ||
+                    !wellKnownTypeProvider.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisCompilation, out var compilationType))
+                {
+                    return;
+                }
+
+                var csharpCompilation = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisCSharpCSharpCompilation);
+                var visualBasicCompilation = wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisVisualBasicVisualBasicCompilation);
+
+                csac.RegisterOperationBlockStartAction(obsac =>
```
sharwell:
```suggestion
                csac.RegisterOperationBlockStartAction(context =>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DoNotUseCompilationGetSemanticModelAnalyzer.cs
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    if (obsac.OwningSymbol is IMethodSymbol methodSymbol &&
+                        methodSymbol.ContainingType.Inherits(diagnosticAnalyzerType))
+                    {
+                        obsac.RegisterOperationAction(oac =>
```
sharwell:
```suggestion
                        obsac.RegisterOperationAction(context =>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
mavasani: @Evangelink I think we should be good to merge once you update the diagnostic resource strings.

PR #3169, opened 2020-01-08
CA1021: Avoid out parameters

### LINKED ISSUE #377: Port FxCop rule CA1021: AvoidOutParameters
**Title:** Avoid out parameters

**Description:**

Passing types by reference (using out or ref) requires experience with pointers, understanding how value types and reference types differ, and handling methods with multiple return values. Also, the difference between out and ref parameters is not widely understood.

**Dependency:** None

**Notes:**

Evangelink:
Fix #377

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidOutParameters.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidOutParameters.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AvoidOutParameters : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1021";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidOutParametersTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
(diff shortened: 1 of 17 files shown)


src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,71 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            analysisContext.RegisterCompilationStartAction(csac =>
+            {
+                var outAttributeType = csac.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesOutAttribute);
```
Evangelink: Maybe it's better if I do this call directly in the `IsOutParameter` as the value is cached.

src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,71 @@
...
+                        sac.ReportDiagnostic(methodSymbol.CreateDiagnostic(Rule));
+                    }
+                }, SymbolKind.Method);
+            });
+        }
+
+        private static bool IsOutParameter(IParameterSymbol parameterSymbol, INamedTypeSymbol? outAttributeSymbol) =>
+            parameterSymbol.RefKind == RefKind.Out ||
+            // Handle VB.NET special case for out parameters
+            (parameterSymbol.RefKind == RefKind.Ref && parameterSymbol.HasAttribute(outAttributeSymbol));
```
Evangelink: I couldn't find any better way for handling this case. It seems that there is an `IsOut` property on the class itself but that's not available through the interface.

Evangelink: FYI `HasAttribute` handles null parameter.

src/Microsoft.CodeQuality.Analyzers/Core/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1352,4 +1352,13 @@
...
   </data>
+  <data name="AvoidOutParametersDescription" xml:space="preserve">
+    <value>Passing types by reference (using 'out' or 'ref') requires experience with pointers, understanding how value types and reference types differ, and handling methods with multiple return values. Also, the difference between 'out' and 'ref' parameters is not widely understood.</value>
+  </data>
+  <data name="AvoidOutParametersMessage" xml:space="preserve">
+    <value>Avoid 'out' parameters as they are not designed for general audience.</value>
+  </data>
+  <data name="AvoidOutParametersTitle" xml:space="preserve">
+    <value>Avoid out parameters</value>
+  </data>
```
Evangelink: Shall we reuse the fxcop messages? If so are they available somewhere or do I need to run the old fxcop to get them?

PR #3173, opened 2020-01-08
UseNameofInPlaceOfString: Remove redundant null-check

### LINKED ISSUE #3023: UseNameofInPlaceOfString: Add coverage for early return
📝 We should make sure this early return line of the analyzer is still covered by tests:

_Originally posted by @sharwell in https://github.com/dotnet/roslyn-analyzers/pull/3005_

Evangelink:
Fix #3023

### DIFF
--- a/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/UseNameofInPlaceOfString.cs
+++ b/src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/UseNameofInPlaceOfString.cs
@@ -61,11 +61,6 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
                 return;
             }
 
-            if (argument.Parameter == null)
-            {
-                return;
-            }
-
             var stringText = (string)argument.Value.ConstantValue.Value;
 
             var matchingParameter = argument.Parameter;


src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/UseNameofInPlaceOfString.cs
```diff
@@ -61,11 +61,6 @@ private void AnalyzeArgument(OperationAnalysisContext context)
                 return;
             }
 
-            if (argument.Parameter == null)
```
Evangelink: The hover popup says that `Parameter` can't be null so I am removing this un-needed/unreachable early exit.

sharwell: The hover popups are not a guarantee. In this case, the definition of `IArgumentOperation` is null oblivious.

Evangelink: Ah too bad! I thought that was something based on nullability declaration,

Evangelink:
@sharwell do you know a way to produce a null parameter symbol from a unit-test?

sharwell:
I do not.

sharwell:
My first guess would be a call site that passes more arguments than the method has parameters.

Evangelink:
From the various tests I have made, it seems that the `OperationKind.Argument` callback only occurs when a valid overload is found so I am really not sure what could be done to have something null. I'll close this PR except if you prefer that I add a `Debug.Assert` and handle the nullability in the switch on the following lines.

src/Microsoft.CodeQuality.Analyzers/Core/Maintainability/UseNameofInPlaceOfString.cs
mavasani: I believe `IArgumentOperation.Parameter` cannot be null. @333fred to confirm.

sharwell:
I would prefer to leave it as-is for now, but consider revisiting once we make more progress on the nullability annotations in the IOperation APIs (actively being worked on).

PR #3176, opened 2020-01-08
CA1021: Avoid out parameters

### LINKED ISSUE #377: Port FxCop rule CA1021: AvoidOutParameters

Evangelink:
Fix #377


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,70 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            analysisContext.RegisterCompilationStartAction(csac =>
+            {
+                var outAttributeType = csac.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesOutAttribute);
```
Evangelink: Maybe it's better if I do this call directly in the `IsOutParameter` as the value is cached.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,70 @@
...
+                        sac.ReportDiagnostic(methodSymbol.CreateDiagnostic(Rule));
+                    }
+                }, SymbolKind.Method);
+            });
+        }
+
+        private static bool IsOutParameter(IParameterSymbol parameterSymbol, INamedTypeSymbol? outAttributeSymbol) =>
+            parameterSymbol.RefKind == RefKind.Out ||
+            // Handle VB.NET special case for out parameters
+            (parameterSymbol.RefKind == RefKind.Ref && parameterSymbol.HasAttribute(outAttributeSymbol));
```
Evangelink: I couldn't find any better way for handling this case. It seems that there is an `IsOut` property on the class itself but that's not available through the interface.

Evangelink: FYI `HasAttribute` handles null parameter.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1352,4 +1352,13 @@
...
   </data>
+  <data name="AvoidOutParametersDescription" xml:space="preserve">
+    <value>Passing types by reference (using 'out' or 'ref') requires experience with pointers, understanding how value types and reference types differ, and handling methods with multiple return values. Also, the difference between 'out' and 'ref' parameters is not widely understood.</value>
+  </data>
+  <data name="AvoidOutParametersMessage" xml:space="preserve">
+    <value>Avoid 'out' parameters as they are not designed for general audience.</value>
+  </data>
+  <data name="AvoidOutParametersTitle" xml:space="preserve">
+    <value>Avoid out parameters</value>
+  </data>
```
Evangelink: Shall we reuse the fxcop messages? If so are they available somewhere or do I need to run the old fxcop to get them?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,70 @@
...
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidOutParametersDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(RuleId,
+                                                                             s_localizableTitle,
+                                                                             s_localizableMessage,
+                                                                             DiagnosticCategory.Design,
+                                                                             DiagnosticHelpers.DefaultDiagnosticSeverity,
+                                                                             isEnabledByDefault: DiagnosticHelpers.EnabledByDefaultIfNotBuildingVSIX,
```
mavasani: I think we want this rule to be disabled by default - out parameters are very common and this will potentially make the package noisy.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,70 @@
...
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            analysisContext.RegisterCompilationStartAction(csac =>
+            {
+                var outAttributeType = csac.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesOutAttribute);
+
+                csac.RegisterSymbolAction(sac =>
+                {
+                    var methodSymbol = (IMethodSymbol)sac.Symbol;
```
mavasani: This rule should only be enabled on the public surface area by default. See https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1021?view=vs-2019#cause:

`A public or protected method in a public type has an out parameter.`

mavasani: Additionally, the rule should respect `api_surface` editorconfig option to allow users to configure the analyzed surface area. You should invoke `ISymbolExtensions.MatchesConfiguredVisibility` API to get this support. See https://github.com/dotnet/roslyn-analyzers/search?q=MatchesConfiguredVisibility&unscoped_q=MatchesConfiguredVisibility

mavasani: Please also update https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#analyzed-api-surface and file an issue on https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1021 to add a sub-section about API configurability. Similar to https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1000?view=vs-2019#configurability

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                    var numberOfOutParams = methodSymbol.Parameters.Count(p => IsOutParameter(p, outAttributeType));
+
+                    if (numberOfOutParams >= 1 && !IsTryPatternMethod(methodSymbol, outAttributeType, numberOfOutParams))
```
mavasani: Should you also special case Deconstruct method? https://github.com/dotnet/roslyn-analyzers/issues/377#issuecomment-505138673

Evangelink: I saw the comment but forgot to add the related tests, thanks for the catch!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
Evangelink: If there isn't any better way, I think that it might be worth to have such method directly available in Roslyn. WDYT?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
```diff
@@ -0,0 +1,109 @@
...
+                    !IsDeconstructPattern(methodSymbol))
+                {
+                    analysisContext.ReportDiagnostic(methodSymbol.CreateDiagnostic(Rule));
+                }
+            }
+
+            private bool IsOutParameter(IParameterSymbol parameterSymbol) =>
+                parameterSymbol.RefKind == RefKind.Out ||
+                // Handle VB.NET special case for out parameters
+                (parameterSymbol.RefKind == RefKind.Ref && parameterSymbol.HasAttribute(outAttributeSymbol));
```
Evangelink: If there isn't a better way to handle this case, I think that is might be worth having this extension method directly available in Roslyn, WDYT?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: I think it is fine to use this message.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidOutParameters.cs
mavasani: Yes, might be worth filing an API request in Roslyn.

PR #3180, opened 2020-01-08
Just reformatting some comments in unit tests

dotpaul:


dotpaul:
Thanks for the super fast reviews!

PR #3190, opened 2020-01-10
Adjust the default severities and enabled by default values for all t…

mavasani:
…he CA rules packaged in Microsoft.CodeAnalysis.NetAnalyzers package as per the internal triage done for these rules.

These values should be unchanged in Microsoft.CodeAnalysis.FxCopAnalyzers package (except for few fixes for security rules, which I have confirmed with @dotpaul)

Tips for review:

1. All the product changes are in the rule descriptors. These get auto-translated into the .sarif and .md documentation files by a tool that runs over analyzer assemblies after cibuild. I would suggest only reviewing the changes to .sarif files (.md file changes don't have nice diff views) for this PR.
2. @dotpaul - you may want to review the changes to Microsoft.CodeAnalysis.FxCopAnalyzers.sarif to ensure that all changes are expected, especially for security rules. Primary changes should be fixups of help links, add missing telemetry tags and fixup of enabled by default/dataflow tags for some security rules.
3. @sharwell @genlu - you may want to review the changes to Microsoft.CodeAnalysis.NetAnalyzers.sarif to double check the default severity and isEnabledByDefault changes as per our triage spreadsheet.


mavasani:
Attaching the triaged CA rules spreadsheet: [CA_Rules.xlsx](https://github.com/dotnet/roslyn-analyzers/files/4046437/CA_Rules.xlsx)

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif
```diff
@@ -153,7 +154,7 @@
           "id": "CA1507",
           "shortDescription": "Use nameof to express symbol names",
           "fullDescription": "Using nameof helps keep your code valid when refactoring.",
-          "defaultLevel": "warning",
+          "defaultLevel": "note",
```
mavasani: I plan to change "note" to "suggestion" and "hidden" to "silent" in the sarif generation tool as a separate PR.

src/Utilities/Compiler/RuleLevel.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+#if NET_ANALYZERS || FXCOP_ANALYZERS
+
+namespace Microsoft.CodeAnalysis
+{
+    internal enum RuleLevel
```
mavasani: @sharwell @genlu You may want to start the review from this file. This closely matches the rule levels that we decided. I have excluded the SDL required rule level and made the security rules use `BuildWarning`, which is basically level 2 in the spreadsheet.

`IdeSuggestion` and `IdeHidden_BulkConfigurable` are basically rule level 3 in the spreadsheet, with the former being no false positive/useful as IDE suggestion case and latter being potential false positive case.

`Disabled` and `CandidateForRemoval` are levels 4 and 5 respectively, as per the spreadsheet.

src/Utilities/Compiler/DiagnosticDescriptorHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Diagnostics;
+using System.Linq;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+
+namespace Analyzer.Utilities
+{
+    internal static partial class DiagnosticDescriptorHelper
+    {
+        public static DiagnosticDescriptor Create(
```
mavasani: Core helper that maps RuleLevel to default severity and enabled by default.

src/NetAnalyzers/Directory.Build.props
```diff
@@ -0,0 +1,10 @@
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+  <Import Project="..\Directory.Build.props"/>
+
+  <PropertyGroup>
+    <FxCopAnalyzersProject Condition="$(MSBuildProjectName.StartsWith('Microsoft.CodeQuality')) OR $(MSBuildProjectName.StartsWith('Microsoft.NetCore')) OR $(MSBuildProjectName.StartsWith('Microsoft.NetFramework'))">true</FxCopAnalyzersProject>
+	<DefineConstants Condition="'$(FxCopAnalyzersProject)' == 'true'">$(DefineConstants),FXCOP_ANALYZERS</DefineConstants>
```
genlu: What's the purpose of these constants?

genlu:
>           "defaultLevel": "hidden",

CA1065 is marks as off in the spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:101 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

src/NetAnalyzers/Directory.Build.props
mavasani: `FXCOP_ANALYZERS` - used for our currently shipping FxCopAnalyzers package, where we want to retain current packaging names, rule severities and enabled by default values.

`NET_ANALYZERS` - used for new NetAnalyzers package, where we want new values as per our triage.

Both these packages share the same underlying source files, so we need some preprocessor directives to distinguish and generate respective severities and enabled by default values.

mavasani:
> CA1065 is marks as off in the spreadsheet

Ah, thanks. I will fix.

genlu:
How's the value of `default level` and `enabled by default` decided for rules don't have proposed values in the spreadsheet (e.g. CA1051)?

genlu:
>             "isEnabledByDefault": true,

I thought all security rules will be off by default?

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:4386 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

mavasani:
> How's the value of default level and enabled by default decided for rules don't have proposed values in the spreadsheet (e.g. CA1051)?

I set most of these to `RuleLevel.IdeHidden_BulkConfigurable`, so they are not enabled by default, but bulk configurable by category.

genlu:
>           "defaultLevel": "note",

CA2224 is level5 and marked as off by default in the spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:4344 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

mavasani:
> I thought all security rules will be off by default?

I had an offline chat with @dotpaul - he will triage the enabled by default values for all the security rules in the NetAnalyzers package. For now, we decided to keep the current enabled by default values  for all security rules in both FxCopAnalyzers and NetAnalyzers package.

genlu:
>             "isEnabledByDefault": true,

CA2218 is off in spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:4329 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

genlu:
>             "isEnabledByDefault": true,

CA1065 is off in spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:4087 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

mavasani:
> CA2218 is off in spreadsheet

I forgot to add a comment to the spreadsheet. We marked this as candidate for removal as C# compiler already reports an equivalent CS warning. However, VB compiler does not report any such diagnostic. CA2218 has a VB-only implementation in the repo, so I think we should keep this analyzer and update the spreadsheet. Same holds for CA2224. These 2 are the only VB-only analyzers in our repo.

Thoughts @sharwell @genlu?

genlu:
>             "isEnabledByDefault": true,

CA1822 is marked off in spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:1809 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

mavasani:
> CA1822 is marked off in spreadsheet

Hmm, it is marked as level 3 with no false positives, so I think I got the third column value wrong, it should be enabled as a suggestion instead of being off. This rule is Mark members static: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1822?view=vs-2019

genlu:
>             "isEnabledByDefault": true,

CA1031 is off in spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:652 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

genlu:
>             "isEnabledByDefault": false,

CA1014 is level 4/on+hidden in the spreadsheet

---
Refers to: src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif:492 in aa5cb16. [](commit_id = aa5cb161ee5799aa1a1e199a7792f84fc9208edc, deletion_comment = False)

mavasani:
> CA1014 is level 4/on+hidden in the spreadsheet

We had a comment "Need to update templates to add CLSCompliant(false)". I presume we should not enable it by default till we change the templates? I will file a tracking issue to switch it to on once we update the templates.

genlu:
Done with the pass. I have gone through Microsoft.CodeAnalysis.NetAnalyzers.sarif and all utility code changes

mavasani:
> CA1031 is off in spreadsheet

Ah, I should fix it.

src/NetAnalyzers/Directory.Build.props
genlu: I see. So basically this is just to single out MS.CA.*.NetAnalyzers

mavasani:
I have pushed a change addressing the feedback. Summary:

1. CA1065 - now fixed to be off by default
2. CA2218 and CA2224 - marked for deprecation in spreadsheet, but has VB only implementations which should be retained. So, the rules are still enabled as suggestions and spreadsheet should be changed.
3. Security rules: PR retains the current enabled by default values, except for DFA security rules which were accidentally on by default.
4. CA1822 - Spreadsheet needs to be fixed so the column that I computed from triaged values is correct
5. CA1031 - now fixed to be off by default
6. CA1014 - Spreadsheet says that should be enabled by default as a suggestion, but needs templates to be fixed before this can be done. So, the rule will be off by default right now and I will file a tracking issue to enable it to suggestion once templates are fixed.

mavasani:
@genlu Let me know if https://github.com/dotnet/roslyn-analyzers/pull/3190#issuecomment-573807925 sounds reasonable

genlu:
Sounds good to me. Thx

mavasani:
@genlu Any further feedback? @dotpaul would you be able to review the changes to FxCopAnalyzers.sarif?

mavasani:
Thank you @genlu @dotpaul

PR #3191, opened 2020-01-10
CA1822: Update rule to flag more cases

### LINKED ISSUE #3019: CA1822 doesn't suggest method be made static if called from xUnit fact
#### Analyzer package

#### Package Version

v2.9.7

#### Diagnostic ID

CA1822 Mark members as static

#### Repro steps

```c#
using Xunit;

public class Program
{
    [Fact]
    private void M()
    {
        N();
    }

    private void N() // expected suggestion to convert to static
    {
    }
}
```

#### Expected behavior

Suggests method `N` should be static.

#### Actual behavior

No suggestion.

Removing the `FactAttribute` from `M` causes the suggestion to appear as expected.

Evangelink:
Fix #3019 

With this change the analyzer will look for invocations located in _ALL_ methods but will keep raising issues only on correct ones. Thanks to this change we shall be able to report a lot more cases where the method could be made `static`.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -6,7 +6,6 @@ using System.Collections.Immutable;
 using System.Linq;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
-using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -577,6 +577,70 @@ End Class
...
+                TestState =
+                {
+                    Sources =
+                    {
+                        @"
+using Xunit;
+
+public class Program
+{
+    [Fact]
```
Evangelink: I only made a test with `Fact` as it is the one reported in the ticket. Is it worth adding test cases for all of the exit conditions of the `ShouldAnalyze` method?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -70,29 +69,36 @@ public override void Initialize(AnalysisContext analysisContext)
 
                 compilationContext.RegisterOperationBlockStartAction(blockStartContext =>
                 {
-                    if (!(blockStartContext.OwningSymbol is IMethodSymbol methodSymbol) || !ShouldAnalyze(methodSymbol, wellKnownTypeProvider, skippedAttributes))
```
mavasani: Wow, I wonder how it took so long to discover this bug. Thanks!

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
mavasani: I think this coverage is fine.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: It was probably simply not reported.

Issue #3174, opened 2020-01-08
[Discussion] Specific attributes


Evangelink:
Hi there,

I don't know whether this is the right repository for this discussion, it might be better on `roslyn` or maybe even directly on `csharp-lang`.

With the increased use of analyzers, I see more and more cases where having some specific annotations would help tailoring the end-user experience. For instance, I have seen a couple of issues related to having some end-user exclusion option for some rules. Adding the exclusion option in the `.editorconfig` file is great but the same type/member/... would need to be added for a lot of rules whereas if it was possible to annotate some behavior _ALL_ other analyzers could read and use this metadata.

For example, on https://github.com/dotnet/roslyn-analyzers/issues/2983, having the ability for users to exclude some specific attributes is nice but it would be even better if the developer who created the attribute was able to annotate it with something to say whether or not it accesses something on the instance.

Of course, if we had a perfect thing we could do some dataflow analysis directly on the binary code and we wouldn't need any input from the developer as it would be based only on the real use. Note that even if it was possible I still like the idea to _annotate_/_document_ my intent and have an analyzer (or some tests) checking that I have respected my intent!

Issue #2371, opened 2019-04-18
PerformanceSensitiveAttribute conflicts with IVTs


sharwell:
#### Analyzer package

Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers

#### Package Version

Example: 3.0.0-beta2.19218.3+e96bad97

#### Diagnostic ID

n/a

#### Repro steps

1. Create a project First, which has IVT visible to Second
2. Create a project Second, with a reference to First
3. Install the analyzer package in both projects

#### Expected behavior

No warnings.

#### Actual behavior

> Warning CS0436 The type 'PerformanceSensitiveAttribute' in 'C:\Users\sam\.nuget\packages\microsoft.codeanalysis.performancesensitiveanalyzers\3.0.0-beta2.19218.3\build\PerformanceSensitiveAttribute.cs' conflicts with the imported type 'PerformanceSensitiveAttribute' in 'Microsoft.CodeAnalysis, Version=42.42.42.42, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. Using the type defined in 'C:\Users\sam\.nuget\packages\microsoft.codeanalysis.performancesensitiveanalyzers\3.0.0-beta2.19218.3\build\PerformanceSensitiveAttribute.cs'.

genlu:
This should be resolved by setting `PerformanceSensitiveAttribute` to true only at the root of IVT tree.

See https://github.com/dotnet/roslyn/blob/master/eng/targets/Settings.props#L139
```xml
  <!-- 
    Don't inject PerformanceSensitiveAttribute source by default to avoid duplicate definitions caused by IVT.
    This needs to be set to true in projects at the root of IVT trees, in order for PerformanceSensitiveAnalyzers to work.
  -->
  <PropertyGroup Condition="'$(GeneratePerformanceSensitiveAttribute)' == ''">
    <GeneratePerformanceSensitiveAttribute>false</GeneratePerformanceSensitiveAttribute>
  </PropertyGroup>
```

PR #3193, opened 2020-01-10
[WIP] Add an annotation package

genlu:


Evangelink:
Hi @genlu,

It could be interesting to have this annotation project/package common to the various analyzers instead of having it specific to each one (specifically performance in this case). This relates to https://github.com/dotnet/roslyn-analyzers/issues/3174

sharwell:
Why is the previous approach insufficient? The goal was to eliminate the need for additional assemblies related to the attributes.

genlu:
@sharwell because of IVT causing conflicts https://github.com/dotnet/roslyn-analyzers/issues/2371. I couldn't think of a better way to solve this issue than converting it to assembly. Note that the assembly reference wouldn't be added to the built binaries unless the constant is defined.

sharwell:
It can be resolved using conditional include logic, such as the `GenerateNullableAttributes` property here:

https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator/blob/c14d291d32286d8db1477e668059207145a87a52/TunnelVisionLabs.ReferenceAssemblyAnnotator/TunnelVisionLabs.ReferenceAssemblyAnnotator.targets#L16-L21

sharwell:
It looks like we already have a `GeneratePerformanceSensitiveAttribute` property, which can be set false for the second project in the example for #2371.

genlu:
That would require user to understand how this package works and then set an additional property in each project that has IVT though.

genlu:
> It could be interesting to have this annotation project/package common to the various analyzers instead of having it specific to each one (specifically performance in this case). This relates to #3174

I think it really depends on how the analyzers are shipped. It might make sense to have a common annotation package when consuming analyzers bundled together in some way, for example, the fxcop analyzers meta package. However, if those analyzer packages each has a specific use case (thus not bundled together, like this one), having a separate annotation package seem to be better.

BTW, we are still deciding whether to use a separate assembly for the attributes or stick to the current source injection approach (comments welcomed). The discussion here would be moot if we ended up not taking this change :)

genlu:
Also tag @mavasani for additional inputs on how to distribute annotation type, i.e. source injection vs additional assembly

mavasani:
Couple of thoughts (I am pretty sure you feel somewhat same):

1. Repro steps mentioned in https://github.com/dotnet/roslyn-analyzers/issues/2371 should work as-is without requiring any project file edits for the user
2. I do not like the idea of a separate assembly for attributes needed by analyzers, especially given it leaks at bunch of places: UI (Analyzers node), project file (Analyzer ItemGroup), etc.

Having said that, I don't think I have a good suggestion on how to accomplish this.

genlu:
@mavasani I agree requiring manual changes to proj file just to use the analyzer isn't good experience. That's why I'm leaning toward the annotation assembly approach, even with some of it's downsides.

>  I do not like the idea of a separate assembly for attributes needed by analyzers, especially given it leaks at bunch of places: UI (Analyzers node), project file (Analyzer ItemGroup), etc.

Yeah, the reference nodes UI in VS does get crowded if we use an additional annotation assembly (see screenshot below), but the proj file is clean (at least for SDK style projects). The way the changes in this PR works is, users just reference the annotation package which has a dependency to the analyzer package so it will be pulled in automatically.

image

@sharwell thoughts?

mavasani:
@genlu If we do go the path of annotation package, we should make sure they are conditionally included only for debug builds. I also agree with @Evangelink that eventually we may want a common annotation package, at least for the primary analyzer packages that we ship: Microsoft.CodeAnalysis.FxCopAnalyzers and Microsoft.CodeAnalysis.NetAnalyzers.

genlu:
> we should make sure they are conditionally included only for debug builds

In terms of reference in metadata, it's fine because the attribute is declared as `[Conditional("EMIT_CODE_ANALYSIS_ATTRIBUTES")]`, so it won't be emitted unless the constant is defined. However, user needs to add `PrivateAssets="all"` to their PackageReference to avoid the annotation package becoming a dependency say if they are building a nuget package. 

I'm wondering if there a nuspec setting similar to `developmentDependency` but include compile assets? Otherwise, this comes down to "_source injection with proj change if there's IVT_" vs. "_annotation assembly with project change if transitive dep is undesired_". The first option seems better to me now, especially if we ensured this scenario well is well-documented. (FWIW, [resharper](https://www.jetbrains.com/help/resharper/Code_Analysis__Annotations_in_Source_Code.html#) mainly relies the annotation assembly approach, and people seems fine with it)

src/PerformanceSensitiveAnalyzers/Annotations/PerformanceSensitiveAttribute.cs
```diff
@@ -15,7 +15,7 @@ namespace Roslyn.Utilities
     /// </remarks>
     [Conditional("EMIT_CODE_ANALYSIS_ATTRIBUTES")]
     [AttributeUsage(AttributeTargets.Constructor | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Field, AllowMultiple = true, Inherited = false)]
-    internal sealed class PerformanceSensitiveAttribute : Attribute
+    public sealed class PerformanceSensitiveAttribute : Attribute
```
mjsabby: can this be Microsoft.CodeAnalysis and not Roslyn.Utilities?

genlu: I think that makes sense. @sharwell thoughts?

sharwell:
I would prefer to look for better solutions under the original packaging rather than try and distribute this as a new package, especially considering the fact that PerformanceSensitiveAnalyzers isn't shipping as a stable package yet _and_ it hasn't been updated to consider the various properties of this attribute in its analysis.

Issue #2133, opened 2019-02-15
LINQ syntax artificially inflates class coupling counts for CA1506


j2jensen:
#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

Example: v2.6.3

#### Diagnostic ID

[CA1506](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1506-avoid-excessive-class-coupling?view=vs-2017)

#### Repro steps

Use a block of LINQ code:

    using System.Collections.Generic;
    using System.Linq;

    public static class Ca1506Tester
    {
        public static IEnumerable<int> TestCa1506()
        {
            var ints = new[] { 1, 2 };
            return from a in ints
                   from b in ints
                   from c in ints
                   from d in ints
                   from e in ints
                   from f in ints
                   from g in ints 
                   from h in ints
                   from i in ints
                   from j in ints
                   from k in ints
                   from l in ints
                   from m in ints
                   from n in ints
                   from o in ints
                   from p in ints
                   select p;
        }
    }

#### Expected behavior

In terms of code maintainability, this method is really only coupled with a handful of types (`Enumerable`, `IEnumerable<>`, `Array`, and `int`. It shouldn't trigger CA1506.

#### Actual behavior

CA1506 is triggered with a message like this:
> error CA1506: 'TestCa1506' is coupled with '60' different types from '4' different namespaces. Rewrite or refactor the code to decrease its class coupling below '41'.

Of course this example is a little contrived, but in normal methods if someone were writing a linq query navigating across object models rather than ints, LINQs tendency to artificially inflate the numbers will add to whatever natural class coupling is involved and quickly push a method or a class over its reasonable threshold.

As an implementation detail, we know that each `from` is translated to a `SelectMany` call that creates a new anonymous type. I suspect that each of these anonymous types adds a class coupling that gets counted by the rule. I don't know if it's possible to differentiate these auto-generated anonymous types from other types, but it seems like it may be reasonable to exclude anonymous types from class coupling counts generally.

mavasani:
> it seems like it may be reasonable to exclude anonymous types from class coupling counts generally

[...]

PR #3194, opened 2020-01-10
Make code metrics ignore anonymous types when counting class coupling.

j2jensen:
This aims to fix https://github.com/dotnet/roslyn-analyzers/issues/2133 by not counting anonymous types when counting class coupling.

I still need to check whether classes generated for closures are getting counted by metrics, but since this is my first time contributing to this project I wanted to get early feedback on a small pull request.


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
```diff
@@ -550,6 +550,29 @@ class C2 : C1 { }
...
+{
+    object M1(C1 c)
+    {
+        return new {a = 1};
+    }
+}
+";
+
+            var expectedMetricsText = @"
+Assembly: (Lines: 8, ExecutableLines: 1, MntIndex: 94, CycCxty: 1, CoupledTypes: {C1}, DepthInherit: 1)
```
mavasani: I presume the test would fail without your product fix, correct? CoupledTypes should have listed the anonymous type?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
```diff
@@ -550,6 +550,29 @@ class C2 : C1 { }
             VerifyCSharp(source, expectedMetricsText);
         }
 
+        [Fact]
```
mavasani:
```suggestion
        [Fact, WorkItem(2133, "https://github.com/dotnet/roslyn-analyzers/issues/2133")]
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
j2jensen: Yes, I used Test-Driven Development, writing this test first and verifying that it failed as expected (listing the anonymous type). Then I got the test passing by changing the MetricsHelper.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
```diff
@@ -280,6 +280,11 @@ private static void AddCoupledNamedTypesCore(ImmutableHashSet<INamedTypeSymbol>.
             // Compat
             static bool isIgnoreableType(INamedTypeSymbol namedType)
```
Evangelink: I think that the compiler generated types shall also be excluded:
`namedType.GetAttributes().Any(a => a.AttributeClass.Equals(compilerGeneratedAttributeType) || a.AttributeClass.Equals(generatedCodeAttributeType))`

Evangelink:
tagging @mavasani

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
mavasani: Sounds reasonable to me.

j2jensen: @Evangelink I'm working on some unit tests and fixes for what you're describing here. My plan for generics is to include the Unbound generic type (`IEnumerable<>`) in the list of coupled types instead of the actual generic type (`IEnumerable<int>`). Does that sound good?

There's currently an optimization where we avoid recursively checking the generic sub-types if we've already seen the exact generic type we're looking at. Since I won't be adding the actual generic type to the builder anymore, the simplest approach would be to drop that optimization. In theory this could yield degraded performance in code that has a lot of references to nested generic types (`G<G<G<G<a>>>>`), but I think that will be rare. Does that sound all right to you?

j2jensen: All right, I've fixed the code in the way I describe above, and added a couple of tests, including one for the specific code I used in #2133 .

The `ConditionalLogic_CoalesceAndConditionalAccess` test previously expected that the generic nullable type (e.g. `bool?`) would be referenced. In keeping with the approach I described, I'm changing that to expect the Unbound generic type instead (e.g. `Nullable<>`, represented as `T?` in the output). But I wanted to check that we want `Nullable<>` to count at all, and not to be treated as an ignored type, the way we ignore `int`, `bool`, etc.? Same goes for `Action` and `Func<>`: we are no longer generating separate coupled types for each generic type of Action and Func, so I consider my original issue fixed regardless, but an unbound generic type is being counted if the code has a lambda or LINQ expression.

In terms of compiler-generated code, it appears that there's nothing showing for closures, so I'm not including the attribute check you mentioned for this pull request. I am curious to see what happens if we use `yield return` or `async`/`await`, but I don't think that should necessarily be something I address in this pull request.

j2jensen:
FYI I was able to verify that compiler-generated types produced from `async`/`await` and `yield return` patterns are already not counted as coupled types, so I'm ready for this PR and issue to be completed and resolved.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
```diff
@@ -265,21 +265,30 @@ static bool hasConditionalLogic(IOperation operation)
...
         {
             if (typeOpt is INamedTypeSymbol usedType &&
-                !isIgnoreableType(usedType) &&
-                builder.Add(usedType))
+                !isIgnoreableType(usedType))
             {
-                if (usedType.IsGenericType)
+                if (usedType.IsGenericType && !usedType.IsUnboundGenericType)
                 {
+                    builder.Add(usedType.ConstructUnboundGenericType());
```
Evangelink: @mavasani will correct me but I think that's better to use the `OriginalDefinition` property. You can use it in both the if and else as it will return the same value if there is no generics.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
```diff
@@ -550,6 +551,124 @@ class C2 : C1 { }
...
+";
+
+            var expectedMetricsText = @"
+Assembly: (Lines: 12, ExecutableLines: 4, MntIndex: 88, CycCxty: 2, CoupledTypes: {C1, C2, G<>}, DepthInherit: 1)
+   C: (Lines: 11, ExecutableLines: 4, MntIndex: 77, CycCxty: 1, CoupledTypes: {C1, C2, G<>}, DepthInherit: 1)
+      C.M1(): (Lines: 7, ExecutableLines: 4, MntIndex: 77, CycCxty: 1, CoupledTypes: {C1, C2, G<>})
+   G<T>: (Lines: 1, ExecutableLines: 0, MntIndex: 100, CycCxty: 1, DepthInherit: 1)
+";
+
+            VerifyCSharp(source, expectedMetricsText, true);
```
Evangelink: @mavasani Looks like this test class wasn't updated to use the new test framework. `C1` and `C2` are not declared yet there is no compiler error.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
```diff
@@ -265,21 +265,30 @@ static bool hasConditionalLogic(IOperation operation)
         private static void AddCoupledNamedTypesCore(ImmutableHashSet<INamedTypeSymbol>.Builder builder, ITypeSymbol typeOpt)
         {
             if (typeOpt is INamedTypeSymbol usedType &&
-                !isIgnoreableType(usedType) &&
-                builder.Add(usedType))
```
mavasani: You can change this to `builder.Add(usedType.OriginalDefinition)` and will likely not need any other changes in this method.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
j2jensen: The `true` being passed as the final argument to VerifyCSharp seems to indicate that we expect compiler errors to happen. The analyzer still appears to be able to produce meaningful information, even though it sees some of the types as having type errors. I was following the patterns that I saw in nearby methods, but if you'd prefer I can add all the necessary classes to prevent the test from failing.

Evangelink: Up to you but that's fine like this I think. We have switch to a different "test framework" so I thought that might have some parts broken hecne this comment.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
mavasani: Ping @j2jensen

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
mavasani: I agree with @Evangelink - I think it seems best to not have compile errors for this test. If you wish to also test the functionality in presence of compiler errors, please add another test with appropriate test name.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
j2jensen: Sorry it has taken me a while. I haven't forgotten this, just haven't made time for it. I should have a chance to work on it again this Thursday.

j2jensen: Changed to use OriginalDefinition.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
j2jensen: Fixed compile errors on all the test methods I added. I also made one change to the base class to make it easier to track down unexpected diagnostic messages in case of failure.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
j2jensen: I changed it to `builder.Add(usedType.OriginalDefinition)` and reverted most of the code back to the way it was. We still need to take the call out of the `if` expression, though, to ensure that different generic arguments on the same generic types (`G<C1>` and `G<C2>`) cause their generic arguments' types (`C1` and `C2`) to be included.

j2jensen:
I made most of the changes recommended. I did not get around to adding tests to `CodeMetricsAnalyzerTests.cs`. If that's important, it'll have to wait until I get free time to work on it again. Other than that, I think this is ready.

j2jensen:
I found time tonight to add a test to `CodeMetricsAnalyzerTests.cs`. Ready for another review.

src/Test.Utilities/CodeMetricsTestsBase.cs
```diff
@@ -68,7 +68,14 @@ private void Verify(string[] sources, string expectedMetricsText, bool expectDia
             var project = CreateProject(sources, language);
             var compilation = project.GetCompilationAsync(CancellationToken.None).Result;
             var diagnostics = compilation.GetDiagnostics().Where(d => d.Severity == DiagnosticSeverity.Warning || d.Severity == DiagnosticSeverity.Error);
-            Assert.Equal(expectDiagnostics, diagnostics.Any());
```
mavasani: Why was this change required?

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
mavasani: Can we still restore the bail out condition where build.Add returned false for a non-generic type?

mavasani: Ah, I see the only code within the if statement runs on generic types, so we should be fine.

src/Test.Utilities/CodeMetricsTestsBase.cs
j2jensen: It wasn't required to make the tests pass or anything, but I found it useful when a test that was not expecting diagnostics to appear caused some to appear, to have a list of those diagnostics in the output instead of just "expected false but found true."

If you don't like it, I can easily back it out.

j2jensen:
> Changed LGTM
> 
> Unit tests are failing on CI as you seem to have a missing using in one of your test files. I can merge once you can fix it up. Thanks for the contribution!

Looks like the test classes were changed since I started this work. The `using` wasn't missing on my branch, but it got removed upstream. I'll get everything synced up. @mavasani is merge okay, or do you prefer rebase?

j2jensen:
@mavasani do you need anything else from me? This looks ready to merge.

PR #3200, opened 2020-01-13
Convert more tests to the new test framework

sharwell:
This change leverages stability improvements in the new test framework releases to convert most of the remaining tests. Following this change, the only type derived from `DiagnosticAnalyzerTestBase` is `PropertySetAnalysisTests`. I started the process for updating MetaCompilation.Analyzers as well, which is much easier with the latest test framework updates.

I can break this change up if necessary.


src/MetaCompilation.Analyzers/UnitTests/UnitTests.cs
```diff
@@ -1,33 +1,34 @@
...
-    public class UnitTest : CodeFixVerifier
+    public class UnitTest
     {
         private const string s_messagePrefix = "T: ";
 
         #region default no diagnostics tests
         // no diagnostics
         [Fact]
-        public void TestMethod1()
+        public async Task TestMethod1()
```
sharwell: 📝 The tests are converted as follows:

1. Make the test method asynchronous.
2. Delete the previous `expected` definition and call to `VerifyCSharpDiagnostic`. Make a note of the diagnostic ID field used for creating `expected`, as it will be needed below.
3. Replace the call to `VerifyCSharpFix` with the following:

    ```csharp
    await new VerifyCS.Test
    {
        TestState =
        {
            Sources = { test },
            ExpectedDiagnostics =
            {
            },
        },
        FixedState =
        {
            Sources = { fixtest },
            ExpectedDiagnostics =
            {
            },
        },
        CodeFixEquivalenceKey = null,
    }.RunAsync();
    ```

4. Locate the code fix registration in **CodeFixProvider.cs** for the diagnostic seen in step 2, and find the equivalence key string. Copy that string and paste it over the `null` used in the placeholder from step 3.
5. Run the test to get the expected diagnostics for the test state. Copy the block from the test output to the `ExpectedDiagnostics` initializer, and replace the identifier for the diagnostic ID with the one seen in step 2.
6. Run the test again to get the expected diagnostics for the fixed state. Copy the block from the test output to the `ExpectedDiagnostics` initializer. If any diagnostics are reported by the analyzer (as opposed to compilation errors), look up the corresponding field in **DiagnosticAnalyzer.cs** to find the correct identifier.

I converted one `#region` at a time, since step 2 uses the same identifier for all tests within each region.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/InternalImplementationOnlyTests.cs
```diff
@@ -71,35 +102,282 @@ class Foo : IFoo { }
...
+            {
+                TestState =
+                {
+                    Sources = { source },
+                    ExpectedDiagnostics =
+                    {
+                        // Test0.cs(3,7): error RS1009: Type Foo cannot implement interface ISymbol because ISymbol is not available for public implementation.
+                        VerifyCS.Diagnostic().WithSpan(3, 7, 3, 10).WithArguments("Foo", "ISymbol"),
+                        // Test0.cs(3,13): error CS0535: 'Foo' does not implement interface member 'ISymbol.Accept(SymbolVisitor)'
+                        DiagnosticResult.CompilerError("CS0535").WithSpan(3, 13, 3, 43).WithArguments("Foo", "Microsoft.CodeAnalysis.ISymbol.Accept(Microsoft.CodeAnalysis.SymbolVisitor)"),
```
mavasani: Is there a test mode where caller can request to skip compiler diagnostic validation? This seems to be a case where forcing compiler diagnostic validation is leading to unnecessary noise in test baseline.

sharwell: There is such a mode, but it should not be used. The solution for this scenario is updating the input code to fully implement the interface with `=> throw null` bodies.

PR #3215, opened 2020-01-16
Remove foo, bar... in tests

### LINKED ISSUE #3213: Do not use "Foo" or "Bar" as name in tests
These names should not be used in tests as they generates some issues internally.

Evangelink:
Fix #3213


mavasani:
@Evangelink Tests seem to be hanging in this PR. Can you try locally?

Evangelink:
It seems to be ok from the `Test Explorer` but not from the command line. That's quite weird because there is no functional change...

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/InternalImplementationOnlyTests.cs
```diff
@@ -517,265 +517,249 @@ public async Task Basic_VerifyISymbol()
         {
             var source = @"
 ' Causes many compile errors, because not all members are implemented.
-Class Foo 
+Class C1
     Implements Microsoft.CodeAnalysis.ISymbol
 End Class
-Class Bar
+Class C2
```
Evangelink: Using `SomeClass` and `SomeOtherClass` here was causing the test to hang... I couldn't find out why (I didn't spent too long debugging it).

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/InternalImplementationOnlyTests.cs
```diff
@@ -517,265 +517,249 @@ public async Task Basic_VerifyISymbol()
...
         public async Task Basic_VerifyIOperation()
         {
             var source = @"
 ' Causes many compile errors, because not all members are implemented.
-Class Foo 
+Class C1
     Implements Microsoft.CodeAnalysis.IOperation
 End Class
-Class Bar
+Class C2
```
Evangelink: Using `SomeClass` and `SomeOtherClass` here was causing the test to hang... I couldn't find out why (I didn't spent too long debugging it).

PR #3230, opened 2020-01-21
Improve code fixes by automatically adding imports

sharwell:


src/Roslyn.Diagnostics.Analyzers/Core/ImportingConstructorShouldBeObsoleteCodeFixProvider.cs
```diff
@@ -206,7 +207,7 @@ private static SyntaxNode GenerateDescriptionArgument(SyntaxGenerator generator,
                 mefConstructionType.GetMembers("ImportingConstructorMessage").OfType<IFieldSymbol>().Any())
             {
                 attributeArgument = generator.MemberAccessExpression(
-                    generator.TypeExpressionForStaticMemberAccess(mefConstructionType),
+                    generator.TypeExpression(mefConstructionType).WithAddImportsAnnotation(),
```
sharwell: 📝 This is not a correct syntax tree, but AddImportsAnnotation only works for direct type references.

PR #3231, opened 2020-01-21
Emit the default severity of the rules in the .md documentation files

mavasani:
This is now useful as Microsoft.CodeAnalysis.NetAnalyzers project has rules with varying default severity.


src/Tools/GenerateAnalyzerRulesets/Program.cs
```diff
@@ -296,7 +296,7 @@ void createAnalyzerDocumentationFile()
                         description = descriptor.MessageFormat.ToString();
                     }
 
-                    builder.AppendLine($"{index} | {ruleIdWithHyperLink} | {descriptor.Title} | {descriptor.Category} | {descriptor.IsEnabledByDefault} | {hasCodeFix} | {description} |");
+                    builder.AppendLine($"{index} | {ruleIdWithHyperLink} | {descriptor.Title} | {descriptor.Category} | {descriptor.IsEnabledByDefault} | {descriptor.DefaultSeverity} | {hasCodeFix} | {description} |");
```
mavasani: This file has the only real tooling change. Everything else is auto-generated.

Issue #3212, opened 2020-01-16
CA2000 - Exclude StringReader by default


ashmind:
#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

[CA2000](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2000)

#### Repro steps

```
var x = new StringReader("abc");
```

#### Expected behavior
`StringReader` is a well-known type that does not need to be disposed, so no warning.

#### Actual behavior
`CA2000` is reported.

Evangelink:
That seems reasonable to me. We could either hardcode the exclusion or make this type part of the default value for the option (i.e. `dotnet_code_quality.CA2000.excluded_symbol_names = System.IO.StringReader`).

Tagging @mavasani

Evangelink:
Ping @mavasani

sharwell:
We can hard-code this exclusion. The lack of unmanaged resources is [documented for this type](https://docs.microsoft.com/en-us/dotnet/api/system.io.stringreader?view=netframework-4.8) and cannot be changed in the future.

Evangelink:
@sharwell Any preference between excluding the type post-processing or adding the type directly in the dispose helper processing part? I would say the latter but this could have consequences on other rules so I am not sure.

Evangelink:
Another interesting path could be to add this type to the list of `excluded_symbol_names` for the rule as we have allowed a custom exclusion for this kind of cases in #3085 .

mavasani:
@Evangelink I think it is fine to add the exclusion in the dispose helper or the rule, whichever is convenient.

PR #3233, opened 2020-01-22
Update CA1501 to allow user specific hierarchy exclusions

### LINKED ISSUE #1839: CA1501 AvoidExcessiveInheritance reports on all WPF Window/UserControl classes
#### Analyzer package

#### Package Version

v2.6.2

#### Diagnostic ID

[CA1501](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1501-avoid-excessive-inheritance?view=vs-2017)

#### Repro steps

1. Create WPF application with .xaml pages which inherit from Window or UserControl
2, Compile

#### Expected behavior

Build succeeds

#### Actual behavior

Build error reported:

error CA1501: 'SomeTab' has an object hierarchy '9' levels deep within the defining module. If possible, eliminate base classes within the hierarchy to decrease its hierarchy level below '6': 'UserControl, ContentControl, Control, FrameworkElement, UIElement, Visual, DependencyObject, DispatcherObject, Object'

Evangelink:
Fix #1839


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -164,7 +170,7 @@ void analyzeMetricsData(CodeAnalysisMetricData codeAnalysisMetricData)
                             var arg1 = symbol.Name;
                             var arg2 = codeAnalysisMetricData.DepthOfInheritance;
                             var arg3 = inheritanceThreshold + 1;
-                            var arg4 = string.Join(", ", ((INamedTypeSymbol)symbol).GetBaseTypes().Select(t => t.Name));
+                            var arg4 = string.Join(", ", ((INamedTypeSymbol)symbol).GetBaseTypes(t => !inheritanceExcludedTypes.Contains(t)).Select(t => t.Name));
```
Evangelink: It might be better to store the base types we have kept directly in the metrics data rather than doing multiple calculations.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -144,7 +144,13 @@ public override void Initialize(AnalysisContext analysisContext)
...
 
                 // Compute code metrics.
-                var computeTask = CodeAnalysisMetricData.ComputeAsync(compilationContext.Compilation, compilationContext.CancellationToken);
+                var inheritanceExcludedTypes = compilationContext.Options.GetInheritanceExcludedTypeNamesOption(CA1501Rule, compilationContext.Compilation, compilationContext.CancellationToken);
+
+                var metricsAnalysisContext = new CodeMetricsAnalysisContext(compilationContext.Compilation, compilationContext.CancellationToken)
+                {
+                    IsExcludedFromInheritanceCountFunc = namedType => inheritanceExcludedTypes.Contains(namedType)
+                };
+                var computeTask = CodeAnalysisMetricData.ComputeAsync(metricsAnalysisContext);
```
Evangelink: I have decided not to pass the options to the metric classes but instead to use a `Func` on some metrics context class I have created. I did so because I couldn't find any cache for the reading of the options.

src/Utilities/Compiler/CodeMetrics/CodeMetricsAnalysisContext.cs
```diff
@@ -0,0 +1,33 @@
...
+
+#if HAS_IOPERATION
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Microsoft.CodeAnalysis.CodeMetrics
+{
+    public sealed class CodeMetricsAnalysisContext
```
Evangelink: I have tried different ways of passing down the user option to the named type metric calculator and this one was IMO the best solution in term of reading and evolution. Please feel free to suggest another approach if you don't like this one.

src/Utilities/Compiler/CodeMetrics/CodeMetricsAnalysisContext.cs
```diff
@@ -0,0 +1,33 @@
...
+            CancellationToken = cancellationToken;
+            _semanticModelMap = new ConcurrentDictionary<SyntaxTree, SemanticModel>();
+            IsExcludedFromInheritanceCountFunc = isExcludedFromInheritanceCountFunc ?? (x => false); // never excluded by default
+        }
+
+        public Compilation Compilation { get; }
+        public CancellationToken CancellationToken { get; }
+        public Func<INamedTypeSymbol, bool> IsExcludedFromInheritanceCountFunc { get; internal set; }
+
+        internal SemanticModel GetSemanticModel(SyntaxNode node)
```
Evangelink: This was `internal` on the `SemanticModelProvider`.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -207,6 +214,13 @@ bool TryParse(string s, out SymbolNamesOption option)
                 option = SymbolNamesOption.Create(names, compilation, namePrefixOpt, getTypeAndSuffixFunc);
                 return true;
             }
+
+            SymbolNamesOption GetDefaultValue()
+#pragma warning disable CS8604 // Possible null reference argument.
+                => !string.IsNullOrEmpty(optionDefaultValue) && TryParse(optionDefaultValue, out var option)
```
Evangelink: Looks like `string.IsNullOrEmpty` is not handled by the symbolic execution.

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -134,5 +134,19 @@ internal static partial class EditorConfigOptionNames
...
+        /// Allowed name formats:
+        ///   1. Type or namespace name (includes all types with the name, regardless of the containing type or namespace and all types whose namespace contains the name)
+        ///   2. Type or namespace name ending with a wildcard symbol (includes all types whose name starts with the given name, regardless of the containing type or namespace
+        ///      and all types whose namespace contains the name)
+        ///   3. Fully qualified names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format
+        ///      with an optional "T:" prefix for types or "N:" prefix for namespaces.
+        ///   4. Fully qualified wildcard names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format
+        ///      with an optional "T:" prefix for types or "N:" prefix for namespaces and ending with the wildcard symbol.
+        /// </summary>
+        public const string InheritanceExcludedTypeNames = "inheritance_excluded_type_names";
```
Evangelink: We allow namespace exclusions so maybe `inheritance_excluded_type_names` isn't the best, WDYT?

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -12,23 +15,32 @@ namespace Analyzer.Utilities
...
+        private const SymbolKind AllKinds = SymbolKind.ErrorType;
+        private const char WildcardChar = '*';
+
+        private static readonly KeyValuePair<string, string?> NoWildcardMatch = default;
+
         public static readonly SymbolNamesOption Empty = new SymbolNamesOption();
 
         private readonly ImmutableDictionary<string, string?> _names;
         private readonly ImmutableDictionary<ISymbol, string?> _symbols;
+        private readonly ImmutableDictionary<SymbolKind, ImmutableDictionary<string, string?>> _wildcardNamesBySymbolKind;
```
Evangelink: I can't have `null` as the generic entry to try to match the various symbol kinds so I have added a constant to represent this state using the `ErrorType`.

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -98,32 +154,145 @@ private SymbolNamesOption()
...
 
         public bool Equals(SymbolNamesOption? other)
-            => other != null && _names.IsEqualTo(other._names) && _symbols.IsEqualTo(other._symbols);
+            => other != null && _names.IsEqualTo(other._names) && _symbols.IsEqualTo(other._symbols) && _wildcardNamesBySymbolKind.IsEqualTo(other._wildcardNamesBySymbolKind);
 
         public override int GetHashCode()
-            => HashUtilities.Combine(HashUtilities.Combine(_names), HashUtilities.Combine(_symbols));
+            => HashUtilities.Combine(HashUtilities.Combine(_names), HashUtilities.Combine(_symbols), HashUtilities.Combine(_wildcardNamesBySymbolKind));
+
+        private bool TryGetFirstWildcardMatch(ISymbol symbol, out KeyValuePair<string, string?> firstMatch)
```
Evangelink: This is the "simplest" algo I have found to match the various cases I wanted to handle. I don't think this performs particularly well. If we decided to stick with this algo we might want to have some cache of symbol <-> result to avoid doing the full match everytime.

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -98,32 +154,145 @@ private SymbolNamesOption()
...
+        {
+            firstMatch = NoWildcardMatch;
+
+            if (_wildcardNamesBySymbolKind.IsEmpty)
+            {
+                return false;
+            }
+
+            Debug.Assert(symbol.Kind == SymbolKind.Event || symbol.Kind == SymbolKind.Field || symbol.Kind == SymbolKind.Method || symbol.Kind == SymbolKind.NamedType || symbol.Kind == SymbolKind.Namespace || symbol.Kind == SymbolKind.Property);
```
Evangelink: I rely on the symbol kind but wasn't sure if it was best to play defensive and go for a bail-out condition.

src/Utilities/Compiler/PooledObjects/PooledDictionary.cs
```diff
@@ -41,6 +41,24 @@ private PooledDictionary(ObjectPool<PooledDictionary<K, V>>? pool, IEqualityComp
             return result;
         }
 
+        public ImmutableDictionary<TKey, TValue> ToImmutableDictionaryAndFree<TKey, TValue>(
```
Evangelink: I have kept the 2 full implementations but I could refactor the parameterless method to call this one.

Evangelink:
The API text file hasn't been updated yet.

src/Utilities/Compiler/CodeMetrics/CodeAnalysisMetricData.cs
```diff
@@ -180,66 +179,65 @@ void appendChildren(string indent)
         }
 
         /// <summary>
-        /// Computes <see cref="CodeAnalysisMetricData"/> for the given <paramref name="compilation"/>.
+        /// Computes <see cref="CodeAnalysisMetricData"/> for the given <paramref name="context"/>.
         /// </summary>
-        public static Task<CodeAnalysisMetricData> ComputeAsync(Compilation compilation, CancellationToken cancellationToken)
+        public static Task<CodeAnalysisMetricData> ComputeAsync(CodeMetricsAnalysisContext context)
```
mavasani: Lets try to preserve the existing public API surface of code under  `CodeMetrics` folder. We insert the utilities assembly produced by the repo into VisualStudio, and this will cause a build break when we try to bump up the version number. Suggestion:

```csharp
[Obsolete("Use ComputeAsync(CodeMetricsAnalysisContext) instead")
public static Task<CodeAnalysisMetricData> ComputeAsync(Compilation compilation, CancellationToken cancellationToken)
{
   // Create new CodeMetricsAnalysisContext and invoke the new API
}

public static Task<CodeAnalysisMetricData> ComputeAsync(CodeMetricsAnalysisContext context)
{
```

src/Utilities/Compiler/Analyzer.Utilities.projitems
```diff
@@ -32,7 +32,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)CodeMetrics\CodeAnalysisMetricData.PropertyMetricData.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)CodeMetrics\ComputationalComplexityMetrics.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)CodeMetrics\MetricsHelper.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)CodeMetrics\SemanticModelProvider.cs" />
```
mavasani: Is this file deleted in this PR or is it still used somewhere?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -144,7 +144,13 @@ public override void Initialize(AnalysisContext analysisContext)
                 }
 
                 // Compute code metrics.
-                var computeTask = CodeAnalysisMetricData.ComputeAsync(compilationContext.Compilation, compilationContext.CancellationToken);
+                var inheritanceExcludedTypes = compilationContext.Options.GetInheritanceExcludedTypeNamesOption(CA1501Rule, compilationContext.Compilation, compilationContext.CancellationToken);
+
+                var metricsAnalysisContext = new CodeMetricsAnalysisContext(compilationContext.Compilation, compilationContext.CancellationToken)
+                {
+                    IsExcludedFromInheritanceCountFunc = namedType => inheritanceExcludedTypes.Contains(namedType)
```
mavasani: Just curious: any reason not to pass the delegate as the 3rd constructor argument?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -144,7 +144,13 @@ public override void Initialize(AnalysisContext analysisContext)
                 }
 
                 // Compute code metrics.
-                var computeTask = CodeAnalysisMetricData.ComputeAsync(compilationContext.Compilation, compilationContext.CancellationToken);
+                var inheritanceExcludedTypes = compilationContext.Options.GetInheritanceExcludedTypeNamesOption(CA1501Rule, compilationContext.Compilation, compilationContext.CancellationToken);
```
mavasani: Using just `CA1501Rule` seems a bit weird. Should we instead make this option rule ID agnostic, and just accept category based or generic configuration? At least we should document it such so users do not try to qualify the option key name with the rule ID.

Evangelink: We could use a syntax that is set in the additional text file instead of the `.editorconfig` (or at least maybe in addition). I don't really get the part with the rule id agnostic because that's how it's working for other rules. Besides, this option is really specific to this rule so I have some trouble with finding a generic way to handle this option.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
Evangelink: I can't say that's a good reason, I just thought that it wasn't scaling well to have all the extra func that we might add in the future as extra args.

src/Utilities/Compiler/Analyzer.Utilities.projitems
Evangelink: It's been renamed as `CodeMetricsAnalysisContext`. It was used only for code metrics so it seemed ok to do a rename instead of adding a new class.

Evangelink:
I will wait for the review to move forward before doing the update of the public API.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -102,14 +105,347 @@ End Class
...
 ";
             DiagnosticResult[] expected = new[] {
-                // Test0.vb(5, 7): warning CA1501: 'BaseClass' has an object hierarchy '2' levels deep within the defining module. If possible, eliminate base classes within the hierarchy to decrease its hierarchy level below '2': 'BaseClass, Object'
-                GetBasicCA1501ExpectedDiagnostic(5, 7, "FirstDerivedClass", 2, 2, "BaseClass, Object")};
+                GetBasicCA1501ExpectedDiagnostic(5, 7, "FirstDerivedClass", 1, 1, "BaseClass")};
             await VerifyBasicAsync(source, additionalText, expected);
         }
 
+        [Fact, WorkItem(1839, "https://github.com/dotnet/roslyn-analyzers/issues/1839")]
+        public async Task CA1501_ExcludesTypesInSystemNamespaceByDefault()
```
mavasani: I believe this test will pass even before your changes as the default allowed inheritance hierarchy is higher than 1. You probably want this test to supply an additional text like in the previous test so it fails prior to your fix and passes afterwards.

mavasani: Ah, seems like `System.Windows.Forms.UserControl` already has a base type chain that would have made this test fail before your fix. Seems fine to keep it as-is.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -102,14 +105,347 @@ End Class
...
+        {
+            var editorConfigText = "dotnet_code_quality.CA1501.inheritance_excluded_type_names = T:SomeClass";
+
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources = { "public class MyUC : System.Windows.Forms.UserControl {}" },
+                    AdditionalFiles = { (".editorconfig", editorConfigText) },
+                    ExpectedDiagnostics = { GetCSharpCA1501ExpectedDiagnostic(1, 14, "MyUC", 7, 6, "UserControl, ContainerControl, ScrollableControl, Control, Component, MarshalByRefObject, Object") },
```
mavasani: I think we probably want `additional_inheritance_excluded_symbol_names`:

1. Users can exclude even namespaces, not just types
2. I don't think users would want to remove the exclusion for `System.*` when adding a new value. We should probably just assume `System.*` is always excluded.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -102,14 +105,347 @@ End Class
...
+        [InlineData("dotnet_code_quality.CA1501.inheritance_excluded_type_names = User*ontrol")]
+        public async Task CA1501_InvalidUserConfigurationOverridesDefaultValueButIsIgnored(string editorConfigText)
+        {
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources = { "public class MyUC : System.Windows.Forms.UserControl {}" },
+                    AdditionalFiles = { (".editorconfig", editorConfigText) },
+                    ExpectedDiagnostics = { GetCSharpCA1501ExpectedDiagnostic(1, 14, "MyUC", 7, 6, "UserControl, ContainerControl, ScrollableControl, Control, Component, MarshalByRefObject, Object") },
```
mavasani: Same here, I think we should still continue excluding `System.*`

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -102,14 +105,347 @@ End Class
...
+                        GetCSharpCA1501ExpectedDiagnostic(5, 18, "C1", 2, 2, "SomeClass, Object"),
+                        GetCSharpCA1501ExpectedDiagnostic(8, 18, "SomeClass2", 2, 2, "SomeClass1, Object"),
+                        GetCSharpCA1501ExpectedDiagnostic(9, 18, "C2", 3, 2, "SomeClass2, SomeClass1, Object"),
+                    }
+                },
+            }.RunAsync();
+        }
+
+        [Fact, WorkItem(1839, "https://github.com/dotnet/roslyn-analyzers/issues/1839")]
+        public async Task CA1501_WildcardTypePrefixWithNamespace()
```
mavasani: Can this test be combined with the previous test and use InlineData and Theory attributes for different editorconfig specifications?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
Evangelink: > I don't think users would want to remove the exclusion for System.* when adding a new value. We should probably just assume System.* is always excluded.

I was thinking that people at Microsoft might be interested. Assuming they are using this analyzer.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
Evangelink: I am fine with adding it as extra ctor argument and wait until we see if we have some "scaling" problem. In the meantime this will be clearer.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
Evangelink: I have addressed point 1 let me know if you think that `System.*` exclusion shall be always activated.

Evangelink:
@sharwell When you have some time, I'd be happy to also have your input on this PR (especially for the wildcard match method).

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
mavasani: I would think the number of external people who will accidentally get affected when they configure a new exclusion would be much broader then those working in System.* namespace. We can possibly add a separate option to remove the System.* exclusion.

Evangelink: Fine by me, just as a reminder, with this change we will cause a breaking change.

Evangelink: Done

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
Evangelink: @mavasani what's your recommendation?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -144,7 +144,12 @@ public override void Initialize(AnalysisContext analysisContext)
                 }
 
                 // Compute code metrics.
-                var computeTask = CodeAnalysisMetricData.ComputeAsync(compilationContext.Compilation, compilationContext.CancellationToken);
+                var inheritanceExcludedTypes = compilationContext.Options.GetInheritanceExcludedSymbolNamesOption(CA1501Rule, compilationContext.Compilation,
+                    compilationContext.CancellationToken);
+
+                var metricsAnalysisContext = new CodeMetricsAnalysisContext(compilationContext.Compilation, compilationContext.CancellationToken,
+                    namedType => inheritanceExcludedTypes.Contains(namedType));
+                var computeTask = CodeAnalysisMetricData.ComputeAsync(metricsAnalysisContext);
```
mavasani: Can you open a tracking issue to update the command line Metrics tool to also respect this option: https://github.com/dotnet/roslyn-analyzers/blob/d9c93107393265b20a982c7445184a57e3c47003/src/Tools/Metrics/Program.cs#L319

At the minimum, we should update it to exclude System.* types. Please also make a note that we should change the internal Visual Studio code base for "Analyzer -> Calculate Code Metrics" command to also have similar change so this is consistent everywhere.

src/Tools/Metrics/Program.cs
```diff
@@ -316,7 +316,7 @@ async Task computeProjectMetricDataAsync(MSBuildWorkspace workspace, string proj
 
                 cancellation.ThrowIfCancellationRequested();
                 var compilation = await project.GetCompilationAsync(CancellationToken.None).ConfigureAwait(false);
-                var metricData = await CodeAnalysisMetricData.ComputeAsync(compilation.Assembly, compilation, CancellationToken.None).ConfigureAwait(false);
+                var metricData = await CodeAnalysisMetricData.ComputeAsync(compilation.Assembly, new CodeMetricsAnalysisContext(compilation, CancellationToken.None)).ConfigureAwait(false);
```
mavasani: Yep, we need to update this one in a follow-up PR. We can probably just link in all the files that are needed for parsing editorconfig options into this project, and create a dummy diagnostic descriptor for CA1501 to pass in the correct func here.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -205,31 +205,70 @@ static SymbolNamesOption.NameParts GetParts(string name)
             }
         }
 
+        public static SymbolNamesOption GetInheritanceExcludedSymbolNamesOption(
```
mavasani: I suggest changing this method to take the default forced value as a parameter, and let the caller provide this value.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
Evangelink: See #3275

src/Tools/Metrics/Program.cs
Evangelink: See #3275

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -41,14 +53,58 @@ private SymbolNamesOption()
...
+                    (numberOfWildcards == 1 && symbolName[symbolName.Length - 1] != WildcardChar))
+                {
+                    continue;
+                }
+
+                if (numberOfWildcards == 1)
+                {
+                    Debug.Assert(parts.SymbolName[parts.SymbolName.Length - 1] == WildcardChar);
+
+                    if (parts.SymbolName[1] != ':')
```
mavasani: Can this throw an `IndexOutOfRangeException`?

Evangelink: I guess it could for some invalid data entry. I'll go defensive.

src/Utilities/Compiler/Options/SymbolNamesOption.cs
mavasani: Can you add a doc comment with examples for the different combinations of key-value pairs that can show up in this dictionary?

docs/Analyzer Configuration.md
```diff
@@ -455,3 +455,30 @@ Examples:
...
+
+Configurable Rules: [CA1501](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1501)
+
+Option Values: Names of types or namespaces (separated by '|'), such that the type or type's namespace doesn't count in the inheritance hierarchy tree.
+Allowed symbol name formats:
+  1. Type or namespace name (includes all types with the name, regardless of the containing type or namespace and all types whose namespace contains the name)
+  2. Type or namespace name ending with a wildcard symbol (includes all types whose name starts with the given name, regardless of the containing type or namespace and all types whose namespace contains the name)
+  3. Fully qualified names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format with an optional "T:" prefix for types or "N:" prefix for namespaces.
+  4. Fully qualified wildcard names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format with an optional "T:" prefix for types or "N:" prefix for namespaces and ending with the wildcard symbol.
```
mavasani: This is not completely true, right? We support all prefixes from the symbol's documentation ID format, but will fail if say someone were to provide `M:MyMethod(System.String)`, which is a valid documentation ID format, but not something we support as we only do a symbol name based search by walking the symbol's containing symbol chain.

Maybe better to not mention documentation ID format in this bullet point, and just state fully qualified name?

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -134,5 +134,19 @@ internal static partial class EditorConfigOptionNames
...
+        /// <summary>
+        /// Names of types or namespaces (separated by '|'), such that the type or type's namespace doesn't count in the inheritance hierarchy tree.
+        /// Configurable rules: CA1501 (https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1501)
+        /// Allowed name formats:
+        ///   1. Type or namespace name (includes all types with the name, regardless of the containing type or namespace and all types whose namespace contains the name)
+        ///   2. Type or namespace name ending with a wildcard symbol (includes all types whose name starts with the given name, regardless of the containing type or namespace
+        ///      and all types whose namespace contains the name)
+        ///   3. Fully qualified names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format
+        ///      with an optional "T:" prefix for types or "N:" prefix for namespaces.
+        ///   4. Fully qualified wildcard names in the symbol's documentation ID format: https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format
```
mavasani: Should also fix up this comment if you update the documentation file.

src/Utilities/Compiler/Options/SymbolNamesOption.cs
```diff
@@ -41,14 +53,58 @@ private SymbolNamesOption()
...
+            var wildcardNamesBuilder = PooledDictionary<SymbolKind, PooledDictionary<string, string?>>.GetInstance();
 
             foreach (var symbolName in symbolNames)
             {
                 var parts = getSymbolNamePartsFunc != null
                     ? getSymbolNamePartsFunc(symbolName)
                     : new NameParts(symbolName);
 
-                if (parts.SymbolName.Equals(".ctor", StringComparison.Ordinal) ||
+                var numberOfWildcards = symbolName.Count(c => c == WildcardChar);
```
mavasani: Nit: this method seems to have become big enough to warrant breaking it into different local functions for each bucket (symbols, names, wildcard based names)

src/Utilities/Compiler/Options/SymbolNamesOption.cs
mavasani: Yeah, agreed, we need to build up a cache to avoid iterating over the same containing types and namespaces again. I think you can use data structures and logic similar to below used in RestrictedIVT analyzer: https://github.com/dotnet/roslyn-analyzers/blob/3cc0345042a39d23c8b82fd02d0d2dd298dc92cb/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/RestrictedInternalsVisibleToAnalyzer.cs#L196-L258

Maintain a lazyily computed dictionary, say `ConcurrentDictionary<INamespaceOrTypeSymbol, /*isWildCardMatch*/bool>`, and update it with true or false value whenever you encounter a new containing type or namespace.

mavasani: I think rest of the logic looks fine. We can perf tune the rest if we see performance issues.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
Evangelink: Ok so we don't always for the `N:System.*` pattern, do we?

docs/Analyzer Configuration.md
Evangelink: Sounds good to me but we have the same _incorrect_ behavior for all the recently added options.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
Evangelink: Nevermind this was a stupid comment.

docs/Analyzer Configuration.md
mavasani: > we have the same incorrect behavior for all the recently added options

Which ones? I thought none of the previous options were based on wildcards and always used the public APIs on DocumentationCommentId to find the matching symbol for a specified name in documentation ID format.

docs/Analyzer Configuration.md
```diff
@@ -474,3 +474,30 @@ Examples:
...
+### Inheritance excluded type or namespace names
+Option Name: `additional_inheritance_excluded_symbol_names`
+
+Configurable Rules: [CA1501](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1501)
+
+Option Values: Names of types or namespaces (separated by '|'), such that the type or type's namespace doesn't count in the inheritance hierarchy tree.
+Allowed symbol name formats:
+  1. Type or namespace name (includes all types with the name, regardless of the containing type or namespace and all types whose namespace contains the name)
+  2. Type or namespace name ending with a wildcard symbol (includes all types whose name starts with the given name, regardless of the containing type or namespace and all types whose namespace contains the name)
+  3. Fully qualified type or namespace name with an optional "T:" prefix for type or "N:" prefix for namespace (includes all types with the exact type match or the exact containing namespace match)
```
mavasani: Why was the 3rd bullet point changed to exclude documentation ID format and the help link? Only the wildcard based option value does not abide by that format. Non-wildcard based options still must be in that format and we should document these similar to other options.

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -143,5 +143,17 @@ internal static partial class EditorConfigOptionNames
...
         public const string AdditionalRequiredGenericInterfaces = "additional_required_generic_interfaces";
+
+        /// <summary>
+        /// Names of types or namespaces (separated by '|'), such that the type or type's namespace doesn't count in the inheritance hierarchy tree.
+        /// Configurable rules: CA1501 (https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1501)
+        /// Allowed name formats:
+        ///   1. Type or namespace name (includes all types with the name, regardless of the containing type or namespace and all types whose namespace contains the name)
+        ///   2. Type or namespace name ending with a wildcard symbol (includes all types whose name starts with the given name, regardless of the containing type or namespace
+        ///      and all types whose namespace contains the name)
+        ///   3. Fully qualified type or namespace name with an optional "T:" prefix for type or "N:" prefix for namespace (includes all types with the exact type match or the exact containing namespace match)
```
mavasani: Need to restore even this comment to mention documentation ID format for 3.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -43,89 +81,287 @@ private SymbolNamesWithValueOption()
...
 
         public bool Equals(SymbolNamesWithValueOption<TValue>? other)
-            => other != null && _names.IsEqualTo(other._names) && _symbols.IsEqualTo(other._symbols);
+            => other != null && _names.IsEqualTo(other._names) && _symbols.IsEqualTo(other._symbols) && _wildcardNamesBySymbolKind.IsEqualTo(other._wildcardNamesBySymbolKind);
 
         public override int GetHashCode()
-            => HashUtilities.Combine(HashUtilities.Combine(_names), HashUtilities.Combine(_symbols));
+            => HashUtilities.Combine(HashUtilities.Combine(_names), HashUtilities.Combine(_symbols), HashUtilities.Combine(_wildcardNamesBySymbolKind));
+
+        private bool TryGetFirstWildcardMatch(ISymbol symbol, out KeyValuePair<string, TValue> firstMatch)
```
mavasani: I see a bunch of return statements in this method where nothing is added/updated in the cache, while some of them do. It is also not clear if the intention is to cache values only for namespace and type symbols or for all symbols. I think former might be more suitable to keep the cache size to reasonable.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -43,89 +81,287 @@ private SymbolNamesWithValueOption()
...
+
+                firstMatch = _wildcardNamesBySymbolKind[kind].FirstOrDefault(x => symbolName.StartsWith(x.Key, StringComparison.Ordinal));
+
+                if (string.IsNullOrWhiteSpace(firstMatch.Key))
+                {
+                    return false;
+                }
+
+                var match = firstMatch;
+                _wildcardMatchResult.AddOrUpdate(symbol, firstMatch, (s, kvp) => match);
```
mavasani: This method is another example where it is not clear to me when and why cache is updated and why is it not updated on the the false returning paths.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -43,89 +81,287 @@ private SymbolNamesWithValueOption()
...
+            if (TryGetSymbolPartialMatch(symbolFullNameBuilder, symbol, out firstMatch))
+            {
+                return true;
+            }
+
+            // ...Now try a partial check looping on the symbol's containing type...
+            INamedTypeSymbol? currentType = symbol.ContainingType;
+            while (currentType != null)
+            {
+                if (TryGetSymbolPartialMatch(symbolFullNameBuilder, currentType, out firstMatch))
```
mavasani: I would have taken a much more simpler implementation approach. The cache is defined with `INamespaceOrTypeSymbol` as the key. In both the while loops here that walk the containing type and namespace chain, use `_wildcardMatchResult.GetOrAdd` to get this bool. The func passed for Add would invoke TryGetSymbolPartialMatch. `_wildcardMatchResult` does not need to be used at any other place except inside these two loops.

Additionally, also consider making the local functions static in this method to avoid accidental capture or update. You can probably pass `_wildcardNamesBySymbolKind` as an explicit argument to local function invocations.

docs/Analyzer Configuration.md
Evangelink: I think I got confused to what was your suggestion. I thought that you wanted me to remove the documentation ID reference for all bullet points because we were accepting only types and namespaces. If this was your suggestion then yes we have added a couple of rules for which we don't support all member kinds allowed by the doc.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
Evangelink: @mavasani Shall we ignore this point for now? Is it worth adding a ticket?

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
Evangelink: You are correct I have missed some parts and I forgot to reuse the cache at the various steps.

docs/Analyzer Configuration.md
Evangelink: Done

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
Evangelink: Done

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -244,31 +244,72 @@ static SymbolNamesWithValueOption<INamedTypeSymbol>.NameParts GetParts(string na
             }
         }
 
+        public static SymbolNamesWithValueOption<Unit> GetInheritanceExcludedSymbolNamesOption(
+            this AnalyzerOptions options,
+            DiagnosticDescriptor rule,
+            Compilation compilation,
+            string defaultForcedValue,
```
Evangelink: There is actually a problem with having `defaultForcedValue` as a method parameter. Because of the caching mechanism if there are multiple calls for the same rule with different values only the first one will be taken into account. I am facing this problem when trying to mimic this for #3212

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -144,7 +144,12 @@ public override void Initialize(AnalysisContext analysisContext)
                 }
 
                 // Compute code metrics.
-                var computeTask = CodeAnalysisMetricData.ComputeAsync(compilationContext.Compilation, compilationContext.CancellationToken);
+                var inheritanceExcludedTypes = compilationContext.Options.GetInheritanceExcludedSymbolNamesOption(CA1501Rule, compilationContext.Compilation,
```
mavasani: Nit: add a comment on why `System.*` is being excluded?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
mavasani: Yes, possibly. I am fine to wait till we see a performance issue.

src/Utilities.UnitTests/Options/SymbolNamesWithValueOptionTests.cs
```diff
@@ -0,0 +1,361 @@
...
+
+            var options = SymbolNamesWithValueOption<Unit>.Create(symbolNames, compilation, null, null);
+
+            // Act
+            var isFound = options.Contains(symbol);
+
+            // Assert
+            Assert.True(isFound);
+            Assert.True(options._wildcardMatchResult.ContainsKey(symbol));
+        }
```
mavasani: Thanks for writing these exhaustive unit tests!

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -1,34 +1,73 @@
...
+using System.Linq;
+using System.Text;
 using Analyzer.Utilities.Extensions;
 using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis;
 
 namespace Analyzer.Utilities
 {
-    internal sealed class SymbolNamesWithValueOption<TValue> : IEquatable<SymbolNamesWithValueOption<TValue>?>
+    internal class SymbolNamesWithValueOption<TValue> : IEquatable<SymbolNamesWithValueOption<TValue>?>
```
mavasani: Nit: can this still be sealed?

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
mavasani: I still find the logic very confusing. We should not have so many points in code where you are updating or reading from the cache. There should be a single GetOrCompute point at the top which accesses the cache, and then invokes a core compute function which does not access/update the cache. Let me know if you'd prefer if I create a simple commit on top of your branch to convey what I mean.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
Evangelink: Oh yes! I removed the `sealed` because I wanted first to go for an inherited class for tests but then decided to go for `internal` instead.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
Evangelink: I haven't finished the refactoring, for example methods and indexers properties are matching is still not working properly so I will do one more commit to try and handle this. 

I am more than happy for you to do (a) commit(s) on this PR or any. This is exactly the kind of "problem solving" situation where a chat and a whiteboard are useful.

src/Utilities.UnitTests/Options/SymbolNamesWithValueOptionTests.cs
Evangelink: You are welcome. The code I have added is too complex so adding tests only through the actual rules wasn't giving me enough confidence.

Evangelink:
It might be worth marking this PR as `blocked` while we figure out how to improve the wildcard matching algo.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
mavasani: > for example methods and indexers properties are matching is still not working properly

I think the hot path here is for containing types and symbols. IMO, it should be fine to start with caching just those symbol kinds.

mavasani: @Evangelink I think I have blocked this PR for long enough now without much action from my side. Lets do the following to get traction here:

1. Given this code path is only taken when user provides a wild card option value, it will not affect normal analysis scenarios. So any performance concern here is non-blocking for the user. I think that is sufficient to postpone the performance tuning and cleanup work for this code path.
2. Can you please file a tracking issue with following notes and tag me?
   1. Consider revisiting this code for code refactoring/cleanup for easier future maintenance.
   2. Consider tuning the cache size and hit/misses here to improve performance.
3. Please resolve the merge conflicts on the PR. I am going to mark this PR approved and move forward with the merge.

mavasani:
@Evangelink There are build failures.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
Evangelink: See #3403

Evangelink:
Documentation update request: https://github.com/MicrosoftDocs/visualstudio-docs/issues/4984

Issue #2579, opened 2019-06-13
Convert CompilationEnd analyzers to SymbolStart/SymbolEnd analyzers


mavasani:
We should attempt to minimize the use of compilation end API - majority of these can be converted into SymbolStart/SymbolEnd analyzers, this new analyzer API is available on Microsoft.CodeAnalysis 3.0.0. This will enable us to convert these into build + intellisense analyzers, instead of them being build-only analyzers.

mavasani:
None of the analyzers in Microsoft.CodeAnalysis.NetAnalyzers and Microsoft.CodeAnalysis.FxCopAnalyzers are now based on CompilationEnd action. I am going to mark this work as complete, and track any similar work in other analyzer packages separately, if and when required.

PR #3235, opened 2020-01-22
Rewrite CA2213 (DisposableFieldsShouldBeDisposed) analyzer to be a SymbolStart/End analyzer

mavasani:
Work towards reducing CompilationEnd analyzers in the repo: #2579


mavasani:
FYI: Much easier to review with https://github.com/dotnet/roslyn-analyzers/pull/3235/files?w=1

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposableFieldsShouldBeDisposed.cs
```diff
@@ -41,172 +41,217 @@ public override void Initialize(AnalysisContext context)
...
-                compilationContext.RegisterOperationAction(_ => hasErrors = true, OperationKind.Invalid);
+            // Instance fields initialized in constructor/method body with a locally created disposable object must be disposed.
+            symbolStartContext.RegisterOperationBlockStartAction(OnOperationBlockStart);
 
-                // Disposable fields with initializer at declaration must be disposed.
-                compilationContext.RegisterOperationAction(operationContext =>
+            // Report diagnostics at symbol end.
+            symbolStartContext.RegisterSymbolEndAction(OnSymbolEnd);
+
+            return;
```
Evangelink: Do you need this return?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposableFieldsShouldBeDisposed.cs
```diff
@@ -41,172 +41,217 @@ public override void Initialize(AnalysisContext context)
...
-                        }
-                    }
-                },
-                OperationKind.FieldInitializer);
+                return false;
+            }
+
+            void OnFieldInitializer(OperationAnalysisContext operationContext)
+            {
+                RoslynDebug.Assert(disposeAnalysisHelper != null);
```
Evangelink: Is there any guideline on when to use `RoslynDebug` or `Debug`?

mavasani: `RoslynDebug.Assert` is something that is nullable annotated, so we use it only when we get a false nullable warning from the C# compiler. Other alternative is to use the `!` suppression operator at use sites for this variable.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposableFieldsShouldBeDisposed.cs
mavasani: It is generally recommended to have an explicit return statement separating the core logic and local functions, which are primarily implementation details. We do so even in roslyn repo.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposableFieldsShouldBeDisposed.cs
Evangelink: Thanks

Issue #1818, opened 2018-09-17
Classes implementing IDataReader should not have have enumerable diagnostics


jnm2:
Analyzer package: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)
Package Version: 2.6.2-beta2
Diagnostic ID: CA1010, CA1710

When implementing DbDataReader, IDataReader, or maybe even the base interface IDataRecord, it's rare to treat it as a collection even though IEnumerable is a base interface of IDataRecord. The name of a data reader should not end with 'Collection' and implementing the generic `IEnumerable<>` is not possible since the schema varies at runtime.

Repro code:

```c#
public sealed class CustomDbDataReader : DbDataReader
{
    // Ordinary DbDataReader implementation 
}
```

> CA1010 Collection 'CustomDbDataReader' directly or indirectly inherits 'IEnumerable' without implementing 'IEnumerable\<T>'. Publicly-visible collections should implement the generic version to broaden usability.

> CA1710 Rename CustomDbDataReader to end in 'Collection'.

Evangelink:
@mavasani I was starting to tackle this issue when I realized that it might be better to ask for a `DataReader` suffix instead of skipping data reader classes. WDYT?

Evangelink:
@mavasani I'd like your input here before doing a PR in a direction or another.

Evangelink:
@mavasani ping

Evangelink:
@mavasani Same here. With the decision on #3065 I think this issue shall be addressed.

mavasani:
This rule is now configurable: https://github.com/dotnet/roslyn-analyzers/pull/3148

jnm2:
@mavasani Are classes implementing `IDataReader` now excluded from these diagnostics out of the box?

Evangelink:
They are not but you now can add a custom suffix. @mavasani I actually didn't allow empty string, it might be better to allow it to help providing the exclusion feature.

mavasani:
That sounds reasonable - might be worth also taking a compat break and having the default value for [exclude-indirect-base-types](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#exclude-indirect-base-types) be `true`, given that seems to be a more reasonable behavior.

Evangelink:
@mavasani This can now be closed.

@jnm2 This is now excluded by default.

PR #3237, opened 2020-01-23
Update CA1710 to allow empty suffix

Evangelink:
Contributes to #1818 and allows for empty suffix as a way to override some of the built-in suffixes.


docs/Analyzer Configuration.md
```diff
@@ -452,5 +452,6 @@ Examples:
 
 | Option Value | Summary |
 | --- | --- |
-|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass->Class` | All types inheriting from MyClass are expected to have the 'Class' suffix |
-|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass->Class|MyNamespace.IPath->Path` | All types inheriting from 'MyClass' are expected to have the 'Class' suffix AND all types implementing 'MyNamespace.IPath' are expected to have the 'Path' suffix. |
\ No newline at end of file
+|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass->Class` | All types inheriting from' MyClass' are expected to have the 'Class' suffix |
```
Evangelink: I have added single quotes around `MyClass`.

docs/Analyzer Configuration.md
```diff
@@ -452,5 +452,6 @@ Examples:
...
+|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass->Class|MyNamespace.IPath->Path` | All types inheriting from 'MyClass' are expected to have the 'Class' suffix AND all types implementing 'MyNamespace.IPath' are expected to have the 'Path' suffix. |
+|`dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader->` | Allows to override built-in suffixes, in this case, all types implementing 'IDataReader' are no longer expected to end in 'Collection' |
```
mavasani: I think we should not require the input to have a trailing `->`, so even the below setting should produce the same result.
```suggestion
|`dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader` | Allows to override built-in suffixes, in this case, all types implementing 'IDataReader' are no longer expected to end in 'Collection' |
```

mavasani: Or maybe for clarity we can allow some special character after the `->`, say `_` or some such?

Evangelink: I like the second option.

Evangelink: The symbol `_` was a good way to express discard but the problem is that it is a valid class name suffix (in term of what is allowed) even if it's not a nice suffix.

docs/Analyzer Configuration.md
mavasani:
```suggestion
|`dotnet_code_quality.CA1710.additional_required_suffixes = MyClass->Class` | All types inheriting from 'MyClass' are expected to have the 'Class' suffix |
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1380,6 +1380,261 @@ Inherits Dictionary(Of String, String)
             }.RunAsync();
         }
 
+        [Fact, WorkItem(1818, "https://github.com/dotnet/roslyn-analyzers/issues/1818")]
+        public async Task CA1710_AllowEmptySuffix()
+        {
+            var editorConfigText = @"dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader->{}";
```
mavasani: Can you make this a Theory based test and also verify the behavior for different use of whitespaces, say `T:System.Data.IDataReader -> {}` and `T:System.Data.IDataReader-> { } `?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffixTests.cs
```diff
@@ -1452,6 +1452,283 @@ End Sub
...
+        // Following values are ok
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader->{}")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader-> {}")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader->{} ")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader-> {} ")]
+        // Following values are not ok
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader->{ }")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader-> { }")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader->{ } ")]
+        [InlineData("dotnet_code_quality.CA1710.additional_required_suffixes = T:System.Data.IDataReader-> { } ")]
```
Evangelink: Not sure if you were expecting these cases to be supported or not.

mavasani: All of these values should work. We should trim the leading and trailing whitespaces in each value.

Evangelink: I'm already doing a trim (start and end). The values I have highlighted here are the ones for which the user has put space between the curly braces. Is it something we want to handle? If so how many spaces shall be allowed?

mavasani: IMO All spaces should be ignored. I think you can handle trimmed values starting and ending with curliest and any number of spaces in between as empty value.

PR #3244, opened 2020-01-24
Update CA2000 to exclude some types

### LINKED ISSUE #3085: Feature Request: IDE0067/CA2000 - Exclude only certain types
I think both IDE0067 and CA2000 are too valuable to turn off.

However, sometimes types that don't really have unmanaged ressources implement IDisposable for whatever reasons. I cant change those classes, they are part of a library. I can exclude them but manually setting pragmas around hundreds and more cases is really detrimental to the readability of the code.

**My wish:**
Being able to configure IDE0067 and CA2000 to exclude types, like they exclude Task.

PS:
The type in question is HalconDotNet.HTuple. Pretty niche, i know.

Evangelink:
Fix #3085

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
@@ -11774,5 +11774,88 @@ class Test
     }
 }");
         }
+
+        [Fact, WorkItem(3085, "https://github.com/dotnet/roslyn-analyzers/issues/3085")]
+        public async Task LocalInvocationOfAnExcludedType_NoDiagnostic()
+        {
+            string editorConfigText = $"dotnet_code_quality.{DisposeObjectsBeforeLosingScope.RuleId}.excluded_symbol_names = T:A";
+
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources =
+                    {
+                        @"
+using System;
+
+class A : IDisposable
+{
+    public void Dispose()
+    {
+    }
+}
+
+class B : IDisposable
+{
+    public void Dispose()
+    {
+    }
+}
+
+class Test
+{
+    void M1()
+    {
+        var a = new A();
+        var b = new B();
+    }
+}
+",
+                    },
+                    AdditionalFiles = { (".editorconfig", editorConfigText) },
+                    ExpectedDiagnostics =
+                    {
+                        GetCSharpResultAt(23, 17, "new B()"),
+                    }
+                }
+            }.RunAsync();
+
+            await new VerifyVB.Test
+            {
+                TestState =
+                {
+                    Sources =
+                    {
+                        @"
+Imports System
+
+Class A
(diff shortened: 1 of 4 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.cs
```diff
@@ -62,6 +62,7 @@ private DisposeAnalysis(DisposeAnalysisDomain analysisDomain, DisposeDataFlowOpe
                 disposeOwnershipTransferLikelyTypes, disposeOwnershipTransferAtConstructor,
                 disposeOwnershipTransferAtMethodCall, trackInstanceFields, exceptionPathsAnalysis,
                 performCopyAnalysis: analyzerOptions.GetCopyAnalysisOption(rule, defaultValue: performCopyAnalysisIfNotUserConfigured, cancellationToken),
+                excludedSymbols: analyzerOptions.GetExcludedSymbolNamesOption(rule, wellKnownTypeProvider.Compilation, cancellationToken),
```
Evangelink: This is the "deepest" location where we still have access to the diagnostic.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
```diff
@@ -11774,5 +11774,88 @@ class Test
     }
 }");
         }
+
+        [Fact, WorkItem(3085, "https://github.com/dotnet/roslyn-analyzers/issues/3085")]
+        public async Task LocalInvocationOfAnExcludedType_NoDiagnostic()
```
Evangelink: Do we want more tests for the syntax of `excluded_symbol_names`?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
```diff
@@ -97,6 +97,12 @@ protected override DisposeAbstractValue HandleInstanceCreation(IOperation creati
                     return defaultValue;
                 }
 
+                // Handle user option for additional excluded types
+                if (DataFlowAnalysisContext.ExcludedSymbols.Contains(instanceType))
```
Evangelink: I am only handling fully qualified types and type names, let me know if you think that's worth also handling namespaces.

mavasani: I think this is fine.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
mavasani: I think one test is good enough here.

PR #3246, opened 2020-01-27
Convert last tests to use Test.Utilities.CSharpCodeFixVerifier

### LINKED ISSUE #3206: Test framework - difference in expected diagnostics
Doing some cleanup in the tests I have noticed that the behavior of `Microsoft.CodeAnalysis.CSharp.Testing.XUnit.CodeFixVerifier` and `Test.Utilities.CSharpCodeFixVerifier` is different. The same is true for the `VisualBasic` pendant.

For example, in the class `DoNotMixAttributesFromDifferentVersionsOfMEFTests`, the test `NoDiagnosticCases_UnresolvedTypes` will return 2 diagnostics with the first verifier and only 1 with the second. In the class `PartsExportedWithMEFv2MustBeMarkedAsSharedTests`, the test `NoDiagnosticCases_UnresolvedTypes` will return 5 diagnostics with the first and none with the second.

These 2 tests are the only failing tests but I suspect that we might have more non-failing tests impacted by this difference.

Tagging @sharwell as you know quite a lot on the test framework.

Evangelink:
Fix #3206 thanks to @sharwell 's comment.

### DIFF
--- a/src/Test.Utilities/AdditionalMetadataReferences.cs
+++ b/src/Test.Utilities/AdditionalMetadataReferences.cs
@@ -36,9 +36,6 @@ namespace Test.Utilities
         public static ReferenceAssemblies DefaultWithWinForms { get; } = Default
             .AddAssemblies(ImmutableArray.Create("System.Windows.Forms"));
 
-        public static ReferenceAssemblies DefaultWithoutCodeAnalysis { get; } = ReferenceAssemblies.Default
-            .AddAssemblies(ImmutableArray.Create("System.Xml.Data"));
-
         public static MetadataReference SystemCollectionsImmutableReference { get; } = MetadataReference.CreateFromFile(typeof(ImmutableHashSet<>).Assembly.Location);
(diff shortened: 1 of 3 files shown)


src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -36,6 +36,9 @@ public static class AdditionalMetadataReferences
         public static ReferenceAssemblies DefaultWithWinForms { get; } = Default
             .AddAssemblies(ImmutableArray.Create("System.Windows.Forms"));
 
+        public static ReferenceAssemblies DefaultWithoutCodeAnalysis { get; } = ReferenceAssemblies.Default
+            .AddAssemblies(ImmutableArray.Create("System.Xml.Data"));
```
sharwell: 💭 At this point, would it make sense to just use `ReferenceAssemblies.Default` directly?

Evangelink: Makes sense to me.

PR #3247, opened 2020-01-27
Remove DiagnosticAnalyzerTestBase

Evangelink:
Fix #3089

PR #3249, opened 2020-01-29
Remove DiagnosticAnalyzerTestBase

Evangelink:
Relates to #3247

### DIFF
--- a/src/Utilities.UnitTests/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysisTests.cs
+++ b/src/Utilities.UnitTests/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysisTests.cs
@@ -6,6 +6,7 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -14,6 +15,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PointsToAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.ValueContentAnalysis;
+using Microsoft.CodeAnalysis.Testing;
 using Microsoft.CodeAnalysis.Text;
 using Test.Utilities;
 using Xunit;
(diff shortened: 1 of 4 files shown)


src/Utilities.UnitTests/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysisTests.cs
```diff
@@ -1274,15 +1276,161 @@ public PropertySetAnalysisTests(ITestOutputHelper output)
...
-        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+        protected static readonly CompilationOptions s_CSharpDefaultOptions = new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
+
+        internal const string DefaultFilePathPrefix = "Test";
+        internal const string CSharpDefaultFileExt = "cs";
+        protected static readonly string CSharpDefaultFilePath = DefaultFilePathPrefix + 0 + "." + CSharpDefaultFileExt;
+
+        private const string TestProjectName = "TestProject";
+
+        protected static Project CreateProject(string[] sources)
```
mavasani: Actually, is this code copied from DiagnosticAnalyzerTestBase? If so, is it being used anywhere now? If not, can it be deleted from DiagnosticAnalyzerTestBase?

mavasani: I see that the helpers added here are still used at other places in 2.9.x branch. In that case, I don't think it makes sense to clone them here in 2.9.x and it seems reasonable to make this clone in master branch in the PR that deletes `DiagnosticAnalyzerTestBase`. Sorry for the extra work here.

mavasani: I'd let @dotpaul decide what he prefers though, since he is likely the one to make changes to this file and would have to handle merge conflicts in future.

sharwell: Yeah, this is a weird situation. A copy of the code needs to be made in security analyzer tests (with these changes typically occurring in 2.9.x), but the overall refactoring is only applicable in the master branch. Overall I believe the code copy in 2.9.x is most consistent with other work to date. It fully eliminates the security analyzers' use of `DiagnosticAnalyzerTestBase` in 2.9.x, and then we will update master for the remaining tests.

mavasani: Sounds reasonable to me.

dotpaul: Sounds fine to me too.

PR #3255, opened 2020-01-31
Remove Humanizer based enum name pluralization rules from NetAnalyzer…

mavasani:
…s package

This allows us to reduce an additional dependency from this package. In future, if we get customer requests for these rules, we may implement a local algorithm for pluralization logic OR add back the reference to Humanizer.

Given that we still ship these rules in FxCopAnalyzers package, and we don't have separate unit tests for analyzers shipping in that package, I have retained the analyzer and its unit test in the NetAnalyzers unit test project.


src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -63,149 +63,147 @@ Sr. No. | Rule ID | Title | Category | Enabled | Severity | CodeFix | Descriptio
...
-206 | [CA5401](https://docs.microsoft.com/visualstudio/code-quality/ca5401) | Do not use CreateEncryptor with non-default IV | Security | False | Warning | False | Symmetric encryption should always use a non-repeatable initialization vector to prevent dictionary attacks. |
-207 | [CA5402](https://docs.microsoft.com/visualstudio/code-quality/ca5402) | Use CreateEncryptor with the default IV  | Security | False | Warning | False | Symmetric encryption should always use a non-repeatable initialization vector to prevent dictionary attacks. |
-208 | [CA5403](https://docs.microsoft.com/visualstudio/code-quality/ca5403) | Do not hard-code certificate | Security | False | Warning | False | Hard-coded certificates in source code are vulnerable to being exploited. |
+63 | [CA1715](https://docs.microsoft.com/visualstudio/code-quality/ca1715) | Identifiers should have correct prefix | Naming | True | Hidden | False | Identifiers should have correct prefix |
```
sharwell: We need to get rid of the numbers down the left side...

mavasani: Yes :)

PR #3257, opened 2020-01-31
CA1810: Field assignment from event subscription in static ctor

### LINKED ISSUE #3138: CA1810 should not warn when assigning field in event handler.
```cs
class C
{
    private static string? s;

    //     ↓ CA1810 should not warn here
    static C()
    {
        Console.CancelKeyPress += (o, e) => s = string.Empty;
    }
}
```

### LINKED ISSUE #3544: Convert analyzer CA1810 to be based on IOperation
Ideally, we would just convert this analyzer to be based on IOperation, and that would avoid language specific checks and also improve its performance. We can then also use our existing helpers to check if an operation has an ancestor operation with `OperationKind.AnonymousFunction`. However, that change would probably be outside the scope of this PR. @Evangelink - can you file a separate issue to convert this to an operation analyzer?

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/pull/3257_

Evangelink:
…tic ctor

Fix #3138 and fix #3544


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
```diff
@@ -155,6 +156,23 @@ End Class
 ");
         }
 
+        [Fact, WorkItem(3138, "https://github.com/dotnet/roslyn-analyzers/issues/3138")]
+        public async Task CA1810_EventSubscriptionInConstructor()
```
Evangelink: I couldn't replicate the use case for vb.net (cannot access to `s` from the local lambda declaration).

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.cs
```diff
@@ -17,7 +17,10 @@ public sealed class CSharpInitializeStaticFieldsInlineAnalyzer : InitializeStati
         protected override bool InitialiesStaticField(SyntaxNode node, SemanticModel semanticModel, CancellationToken cancellationToken)
         {
             var assignmentNode = (AssignmentExpressionSyntax)node;
-            return semanticModel.GetSymbolInfo(assignmentNode.Left, cancellationToken).Symbol is IFieldSymbol leftSymbol && leftSymbol.IsStatic;
+
+            return assignmentNode.FirstAncestorOrSelf<AssignmentExpressionSyntax>(x => x.IsKind(SyntaxKind.AddAssignmentExpression)) == null &&
```
Evangelink: Handles only add assignment because I couldn't find any valid use-case where we would like to unsubscribe from the static ctor.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
sharwell: 💭 

```
Public Class C
    Private Shared s As string
    
    Shared Sub New()
        AddHandler Console.CancelKeyPress,
            Sub(o, e)
                s = ""
            End Sub
    End Sub

End Class
```

Evangelink: Stupid me, I missed the `Shared` on the field declaration...

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.cs
sharwell: Why not generalize it to any lambda?

Evangelink: I was thinking about that but I thought that would be too wide exclusion. Most of the lambdas could be put in function removing this issue. But I'm happy to change the implem.

sharwell: Do you have an example of such a lambda? It seems like CA1810 should be handling the "easy" cases, while more complicated cases (enforcing the lack of `beforefieldinit`) should be covered by PerformanceSensitiveAnalyzers.

Evangelink: I was thinking about some patterns I have seen where people did initialization of `Func` and `Action` in the static ctor. I don't really think that matters much especially in regard to the second part of your comment.

sharwell: Ah, I wasn't referring to the initialization of a field with a delegate type. I was referring to a lambda in the static constructor which contains a nested statement that assigns a value to a field. That assignment cannot trivially be moved out of the lambda to become an inline initialization.

sharwell: Example:

```
class C {
  static int s;
  static C() {
    Task.Run(() => s = 3);
  }
}
```

Evangelink: Makes sense, I'll update the PR

Evangelink:
@sharwell it seems that the `re-request review` button is not working on this PR so I am pinging you.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
```diff
@@ -169,6 +169,7 @@ class C
     static C()
     {
         Console.CancelKeyPress += (o, e) => s = string.Empty;
+        Console.CancelKeyPress -= (o, e) => s = string.Empty;
```
sharwell: 💡 I see tests for most lambda forms, but would like to see one for anonymous function syntax:

```csharp
delegate { s = string.Empty; }
```
```

File: src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.cs [hunk 4]
```diff
@@ -18,7 +18,7 @@ protected override bool InitialiesStaticField(SyntaxNode node, SemanticModel sem
         {
             var assignmentNode = (AssignmentExpressionSyntax)node;
 
-            return assignmentNode.FirstAncestorOrSelf<AssignmentExpressionSyntax>(x => x.IsKind(SyntaxKind.AddAssignmentExpression)) == null &&
+            return assignmentNode.FirstAncestorOrSelf<LambdaExpressionSyntax>() == null &&
```
sharwell: @mavasani Do we have a helper method for "is inside lambda or anonymous function"?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicInitializeStaticFieldsInline.vb
```diff
@@ -21,16 +21,12 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
         Protected Overrides Function InitialiesStaticField(node As SyntaxNode, semanticModel As SemanticModel, cancellationToken As CancellationToken) As Boolean
             Dim assignmentStatement = DirectCast(node, AssignmentStatementSyntax)
 
-            If assignmentStatement.FirstAncestorOrSelf(Of AddRemoveHandlerStatementSyntax)(AddressOf IsAddHandler) IsNot Nothing Then
+            If assignmentStatement.FirstAncestorOrSelf(Of LambdaExpressionSyntax)() IsNot Nothing Then
```
sharwell: 💭 The same would apply here for using a helper

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.cs
mavasani: I think we have an IOperation based helper for it, but not the syntax one. I know we have such a helper in Roslyn repo though.

Evangelink: Are you thinking about this? https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Syntax/LambdaUtilities.cs#L135

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
Evangelink: What's the equivalent for vbnet?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.cs
Evangelink: Actually might be this one https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Syntax/LambdaUtilities.cs#L17

Evangelink: Let me know which helper you were thinking about so that I can add it to this PR and get it in.

Evangelink:
Did you guys managed to find which helper shall be used?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.cs
Evangelink: Ping @mavasani @sharwell

mavasani: Ideally, we would just convert this analyzer to be based on IOperation, and that would avoid language specific checks and also improve its performance. We can then also use our existing helpers to check if an operation has an ancestor operation with `OperationKind.AnonymousFunction`. However, that change would probably be outside the scope of this PR. @Evangelink - can you file a separate issue to convert this to an operation analyzer?

Evangelink: https://github.com/dotnet/roslyn-analyzers/issues/3544

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInline.cs
```diff
@@ -45,39 +46,46 @@ public abstract class InitializeStaticFieldsInlineAnalyzer<TLanguageKindEnum> :
...
-                    !methodSym.IsStatic ||
-                    methodSym.MethodKind != MethodKind.SharedConstructor)
+                if (!(context.OwningSymbol is IMethodSymbol method) ||
+                    !method.IsStatic ||
+                    method.MethodKind != MethodKind.StaticConstructor)
                 {
                     return;
                 }
 
                 var initializesStaticField = false;
```
Evangelink, line 63: Is it safe to only use a `bool` for a multi-threaded context? I have kept what was used previously.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInline.cs
```diff
@@ -45,39 +46,46 @@ public abstract class InitializeStaticFieldsInlineAnalyzer<TLanguageKindEnum> :
...
+                    if (initializesStaticField)
+                    {
+                        return;
+                    }
+
+                    var assignment = (IAssignmentOperation)context.Operation;
+
+                    if (assignment.Target is IFieldReferenceOperation fieldReference &&
+                        fieldReference.Member.IsStatic &&
+                        assignment.GetAncestor<IAnonymousFunctionOperation>(OperationKind.AnonymousFunction) is null)
```
Evangelink, line 75: Actual fix is this line.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
sharwell: 💭 I don't think there is one

Evangelink: Ok! Thanks!

Evangelink:
Ping @mavasani

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInline.cs
```diff
@@ -45,39 +46,46 @@ public abstract class InitializeStaticFieldsInlineAnalyzer<TLanguageKindEnum> :
...
 
-                codeBlockStartContext.RegisterCodeBlockEndAction(codeBlockEndContext =>
+                context.RegisterOperationBlockEndAction(context =>
                 {
                     if (initializesStaticField)
                     {
-                        DiagnosticDescriptor descriptor = methodSym.ContainingType.IsReferenceType ? CA1810Rule : CA2207Rule;
-                        Diagnostic diagnostic = Diagnostic.Create(descriptor, methodSym.Locations[0], methodSym.ContainingType.Name);
-                        codeBlockEndContext.ReportDiagnostic(diagnostic);
+                        context.ReportDiagnostic(
```
mavasani, line 85: I think we will end up reporting a diagnostic if a type has one static field that is assigned a delegate and one static field with a simple initializer. The correct fix would be to use a separate bool variable to track if any field is assigned a delegate, and if so bail out from the block end action.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
```diff
@@ -155,6 +155,200 @@ End Class
 ");
         }
 
+        [Fact, WorkItem(3138, "https://github.com/dotnet/roslyn-analyzers/issues/3138")]
+        public async Task CA1810_EventLambdas_NoDiagnostic()
```
mavasani, line 159: Please add the following unit test:

```csharp
using System;
class C
{
    private static string s1;
    private static string s2;

    static C()
    {
        Console.CancelKeyPress += (o, e) => s1 = string.Empty;
        s2 = string.Empty;
    }
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInline.cs
mavasani: Theoretically, we can still flag methods which have at least one static field initialized directly, but if user already needs a static cctor to initialize one of the fields inside a lambda, they might prefer initializing all static fields in the static cctor. Best to just bail out for the case where any one field is assigned inside a lambda.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInline.cs
```diff
@@ -45,39 +46,53 @@ public abstract class InitializeStaticFieldsInlineAnalyzer<TLanguageKindEnum> :
...
                     return;
                 }
 
                 var initializesStaticField = false;
-                codeBlockStartContext.RegisterSyntaxNodeAction(syntaxContext =>
+                var isStaticCtorMandatory = false;
+                context.RegisterOperationAction(context =>
                 {
-                    if (!initializesStaticField)
+                    if (isStaticCtorMandatory || initializesStaticField)
```
mavasani, line 67: This will lead to a race condition if first callback is for a regular static field assignment that sets initializesStaticField to true. We should remove this if statement.

PR #3258, opened 2020-01-31
CA1010: Allows users to provide additional interfaces

### LINKED ISSUE #1490: CA1010 should analyze IDictionary as well
Currently, CA1010 only reports on IList, ICollection, and IEnumerable implementations missing the generic version. We should also look into reporting on dictionary types as well. The original FxCop implementation disregards dictionary types "to reduce noise", http://index/?query=CollectionsShouldImplemen&rightProject=DesignRules&file=CollectionsShouldImplementGenericInterface.cs&line=24, but we should investigate if this will actually make the analyzer more noisy.

Evangelink:
Fix #1490


src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -128,57 +128,57 @@ static bool TryParseValue(string value, out ImmutableHashSet<TEnum> result)
             return analyzerConfigOptions.GetOptionValue(optionName, rule, uint.TryParse, defaultValue);
         }
 
-        public static SymbolNamesOption GetNullCheckValidationMethodsOption(
+        public static SymbolNamesOption<byte> GetNullCheckValidationMethodsOption(
```
Evangelink: I am not a big fan of the `<byte>` but I couldn't find anything nice except for adding a `Unit` class.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionsShouldImplementGenericInterface.cs
```diff
@@ -95,95 +109,41 @@ public override void Initialize(AnalysisContext analysisContext)
                 return;
             }
 
-            var allInterfacesStatus = default(CollectionsInterfaceStatus);
-            foreach (var @interface in namedTypeSymbol.AllInterfaces)
+            var allInterfaces = ImmutableHashSet.CreateRange(namedTypeSymbol.AllInterfaces.Select(i => i.OriginalDefinition));
```
mavasani:
```suggestion
            using var allInterfaces = PooledHashSet.GetInstance(namedTypeSymbol.AllInterfaces.Select(i => i.OriginalDefinition));
```

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -193,41 +193,80 @@ static SymbolNamesOption.NameParts GetParts(string name)
...
+                }
+
+                var matchingSymbols = DocumentationCommentId.GetSymbolsForDeclarationId(genericInterfaceFullName, compilation);
+
+                if (matchingSymbols.Length != 1 ||
+                    !(matchingSymbols[0] is INamedTypeSymbol namedType) ||
+                    namedType.TypeKind != TypeKind.Interface ||
+                    !namedType.IsGenericType)
+                {
+                    // Invalid matching type so we assume there was no associated type
```
mavasani: We may want to eventually report a diagnostic here so user can fix his configuration. It is fine to not do it here as it seems applicable for all our option values.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -128,57 +128,57 @@ static bool TryParseValue(string value, out ImmutableHashSet<TEnum> result)
...
+        public static SymbolNamesOption<byte> GetExcludedTypeNamesWithDerivedTypesOption(
             this AnalyzerOptions options,
             DiagnosticDescriptor rule,
             Compilation compilation,
             CancellationToken cancellationToken)
-            => options.GetSymbolNamesOption(EditorConfigOptionNames.ExcludedTypeNamesWithDerivedTypes, namePrefixOpt: "T:", rule, compilation, cancellationToken);
+            => options.GetSymbolNamesOption<byte>(EditorConfigOptionNames.ExcludedTypeNamesWithDerivedTypes, namePrefixOpt: "T:", rule, compilation, cancellationToken);
 
-        public static SymbolNamesOption GetDisallowedSymbolNamesOption(
+        public static SymbolNamesOption<byte> GetDisallowedSymbolNamesOption(
```
mavasani: I am wondering if the base type can be a non-generic type `SymbolNamesOption` and the sub-type can be `SymbolNamesOption<T>` so only the options which needs a key-value mapping will have to specify the type argument. What do you think?

mavasani: Probably also call it `SymbolNamesWithValuesOption<T>` or some such and move the `TryGetValue` function to this type.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
Evangelink: #3267

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionsShouldImplementGenericInterface.cs
Evangelink: Suggested overload doesn't exist but I will add it.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
Evangelink: The rename is a good point. Regarding the base class not generic, I was thinking about that but I couldn't managed to do some implementation without too much duplication. Actually I did, but that was making the base handling a dictionary with `object` as value and so potentially forcing some unneeded boxing. Feel free to make a suggestion if you think I am missing something.

src/Microsoft.CodeAnalysis.AnalyzerUtilities/PublicAPI.Unshipped.txt
```diff
@@ -709,3 +709,12 @@ virtual Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DataFlowOperationVisitor<TA
 virtual Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DataFlowOperationVisitor<TAnalysisData, TAnalysisContext, TAnalysisResult, TAbstractAnalysisValue>.VisitUnaryOperatorCore(Microsoft.CodeAnalysis.Operations.IUnaryOperation operation, object argument) -> TAbstractAnalysisValue
 virtual Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PredicatedAnalysisData<TKey, TValue>.ApplyPredicatedData(Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DictionaryAnalysisData<TKey, TValue> coreAnalysisData, Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DictionaryAnalysisData<TKey, TValue> predicatedData) -> void
 virtual Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PredicatedAnalysisData<TKey, TValue>.RemoveEntryInPredicatedData(TKey key, Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DictionaryAnalysisData<TKey, TValue> predicatedData) -> void
+Analyzer.Utilities.Unit
```
mavasani: Please make this an internal type and revert the changes to the PublicAPI files. We only want to expose public APIs for functionality, not for utilities from this package.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
mavasani: Can we have an `AbstractSymbolNamesOption` which has all the core logic, with sub-types `SymbolNamesOption` and `SymbolNamesWithValueOption<T>` maintaining their required datastructures (set in the former, and dictionary for the latter)?

Evangelink: We can but there won't be much code common between the 2, I'll add a separate commit so you can decide what you think is best.

src/Microsoft.CodeAnalysis.AnalyzerUtilities/PublicAPI.Unshipped.txt
Evangelink: Makes sense, I didn't pay attention here...

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
mavasani: Ah, in that case it is fine to keep it as is.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionsShouldImplementGenericInterface.cs
```diff
@@ -95,95 +110,45 @@ public override void Initialize(AnalysisContext analysisContext)
...
+
+                if (allInterfaces.Contains(kvp.Key) && !allInterfaces.Contains(kvp.Value))
+                {
+                    ReportDiagnostic(kvp.Key, kvp.Value);
+                    return;
+                }
             }
-            else
+
+            void ReportDiagnostic(INamedTypeSymbol @interface, INamedTypeSymbol genericInterface)
```
mavasani: Nit: please add an explicit return statement before start of local function definitions.

Evangelink: Is it for reading purpose or is there any impact?

mavasani: Yep, just a readability suggestion.

Evangelink:
@mavasani I have 2 side questions:

1. Shall I open a ticket for the rename of the class? 

> I feel that with this user option the class is no longer strictly for collections and should be renamed as it is now recommending generic interface over interface for any functionality.

2. Shall I open a ticket for the doc update?

mavasani:
Yes, both sound reasonable follow-up items to me. Thanks!

PR #3262, opened 2020-02-04
PublicAPI: Keep found APIs regardless of nullability

jcouv:
The test added in this PR was running into an issue with the test framework iterating on fixes. That's because the fixer was incorrectly removing APIs already marked with nullability information.

This PR adjusts the logic for recognizing APIs (to decide whether they were removed or not) to account for nullability format.


mavasani:
> The test added in this PR was running into an issue with the test framework iterating on fixes. That's because the fixer was incorrectly removing APIs already marked with nullability information.

@Evangelink @sharwell We may want to move the tests over to the new test framework.

mavasani:
Ah, the tests for this analyzer/fixer do seem to be on the new testing library. In that case, @jcouv do you have a repro for the infinite iteration that you can share with us to investigate?

jcouv:
> do you have a repro for the infinite iteration that you can share with us to investigate?

When I run the added tests without the fix in this PR (`TestSimpleMissingMember_Fix_WithNullability2`), then the iterative behavior of the test framework ran into a built-in limit and failed the test. The test framework prevented infinite iteration which the analyzer/fixer would have caused.

So I don't think there is anything to investigate from the side of the test framework. It behaved well and caught the issue gracefully.

PR #3270, opened 2020-02-06
Port "Prevent reporting HAA0101 when no value for params is provided"

genlu:
HAA0101 was incorrectly implemented because the statement This results in an array allocation even if no parameter is passed in for the params parameter is not always true. If no parameter is passed then the compiler will insert Array.Empty<T>() in the method invocation. This PR fixes this issue.

Ported from https://github.com/microsoft/RoslynClrHeapAllocationAnalyzer/pull/78
This is originally committed by @cezarypiatek

FYI @mjsabby

### DIFF
--- a/src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
+++ b/src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
@@ -25,7 +25,7 @@ public class MyClass
     public void Testing()
     {
 
-        Params();
+        Params(); //no allocation, because compiler will implicitly substitute Array<int>.Empty
         Params(1, 2);
         Params(new [] { 1, 2}); // explicit, so no warning
         ParamsWithObjects(new [] { 1, 2}); // explicit, but converted to objects, so stil la warning?!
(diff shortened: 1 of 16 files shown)


src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -25,7 +25,7 @@ public class MyClass
     public void Testing()
     {
 
-        Params();
+        Params(); //no allocation, because compiler will implicitly substitute Array<int>.Empty
```
sharwell:
```suggestion
        Params(); //no allocation, because compiler will implicitly substitute Array.Empty<int>()
```

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -25,7 +25,7 @@ public class MyClass
     public void Testing()
```
sharwell: 💡 A new test needs to be added that uses the `net45` reference assemblies. These do not include `Array.Empty<T>()`, and the compiler may still insert an allocation.

genlu: Could you file an issue for that please?

sharwell: I filed #3272. Keep in mind that the new test library includes reference assemblies so it's pretty easy to write a test targeting a specific version.

PR #3271, opened 2020-02-06
Port "Verify symbol type when reporting HAA0602"

genlu:
Ported from https://github.com/microsoft/RoslynClrHeapAllocationAnalyzer/pull/79. 
Originally implemented by @cezarypiatek

FYI @mjsabby

### DIFF
--- a/src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
+++ b/src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
@@ -334,7 +334,7 @@ namespace Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers
 
         private static bool IsStructInstanceMethod(SyntaxNode? node, SemanticModel semanticModel, CancellationToken cancellationToken)
         {
-            if (node.IsKind(SyntaxKind.AnonymousMethodExpression) || node.IsKind(SyntaxKind.ParenthesizedLambdaExpression))
+            if (node.IsKind(SyntaxKind.AnonymousMethodExpression) || node.IsKind(SyntaxKind.ParenthesizedLambdaExpression) || node.IsKind(SyntaxKind.SimpleLambdaExpression))
             {
                 return false;
             }
--- a/src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
+++ b/src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
@@ -325,12 +325,22 @@ namespace Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers
                     }
                 }
 
-                var symbolInfo = semanticModel.GetSymbolInfo(node, cancellationToken).Symbol;
-                if (symbolInfo?.ContainingType?.IsValueType == true && !insideObjectCreation)
+                if (IsStructInstanceMethod(node!, semanticModel, cancellationToken))
                 {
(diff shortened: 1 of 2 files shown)


src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
```diff
@@ -325,12 +325,22 @@ private static void CheckDelegateCreation(SyntaxNode node, TypeInfo typeInfo, Se
...
+                if (IsStructInstanceMethod(node!, semanticModel, cancellationToken))
                 {
                     reportDiagnostic(Diagnostic.Create(DelegateOnStructInstanceRule, location, EmptyMessageArgs));
                 }
             }
         }
+
+        private static bool IsStructInstanceMethod(SyntaxNode? node, SemanticModel semanticModel, CancellationToken cancellationToken)
+        {
+            if (node.IsKind(SyntaxKind.AnonymousMethodExpression) || node.IsKind(SyntaxKind.ParenthesizedLambdaExpression))
```
sharwell: ❓ Is this still correct if the lambda captures `this`?

src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
```diff
@@ -325,12 +325,22 @@ private static void CheckDelegateCreation(SyntaxNode node, TypeInfo typeInfo, Se
...
+            {
+                return false;
+            }
+
+            var symbolInfo = semanticModel.GetSymbolInfo(node, cancellationToken).Symbol;
+            return symbolInfo != null && symbolInfo.Kind == SymbolKind.Method && symbolInfo.IsStatic == false && symbolInfo.ContainingType?.IsValueType == true;
```
sharwell:
```suggestion
            return symbolInfo?.Kind == SymbolKind.Method && symbolInfo.IsStatic == false && symbolInfo.ContainingType?.IsValueType == true;
```

src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
genlu: That would be a boxing by itself, right? I think this check is only for wrapping an instance method on struct

sharwell: 💭 Is this missing `SyntaxKind.SimpleLambdaExpression`?

src/PerformanceSensitiveAnalyzers/UnitTests/TypeConversionAllocationAnalyzerTests.cs
```diff
@@ -686,5 +692,120 @@ public void SomeMethod()
...
+public struct MyStruct
+{
+    [PerformanceSensitive(""uri"")]
+    public void Testing()
+    {
+        var ints = new[] { 5, 4, 3, 2, 1 };
+        Array.Sort(ints, delegate(int x, int y) { return x - y; });
+        Array.Sort(ints, (x, y) => x - y);
+        DoSomething(() => throw new Exception());
+        DoSomething(delegate() { throw new Exception(); });
```
sharwell: 💡 This needs a test with the simple lambda syntax as well (`x => throw new Exception()`)

src/PerformanceSensitiveAnalyzers/CSharp/TypeConversionAllocationAnalyzer.cs
sharwell: It turns out it's not possible to box from within a lambda in a struct, so we're good there.

PR #3279, opened 2020-02-07
Remove minimal implementations

Evangelink:
Fix #3276 

I have the following 2 errors that I don't manage to get over:

1. `error CS0433: The type 'HttpClient' exists in both 'System.Net.Http, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and 'System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'`

2. `error CS0234: The type or namespace name 'Security' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?)`

Issue #3276, opened 2020-02-06
Remove the MinimalImplementations.IAsyncDisposable class and use Microsoft.Bcl.AsyncInterfaces instead


Evangelink:
Prior to the TFM change, we should remove the MinimalImplementations.IAsyncDisposable class and add a NuGet reference to Microsoft.Bcl.AsyncInterfaces instead.

_Originally posted by @sharwell in https://github.com/dotnet/roslyn-analyzers/pull/3273_

Evangelink:
Shall we get rid of all the minimal implementation?

sharwell:
> Shall we get rid of all the minimal implementation?

Ideally, yes. Referencing the true packages produces more accurate tests.

PR #3284, opened 2020-02-10
Remove minimal implementations for security analyzers

Evangelink:
Contributes to #3276, see #3279


src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -29,6 +30,18 @@ public static class AdditionalMetadataReferences
         public static ReferenceAssemblies DefaultWithNewtonsoftJson { get; } = Default
             .AddPackages(ImmutableArray.Create(new PackageIdentity("Newtonsoft.Json", "10.0.1")));
 
+        public static ReferenceAssemblies DefaultWithWinForms { get; } = Default
+            .AddAssemblies(ImmutableArray.Create("System.Windows.Forms"));
+
+        public static ReferenceAssemblies DefaultWithWinHttpHandler { get; } = ReferenceAssemblies.NetStandard.NetStandard20
```
Evangelink: NetFramework assemblies were causing an issue about `HttpClient` being imported twice.

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -29,6 +30,18 @@ public static class AdditionalMetadataReferences
...
+            .AddPackages(ImmutableArray.Create(new PackageIdentity("System.Net.Http.WinHttpHandler", "4.7.0")));
+
+        public static ReferenceAssemblies DefaultWithAspNetCoreMvc { get; } = Default
+            .AddPackages(ImmutableArray.Create(
+                new PackageIdentity("Microsoft.AspNetCore", "1.1.7"),
+                new PackageIdentity("Microsoft.AspNetCore.Mvc", "1.1.8"),
+                new PackageIdentity("Microsoft.AspNetCore.Http", "1.1.2")));
```
Evangelink: More recent versions of the framework are causing some tests to fail.

Evangelink:
Oh yeah I forgot to mention that I cannot find the package containing the `CurlHandler`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheckTests.cs
```diff
@@ -22,10 +21,8 @@ private async Task VerifyCSharpWithDependenciesAsync(string source, params Diagn
         {
             var csharpTest = new VerifyCS.Test
             {
-                TestState =
-                {
-                    Sources = { source, SystemNetHttpApis.CSharp },
```
sharwell: 💭 System.Net.Http is included by default in netstandard2.0. Are you seeing errors if we just let it use default?

mavasani:
Tagging @dotpaul for review.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheckTests.cs
Evangelink: Yes there were errors about `CurlHandler`, `WinHttpHandler` and a third one that I don't remember.

dotpaul:
> Oh yeah I forgot to mention that I cannot find the package containing the `CurlHandler`.

Hrm, looks like CurlHandler was internal (https://github.com/dotnet/corefx/blob/d3911035f2ba3eb5c44310342cc1d654e42aa316/src/System.Net.Http/src/System/Net/Http/CurlHandler/CurlHandler.SslProvider.Linux.cs#L21) and I can't find it in the new dotnet/runtime repo. I'd be fine with removing the CurlHandler parts from the rule itself. Or I can do it you'd prefer not to.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheckTests.cs
Evangelink: I confirm that there are errors when I don't use the custom ref assemblies. I have even tried to force `ReferenceAssemblies.NetStandard.Netstarnd2.0`

dotpaul:
@dotnet-bot retest please

PR #3298, opened 2020-02-13
Update CA1720: Do not report on identifier 'obj'

### LINKED ISSUE #1823: Identifier 'obj' contains type name
Analyzer package: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)
Package Version: 2.6.2-beta2
Diagnostic ID: CA1720

```c#
public static IDisposable ExpectPropertyChanged(this INotifyPropertyChanged obj, int callCount)
{
    // ...
}
```

> CA1720 Identifier 'obj' contains type name

'obj' does not contain a type name in a sense that I understand. Also, I picked `obj` because is a recognized parameter name used by the BCL, e.g. object.Equals. Is this to be avoided? Would something like `observableObject` in fact be better even though it seems less consistent with the BCL?

Evangelink:
Fix #1823


Evangelink:
Tagging @mavasani

PR #3299, opened 2020-02-13
Update CA2000 to exclude System.IO.StringReader

### LINKED ISSUE #3212: CA2000 - Exclude StringReader by default

Evangelink:
Fix #3212


src/Utilities/Compiler/Options/CategorizedAnalyzerConfigOptions.cs
```diff
@@ -113,7 +113,7 @@ public T GetOptionValue<T>(string optionName, DiagnosticDescriptor rule, TryPars
                 return defaultValue;
             }
 
-            return (T)_computedOptionValuesMap.GetOrAdd($"{rule.Id}.{optionName}", _ => ComputeOptionValue(optionName, rule, tryParseValue, defaultValue))!;
+            return (T)_computedOptionValuesMap.GetOrAdd($"{rule.Id}.{optionName}.{defaultValue?.GetHashCode()}", _ => ComputeOptionValue(optionName, rule, tryParseValue, defaultValue))!;
```
Evangelink: I had to update the cache to allow callers with a different "forced value" to have a different result and not the same cache entry. We might want to have a constraint on `T` if we go-on with the `GetHashCode` method. Alternatively, we could change the method signature to have the default value being a `string` and not a `T` and we would call `TryParse` on this `string` entry.

Anyways let me know which direction you think is best.

Evangelink:
Tagging @sharwell @mavasani

Evangelink:
ping @sharwell

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -247,28 +248,51 @@ static SymbolNamesWithValueOption<INamedTypeSymbol>.NameParts GetParts(string na
         private static SymbolNamesWithValueOption<TValue> GetSymbolNamesWithValueOption<TValue>(
             this AnalyzerOptions options,
             string optionName,
-            string? namePrefixOpt,
```
mavasani: I think it is preferable to keep this as a required parameter so every new callsite has to make an explicit decision on supported name prefix.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.cs
```diff
@@ -62,7 +62,8 @@ private DisposeAnalysis(DisposeAnalysisDomain analysisDomain, DisposeDataFlowOpe
                 disposeOwnershipTransferLikelyTypes, disposeOwnershipTransferAtConstructor,
                 disposeOwnershipTransferAtMethodCall, trackInstanceFields, exceptionPathsAnalysis,
                 performCopyAnalysis: analyzerOptions.GetCopyAnalysisOption(rule, defaultValue: performCopyAnalysisIfNotUserConfigured, cancellationToken),
-                excludedSymbols: analyzerOptions.GetExcludedSymbolNamesWithValueOption(rule, wellKnownTypeProvider.Compilation, cancellationToken),
+                // StringReader doesn't need to be disposed: https://docs.microsoft.com/en-us/dotnet/api/system.io.stringreader?view=netframework-4.8
+                excludedSymbols: analyzerOptions.GetExcludedSymbolNamesWithValueOption(rule, wellKnownTypeProvider.Compilation, cancellationToken, defaultForcedValue: $"T:{WellKnownTypeNames.SystemIOStringReader}"),
```
mavasani: Wouldn't it be simpler to just update this existing special case for `Task` to also include this new type: https://github.com/dotnet/roslyn-analyzers/blob/81512992e784b6108a88d93df0ffe0f72ca0af5c/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs#L93-L98

src/Utilities/Compiler/Options/CategorizedAnalyzerConfigOptions.cs
mavasani: Yeah, this looks fragile. I'd rather avoid all changes in this PR to any of the options infrastructure and just make an isolated fix to Dispose analysis, as suggested in https://github.com/dotnet/roslyn-analyzers/pull/3299/files#r384039514

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
```diff
@@ -11956,5 +11956,92 @@ void M2()
...
+
+Public Class C
+    Public Sub New()
+        Dim x = new StringReader(""abc"")
+    End Sub
+End Class");
+        }
+
+        [Fact, WorkItem(3212, "https://github.com/dotnet/roslyn-analyzers/issues/3212")]
+        public async Task StringReader_CustomSymbolExclusion_NoDiagnostic()
```
Evangelink: I have kept this test just to ensure that there is no impact when changing the excluded_symbols. If you think that's useless, I will remove it.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -3638,6 +3639,11 @@ public sealed override TAbstractAnalysisValue VisitLocalFunction(ILocalFunctionO
         /// </summary>
         protected INamedTypeSymbol? GenericIEquatableNamedType { get; }
 
+        /// <summary>
+        /// <see cref="INamedTypeSymbol"/> for <see cref="System.IO.StringReader"/>
+        /// </summary>
+        protected INamedTypeSymbol? StringReaderType { get; }
```
mavasani: You need to add this to the public API surface for AnalyzerUtilities package.

mavasani: `F:\workspace\_work\1\s\src\Utilities\FlowAnalysis\FlowAnalysis\Framework\DataFlow\DataFlowOperationVisitor.cs(3645,56): error RS0016: Symbol 'StringReaderType.get' is not part of the declared API.`

Evangelink: Yes sorry I missed the warning in VS :disappointed:

PR #3300, opened 2020-02-13
Update CA1716 to stop reporting on parameters

### LINKED ISSUE #1858: CA1716 warns for vb.net keyword in c#
#### Analyzer package

Example: [Microsoft.CodeQuality.Analyzers]

#### Package Version

v2.6.2

#### Diagnostic ID

Example: [CA1716](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1716-identifiers-should-not-match-keywords)

#### Repro steps

1. Create a abstract class in **C#** with method that attribute named step  (class library with .net 4.7.1 in VS15.8.7)

```C#
public abstract class myBase
    {
        public virtual void MyStep(CustomActionStep step)
        {
           // Do somthing
        }
    }
```

#### Expected behavior
Because it is c# and step is a vb.net keyword, it should not be a Problem

#### Actual behavior
Warns me

Evangelink:
Fix #1858


Evangelink:
@mavasani I added a comment on the issue to try to understand which direction you/we wanted to take and thought that with the timezone it might be worth for me to actually add the code for the "implementation" that was making the more sense to me (in case I got it right and that you could merge). You can mark the PR as blocked and I will wait for a decision/consensus on the issue before updating the PR. Sorry if it had caused some useless work on your side.

Evangelink:
Overruled by #3428

PR #3303, opened 2020-02-13
HAA0101 add unit test for empty 'params' in net45

### LINKED ISSUE #3272: HAA0101 should be reported for empty 'params' when targeting net45
💡 A new test needs to be added that uses the `net45` reference assemblies. These do not include `Array.Empty<T>()`, and the compiler may still insert an allocation.

_Originally posted by @sharwell in https://github.com/dotnet/roslyn-analyzers/pull/3270_

Evangelink:
Fix #3272

@sharwell is it what you were expecting?

### DIFF
--- a/src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
+++ b/src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
@@ -79,6 +79,11 @@ namespace Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers
             var arguments = invocationExpression.ArgumentList.Arguments;
             if (arguments.Count == methodInfo.Parameters.Length - 1)
             {
+                // Up to net45 the System.Array.Empty<T> singleton didn't existed so an empty params array was still causing some memory allocation.
+                if (semanticModel.Compilation.GetSpecialType(SpecialType.System_Array).GetMembers("Empty").Length == 0)
+                {
+                    reportDiagnostic(Diagnostic.Create(ParamsParameterRule, invocationExpression.GetLocation(), EmptyMessageArgs));
+                }
                 return;
             }
 
--- a/src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
+++ b/src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
@@ -2,6 +2,7 @@
 
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers;
+using Microsoft.CodeAnalysis.Testing;
 using Test.Utilities;
 using Xunit;
 using VerifyCS = Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.UnitTests.CSharpPerformanceCodeFixVerifier<
@@ -146,6 +147,31 @@ public class MyClass
             await VerifyCS.VerifyAnalyzerAsync(source,
                 // Test0.cs(12,9): warning HAA0102: Non-overridden virtual method call on a value type adds a boxing or constrained instruction
                 VerifyCS.Diagnostic(CallSiteImplicitAllocationAnalyzer.ValueTypeNonOverridenCallRule).WithLocation(12, 9));
+        }
+
+        [Fact, WorkItem(3272, "https://github.com/dotnet/roslyn-analyzers/issues/3272")]
+        public async Task C()
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.NetFramework.Net45.Default,
+                TestCode = @"
+using System;
+using Roslyn.Utilities;
+
+public class MyClass
+{
+    [PerformanceSensitive(""uri"")]
+    public void Testing()
+    {
+        Params(); //no allocation
+    }
+
+    public void Params(params int[] args)
+    {
+    }
+}",
+            }.RunAsync();
         }
     }
 }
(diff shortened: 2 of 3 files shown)


genlu:
Thanks! I think the actual issue is we are not checking for the existence of `Array.Empty<T>` in the analyzer, so in case of .net framework versions prior to 4.6, we would have false negative.

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -147,5 +148,30 @@ public void SomeMethod()
...
+    public void Testing()
+    {
+        Params(); //no allocation
+    }
+
+    public void Params(params int[] args)
+    {
+    }
+}",
+            }.RunAsync();
```
genlu: Test code look good to me, just need to call though `VerifyCS.VerifyAnalyzerAsync` which will include attribute source.

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -147,5 +148,30 @@ public void SomeMethod()
                 // Test0.cs(12,9): warning HAA0102: Non-overridden virtual method call on a value type adds a boxing or constrained instruction
                 VerifyCS.Diagnostic(CallSiteImplicitAllocationAnalyzer.ValueTypeNonOverridenCallRule).WithLocation(12, 9));
         }
+
+        [Fact, WorkItem(3272, "https://github.com/dotnet/roslyn-analyzers/issues/3272")]
+        public async Task C()
```
genlu: Test name

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -147,5 +148,30 @@ public void SomeMethod()
...
+                TestCode = @"
+using System;
+using Roslyn.Utilities;
+
+public class MyClass
+{
+    [PerformanceSensitive(""uri"")]
+    public void Testing()
+    {
+        Params(); //no allocation
```
sharwell: ❗️ This is not correct. When targeting **net45** there _is_ an allocation here.

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
sharwell: The full form is required since we are changing the reference assemblies.

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
Evangelink: My bad, I went for a quick test and forgot to review... I will update!

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -52,6 +53,44 @@ public void ParamsWithObjects(params object[] args)
...
+
+    public void Params(params int[] args)
+    {
+    }
+}",
+                        ("PerformanceSensitiveAttribute.cs", VerifyCS.PerformanceSensitiveAttributeSource)
+                    },
+                    ExpectedDiagnostics =
+                    {
+                        VerifyCS.Diagnostic(CallSiteImplicitAllocationAnalyzer.ParamsParameterRule).WithLocation(10, 9),
```
Evangelink: Alright, so now, the test is failing because no issue is raised. In the `CheckParam` method when debugging the net45 case and the other case I don't seem to see any difference:

https://github.com/dotnet/roslyn-analyzers/blob/master/src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs#L79-L83

Evangelink: The only thing I can see (with my knowledge) is to access to the referenced assemblies through the compilation and to check the .net version using `mscorlib`. Then it that's earlier than net45 we do raise. Is it what you are thinking about?

sharwell: The allocation is only avoided if `System.Array.Empty<T>()` is defined.

Evangelink:
Thank you @sharwell! You did basically all the job of the PR but hey happy to have "helped" :D

src/PerformanceSensitiveAnalyzers/CSharp/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -79,6 +79,11 @@ private static void CheckParam(InvocationExpressionSyntax invocationExpression,
             var arguments = invocationExpression.ArgumentList.Arguments;
             if (arguments.Count == methodInfo.Parameters.Length - 1)
             {
+                // Up to net45 the System.Array.Empty<T> singleton didn't existed so an empty params array was still causing some memory allocation.
+                if (semanticModel.Compilation.GetSpecialType(SpecialType.System_Array).GetMembers("Empty").Length == 0)
```
Evangelink: I have tried to extract this to the `AnalyzeNode` method but this was actually causing a few more un-needed access. The best thing would be to have this check done once at the base-class level but I wasn't sure this was ok for you to update everything.

sharwell: I'm OK with this for now since it's already gated on the attribute.

Evangelink:
ping @genlu

Issue #2703, opened 2019-07-26
CA2000 on using declaration with object initializer


sectax:
#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

v2.9.3 (Latest)

#### Diagnostic ID

[CA2000](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2000-dispose-objects-before-losing-scope)

#### Repro steps
1. Create a library using netstandard2.0 and `<LangVersion>preview</LangVersion>`
2. Insert following snippet code

```cs
public static void RaisesWarning()
{
    using var aesEncryption = new RijndaelManaged
    {
        KeySize = 128
    };
}

public static void DoesNotRaiseWarning()
{
    using (var aesEncryption = new RijndaelManaged
    {
        KeySize = 128
    }) {}
}
```

#### Expected behavior
Either both methods generate CA2000, or neither.

#### Actual behavior
Only the method with a using declaration generates a CA2000 warning.

mavasani:
This is due to missing IOperation support for using declarations in Roslyn: https://github.com/dotnet/roslyn/issues/32100

mavasani:
I will add a workaround to the analyzer for this case.

PR #3309, opened 2020-02-14
Fix for CA2000 not detected if nameof() is in the scope

### LINKED ISSUE #3297: CA2000 not detected when using nameof() inside the method
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

Example: [CA2000](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2000)

#### Repro steps

1. Create an empty solution from the .NET Framework 4.8 Console Application.
2. Add "Microsoft.CodeAnalysis.FxCopAnalyzers" version "2.9.8".
3. Add the following code:

```csharp
    public class Baz
    {
        // doesn't trigger CA2000
        public string Foo()
        {
            var streamReader = new StreamReader(Stream.Null);
            Console.WriteLine(nameof(Foo));
            return streamReader.ReadToEnd();
        }
    
        // triggers CA2000
        public string Bar()
        {
            var streamReader = new StreamReader(Stream.Null);
            Console.WriteLine("Bar");
            return streamReader.ReadToEnd();
        }
    }
```

#### Expected behavior

Both the `Bar` and `Foo` methods get flagged for CA2000 violation.

#### Actual behavior

Only the `Bar` method gets flagged for CA2000 violation.

tompazourek:
fixes #3297

Basically any scope that contains a `nameof` operator will be ignored by the CA2000:DisposeObjectsBeforeLosingScope rule. And, in my opinion, `nameof` is used quite often (parameter checks, exception messages, logging, etc.), so it's quite a serious bug.

This bug was introduced by #2704 (workaround for issue #2703), see also: https://github.com/dotnet/roslyn/issues/32100. The contents of the `nameof` expression are an `OperationKind.None`, which is why the hack breaks it. I added additional check for `INameOfOperation` specifically.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs
@@ -13,6 +13,7 @@ using Microsoft.CodeAnalysis.FlowAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DisposeAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PointsToAnalysis;
+using Microsoft.CodeAnalysis.Operations;
 
 namespace Microsoft.NetCore.Analyzers.Runtime
 {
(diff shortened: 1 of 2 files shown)


mavasani:
Tagging @333fred - the workaround looks fine to me, but I think we want to fix the IOperation tree for `nameof(TypeName)` to not have an `Operation.None` as a child node of `INameOfOperation`. I'll file a bug in Roslyn repo with a unit test.

mavasani:
@tompazourek Thanks for the contribution!

mavasani:
Filed https://github.com/dotnet/roslyn/issues/41963 for the underlying IOperation issue with nameof expressions.

Issue #3305, opened 2020-02-13
CA2000 needs to be emitted when type has Dispose method


lostmsu:
#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

[CA2000](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2000)

#### Repro steps

```csharp
ref struct Disposable { public void Dispose(){} }
void Do(){ var x = new Disposable(); }
```

#### Expected behavior
CA2000 for `x`

#### Actual behavior
No CA2000 for `x`

#### Notes
C# 8.0 added ability for `using` block and statement to call `Dispose` method by name when the type does not implement `IDisposable`.

sharwell:
This applies _only_ to `ref struct`.

PR #3311, opened 2020-02-15
CA2000 for ref structs, that are not disposed properly

lostmsu:
fixes https://github.com/dotnet/roslyn-analyzers/issues/3305

In this fix:
- three new tests: `LocalWithRefStructDisposableAssignment_NotDisposed_Diagnostic`, `LocalWithRefStructDisposableAssignment_Internal_NotDisposed_Diagnostic` and `LocalWithRefStructDisposableAssignment_Disposed_NoDiagnostic` (C# only, `protected` is not covered as structs can not declare new `protected` methods)
- `GetDisposeMethodKind` returns `DisposeMethodKind.Dispose` for parameterless `Dispose` instance method on ref-only structs
- `IsDisposable` returns `true` for ref-only structs with non-private parameterless `Dispose` instance method
- `PointsToAnalysis` and `DisposeDataFlowOperationVisitor` track ref-only value types in addition to reference types

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
@@ -2160,6 +2160,37 @@ class Test
         }
 
         [Fact, WorkItem(3305, "https://github.com/dotnet/roslyn-analyzers/issues/3305")]
+        public async Task LocalWithRefStructDisposableAssignment_Internal_NotDisposed_Diagnostic()
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
+                TestCode = @"
+using System;
+
+ref struct RefStructDisposable
+{
+    internal void Dispose()
+    {
+    }
+}
+
+class Test
+{
+    public static void M1()
+    {
+        var e = new RefStructDisposable();
+    }
+}
+",
+                ExpectedDiagnostics =
+                {
+                    GetCSharpResultAt(15, 17, "new RefStructDisposable()"),
+                }
+            }.RunAsync();
+        }
+
+        [Fact, WorkItem(3305, "https://github.com/dotnet/roslyn-analyzers/issues/3305")]
         public async Task LocalWithRefStructDisposableAssignment_Disposed_NoDiagnostic()
         {
             await new VerifyCS.Test
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
@@ -2128,6 +2128,64 @@ End Class",
             }.RunAsync();
         }
 
+        [Fact, WorkItem(3305, "https://github.com/dotnet/roslyn-analyzers/issues/3305")]
+        public async Task LocalWithRefStructDisposableAssignment_NotDisposed_Diagnostic()
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
+                TestCode = @"
+using System;
+
+ref struct RefStructDisposable
+{
+    public void Dispose()
+    {
+    }
+}
+
+class Test
+{
+    public static void M1()
+    {
+        var e = new RefStructDisposable();
+    }
+}
+",
+                ExpectedDiagnostics =
+                {
+                    GetCSharpResultAt(15, 17, "new RefStructDisposable()"),
+                }
+            }.RunAsync();
+        }
+
+        [Fact, WorkItem(3305, "https://github.com/dotnet/roslyn-analyzers/issues/3305")]
+        public async Task LocalWithRefStructDisposableAssignment_Disposed_NoDiagnostic()
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
+                TestCode = @"
+using System;
+
+ref struct RefStructDisposable
+{
+    public void Dispose()
+    {
+    }
+}
+
+class Test
+{
+    public static void M1()
+    {
+        var e = new RefStructDisposable();
+        e.Dispose();
+    }
+}"
+            }.RunAsync();
+        }
+
         [Fact]
         public async Task ArrayElementWithDisposableAssignment_NoDiagnostic()
         {
--- a/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
@@ -283,7 +283,7 @@ namespace Analyzer.Utilities.Extensions
                 if (IsDisposeImplementation(method, iDisposable) ||
                     (Equals(method.ContainingType, iDisposable) &&
                      method.HasDisposeMethodSignature())
-#if NET_ANALYZERS
+#if CODEANALYSIS_V3_OR_BETTER
                     || (method.ContainingType.IsRefLikeType &&
                      method.HasDisposeSignatureByConvention())
 #endif
--- a/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
@@ -199,7 +199,7 @@ namespace Analyzer.Utilities.Extensions
         public static bool HasDisposeSignatureByConvention(this IMethodSymbol method)
         {
             return method.HasDisposeMethodSignature()
-                && !method.GetSymbolModifiers().HasFlag(SymbolModifiers.Static)
+                && !method.IsStatic
                 && !method.IsPrivate();
         }
 
--- a/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
@@ -194,6 +194,16 @@ namespace Analyzer.Utilities.Extensions
         }
 
         /// <summary>
+        /// Checks if the given method matches Dispose method convention and can be recognized by "using".
+        /// </summary>
+        public static bool HasDisposeSignatureByConvention(this IMethodSymbol method)
+        {
+            return method.HasDisposeMethodSignature()
+                && !method.GetSymbolModifiers().HasFlag(SymbolModifiers.Static)
+                && !method.IsPrivate();
+        }
+
+        /// <summary>
         /// Checks if the given method has the signature "void Dispose(bool)".
         /// </summary>
         public static bool HasDisposeBoolMethodSignature(this IMethodSymbol method)
@@ -272,7 +282,12 @@ namespace Analyzer.Utilities.Extensions
             {
                 if (IsDisposeImplementation(method, iDisposable) ||
                     (Equals(method.ContainingType, iDisposable) &&
-                     method.HasDisposeMethodSignature()))
+                     method.HasDisposeMethodSignature())
+#if NET_ANALYZERS
+                    || (method.ContainingType.IsRefLikeType &&
+                     method.HasDisposeSignatureByConvention())
+#endif
+                )
                 {
                     return DisposeMethodKind.Dispose;
                 }
--- a/src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
@@ -126,7 +126,8 @@ namespace Analyzer.Utilities.Extensions
         }
 
         /// <summary>
-        /// Indicates if the given <paramref name="type"/> is a reference type that implements <paramref name="iDisposable"/> or System.IAsyncDisposable or is <see cref="IDisposable"/> or System.IAsyncDisposable type itself.
+        /// Indicates if the given <paramref name="type"/> is disposable,
+        /// and thus can be used in a <code>using</code> or <code>await using</code> statement.
         /// </summary>
         public static bool IsDisposable(this ITypeSymbol type,
             INamedTypeSymbol? iDisposable,
@@ -280,11 +281,7 @@ namespace Analyzer.Utilities.Extensions
         }
 
         public static bool HasValueCopySemantics(this ITypeSymbol typeSymbol)
-            => typeSymbol.IsValueType
-#if NET_ANALYZERS
-                && !typeSymbol.IsRefLikeType
-#endif
-            || typeSymbol.SpecialType == SpecialType.System_String;
+            => typeSymbol.IsValueType || typeSymbol.SpecialType == SpecialType.System_String;
 
         public static bool IsNonNullableValueType([NotNullWhen(returnValue: true)] this ITypeSymbol? typeSymbol)
             => typeSymbol != null && typeSymbol.IsValueType && typeSymbol.OriginalDefinition.SpecialType != SpecialType.System_Nullable_T;
--- a/src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
@@ -139,7 +139,7 @@ namespace Analyzer.Utilities.Extensions
                     || IsInterfaceOrImplementsInterface(type, iAsyncDisposable);
             }
 
-#if NET_ANALYZERS
+#if CODEANALYSIS_V3_OR_BETTER
             if (type.IsRefLikeType)
             {
                 return type.GetMembers("Dispose").OfType<IMethodSymbol>()
--- a/src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
@@ -132,9 +132,21 @@ namespace Analyzer.Utilities.Extensions
             INamedTypeSymbol? iDisposable,
             INamedTypeSymbol? iAsyncDisposable)
         {
-            return type.IsReferenceType &&
-                (IsInterfaceOrImplementsInterface(type, iDisposable) ||
-                 IsInterfaceOrImplementsInterface(type, iAsyncDisposable));
+            if (type.IsReferenceType)
+            {
+                return IsInterfaceOrImplementsInterface(type, iDisposable)
+                    || IsInterfaceOrImplementsInterface(type, iAsyncDisposable);
+            }
+
+#if NET_ANALYZERS
+            if (type.IsRefLikeType)
+            {
+                return type.GetMembers("Dispose").OfType<IMethodSymbol>()
+                    .Any(method => method.HasDisposeSignatureByConvention());
+            }
+#endif
+
+            return false;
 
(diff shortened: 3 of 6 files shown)


src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -272,7 +282,12 @@ public static DisposeMethodKind GetDisposeMethodKind(this IMethodSymbol method,
             {
                 if (IsDisposeImplementation(method, iDisposable) ||
                     (Equals(method.ContainingType, iDisposable) &&
-                     method.HasDisposeMethodSignature()))
+                     method.HasDisposeMethodSignature())
+#if NET_ANALYZERS
+                    || (method.ContainingType.IsRefLikeType &&
```
Evangelink: I would move the ref struct check inside of the `HasDisposeSignatureByConvention` helper.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -193,6 +193,16 @@ private static bool HasDisposeMethodSignature(this IMethodSymbol method)
...
         }
 
+        /// <summary>
+        /// Checks if the given method matches Dispose method convention and can be recognized by "using".
+        /// </summary>
+        public static bool HasDisposeSignatureByConvention(this IMethodSymbol method)
+        {
+            return method.HasDisposeMethodSignature()
+                && !method.GetSymbolModifiers().HasFlag(SymbolModifiers.Static)
+                && !method.IsPrivate();
```
Evangelink: Is the `Dispose` convention working for something else than `public`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
```diff
@@ -2128,6 +2128,64 @@ End Function
...
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
+                TestCode = @"
+using System;
+
+ref struct RefStructDisposable
+{
+    public void Dispose()
```
Evangelink: Is it possible to have `DisposeAsync` by convention?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: I'd keep it as is:
1. It would not be immediately clear from the current name, that it also performs type check on the parent type.
2. In both callsites by the time this method is called, it is already known, that type is ref-like, and in one case an extra logic has been computed (scanned for method by name).

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: Apparently, the answer is yes. This compiles:
```csharp
ref struct InternallyDisposable { internal void Dispose() { } }
...
using var internallyDisposable = new InternallyDisposable();
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
lostmsu: Yes and no.

You can declare the method, and there would be no `using`-related error in `await using var disposable = new RefLikeAsyncDisposable()`.

But there will be another error: at this moment ref-like types are completely prohibited in async methods. So `var disposable` will lead to `CS4012`: Parameters or locals of type 'RefLikeAsyncDisposable' cannot be declared in async methods or lambda expressions.

Evangelink: Might be worth adding a test for this case with an expected error which could be fixed in the future, WDYT?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Evangelink: Might be worth adding a test for this case too to avoid regression

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Evangelink: Makes sense.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
lostmsu: I don't understand what kind of test do you suggest exactly.

Due to the `CS4012` I don't think this PR needs to be concerned of `async` for now.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: Added one for `internal`. `protected` is undefined/impossible at the moment too, as structs can't declare new `protected` members.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
Evangelink: I was thinking about something like (code was written directly so might have some errors):

```csharp
await new VerifyCS.Test
{
	ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
	TestCode = @"
using System;
ref struct RefStructDisposableAsync
{
    public ValueTask DisposeAsync()
    {
    }
}
class Test
{
    public static async Task M1()
    {
        await using var e = new RefStructDisposableAsync();
    }
}
",
		ExpectedDiagnostics =
		{
			VerifyCS.CompilerError("CS4012").WithLocation(),
		}
	}.RunAsync();
}
```

lostmsu: I think it should be fine without tests for `Async` variants for now. If it is ever allowed in the future code style and semantics might change a bit.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -272,7 +282,12 @@ public static DisposeMethodKind GetDisposeMethodKind(this IMethodSymbol method,
             {
                 if (IsDisposeImplementation(method, iDisposable) ||
                     (Equals(method.ContainingType, iDisposable) &&
-                     method.HasDisposeMethodSignature()))
+                     method.HasDisposeMethodSignature())
+#if NET_ANALYZERS
```
mavasani: Why `#if NET_ANALYZERS`? This will cause this fix not to appear in FxCop analyzers package. I think you probably `#if HAS_IOPERATION`, which should prevent it.

```suggestion
#if HAS_IOPERATION
```

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -132,9 +132,21 @@ public static bool DerivesFrom([NotNullWhen(returnValue: true)] this ITypeSymbol
...
-            return type.IsReferenceType &&
-                (IsInterfaceOrImplementsInterface(type, iDisposable) ||
-                 IsInterfaceOrImplementsInterface(type, iAsyncDisposable));
+            if (type.IsReferenceType)
+            {
+                return IsInterfaceOrImplementsInterface(type, iDisposable)
+                    || IsInterfaceOrImplementsInterface(type, iAsyncDisposable);
+            }
+
+#if NET_ANALYZERS
```
mavasani:
```suggestion
#if HAS_IOPERATION
```

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -132,9 +132,21 @@ public static bool DerivesFrom([NotNullWhen(returnValue: true)] this ITypeSymbol
...
+            if (type.IsRefLikeType)
```
mavasani: Please update the doc comment accordingly.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
mavasani: Can we extract out a helper `IsDisposeMethodOfRefLikeType(IMethodSymbol)` and use it here, as well as in `IsDisposable` method below as `return type.GetMembers("Dispose").OfType<IMethodSymbol>().Any(IsDisposeMethodOfRefLikeType);`

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -268,7 +280,11 @@ public static bool IsAttribute(this ITypeSymbol symbol)
         }
 
         public static bool HasValueCopySemantics(this ITypeSymbol typeSymbol)
-            => typeSymbol.IsValueType || typeSymbol.SpecialType == SpecialType.System_String;
+            => typeSymbol.IsValueType
+#if NET_ANALYZERS
```
mavasani:
```suggestion
#if HAS_IOPERATION
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.cs
```diff
@@ -79,7 +79,8 @@ private PointsToAnalysis(PointsToAnalysisDomain analysisDomain, PointsToDataFlow
         }
 
         internal static bool ShouldBeTracked(ITypeSymbol typeSymbol) => typeSymbol.IsReferenceTypeOrNullableValueType() ||
-            typeSymbol is ITypeParameterSymbol typeParameter && !typeParameter.IsValueType;
+            typeSymbol?.HasValueCopySemantics() == false ||
```
mavasani: The change here feels a bit conflicting for `string` types as we do want to track them, but `HasValueCopySemantics` returns false for it. Can we instead just have a new extension method on `ITypeSymbol` that explicitly checks for ref like types, say `IsRefLikeType()`, which delegates to the property when available, and returns false otherwise?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
```diff
@@ -63,7 +63,7 @@ protected override void SetAbstractValue(AbstractLocation location, DisposeAbstr
             {
                 if (!location.IsNull &&
                     location.LocationTypeOpt != null &&
-                    !location.LocationTypeOpt.IsValueType &&
+                    !location.LocationTypeOpt.HasValueCopySemantics() &&
```
mavasani: I think you can avoid having any value type or value copy semantic check here. `IsDisposable` check just below this line should ensure this.

mavasani:
```suggestion
```

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -193,6 +193,16 @@ private static bool HasDisposeMethodSignature(this IMethodSymbol method)
                 method.ReturnsVoid && method.Parameters.IsEmpty;
         }
 
+        /// <summary>
+        /// Checks if the given method matches Dispose method convention and can be recognized by "using".
+        /// </summary>
+        public static bool HasDisposeSignatureByConvention(this IMethodSymbol method)
+        {
+            return method.HasDisposeMethodSignature()
+                && !method.GetSymbolModifiers().HasFlag(SymbolModifiers.Static)
```
mavasani:
```suggestion
                && !method.IsStatic
```

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: I was not sure what conditional to use for compatibility, and neither of the conditionals are documented.

I used this `#if` because, apparently, not all projects that include these files reference a version of `Microsoft.CodeAnalysis.Common`, that has `IsRefLikeType` property on `ITypeSymbol`.

Can you confirm, that `HAS_IOPERATION` is more meaningful for this use case, and why?

P.S. roslyn-analyzers team should probably document the conditionals and the rationale behind them somewhere it could be easily found.

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
lostmsu: I can add the description of this specific case to the doc comment (e.g. something like _or is a ref-like type with a non-private void Dispose() method_), but I feel like if it would be better if the doc comment spoke about applicability of `using` statement rather than listed all the specific cases it is looking for. E.g. I'd replace the whole comment with _Indicates if the given type is disposable (can be used in a `using` or `await using` statement)._ What do you think?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: I can extract the method for clarity, but what would be the advantage of using it in `IsDisposable` over `HasDisposeSignatureByConvention`?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.cs
lostmsu: @mavasani perhaps I did not understand the purpose of `HasValueCopySemantics`. From my understanding, it is incorrect to say, that `System.String` type has value copy semantics. Rather it is an immutable reference type.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
lostmsu: My understand of this check for `IsValueType` was that it has to do with the intent. E.g. if the value is being copied, it should not be tracked regardless of it being disposable or not.

In fact, if this check is completely removed, this would lead a change of behavior in this analyzer for `struct` types, that implement `IDisposable`. From my understanding, because this check is present, they are not being tracked.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.cs
mavasani: The helper method just intends to check if assigning a variable of the given type, say `a = b` causes `a` and `b` to  become reference equals or value equals.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
mavasani: Sounds reasonable to me.

src/Utilities/Compiler/Extensions/ITypeSymbolExtensions.cs
mavasani: Sure, that sounds a reasonable change to me.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
mavasani: I'll let you make the final call here as all of these are private methods. I was just providing a suggestion that might avoid duplicated code with `#if` directives.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
mavasani: Yeah, these directives are really unfortunate, and needed primarily because we have couple of analyzer packages that target a very early version of Microsoft.CodeAnalysis and import this shared project. I can clean this up in a follow-up PR to make it more explicit. For now, switching to `HAS_IOPERATION` will ensure that such analyzer packages on older Microsoft.CodeAnalysis will be excluded.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.cs
lostmsu: @mavasani seems quite unfortunate in terms of naming and the expectations. I'll probably just introduce `IsPassedByReference` and use that then.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: This suggestion is not working. The following projects seem to have this outdated dependency lacking `IsRefLikeType` and define `HAS_IOPERATION`: PerformanceSensitiveAnalyzers and BannedApiAnalyzers

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.cs
lostmsu: Reverted this change in favor of explicitly checking for `IsRefLikeType ` in `PointsToAnalysis` and `DisposeDataFlowOperationVisitor`

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
lostmsu: @mavasani any suggestions regarding `HAS_IOPERATION` here in `PerformanceSensitiveAnalyzers` and `BannedApiAnalyzers`?

mavasani: Those packages should not be needing these APIs. It is fine if this functionality does not exist for those analyzer packages.

lostmsu: @mavasani problem is they define `HAS_IOPERATION`, so I can not use it to skip `IsRefLikeType` check there.

mavasani: Ah, thanks for the clarification. It seems reasonable to define a new repo wide variable, say `NOT_ON_LATEST_MICROSOFT_CODE_ANALYSIS` or some such, and all packages which override their referenced Microsoft.CodeAnalysis package reference version to a lower one then defined in the repo's Versions.props can define this constant in their project file.

lostmsu: @mavasani ok, added `CODEANALYSIS_V3_OR_BETTER` conditional compilation symbol.

lostmsu:
@mavasani @Evangelink do I need to do anything to progress this further?

mavasani:
Sorry for the delay in merging the PR and thanks for your contribution!

PR #3315, opened 2020-02-24
Update CA1506: Ignore compiler generated types

### LINKED ISSUE #2133: LINQ syntax artificially inflates class coupling counts for CA1506

Evangelink:
Fix #2133

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
@@ -457,6 +457,51 @@ CA1506: 2
             await VerifyCSharpAsync(source, additionalText, expected);
         }
 
+        [Fact, WorkItem(2133, "https://github.com/dotnet/roslyn-analyzers/issues/2133")]
+        public void CA1506_Configuration_CSharp_Linq()
+        {
+            var source = @"
+using System.Linq;
+using System.Collections.Generic;
+class C
+{
+    IEnumerable<int> TestCa1506()
+    {
+        var ints = new[] { 1, 2 };
+        return from a in ints
+               from b in ints
+               from c in ints
+               from d in ints
+               from e in ints
+               from f in ints
+               from g in ints 
+               from h in ints
+               from i in ints
+               from j in ints
+               from k in ints
+               from l in ints
+               from m in ints
+               from n in ints
+               from o in ints
+               from p in ints
+               select p;
+    }
+}
+";
+            string additionalText = @"
+# FORMAT:
+# 'RuleId'(Optional 'SymbolKind'): 'Threshold'
+
+CA1506: 2
+";
+            DiagnosticResult[] expected = new[] {
+                // Test0.cs(4,7): warning CA1506: 'C' is coupled with '4' different types from '3' different namespaces. Rewrite or refactor the code to decrease its class coupling below '3'.
+                GetCSharpCA1506ExpectedDiagnostic(4, 7, "C", 4, 3, 3),
+                // Test0.cs(4,10): warning CA1506: 'TestCa1506' is coupled with '4' different types from '3' different namespaces. Rewrite or refactor the code to decrease its class coupling below '3'.
+                GetCSharpCA1506ExpectedDiagnostic(6, 22, "TestCa1506", 4, 3, 3)};
+            VerifyCSharp(source, GetAdditionalFile(additionalText), expected);
+        }
+
         [Fact]
         public async Task CA1506_Configuration_Basic_VerifyDiagnostic()
         {
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
@@ -604,6 +604,128 @@ CA1506(Type): 10
             await VerifyBasicAsync(source, additionalText, expected);
         }
 
+        [Fact, WorkItem(2133, "https://github.com/dotnet/roslyn-analyzers/issues/2133")]
+        public async Task CA1506_CountCorrectlyGenericTypes()
+        {
+            await VerifyCSharpAsync(@"
+using System.Collections.Generic;
+
+public class A {}
+public class B {}
+
+public class C
+{
+    private IEnumerable<A> a;
+    private IEnumerable<B> b;
+}",
+@"
+# FORMAT:
+# 'RuleId'(Optional 'SymbolKind'): 'Threshold'
+
+CA1506: 2
+",
+                GetCSharpCA1506ExpectedDiagnostic(7, 14, "C", 3, 2, 3));
+
+            await VerifyBasicAsync(@"
+Imports System.Collections.Generic
+
+Public Class A
+End Class
+
+Public Class B
+End Class
+
+Public Class C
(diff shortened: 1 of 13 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
```diff
@@ -1900,6 +1878,28 @@ class C1
             VerifyCSharp(source, expectedMetricsText);
         }
 
+        [Fact, WorkItem(2133, "https://github.com/dotnet/roslyn-analyzers/issues/2133")]
+        public void MethodWithAnonymousType()
+        {
+            var source = @"
+class C
+{
+    object M1(C1 c)
```
mavasani: Why was the test moved and new parameter of an undefined type `C1` added? Can we revert this test change?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -559,6 +559,128 @@ End Namespace
             await VerifyBasicAsync(source, additionalText, expected);
         }
 
+        [Fact, WorkItem(2133, "https://github.com/dotnet/roslyn-analyzers/issues/2133")]
+        public async Task CA1506_CountCorrectlyGenericTypes()
```
mavasani: Can you also add this test to `CodeMetricsComputationTests.cs`?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
```diff
@@ -559,6 +559,128 @@ End Namespace
...
+               from n in ints
+               from o in ints
+               from p in ints
+               select p;
+    }
+}");
+        }
+
+        [Fact, WorkItem(2133, "https://github.com/dotnet/roslyn-analyzers/issues/2133")]
+        public async Task CA1506_ExcludeCompilerGeneratedTypes()
```
mavasani: Can you also add this test to `CodeMetricsComputationTests.cs`?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
Evangelink: I did the PR a while ago and it seems that I picked the wrong method during the merge. I am reverting this change.

src/Test.Utilities/CodeMetricsTestsBase.cs
```diff
@@ -33,14 +35,18 @@ protected Project CreateProject(string[] sources, string language = LanguageName
...
 #pragma warning disable CA2000 // Dispose objects before losing scope - Current solution/project takes the dispose ownership of the created AdhocWorkspace
             var solution = new AdhocWorkspace()
 #pragma warning restore CA2000 // Dispose objects before losing scope
                 .CurrentSolution
                 .AddProject(projectId, TestProjectName, TestProjectName, language)
                 .WithProjectCompilationOptions(projectId, options)
-                .AddMetadataReference(projectId, s_corlibReference)
-                .AddMetadataReference(projectId, s_systemCoreReference);
+                .AddMetadataReferences(projectId, references);
+                //.AddMetadataReference(projectId, s_corlibReference)
```
mavasani: Why the commented out code?

src/Test.Utilities/CodeMetricsTestsBase.cs
```diff
@@ -33,14 +35,18 @@ protected Project CreateProject(string[] sources, string language = LanguageName
 
             var projectId = ProjectId.CreateNewId(debugName: TestProjectName);
 
+            var defaultReferences = ReferenceAssemblies.NetFramework.Net48.Default;
+            var references = Task.Run(() => defaultReferences.ResolveAsync(LanguageNames.CSharp, CancellationToken.None)).GetAwaiter().GetResult();
```
mavasani: Can we avoid this test change in this PR?

Evangelink: Well I need to change the references because of the test with the compiler generated attributes.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
Evangelink: Done

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzerTests.cs
Evangelink: Done

src/Test.Utilities/CodeMetricsTestsBase.cs
Evangelink: See this test https://github.com/dotnet/roslyn-analyzers/pull/3315/files#diff-de51a4bb7ce3a178f017a5b53b974678R639

mavasani: Sounds good.

src/Test.Utilities/CodeMetricsTestsBase.cs
```diff
@@ -33,14 +33,16 @@ protected Project CreateProject(string[] sources, string language = LanguageName
...
```
mavasani:
```suggestion
            var references = Task.Run(() => defaultReferences.ResolveAsync(language, CancellationToken.None)).GetAwaiter().GetResult();
```

Evangelink:
@mavasani ready to be merged

PR #3317, opened 2020-02-25
Switch all the enabled by default warnings in NetAnalyzers package to…

mavasani:
… not be build warnings by default.

As per the internal discussion, we will start this package with _no enabled by default warning analyzers_. Each analyzer needs to be brought up to a design meeting/triage to be promoted to a warning analyzer.


mavasani:
Tagging @dotnet/dotnet-analyzers - this PR ensures that the candidate analyzer assembly for insertion in the SDK has no enabled by default warnings to begin with.

PR #3318, opened 2020-02-25
Update CA1034 to allow builder pattern

### LINKED ISSUE #3033: CA1034: Exempt nested classes named Builder, for the builder design pattern
[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers) v2.9.7 (Latest)

I would like FxCop analyzers to exempt nested non-Enum classes named Builder from the CA1034 check.  Classes with this name are useful for implementing the so-called "[builder pattern](https://stackoverflow.com/questions/4416745/builder-design-pattern/57419064#57419064)" for building thread-safe objects, such as settings, that will not be further modified.  One notable use of this pattern I am aware of is the Android class library.

This class involves a public nested class, called the _builder class_ and usually named `Builder`, and a private constructor in the outer class (here called the _building constructor_) that takes a builder object and initializes the thread-safe object.

I am aware that the builder pattern can be implemented without nesting, and [I have done so](https://github.com/peteroupc/MailLib/blob/master/MailLib/PeterO/Mail/MediaTypeBuilder.cs) several times before, but this requires the building constructor to be internal or public, rather than private.  (A side note: For best results, all the fields or properties in the outer class have to be `readonly`.)

Evangelink:
Fix #3033

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/NestedTypesShouldNotBeVisible.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/NestedTypesShouldNotBeVisible.cs
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Immutable;
 using System.Linq;
-using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.Diagnostics;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
 
 namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
 {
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/NestedTypesShouldNotBeVisible.cs
```diff
@@ -102,6 +103,14 @@ public override void Initialize(AnalysisContext analysisContext)
                                 return;
                             }
 
+                            // Allow builder pattern
+                            if (nestedType.Name.EndsWith("Builder", StringComparison.Ordinal) &&
+                                containingType.ContainingType == null &&
+                                !containingType.InstanceConstructors.Any(ctor => ctor.IsExternallyVisible()))
```
Evangelink: Do you want to have other constraints? Do you want to keep the depth constraint I have added?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/NestedTypesShouldNotBeVisible.cs
```diff
@@ -50,12 +51,12 @@ public override void Initialize(AnalysisContext analysisContext)
             analysisContext.RegisterCompilationStartAction(
                 compilationStartContext =>
                 {
-                    Compilation compilation = compilationStartContext.Compilation;
```
mavasani: Why the changes to explicitly use `wellKnownTypeProvider` here, when the underlying implementation for `GetOrCreateTypeByMetadataName` already uses it?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/NestedTypesShouldNotBeVisible.cs
mavasani: I think this looks fine.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/NestedTypesShouldNotBeVisible.cs
Evangelink: The idea was to avoid the overhead of the cache call.

PR #3319, opened 2020-02-25
Create and use Netstandard detection helper

### LINKED ISSUE #3017: Standardize detection of .NET Standard
https://github.com/dotnet/roslyn-analyzers/pull/3011#discussion_r343315242
https://github.com/dotnet/roslyn-analyzers/pull/2972#discussion_r339068188

Evangelink:
Fix #3017 by cherry-picking (and merging) the helper written by @duracellko in #2972 and using it in CA2229.


src/Utilities/Compiler/Extensions/CompilationExtensions.cs
```diff
@@ -27,5 +32,34 @@ internal static class CompilationExtensions
...
         internal static bool TryGetOrCreateTypeByMetadataName(this Compilation compilation, string fullTypeName, [NotNullWhen(returnValue: true)] out INamedTypeSymbol? namedTypeSymbol) =>
             WellKnownTypeProvider.GetOrCreate(compilation).TryGetOrCreateTypeByMetadataName(fullTypeName, out namedTypeSymbol);
+
+        /// <summary>
+        /// Gets a value indicating, whether the compilation of assembly targets .NET Framework.
+        /// This method differentiates between .NET Framework and other frameworks (.NET Core, .NET Standard, .NET 5 in future).
+        /// </summary>
+        /// <param name="compilation">The compilation</param>
+        /// <returns><c>True</c> if the compilation targets .NET Framework; otherwise <c>false</c>.</returns>
+        internal static bool DoesTargetDotNetFramework(this Compilation compilation)
```
mavasani:
```suggestion
        internal static bool TargetsDotNetFramework(this Compilation compilation)
```

mavasani: Tagging @stephentoub @sharwell to confirm if the implementation looks fine.

Evangelink: Maybe we should drop this helper and use the following one instead? https://github.com/dotnet/roslyn-analyzers/blob/master/src/NetAnalyzers/Core/Microsoft.NetFramework.Analyzers/Helpers/SecurityDiagnosticHelpers.cs#L281

Evangelink: Ping @sharwell @stephentoub

Evangelink: @sharwell, @stephentoub could you guys have a look at the implementation?

stephentoub: Is all of the strong naming and version checking necessary?  I'd have expected just the "mscorlib" check to be sufficient, but presumably you found cases where it's not?  FWIW, this is what we use in dotnet/runtime in our tests:
https://github.com/dotnet/runtime/blob/4bdf4684ddf5636c3d8ed8cd7231e40fc069ab0a/src/libraries/Common/tests/CoreFx.Private.TestUtilities/System/PlatformDetection.Windows.cs#L22
but that likely won't suffice here, since it requires the System.Runtime.InteropServices.RuntimeInformation library.

Evangelink: I'm not sure, I simply did a cherry-pick of the helper from another PR. I guess @sharwell is thinking about something in particular when reading his comment https://github.com/dotnet/roslyn-analyzers/pull/2972#discussion_r339068188

duracellko: When I implemented the .NET Framework detection I assumed following:

1. Public key used for strong name of .NET Framework assemblies will never be used outside of .NET Framework (and Silverlight). Therefore it is pretty strong indicator of .NET Framework.
2. System.Object type is not in mscorlib assembly in .NET Core or .NET Standard. But it's not sure, if it is going to stay in the future. Therefore I thought that assembly name and version may not be enough to compare.

Evangelink: @sharwell, @stephentoub could you guys let me know how you want to proceed with this check?

Evangelink: @mavasani Could you check with @sharwell / @stephentoub what would be the right way to move forward with this PR?

mavasani: @stephentoub @sharwell Let us know if you'd prefer any modifications to the checks done by this helper.

PR #3321, opened 2020-02-25
Update CA1801 to ignore serialization methods

### LINKED ISSUE #3039: CA1801 emitted for deserialization constructor when implementing ISerializable.
#### Analyzer package

#### Package Version

v2.9.7 (Latest)

#### Diagnostic ID

Example: [1801](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1801?view=vs-2019)

#### Repro steps

Consider the following class:
```
   [Serializable]
   public class Class1 : ISerializable
   {
      private int _i;

      protected Class1(SerializationInfo info, StreamingContext context)
      {
         _i = (int)info.GetValue("i", typeof(int));
      }

      public void GetObjectData(SerializationInfo info, StreamingContext context)
      {
         if (info == null)
            throw new ArgumentNullException(nameof(info), $"{nameof(info)} is null.");
       
         info.AddValue("i", _i);
      }
   }
```

This warns about `context` being an unused parameter in the ctor, and suggests removing it. But the `ISerializable` interface implies that this ctor exists with exactly that signature according to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.iserializable?view=netframework-4.8).  

#### Expected behavior

No warning about unused parameters suggesting to remove them should be emitted.

#### Actual behavior

CA1801 is emitted for the `context` parameter of the ctor.

Evangelink:
Fix #3039


src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -320,6 +320,19 @@ public static bool IsOnDeserializationImplementation([NotNullWhen(returnValue: t
...
+                method.Parameters.Length == 2 &&
+                method.Parameters[0].Type.Equals(serializationInfoType) &&
+                method.Parameters[1].Type.Equals(streamingContextType);
+
+        public static bool IsGetObjectData([NotNullWhen(returnValue: true)] this IMethodSymbol? method, INamedTypeSymbol? serializationInfoType, INamedTypeSymbol? streamingContextType)
+            => method?.Name == "GetObjectData" &&
+                method.ReturnsVoid &&
+                method.Parameters.Length == 2 &&
+                method.Parameters[0].Type.Equals(serializationInfoType) &&
+                method.Parameters[1].Type.Equals(streamingContextType);
```
Evangelink: This isn't taking into account accessibility nor parameter names. Do we want to be strict?

Also, is it worth creating a `WellKnownMethods` static class and move all the specific construct there?

mavasani: Can this extension method also be used here: https://github.com/dotnet/roslyn-analyzers/blob/5be2c0cf63a36b43a2736587c02c37a83dfa41e1/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ParameterValidationAnalysis/ParameterValidationAnalysis.ParameterValidationDataFlowOperationVisitor.cs#L287-L306

mavasani: And also here: https://github.com/dotnet/roslyn-analyzers/blob/5be2c0cf63a36b43a2736587c02c37a83dfa41e1/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SerializationRulesDiagnosticAnalyzer.cs#L200-L206

Evangelink:
@mavasani Any way to retrigger the various steps? Except from pushing an empty commit.

Evangelink:
Ready to be merged @mavasani

PR #3325, opened 2020-02-26
Update CA1304 to allow user exclusions

### LINKED ISSUE #3204: Is there any way I can add to CA1304's list of ignored calls?
I have a type that wraps a `ResourcesManager` object. The type has two methods, each wrapping a call to  `GetString` method. One of the method has a `CultureInfo` argument, the other have not and uses `GetString` without a specific culture, hence falls on the thread's current culture.

Now the CA1304 documentation specifies that calls to `ResourcesManager.GetString` are ignored by the rule. But since I'm wrapping the call under a different method, the analyzer now warns me about a possible problem with calling the method without a `CultureInfo` argument, even though having a different result depending on culture is the point.

I understand why this happens and that the rule can't possibly know about my API's intent. However, I don't want to disable the rule every time I'm using the method, nor do I want to ignore it as a whole in my projects.

Is there any way I can add the Wrapper.MethodThatCallsGetString method to the list of calls that CA1304 ignores? An attribute, perhaps?

Thanks.

Evangelink:
Fix #3204


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyCultureInfo.cs
```diff
@@ -41,43 +41,48 @@ public override void Initialize(AnalysisContext analysisContext)
 
             analysisContext.RegisterCompilationStartAction(csaContext =>
             {
-                var obsoleteAttributeType = csaContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemObsoleteAttribute);
+                var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(csaContext.Compilation);
```
mavasani: Nit: Can we avoid pulling in `wellKnownTypeProvider` into the analyzers? I don't believe saving an additional `WellKnownTypeProvider.GetOrCreate` for each well known type gives any real performance benefit, and ends up unnecessarily having the analyzer deal with the provider. I think prior implementation has better readability.

Evangelink: Fine by me! Do you want to always avoid using the `WellKnownTypeProvider`?

mavasani: Lets avoid it until we really need it (say like an argument passed to some existing API being invoked). If we find this is indeed a performance bottleneck, we can likely do a repo wide change to have analyzers use WellKnownTypeProvider explicitly when accessing more than one well known type.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SpecifyCultureInfoTests.cs
```diff
@@ -652,14 +653,89 @@ End Sub
 End Class");
         }
 
-        private static DiagnosticResult GetCSharpResultAt(int line, int column, string arg1, string arg2, string arg3) =>
+        [Theory, WorkItem(3204, "https://github.com/dotnet/roslyn-analyzers/issues/3204")]
+        // Diagnostics
+        [InlineData("")]
+        [InlineData("dotnet_code_quality.CA1304.excluded_symbol_names = T:C")]
```
mavasani: Why does this setting trigger a violation?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyCultureInfo.cs
```diff
@@ -42,42 +42,46 @@ public override void Initialize(AnalysisContext analysisContext)
...
                 var cultureInfoType = csaContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemGlobalizationCultureInfo);
-                if (cultureInfoType != null)
+
+                if (cultureInfoType == null)
                 {
-                    csaContext.RegisterOperationAction(oaContext =>
+                    return;
+                }
+
+                var excludedSymbols = csaContext.Options.GetExcludedSymbolNamesWithValueOption(Rule, csaContext.Compilation, csaContext.CancellationToken);
```
mavasani: Ah the problem is you are directly invoking this method, instead of invoking the below extension method: https://github.com/dotnet/roslyn-analyzers/blob/3cc0345042a39d23c8b82fd02d0d2dd298dc92cb/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L154-L162

This extension method ensures that it skips analysis for a symbol even when any of it's containing symbol is marked for exclusion, which is what the user really desires.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SpecifyCultureInfoTests.cs
Evangelink: I have only handled method names in here: https://github.com/dotnet/roslyn-analyzers/pull/3325/files#diff-52e48dbabcedc53847defe0a973ebfb8R58

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyCultureInfo.cs
Evangelink: Ah! I'll fix this now

Issue #2994, opened 2019-11-01
CA1065 should not be raised for ArgumentNullException in IEqualityComparer<T>.GetHashCode


dfederm:
#### Analyzer package
Microsoft.CodeAnalysis.FxCopAnalyzers

#### Package Version
v2.9.2

#### Diagnostic ID
CA1065

#### Repro steps
```
    public class SomeEqualityComparer : IEqualityComparer<string>
    {
        public bool Equals(string x, string y)
        {
            return true;
        }

        public int GetHashCode(string obj)
        {
            if (obj == null)
            {
                // This raises CA1065!
                throw new ArgumentNullException(nameof(obj));
            }

            return 0;
        }
    }
```

#### Expected behavior
No CA violation raised.

The documented expectation for `IEqualityComparer<T>.GetHashCode` is to throw null when given a null object. See: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iequalitycomparer-1.gethashcode?redirectedfrom=MSDN&view=netframework-4.8#exceptions

#### Actual behavior
CA1065 is raised.

sharwell:
Even though the documentation allows for this exception, I do not agree that it is expected. I would only be in favor of this change if CA1065 continued to be reported for `GetHashCode` methods implementing `IEqualityComparer<T>` for a nullable type `T`. In other words, the exception would _only_ be made for the case where _both_ of these conditions are met:

1. Nullable reference types is enabled
2. The type implements `IEqualityComparer<T>` for a non-nullable reference type `T`

dfederm:
@sharwell Are you implying the documentation is wrong then? The "correct" way to implement `IEqualityComparer<T>.GetHashCode` would be to throw on `null` (assuming `T` is a reference type).

Regarding your suggestion, I don't think it makes sense. You're suggesting that an exception only be allowed to throw if nullable reference types is enabled, and then `T` is non-nullable, but wouldn't that be enforced by the compiler regardless?

sharwell:
> Are you implying the documentation is wrong then?

I am saying the documentation is misleading compared to what I would consider a best practice. Equality comparers are more versatile if they allow null for `GetHashCode`, similar to how `EqualityComparer<T>.Default` allows null. However, if the equality comparer explicitly implements a type that does not allow null, then it would be fine to throw this exception if null was passed to `GetHashCode`.

[...]

PR #3328, opened 2020-02-26
CA1065: Add unit test throwing ArgumentNullException in GetHashCode

Evangelink:
Relates to #2994


mavasani:
@Evangelink Is their any conclusion on the underlying issue?

Evangelink:
@mavasani No! The user is ok to send a reproducer with proprietary code only to MS employees (which I am not) so I couldn't find what was the issue. This PR was just showing that the user public/small repro wasn't causing any issue in newer version of the analyzer.

Issue #1241, opened 2017-06-23
False report CA1821 (Remove empty Finalizers)


mavasani:
See https://github.com/dotnet/roslyn/pull/19806/files#diff-4fddad076e8fb11cecceca523ee2aeb5R120

mavasani:
@nguerrera Thanks, you are right that the analyzer is reporting a valid issue on release build. However, there is still an issue in the analyzer - it shouldn't fire if the enclosing method is also conditionally excluded. For example, the following still fires the diagnostic in both release and debug builds.

#if DEBUG
        ~InvisibleEditor()
        {
            Debug.Assert(Environment.HasShutdownStarted, GetType().Name + " was leaked without Dispose being called.");
        }
#endif

sharwell:
> For example, the following still fires the diagnostic in both release and debug builds.

How does the diagnostic fire in this case for release builds?

333fred:
The problem here is that we explicitly flag bodies with just methods marked as conditional, but we don't then check to see if the method is excluded conditionally. I've filed https://github.com/dotnet/roslyn/issues/24569 for exposing the ability to get preprocessor info for a given syntax node, so we can check to see if the method is guarded by `#if DEBUG` flags.

sharwell:
> The problem here is that we explicitly flag bodies with just methods marked as conditional, but we don't then check to see if the method is excluded conditionally.

My understanding is you currently treat all conditional methods as not included in the compilation output. After the change, you will start to treat them as included if the conditional symbol applied to the method definition is defined at the point of the call. Is this correct?

333fred:
Yep, that's correct. The idea is that if the user has a method with just `Debug` calls in it, we can then tell them to wrap with `#if DEBUG`.

Cyberboss:
Here is another repro: https://github.com/tgstation/tgstation-server/blob/master/src/Tgstation.Server.Host/Components/Watchdog/SessionController.cs#L230

Evangelink:
@mavasani We could use a workaround while waiting for a more reliable solution. My idea is to count invocations to method with the conditional attribute if the destructor has any directive.

This would result in potentially some false negative but would certainly kill the false positives raised here. Besides, I think that the current implementation might be a little to strict anyways because a method marked with `ConditionalAttribute` specifies which pre-processor symbol is expected while our current implementation rules them all out. WDYT?

sharwell:
[...]

PR #3341, opened 2020-03-03
Update CA1821 to handle finalizer in debug directive

### LINKED ISSUE #3336: CA1821 doesn't handle debug method call in debug context
#### Analyzer package

#### Package Version

v2.9.8 (Latest) and master

#### Diagnostic ID

Example: [CA1821](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1821)

#### Repro steps

Current unit test is broken as it gives the impression the code is handling the case where it is not.

Current test:

```csharp
public async Task CA1821CSharpTestRemoveEmptyFinalizersWithDebugFailAndDirective()
{
	await VerifyCS.VerifyAnalyzerAsync(@"
public class Class1
{
#if DEBUG
	// Violation will not occur because the finalizer will exist and 
	// contain code when the DEBUG directive is present. When the 
	// DEBUG directive is not present, the finalizer will not exist, 
	// and therefore not be empty.
	~Class1()
	{
		System.Diagnostics.Debug.Fail(""Finalizer called!"");
    }
#endif
}
");
}
```

Test as it should be:

```csharp
public async Task CA1821CSharpTestRemoveEmptyFinalizersWithDebugFailAndDirective()
{
	await new VerifyCS.Test
	{
		TestCode = @"
public class Class1
{
#if DEBUG
	// Violation will not occur because the finalizer will exist and 
	// contain code when the DEBUG directive is present. When the 
	// DEBUG directive is not present, the finalizer will not exist, 
	// and therefore not be empty.
	~Class1()
	{
		System.Diagnostics.Debug.Fail(""Finalizer called!"");
    }
#endif
}
",
		SolutionTransforms =
		{
			(solution, projectId) =>
			{
				var project = solution.GetProject(projectId);
				var parseOptions = ((CSharpParseOptions)project.ParseOptions).WithPreprocessorSymbols("DEBUG");

				return project.WithParseOptions(parseOptions)
					.Solution;
			},
		}
	}.RunAsync();
}
```

#### Expected behavior

No diagnostic is raised.

#### Actual behavior

A diagnostic is raised.

Evangelink:
Fix #3336 and relates to #1241 

Tagging @sharwell


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RemoveEmptyFinalizersTests.cs
```diff
@@ -216,6 +218,86 @@ public class Class1
...
+    }
+#endif
+}
+",
+                SolutionTransforms =
+                {
+                    (solution, projectId) =>
+                    {
+                        var project = solution.GetProject(projectId);
+                        var parseOptions = ((CSharpParseOptions)project.ParseOptions).WithPreprocessorSymbols("DEBUG");
```
sharwell: 😕 Is this supposed to be `RELEASE`?

Evangelink: Yes, this should be `RELEASE`. Maybe the test should be skipped and contain the expected diagnostic. The idea of the test was to show that we currently disregard the actual symbol matching between the `#if` directive and the `ConditionalAttribute`. Another example would be to have `#if NET_20` instead of `#if RELEASE` and the finalizer containing a call to a method marked with the conditional let's say `NETCORE_APP_3.1` there wouldn't be any diagnostic raised while it should.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RemoveEmptyFinalizers.cs
```diff
@@ -42,20 +42,39 @@ public override void Initialize(AnalysisContext analysisContext)
...
-                        !blockOperation.HasAnyExplicitDescendant(op => CanDescendIntoOperation(op, conditionalAttributeType)))
+                    if (context.OperationBlocks.Length != 1 ||
+                        !(context.OperationBlocks[0] is IBlockOperation blockOperation) ||
+                        !(context.OwningSymbol is IMethodSymbol methodSymbol) ||
+                        !methodSymbol.IsDestructor())
+                    {
+                        return;
+                    }
+
+                    var isMethodSurroundedWithDirective = blockOperation.Syntax.Parent.ContainsDirectives;
```
Evangelink: Shall it be "surrounded by" instead of "surrounded with"?

Evangelink:
ping @sharwell

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RemoveEmptyFinalizersTests.cs
```diff
@@ -216,6 +218,86 @@ public class Class1
...
+        [Fact, WorkItem(1241, "https://github.com/dotnet/roslyn-analyzers/issues/1241")]
+        public async Task CA1821_DebugFailAndReleaseDirective_NoDiagnostic_FalsePositive()
+        {
+            await new VerifyCS.Test
+            {
+                TestCode = @"
+public class Class1
+{
+#if RELEASE
+	// There should be a diagnostic because in release the finalizer will be empty.
```
sharwell, line 256: Where is the issue tracking this failure?

Evangelink: It's the issue #1241 mentioned in the `WorkItem`.

sharwell: Thanks, I didn't realize the issue was still open

PR #3342, opened 2020-03-04
Use Visual Studio 2019 for builds

sharwell:


mavasani:
Seems builds are failing even without executing.

mavasani:
Tagging @JoeRobich in case he has ideas on how to resolve this.

JoeRobich:
@sharwell Just curious, Is there a catalog of queues that we can specify?

sharwell:
@JoeRobich I found it here: https://github.com/dotnet/arcade-pool-provider/blob/master/src/appsettings.dncengpublic-prod.json

PR #3344, opened 2020-03-04
Ignore XamlGeneratedNamespace.GeneratedInternalTypeHelper

tmat:
Workaround for https://github.com/dotnet/wpf/issues/2690

### DIFF
--- a/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
+++ b/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
@@ -339,6 +339,13 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
 
                 void reportDeclareNewApi(ISymbol symbol, bool isImplicitlyDeclaredConstructor, string publicApiName)
                 {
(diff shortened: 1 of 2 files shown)


src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -339,6 +339,13 @@ void reportDiagnosticAtLocations(DiagnosticDescriptor descriptor, ImmutableDicti
 
                 void reportDeclareNewApi(ISymbol symbol, bool isImplicitlyDeclaredConstructor, string publicApiName)
                 {
+                    // TODO: workaround for https://github.com/dotnet/wpf/issues/2690
```
mavasani: Do we need to file a tracking issue in roslyn-analyzers repo to revert this change once https://github.com/dotnet/wpf/issues/2690 is fixed? If not, can you add a comment to https://github.com/dotnet/wpf/issues/2690 that this workaround needs to be reverted once that issue is fixed?

tmat: https://github.com/dotnet/roslyn-analyzers/issues/3345

PR #3348, opened 2020-03-05
Update CA1062 to support Guard fluent validation

### LINKED ISSUE #3340: CA1062 should support fluent syntax
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

CA1062

#### Repro steps
```csharp        
public static void Foo(string bar)
{
       Guard.Argument(bar, nameof(bar)).NotNull();
       var length = bar.Length; // Warning CA1062
}
```

#### Expected behavior

Add configuration options to support fluent syntaxes.
Maybe something like dotnet_code_quality.null_check_validation_methods = Argument->NotNull|...

#### Actual behavior

Only 'usual' methods that check directly the value can be marked as null_check_validation_methods.

Evangelink:
Fix #3340

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ParameterValidationAnalysis/ParameterValidationAnalysis.ParameterValidationDataFlowOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ParameterValidationAnalysis/ParameterValidationAnalysis.ParameterValidationDataFlowOperationVisitor.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ParameterValidationAnalysis/ParameterValidationAnalysis.ParameterValidationDataFlowOperationVisitor.cs
```diff
@@ -352,8 +353,22 @@ private void ProcessRegularInvocationOrCreation(IMethodSymbol targetMethod, Immu
                         }
                     }
                 }
+
+                // Basic workaround to allow use of Guard fluent syntax as per https://github.com/dotnet/roslyn-analyzers/issues/3340
+                if (targetMethod.OriginalDefinition.Name == "NotNull" &&
```
Evangelink: Potential additional checks:

1. Enclosing class is `static`
2. `NotNull` method is `static`
3. `Argument` method is `static`
4. `NotNull` parameter is of type `ArgumentInfo`.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ParameterValidationAnalysis/ParameterValidationAnalysis.ParameterValidationDataFlowOperationVisitor.cs
```diff
@@ -352,8 +353,22 @@ private void ProcessRegularInvocationOrCreation(IMethodSymbol targetMethod, Immu
...
+                    notNullInvocation.Arguments.Length >= 1 &&
+                    notNullInvocation.Arguments[0].Value is IInvocationOperation argumentInvocation &&
+                    argumentInvocation.TargetMethod.Name == "Argument" &&
+                    argumentInvocation.Arguments.Length >= 1)
+                {
+                    MarkValidatedLocations(argumentInvocation.Arguments[0]);
+                }
             }
 
+            private readonly ConcurrentDictionary<IArgumentOperation, bool> a = new ConcurrentDictionary<IArgumentOperation, bool>();
```
Evangelink: Un-needed

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/ParameterValidationAnalysis/ParameterValidationAnalysis.ParameterValidationDataFlowOperationVisitor.cs
Evangelink: I could also add one more test for the first argument being of type `T` because we don't handle overload of the `Argument` method that takes an `Expression<Func<T>>` argument.

PR #3351, opened 2020-03-06
Update CA1801 to handle arrowed methods throwing

### LINKED ISSUE #2846: CA1801: False report for unimplemented method with expression body and return value
#### Analyzer package

#### Analyzer

ReviewUnusedParametersAnalyzer

#### Diagnostic ID

[CA1801](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1801-review-unused-parameters)

#### Repro steps

Define method with unused parameter that returns a value and throws NotImplementedException using expression bodied syntax.

#### Code sample
```
// Error CA1801  Parameter value of method Foo is never used.Remove the parameter or use it in the method body.
public int Foo(int value) => throw new NotImplementedException();

// ok
public void Bar(int value) => throw new NotImplementedException();

// ok
public int Baz(int value)
{
}

// Error CA1801  Parameter value of method Foo is never used.Remove the parameter or use it in the method body.
public int Foo2(int value) => throw new NotSupportedException();

// ok
public void Bar2(int value) => throw new NotSupportedException();

// ok
public int Baz2(int value)
{
    throw new NotSupportedException();
}
```

Evangelink:
Fix #2846

### DIFF
--- a/src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
+++ b/src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
@@ -2,7 +2,6 @@
 
 #if HAS_IOPERATION
 
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
--- a/src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
+++ b/src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
@@ -60,6 +60,18 @@ namespace Analyzer.Utilities.Extensions
                         innerOperation = exprStatement.Operation;
                     }
 
+                    // Special case for methods with a return type and using arrowed expression where the throw-statement
+                    // is wrapped.
+                    // For example: public int GetSomething(int val) => throw new NotImplementedException();
+                    if (innerOperation.Kind == OperationKind.Return &&
+                        innerOperation is IReturnOperation @return &&
+                        @return.ReturnedValue.Kind == OperationKind.Conversion &&
+                        @return.ReturnedValue is IConversionOperation conversion &&
+                        conversion.Operand.Kind == OperationKind.Throw)
+                    {
+                        innerOperation = conversion.Operand;
+                    }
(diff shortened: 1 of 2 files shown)


src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
```diff
@@ -60,6 +60,18 @@ static bool IsSingleStatementBody(IBlockOperation body)
                         innerOperation = exprStatement.Operation;
                     }
 
+                    // Special case for methods with a return type and using arrowed expression where the throw-statement
```
Evangelink: I am wondering if some ticket shall be created on roslyn's side or if this is the expected behavior. This is not the behavior I was expecting but I can understand why it is like this. 

Note that as the helper is updated we might have some impact on `MarkMembersAsStatic` (the only other rule relying on this helper).

mavasani: Can we instead define a variant of `GetTopmostExplicitDescendants`, say `bool TryGetSingleTopmostExplicitDescendant(IOperation operation, out var singleTopmostExplicitOperation)`: https://github.com/dotnet/roslyn-analyzers/blob/8dd4be4fa590db4512a1089c4df1b970e0d4a527/src/Utilities/Compiler/Extensions/IOperationExtensions.cs#L154-L159

This entire if-body can then probably be implemented as:
```csharp
                    if (TryGetSingleTopmostExplicitDescendant(out var innerOperation) &&
                        innerOperation is IThrowOperation throwOperation &&
                        throwOperation.GetThrownExceptionType() is ITypeSymbol createdExceptionType)
                    {
                        if (Equals(context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNotImplementedException), createdExceptionType.OriginalDefinition)
                            || Equals(context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNotSupportedException), createdExceptionType.OriginalDefinition))
                        {
                            return true;
                        }
                    }
```

mavasani: Also see https://github.com/dotnet/roslyn/pull/42408 for similar bug fix in Roslyn.

PR #3352, opened 2020-03-06
RS0046: Avoid Opt suffix in nullable-enabled code

### LINKED ISSUE #2920: Avoid 'Opt' suffix in nullable-enabled code
This will help with cleaning up the code base after nullable reference types is adopted.

https://github.com/dotnet/roslyn/pull/39053#discussion_r333089641

Evangelink:
Fix #2920


src/Roslyn.Diagnostics.Analyzers/UnitTests/AvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Roslyn.Diagnostics.CSharp.Analyzers.CSharpAvoidOptSuffixForNullableEnableCode,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Roslyn.Diagnostics.Analyzers.UnitTests
+{
+    public class AvoidOptSuffixForNullableEnableCodeTests
+    {
+        [Fact]
+        public async Task RS0037_CSharp8_NullableEnabledCode_Diagnostic()
```
Evangelink: Test is failing with `error AD0001: Analyzer 'Roslyn.Diagnostics.CSharp.Analyzers.CSharpAvoidOptSuffixForNullableEnableCode' threw an exception of type 'System.MissingMethodException' with message 'Method not found: 'Microsoft.CodeAnalysis.CSharp.NullableContextOptions Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions.get_NullableContextOptions()'.'.`

The tests are using Roslyn 3.5 so I think that everything shall be ok.

src/Roslyn.Diagnostics.Analyzers/UnitTests/AvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    public void Method1(string? sOpt)
+    {
+    }
+}",
+                LanguageVersion = LanguageVersion.CSharp8
+            }.RunAsync();
+        }
+
+        [Fact]
+        public async Task RS0037_CSharp8_NonNullableEnabledCode_NoDiagnostic()
```
Evangelink: Same `AD0001` error.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,26 @@
...
+    public sealed class CSharpAvoidOptSuffixForNullableEnableCode : AvoidOptSuffixForNullableEnableCode
+    {
+        protected override bool IsNullableEnabledContext(CompilationOptions compilationOptions, IEnumerable<ParseOptions?> parseOptions)
+        {
+            var hasAnyCSharp8Part = parseOptions.OfType<CSharpParseOptions>().Any(option => option.LanguageVersion >= LanguageVersion.CSharp8);
+
+            return hasAnyCSharp8Part && IsNullableEnabledContext(compilationOptions);
+        }
+
+        private static bool IsNullableEnabledContext(CompilationOptions compilationOptions)
```
Evangelink: Split the method to avoid method not found exception on older roslyn versions.

src/Roslyn.Diagnostics.Analyzers/UnitTests/AvoidOptSuffixForNullableEnableCodeTests.cs
Evangelink: @sharwell Any idea what could be causing this `MissingMethodException`?

mavasani:
@Evangelink 
@sharwell is the expert in this space. I will let him review/help you out here.

mavasani:

src/Roslyn.Diagnostics.Analyzers/Core/AvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,49 @@
...
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSymbolAction(context =>
+            {
+                if (context.Symbol.Name.EndsWith("Opt", System.StringComparison.Ordinal) &&
+                    context.Symbol.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is { } node &&
+                    context.Compilation.GetSemanticModel(node.SyntaxTree).GetNullableContext(node.SpanStart).AnnotationsEnabled())
```
Evangelink, line 42: I get a warning `RS1030` for calling `Compilation.GetSematicModel` but `GetNullableContext` only exists on the `SemanticModel`. Should I go for a C# specific analyzer based on syntax analysis?

sharwell: > Should I go for a C# specific analyzer based on syntax analysis?

Normally I would be looking for another way, but in this case only C# supports nullable reference types.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,47 @@
...
+            }, SyntaxKind.Parameter);
+
+            context.RegisterSyntaxNodeAction(context =>
+            {
+                var fieldDeclaration = (FieldDeclarationSyntax)context.Node;
+
+                foreach (var variable in fieldDeclaration.Declaration.Variables)
+                {
+                    if (variable.Identifier.Text.EndsWith("Opt", System.StringComparison.Ordinal) &&
+                        context.SemanticModel.GetNullableContext(variable.SpanStart).AnnotationsEnabled())
```
Evangelink, line 39: In theory it's possible to have some parts of a field declaration after the `#nullable enable` but I don't think that would happen in Roslyn so we might want to check the context at the field declaration level rather than each variable.

Evangelink:
@sharwell, @mavasani PR ready for review.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpAvoidOptSuffixForNullableEnableCode : AvoidOptSuffixForNullableEnableCode
```
mavasani, line 14: Nullable support for VB is not likely in near future. Can we merge `AvoidOptSuffixForNullableEnableCode` into this type and only have `CSharpAvoidOptSuffixForNullableEnableCode`?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,47 @@
...
+            context.RegisterSyntaxNodeAction(context =>
+            {
+                var parameter = (ParameterSyntax)context.Node;
+
+                if (parameter.Identifier.Text.EndsWith("Opt", System.StringComparison.Ordinal) &&
+                    context.SemanticModel.GetNullableContext(parameter.SpanStart).AnnotationsEnabled())
+                {
+                    context.ReportDiagnostic(parameter.Identifier.CreateDiagnostic(Rule));
+                }
+            }, SyntaxKind.Parameter);
```
mavasani, line 30: Can we also handle local symbols? Given that locals can we declared with multiple different syntax nodes, it might be good to write this as an operation action for `OperationKind.VariableDeclarator`.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            context.RegisterSyntaxNodeAction(context =>
+            {
+                var parameter = (ParameterSyntax)context.Node;
+
+                if (parameter.Identifier.Text.EndsWith("Opt", System.StringComparison.Ordinal) &&
+                    context.SemanticModel.GetNullableContext(parameter.SpanStart).AnnotationsEnabled())
+                {
+                    context.ReportDiagnostic(parameter.Identifier.CreateDiagnostic(Rule));
+                }
```
mavasani, line 29: Can this be extracted into a shared static helper or local function that takes `name`, `syntax` and `addDiagnostic` as parameters?

```suggestion
                if (parameter.Identifier.Text.EndsWith("Opt", System.StringComparison.Ordinal) &&
                    context.SemanticModel.GetNullableContext(parameter.SpanStart).AnnotationsEnabled())
                {
                    context.ReportDiagnostic(parameter.Identifier.CreateDiagnostic(Rule));
                }
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,47 @@
...
+            }, SyntaxKind.Parameter);
+
+            context.RegisterSyntaxNodeAction(context =>
+            {
+                var fieldDeclaration = (FieldDeclarationSyntax)context.Node;
+
+                foreach (var variable in fieldDeclaration.Declaration.Variables)
```
mavasani, line 36: I think this will also get handled if we register an operation action for `OperationKind.VariableDeclarator`?

src/Roslyn.Diagnostics.Analyzers/Core/AvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,30 @@
...
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdMessage), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdDescription), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            RoslynDiagnosticIds.AvoidOptSuffixForNullableEnableCodeRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.RoslynDiagnosticsDesign,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: false,
```
mavasani, line 23: Why? This looks to be a good rule!

```suggestion
            isEnabledByDefault: true,
```

src/Roslyn.Diagnostics.Analyzers/UnitTests/AvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            await new VerifyCS.Test
+            {
+                LanguageVersion = LanguageVersion.CSharp8,
+                TestCode =
+                        @"
+#nullable enable
+
+public class Class1
+{
+    private Class1? [|_instanceOpt|];
```
mavasani, line 26: We should write a code fix that invokes [Renamer.RenameSymbolAsync](http://sourceroslyn.io/#Microsoft.CodeAnalysis.Workspaces/Rename/Renamer.cs,17). I think this is **absolutely** required as most users would prefer to do a FixAll operation to get rid of the opt suffix.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Rename;
+using Microsoft.CodeAnalysis.Text;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider))]
```
Evangelink, line 16: I have seen a few patterns used for the codefix providers and I am not sure which one shall be used.

src/Roslyn.Diagnostics.Analyzers/UnitTests/CSharpAvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+public class Class1
+{
+    private Class1? _instance, otherInstance;
+
+    public void Method1(string? s)
+    {
+        string? other, otherLocalOpt;
+    }
+}",
```
Evangelink, line 55: This isn't working and I am not sure what's going on. Breakpoints on the codefix provider are never reached.

mavasani: @sharwell?

sharwell: What's the test failure message? If it fails during diagnostic validation, the code fixes will not have run yet.

Evangelink: Failure message:

```
  Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessageException : Context: Diagnostics of fixed state
    Mismatch between number of diagnostics returned, expected "0" actual "1"
    
    Diagnostics:
    // /0/Test0.cs(10,24): warning RS0046: Avoid the 'Opt' suffix in a nullable-enabled code
    VerifyCS.Diagnostic().WithSpan(10, 24, 10, 37),
```

Evangelink: My fixed code is invalid, I forgot to rename the variables!!! On previous versions of the test sdk the messages were a lot better we used to have the full diff displayed which was making comparison easier.

Evangelink:
Guys sorry for the slow pace PR, I think the PR is now ready. @mavasani Is there some doc to update for the RS analyzers?

Evangelink:
Ping @mavasani

Evangelink:
@sharwell @mavasani let me know if there is anything else to do.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+            return Task.CompletedTask;
+        }
+
+        private static async Task<Document> RemoveOptSuffixOnVariableAsync(Document document, TextSpan sourceSpan, CancellationToken cancellationToken)
+        {
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+
+            var variable = root.FindNode(sourceSpan, getInnermostNodeForTie: true);
+            var variableSymbol = semanticModel.GetDeclaredSymbol(variable, cancellationToken);
```
mavasani, line 48: This can return `null` for erroneous code. I would recommend moving these 4 lines that compute the `variableSymbol` inside the loop before `RegisterCodeFix` and add a null check for `variableSymbol` before invoking `RegisterCodeFix`

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            var newSolution = await Renamer.RenameSymbolAsync(
+                    document.Project.Solution,
+                    variableSymbol,
+                    variableSymbol.Name.Substring(0, variableSymbol.Name.Length - CSharpAvoidOptSuffixForNullableEnableCode.OptSuffix.Length),
```
mavasani, line 53: This can lead to name clash with an existing symbol in scope with matching name (for example, `string? name, nameOpt;`). I would recommend invoking [SemanticModel.LookupSymbols](http://sourceroslyn.io/#Microsoft.CodeAnalysis/Compilation/SemanticModel.cs,462) API for the new name without `Opt` suffix before invoking `RegisterCodeFix` so we don't offer a code fix for such cases and let users manually fix it up.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
mavasani: Also recommend adding a defensive check for null value returned by `FindNode` before invoking `GetDeclaredSymbol`

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+            => WellKnownFixAllProviders.BatchFixer;
+
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var title = RoslynDiagnosticsAnalyzersResources.TestExportsShouldNotBeDiscoverableCodeFix;
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
```
mavasani, line 32: Please do no use `CodeAction.Create` - this should have fired [RS0005](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs#L11), but seems it does not (please file an issue for that).

Use the pattern like this: https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/OverrideEqualsOnOverloadingOperatorEquals.Fixer.cs#L65-L72

https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/OverrideEqualsOnOverloadingOperatorEquals.Fixer.cs#L46-L51

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+    public sealed class CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider : CodeFixProvider
+    {
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(CSharpAvoidOptSuffixForNullableEnableCode.Rule.Id);
+
+        public override FixAllProvider GetFixAllProvider()
+            => WellKnownFixAllProviders.BatchFixer;
+
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var title = RoslynDiagnosticsAnalyzersResources.TestExportsShouldNotBeDiscoverableCodeFix;
```
mavasani, line 27:
```suggestion
            var title = RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdTitleCodeFixTitle;
```

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs
```diff
@@ -49,7 +49,7 @@ internal static class RoslynDiagnosticIds
         public const string DoNotCallGetTestAccessorRuleId = "RS0043";
         public const string CreateTestAccessorRuleId = "RS0044";
         public const string ExposeMemberForTestingRuleId = "RS0045";
-
+        public const string AvoidOptSuffixForNullableEnableCodeRuleId = "RS0046";
```
mavasani, line 52: This was intentionally separated by a blank line for a new ID range used for temporary `RS01xx` diagnostics.

```suggestion
        public const string AvoidOptSuffixForNullableEnableCodeRuleId = "RS0046";
        
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+                var variableSymbol = semanticModel.GetDeclaredSymbol(variable, context.CancellationToken);
+                if (variableSymbol == null)
+                {
+                    continue;
+                }
+
+                var newName = variableSymbol.Name.Substring(0, variableSymbol.Name.Length - CSharpAvoidOptSuffixForNullableEnableCode.OptSuffix.Length);
+
+                // There is no symbol matching the new name so we can register the codefix
+                if (semanticModel.LookupSymbols(diagnostic.Location.SourceSpan.Start, variableSymbol.ContainingType, newName).IsEmpty)
```
Evangelink, line 53: I am not entirely sure to understand which value should be used as `position` (first argument) of this method call.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
Evangelink: See #3684

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+            var title = RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdTitleCodeFixTitle;
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+                if (root == null)
+                {
+                    continue;
+                }
+
```
mavasani, line 36: root cannot be null
```suggestion
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
mavasani: I think this is fine.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+            var newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, variableSymbol, newName, document.Project.Solution.Options, cancellationToken)
+                .ConfigureAwait(false);
+
+            return newSolution.GetDocument(document.Id)!;
+        }
+
+        // Needed for Telemetry (https://github.com/dotnet/roslyn/issues/4919)
+        private class MyCodeAction : DocumentChangeAction
+        {
+            public MyCodeAction(string title, Func<CancellationToken, Task<Document>> createChangedDocument, string equivalenceKey)
```
mavasani, line 76:
```suggestion
            public MyCodeAction(string title, Func<CancellationToken, Task<Solution>> createChangedSolution, string equivalenceKey)
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+        private static async Task<Document> RemoveOptSuffixOnVariableAsync(Document document, ISymbol variableSymbol, string newName, CancellationToken cancellationToken)
+        {
+            var newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, variableSymbol, newName, document.Project.Solution.Options, cancellationToken)
+                .ConfigureAwait(false);
+
+            return newSolution.GetDocument(document.Id)!;
+        }
+
+        // Needed for Telemetry (https://github.com/dotnet/roslyn/issues/4919)
+        private class MyCodeAction : DocumentChangeAction
```
mavasani, line 74:
```suggestion
        private class MyCodeAction : SolutionChangeAction
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+
+        private static async Task<Document> RemoveOptSuffixOnVariableAsync(Document document, ISymbol variableSymbol, string newName, CancellationToken cancellationToken)
+        {
+            var newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, variableSymbol, newName, document.Project.Solution.Options, cancellationToken)
+                .ConfigureAwait(false);
+
+            return newSolution.GetDocument(document.Id)!;
+        }
```
mavasani, line 71:
```suggestion
        private static Task<Solution> RemoveOptSuffixOnVariableAsync(Document document, ISymbol variableSymbol, string newName, CancellationToken cancellationToken)
            => Renamer.RenameSymbolAsync(document.Project.Solution, variableSymbol, newName, document.Project.Solution.Options, cancellationToken);
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
mavasani: Please add a test where lookup symbols will return non-empty symbols, i.e. we bail out due to name clash.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+                var variable = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+                if (variable == null)
+                {
+                    continue;
+                }
+
+                var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+                var variableSymbol = semanticModel.GetDeclaredSymbol(variable, context.CancellationToken);
```
mavasani, line 44: Can you try to write some test with error code where this can return null? Honestly, I am not sure but worth trying out few code snippets.

src/Roslyn.Diagnostics.Analyzers/UnitTests/CSharpAvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                FixedCode = @"
+#nullable enable
+
+public class Class1
+{
+    private Class1? _instance, otherInstance;
+
+    public void Method1(string? s)
+    {
+        string? local, otherLocal;
```
mavasani, line 41: Might be worth adding read references to these symbols to ensure we are invoking the renamer and not just replacing the identifier text of the declaration.

src/Roslyn.Diagnostics.Analyzers/UnitTests/CSharpAvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -0,0 +1,110 @@
...
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Roslyn.Diagnostics.CSharp.Analyzers.CSharpAvoidOptSuffixForNullableEnableCode,
+    Roslyn.Diagnostics.CSharp.Analyzers.CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider>;
+
+namespace Roslyn.Diagnostics.Analyzers.UnitTests
+{
+    public class CSharpAvoidOptSuffixForNullableEnableCodeTests
+    {
+        [Fact]
+        public async Task RS0046_CSharp8_NullableEnabledCode_Diagnostic()
```
mavasani, line 15: Can you also add a test to verify no diagnostic for below case:

```csharp
#nullable enable
public class Class1
{
    private Class1 _instanceOpt, otherInstanceOpt;
    public void Method1(string sOpt)
    {
        string localOpt, otherLocalOpt;
    }
}
```

mavasani: Can you also add positive and negative tests for nullable value types and value types in nullable enable context?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                var variableDeclarator = (VariableDeclaratorSyntax)context.Node;
+                ReportOnInvalidIdentifier(variableDeclarator.Identifier, context.SemanticModel, context.ReportDiagnostic);
+            }, SyntaxKind.VariableDeclarator);
+        }
+
+        private static void ReportOnInvalidIdentifier(SyntaxToken identifier, SemanticModel semanticModel, Action<Diagnostic> reportAction)
+        {
+            if (identifier.Text.EndsWith(OptSuffix, StringComparison.Ordinal) &&
+                semanticModel.GetNullableContext(identifier.SpanStart).AnnotationsEnabled())
```
mavasani, line 62: I think we want to also check if the symbol type is a nullable reference type here.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
Evangelink: I am having some trouble with the test. It is complaining about a mismatch of diagnostic but I have declared them at the correct locations... Any idea what's wrong?

src/Roslyn.Diagnostics.Analyzers/UnitTests/CSharpAvoidOptSuffixForNullableEnableCodeTests.cs
Evangelink: A diagnostic is reported on your example. I don't know how to tell if the symbol is nullable or not, any idea?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
Evangelink: Also, it seems that the `SemanticModel.LookupSymbols` is only looking backward so we end-up with wrong results/state for other parameters in a method declaration. Or maybe I am not passing the right arguments to the call.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
Evangelink: I have tried a couple of tests and they all return a non-null symbol. If you have any suggestion I am happy to add a test. @sharwell Any idea?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
mavasani: Can you try to find the ancestor `BlockSytnax` and give the span end position as the position argument?

Evangelink: I have tried with `Start` and `End` of the ancestor `BlockSyntax` and that doesn't solve the issue (on the opposite, the codefix no longer works).

mavasani: Can you file a tracking issue? I think this is possiblly a corner case enough to not block this PR.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// RS0046: Avoid 'Opt' suffix for nullable enable code
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpAvoidOptSuffixForNullableEnableCode : DiagnosticAnalyzer
+    {
+        internal const string OptSuffix = "Opt";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdTitle), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdMessage), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdDescription), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
```
Youssef1313, line 26: The "RuleId" part in the following names seems inconsistent with the other analyzers. I also think the name reads better without it.

- `AvoidOptSuffixForNullableEnableCodeRuleIdTitle`
- `AvoidOptSuffixForNullableEnableCodeRuleIdMessage`
- `AvoidOptSuffixForNullableEnableCodeRuleIdDescription`

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,74 @@
...
+        public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSyntaxNodeAction(context =>
+            {
+                var parameter = (ParameterSyntax)context.Node;
```
Youssef1313, line 48: Is this cast safe?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (variableSymbol == null)
+                {
+                    continue;
+                }
+
+                var newName = variableSymbol.Name.Substring(0, variableSymbol.Name.Length - CSharpAvoidOptSuffixForNullableEnableCode.OptSuffix.Length);
```
Youssef1313, line 45: I think this could throw an exception if the `variableSymbol.Name` is shorter than `OptSuffix`?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    /// <summary>
+    /// RS0046: Avoid 'Opt' suffix for nullable enable code
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpAvoidOptSuffixForNullableEnableCode : DiagnosticAnalyzer
```
Youssef1313, line 20: Why not inherit from a base "AvoidOptSuffixForNullableEnableCode" (that will inturn inherit from DiagnosticAnalyzer) instead of direct inheritance from DiagnosticAnalyzer?

That will allow easily adding a VB analyzer.

Evangelink: I started with the base class but after some chat we decided it was unlikely that the case will happen for vbnet. Note that this analyzer is mainly (only) meant to help on roslyn.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
Evangelink: Yes because I am only passing `SyntaxKind.Parameter`.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
Evangelink: That's probably better to check! Thanks

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
Evangelink: See #3680 3707

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    /// <summary>
+    /// RS0046: Avoid 'Opt' suffix for nullable enable code
```
mavasani, line 17: I think we may want to eventually promote this to a CA rule as an IDE suggestion by default. Can you open a tracking issue for port?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+    public sealed class CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider : CodeFixProvider
+    {
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(CSharpAvoidOptSuffixForNullableEnableCode.Rule.Id);
+
+        public override FixAllProvider GetFixAllProvider()
+            => WellKnownFixAllProviders.BatchFixer;
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var title = RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeRuleIdTitleCodeFixTitle;
```
mavasani, line 27:
```suggestion
            var title = RoslynDiagnosticsAnalyzersResources.AvoidOptSuffixForNullableEnableCodeCodeFixTitle;
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+                .ConfigureAwait(false);
+
+        // Needed for Telemetry (https://github.com/dotnet/roslyn/issues/4919)
+        private class MyCodeAction : SolutionChangeAction
+        {
+            public MyCodeAction(string title, Func<CancellationToken, Task<Solution>> createChangedSolution, string equivalenceKey)
+                : base(title, createChangedSolution, equivalenceKey)
+            {
+            }
+        }
```
mavasani, line 71: You can delete this for Roslyn.Diagnostics.Analyzers as Sam recommended and revert back to `CodeAction.Create`
```suggestion
```

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
Evangelink: See #3710

mavasani: Sounds reasonable- let close the issue and we can revisit if we get a request.

PR #3360, opened 2020-03-09
Update CA1711 to report on enum with Flag or Flags suffix

### LINKED ISSUE #3355: "Flag" or "Flags" Suffix for Enums in CA1711
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

Example: [CA1711](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1711?view=vs-2019)

#### Repro steps

1. Create a new C# project, install FxCops and configure to AllRulesDefault
2. Paste the following in a namespace in a .cs file:

```
    public enum BitFlag
    {
        None,
    }

    [Flags]
    public enum BitFlags
    {
        None,
    }

```

#### Expected behavior
Warning, as this is not allowed in the [Design Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations):

> DO NOT use "Flag" or "Flags" suffixes in enum type names.

#### Actual behavior
No warning.

Evangelink:
Fix #3355


Evangelink:
> 
> 
> This is a breaking change in CA1711, though is reasonable as per the design guidelines. I am also concerned that this might make the rule noisy. Can we add options, say `allowed_suffixes` and/or `additional_required_suffixes`, so that users can customize this behavior to meet their requirements?

Not sure if you want me to keep current hard-coded change in addition to this new user-option? In which case, isn't it better to have a separate ticket + PR for the user-option?

Or do you want me to revert the change and only add the user-option? I think (and you seem to agree) that the behavior matches the rule description so it feels weird to ask users to always provide this option. Maybe your preference is that the rule description gets updated?

mavasani:
Given the rule is not enabled by default, I am fine with the change. However, I am concerned that anyone who has enabled the rule would be broken with no alternative but to turn off the rule if they don't like this change. Can we add the option(s) which enables the user to switch back to the state prior to this change in this PR itself?

Evangelink:
As discussed I have handled only `allowed_suffixes` as the other option will require more thoughts on how to define it.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotHaveIncorrectSuffix.cs
```diff
@@ -147,12 +149,17 @@ public override void Initialize(AnalysisContext analysisContext)
                                 return;
                             }
 
+                            var allowedSuffixes = namedTypeSymbol.DeclaringSyntaxReferences
+                                .Select(syntaxRef => symbolAnalysisContext.Options.GetStringOptionValue(EditorConfigOptionNames.AllowedSuffixes, TypeNoAlternateRule, syntaxRef.SyntaxTree, symbolAnalysisContext.Compilation, symbolAnalysisContext.CancellationToken))
```
Evangelink, line 153: We might have multiple syntax-trees and I wasn't sure how to best handle it so I decided to merge the various results. Let me know if you prefer something else.

Evangelink, line 153: I was thinking about returning `ImmutableHashSet<string>` instead of string but that doesn't seem to bring much.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotHaveIncorrectSuffixTests.cs
```diff
@@ -2015,6 +2015,79 @@ public async Task CA1711_Basic_Diagnostic_FieldNameEndsWithEx()
...
+
+            await VerifyVB.VerifyAnalyzerAsync(@"
+Public Enum SomeEnumFlags
+    X
+End Enum",
+                GetCSharpResultAt(2, 13, IdentifiersShouldNotHaveIncorrectSuffixAnalyzer.TypeNoAlternateRule, "SomeEnumFlags", IdentifiersShouldNotHaveIncorrectSuffixAnalyzer.FlagsSuffix));
+        }
+
+        [Fact]
+        public async Task CA1711_AllowedSuffixes_NoDiagnostic()
```
Evangelink, line 2053: I have added only one test for this new option, let me know if you want some specific tests.

Evangelink:
@mavasani anything else to do on this PR?

Evangelink:
Ping @mavasani

docs/Analyzer Configuration.md
```diff
@@ -693,3 +693,12 @@ Examples:
 |`dotnet_code_quality.CA1806.additional_use_results_methods = MyMethod1\|MyMethod2` | Matches all methods named either 'MyMethod1' or 'MyMethod2' in the compilation
 |`dotnet_code_quality.CA1806.additional_use_results_methods = M:NS.MyType.MyMethod(ParamType)` | Matches specific method 'MyMethod' with given fully qualified signature
 |`dotnet_code_quality.CA1806.additional_use_results_methods = M:NS1.MyType1.MyMethod1(ParamType)\|M:NS2.MyType2.MyMethod2(ParamType)` | Matches specific methods 'MyMethod1' and 'MyMethod2' with respective fully qualified signature
+
+### Allowed suffixes
+Option Name: `allowed_suffixes`
+
+Configurable Rules: [CA1711](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1711)
```
Youssef1313, line 700: nit: prevent localized links

```suggestion
Configurable Rules: [CA1711](https://docs.microsoft.com/visualstudio/code-quality/ca1711)
```

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -175,5 +175,11 @@ internal static partial class EditorConfigOptionNames
         ///      with an optional "M:" prefix.
         /// </summary>
         public const string AdditionalUseResultsMethods = "additional_use_results_methods";
+
+        /// <summary>
+        /// String option to configure allowed suffixed (separated by '|').
+        /// Configurable rule: CA1711 (https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1711).
```
Youssef1313, line 181:
```suggestion
        /// Configurable rule: CA1711 (https://docs.microsoft.com/visualstudio/code-quality/ca1711).
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotHaveIncorrectSuffix.cs
mavasani: Lets just stick to the first tree, see https://github.com/dotnet/roslyn-analyzers/issues/3494 for context.

Also its cheaper to use `symbol.Locations[0].SyntaxTree` instead of `DeclaringSyntaxReferences`, I think the latter does not cache things and/or has more allocations.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotHaveIncorrectSuffix.cs
```diff
@@ -224,7 +248,12 @@ public override void Initialize(AnalysisContext analysisContext)
 
                     string name = memberSymbol.Name;
 
-                    if (name.HasSuffix(ExSuffix))
+                    var allowedSuffixes = memberSymbol.DeclaringSyntaxReferences
```
mavasani, line 251: You need the same change here.

Evangelink: 🤦 Sorry

PR #3362, opened 2020-03-09
Update CA1012 to report on protected internal constructors

### LINKED ISSUE #3358: Check for protected internal constructors in CA1012
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

Example: [CA1012](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1012?view=vs-2019)

#### Repro steps

1. Create a new C# project
2. Install FxCops Analyzers
3. Configure CA1012 as a warning
4. Paste the following code in a .cs file:

```
    public abstract class Book
    {
        protected internal Book()
        {
        }
    }
``` 

#### Expected behavior
Warning, as this isn't allowed in the [Design Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/abstract-class):

> DO NOT define public or protected internal constructors in abstract types.

#### Actual behavior
No warning.

Evangelink:
Fix #3358


mavasani:
I believe this rule description indicates it is not intended to flag protected internal ctors: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1012?view=vs-2019. A ctor might have been marked protected internal to prevent it being instantiated from outside the assembly. Even if we wanted to flag such a case and recommend other design approaches, this would not fall under the CA1012 bucket as per current rule guidelines. I am also concerned this can have side effects of making an internal API a public API by switching protected internal to protected.

Evangelink:
Makes sense! @mavasani let me know if you think you will discuss this quickly and so it's better to keep the PR open or if you think I can close and we will handle it later.

mavasani:
@Evangelink - I am not sure. Personally, I don't see clarity on design guidelines why it suggests `protected internal` as not recommended. Is the alternative to use `private protected`? Any case, I think we should not change CA1012 and introduce a breaking change in the rule for this case. However, that is my personal opinion.

PR #3364, opened 2020-03-10
Update CA1806 to report on linq methods

### LINKED ISSUE #3363: Prevent mistakes by warning when LINQ method return value is not used, e.g. `list.Append(item);`
I think there are similar analyzers that warn if you don't use the return value of methods on immutable collections and Roslyn's immutable objects.

Something like this?

```cs
var list = new List<int>();

// ⚠ This method returns a new sequence which is not used. Did you mean 'list.Add'?
//   ↓↓↓↓↓↓↓↓↓↓
list.Append(42);
```

This should also apply to all LINQ methods known to be pure, like `Enumerable.Where`.

Evangelink:
Fix #3363


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -116,6 +117,15 @@ public sealed class DoNotIgnoreMethodResultsAnalyzer : DiagnosticAnalyzer
                                                                              isPortedFxCopRule: true,
                                                                              isDataflowRule: false);
 
+        internal static DiagnosticDescriptor LinqMethodRule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                             s_localizableTitle,
+                                                                             s_localizableMessageLinqMethod,
+                                                                             DiagnosticCategory.Performance,
+                                                                             RuleLevel.IdeSuggestion,
+                                                                             description: s_localizableDescription,
+                                                                             isPortedFxCopRule: false,
```
Evangelink: I have set `false` as this is a new behavior but the rule ID is reused so maybe we should stick to `true`.

mavasani: I'd still stick with `true` here and retain consistency for the flag for a given rule ID, as this flag is used to generate a custom tag, which is then used in some post-build tools to generate documentation and such.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1370,6 +1370,9 @@
   <data name="EnumShouldNotHaveDuplicatedValuesTitle" xml:space="preserve">
     <value>Enums values should not be duplicated</value>
   </data>
+  <data name="DoNotIgnoreMethodResultsMessageLinqMethod" xml:space="preserve">
+    <value>{0} calls {1} but does not use the value the method returns. Linq methods are known to not have side effects. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</value>
```
mavasani, line 1374:
```suggestion
    <value>'{0}' calls '{1}' but does not use the value the method returns. Linq methods are known to not have side effects. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</value>
```

mavasani: At some point it might be worthwhile to write an analyzer/fixer that catches and auto-fixes these - the resx file would be expected to be passed as an additional file...

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -176,6 +187,10 @@ public override void Initialize(AnalysisContext analysisContext)
                                 {
                                     rule = PureMethodRule;
                                 }
+                                else if (targetMethod.ContainingType.Equals(enumerableType))
```
mavasani, line 190: There does not seem to be anything specific to linq methods in the implementation. Should we just update the rule name, message, etc. to use the term `Enumerable` instead of `Linq`?

mavasani: Ah, I missed `enumerableType` here is actually `SystemLinqEnumerable`. Can the local name be changed to `linqEnumerableType`?

PR #3366, opened 2020-03-11
CA1045: Do not pass types by reference

### LINKED ISSUE #391: Port FxCop rule CA1045: DoNotPassTypesByReference
**Title:** Do not pass types by reference

**Description:**

Passing types by reference (using out or ref) requires experience with pointers, understanding how value types and reference types differ, and handling methods that have multiple return values. Library architects who design for a general audience should not expect users to master working with out or ref parameters.

**Dependency:** None

**Notes:**

Evangelink:
Fix #391


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
```diff
@@ -0,0 +1,70 @@
...
+                    var methodSymbol = (IMethodSymbol)context.Symbol;
+
+                    // FxCop compat: only analyze externally visible symbols by default.
+                    if (!methodSymbol.MatchesConfiguredVisibility(context.Options, Rule, context.CancellationToken))
+                    {
+                        return;
+                    }
+
+                    if (methodSymbol.IsOverride ||
+                        methodSymbol.IsImplementationOfAnyInterfaceMember())
```
Evangelink: I thought that was better to have a method callback rather than a parameter one to avoid doing multiple time this check.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
```diff
@@ -0,0 +1,70 @@
...
+                    {
+                        return;
+                    }
+
+                    foreach (var parameterSymbol in methodSymbol.Parameters)
+                    {
+                        // VB.NET out is a ref parameter with the OutAttribute
+                        if (parameterSymbol.RefKind == RefKind.Ref &&
```
Evangelink: We have used this pattern somewhere else in the code, do you want me to extract it out as an helper?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1370,4 +1370,13 @@
   <data name="EnumShouldNotHaveDuplicatedValuesTitle" xml:space="preserve">
     <value>Enums values should not be duplicated</value>
   </data>
+  <data name="DoNotPassTypesByReferenceMessage" xml:space="preserve">
+    <value>Consider a design that does not require that '{0}' be a reference parameter.</value>
```
Evangelink: Message is copied from legacy fxcop.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1370,4 +1370,13 @@
...
+  </data>
+  <data name="DoNotPassTypesByReferenceTitle" xml:space="preserve">
+    <value>Do not pass types by reference</value>
+  </data>
+  <data name="DoNotPassTypesByReferenceDescription" xml:space="preserve">
+    <value>Passing types by reference (using out or ref) requires experience with pointers, understanding how value types and reference types differ, and handling methods that have multiple return values. Also, the difference between out and ref parameters is not widely understood.</value>
```
Evangelink: First paragraph of the web page.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReferenceTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        public async Task CA1045_PInvokeMethod_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+using System.Runtime.InteropServices;
+
+public class Class1
+{
+    [DllImport(""Advapi32.dll"", CharSet=CharSet.Auto)]
+    public static extern Boolean FileEncryptionStatus(String filename, ref UInt32 status);
```
Evangelink: I would have expected this case not to raise but it is raising with legacy fxcop so I kept the same behavior.

Note that usually the pinvoke are private and so won't be subject to the diagnostic with the default visibility.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
```diff
@@ -0,0 +1,70 @@
...
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotPassTypesByReferenceTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotPassTypesByReferenceMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotPassTypesByReferenceDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Design,
+            RuleLevel.IdeSuggestion,
```
Evangelink: What shall be the default level?

stephentoub: From my perspective, Disabled :-)  It's nice that we're aiming to port all of the old FxCop rules so that folks that still want them can use them.  But this rule in particular feels very outdated, and I expect it will drive the wrong behaviors.  For example, my understanding is this would warn any time someone tries to write a method that uses the Try pattern.

mavasani: I agree - most of the ported FxCop rules should be disabled by default, so users can turn them on explicitly if they desire. Only a handful of these would be on by default suggestions, if they are indeed helpful code improvements, such that analyzer has no noise.

mavasani: NOTE: None of the ported FxCop rules should be enabled warnings by default - I am going to write a special analyzer for this repo which helps enforce that only design team approved analyzers are marked as enabled warnings.

Evangelink: @stephentoub This rule only looks at `ref` not `out` parameters so the Try pattern is not impacted here. This is the rule `AvoidOutParameters` that takes care of it.

stephentoub: Ok.  Then lazy initialization methods :)

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReferenceTests.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        private static DiagnosticResult GetCA1045CSharpResultAt(int line, int column, string parameterName)
+            => VerifyCS.Diagnostic()
+                .WithLocation(line, column)
+                .WithArguments(parameterName);
+
+        private static DiagnosticResult GetCA1045BasicResultAt(int line, int column, string parameterName)
+            => VerifyVB.Diagnostic()
+                .WithLocation(line, column)
+                .WithArguments(parameterName);
```
sharwell: 💡 Consider using markup syntax for all tests, except for a single test which verifies the message includes the correct parameter name.

sharwell: 💡 Use `WithSpan` instead of `WithLocation` for these helpers, so the full diagnostic location is clear in the tests.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
sharwell:
```suggestion
    <value>Passing types by reference requires experience with pointers, understanding how value types and reference types differ, and handling methods that have multiple return values. Also, the difference between 'out', 'ref', and 'in' parameters is not widely understood.</value>
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReferenceTests.cs
```diff
@@ -0,0 +1,171 @@
...
+    Implements Interface1
+
+    Public Sub Method1(ByRef s As String) Implements Interface1.Method1 ' ... but not here
+    End Sub
+End Class",
+                GetCA1045BasicResultAt(3, 23, "s"));
+        }
+
+        [Fact]
+        public async Task CA1045_VBNET_ByRefOutParameter_NoDiagnostic()
```
sharwell: 💡 Consider updating this test to also cover `out` parameters for C#

sharwell: 💡 Consider updating this test to use methods with one `ref` parameter and one `out` parameter, so it's clear that the distinguishing characteristic of the `out` parameter is the use of `out`, as opposed to an unintentional situation like limited visibility of the method causing the analyzer to ignore it.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReferenceTests.cs
```diff
@@ -0,0 +1,171 @@
...
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.DoNotPassTypesByReference,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.DoNotPassTypesByReference,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.UnitTests
+{
+    public class DoNotPassTypesByReferenceTests
```
sharwell: 💡 Consider adding tests covering `in` parameters (C# only)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotPassTypesByReferenceTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotPassTypesByReferenceMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotPassTypesByReferenceDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Design,
```
sharwell: @mavasani Can we place this rule in a new category `LegacyDesign` to indicate that it is no longer a recommended design rule? I worry the rate of false positives would exceed the benefit provided by the rule.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReferenceTests.cs
Evangelink: @sharwell is this recommendation targeted to this PR or a more general advice on this repo?

sharwell: More a suggestion for new tests.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
Evangelink: I have added this new category and applied it for this rule. Feel free to revert the commit if you don't want it.

mavasani:

src/Utilities/Compiler/DiagnosticCategory.cs
```diff
@@ -5,6 +5,7 @@ namespace Analyzer.Utilities
     internal static class DiagnosticCategory
     {
         public const string Design = nameof(Design);
+        public const string LegacyDesign = nameof(LegacyDesign);
```
mavasani: Please revert - we should be just using "Design" here.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
sharwell: I discussed this with @mavasani, and concluded that the category-based configuration for rules will not apply automatically to rules that are disabled by default, which makes it less likely that this rule will be accidentally enabled. This behavior was already covered by the following test case, but @mavasani is updating the documentation for this test case to call out its importance for analyzers like this so we don't unintentionally change that behavior in the future.

https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Test/CommandLine/CommandLineTests.cs#L11603-L11608

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPassTypesByReference.cs
```diff
@@ -0,0 +1,70 @@
...
+            context.RegisterCompilationStartAction(context =>
+            {
+                var outAttributeType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesOutAttribute);
+
+                context.RegisterSymbolAction(context =>
+                {
+                    var methodSymbol = (IMethodSymbol)context.Symbol;
+
+                    // FxCop compat: only analyze externally visible symbols by default.
+                    if (!methodSymbol.MatchesConfiguredVisibility(context.Options, Rule, context.CancellationToken))
```
mavasani: Need to update https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#analyzed-api-surface, add a test for `api_surface` option and also file an issue on docs repo to mention about configurability: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1045?view=vs-2019

Evangelink: Done!

PR #3183, opened 2020-01-09
Review diagnostic messages for missing single quotes around symbol names

Evangelink:
Fix #1714 

I did a first review of all the diagnostic messages to find missing quotes. I have deliberately skipped all description messages (let me know if you want me to update them).

I have, for now, only changed the resource files and when we agree upon the texts I'll update the various translation files.

Doing this review of the messages, I have a couple of remarks to make:

- There is no consistency on whether title and messages shall end (or not) with a period.

- Some of the messages are using a pascal case rule for all the words of the message, this probably reads weird when we happen to have the issue raised.

- There is no consistency on how attributes are referred to. We can find `Use [XXX]`, `Use [XXXAttribute]`, `Add XXX attribute` or `Add XXXAttribute`. It might be good to always use the same style.

Issue #1714, opened 2018-06-06
Audit diagnostic messages for missing quotes around symbol names


mavasani:
I think quite a few of our diagnostic messages were created by an automated tool that ported messages from old FxCop implementation. For example, see https://github.com/dotnet/roslyn-analyzers/issues/1713#issuecomment-395051433.

Evangelink:
@mavasani Do you have some guidelines around when the "thing" should be single quoted? Is it like every time this is a keyword or a type/method/property/... name?

Evangelink:
ping @mavasani

mavasani:
Yes - I think at the very least all symbol names in messages should be within single quotes for clarity.

Evangelink:
To sum-up some of the exchanges on the 1st draft PR:

- Prefer `SomeAttribute` for diagnostics, and `[Some]` for code fixes.

- Prefer no '.' for single-sentence titles and messages (and avoid multi-sentence unless it substantially detracts from clarity)

- Use full sentences with '.' for descriptions

- Do not use pascal case for all words

PR #3368, opened 2020-03-13
Add DoNotUseReferenceEqualsWithValueTypesAnalyzer

bartonjs:
This analyzer detects when an argument into object.ReferenceEquals
is a value type that gets boxed as part of the argument.

Contributes to https://github.com/dotnet/runtime/issues/30740
cc @dotnet/dotnet-analyzers


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+        internal const string RuleId = "CA2013";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotUseReferenceEqualsWithValueTypesTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotUseReferenceEqualsWithValueTypesMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotUseReferenceEqualsWithValueTypesDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.MicrosoftCodeAnalysisCorrectness,
```
mavasani: This category is actually meant for analyzers for correct usage of APIs from Microsoft.CodeAnalysis NuGet package. See https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt. We also have an analyzer in the repo that would have flagged this category mismatch and enforces the ID range to category mapping, but that analyzer recently broke - let me resurrect that.

Meanwhile, you may want to follow up with @terrajobst on the category you guys would like to use. You can either re-use one of the FxCop categories from that file and pick the next available ID from that categories allocated ID range OR define a new category. If we decide to go with IDs that are not following the 2 letter prefix followed by an interger, then I can remove this ID-range mapping validation for the new category.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -930,6 +930,15 @@
...
     <value>Call to obsolete key derivation function {0}.{1}</value>
   </data>
+  <data name="DoNotUseReferenceEqualsWithValueTypesDescription" xml:space="preserve">
+    <value>Value type typed arguments are uniquely boxed for each call to this method, therefore the result is always false.</value>
+  </data>
+  <data name="DoNotUseReferenceEqualsWithValueTypesMessage" xml:space="preserve">
+    <value>The argument for the '{0}' parameter to ReferenceEquals evaluates to the value type '{1}'. Due to value boxing, this call to ReferenceEquals will always return false.</value>
+  </data>
+  <data name="DoNotUseReferenceEqualsWithValueTypesTitle" xml:space="preserve">
+    <value>Do not use ReferenceEquals with value types</value>
```
stephentoub: All the other messages end with a period.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// CA1307: Specify StringComparison
```
stephentoub: Copy/paste error

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotUseReferenceEqualsWithValueTypesTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotUseReferenceEqualsWithValueTypesMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotUseReferenceEqualsWithValueTypesDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.MicrosoftCodeAnalysisCorrectness,
+            RuleLevel.BuildWarning,
```
mavasani: This is going to be the first enabled by default warning from this package. I am going to write an analyzer that enforces that every enabled by default build warning is tracked explicitly and has an associated github issue with the approval tag.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+                                    argument.Parameter.Name,
+                                    val.Type.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat)));
+                        }
+                    }
+                },
+                OperationKind.Invocation);
+            });
+        }
+
+        private static ParameterInfo GetParameterInfo(INamedTypeSymbol type, bool isArray = false, int arrayRank = 0, bool isParams = false)
```
mavasani: Consider making the last three parameters of `ParameterInfo.GetParameterInfo` optional and avoiding this private static method.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+                    if (!referenceEqualsMethod.Equals(targetMethod))
+                    {
+                        return;
+                    }
+
+                    foreach (var argument in invocationExpression.Arguments)
+                    {
+                        var val = argument.Value;
+
+                        if (val is IConversionOperation conversion)
```
mavasani: To handle multiple conversions - for example an implicit conversion to a value type on top of a user defined conversion.
```suggestion
                        val = val.WalkDownConversion();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+                        {
+                            val = conversion.Operand;
+                        }
+
+                        if (val.Type.IsValueType)
```
mavasani: I believe argument type can be null here, for example `ReferenceEquals(null, a);` Please add a test as well.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// CA1307: Specify StringComparison
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotUseReferenceEqualsWithValueTypesAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2013";
```
mavasani: `CA2013` maps to `DiagnosticCategory.Reliability` from https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt, and is indeed the next available ID in the range for that category. I think it might be fine to just change the category in the descriptor to Reliability for now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -930,6 +930,15 @@
   <data name="DoNotUseObsoleteKDFAlgorithmMessage" xml:space="preserve">
     <value>Call to obsolete key derivation function {0}.{1}</value>
   </data>
+  <data name="DoNotUseReferenceEqualsWithValueTypesDescription" xml:space="preserve">
+    <value>Value type typed arguments are uniquely boxed for each call to this method, therefore the result is always false.</value>
+  </data>
+  <data name="DoNotUseReferenceEqualsWithValueTypesMessage" xml:space="preserve">
+    <value>The argument for the '{0}' parameter to ReferenceEquals evaluates to the value type '{1}'. Due to value boxing, this call to ReferenceEquals will always return false.</value>
```
mavasani: Wondering if the message should have the user actionable part at the beginning. Also wondering if the parameter name in the message is actually helpful? Maybe we can avoid it to simplify the message.

```suggestion
    <value>Do not pass an argument with value type '{0}' to 'ReferenceEquals'. Due to value boxing, this call to 'ReferenceEquals' will always return 'false'.</value>
```

I would be fine to even go a step further and simplify to `Do not pass a value type argument to 'ReferenceEquals'. Due to value boxing, this call to 'ReferenceEquals' will always return 'false'.`. I'll let you make a call on the preferred message.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Yeah we are very inconsistent in whether or not periods are used in our title/message/description strings. I think we should follow @sharwell's proposal from https://github.com/dotnet/roslyn-analyzers/pull/3183#issuecomment-576724513, which is summarized by @Evangelink in https://github.com/dotnet/roslyn-analyzers/issues/1714#issuecomment-580647362

mavasani: Another place where an analyzer/fixer to enforce this would be a much better route. @Evangelink given your initial interest in tackling #1714, would you be interested in writing this analyzer/fixer? If so, we can sync up offline on the design.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
mavasani: https://github.com/dotnet/roslyn-analyzers/pull/3370 - @bartonjs once that PR goes in, I'll have a follow-up PR to move the repo to consume the new analyzer package with that enhancement. At that point, you can pull in the latest sources into your branch and confirm that you get an RSxxxx warning here that your chosen rule ID does not fall in the allowed range ID for the provided rule category.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
bartonjs: Really I think the question here is just "is the last step a boxing conversion". `ReferenceEquals((SomeCacheType)5, other)` is sensible if SomeCacheType is a reference type and the conversion operator was doing a dictionary lookup by the integer ID; so if WalkDownConversion is going to go all the way to the `5` then it seems like the wrong approach.

mavasani: Sounds good!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
bartonjs: My scan showed that things seemed to be title: no period, other: yes period.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypesTests.cs
```diff
@@ -0,0 +1,246 @@
...
+{
+    class TestClass
+    {
+        private static bool TestMethod(string test)
+        {
+            return ReferenceEquals(IntPtr.Zero, test);
+        }
+    }
+}",
+                GetCSharpResultAt(10, 36, "objA", "System.IntPtr"));
```
mavasani: I believe the test framework also supports annotating the source such as `return ReferenceEquals([|IntPtr.Zero|], test);`, so you can avoid passing in this explicit expected diagnostic and the framework will decode the annotation and validate the diagnostic. I am fine with your current form as well, if that is by preference.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
bartonjs: I feel like having the type name in the message is helpful, especially for non-IDE experiences.

The line and character positions do help solve the "which argument" problem, so I can see leaving that one out.

So I think I'm leaning toward your first suggestion ("Do not pass an argument with value type '{0}' to 'ReferenceEquals'. Due to value boxing, this call to 'ReferenceEquals' will always return 'false'.")

Joe4evr:
In a related vein, should there also be an analyzer warning when passing the newly added `ReferenceEqualityComparer` to something that is keyed by a value type?

mavasani:
> In a related vein, should there also be an analyzer warning when passing the newly added ReferenceEqualityComparer to something that is keyed by a value type?

Seems reasonable. Should we just fold that case also in the analyzer you are adding, and just use a different message, but same ID? If so, you can create an additional descriptor in the analyzer with same arguments, but just a different messageFormat string, and it should work.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
Evangelink: @mavasani Yep I'd be happy to help, let's have a chat later on.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
```diff
@@ -0,0 +1,95 @@
...
+                        {
+                            val = conversion.Operand;
+                        }
+
+                        if (val.Type?.IsValueType == true)
+                        {
+                            operationContext.ReportDiagnostic(
+                                val.CreateDiagnostic(
+                                    Rule,
+                                    val.Type.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat)));
```
mavasani: We cannot use this error message format because the analyzer is supported for both VB and C#. Generally we just use the simple `type.Name` in messages.

src/Microsoft.NetCore.Analyzers/Microsoft.NetCore.Analyzers.sarif
```diff
@@ -344,6 +344,25 @@
             ]
           }
         },
+        "CA2013": {
+          "id": "CA2013",
+          "shortDescription": "Do not use ReferenceEquals with value types",
+          "fullDescription": "Value type typed arguments are uniquely boxed for each call to this method, therefore the result is always false.",
```
mavasani: Did you edit this file manually? If so, it needs to be reverted. This is auto-generated file that you can update locally by ruuning `msbuild /v:m /t:pack nuget\NetAnalyzers\Microsoft.CodeAnalysis.NetAnalyzers.Package.csproj` or `eng\common\cibuild.cmd` from the root of the repo. It is fine if these files are not updated as a part of your PR as well, we normally re-generate this periodically.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypes.cs
bartonjs: I was concerned about that, but a lot of analyzers seem to use that format, none use the VB format, and the VB tests say that it says "Integer" instead of "int", so whatever the printer is it seems to understand which language applies.

bartonjs:
I added the simple detection for ReferenceEqualityComparer's Equals method (I didn't do dataflow to find out if an `IEqualityComparer` or `IEqualityComparer<object>` was provably a ReferenceEqualityComparer).

I'm not sure how to add tests that involve types that were recently introduced... is there a best practice?  (I did manually verify it with a project in dotnet/runtime that can see ReferenceEqualityComparer)

mavasani:
> I'm not sure how to add tests that involve types that were recently introduced... is there a best practice? (I did manually verify it with a project in dotnet/runtime that can see ReferenceEqualityComparer)

@sharwell?

sharwell:
@bartonjs The test would need to set the reference assemblies as appropriate for the API in question. Eventually this will be simple, such as the following:

```csharp
await new VerifyCS.Test
{
  ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp20, // netcoreapp2.0
}.RunAsync();
```

Can you clarify exactly which version you want to test against? We can use that information to construct a custom `ReferenceAssemblies` instance for the test in question.

bartonjs:
> Can you clarify exactly which version you want to test against?

.NET 5, Feb14 or newer :smile:.

sharwell:
Here's the definition for **netcoreapp2.1**:

https://github.com/dotnet/roslyn-sdk/blob/18d931b1e05151ffd189a362ce86b96f01f08c73/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs#L760-L762

I'm guessing the new TFM is **netcoreapp5.0**? Assuming the reference assembly package for this TFM is still **Microsoft.NETCore.App** and the package still has the same internal structure as **netcoreapp2.1**, there are two changes that are needed:

1. We need to know the package version corresponding to "Feb14 or newer".
2. We need to know the NuGet feed containing the package. If it's not nuget.org, then we'll also need to update the `ResolveAsync` implementation to support custom feeds.

bartonjs:
I think we're in the process of prepping preview 1, so it's probably sensible to just wait for that to finish; but it doesn't look like 3.0/3.1/5.0 use the Microsoft.NETCore.App package anymore, it's just intrinsically there, or not (which may be part of why no NetCoreApp30 or NetCoreApp31 test target got defined?)

sharwell:
Perhaps https://www.nuget.org/packages/Microsoft.NETCore.App.Ref/5.0.0-preview.1.20120.5?

bartonjs:
@sharwell  Yeah, pretty much that one. ::facepalm::

sharwell:
@bartonjs You'll be able to target this as soon as dotnet/roslyn-sdk#388 is available (it specifically needs the change to `ResolveAsync`), or you can wait for dotnet/roslyn-sdk#490 and not even need to define a custom `ReferenceAssemblies` instance.

sharwell:
@bartonjs I just updated the SDK for dotnet/roslyn-analyzers. You can now use `ReferenceAssemblies.NetCore.NetCoreApp50` for your test.

src/Test.Utilities/CSharpCodeFixVerifier`2.cs
```diff
@@ -51,5 +51,20 @@ public static async Task VerifyCodeFixAsync(string source, DiagnosticResult[] ex
             test.ExpectedDiagnostics.AddRange(expected);
             await test.RunAsync();
         }
+
+        public static async Task VerifyAnalyzerAsync(
```
sharwell: 📝 New helper methods should not be added to this class. I will remove this in a follow-up PR.

bartonjs:
@mavasani Would a fixer that replaces the triggering expressions with `false` be a reasonable fixer?  For newly-written code the error is likely "you meant to pass a different argument", so I don't know that the fixer would really be "helpful"... but the half-dozen violations in dotnet/runtime were all several years old and I'm just doing dead-code deletion.

mavasani:
Yeah, I am not sure there is a good codefix that would be helpful here.

PR #3370, opened 2020-03-13
Enhance DiagnosticDescriptorCreationAnalyzer to support analysis of h…

mavasani:
…elper methods for creating diagnostic descriptors

This ensures that the analyzers in Microsoft.CodeAnalysis.NetAnalyzers which use a helper method for creating descriptors get all the RS diagnostics for correct descriptor usage: RS1017 (DiagnosticIdMustBeAConstant), RS1019 (UseUniqueDiagnosticId) RS1018 (DiagnosticIdMustBeInSpecifiedFormat) and RS1020 (UseCategoriesFromSpecifiedRange)

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
@@ -207,32 +207,25 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
                 var idToAnalyzerMap = new ConcurrentDictionary<string, ConcurrentDictionary<string, ConcurrentBag<Location>>>();
                 compilationContext.RegisterOperationAction(operationAnalysisContext =>
                 {
-                    if (!(((IFieldInitializerOperation)operationAnalysisContext.Operation).Value is IObjectCreationOperation objectCreation))
+                    var fieldInitializer = (IFieldInitializerOperation)operationAnalysisContext.Operation;
+                    if (!TryGetDescriptorCreateMethodAndArguments(fieldInitializer, diagnosticDescriptorType, out var creationMethod, out var creationArguments))
                     {
                         return;
                     }
 
-                    var ctor = objectCreation.Constructor;
-                    if (ctor == null ||
-                        !diagnosticDescriptorType.Equals(ctor.ContainingType) ||
-                        !diagnosticDescriptorType.InstanceConstructors.Any(c => c.Equals(ctor)))
-                    {
-                        return;
-                    }
-
-                    AnalyzeTitle(operationAnalysisContext, objectCreation);
-                    AnalyzeHelpLinkUri(operationAnalysisContext, objectCreation);
-                    AnalyzeCustomTags(operationAnalysisContext, objectCreation);
+                    AnalyzeTitle(creationMethod, fieldInitializer, operationAnalysisContext.ReportDiagnostic);
+                    AnalyzeHelpLinkUri(operationAnalysisContext, creationArguments);
+                    AnalyzeCustomTags(operationAnalysisContext, creationArguments);
 
                     string? categoryOpt = null;
                     if (!checkCategoryAndAllowedIds ||
-                        !TryAnalyzeCategory(operationAnalysisContext, objectCreation,
+                        !TryAnalyzeCategory(operationAnalysisContext, creationArguments,
                             additionalTextOpt!, categoryAndAllowedIdsMap!, out categoryOpt, out var allowedIdsInfoList))
                     {
                         allowedIdsInfoList = default;
                     }
 
-                    AnalyzeRuleId(operationAnalysisContext, objectCreation, additionalTextOpt,
+                    AnalyzeRuleId(operationAnalysisContext, creationArguments, additionalTextOpt,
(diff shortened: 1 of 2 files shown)


mavasani:
Ping @genlu @sharwell @dotpaul for reviews.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
```diff
@@ -291,6 +291,65 @@ public override void Initialize(AnalysisContext context)
...
+            return ImmutableArray.Create(descriptor);
+        }
+    }
+
+    public override void Initialize(AnalysisContext context)
+    {
+    }
+}" + CSharpDiagnosticDescriptorCreationHelper,
+                GetCSharpRS1017ExpectedDiagnostic(14, 43, "descriptor"),
+                GetCSharpRS1019ExpectedDiagnostic(38, 43, "DuplicateDiagnosticId", "MyAnalyzer"));
```
sharwell: 💡 Consider using markup syntax so it's easier to see the full span where these are getting reported.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -276,28 +269,57 @@ public override void Initialize(AnalysisContext context)
             });
         }
 
-        private static void AnalyzeTitle(OperationAnalysisContext operationAnalysisContext, IObjectCreationOperation objectCreation)
+        private static bool TryGetDescriptorCreateMethodAndArguments(
+            IFieldInitializerOperation fieldInitializer,
+            INamedTypeSymbol diagnosticDescriptorType,
+            out IMethodSymbol creationMethod,
```
sharwell:
```suggestion
            [NotNullWhen(true)] out IMethodSymbol? creationMethod,
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -207,32 +207,25 @@ public override void Initialize(AnalysisContext context)
                 var idToAnalyzerMap = new ConcurrentDictionary<string, ConcurrentDictionary<string, ConcurrentBag<Location>>>();
                 compilationContext.RegisterOperationAction(operationAnalysisContext =>
                 {
-                    if (!(((IFieldInitializerOperation)operationAnalysisContext.Operation).Value is IObjectCreationOperation objectCreation))
+                    var fieldInitializer = (IFieldInitializerOperation)operationAnalysisContext.Operation;
```
sharwell: 💭 Does this work for properties with this form? In theory the backing field initialization could be covered by one of these operations, but I'm not sure if it actually is.

```csharp
static DiagnosticDescriptor Rule { get; } = new DiagnosticDescriptor(...);
```

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
mavasani: I have kept it consistent with the existing tests in this file. We can clean up the entire file in a follow-up PR.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: This file is not yet nullable enabled.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: Nope, that wasn't supported earlier either. I can file a separate feature request to support that case.

mavasani:
Thanks!

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
sharwell: I thought the whole repository was?

mavasani: Oh yes, sorry. Switching between Roslyn and RoslynAnalyzers repo confused me. Let me fix this in a follow-up PR.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
sharwell: Can you file this issue? I believe it's a common pattern for some third-party analyzers.

PR #3372, opened 2020-03-16
Implement RS0038 (Prefer null literal)

sharwell:
Prefer `null` to `default(T)` (or `default`) when `T` is a reference type.

Additional cases that need to be accounted for:

* [x] Use `(T)null` where necessary to avoid overload resolution errors
* [x] ~~Use `nameof(T.Member)` instead of `nameof(default(T).Member)` in the code fix~~ Not valid C# syntax.

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/CSharp/PreferNullLiteral.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/CSharp/PreferNullLiteral.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class PreferNullLiteral : DiagnosticAnalyzer
+    {
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.PreferNullLiteralTitle), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.PreferNullLiteralMessage), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.PreferNullLiteralDescription), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            RoslynDiagnosticIds.PreferNullLiteralRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.RoslynDiagnosticsMaintainability,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: s_localizableDescription,
+            helpLinkUri: null,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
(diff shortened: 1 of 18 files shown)


stephentoub:
How are we thinking about whether a rule like this is put into the Roslyn-specific library?  Seems it'd be  generally useful (as opt-in) for many .NET projects, no?

Evangelink:
@sharwell I am curious about the reasoning for this rule, is it only styling preference or is there any kind of impact (performance or anything else)?

sharwell:
@stephentoub The rule could be moved to a more widely available package in the future. I am implementing it right now to simplify my code reviews in dotnet/roslyn, so I didn't want to block it on that discussion.

@Evangelink It's a readability code style rule.

src/Roslyn.Diagnostics.Analyzers/CSharp/PreferNullLiteral.cs
```diff
@@ -0,0 +1,58 @@
...
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.PreferNullLiteralMessage), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.PreferNullLiteralDescription), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            RoslynDiagnosticIds.PreferNullLiteralRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.RoslynDiagnosticsMaintainability,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
```
mavasani: I believe this rule should be disabled by default, and only enabled explicitly if required.

sharwell: This rule is already disabled by default on the grounds that it is not in a standard package.

mavasani: I think this package is used by other internal repos: roslyn-analyzers, possibly project-system too. Not a strong pushback here though...

src/Roslyn.Diagnostics.Analyzers/UnitTests/PreferNullLiteralTests.cs
```diff
@@ -0,0 +1,264 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
```
mavasani: Can you please add at least one sanity test for VB?

src/Roslyn.Diagnostics.Analyzers/CSharp/PreferNullLiteralCodeFixProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Simplification;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic, Name = nameof(PreferNullLiteralCodeFixProvider))]
```
mavasani: Ah, seems like we actually only support C# for the analyzer - can you remove VB from this list for the fixer then?

src/Roslyn.Diagnostics.Analyzers/UnitTests/PreferNullLiteralTests.cs
sharwell: ➡️ Covered by https://github.com/dotnet/roslyn-analyzers/pull/3372#discussion_r393687493

src/Roslyn.Diagnostics.Analyzers/CSharp/PreferNullLiteral.cs
sharwell: It is, but that's always come with the caveat that the rules were originally intended for Roslyn so they may need to be reconfigured for use outside of that.

src/Roslyn.Diagnostics.Analyzers/CSharp/PreferNullLiteralCodeFixProvider.cs
sharwell: Fixed

PR #3386, opened 2020-03-18
PublicAPI: tag APIs with oblivious types

jcouv:
PublicAPI has the ability to track nullable (`?`) and non-nullable (`!`) reference types when the public API file starts with `#nullable enable` (PR https://github.com/dotnet/roslyn-analyzers/pull/3125).
But this is insufficient to track progress on annotating a project, because of value types.

The solution is to flag APIs that use oblivious reference types with a `~` prefix.

In an un-annotated file, you'll just have plain APIs, such as `C.Field -> string`.
In an annotated file (`#nullable enable`), you can have the following:
1. `C.Field -> string` (for folks who started using `#nullable enable` already)
2. `~C.Field -> string` (new/preferred format)
3. `C.Field -> string?`
4. `C.Field -> string!`

Note that we offer to annotate entries like (1) to add the oblivious marker (like 2).


src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -83,6 +84,8 @@ internal ApiData(ImmutableArray<ApiLine> apiList, ImmutableArray<RemovedApiLine>
 
         private sealed class Impl
         {
+            private const char obliviousMarker = '~';
```
jaredpar: Should be pascal cased

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -189,6 +192,11 @@ private void OnSymbolActionCore(ISymbol symbol, Action<Diagnostic> reportDiagnos
                 }
             }
 
+            static string WithObliviousMarker(string name)
```
jaredpar:
```suggestion
            internal static string WithObliviousMarker(string name)
```

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -215,21 +224,38 @@ private void OnSymbolActionCore(ISymbol symbol, Action<Diagnostic> reportDiagnos
                 }
 
                 ApiLine foundApiLine;
+                bool symbolUsesOblivious = false;
                 if (_useNullability)
                 {
+                    symbolUsesOblivious  = UsesOblivious(symbol);
```
jaredpar:
```suggestion
                    symbolUsesOblivious = UsesOblivious(symbol);
```

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -503,11 +544,19 @@ private string GetSiblingNamesToRemoveFromUnshippedTextCore(ISymbol symbol)
                 return string.Empty;
             }
 
+            static bool UsesOblivious(ISymbol symbol)
```
jaredpar:
```suggestion
            internal static bool UsesOblivious(ISymbol symbol)
```

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -703,6 +754,84 @@ private static bool CanTypeBeExtendedPubliclyImpl(ITypeSymbol type)
                         m => m.DeclaredAccessibility != Accessibility.Internal && m.DeclaredAccessibility != Accessibility.Private && m.DeclaredAccessibility != Accessibility.ProtectedAndInternal
                     );
             }
+
+            /// <summary>
+            /// Various Visit* methods return true if an oblivious reference type is detected.
+            /// </summary>
+            private class ObliviousDetector : SymbolVisitor<bool>
```
jaredpar:
```suggestion
            private sealed class ObliviousDetector : SymbolVisitor<bool>
```

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: Will fix. I was relying on editor to guide me on that, but maybe editorconfig is not configured in this repo...

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -703,6 +754,84 @@ private static bool CanTypeBeExtendedPubliclyImpl(ITypeSymbol type)
...
+                        if (Visit(parameter.Type))
+                        {
+                            return true;
+                        }
+                    }
+
+                    return false;
+                }
+
+                public override bool VisitNamedType(INamedTypeSymbol symbol)
```
jaredpar: Where do type parameters which have no constraints get handled in this checker? Sorry, feel like I'm missing the obvious.

src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
```diff
@@ -85,6 +85,21 @@ private static SourceText AddSymbolNamesToSourceText(SourceText sourceText, IEnu
 
             SourceText newSourceText = sourceText.Replace(new TextSpan(0, sourceText.Length), string.Join(Environment.NewLine, lines) + GetEndOfFileText(sourceText));
             return newSourceText;
+
+            // Insert name at the first suitable position
+            static void InsertInList(List<string> list, string name)
```
jaredpar: Nit: other local functions in this PR use camel case

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: I didn't follow. It's okay to have unconstrained type parameters in your API (we shouldn't warn about that).

jaredpar: What if the unconsrtained type parameter is in an oblivious context

jaredpar: Basically this case:

```cs
#nullable enable
class ReallyMean<
#nullable disable
T> { }
```

Missing where we catch this case (or possibly it just doesnt' matter)

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -703,6 +754,84 @@ private static bool CanTypeBeExtendedPubliclyImpl(ITypeSymbol type)
...
+            {
+                private bool _isTypeDeclaration = true;
+
+                public override bool VisitField(IFieldSymbol symbol)
+                {
+                    _isTypeDeclaration = false;
+                    return Visit(symbol.Type);
+                }
+
+                public override bool VisitMethod(IMethodSymbol symbol)
```
sharwell: ❔ What about oblivious types in generic type constraints?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -703,6 +754,84 @@ private static bool CanTypeBeExtendedPubliclyImpl(ITypeSymbol type)
...
+                {
+                    _isTypeDeclaration = false;
```
sharwell: ❔ Why are we not restoring this on return? What about delegating to a different visitor instance for descent into these?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: Ah, I think that's probably one of the cases Jared had in mind (`where T : class~` vs. `where T : class!` or `where T : class?` kinda situations). I'll address

src/PublicApiAnalyzers/UnitTests/AnnotatePublicApiAnalyzerTests.cs
```diff
@@ -246,6 +246,45 @@ public class C
             await VerifyCSharpAdditionalFileFixAsync(source, shippedText, unshippedText, newShippedApiText: shippedText, fixedUnshippedText);
         }
 
+        [Fact]
+        public async Task TestAddAndRemoveMembers_CSharp_Fix_WithAddedNullability_WithoutOblivious()
```
sharwell: 💡 It would be good to have a test showing that the code fix annotates _shipped_ files in-place.

src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
```diff
@@ -145,6 +146,16 @@
   <data name="ShouldAnnotateApiFilesDescription" xml:space="preserve">
     <value>All public types and members should be declared with nullability annotations in PublicAPI.txt. This draws attention to API nullability changes in the code reviews and source control history, and helps prevent breaking changes.</value>
   </data>
+
+  <data name="ObliviousPublicApiTitle" xml:space="preserve">
+    <value>Public members should not use oblivious types</value>
+  </data>
+  <data name="ObliviousPublicApiMessage" xml:space="preserve">
+    <value>Symbol '{0}' uses some oblivious reference types.</value>
```
sharwell: We might find it helpful in the future to switch to this form:

```suggestion
    <value>Symbol '{0}' uses oblivious reference type '{1}'</value>
```

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: We can't put annotations in that position `class ReallyMean<T$$>`, be it `?`, `!` or `~` (in spirit).

src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
jcouv: I'd rather not. The tracking of which part or parts caused this warning would be tricky to do. Also, in practice, I'm not sure it's often ambiguous (either you put `#nullable enable` in the file or you didn't).

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
jcouv: Good idea. Thanks!

src/PublicApiAnalyzers/UnitTests/AnnotatePublicApiAnalyzerTests.cs
jcouv: Added

src/Utilities/Compiler/Lightup/ITypeSymbolExtensions.cs
```diff
@@ -11,11 +11,16 @@ internal static class ITypeSymbolExtensions
...
+            = LightupHelpers.CreateSymbolPropertyAccessor<ITypeParameterSymbol, NullableAnnotation>(typeof(ITypeParameterSymbol), nameof(ReferenceTypeConstraintNullableAnnotation), fallbackResult: Lightup.NullableAnnotation.None);
 
         public static NullableAnnotation NullableAnnotation(this ITypeSymbol typeSymbol)
             => s_nullableAnnotation(typeSymbol);
 
         public static ITypeSymbol WithNullableAnnotation(this ITypeSymbol typeSymbol, NullableAnnotation nullableAnnotation)
             => s_withNullableAnnotation(typeSymbol, nullableAnnotation);
+
+        public static NullableAnnotation ReferenceTypeConstraintNullableAnnotation(this ITypeParameterSymbol typeParameterSymbol)
+            => s_referenceTypeConstraintNullableAnnotation(typeParameterSymbol);
```
sharwell: 💡 Can you move this to `ITypeParameterSymbolExtensions` for the lightup layer?

jcouv: Moved to `ITypeParameterSymbolExtensions` and spotted/addressed more issues with detecting oblivious in type constraints. Thanks

PR #3387, opened 2020-03-18
CA1002: Do not expose generic lists

### LINKED ISSUE #369: Port FxCop rule CA1002: DoNotExposeGenericLists
**Title:** Do not expose generic lists

**Description:**

System.Collections.Generic.List&lt;(Of &lt;(T&gt;)&gt;) is a generic collection that is designed for performance, not inheritance. Therefore, List does not contain any virtual members. The generic collections that are designed for inheritance should be exposed instead.

**Dependency:** None

**Notes:**

Evangelink:
Fix #369


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotExposeGenericListsTests.cs
```diff
@@ -0,0 +1,274 @@
...
+        public async Task CA1002_ExtensionMethods()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System.Collections.Generic;
+
+namespace Namespace1
+{
+    public static class Helper
+    {
+        public static bool Ext1(this List<int> list) => true;
```
Evangelink: This exclusion is not part of the original implementation but that seems weird to get a penalty to extend `List<T>` so I added an exclusion. Please feel free to let me know if you think this shouldn't be done.

mavasani:

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotExposeGenericLists.cs
```diff
@@ -0,0 +1,89 @@
...
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            analysisContext.RegisterCompilationStartAction(context =>
+            {
+                var genericListType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericList1);
+
+                context.RegisterSymbolAction(context =>
+                {
+                    var symbol = context.Symbol;
+                    var methodSymbol = symbol as IMethodSymbol;
```
mavasani, line 45: Did you mean to register separate actions for each symbol kind? It seems better to move the shared code in a separate helper and register separate actions for each symbol kind that needs different checks.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotExposeGenericLists.cs
```diff
@@ -0,0 +1,89 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            analysisContext.RegisterCompilationStartAction(context =>
+            {
+                var genericListType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericList1);
```
mavasani, line 40:
```suggestion
                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericList1, out var genericListType))
                {
                   return;
                }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotExposeGenericLists.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                context.RegisterSymbolAction(context =>
+                {
+                    var symbol = context.Symbol;
+                    var methodSymbol = symbol as IMethodSymbol;
+
+                    // FxCop compat: only analyze externally visible symbols by default.
```
mavasani, line 47: Thanks!

mavasani: Please update https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#analyzed-api-surface and file a doc issue to mention api surface configurability for this rule. Same for other rules that you are supporting this option.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotExposeGenericLists.cs
Evangelink: Do you mean replacing 

```csharp
context.RegisterSymbolAction(ctx => {}, SymbolKind.Field, SymbolKind.Property, SymbolKind.Method);
``` 

with 

```csharp
context.RegisterSymbolAction(Analyze, SymbolKind.Field);
context.RegisterSymbolAction(Analyze, SymbolKind.Property);
context.RegisterSymbolAction(Analyze, SymbolKind.Method);

private static Analyze() {}
```

mavasani: I meant having separate `AnalyzeMethod`, `AnalyzeField`, etc. for each symbol kind, so you can directly cast to `IMethodSymbol` in `AnalyzeMethod` and have method specific logic only in `AnalyzeMethod`.

Evangelink: Ah ok, yeah makes sense I have overused the helpers :)

PR #3391, opened 2020-03-18
Add suppressor RS0039 (Relax test naming)

sharwell:
Builds on #3390

Asynchronous test methods do not require the 'Async' suffix.


src/Roslyn.Diagnostics.Analyzers/Core/RelaxTestNamingSuppressor.cs
```diff
@@ -0,0 +1,55 @@
...
+        private static readonly LocalizableString s_localizableJustification = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.RelaxTestNamingSuppressorJustification), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+
+        internal static readonly SuppressionDescriptor Rule =
+            new SuppressionDescriptor(Id, SuppressedDiagnosticId, s_localizableJustification);
+
+        public override ImmutableArray<SuppressionDescriptor> SupportedSuppressions { get; } = ImmutableArray.Create(Rule);
+
+        public override void ReportSuppressions(SuppressionAnalysisContext context)
+        {
+            if (!(context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitFactAttribute) is { } factAttribute))
```
Evangelink: I am probably not using the right keywords but I can't find the explanation of the `XXX is {} name` syntax, could you enlighten me?

sharwell: It's like `var name`, except `{ } name` is only true when `name` is non-null.

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs
```diff
@@ -42,5 +42,6 @@ internal static class RoslynDiagnosticIds
         public const string AnnotatePublicApiRuleId = "RS0036";
         public const string ShouldAnnotateApiFilesRuleId = "RS0037";
         public const string PreferNullLiteralRuleId = "RS0038";
+        public const string RelaxTestNamingSuppressionRuleId = "RS0039";
```
mavasani: Should we use a different prefix for suppression IDs?

src/Roslyn.Diagnostics.Analyzers/Core/RelaxTestNamingSuppressor.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Roslyn.Diagnostics.Analyzers.Extensions;
+
+namespace Roslyn.Diagnostics.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class RelaxTestNamingSuppressor : DiagnosticSuppressor
+    {
+        private const string Id = RoslynDiagnosticIds.RelaxTestNamingSuppressionRuleId;
+        private const string SuppressedDiagnosticId = "VSTHRD200";
```
mavasani: Doc comment on the ID with pointer to its documentation?

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/INamedTypeSymbolExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis;
+
+namespace Roslyn.Diagnostics.Analyzers.Extensions
+{
+    internal static class INamedTypeSymbolExtensions
+    {
+        public static bool IsTestAttribute(this INamedTypeSymbol attributeClass, ConcurrentDictionary<INamedTypeSymbol, bool> knownTestAttributes, INamedTypeSymbol factAttribute)
```
mavasani: Should we extract the test attribute list from here into some helper type that can then be shared? https://github.com/dotnet/roslyn-analyzers/blob/f06e25d8ac47ae5a731107c7c82d9d09f343b8ba/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs#L293-L310

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/INamedTypeSymbolExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            if (knownTestAttributes.TryGetValue(attributeClass, out var isTest))
+                return isTest;
+
+            return knownTestAttributes.GetOrAdd(attributeClass, ExtendsFactAttribute(attributeClass, factAttribute));
+        }
+
+        private static bool ExtendsFactAttribute(INamedTypeSymbol namedType, INamedTypeSymbol factAttribute)
```
mavasani: You can use the `Inherits` extension method: https://github.com/dotnet/roslyn-analyzers/blob/f06e25d8ac47ae5a731107c7c82d9d09f343b8ba/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs#L199

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/INamedTypeSymbolExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis;
+
+namespace Roslyn.Diagnostics.Analyzers.Extensions
+{
+    internal static class INamedTypeSymbolExtensions
```
mavasani: Can these extension methods be added here: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/Extensions/INamedTypeSymbolExtensions.cs

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+using Microsoft.CodeAnalysis;
+
+namespace Roslyn.Diagnostics.Analyzers.Extensions
+{
+    internal static class IMethodSymbolExtensions
```
mavasani: Same for this - consider moving to existing extension method file in shared analyzer utilities project.

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs
sharwell: I couldn't think of a good reason to do so, or a good prefix.

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/INamedTypeSymbolExtensions.cs
sharwell: Right now this suppressor is specific to dotnet/roslyn, which uses XUnit.

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/INamedTypeSymbolExtensions.cs
sharwell: I used `DerivesFrom` and optimized it for the clear base class case.

src/Roslyn.Diagnostics.Analyzers/Core/Extensions/INamedTypeSymbolExtensions.cs
sharwell: I didn't evaluate the impact of all of these on the suppressor, so I focused on the primary known case.

PR #3392, opened 2020-03-18
Inline test helper

sharwell:
Inline a `VerifyAnalyzerAsync` helper method to avoid bloating the shared analyzer entry points.

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypesTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypesTests.cs
@@ -341,9 +341,10 @@ End Namespace");
         [Fact]
         public async Task Comparer_ReferenceTypesAreOK()
         {
-            await VerifyCS.VerifyAnalyzerAsync(
-                ReferenceAssemblies.NetCore.NetCoreApp50,
-                @"
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp50,
+                TestCode = @"
 using System;
 using System.Collections.Generic;
 
@@ -356,11 +357,13 @@ namespace TestNamespace
             return ReferenceEqualityComparer.Instance.Equals(string.Empty, test);
         }
     }
-}");
+}",
+            }.RunAsync();
 
-            await VerifyVB.VerifyAnalyzerAsync(
-                ReferenceAssemblies.NetCore.NetCoreApp50,
-                @"
+            await new VerifyVB.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp50,
+                TestCode = @"
 Imports System
 Imports System.Collections.Generic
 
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypesTests.cs
```diff
@@ -341,9 +341,10 @@ End Class
         [Fact]
         public async Task Comparer_ReferenceTypesAreOK()
         {
-            await VerifyCS.VerifyAnalyzerAsync(
-                ReferenceAssemblies.NetCore.NetCoreApp50,
-                @"
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp50,
```
mavasani: Isn't this something that will be needed by majority of new tests that will be added? I am wondering if we should create a common helper such as `VerifyAnalyzerWithNetCoreApp50Async` or such. I will let @bartonjs chime in here.

bartonjs: I think that a lot of the new analyzers are going to end up with the minimum platform reference assembly requirements.

Without fully understanding things, this either seems common (and should stay as a shared helper) or "nothing is common" and the helper should be fully deleted.

sharwell: We have three main options:

1. We could update the default reference assemblies to use `NetCoreApp50` instead of the current default.
1. If the analyzers are only expected to work on .NET 5, we can put the tests for those analyzers in a package that runs on `net5` and update `AdditionalMetadataReferences.Default` to use `NetCoreApp50` for this case. The code fix verifiers use this by default.
1. We could create a new type like `CSharpCodeFixVerifier<TAnalyzer, TCodeFix>` which is the new core verifier type for the new tests (similar to how we have separate a verifier for testing security rules).

sharwell:
> Is it worth doing similar for the other tests? So that there is only one way to write the tests?

This is what we already have:

* Tests that use the default behavior and just have input&rarr;output can use the helper methods
* Tests that need to modify the default behavior use the full-form tests

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseReferenceEqualsWithValueTypesTests.cs
Evangelink: I like 2/ as it goes in the line of what is (will be) done for the support of netcoreapp3.1

sharwell:
@bartonjs @stephentoub we can consider adding methods to this class in the future, but each proposed addition will need to go through a separate review process to understand where it fits into the various testing layers (between a one-off test, a shared helper, or an addition to the testing library itself). I'm merging this for now to restore things to the initial expected state, and we can take this through a full review if we want to add it again in the future.

sharwell:
@bartonjs can I get a ✔️ to move this forward? I don't want this to stall while we figure out the eventual review process.

Issue #3384, opened 2020-03-17
CA2012 false positive for null conditional


sharwell:
#### Package Version

3.0.0-beta2.20167.4

#### Diagnostic ID

CA2012 (Use ValueTasks correctly)

#### Repro steps

https://github.com/dotnet/roslyn/blob/e0a1787d544c8e7c5fef2007dbc8ef6592d8c028/src/Workspaces/SharedUtilitiesAndExtensions/Workspace/Core/Utilities/AsyncSymbolVisitor%601.cs#L15

```csharp
return symbol?.Accept(this) ?? new ValueTask<TResult>(DefaultResult);
```

#### Expected behavior

No diagnostic.

#### Actual behavior

> CA2012: ValueTask instances returned from method calls should be directly awaited, returned, or passed as an argument to another method call. Other usage, such as storing an instance into a local or a field, is likely an indication of a bug, as ValueTask instances must only ever be consumed once.

stephentoub:
ISymbol.Accept returns a ValueTask?

sharwell:
@stephentoub yes that is correct

stephentoub:
Ok.  As an aside, "AcceptAsync" ;)

PR #3393, opened 2020-03-18
Fix UseValueTasksCorrectly handling of null conditionals

stephentoub:
Just treat it the same as a ternary and walk up to the parent.

Also some minor refactoring just to simplify/clarify things.

And explicitly added in handling of discards.

Fixes https://github.com/dotnet/roslyn-analyzers/issues/3384
cc: @sharwell


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -1074,6 +1097,28 @@ public void StoreIntoFields()
...
+                using System;
+                using System.Threading.Tasks;
+
+                class C
+                {
+                    public void Discards()
+                    {
+                        _ = Helpers.ReturnsValueTask();
+                        _ = Helpers.ReturnsValueTaskOfT<string>();
+                        _ = Helpers.ReturnsValueTaskOfInt();
```
sharwell: 💡 Consider adding lines showing that calling `AsTask()` or `Preserve()` on each of these will not result in a diagnostic when assigned to a discard.

Issue #2700, opened 2019-07-26
CA2007 Should respect ValueTask as well


stephentoub:
The current DoNotDirectlyAWaitATaskAnalyzer code only pays attention to `Task` and `Task{TResult}`:
https://github.com/dotnet/roslyn-analyzers/blob/c9d6579f31fd4ad45f3d23187189657343b395a4/src/Microsoft.CodeQuality.Analyzers/Core/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs#L93-L96

It should also be augmented to pay attention to `ValueTask` and `ValueTask{TResult}` if they're available in the compilation unit.

mavasani:
Tagging @AArnott - we are actually planning on adding VS Threading analyzer package as a dependency of FxCop analyzers package, and deprecate CA2007 in favor of a similar rule in the threading package that handles both Task and ValueTask.

AArnott:
Yes, our [VSTHRD111](https://github.com/microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD111.md) rule already supports `ValueTask` for this scenario.

MithrilMan:
What's the suggested approach to solve the issue?
I'm using 
`# CA2007: Consider calling ConfigureAwait on the awaited task
dotnet_diagnostic.CA2007.severity = error` and I'd like to have errors on ValueTask too, is it possible to have that with something to install on top of fxCopAnalyzers?

PR #3397, opened 2020-03-18
Make CA2007 work for ValueTask as well

stephentoub:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/2700


sharwell:
😂 There is some code in Roslyn that uses `ValueTask<TResult>` specifically to avoid CA2007 in JTF-aware code.

PR #3400, opened 2020-03-19
Fix build warnings and update compiler

sharwell:


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -44,7 +44,7 @@ public override void Initialize(AnalysisContext analysisContext)
 
             analysisContext.RegisterCompilationStartAction(startContext =>
             {
-                var instantiatedTypes = new ConcurrentDictionary<INamedTypeSymbol, object?>();
+                ConcurrentDictionary<INamedTypeSymbol, object?> instantiatedTypes = new ConcurrentDictionary<INamedTypeSymbol, object?>();
```
jcouv: Those explicit types were not necessary, are they?

Evangelink: The commit introducing these changes is named "Use explicit type where compiler fails to evaluate nullability" so I think they matters.

sharwell: @jcouv yes, they are necessary. `var` doesn't work with local functions, at least not here.

mavasani: Do we need to file a compiler bug?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotPassLiteralsAsLocalizedParameters.cs
```diff
@@ -66,7 +66,7 @@ public override void Initialize(AnalysisContext context)
                         return;
                     }
 
-                    var lazyValueContentResult = new Lazy<DataFlowAnalysisResult<ValueContentBlockAnalysisResult, ValueContentAbstractValue>?>(
+                    Lazy<DataFlowAnalysisResult<ValueContentBlockAnalysisResult, ValueContentAbstractValue>?> lazyValueContentResult = new Lazy<DataFlowAnalysisResult<ValueContentBlockAnalysisResult, ValueContentAbstractValue>?>(
```
mavasani: Same here - this seems unfortunate.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotReferSelfInSerializableClass.cs
```diff
@@ -66,14 +66,14 @@ public sealed override void Initialize(AnalysisContext context)
                         return;
                     }
 
-                    var forwardGraph = new ConcurrentDictionary<ISymbol, ConcurrentDictionary<ISymbol, bool>>();
-                    var invertedGraph = new ConcurrentDictionary<ISymbol, ConcurrentDictionary<ISymbol, bool>>();
+                    ConcurrentDictionary<ISymbol, ConcurrentDictionary<ISymbol, bool>> forwardGraph = new ConcurrentDictionary<ISymbol, ConcurrentDictionary<ISymbol, bool>>();
```
mavasani: Please retarget all security rule changes to 2.9.x branch. Also, there is no nullability involved here, so why was this required?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
sharwell: I think there are already bugs filed for improving local function support. It hit Roslyn too.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotPassLiteralsAsLocalizedParameters.cs
sharwell: ➡️ Covered by https://github.com/dotnet/roslyn-analyzers/pull/3400#discussion_r394759842. The compiler requires this change.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotReferSelfInSerializableClass.cs
sharwell: It's not practical to backport this change. The 2.9.x branch doesn't support this toolchain or compiler version so it would be a large backport of much more than just this pull request.

I don't believe these changes will be a problem in practice. /cc @dotpaul for awareness.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
sharwell: ➡️ https://github.com/dotnet/roslyn/issues/41689 and https://github.com/dotnet/roslyn/issues/41705

PR #3401, opened 2020-03-19
CA1713: Events should not have 'Before' or 'After' prefix

### LINKED ISSUE #449: Port FxCop rule CA1713: EventsShouldNotHaveBeforeOrAfterPrefix
**Title:** Events should not have before or after prefix

**Description:**

The name of an event starts with ""Before"" or ""After"". To name related events that are raised in a specific sequence, use the present or past tense to indicate the relative position in the sequence of actions.

**Dependency:** None

**Notes:**

See if StyleCop has implemented it

Evangelink:
Fix #449.


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
```diff
@@ -0,0 +1,55 @@
...
+                if (IsViolatingPrefix(context.Symbol.Name, AfterKeyword) ||
+                    IsViolatingPrefix(context.Symbol.Name, BeforeKeyword))
+                {
+                    context.ReportDiagnostic(context.Symbol.CreateDiagnostic(Rule));
+                }
+            }, SymbolKind.Event);
+        }
+
+        private static bool IsViolatingPrefix(string eventName, string prefix)
+            => eventName.StartsWith(prefix, StringComparison.Ordinal) &&
```
Evangelink, line 51: @mavasani I guess the old implementation was also using the `WordParser` here, isn't it?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
```diff
@@ -0,0 +1,55 @@
...
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.EventsShouldNotHaveBeforeOrAfterPrefixTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.EventsShouldNotHaveBeforeOrAfterPrefixMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.EventsShouldNotHaveBeforeOrAfterPrefixDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Naming,
+            RuleLevel.IdeSuggestion,
```
mavasani, line 28:
```suggestion
            RuleLevel.Disabled,
```

mavasani: Any heuristic based rule should be disabled by default in .NET analyzers

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
mavasani: Surprisingly not: `if (name.StartsWith("Before", StringComparison.Ordinal) || name.StartsWith("After", StringComparison.Ordinal))`

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefixTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        [Fact]
+        public async Task CA1713_EventNameStartsWithAfter_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+public class Class1
+{
+    public event EventHandler AfterClose;
+}",
+            GetCSharpResultAt(5, 31));
```
mavasani, line 26: Can the tests be changed to be markup based?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
```diff
@@ -0,0 +1,55 @@
...
+                    IsViolatingPrefix(context.Symbol.Name, BeforeKeyword))
+                {
+                    context.ReportDiagnostic(context.Symbol.CreateDiagnostic(Rule));
+                }
+            }, SymbolKind.Event);
+        }
+
+        private static bool IsViolatingPrefix(string eventName, string prefix)
+            => eventName.StartsWith(prefix, StringComparison.Ordinal) &&
+                eventName.Length > prefix.Length &&
```
mavasani, line 52: This seems like a good check, even though it was not done in legacy implementation.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
```diff
@@ -0,0 +1,55 @@
...
+                char.IsUpper(eventName[prefix.Length]);
```
mavasani, line 53: I think we should remove this check. If we want to be smarter, we should switch to word parser instead.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
Evangelink: Ok!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
```diff
@@ -25,31 +22,31 @@ public sealed class EventsShouldNotHaveBeforeOrAfterPrefix : DiagnosticAnalyzer
...
         public override void Initialize(AnalysisContext analysisContext)
         {
             analysisContext.EnableConcurrentExecution();
             analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
             analysisContext.RegisterSymbolAction(context =>
             {
-                if (IsViolatingPrefix(context.Symbol.Name, AfterKeyword) ||
-                    IsViolatingPrefix(context.Symbol.Name, BeforeKeyword))
+                var wordParse = new WordParser(context.Symbol.Name, WordParserOptions.SplitCompoundWords);
```
mavasani, line 41: We don't want to do this unless we know it is likely that the symbol name starts with "Before" or "After". Can we have a cheaper bail out check upfront?

Evangelink: Sure! Do you think the `WordParser` is valuable? I could revert to my previous implementation if needed.

mavasani: It seems fine to use `WordParser`. Just avoid the perf hit of invoking it for every symbol.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EventsShouldNotHaveBeforeOrAfterPrefix.cs
```diff
@@ -0,0 +1,64 @@
...
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            analysisContext.RegisterSymbolAction(context =>
+            {
+                var eventName = context.Symbol.Name;
+
+                if (!eventName.StartsWith(BeforeKeyword, StringComparison.Ordinal) &&
```
Evangelink, line 47: @mavasani Is it enough or shall I use the span technique?

PR #3404, opened 2020-03-19
Update RS0030 to report on usage of overriden banned APIs

### LINKED ISSUE #3295: Banning a virtual/abstract method should also ban usages of overridden versions of it
The API analyzer does fire on overrides of a method that has been banned, which makes the below pattern very hard to ban.

#### Analyzer package

BannedApiAnalyzers

#### Diagnostic ID

RS0030

#### Repro steps

```
M:Microsoft.VisualStudio.ProjectSystem.OnceInitializedOnceDisposed.Initialize();Did you mean OnceInitializedOnceDisposed.EnsureInitialized?
```

``` C#

public class Component : OnceInitializedOnceDisposed
{
    public string Property
    {
        get 
        {
            Initialized();
            return "Hello";
        }
    }

   protected override void Initialize()
   {
   }
}
```

#### Expected behavior
Warns about incorrect usage of Initialize

#### Actual behavior
Does not warn

Evangelink:
Fix #3295

### DIFF
--- a/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
@@ -348,7 +348,7 @@ namespace Microsoft.CodeAnalysis.BannedApiAnalyzers
                     {
                         yield return currentSymbol;
 
-                        // It's possible to have `IsOverride` true and yet have `GetOverriddeMember` returning false when the code is invalid
+                        // It's possible to have `IsOverride` true and yet have `GetOverriddeMember` returning null when the code is invalid
                         // (e.g. base symbol is not marked as `virtual` or `abstract` and current symbol has the `overrides` modifier).
                         currentSymbol = currentSymbol.IsOverride
                             ? currentSymbol.GetOverriddenMember()?.OriginalDefinition
--- a/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
@@ -326,16 +326,34 @@ namespace Microsoft.CodeAnalysis.BannedApiAnalyzers
             {
                 RoslynDebug.Assert(entryBySymbol != null);
 
-                symbol = symbol.OriginalDefinition;
-
-                if (entryBySymbol.TryGetValue(symbol, out var entry))
+                foreach (var currentSymbol in GetSymbolAndOverridenSymbols(symbol))
                 {
-                    reportDiagnostic(
-                        Diagnostic.Create(
-                            SymbolIsBannedAnalyzer.SymbolIsBannedRule,
-                            syntaxNode.GetLocation(),
-                            symbol.ToDisplayString(SymbolDisplayFormat),
-                            string.IsNullOrWhiteSpace(entry.Message) ? "" : ": " + entry.Message));
+                    if (entryBySymbol.TryGetValue(currentSymbol, out var entry))
+                    {
+                        reportDiagnostic(
+                            Diagnostic.Create(
+                                SymbolIsBannedAnalyzer.SymbolIsBannedRule,
+                                syntaxNode.GetLocation(),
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
```diff
@@ -18,15 +19,15 @@ namespace Microsoft.CodeAnalysis.BannedApiAnalyzers.UnitTests
 
     public class SymbolIsBannedAnalyzerTests
     {
-        private static DiagnosticResult GetCSharpResultAt(int line, int column, DiagnosticDescriptor descriptor, string v3, string v4)
+        private static DiagnosticResult GetCSharpResultAt(int line, int column, DiagnosticDescriptor descriptor, string bannedMemberName, string message)
```
Evangelink: Rename parameters as it wasn't clear what they were supposed to be.

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
```diff
@@ -761,6 +762,147 @@ void M()
...
+
+    public class C2 : C1
+    {
+        public override void Method1() {}
+        public override int Property1 { get; set; }
+        public override event Action Event1;
+
+        public override void Method2()
+        {
+            base.Method2();
```
Evangelink: The current implementation raises a diagnostic on this line but I don't think we should. I think that when an API is banned we either want to prevent all usages, in which case just the `override` definition itself should trigger the diagnostic or we want to prevent calls but we do allow overrides in which case the call to base in the overrides should be fine.

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
```diff
@@ -1176,6 +1318,139 @@ public async Task VisualBasic_BannedClass_DocumentationReference()
...
         }
 
+        [Fact, WorkItem(3295, "https://github.com/dotnet/roslyn-analyzers/issues/3295")]
+        public async Task VisualBasic_BannedAbstractVirtualMemberAlsoBansOverrides()
+        {
+            var source = @"
+Imports System
+
+Namespace N
+    Public MustInherit Class C1
```
Evangelink: It's not possible to have `abstract` or `virtual` events in vbnet.

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
```diff
@@ -326,16 +326,34 @@ void VerifySymbol(Action<Diagnostic> reportDiagnostic, ISymbol symbol, SyntaxNod
...
-                            symbol.ToDisplayString(SymbolDisplayFormat),
-                            string.IsNullOrWhiteSpace(entry.Message) ? "" : ": " + entry.Message));
+                    ISymbol? currentSymbol = symbol.OriginalDefinition;
+
+                    while (currentSymbol != null)
+                    {
+                        yield return currentSymbol;
+
+                        // It's possible to have `IsOverride` true and yet have `GetOverriddeMember` returning false when the code is invalid
+                        // (e.g. base symbol is not marked as `virtual` or `abstract` and current symbol has the `overrides` modifier).
```
mavasani: Nice you handled this case, I was primarily looking for this case being covered in this review :-)

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
mavasani: Can we use the inline annotations in tests for diagnostics? It is very difficult to map line numbers to which code actually fired the diagnostic.

Evangelink: I will need to duplicate the code but yeah sure!

PR #3405, opened 2020-03-20
CA1070: Do not use mutable struct in readonly field

### LINKED ISSUE #2811: New rule: using mutable structs in readonly fields

Evangelink:
Fix #2811

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotUseMutableStructInReadonlyField.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotUseMutableStructInReadonlyField.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+
+namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
+{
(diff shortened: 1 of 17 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotUseMutableStructInReadonlyField.cs
```diff
@@ -0,0 +1,62 @@
...
+                {
+                    var field = (IFieldSymbol)context.Symbol;
+
+                    if (!field.IsReadOnly ||
+                        field.Type?.TypeKind != TypeKind.Struct)
+                    {
+                        return;
+                    }
+
+                    if (field.Type.Equals(spinLockType) ||
```
Evangelink: Only looking at field type as @stephentoub was suggesting. We could make the rule more advanced and look at mutation calls rather than field definition.

mavasani:
How does this play out with https://github.com/dotnet/roslyn-analyzers/pull/2831?

Evangelink:
@mavasani I missed it sorry. Closing this PR.

PR #3406, opened 2020-03-20
Remove duplicate helper

Evangelink:


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
```diff
@@ -914,7 +914,7 @@ private static bool IsSpecialEmptyMember(ISymbol symbol)
             {
                 return symbol.IsStatic &&
                     symbol.Name.Equals("Empty", StringComparison.Ordinal) &&
-                    (symbol.IsReadOnlyFieldOrProperty() || symbol.Kind == SymbolKind.Method) &&
+                    (symbol.IsReadOnly() || symbol.Kind == SymbolKind.Method) &&
```
mavasani: Is this an existing extension method? Will this now return true for readonly structs?

Evangelink: It's an already existing helper in the same class: https://github.com/dotnet/roslyn-analyzers/blob/3cc0345042a39d23c8b82fd02d0d2dd298dc92cb/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L684-L697


FYI I looked at the TODO but with the current version of Roslyn the `IsReadOnly` on method and types is still not available.

mavasani: IMO, it will be safer to handle the TODO in `IsReadOnly` extension by handling `IMethodSymbol` and `ITypeSymbol`, so that extension returns true for any readonly symbol and revert changes currently in the PR so `IsSpecialEmptyMember` does not return true for readonly struct types

Evangelink: Not sure I got what you meant.

mavasani: Retain `IsReadOnlyFieldOrProperty` to only return true for fields and property symbols and update `IsReadOnly` to return true for any possible symbol type. This ensures `IsSpecialEmptyMember` retains its current functionality by continuing to invoke `IsReadOnlyFieldOrProperty`

Evangelink: Ok so closing this PR. As I said, I cannot fix the TODO for the time being, the current version of Microsoft.CodeAnalysis is not recent enough (even if it is 3.0).

PR #3417, opened 2020-03-23
CA2109: Review visible event handlers

### LINKED ISSUE #488: Port FxCop rule CA2109: ReviewVisibleEventHandlers
**Title:** Review visible event handlers

**Description:**

A public or protected event-handling method was detected. Event-handling methods should not be exposed unless absolutely necessary.

**Dependency:** None

**Notes:**

@michaelcfanning: Validate this decision

Evangelink:
Fix #488


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -39,17 +41,43 @@ public abstract class ReviewVisibleEventHandlersAnalyzer : DiagnosticAnalyzer
...
+                    }
+
+                    if (method.IsOverride || method.IsImplementationOfAnyInterfaceMember())
+                    {
+                        return;
+                    }
+
+                    if (method.HasEventHandlerSignature(eventArgsType))
+                    {
+                        var rule = method.HasAttribute(securityPermissionAttributeType)
```
Evangelink: I don't know if there is any other criteria triggering the "security" part of the rule.

mavasani: Seems the security rule fired only for cases which are now deprecated:

```csharp
    void DoSecurityCheck(Method method)
    {      
      ParameterCollection parameters = method.Parameters;

      for (int i = 0; i < parameters.Count; i++)
      {
        Parameter parameter = parameters[i];
        if (RuleUtilities.IsEventArgs(parameter.Type)
          && SecurityRuleUtilities.HasLinkDemandOrDemandSecurityAction(method))
```

mavasani: https://docs.microsoft.com/en-us/dotnet/api/system.security.permissions.securityaction?view=netframework-4.8

I think we should remove the security rule.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -567,7 +567,10 @@ public static bool HasEventHandlerSignature(this IMethodSymbol method, [NotNullW
             => eventArgsType != null &&
                method.Parameters.Length == 2 &&
                method.Parameters[0].Type.SpecialType == SpecialType.System_Object &&
-               method.Parameters[1].Type.DerivesFrom(eventArgsType, baseTypesOnly: true);
+               // FxCop compat: Struct with name ending with "EventArgs" are allowed
+               // + UWP has specific EventArgs not inheriting from 'System.EventArgs'.
+               // See https://github.com/dotnet/roslyn-analyzers/issues/3106
```
mavasani, line 572: Nice!

src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -487,6 +487,7 @@ internal static class WellKnownTypeNames
         public const string SystemXmlXslXslCompiledTransform = "System.Xml.Xsl.XslCompiledTransform";
         public const string SystemXmlXslXslTransform = "System.Xml.Xsl.XslTransform";
         public const string SystemXmlXslXsltSettings = "System.Xml.Xsl.XsltSettings";
+        public const string SystemSecurityPermissionsSecurityPermissionAttribute = "System.Security.Permissions.SecurityPermissionAttribute";
```
mavasani, line 490: Keep the list sorted?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
Evangelink: Do you want to simply remove the rule or make it not raising for netcore + net fwk >= 4.8?

src/Utilities/Compiler/WellKnownTypeNames.cs
Evangelink: I didn't even paid attention it was _finally_ sorted!!! 🎉 🎉

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
mavasani: I'd say let just remove the rule as it is pertains to an obsolete API

src/Utilities/Compiler/WellKnownTypeNames.cs
mavasani: Yep, thanks @carlossanlop!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -20,36 +22,65 @@ public abstract class ReviewVisibleEventHandlersAnalyzer : DiagnosticAnalyzer
...
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray<DiagnosticDescriptor>.Empty;
-        //ImmutableArray.Create(SecurityRule, DefaultRule);
-
-#pragma warning disable RS1025 // Configure generated code analysis
+        internal static DiagnosticDescriptor SecurityRule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessageSecurity,
+            DiagnosticCategory.Security,
```
Evangelink, line 29: @mavasani I guess I'll need to change this to `Design`, isn't it?

mavasani: I'd say lets retain the existing category, which is Security: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2109?view=vs-2019. Avoids us from unnecessarily breaking the category to ID range mapping, and the description of the rule does indicate it will still be flagging a security concern.

> Do not expose event-handling methods unless absolutely necessary. An event handler, a delegate type, that invokes the exposed method can be added to any event as long as the handler and event signatures match. Events can potentially be raised by any code, and are frequently raised by highly trusted system code in response to user actions such as clicking a button. Adding a security check to an event-handling method does not prevent code from registering an event handler that invokes the method.

src/Utilities/Compiler/WellKnownTypeNames.cs
carlossanlop: 😃

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -10,46 +11,60 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
...
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessageDefault,
+            DiagnosticCategory.Security,
+            RuleLevel.Disabled,
+            description: s_localizableDescription,
+            isPortedFxCopRule: true,
+            isDataflowRule: false,
+            isEnabledByDefaultInFxCopAnalyzers: false);
```
mavasani, line 32: I feel this will be a valuable rule to be enabled by default in fxcop analyzers.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -10,46 +11,60 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
...
+                        return;
+                    }
+
+                    if (!method.HasEventHandlerSignature(eventArgsType))
+                    {
+                        return;
+                    }
+
+                    // FxCop compat: only analyze externally visible symbols by default.
+                    if (!method.MatchesConfiguredVisibility(context.Options, Rule, context.Compilation, context.CancellationToken))
```
mavasani, line 60: I believe this is one rule which is always applicable only for externally visible symbols, i.e. the `api_surface` option should not be configurable here. We should directly check if method is externally visible.

Evangelink: Makes sense!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
Evangelink: IdeSuggestion? or Warning?

mavasani: No, I meant just remove `isEnabledByDefaultInFxCopAnalyzers: false`

mavasani: Lets keep it disabled for now for NetAnalyzers package.

src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -277,6 +277,7 @@ internal static class WellKnownTypeNames
         public const string SystemSecurityCryptographyX509CertificatesX509Certificate2 = "System.Security.Cryptography.X509Certificates.X509Certificate2";
         public const string SystemSecurityCryptographyX509CertificatesX509Chain = "System.Security.Cryptography.X509Certificates.X509Chain";
         public const string SystemSecurityCryptographyX509CertificatesX509Store = "System.Security.Cryptography.X509Certificates.X509Store";
+        public const string SystemSecurityPermissionsSecurityPermissionAttribute = "System.Security.Permissions.SecurityPermissionAttribute";
```
Evangelink, line 280: No longer used, I'll remove this.

docs/Analyzer Configuration.md
```diff
@@ -72,33 +72,34 @@ Configurable Rules:
 [CA1044](https://docs.microsoft.com/visualstudio/code-quality/ca1044), 
 [CA1045](https://docs.microsoft.com/visualstudio/code-quality/ca1045),
 [CA1047](https://docs.microsoft.com/visualstudio/code-quality/ca1047),
-[CA1051](https://docs.microsoft.com/visualstudio/code-quality/ca1051),
```
Evangelink, line 75: I'll fix this merge error.

Evangelink:
@mavasani Looks like the build is stuck.

mavasani:
Seems like build did succeed: https://github.com/dotnet/roslyn-analyzers/pull/3417/checks, but status is not correctly updated.

docs/Analyzer Configuration.md
```diff
@@ -94,6 +94,7 @@ Configurable Rules:
 [CA1802](https://docs.microsoft.com/visualstudio/code-quality/ca1802),
 [CA1815](https://docs.microsoft.com/visualstudio/code-quality/ca1815),
 [CA1819](https://docs.microsoft.com/visualstudio/code-quality/ca1819),
+[CA2109](https://docs.microsoft.com/visualstudio/code-quality/ca2109),
```
mavasani, line 97: This is no longer true, correct? The rule is not configurable.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlersTests.cs
```diff
@@ -1,8 +1,105 @@
...
+    Public Class RoutedEventArgs
+    End Class
+End Namespace
+
+Public Class Program
+    Public Sub [|Handler1|](ByVal sender As Object, ByVal args As Windows.UI.Xaml.RoutedEventArgs)
+    End Sub
+End Class");
+        }
+
```
mavasani, line 94: Please add no diagnostic cases when the event handler is not externally visible.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -10,46 +11,58 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
...
+
+            analysisContext.RegisterCompilationStartAction(context =>
+            {
+                var eventArgsType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemEventArgs);
+
+                context.RegisterSymbolAction(context =>
+                {
+                    var method = (IMethodSymbol)context.Symbol;
+
+                    if (method.IsOverride || method.IsImplementationOfAnyInterfaceMember())
```
mavasani, line 48: Is it possible to write a test for these checks?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -10,46 +11,58 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
...
+                    {
+                        return;
+                    }
+
+                    if (!method.HasEventHandlerSignature(eventArgsType))
```
mavasani, line 53: Can we have a test where this if condition succeeds and we bail out here?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
mavasani: `method.IsImplementationOfAnyInterfaceMember` seems like the most expensive check here. Can this if statement be moved below the other two if statements?

Evangelink:
@mavasani Could you review?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -10,46 +11,58 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
...
+                context.RegisterSymbolAction(context =>
+                {
+                    var method = (IMethodSymbol)context.Symbol;
+
+                    if (!method.IsExternallyVisible())
+                    {
+                        return;
+                    }
+
+                    if (!method.HasEventHandlerSignature(eventArgsType))
```
mavasani, line 53: Can you update `HasEventHandlerSignature`:
1. Update doc comments that still mention that second parameter type is or derives from `eventArgsType`
2. Remove the `[NotNullWhen(returnValue: true)]` attribute on `eventArgsType` parameter?

https://github.com/dotnet/roslyn-analyzers/blob/2763ec7c3d235319e15c3d5d9db111e1715f9be4/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs#L561-L566

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
```diff
@@ -10,46 +11,58 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
...
-
-        private static readonly LocalizableString s_localizableMessageSecurity = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.ReviewVisibleEventHandlersMessageSecurity), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
         private static readonly LocalizableString s_localizableMessageDefault = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.ReviewVisibleEventHandlersMessageDefault), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
         private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.ReviewVisibleEventHandlersDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
 
-        internal static DiagnosticDescriptor SecurityRule = DiagnosticDescriptorHelper.Create(RuleId,
```
mavasani, line 23: Can you please file a doc issue/submit doc PR to update https://docs.microsoft.com/visualstudio/code-quality/ca2109?view=vs-2019 to mention that the link demand part is now deprecated?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
Evangelink: 1. I will update the comment to say we also mention the name ends with

2. The first line of the method is still `=> eventArgsType != null &&` so that seems to be fair keep it. Did you meant that you feel like we can get rid of the null check and allow the method to return `true` even when `EventArgs` type was not found? I have some doubts because if we can't find this type in system we are most likely screwed anyways for something else. Let me know your preference.

3. The helper method mentions that the method returns `void` which isn't tested. Shall I update the doc or the method? I would be in favor of updating the implementation to check for `void`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlers.cs
Evangelink: https://github.com/MicrosoftDocs/visualstudio-docs/issues/5248

Evangelink:
Ping @mavasani

PR #3418, opened 2020-03-23
Update CA1508 to avoid FP with variable declarator in using

### LINKED ISSUE #2985: CA1508 False positive on using statement
#### Analyzer package

#### Package Version

v2.9.6 (Latest)

#### Diagnostic ID

CA1508 - Avoid Dead Conditional Code

#### Repro steps

image

```
		static void Test(byte[] data)
		{
			using (var ms = new MemoryStream(data))
			{
				Console.WriteLine(ms);
			}
		}

		static void TestError(byte[] data)
		{
			using (var ms = new MemoryStream(data))
			{
				Console.WriteLine(bool.Parse(ms.ToString()));
			}
		}
```

#### Expected behavior

No CA1508 on the "new MemoryStream(data)" part

It's not clear what the rule wants me to remove.  I suspect it should be ignoring the generated check added by the using block.

Note that this only seems to happen if you pass the using object to another function, or deference it?  The Test(...) method doesn't show the error.

#### Actual behavior

CA1508

### LINKED ISSUE #3063: CA1508 (Avoid Dead Conditional Code) using statement continues to give false positive
The continuation of #2250 as the latest version still has this (or similar) issue.

Would be great if you could prioritize the resolution as an internal partner's ask, Identity Platform Engineering would be grateful.

#### Analyzer package

#### Package Version

v2.9.7 (Latest)

#### Diagnostic ID

CA1508 (no article at https://docs.microsoft.com/en-us/visualstudio/code-quality/)

#### Repro steps

```c#
using (var fabricClient = new FabricClient())
{
}
```

#### Expected behavior

No error

#### Actual behavior

>'fabricClient = new FabricClient()' is never 'null'

More examples:

image

Evangelink:
Fix #3063 and fix #2985

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidDeadConditionalCode.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidDeadConditionalCode.cs
@@ -3,7 +3,6 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
-using System.Linq;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidDeadConditionalCode.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidDeadConditionalCode.cs
@@ -99,8 +99,7 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
                                 continue;
                             }
 
-                            var ops = cfg.DescendantOperations().ToArray();
-                            foreach (var operation in ops)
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidDeadConditionalCode.cs
```diff
@@ -151,9 +152,10 @@ public override void Initialize(AnalysisContext context)
                                         }
 
                                         var originalOperation = operationRoot.SemanticModel.GetOperation(operation.Syntax, operationBlockContext.CancellationToken);
-                                        if (originalOperation is IAssignmentOperation)
+                                        if (originalOperation is IAssignmentOperation ||
+                                            originalOperation is IVariableDeclaratorOperation)
                                         {
-                                            // Skip compiler generated IsNull operation for assignment within a using.
+                                            // Skip compiler generated IsNull operation for assignment/variable declaration within a using.
```
Evangelink: I would have expected that we check that there is a parent operation being the using. I am assuming that the code was correctly handling the assignment in the using so I have simply added the variable declarator case.

PR #3420, opened 2020-03-23
Implement RS0042 (Do not copy value)

sharwell:
At a high level, this pull request approaches non-copyable value type restrictions by preventing all use of such types (similar to Banned API analyzer), followed by selectively allowing known safe patterns.

For the implementation, I have been iterating over **Roslyn.sln** using AnalyzerRunner to log all diagnostics produced by this analyzer.

> .\\artifacts\\bin\\AnalyzerRunner\\Release\\netcoreapp3.1\\AnalyzerRunner.exe ..\\roslyn-analyzers\\artifacts\\bin\\Roslyn.Diagnostics.Analyzers\\Debug\\netstandard2.0\\ .\\Roslyn.sln /stats /concurrent /a DoNotCopyValue /log DoNotCopyValue.txt

Supersedes #2831


src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
```diff
@@ -0,0 +1,676 @@
...
+            s_localizableTitle,
+            s_localizableUnsupportedUseMessage,
+            DiagnosticCategory.RoslynDiagnosticsReliability,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: s_localizableUnsupportedUseDescription,
+            helpLinkUri: null,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        internal static DiagnosticDescriptor NoBoxingRule = new DiagnosticDescriptor(
```
sharwell: 📝 Originally the analyzer reported all boxing of non-copyable types. This was particularly problematic for struct enumerators, so the rule was subsequently modified to allow _moving_ a value into a box, and blocking the unboxing. Boxed non-copyable values may be used via `Unsafe.Unbox<T>`, which returns a reference to the boxed value without copying it out of the box.

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
```diff
@@ -0,0 +1,676 @@
...
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.RoslynDiagnosticsReliability,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: s_localizableDescription,
+            helpLinkUri: null,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        internal static DiagnosticDescriptor UnsupportedUseRule = new DiagnosticDescriptor(
```
sharwell: 📝 This rule is reported for cases where a non-copyable value type is used in a manner that the analyzer does not understand. Eventually, we want more specific diagnostics to be reported for invalid use cases, and no diagnostics to be reported for valid use cases.

jkotas:
This looks very similar to https://github.com/ufcpp/NonCopyableAnalyzer . What do you think about using an attribute to mark the non-copyable structs?

sharwell:
@jkotas The manner in which non-copyable types are identified is separate from how they are handled after identification. For now, I'm experimenting at a rather large scale in Roslyn.sln by treating all mutable value types as non-copyable.

mavasani:

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
```diff
@@ -0,0 +1,1358 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterCompilationStartAction(context =>
+            {
+                var cache = new NonCopyableTypesCache(context.Compilation);
+                context.RegisterOperationBlockAction(context => AnalyzeOperationBlock(context, cache));
```
sharwell: 📝 This needs to also have a symbol analyzer to ensure non-copyable types are not used incorrectly (e.g. the early checks would be for the use of a non-copyable type as a generic argument).

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
mavasani: Will this rule be removed before you merge this PR?

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
```diff
@@ -0,0 +1,1443 @@
...
+
+            public NonCopyableWalker(OperationBlockAnalysisContext context, NonCopyableTypesCache cache)
+            {
+                _context = context;
+                _cache = cache;
+            }
+
+            public override void VisitAddressOf(IAddressOfOperation operation)
+            {
+                CheckTypeInUnsupportedContext(operation);
```
mavasani, line 164: Do you override every single method on `OperationWalker`, with this being the default implementation?

src/Roslyn.Diagnostics.Analyzers/UnitTests/DoNotCopyValueTests.cs
```diff
@@ -0,0 +1,697 @@
...
+        _ = local.Target;
+        _ = reflocal.Target;
+        _ = reflocal.Target;
+
+        _ = readonlyfield.Target;
+        _ = refreadonlylocal.Target;
+    }
+}
+",
+                // /0/Test0.cs(23,13): warning RS0042: Unsupported use of non-copyable type 'System.Runtime.InteropServices.GCHandle' in 'FieldReference' operation
```
mavasani, line 233: Can we use markup based syntax here to easily map the flagged code?

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
sharwell: Yes. I do that until I manually update the method with a more complete implementation.

src/Roslyn.Diagnostics.Analyzers/UnitTests/DoNotCopyValueTests.cs
sharwell: Yes, the new markup features finally allow this to be cleaned up.

sharwell: Done!

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
sharwell: No. It can be removed if we ever get to a case where all the potential copies are covered by other messages.

PR #3425, opened 2020-03-25
PublicAPI: Fix spurious oblivious warning on nested enum

jcouv:
Follows up on https://github.com/dotnet/roslyn-analyzers/pull/3386 (tracking oblivious APIs)

In trying the analyzer on roslyn, I found two bugs:
- we would produce a spurious warning on nested enums in `#nullable enable` context because of incorrect detection of oblivious types (more details below)
- in the upgrade of oblivious APIs previously marked with `~`, we would not recognize the existing API line as corresponding to a newly annotated member (the `~` needed to be trimmed)

Sorry I missed those scenarios in my previous PR :-(

Detection of oblivious types:
In `class C { enum E { None } }`, we would look at the field `C.E.None` which returns a `C.E` and see an oblivious on the containing type. That is not a place where an annotation `?` could go, therefore that is not a place where oblivious matters.

@mavasani @sharwell for review. Thanks

### DIFF
--- a/src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
+++ b/src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
@@ -1370,26 +1370,6 @@ C.C() -> void
         }
 
         [Fact]
-        public async Task TestAddAndRemoveMembers_CSharp_Fix_WithAddedNullability_WithOblivious()
-        {
-            var source = @"
-#nullable enable
-public class C
-{
-    public string? {|RS0016:ChangedField|};
-}
-";
-            var shippedText = $@"{DeclarePublicApiAnalyzer.NullableEnable}";
-            var unshippedText = @"C
-C.C() -> void
-{|RS0017:~C.ChangedField -> string|}";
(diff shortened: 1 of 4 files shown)


src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1285,26 +1369,6 @@ public class C
             await VerifyCSharpAdditionalFileFixAsync(source, shippedText, unshippedText, fixedUnshippedText);
         }
 
-        [Fact]
-        public async Task TestAddAndRemoveMembers_CSharp_Fix_WithAddedNullability_WithOblivious()
```
jcouv: :memo: in using the feature, I realized this is not the right experience: the RS0017 (API removed) shouldn't be there, and the RS0016 (API added) should be a RS0036 (API annotated) instead.

So this test was moved. `LegacyAPIWithObliviousMarkerGetsAnnotatedAsNullable` in `AnnotatePublicApiAnalyzerTests.cs` file shows the new behavior, which is to annotate in-place.

jcouv:
@mavasani @sharwell for review. Thanks

PR #3428, opened 2020-03-25
Update CA1716 to allow configuration of analyzed symbol kinds

### LINKED ISSUE #1858: CA1716 warns for vb.net keyword in c#

Evangelink:
Fix #1858

Documentation update request: https://github.com/MicrosoftDocs/visualstudio-docs/issues/4982


docs/Analyzer Configuration.md
```diff
@@ -501,3 +501,14 @@ Examples:
...
 |`dotnet_code_quality.CA1501.additional_inheritance_excluded_symbol_names = N:My*` | Matches all types whose containing namespace starts with 'My' and all types from the 'System' namespace |
+
+### Analyzed symbol kinds
+Option Name: `analyzed_symbol_kinds`
+
+Configurable Rules: [CA1716](https://docs.microsoft.com/en-us/visualstudio/code-quality/CA1716)
+
+Option Values: One or more fields of enum [Microsoft.CodeAnalysis.SymbolKind](https://roslynsourceindex.azurewebsites.net/#Microsoft.CodeAnalysis/Symbols/SymbolKind.cs,30fd9c0834bef6ff) as a comma separated list.
+
+Default Value: Namespace, NamedType, Method, Property, Event, Parameter
```
mavasani, line 512: We should state that the default is rule specific, and for CA1716 this is the default value.

PR #3429, opened 2020-03-25
Update CA1303 to configure the use of the naming heuristic

### LINKED ISSUE #2933: CA1303 is more irritating than helpful
#### Analyzer package

[Microsoft.NetCore.Analyzers](https://www.nuget.org/packages/Microsoft.NetCore.Analyzers/)

#### Package Version

v2.9.6 (Latest)

#### Diagnostic ID

[CA1303](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1303-do-not-pass-literals-as-localized-parameters?view=vs-2019)

#### Repro steps

1. Create a function with `string text` as a parameter.
2. Call the function with a string literal.

#### Expected behavior
Should trigger only on developer-specified parameter names, or only those tagged with the Localizable attribute. As is, it's more irritating than helpful.

#### Actual behavior
Triggers every time, whether you want it to or not, and is not configurable in any way without turning off the entire rule.

----

This was previously mentioned [here](https://github.com/MicrosoftDocs/visualstudio-docs/issues/3943) but the user never followed up on it. Two or three separate rules would be a good start, and being able to configure the parameter name list would be extremely helpful. The System.Exception customization in #2602 doesn't really do anything when you're working with (or designing) libraries that modify language-invariant resources.

The use of the name "text" as a parameter is ubiquitous in libraries that have nothing to do with UI or other language-specific resources. If, say, I were programatically generating code and the program had an AddText(string text) method, I would certainly never want to localize "foreach" to "pourchaque"! Similarly, as a developer, I don't want to be limited to having to design AddText(string txtCuzRoslynAnalyzersSaysNoT_e_x_t) functions, just to be sure I don't use the word "text" in a function where that's the most logical name, and don't hit on any names that might get added to the list later on.

It's a well-intentioned rule, just not a very practical one as it stands now.

Evangelink:
Fix #2933

Doc update request: https://github.com/MicrosoftDocs/visualstudio-docs/issues/4981


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotPassLiteralsAsLocalizedParametersTests.cs
```diff
@@ -1255,16 +1283,30 @@ End Class
...
+        [InlineData("Text", false)]
+        [InlineData("Text", true)]
+        [InlineData("Caption", false)]
+        [InlineData("Caption", true)]
         [Theory]
-        public async Task ParameterWithLocalizableName_StringLiteralArgument_Method_Diagnostic(string parameterName)
+        public async Task ParameterWithLocalizableName_StringLiteralArgument_Method_Diagnostic(string parameterName, bool useNameHeuristic)
         {
-            await VerifyCS.VerifyAnalyzerAsync($@"
+            const string editorConfigText = "dotnet_code_quality.CA1303.use_naming_heuristic = true";
```
mavasani: Shouldn't this be:
```suggestion
            const string editorConfigText = $"dotnet_code_quality.CA1303.use_naming_heuristic = {useNameHeuristic}";
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotPassLiteralsAsLocalizedParametersTests.cs
```diff
@@ -1281,10 +1323,25 @@ public void M1(C c)
...
 }}
 ",
-            // Test0.cs(14,13): warning CA1303: Method 'void Test.M1(C c)' passes a literal string as parameter 'param' of a call to 'void C.M(string param)'. Retrieve the following string(s) from a resource table instead: "a".
-            GetCSharpResultAt(14, 13, "void Test.M1(C c)", parameterName, $"void C.M(string {parameterName})", "a"));
+                    },
+                },
+            };
 
-            await VerifyVB.VerifyAnalyzerAsync($@"
+            if (useNameHeuristic)
```
mavasani: Ah, you are relying on the default value being `false` here? Works for me.

Evangelink: Yes sorry I need to find a way to make it more explicit.

Evangelink:
@mavasani I have rewritten the tests that weren't obvious to read/understand. Let me know if the logic stills confuses you. I am quite happy with how they look now.

mavasani:
@Evangelink Can you resolve conflicts and push the changes to tests that you mention in your last comment?

Evangelink:
@mavasani The doc is going to be merged, we should make sure to have the PR merged too.

PR #3430, opened 2020-03-25
Scripts for copying localized strings from master to 2.9.x

dotpaul:
Now after a localization pass in master, we should be able to run the `CopyAllResourceStrings.ps1` script to pick up the changes in the 2.9.x branch.

Here's its output (for making sure there aren't any breaking changes with strings containing message arguments):
```
CollectionsShouldImplementGenericInterfaceMessage
  Old: Collection '{0}' directly or indirectly inherits '{1}' without implementing '{2}'. Publicly-visible collections should implement the generic version to broaden usability.
  New: Type '{0}' directly or indirectly inherits '{1}' without implementing '{2}'. Publicly-visible types should implement the generic version to broaden usability.
MarkMembersAsStaticMessage
  Old: Member {0} does not access instance data and can be marked as static (Shared in VisualBasic)
  New: Member '{0}' does not access instance data and can be marked as static
```


dotpaul:
On second thought, I should pick up the English strings too, in case they've changed in master.

dotpaul:
Thank you @mavasani !

PR #3432, opened 2020-03-26
Add DoNotUseStackallocInLoopsAnalyzer

stephentoub:
Finds stackallocs used within loops and warns that such stackallocs should be moved out of the loop.

Fixes https://github.com/dotnet/runtime/issues/33782
cc: @mavasani, @sharwell, @bartonjs, @jeffhandley


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private static readonly ImmutableArray<SyntaxKind> s_stackallocKinds = ImmutableArray.Create(SyntaxKind.StackAllocArrayCreationExpression);
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
```
Evangelink: Out of curiosity, I have seen this pattern quite often but I am not sure to understand what the point when there is no access to some user-option or known type, could you let me know if there is a special reason?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                compilationContext.RegisterSyntaxNodeAction(syntaxContext =>
```
Evangelink: Can't you do that fully with operations instead of node and then accessing to the operations?

stephentoub: There's no operation for a stackalloc today.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
stephentoub: I'm not understanding the question. What pattern are you referring to? What alternative are you expecting?

Evangelink: My bad, I meant using the `RegisterCompilationStartAction` before the `RegisterSyntaxNodeAction` without doing anything in between while you could register the syntax node action directly.

stephentoub: Ah, no reason, other than I started with as a template another rule that did this.  I can change it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+                    private static void DoWhile() {
+                        do
+                        {
+                            byte* ptr = stackalloc byte[1024];
+                        }
+                        while (true);
+                    }
+                }",
+                VerifyCS.Diagnostic().WithSpan(7, 41, 7, 62),
```
mavasani: For future reference, you can also directly add diagnostic markup inline in source, such as say `{|CA2014:stackalloc byte[1024]|}`

stephentoub: Ok, I'll change these.

PR #3439, opened 2020-03-27
Couple of bug fixes to ReleaseTrackingAnalyzer

mavasani:
1. Use quoted `<Undetected>` instead of unquoted <Undetected> for undetected fields in AnalyzerReleases file - this renders properly in md file viewers.
2. Rename the optional "HelpLink" column to "Notes" column in AnalyzerReleases file and also include the analyzer name in the Notes generated by the code fix.


mavasani:
Thanks @dotpaul

PR #3440, opened 2020-03-27
Add a new diagnostic/code fix to prompt analyzer authors to enable re…

mavasani:
…lease tracking analyzers

EnableReleaseTrackingAnalyzer

Also updated the documentation accordingly.


nuget/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.targets
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <!-- Workaround for https://github.com/dotnet/roslyn/issues/4655 -->
+  <ItemGroup Condition="Exists('$(MSBuildProjectDirectory)\AnalyzerReleases.Shipped.md')" >
+	  <AdditionalFiles Include="AnalyzerReleases.Shipped.md" />
+  </ItemGroup>
+  <ItemGroup Condition="Exists('$(MSBuildProjectDirectory)\AnalyzerReleases.Unshipped.md')" >
+	  <AdditionalFiles Include="AnalyzerReleases.Unshipped.md" />
+  </ItemGroup>
```
sharwell: We should just leave this out. Instead of having the targets file, put a comment at the top of the generated file indicating that the user needs to manually include it in AdditionalFiles.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/ReleaseTrackingAnalyzerTests.cs
```diff
@@ -55,6 +55,70 @@ class MyAnalyzer : DiagnosticAnalyzer
...
+                        (DiagnosticDescriptorCreationAnalyzer.ShippedFileName,
+                            AnalyzerReleaseTrackingFix.ShippedAnalyzerReleaseTrackingFileDefaultContent),
+                        (DiagnosticDescriptorCreationAnalyzer.UnshippedFileName,
+                            AnalyzerReleaseTrackingFix.UnshippedAnalyzerReleaseTrackingFileDefaultContent)
+                    },
+                    ExpectedDiagnostics =
+                    {
+                        new DiagnosticResult(DiagnosticDescriptorCreationAnalyzer.DeclareDiagnosticIdInAnalyzerReleaseRule)
+                            .WithArguments("Id1")
+                            .WithLocation(11, 34),
```
sharwell: 📝 Always use `WithSpan` instead of `WithLocation`

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/ReleaseTrackingAnalyzerTests.cs
```diff
@@ -55,6 +55,70 @@ class MyAnalyzer : DiagnosticAnalyzer
...
+    private static readonly DiagnosticDescriptor descriptor1 =
+        new DiagnosticDescriptor(""Id1"", ""Title1"", ""Message1"", ""Category1"", DiagnosticSeverity.Warning, isEnabledByDefault: true);
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(descriptor1);
+    public override void Initialize(AnalysisContext context) { }
+}";
+            var test = new CSharpCodeFixVerifier<DiagnosticDescriptorCreationAnalyzer, AnalyzerReleaseTrackingFix>.Test()
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.Default,
+                CodeFixTestBehaviors = CodeFixTestBehaviors.FixOne | CodeFixTestBehaviors.SkipFixAllCheck,
```
sharwell: ❓ Why are we doing this? If this is a case of having consecutive code fixes getting applied, then the expected way to address that is setting `CodeActionIndex` and `CodeActionEquivalenceKey` to only consider the behavior of one specific fix.

mavasani: Ah, thanks - I was just confused and did not investigate enough. This was indeed failing due to missing FixAll support. It has now been fixed with the new commit. Thanks for catching!

nuget/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.targets
mavasani: I prefer that to add this tooling support so the user does not have to grok though extra comment and then perform manual steps to achieve the exact same thing. There is no harm in having an additional targets to achieve this workaround. We can delete it once https://github.com/dotnet/roslyn/issues/4655 is fixed.

mavasani:
Note to self: Need to fix a new RS diagnostic once https://github.com/dotnet/roslyn-analyzers/pull/3444 goes in.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/ReleaseTrackingAnalyzerTests.cs
```diff
@@ -55,6 +55,65 @@ class MyAnalyzer : DiagnosticAnalyzer
...
+class MyAnalyzer : DiagnosticAnalyzer
+{
+    private static readonly DiagnosticDescriptor descriptor1 =
+        new DiagnosticDescriptor({|RS2008:""Id1""|}, ""Title1"", ""Message1"", ""Category1"", DiagnosticSeverity.Warning, isEnabledByDefault: true);
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(descriptor1);
+    public override void Initialize(AnalysisContext context) { }
+}";
+
+            var fixedSource = @"
```
sharwell: ❔ Is this the same as `source`? If so, it would be good to just have one copy.

mavasani: Input source has a diagnostic.

sharwell: If it's a diagnostic the code fix claims to be able to fix, the checks will ignore that markup diagnostic during validation of the fixed state. This is because the default `MarkupHandling` value is `IgnoreFixable`:

https://github.com/dotnet/roslyn-sdk/blob/0408271325bb067fd5f84592a545e8280cedff39/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/SolutionState.cs#L65-L70

mavasani: Ah thanks - I have made this change.

mavasani:
@sharwell Any further feedback?

src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
```diff
@@ -112,13 +114,20 @@ CA6000  |  Design  |  Warning | CA6000_AnalyzerName
...
+This can be done in couple of ways, which is detailed in below sections.
+
+### Using the light bulb
+
+![Enable analyzer release tracking](./enable-analyzer-release-tracking.png)
+
+### Manual steps
 This can be done by:
 
 - In Visual Studio, right click project in Solution Explorer, and choose "Add -> New Items", then select "Text File" in "Add new item" dialog.
```
mavasani:
```suggestion
- In Visual Studio, right-click the project in Solution Explorer, choose "Add -> New Item...", and then select "Text File" in the "Add New Item" dialog. Then right-click each file, select "Properties", and choose "C# analyzer additional file" for "Build Action" in the "Properties" window.
```

PR #3443, opened 2020-03-27
Add PreferTypedStringBuilderAppendOverloads analyzer + fixer

stephentoub:
Fixes https://github.com/dotnet/runtime/issues/33787
cc: @bartonjs, @jeffhandley, @mavasani, @sharwell


mavasani:
FYI: You will need to fix a new RS diagnostic once https://github.com/dotnet/roslyn-analyzers/pull/3444 goes in - should be easy to fix with an IDE code fix.

stephentoub:
> FYI: You will need to fix a new RS diagnostic once #3444 goes in - should be easy to fix with an IDE code fix.

Done.

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
```diff
@@ -39,4 +39,3 @@ RS1030 | Do not invoke Compilation.GetSemanticModel() method within a diagnostic
...
 [RS2006](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) | Remove duplicate entries for diagnostic ID between analyzer releases. | MicrosoftCodeAnalysisReleaseTracking | True | Warning | False | Remove duplicate entries for diagnostic ID between analyzer releases. |
 [RS2007](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) | Invalid entry in analyzer release file. | MicrosoftCodeAnalysisReleaseTracking | True | Warning | False | Invalid entry in analyzer release file. |
-[RS2008](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) | Enable analyzer release tracking | MicrosoftCodeAnalysisReleaseTracking | True | Warning | True | Enabling release tracking for analyzer packages helps in tracking and documenting the analyzer diagnostics that ship and/or change with each analyzer release. See details at https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md. |
```
mavasani: I presume this got reverted through some conflict? Maybe need to re-execute pack?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloads.Fixer.cs
```diff
@@ -0,0 +1,63 @@
...
+            CancellationToken cancellationToken = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode expression)
+            {
+                string title = MicrosoftNetCoreAnalyzersResources.PreferTypedStringBuilderAppendOverloadsRemoveToString;
+                context.RegisterCodeFix(
+                    new MyCodeAction(title,
+                        async ct =>
+                        {
+                            if (await doc.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false) is SemanticModel model &&
```
mavasani: `model` should always be non-null for C# and VB (it can be null for other languages such as TS).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloads.Fixer.cs
```diff
@@ -0,0 +1,63 @@
...
+                                model.GetOperationWalkingUpParentChain(expression, cancellationToken) is IArgumentOperation arg &&
+                                arg.Value is IInvocationOperation invoke &&
+                                invoke.Instance?.Syntax is SyntaxNode replacement)
+                            {
+                                DocumentEditor editor = await DocumentEditor.CreateAsync(doc, ct).ConfigureAwait(false);
+                                editor.ReplaceNode(expression, new SyntaxEditor(root, doc.Project.Solution.Workspace).Generator.Argument(replacement));
```
mavasani:
```suggestion
                                editor.ReplaceNode(expression, editor.Generator.Argument(replacement));
```

stephentoub: Oh, cool, I didn't realize DocumentEditor derives from SyntaxEditor.

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
stephentoub: Hmm, not sure.  I'll rebase it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
```diff
@@ -0,0 +1,273 @@
...
+        [InlineData("string", true)]
+        [InlineData("String", true)]
+        [InlineData("ulong", true)]
+        [InlineData("object", false)]
+        [InlineData("char[]", false)]
+        [InlineData("DateTime", false)]
+        [InlineData("DayOfWeek", false)]
+        public async Task ArgumentIsToStringMethodCallOnLocal_CSharp(string receiverType, bool diagnosticExcepted)
+        {
+            DiagnosticResult[] diagnostics = diagnosticExcepted ?
```
mavasani: For such a test, using markup should significantly simplify the test. Your input would be strings such as `[|int|]`, `[|string|]`, etc. for diagnostic cases and just `object`, `char[]`, etc. for non-diagnostic cases.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
```diff
@@ -0,0 +1,273 @@
...
+        [InlineData("int", true)]
+        [InlineData("Int32", true)]
+        [InlineData("string", true)]
+        [InlineData("String", true)]
+        [InlineData("ulong", true)]
+        [InlineData("object", false)]
+        [InlineData("char[]", false)]
+        [InlineData("DateTime", false)]
+        [InlineData("DayOfWeek", false)]
+        public async Task ArgumentIsToStringMethodCallOnLocal_CSharp(string receiverType, bool diagnosticExcepted)
```
bartonjs:
```suggestion
        public async Task ArgumentIsToStringMethodCallOnLocal_CSharp(string receiverType, bool diagnosticExpected)
```

bartonjs: (And the other occurrences)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
stephentoub: That makes sense.  Ok.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
```diff
@@ -0,0 +1,273 @@
...
+        [Fact]
+        public async Task CSharpFixer()
+        {
+            await VerifyCS.VerifyCodeFixAsync(@"
+                using System;
+                using System.Text;
+
+                class C
+                {
+                    public void M1() => new StringBuilder().Append([|42.ToString()|]);
```
bartonjs: Should the squiggle be under the whole argument, or just the `.ToString()` part?  I'd put it just under ToString() since removing those characters is the recovery action.  Consider it as a complex flow:

```C#
public void M1() =>
    new StringBuilder.Append(
        new OtherThing
        {
            PropA = ...,
            PropB = ...,
            PropC = ...,
            PropD = ...,
        }.GetHashCode().ToString());
```

It doesn't need multiple lines of squiggle here.

stephentoub: While this might be a lame argument, I did it under the whole expression as it makes the fixer simpler / easier to get right (and there didn't seem to be a significant usability benefit to just having the squiggle under the ToString()).  The fixer is then passed the whole relevant expression, rather than part of the expression it needs to walk up from to figure out the relevant part.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
bartonjs: Wouldn't that markup expect the error at the variable declaration, instead of where ToString is being called?

mavasani: For future PRs where you have both an analyzer and a fixer and the fixer can fix all diagnostics generated by the analyzer, you don't need to write separate analyzer and fixer tests - just writing the fixer tests for all scenarios should suffice. @sharwell has been strongly pushing for this approach, it does make the set of tests more compact.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
bartonjs: So long as it was a conscious choice :)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
stephentoub: > Wouldn't that markup expect the error at the variable declaration, instead of where ToString is being called?

Actually, yeah, @mavasani, won't that get the span wrong?

mavasani: > Wouldn't that markup expect the error at the variable declaration, instead of where ToString is being called?

Yes, you are correct. You'd still need a separate parameter, but the test would be simpler and avoid magic numbers for diagnostic line/column numbers and spans.

stephentoub: > You'd still need a separate parameter, but the test would be simpler and avoid magic numbers for diagnostic line/column numbers and spans.

How would it avoid magic numbers?  What are you proposing the separate argument be?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
```diff
@@ -0,0 +1,273 @@
...
+            await VerifyCS.VerifyAnalyzerAsync(@"
+                using System;
+                using System.Text;
+
+                class C
+                {
+                    public void M()
+                    {
+                        var sb = new StringBuilder();
+                        sb.AppendLine(42.ToString());
```
bartonjs: Is it correct/worth having this analyzer/fixer report/replace this pattern with `sb.Append(42).AppendLine()`?

Disregarding other pros/cons of the fluent builder pattern, it does happen to work here as a "one expression for another without side effects" solution.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
mavasani:
```csharp
        [Theory]
        [InlineData("int", true)]
        [InlineData("Int32", true)]
        [InlineData("string", true)]
        [InlineData("String", true)]
        [InlineData("ulong", true)]
        [InlineData("object", false)]
        [InlineData("char[]", false)]
        [InlineData("DateTime", false)]
        [InlineData("DayOfWeek", false)]
        public async Task ArgumentIsToStringMethodCallOnLocal_CSharp(string receiverType, bool diagnosticExcepted)
        {
            var arg = diagnosticExcepted ? "[|value.ToString()|]" : "value.ToString()";
            await VerifyCS.VerifyAnalyzerAsync($@"
                using System;
                using System.Text;
                class C
                {{
                    public void M()
                    {{
                        {receiverType} value = default;
                        var sb = new StringBuilder();
                        sb.Append({arg});
                        sb.Insert(42, {arg});
                    }}
                }}", diagnostics);
        }
```

stephentoub: Ok

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloads.cs
```diff
@@ -0,0 +1,121 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>CA1830: Prefer strongly-typed StringBuilder.Append overloads.</summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class PreferTypedStringBuilderAppendOverloads : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1830";
```
mavasani: Once you merge latest, you would now need to bump up the range end/next available ID: https://github.com/dotnet/roslyn-analyzers/blob/78dd233d304b93b5d430b0a96fafe83f8344d932/src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt#L15

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
stephentoub: Interesting.  That seems reasonable.  I'd like to look at it separately.

carlossanlop:
I have a question about this PR, @stephentoub:

The [original issue](https://github.com/dotnet/runtime/issues/33787) was categorized by Immo as "Performance", but this PR adds the analyzer inside the Runtime folder instead of the Performance folder. Why is that?

- Microsoft.CodeAnalysis.NetAnalyzers project
    - Microsoft.NetCore.Analyzers folder
        - Runtime <-
        - Performance

I have [this issue](https://github.com/dotnet/runtime/issues/33790) assigned to me to create an analyzer, and was also categorized as "Performance", so I'd like to know how to decide in which folder should I create my analyzer.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs
```diff
@@ -0,0 +1,264 @@
...
+        [Theory]
+        [InlineData("int", true)]
+        [InlineData("Int32", true)]
+        [InlineData("string", true)]
+        [InlineData("String", true)]
+        [InlineData("ulong", true)]
+        [InlineData("object", false)]
+        [InlineData("char[]", false)]
+        [InlineData("DateTime", false)]
+        [InlineData("DayOfWeek", false)]
```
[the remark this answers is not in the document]
stephentoub: Yup.

stephentoub:
> The original issue was categorized by Immo as "Performance", but this PR adds the analyzer inside the Runtime folder instead of the Performance folder. Why is that?

The categorization that's important is this one:
https://github.com/dotnet/roslyn-analyzers/pull/3443/files#diff-d6a8b8e17ba2b3ecec87bd042b598259R22
The folder isn't publicly visible for folks consuming the analyzer.  I put it in the runtime folder because the general organization of this project is generally around types rather than purpose.  But we can always move things between folders as needed.  @mavasani, any further advice?

mavasani:
Agree with @stephentoub - the important bit is the category set in the descriptor. Ideally, the folder structure will match the rule category, but it has been out of sync for a while. We have wanted to correct it, but never got to prioritize that work. We can even consider flattening the files into a single folder, i.e. put all analyzers under the folder https://github.com/dotnet/roslyn-analyzers/tree/master/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers and remove all sub-folders.

PR #3446, opened 2020-03-28
polish and fix VS instructions for PublicApiAnalyzers

adamralph:


mavasani:
I think we should just mimic the approach taken [here](https://github.com/dotnet/roslyn-analyzers/pull/3440/) - generate a new diagnostic if a project references PublicApiAnalyzers, but does not have shipped and unshipped file, and the code fix should just [drop a targets files](https://github.com/dotnet/roslyn-analyzers/pull/3440/files#diff-7a21e67f11f88594ca8603735b75d332R1-R8) so user requires no manual step. I plan to implement this as soon as https://github.com/dotnet/roslyn-analyzers/pull/3440 goes in.

PR #3454, opened 2020-03-30
CA1046: Do not overload operator equals on reference types

### LINKED ISSUE #392: Port FxCop rule CA1046: DoNotOverloadOperatorEqualsOnReferenceTypes
**Title:** Do not overload operator equals on reference types

**Description:**

For reference types, the default implementation of the equality operator is almost always correct. By default, two references are equal only if they point to the same object.

**Dependency:** None

**Notes:**

Evangelink:
Fix #392


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypesTests.cs
```diff
@@ -0,0 +1,196 @@
...
+            // Doc states that if type behaves as a value-type and has addition/subtraction it might be safe.
+
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public class C
+{
+    public static bool operator [|==|](C left, C right) => true;
+    public static bool operator !=(C left, C right) => true;
+
+    public static C operator +(C left, C right) => left;
+    public static C operator -(C left, C right) => left;
```
Evangelink: I kept the legacy implementation but we could have some FPs here so it might be even better to always ignore cases where there is an add/remove operator, wdyt?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypes.cs
```diff
@@ -0,0 +1,60 @@
...
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotOverloadOperatorEqualsOnReferenceTypesTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotOverloadOperatorEqualsOnReferenceTypesMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.DoNotOverloadOperatorEqualsOnReferenceTypesDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
+        public static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Design,
+            RuleLevel.IdeSuggestion,
```
mavasani, line 25: @sharwell @stephentoub Any concerns in having this rule enabled as an IDE suggestion? Seems the rule will be correct majority of the times..

stephentoub: There are plenty of places where someone overrides Equals on reference types.  Seems like this could have a lot of false positives.  We would definitely disable this in dotnet/runtime (e.g. Version.Equals), presumably in dotnet/roslyn (e.g. Symbol.Equals), presumably in dotnet/aspnetcore (e.g. BindingSource.Equals), etc.  Just doing a regex search for `class .* IEquatable<` to get a sense for how many reference types implement the interface (the number that override Equals is almost certainly more):
https://grep.app/search?q=class%20.%2A%20IEquatable%3C&regexp=true&case=true&filter[lang][0]=C%23

stephentoub: @bartonjs, what's our current design guideline around this?

bartonjs: * CONSIDER overriding Equals to provide value equality if a reference type represents a value. For example, you might want to override Equals in reference types representing numbers or other mathematical entities.
* DO NOT implement value equality on mutable reference types.

So it's right in the latter case, wrong in the former.  Maybe if it can reliably detect mutability?

Of course, that's for public types.  For private types you could want to match on a small number of fields and then have mutable hangers-on, like linked list pointers.

Evangelink: @stephentoub Just in case, this is complaining about operator equal not method equals.

stephentoub: > this is complaining about operator equal not method equals.

Ah, I misread the rule name, thanks.  Even so, providing an operator== is reasonable when you override Equals.

bartonjs: Nevermind, that's for .Equals.  This is op_Equality.

* AVOID overloading equality operators on mutable reference types.
* AVOID overloading equality operators on reference types if the implementation would be significantly slower than that of reference equality.

I'm not exactly sure why we're softer on the operator than the method; apparently those sections are just far enough apart that it's been a cache-miss for 3 editions of the guidelines :)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1391,6 +1391,15 @@
...
     <value>Do not name enum values 'Reserved'</value>
   </data>
+  <data name="DoNotOverloadOperatorEqualsOnReferenceTypesDescription" xml:space="preserve">
+    <value>For reference types, the default implementation of the equality operator is almost always correct. By default, two references are equal only if they point to the same object.</value>
+  </data>
+  <data name="DoNotOverloadOperatorEqualsOnReferenceTypesMessage" xml:space="preserve">
+    <value>'{0}' should not overload the equality operator.</value>
+  </data>
+  <data name="DoNotOverloadOperatorEqualsOnReferenceTypesTitle" xml:space="preserve">
+    <value>Do not overload operator '==' on reference types</value>
```
bartonjs, line 1401: A quick search shows that the operator is named `=` in VB, and `==` in C#; so this message isn't language-agnostic.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypes.cs
```diff
@@ -0,0 +1,60 @@
...
+                    return;
+                }
+
+                // FxCop compat: only analyze externally visible symbols by default.
+                if (!method.MatchesConfiguredVisibility(context.Options, Rule, context.CancellationToken))
+                {
+                    return;
+                }
+
+                context.ReportDiagnostic(method.CreateDiagnostic(Rule, method.ContainingType.Name));
```
bartonjs, line 56: There's a CONSIDER rule for overriding op_Equality for anything that implements `IEquatable<TheirOwnType>`.

CONSIDER not reporting a diagnostic in such cases :smile:.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: I might be wrong but from what I have seen the messages do not seem to be language agnostic. @mavasani What's the "rule"?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypes.cs
Evangelink: > AVOID overloading equality operators on mutable reference types.

I guess if any field isn't `readonly` or any property has a setter we could assume this is a mutable type.

> AVOID overloading equality operators on reference types if the implementation would be significantly slower than that of reference equality.

I don't really see how we could do so...

> There's a CONSIDER rule for overriding op_Equality for anything that implements IEquatable<TheirOwnType>.

Maybe the best is to say report on classes except if it implements `IEquatable`, how does it sound to you?

mavasani: @Evangelink, I think it would be best to start with this rule as disabled by default. We can move the discussion around whether it would be suitable to be enabled as an IDE suggestion by default to a separate issue.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: We should try to keep the messages language agnostic - our existing messages which were ported from original FxCop implementation may violate this at some places, but that you should not encourage us to add more of these. @Evangelink - can you file a tracking issue for all messages that you came across that were not language agnostic?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypes.cs
mavasani: +1

Evangelink: @bartonjs I might be missing something but the only guideline I found suggesting to override operator == on `IEquatable` was for value types. In here this is analyzing only reference type. Do you have any doc?

bartonjs: _Framework Design Guidelines, 2nd ed._, 8.6 (IComparable\<T\> and IEquatable\<T\>), pp. 264

It does seem to be missing from the guidelines published to MSDN/docs.microsoft.com; when 3rd edition releases (later this year) we'll be updating the "MSDN" guidelines-only version to match.

Evangelink: Thank you!

Evangelink:
Ping @mavasani

Evangelink:
@stephentoub @bartonjs Could one of you review the PR? @mavasani thinks you would be more suited to make this PR move forward.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypesTests.cs
bartonjs: If the type is behaving as a number it should be IComparable(Of Itself) and IEquatable(Of Itself).  Since IEquatable will make the warning go away, I don't think there needs to be any special treatment for these operators.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypesTests.cs
```diff
@@ -0,0 +1,227 @@
...
+                TestState =
+                {
+                    Sources =
+                    {
+                        $@"
+public class OuterClass
+{{
+    {accessibility} class C
+    {{
+        public static bool operator [|==|](C left, C right) => true;
```
bartonjs, line 135: It would be good to see the inverse test: That internal doesn't warn without the editorconfig.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1418,6 +1418,15 @@
   <data name="DoNotDeclareEventFieldsAsVirtualTitle" xml:space="preserve">
     <value>Do not declare event fields as virtual</value>
   </data>
+  <data name="DoNotOverloadOperatorEqualsOnReferenceTypesDescription" xml:space="preserve">
+    <value>For reference types, the default implementation of the equality operator is almost always correct. By default, two references are equal only if they point to the same object.</value>
```
bartonjs, line 1422: Do we need to mention the alternative to deleting it?

```suggestion
    <value>For reference types, the default implementation of the equality operator is almost always correct. By default, two references are equal only if they point to the same object. If the operator is providing meaningful value equality, the type should implement the generic IEquatable interface.</value>
```

bartonjs:
The logic LGTM, but it could do with the test to show that the editorconfig option is respected to ignore internal/private types.

@mavasani may have an opinion on the proposed string change.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypesTests.cs
Evangelink: That's ok for me.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypesTests.cs
Evangelink: Done

Evangelink:
@bartonjs Do you think it's worth updating the [CA1046 documentation page](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1046?view=vs-2019) with your comments regarding `IEquatable` and `IComparable`?

Evangelink:
@mavasani PR ready for merge

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypesTests.cs
```diff
@@ -0,0 +1,264 @@
...
+        [InlineData("internal", "dotnet_code_quality.Design.api_surface = all")]
+        // General + Specific analyzer option
+        [InlineData("internal", @"dotnet_code_quality.api_surface = private
+                                  dotnet_code_quality.CA1046.api_surface = all")]
+        // Case-insensitive analyzer option
+        [InlineData("internal", "DOTNET_code_quality.CA1046.API_SURFACE = ALL")]
+        // Invalid analyzer option ignored
+        [InlineData("internal", @"dotnet_code_quality.api_surface = all
+                                  dotnet_code_quality.CA1046.api_surface_2 = private")]
+        public async Task CSharp_ApiSurfaceOption(string accessibility, string editorConfigText)
```
mavasani, line 122: For future reference, I don’t think we need such a huge matrix of tests for api_surface option. Just couple of sanity checks for default behavior (fires for public and not for internal) and configuration (can be made to fire for internal with editorconfig) should suffice.

Evangelink: Ok thank you for the info! That will save me some keystrokes!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotOverloadOperatorEqualsOnReferenceTypes.cs
```diff
@@ -0,0 +1,71 @@
...
+                        return;
+                    }
+
+                    // FxCop compat: only analyze externally visible symbols by default.
+                    if (!method.MatchesConfiguredVisibility(context.Options, Rule, context.Compilation, context.CancellationToken))
+                    {
+                        return;
+                    }
+
+                    context.ReportDiagnostic(method.CreateDiagnostic(Rule, method.ContainingType.Name));
```
mavasani, line 66: Legacy FxCop actually has couple more additional bail out conditions for this rule:

1. It does not report on `IComparable` types
2. It does not report on types that override `System.Object.Equals(object)`

We should do the same given this is a ported FxCop rule.

Evangelink: Out of curiosity, are you accessing the legacy fxcop source code? Is it internal only? When trying to implement such rule it's usually a big pain for me to try and think about all cases that could (or not) lead to an issue being raised or being suppressed.

mavasani: > Is it internal only?

Unfortunately, yes. Feel free to ping me on gitter if you are considering porting any legacy FxCop rule and would like to know about the legacy implementation.

PR #3455, opened 2020-03-30
CA1005: Avoid excessive parameters on generic types

### LINKED ISSUE #371: Port FxCop rule CA1005: AvoidExcessiveParametersOnGenericTypes
**Title:** Avoid excessive parameters on generic types

**Description:**

The more type parameters a generic type contains, the more difficult it is to know and remember what each type parameter represents. It is usually obvious with one type parameter, as in List&lt;T&gt;, and in certain cases that have two type parameters, as in Dictionary&lt;TKey, TValue&gt;. However, if more than two type parameters exist, the difficulty becomes too great for most users.

**Dependency:** None

**Notes:**

Evangelink:
Fix #371


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidExcessiveParametersOnGenericTypes.cs
```diff
@@ -0,0 +1,55 @@
...
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidExcessiveParametersOnGenericTypesTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidExcessiveParametersOnGenericTypesMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AvoidExcessiveParametersOnGenericTypesDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Design,
+            RuleLevel.IdeSuggestion,
```
mavasani: Hmm, I am not sure we should have this as a suggestion by default. I see lot of real world types with more then one parameter. I think this might be better as a disabled diagnostic that anyone with a strong preference on such a design can turn on? @sharwell for thoughts.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidExcessiveParametersOnGenericTypes.cs
```diff
@@ -0,0 +1,55 @@
...
+                var namedType = (INamedTypeSymbol)context.Symbol;
+
+                // FxCop compat: only analyze externally visible symbols by default.
+                if (!namedType.MatchesConfiguredVisibility(context.Options, Rule, context.CancellationToken))
+                {
+                    return;
+                }
+
+                if (namedType.IsGenericType &&
+                    namedType.TypeParameters.Length > 2)
```
mavasani: Should we make this limit configurable?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidExcessiveParametersOnGenericTypes.cs
```diff
@@ -0,0 +1,55 @@
...
+                {
+                    context.ReportDiagnostic(namedType.CreateDiagnostic(Rule, namedType.ToDisplayString()));
```
mavasani:
```suggestion
                    context.ReportDiagnostic(namedType.CreateDiagnostic(Rule, namedType.Name));
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidExcessiveParametersOnGenericTypes.cs
Evangelink: I was thinking about it but at the same time is there any point in having this configurable? I mean maybe based on some new practices/guidelines this number needs to be adjusted but overall I think that either the rule make sense or it doesn't and the adjustable threshold doesn't really change much.

mavasani: Sure, any case we don't need this unless we have user requests...

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1370,6 +1370,15 @@
   <data name="EnumShouldNotHaveDuplicatedValuesTitle" xml:space="preserve">
     <value>Enums values should not be duplicated</value>
   </data>
+  <data name="AvoidExcessiveParametersOnGenericTypesDescription" xml:space="preserve">
+    <value>The more type parameters a generic type contains, the more difficult it is to know and remember what each type parameter represents.</value>
+  </data>
+  <data name="AvoidExcessiveParametersOnGenericTypesMessage" xml:space="preserve">
+    <value>Consider a design where '{0}' has no more than 2 type parameters.</value>
```
mavasani, line 1377: Just in case we decide to make the limit configurable in future.
```suggestion
    <value>Consider a design where '{0}' has no more than {1} type parameters.</value>
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidExcessiveParametersOnGenericTypes.cs
mavasani: Nit: consider defining a special constant field for the limit with value `2`

mavasani:
@Evangelink It is fine to not address the last couple suggestion. Let me know if you planning to address them, I'll wait for your next commit before merging...

Evangelink:
I'll commit in a few minutes :)

mavasani:
@Evangelink seems we have some more merge conflicts...

PR #3459, opened 2020-03-31
Sort WellKnownTypeNames alphabetically

carlossanlop:
No new lines, only sorted the strings. Feel free to double check.
@mavasani Please let me know if this change would cause any problems. I suspect the file is autogenerated and the items are added as they are published. Maybe the order matters?


carlossanlop:
> I think we should retarget this PR to 2.9.x branch to avoid the recurring conflicts for them. The change should flow back into master within a day of being merged into 2.9.x branch.

Thanks for the explanation. Let me close this PR and I'll reopen it for that branch.

PR #3461, opened 2020-03-31
Do not define finalizers for types derived from MemoryManager<T>

buyaa-n:
Fixes dotnet/runtime#33768


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
```diff
@@ -0,0 +1,57 @@
...
+            description: s_localizableDescription,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```
sharwell, line 38: 💭 Seems like this _should_ run on generated code

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
```diff
@@ -0,0 +1,57 @@
...
+            analysisContext.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.NamedType);
+        }
+
+        private static void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
+            var memoryManager1 = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemBuffersMemoryManager1);
```
sharwell, line 45: 💡 This should be obtained just once in a compilation start action. Then register a symbol callback for the case where it is present.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            if (memoryManager1 != null && namedTypeSymbol.DerivesFromOrImplementsAnyConstructionOf(memoryManager1))
+            {
+                var finalizerMethod = namedTypeSymbol.GetMembers().FirstOrDefault(m => m is IMethodSymbol method && method.IsFinalizer());
```
sharwell, line 49: 💡 Consider registering a method callback instead of a named type callback. The `IsFinalizer` check should be extremely simple and fast, and the rest of the code will be cleaner and only run for the rare case of finalizers.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
```diff
@@ -0,0 +1,229 @@
...
+                    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+                    End Sub
+                End Class
+            End Namespace",
+                GetVisualBasicWarningResultAt(9, 45));
+        }
+
+        private DiagnosticResult GetCSharpWarningResultAt(int line, int column)
+            => VerifyCS.Diagnostic(DoNotDefineFinalizersForTypesDerivedFromMemoryManager.Rule)
+                .WithLocation(line, column);
```
sharwell, line 223: 💡 Use `WithSpan` instead of `WithLocation` for all new tests.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
```diff
@@ -0,0 +1,229 @@
...
+                    {
+                        throw new NotImplementedException();
+                    }
+
+                    public override void Unpin() { }
+
+                    protected override void Dispose(bool disposing) { }
+                }
+            }",
+                GetCSharpWarningResultAt(9, 22));
```
sharwell, line 190: 💡 Prefer markup syntax for tests so it's easier to see where the diagnostic is reported in code. If you want to test other specifics of the diagnostics (message, arguments, severity, and/or additional spans), make one test (or a limited few) covering those aspects.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
```diff
@@ -0,0 +1,229 @@
...
+                    End Sub
+
+                    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+                    End Sub
+                End Class
+            End Namespace",
+                GetVisualBasicWarningResultAt(9, 45));
+        }
+
+        private DiagnosticResult GetCSharpWarningResultAt(int line, int column)
```
sharwell, line 221: 💡 In general, new tests do not need the `GetCSharpWarningResultAt` and `GetVisualBasicWarningResultAt` methods. The contents can be inlined for the rare case where this syntax is used.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -7,4 +7,5 @@ CA1069 | Design | Info | EnumShouldNotHaveDuplicatedValues, [Documentation](http
 CA2011 | Reliability | Info | AvoidInfiniteRecursion, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2011)
 CA2012 | Reliability | Hidden | UseValueTasksCorrectlyAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2012)
 CA2013 | Reliability | Warning | DoNotUseReferenceEqualsWithValueTypesAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2013)
-CA2014 | Reliability | Warning | DoNotUseStackallocInLoopsAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2014)
\ No newline at end of file
+CA2014 | Reliability | Warning | DoNotUseStackallocInLoopsAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2014)
+CA2015 | Reliability | Warning | DoNotDefineFinalizersToTypesDerivedFromMemoryManager, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2015)
```
jeffhandley, line 11:
```suggestion
CA2015 | Reliability | Warning | DoNotDefineFinalizersForTypesDerivedFromMemoryManager, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2015)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Linq;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// CA2015: Do not define finalizers to types derived from MemoryManager&lt;T&gt;.
```
jeffhandley, line 13:
```suggestion
    /// CA2015: Do not define finalizers for types derived from MemoryManager&lt;T&gt;.
```

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
mavasani: This entry is auto-generated by the code fix based on the type name used in the analyzer implementation. You may want to also update the analyzer type name here.

buyaa-n: > You may want to also update the analyzer type name here.

Analyzer name changed, forgot this

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
buyaa-n: Seems `GeneratedCodeAnalysisFlags.Analyze` is better suit for that than `ReportDiagnostics` right?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
buyaa-n: Not sure what ` markup syntax` please share some explanation or sample 🙏

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
sharwell: In this case, it would use both flags.

* Analyze means the analyzer should analyze generated code
* ReportDiagnostics means the analyzer driver (e.g. the compiler) should not suppress diagnostics which are reported in generated code

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
sharwell: This is still not resolved. This call should occur in the compilation start action, not in `AnalyzeSymbol`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        private static void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
+            var memoryManager1 = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemBuffersMemoryManager1);
+
+            if (memoryManager1 != null && namedTypeSymbol.DerivesFromOrImplementsAnyConstructionOf(memoryManager1))
+            {
+                namedTypeSymbol.GetMembers().FirstOrDefault(m => CheckMethod(m, context));
```
sharwell, line 58: 💡 This is difficult to read. LINQ is typically used for pure operations (no side effects), but in this case the call to `CheckMethod` is the method that reports diagnostics. Consider using a `foreach` loop here, or replacing `CheckMethod` with something like `TryGetFinalizer` and reporting the diagnostic here if it exists.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
```diff
@@ -0,0 +1,223 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
```
sharwell, line 5: If this analyzer runs on generated code, it will need to use `CSharpSecurityCodeFixVerifier` (which doesn't validate that generated code is excluded).

```suggestion
using VerifyCS = Test.Utilities.CSharpSecurityCodeFixVerifier<
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
```diff
@@ -0,0 +1,223 @@
...
+                    }
+
+                    public override MemoryHandle Pin(int elementIndex = 0)
+                    {
+                        throw new NotImplementedException();
+                    }
+
+                    public override void Unpin() { }
+
+                    ~TestClass() { }
```
sharwell, line 126: 📝 Markup syntax looks like this:

```suggestion
                    [|~TestClass|]() { }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1083,6 +1083,15 @@
...
   </data>
+  <data name="DoNotDefineFinalizersForTypesDerivedFromMemoryManagerDescription" xml:space="preserve">
+    <value>Adding a finalizer for a type derived from MemoryManager&lt;T&gt; creates the possibility that the memory will be freed while it is still in use by a Span&lt;T&gt;.</value>
+  </data>
+  <data name="DoNotDefineFinalizersForTypesDerivedFromMemoryManagerMessage" xml:space="preserve">
+    <value>Adding a finalizer for a type derived from MemoryManager&lt;T&gt; creates the possibility that the memory will be freed while it is still in use by a Span&lt;T&gt;.</value>
+  </data>
+  <data name="DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTitle" xml:space="preserve">
+    <value>Do not define finalizers for types derived from MemoryManager&lt;T&gt;</value>
+  </data>
```
stephentoub, line 1094:
```suggestion
  <data name="DoNotDefineFinalizersForTypesDerivedFromMemoryManagerDescription" xml:space="preserve">
    <value>Adding a finalizer to a type derived from MemoryManager&lt;T&gt; may permit memory to be freed while it is still in use by a Span&lt;T&gt;.</value>
  </data>
  <data name="DoNotDefineFinalizersForTypesDerivedFromMemoryManagerMessage" xml:space="preserve">
    <value>Adding a finalizer to a type derived from MemoryManager&lt;T&gt; may permit memory to be freed while it is still in use by a Span&lt;T&gt;.</value>
  </data>
  <data name="DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTitle" xml:space="preserve">
    <value>Do not define finalizers for types derived from MemoryManager&lt;T&gt;</value>
  </data>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
stephentoub: I agree.  Just using a foreach loop would be easier to read, less code, and less overhead:
```C#
foreach (ISymbol symbol in namedTypeSymbol.GetMembers())
{
    if (symbol is IMethodSymbol method && method.IsFinalizer())
    {
        ... // report diagnostic
        break;
    }
}
```
If you really wanted to use LINQ, you could instead do:
```C#
ISymbol? finalizer = namedTypeSymbol.GetMembers().FirstOrDefault(m => m is IMethodSymbol method && method.IsFinalizer());
if (finalizer != null)
{
    ... // report diagnostic
}
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
jeffhandley: @sharwell Does this markup syntax have the effect of the `VerifyAnalyzerAsync` method automatically checking where the warning would be rendered?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
buyaa-n: Whoops, fixing

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
```diff
@@ -0,0 +1,72 @@
...
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+            analysisContext.RegisterCompilationStartAction(compilationContext =>
```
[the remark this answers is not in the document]
buyaa-n: Isn't it confusing to use `context` for all off them? Especially when they are inside same method?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
buyaa-n: > If you really wanted to use LINQ, you could instead do:
> ISymbol? finalizer = namedTypeSymbol.GetMembers().FirstOrDefault(m => m is IMethodSymbol method && method.IsFinalizer());
> if (finalizer != null)
> {
>     ... // report diagnostic
> }

I was using that at first

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
buyaa-n: Thanks @sharwell, I am wondering about what it does exactly too, as its named markup i guess it just give visual effect for where the errors should reported

mavasani: Yes, the test will fail if the analyzer does not report a diagnostic at the annotated markup span.

mavasani: NOTE: If you have multiple different diagnostic IDs being reported by the analyzer, you can use the markup syntax with prefixed diagnostic ID, like validating `RS2008` below: https://github.com/dotnet/roslyn-analyzers/blob/d02078fe8f2a1051d41b9c5cd6899b2ef6d41649/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/ReleaseTrackingAnalyzerTests.cs#L71

For analyzers that report a single diagnostic ID, the syntax that Sam provided implicitly uses that diagnostic ID for validation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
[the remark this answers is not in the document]
sharwell: It ensures that each use of the variable is correct, because the correct one hides all the incorrect ones from containing scopes.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
sharwell: Yes, it verifies that the diagnostic is reported exactly on the indicated span (as opposed to just requiring that it is within the span.

* If another diagnostic is reported elsewhere, the test will fail
* If the diagnostic is _not_ reported at the marked span, the test will fail
* If the diagnostic does not start _at_ the start of the span, or does not end _at_ the end of the span, the test will fail

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: I have changed from `Adding a finalizer **to** a type derived from` into `Adding a finalizer **for** a type derived from` by @danmosemsft's comment, and it sounded better for me. Not sure if we should update it back, maybe follow the issue wording written by @terrajobst

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
buyaa-n: > Yes, it verifies that the diagnostic is reported exactly on the indicated span (as opposed to just requiring that it is within the span.
> 
> If another diagnostic is reported elsewhere, the test will fail
> If the diagnostic is not reported at the marked span, the test will fail
> If the diagnostic does not start at the start of the span, or does not end at the end of the span, the test will fail

Great, thanks, then I can remove  `.WithSpan(9, 45, 9, 53));`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
stephentoub: Adding x "for" y suggests you can do it externally to y, whereas adding x "to" y suggests it's part of y, which is why I think "to" is better wording here... a finalizer is part of the type, rather than something that can be done externally.

danmosemsft: I agree "to" is better - if I suggested "for", I must have been confused. I can't see my comment for some reason.

jeffhandley: What about "**on** a type"?

@danmosemsft - the comment was on a draft PR in @buyaa-n's fork.  No worries though. 😄

stephentoub: That works, too.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManager.cs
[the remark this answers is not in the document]
jeffhandley: Ah, neat trick -- you're essentially taking the outer context out of scope.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: @danmosemsft comment was for title `Do not define finalizers for types derived from MemoryManager&lt;T&gt;`, my bad that I had changed it for all similar 😛

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotDefineFinalizersForTypesDerivedFromMemoryManagerTests.cs
buyaa-n: Am I missing something, just adding `[||]` around finalizer not working. `[|~TestClass|]() { }`
From error I'm guessing its expecting an argument for the message or so, my warning message not expecting any argument though 
The error:
```
   Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessageException : Context: Diagnostics of test state\r\nMismatch between number of diagnostics returned, expected \"2\" actual \"1\"\r\n\r\nDiagnostics:\r\n// /0/Test0.cs(21,22): warning CA2015: Adding a finalizer to a type derived from MemoryManager<T> may permit memory to be freed while it is still in use by a Span<T>.\r\nVerifyCS.Diagnostic().WithSpan(21, 22, 21, 31).WithArguments(\"Finalize\"),\r\n\r\n\r\nAssert.Equal() Failure\r\nExpected: 2\r\nActual:   1
   at Microsoft.CodeAnalysis.Testing.Verifiers.XUnitVerifier.Equal[T](T expected, T actual, String message) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Testing.Verifiers.XUnit/XUnitVerifier.cs:line 51
```

bartonjs: did you remove the `VerifyCS.Diagnostic(...)` part? If not, it is expecting the diagnostic message twice (once because of the markup, and a second instance because of the manual parameter)

buyaa-n: Oh really, no I only removed `.WithSpan(9, 45, 9, 53)` 😛

PR #3464, opened 2020-04-01
Detect when a range indexer is used on strings or arrays, then immediately implicitly converted to a Span or Memory type.

bartonjs:
# Detects
* CA1381: `ReadOnlySpan<char> slice = str[a..b];`
* CA1382: `ReadOnlySpan<SomeT> slice = arr[a..b];`
* CA1382: `ReadOnlyMemory<SomeT> slice = arr[a..b];`
* CA1383: `Span<SomeT> slice = arr[a..b];`
* CA1383: `Memory<SomeT> slice = arr[a..b];`

as well as using the expression as an argument to an appropriate parameter type.  It would also detect `str[a..b]` to `ReadOnlyMemory<char>`, but there doesn't seem to be an implicit conversion there.

# Does not detect

Explicit conversions:

* `ReadOnlySpan<char> slice = (ReadOnlySpan<char>)str[a..b];`
* `ReadOnlySpan<SomeT> slice = (ReadOnlySpan<SomeT>)arr[a..b];`
* `ReadOnlySpan<SomeT> slice = (Span<SomeT>)arr[a..b];`
* `ReadOnlyMemory<SomeT> slice = (ReadOnlyMemory<SomeT>)arr[a..b];`
* `ReadOnlyMemory<SomeT> slice = (Memory<SomeT>)arr[a..b];`
* `Span<SomeT> slice = (Span<SomeT>)arr[a..b];`
* `Memory<SomeT> slice = (Memory<SomeT>)arr[a..b];`

# Fixes

* CA1381: `ReadOnlySpan<char> slice = str.AsSpan()[a..b];`
* CA1382: `ReadOnlySpan<SomeT> slice = arr.AsSpan()[a..b];`
* CA1382: `ReadOnlyMemory<SomeT> slice = arr.AsMemory()[a..b];`
* CA1383: `Span<SomeT> slice = arr.AsSpan()[a..b];`
* CA1383: `Memory<SomeT> slice = arr.AsMemory()[a..b];`

---

Fixes https://github.com/dotnet/runtime/issues/33829

---

Because the string type is immutable, `ReadOnlySpan<char>` from a string is also immutable, therefore that diagnostic has a very low chance of being a false positive.

For arrays it's possible (but unlikely) that the caller wants the copy to avoid side effects, such as GC lifetime or P/Invokes disagreeing on how ReadOnly the Span was.  Because of that, the array to ReadOnly{Span|Memory} and writable {Span|Memory} are different codes from the string one; and since the read-only version feels like it's _almost_ never wrong, it's yet a third code.


bartonjs:
Right now the tests don't pass on my machine, because it says the fixed code doesn't compile.  If not for that, the change would be ready... so I'm opening the PR to see if it's me, or the repository state.

The fixer could certainly be smarter, and write prettier code.. e.g. str[3..5] => str.AsSpan(3, 2); but I couldn't figure out how to constant fold and it seemed like a nice to have on top, vs the essence of the feature.

bartonjs:
cc @dotnet/dotnet-analyzers

bartonjs:
OK, looks like it's because the tests are using the 3.0.0 version of the compiler package (https://github.com/dotnet/roslyn-analyzers/blob/d02078fe8f2a1051d41b9c5cd6899b2ef6d41649/src/Test.Utilities/Test.Utilities.csproj#L9).  Upgrading it to 3.5.0 (the latest version with no hyphens in it) is now not reporting all of the diagnostics I expect... so now I'm looking into what's up with that.

Thanks for the version pointer, @sharwell.

mavasani:
@bartonjs The core analyzer packages are currently built against Microsoft.CodeAnalysis version 3.0.0 as they do not use any Roslyn APIs from a later version. This enables folks to install the analyzer packages on older compiler/VS. It should be fine to upgrade the test layer to newer Microsoft.CodeAnalysis version, we do this as and when we encounter a scenario such as this one.

mavasani:
@jcouv did something similar for his changes that required the tests to run against latest Microsoft.CodeAnalysis: https://github.com/dotnet/roslyn-analyzers/blob/d02078fe8f2a1051d41b9c5cd6899b2ef6d41649/eng/Versions.props#L24

We should probably rename that property to `MicrosoftCodeAnalysisForTestsVersion` or some such, change it to 3.5.0 and move the test project to that version.

stephentoub:
Why so many diagnostic IDs? Why not just have one? If the goal is just to provide different messages, there are other ways. Similarly, the single rule could be configured for different types.

bartonjs:
> Why so many diagnostic IDs?

That was in my intro.  I believe the string version is "always right" so I have it on as a warning.  Array->ReadOnly as "almost always right" and Array->Writable as "probably right".  Since they had different false positive rates i went with different IDs.  if there're ways to control nuance for the same ID then I'm happy with that model.. I just don't know what it looks like in code or UX.

mavasani:
> Since they had different false positive rates i went with different IDs. if there're ways to control nuance for the same ID then I'm happy with that model.. I just don't know what it looks like in code or UX.

If you want different default severities and/or enabled by default values, using different IDs is the simplest possible approach. We can use [options](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md), but that is primarily used if user wants to control some configuration and is not always discoverable.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseAsSpanInsteadOfRangeIndexer.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var node = root.FindNode(context.Span);
+
+            if (node is null)
+            {
+                return;
+            }
+
+            // The rules are mutually exclusive, so there can't be more than one for the same span:
+            var diagnostic = context.Diagnostics.Single();
```
mavasani, line 43: This is generally not recommended. You could theoretically have another third party analyzer generating the same diagnostic. We have had quite a few bugs in the past where we have customer reports of Single causing an exception.
```suggestion
            var diagnostic = context.Diagnostics.FirstOrDefault();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseAsSpanInsteadOfRangeIndexer.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+                _rootNode = rootNode;
+                _fixer = fixer;
+                EquivalenceKey = ruleId;
+            }
+
+            protected override async Task<Document> GetChangedDocumentAsync(CancellationToken cancellationToken)
+            {
+                var editor = await DocumentEditor.CreateAsync(_document, cancellationToken).ConfigureAwait(false);
+
+                if (_fixer.TrySplitExpression(_rootNode, out var toReplace, out var target, out var arguments))
```
mavasani, line 96: Given that the registered code action can bail out, I'd recommend that you check if TrySplitExpression can succeed prior to invoking `RegisterCodeFix`. Otherwise, light bulb will show up, but the preview/application of the fix will have no changes, which might feel like a bug to the customer.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseAsSpanInsteadOfRangeIndexer.Fixer.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Performance
+{
+    /// <summary>
+    /// CA1831, CA1832, CA1833: Use AsSpan or AsMemory instead of Range-based indexers when appropriate.
+    /// </summary>
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public sealed class CSharpUseAsSpanInsteadOfRangeIndexerFixer : UseAsSpanInsteadOfRangeIndexerFixer
+    {
+        protected override bool TrySplitExpression(
```
mavasani, line 20: I was going to suggest you to use language-agnostic `IOperation` based approach here, but it seems that `input[3..5]` actually generates an `OperationKind.None` node: https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_IFromEndIndexOperation_IRangeOperation.cs#L52

@333fred Do we have a work item tracking the IOperation implementation for this case?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseAsSpanInsteadOfRangeIndexerTests.cs
```diff
@@ -0,0 +1,725 @@
...
+public class TestClass
+{
+    public int TestMethod(" + typeName + @"[] input)
+    {
+        ReadOnlyMemory<" + typeName + @"> tmp = input[3..5];
+        return tmp.Length;
+    }
+}",
+                VerifyCS.Diagnostic(UseAsSpanInsteadOfRangeIndexerAnalyzer.ArrayReadOnlyRule).
+                    WithLocation(8, 32 + typeName.Length).
```
mavasani, line 491: Can we convert all tests to be markup based?

bartonjs: As far as I can tell, markup can't verify the string arguments.  So I have one of each kind of test verifying the arguments, and all of the other tests use markup.

mavasani: Ah got you. Thanks for the explanation. Maybe worth adding a comment so someone in future does not end up converting these to be markup based.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseAsSpanInsteadOfRangeIndexer.cs
```diff
@@ -0,0 +1,211 @@
...
+                    }
+                    else if (operationContext.Operation.Kind == OperationKind.None)
+                    {
+                        // The forward support via the "None" operation kind is only available for C#.
+                        if (operationContext.Compilation.Language != LanguageNames.CSharp)
+                        {
+                            return;
+                        }
+
+                        if (operationContext.Operation.Syntax.RawKind != (int)Microsoft.CodeAnalysis.CSharp.SyntaxKind.ElementAccessExpression)
```
mavasani, line 125: We should now use `Microsoft.CodeAnalysis.CSharp.SyntaxKind` in shared layer. This will cause C# assemblies to be loaded from VB only solution. If you need to access C# specific types, you should change this type to abstract and have the C# sub-type have C# specific logic.

bartonjs: It shouldn't cause an assembly load, since it's an enum (const values are burned in to the caller).  If we're concerned about the ease of creep I could change it to a literal and a comment; which feels better to me than over-engineering a hierarchy here.

eng/Versions.props
```diff
@@ -20,7 +20,7 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
 
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.0.0</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisVersion>3.5.0</MicrosoftCodeAnalysisVersion>
```
mavasani, line 23: This needs to be reverted - this is bumping up the version used to build all analyzer packages in this repo, which is not required given no new API is being consumed. Instead please rename `MicrosoftCodeAnalysisForShippedApisVersion` to `MicrosoftCodeAnalysisForTestsVersion`, update that one to `3.5.0` and then update the relevant test projects to explicit reference `MicrosoftCodeAnalysisForTestsVersion`

bartonjs:
It looks like there's a runtime breaking change in the constructors for Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions between whatever Test.Utilities is building with and 3.5.0, so changing just the one dependency is failing:

```
  Error Message:
   System.TypeInitializationException : The type initializer for 'Test.Utilities
.CodeMetrics.CodeMetricsTestBase' threw an exception.
---- System.MissingMethodException : Method not found: '
Void Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions..ctor(
 Microsoft.CodeAnalysis.OutputKind,
 Boolean,
 System.String,
 System.String,
 System.String,
 System.Collections.Generic.IEnumerable`1<System.String>,
 Microsoft.CodeAnalysis.OptimizationLevel,
 Boolean,
 Boolean,
 System.String,
 System.String,
 System.Collections.Immutable.ImmutableArray`1<Byte>,
 System.Nullable`1<Boolean>,
 Microsoft.CodeAnalysis.Platform,
 Microsoft.CodeAnalysis.ReportDiagnostic,
 Int32,
 System.Collections.Generic.IEnumerable`1<
  System.Collections.Generic.KeyValuePair`2<
    System.String,
    Microsoft.CodeAnalysis.ReportDiagnostic>>,
 Boolean,
 Boolean,
 Microsoft.CodeAnalysis.XmlReferenceResolver,
 Microsoft.CodeAnalysis.SourceReferenceResolver,
 Microsoft.CodeAnalysis.MetadataReferenceResolver,
 Microsoft.CodeAnalysis.AssemblyIdentityComparer,
 Microsoft.CodeAnalysis.StrongNameProvider,
 Boolean,
 Microsoft.CodeAnalysis.MetadataImportOptions,
 Microsoft.CodeAnalysis.CSharp.NullableContextOptions)'.
  Stack Trace:
     at Test.Utilities.CodeMetrics.CodeMetricsTestBase..ctor()
   at Microsoft.CodeAnalysis.CodeMetrics.UnitTests.CodeMetricsComputationTests..
ctor()
----- Inner Stack Trace -----
   at Test.Utilities.CodeMetrics.CodeMetricsTestBase..cctor()
```

... it looks like the namespace for NullableContextOptions changed (there's no CSharp in it anymore).  Trying to figure out how to move forward from here.

bartonjs:
If at first you don't succeed... add reflection?  Testing the workaround now.

mavasani:
@bartonjs - great find! Please do file a bug on Roslyn repo for the breaking change in the constructor, which should get this fixed asap in Roslyn. Thanks!

bartonjs:
It's already known / marked as "can't fix without introducing yet another breaking change".

https://github.com/dotnet/roslyn/issues/39524

src/Test.Utilities/CodeMetricsTestsBase.cs
```diff
@@ -13,7 +15,7 @@ namespace Test.Utilities.CodeMetrics
 {
     public abstract class CodeMetricsTestBase
     {
-        private static readonly CompilationOptions s_CSharpDefaultOptions = new Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
+        private static readonly CompilationOptions s_CSharpDefaultOptions = BuildDefaultCSharpOptions();
```
mavasani, line 18: It will also be fine to move `Test.Utilities.csproj` to a newer Microsoft.CodeAnalysis reference.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseAsSpanInsteadOfRangeIndexer.Fixer.cs
agocke: I seem to remember a substantial amount was implemented, but we may have left it as `OperationKind.None` to prevent considering it "final." I don't remember what still needs to be implemented, if anything.

agocke: I've renamed that issue to clarify that we need to confirm the implemention, not just add tests

PR #3478, opened 2020-04-05
Enabled coverage with deterministic build

MarcoRossignoli:
Test coverage with deterministic build.

@sharwell let me know if update are enough to enable deterministic build or feel free to PR on my branch.
Also I'd like to know if coverage looks correct from your perspective.

On my local it seems to work

coverageFolder

ILSpySample

Repro commands https://github.com/tonerdo/coverlet/issues/363#issuecomment-554003958

cc: @clairernovotny @tmat @tonerdo

### DIFF
--- a/DeterministicBuild.targets
+++ b/DeterministicBuild.targets
@@ -12,7 +12,7 @@ https://github.com/dotnet/sourcelink/issues/572 -->
     <SourceRoot Include="$(NuGetPackageRoot)" />
   </ItemGroup>
 
-  <Target Name="GetPathMap"
+  <Target Name="CoverletGetPathMap"
           DependsOnTargets="InitializeSourceRootMappedPaths"
           Returns="@(_LocalTopLevelSourceRoot)"
           Condition="'$(DeterministicSourcePaths)' == 'true'"
--- a/DeterministicBuild.targets
+++ b/DeterministicBuild.targets
@@ -1,0 +1,24 @@
+<!-- This target must be imported into Directory.Build.targets -->
+<!-- Workaround. Remove once we're on 3.1.300+
+https://github.com/dotnet/sourcelink/issues/572 -->
+<Project>
+  <PropertyGroup>
+    <TargetFrameworkMonikerAssemblyAttributesPath>$([System.IO.Path]::Combine('$(IntermediateOutputPath)','$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)'))</TargetFrameworkMonikerAssemblyAttributesPath>
+  </PropertyGroup>
+  <ItemGroup>
+    <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)"/>
+  </ItemGroup>
+  <ItemGroup>
+    <SourceRoot Include="$(NuGetPackageRoot)" />
+  </ItemGroup>
+
+  <Target Name="GetPathMap"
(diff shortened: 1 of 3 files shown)


sharwell:
I'm seeing some changes in reported coverage that I can't explain yet. Can you send a second pull request that updates coverlet without enabling deterministic build to see if that's the source of the difference?

clairernovotny:
@MarcoRossignoli One thought here: https://github.com/tonerdo/coverlet/compare/master...MarcoRossignoli:deterministicbuild#diff-d0e735819690b729526b502b93107f5eR6

It's possible that the test project may not be deterministic, for other reasons or incompatibilities, but the projects under test are. I think it'll "just work" without the condition in the test project.

Also, may want to rename the target wokround from `GetPathMap` to `CoverletGetPathMap` just to be explicit.

MarcoRossignoli:
> I'm seeing some changes in reported coverage that I can't explain yet.

If you're talking about that 0.31 is possible because every release we update branches alg(to trim roslyn generated onces that our users doesn't like). I'll revert `DeterministicSourcePaths`.

MarcoRossignoli:
@sharwell I updated package/switch let me know.

sharwell:
@MarcoRossignoli it looks like the new package is _more_ accurate, though still lacking compared to e.g. OpenCover. For example, if we consider just **NormalizeStringsToUppercase.cs**:

Master branch:

https://codecov.io/gh/dotnet/roslyn-analyzers/src/51bc5612a8f462c655cc649bf009557ff4e2bb76/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/NormalizeStringsToUppercase.cs

This pull request:

https://codecov.io/gh/dotnet/roslyn-analyzers/src/9c530a286f2462d36fe2bdb0eee98d62bdfedd66/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/NormalizeStringsToUppercase.cs

The main problems still occurring are:

1. It shows coverage is occurring for blank lines.
2. It doesn't accurately show partial coverage for expressions spanning multiple lines. For example, consider the following lines:

    https://github.com/dotnet/roslyn-analyzers/blob/51bc5612a8f462c655cc649bf009557ff4e2bb76/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/NormalizeStringsToUppercase.cs#L60-L62

    If line 60 is showing as partially-covered, then at least one of line 61 or 62 should show as partially- or not-covered. The easiest way would be showing all of 60-62 as partially-covered.

MarcoRossignoli:
@sharwell for "coverage" problem can you fill an issue on coverlet repo pls, better if with repro code so we can improve code coverage?
At the moment my concern is to understand if coverage works with deterministic build.

sharwell:
The new build seems fine. My only reluctance now is the use of an unofficial build.

MarcoRossignoli:
> Once you log in to codecov.io with GitHub, it will work.

🤦‍♂ 

> The new build seems fine. My only reluctance now is the use of an unofficial build.

This build is only for testing, we'll release official package as soon as is ready.

Feel free to close this PR for now.

sharwell:
@MarcoRossignoli I'm looking forward to  the update. I'm also excited by the prospect that the new test support will allow this project to start to cover these edge cases.

MarcoRossignoli:
> I'm also excited by the prospect that the new test support will allow this project to start to cover these edge cases.

Feel free to report issue with some repo(if possible)!

PR #3481, opened 2020-04-07
Consider using const char in Stringbuilder.Append

pgovind:
Fixes https://github.com/dotnet/runtime/issues/33786

Putting it up for an initial review. Once everything looks good, I can duplicate the tests for VB.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            ImmutableArray<Diagnostic> diagnostics = context.Diagnostics;
+            CodeAnalysis.Text.TextSpan diagnosticsSpan = diagnostics.First().Location.SourceSpan;
+
+            // Find the type declaration identified by the diagnostic. Fix only local declarations.
+            IEnumerable<LocalDeclarationStatementSyntax> declarations = root.FindToken(diagnosticsSpan.Start).Parent.AncestorsAndSelf().OfType<LocalDeclarationStatementSyntax>();
```
pgovind, line 34: 1 decision to be made here is if we want to apply the code fix for class fields that are defined as `const string` too(the analyzer already fires). For ex: Look at the `TestClassField` unit test here: https://github.com/dotnet/roslyn-analyzers/pull/3481/files#diff-b1b9befd17d0df8e9eb932ce2b80ea7eR93

I ran into this case in the runtime repo where the automatic code-fix would lead to compile errors for most cases: For ex: The class field (defined as `const string`) could be passed to a method that only accepts a `string`. The code fix would break all such instances.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -0,0 +1,211 @@
...
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringAnalyzer,
+    Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringFixer>;
+
+namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
+{
+    public class PreferConstCharOverConstUnitStringForStringBuilderAppendTests
+    {
+        [Fact]
+        public async Task TestRegularCase()
```
pgovind, line 14: I'll add tests for VB once I get some feedback

pgovind:
@mavasani : When I try to build this PR on the command line(they show up as warnings, not errors, in VS), I get errors like the following:

```
CSC : error AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringAnalyzer' threw an
exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
 [D:\repos\roslyn-analyzers\src\NetAnalyzers\UnitTests\Microsoft.CodeAnalysis.NetAnalyzers.UnitTests.csproj]
```

It looks like some analyzer in the repo is throwing this while building? Any idea why this is happening and/or how I can get more info to fix the bad lines of code?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// Test for single character strings passed in to String.Append
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class PreferConstCharOverConstUnitStringAnalyzer : DiagnosticAnalyzer
+    {
+        public const string RuleId = "CA1831";
```
Evangelink, line 20: [minor] This is usually declared as `internal`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                                      s_localizableTitle,
+                                                                                      s_localizableMessage,
+                                                                                      DiagnosticCategory.Performance,
+                                                                                      RuleLevel.IdeSuggestion,
+                                                                                      s_localizableDescription,
+                                                                                      isPortedFxCopRule: false,
+                                                                                      isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
```
Evangelink, line 34: [cosmetic] `SupportedDiagnostics => ImmutableArray.Create(Rule);` or `SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterSyntaxNodeAction(AnalyzeSymbol, SyntaxKind.InvocationExpression);
```
Evangelink, line 40: It's better to register an invocation operation rather than a node analyzer and then access the symbol.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            InvocationExpressionSyntax invocationExpression = (InvocationExpressionSyntax)context.Node;
+
+            if (!(invocationExpression.Expression is MemberAccessExpressionSyntax memberAccessExpression))
+            {
+                return;
+            }
+
+            // Verify that the current call is to StringBuilder.Append
+            if (memberAccessExpression.Name.ToString() != nameof(StringBuilder.Append))
```
Evangelink, line 53: You shouldn't be doing `Name.ToString()`. Note that when rewriting as an operation analyzer you will have access to the `Name` property.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+
+            // Check that we've called Append on an instance of StringBuilder
+            ISymbol memberSymbol = context.SemanticModel.GetSymbolInfo(memberAccessExpression).Symbol;
+            if (memberSymbol == null)
+            {
+                return;
+            }
+
+            string memberSymbolContainingTypeString = memberSymbol.ContainingType.ToString();
```
Evangelink, line 65: 1. Unused. 
2. You should not check types using the name, instead you should try to retrieve an instance of the type symbol and check for equality. See https://github.com/dotnet/roslyn-analyzers/pull/3387/files#diff-009312362ad767f47f1855e7de22a6c7R38-R43 and https://github.com/dotnet/roslyn-analyzers/pull/3387/files#diff-009312362ad767f47f1855e7de22a6c7R55

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1272,4 +1272,13 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderDescription" xml:space="preserve">
+    <value>Prefer const char over const unit length string in StringBuilder.Append for more performant code</value>
+  </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderMessage" xml:space="preserve">
+    <value>Prefer const char over const unit length string</value>
```
mavasani, line 1279: We normally use a more specific message with specifics about violations so the message is useful even from command line builds. See other examples above.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// Test for single character strings passed in to String.Append
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class PreferConstCharOverConstUnitStringAnalyzer : DiagnosticAnalyzer
```
mavasani, line 18:
```suggestion
    public sealed class PreferConstCharOverConstUnitStringAnalyzer : DiagnosticAnalyzer
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// Test for single character strings passed in to String.Append
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
```
mavasani, line 17:
```suggestion
    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
mavasani: Yes, this should be rewritten as an IOperation analyzer. Please ping me if you need assistance.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        private void TestMethod() 
+        { 
+            StringBuilder sb = new StringBuilder();
+            const char ch = 'a';
+            sb.Append(ch);
+        } 
+    } 
+}";
+
+            await VerifyCS.VerifyCodeFixAsync(input, VerifyCS.Diagnostic(PreferConstCharOverConstUnitStringAnalyzer.Rule).WithLocation(12, 26).WithArguments("ch"), fix);
```
mavasani, line 49: You can avoid passing in explicit diagnostic instance with location and span info for verification by using markup based syntax. For example, see the markup syntax using `[|` and `|]` here: https://github.com/dotnet/roslyn-analyzers/blob/19f9e7ade9c9ec97dfb44841a6ed376df88e4296/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloadsTests.cs#L28

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -0,0 +1,211 @@
...
+    } 
+}";
+
+        [Fact]
+        public async Task TestMultipleDeclarations()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(multipleDeclarations, VerifyCS.Diagnostic(PreferConstCharOverConstUnitStringAnalyzer.Rule).WithLocation(12, 26).WithArguments("ch"), VerifyCS.Diagnostic(PreferConstCharOverConstUnitStringAnalyzer.Rule).WithLocation(12, 36).WithArguments("bb"));
+        }
+
+        private const string classFieldInAppend = @"
```
mavasani, line 75: We normally don't create separate const fields for test snippets. You should include them inline in tests - either define local variable holding the test source or pass the string directly as the argument to Verify invocations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Done. I've seen other PRs where the `OperationAnalysisContext` object has another `RegisterOperationAction` call on it which finally does the real work. Would someone please explain what the `IOperation` is preferred over working on the SyntaxNode directly? Is it a performance concern?

mavasani: Semantic model + syntax tree provides an equivalent semantic view as an IOperation tree (semantic tree). The later is a newer API with primary advantages in space of perf:

1. `Syntax Tree <--> Operation tree` have a many-to-many mapping between nodes, so querying via semantic model + syntax node to fetch semantics is much more expensive then using and navigating the operation tree.
2. Operation tree is generated directly from the bound nodes or the abstract syntax tree that compiler uses as part of compiling code. This means compiler and analyzers share the semantic analysis phase, versus the former approach where coming from syntax, the compiler needs to do additional work to compute requested semantic information.
3. Using IOperation API you can write language-agnostic semantic analyzers that work for all languages that implement the IOperation API (currently C# and VB, but theoretically any language can implement it). SyntaxNode analyzers are generally required when you are doing some kind of syntactic analysis. Most semantic analysis should be implementable with IOperation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Ah, yes, I missed that. Fixed.

pgovind:
I believe this is ready for review again. I rewrote the analyzer and code-fix to using the IOperation based APIs. 

cc @sharwell too

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1272,4 +1272,13 @@
...
   </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderDescription" xml:space="preserve">
+    <value>Prefer const char over const unit length string in StringBuilder.Append for more performant code</value>
+  </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderMessage" xml:space="preserve">
+    <value>Prefer const char over const unit length string in StringBuilder.Append for more performant code</value>
+  </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderTitle" xml:space="preserve">
+    <value>Prefer const char over const unit length string in StringBuilder.Append for more performant code</value>
+  </data>
```
stephentoub, line 1283: I believe we generally make these different, e.g.

Title = "Use StringBuilder.Append(char) when applicable"

Message = "Use StringBuilder.Append(char) instead of StringBuilder(string) when the input is a constant char."

Description = "StringBuilder.Append(char) is more efficient than StringBuilder.Append(string) even when the string is a single character.  When calling Append with a constant, prefer using a constant char rather than a constant string containing one character."

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+        private static void AnalyzeSymbol(OperationAnalysisContext context)
+        {
+            IInvocationOperation invocationOperation = (IInvocationOperation)context.Operation;
+            if (invocationOperation.Arguments.Length < 1)
+            {
+                return;
+            }
+
+            // Check that the object is a StringBuilder
+            if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemTextStringBuilder, out INamedTypeSymbol? stringBuilderType))
```
stephentoub, line 51: In other rules, I've seen this done prior to the callback, and the resulting type then passed into the callback, often by closure.  @sharwell, @mavasani, how expensive is this call?  Is it ok to do it per Invocation callback?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+                return;
+            }
+
+            IMethodSymbol appendStringMethod = stringBuilderType
+                .GetMembers("Append")
+                .OfType<IMethodSymbol>()
+                .WhereAsArray(s =>
+                    s.Parameters.Length == 1 &&
+                    s.Parameters[0].Type.SpecialType == SpecialType.System_String)
+                .FirstOrDefault();
```
stephentoub, line 62: This should definitely be done once and cached prior to the callback.  Example:
https://github.com/dotnet/roslyn-analyzers/blob/1b6fbe0eafcdf4b9f795cbde00365d92e0dfd71f/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferTypedStringBuilderAppendOverloads.cs#L39-L75

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+                return;
+            }
+
+            if (!invocationOperation.TargetMethod.Equals(appendStringMethod))
+            {
+                return;
+            }
+
+            ImmutableArray<IArgumentOperation> arguments = invocationOperation.Arguments;
+            IArgumentOperation firstArgument = arguments[0];
```
stephentoub, line 74: Is there any situation in which Arguments might be empty even though the target method takes one parameter?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                return;
+            }
+
+            ILocalSymbol localArgumentDeclaration = argumentValue.Local;
+
+            // If there are multiple declarations, bail
+            var semanticModel = context.Operation.SemanticModel;
+            var cancellationToken = context.CancellationToken;
+            SyntaxReference declaringSyntaxReference = localArgumentDeclaration.DeclaringSyntaxReferences.First();
```
stephentoub, line 87: Is this guaranteed to always find one, or might First throw?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                if (variableDeclarationOperation.Declarators.Length != 1)
+                {
+                    return;
+                }
+            }
+
+            // Ok, single variable declaration
+            if (localArgumentDeclaration.HasConstantValue && localArgumentDeclaration.ConstantValue is string unitString && unitString.Length == 1)
```
stephentoub, line 110: Is this rule not catching what I'd expect to be the most common case: `Append("s")`?  It's nice that the rule is able to handle declared consts, but in my experience, the far more common case to handle is when the literal passed in is just a single character but with double quotes instead of single quotes.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: I fixed it anyway:) You're likely right that it'd be more performant

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: I'm guarding against that higher up with:

``` csharp
                    if (invocationOperation.Arguments.Length < 1)
                    {
                        return;
                    }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: I can't think of a scenario where it is null, but I changed it to FIrstOrDefault and checking for null now just to be safe.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Yeap, for some reason I decided not to tag that in the analyzer. Fixed now. Opened https://github.com/dotnet/runtime/pull/34843 too to fix all the cases in dotnet/runtime

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+            SeparatedSyntaxList<VariableDeclaratorSyntax> variables = localDeclaration.Variables;
+            foreach (VariableDeclaratorSyntax variable in variables)
+            {
+                Location declarationLocation = variable.GetLocation();
+
+                Optional<object> constant = context.SemanticModel.GetConstantValue(expr);
+                if (constant.HasValue)
+                {
+                    object value = constant.Value;
+                    if (value is string stringValue && stringValue.Length == 1)
```
danmosemsft: Should it flag append of an empty string as well?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+        private static async Task<Document?> HandleVariableDeclarator(IVariableDeclaratorOperation variableDeclaratorOperation, Document doc, SyntaxNode root, CancellationToken cancellationToken)
+        {
+            IVariableDeclarationOperation variableDeclarationOperation = (IVariableDeclarationOperation)variableDeclaratorOperation.Parent;
+            if (variableDeclarationOperation == null)
+            {
+                return null;
+            }
+
+            IVariableDeclarationGroupOperation variableGroupDeclarationOperation = (IVariableDeclarationGroupOperation)variableDeclarationOperation.Parent;
+            if (variableGroupDeclarationOperation.Declarations.Length != 1)
```
carlossanlop, line 78: This is the part that detects if there is more than one variable declared in one line, correct? As in: `string a = "a", b = "b";`

pgovind: It's the next if block. Line 83. In Roslyn, a `VariableDeclaration` statement can have multiple `VariableDeclarators`. I'm not actually sure what it means to have `variableGroupDeclarationOperation.Declarations.Length > 1`, but I have it here just to be safe

pgovind:
cc: @mavasani. This is ready for review when you have some time. Tagging just in case the notification got lost

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: I don't think so. `Empty` is a field in the `string` class. 1) We're not flagging other instances where we have const string fields in classes and 2) Since it's not a `const` var, I'm not sure it's value is available at compile time?

danmosemsft: I guess it would be pretty rare to write `builder.Append("")` and it would have negligible impact anyway.

pgovind: Actually, `builder.Append("")` would get tagged and fixed to builder.Append('') (I think there was an instance of this in dotnet/runtime somewhere). I thought you were talking about builder.Append(string.Empty), which I have not seen in our repos yet :)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                operationContext.ReportDiagnostic(literalArgumentValue.CreateDiagnostic(Rule));
+            }
+        }
+
+        private static void HandleLocalReferenceOperation(ILocalReferenceOperation argumentValue, OperationAnalysisContext operationContext)
+        {
+            ILocalSymbol localArgumentDeclaration = argumentValue.Local;
+
+            // If there are multiple declarations, bail
```
mavasani, line 106: Why are we doing this bailout? I think the entire code block from line 107 to 134 can we removed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            if (literalArgumentValue.ConstantValue.HasValue && literalArgumentValue.ConstantValue.Value is string unitString && unitString.Length == 1)
+            {
+                operationContext.ReportDiagnostic(literalArgumentValue.CreateDiagnostic(Rule));
+            }
+        }
+
+        private static void HandleLocalReferenceOperation(ILocalReferenceOperation argumentValue, OperationAnalysisContext operationContext)
+        {
+            ILocalSymbol localArgumentDeclaration = argumentValue.Local;
```
mavasani, line 104: This is not the declaration site for the local.

```suggestion
            ILocalSymbol localArgumentOperation = argumentValue.Local;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (root.FindNode(context.Span) is SyntaxNode expression)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        title: MicrosoftNetCoreAnalyzersResources.PreferConstCharOverConstUnitStringInStringBuilderMessage,
+                        createChangedDocument: async c =>
+                        {
+                            SemanticModel semanticModel = await doc.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+                            var operation = semanticModel.GetOperationWalkingUpParentChain(expression, cancellationToken);
+                            if (operation is IVariableDeclaratorOperation variableDeclaratorOperation)
```
mavasani, line 37: Is this ever true? From your analyzer code below, you report the diagnostic at the argument of invocation, not at the variable declarator.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+                        equivalenceKey: "PreferConstCharOverConstUnitStringInStringBuilderAppend"),
+                    context.Diagnostics);
+            }
+        }
+
+        private static async Task<Document?> HandleStringLiteral(ILiteralOperation argumentLiteral, Document doc, SyntaxNode root, CancellationToken cancellationToken)
+        {
+            if (argumentLiteral.ConstantValue.HasValue && argumentLiteral.ConstantValue.Value is string unitString && unitString.Length == 1)
+            {
+                DocumentEditor editor = await DocumentEditor.CreateAsync(doc, cancellationToken).ConfigureAwait(false);
```
mavasani, line 59: We should register a code fix only if you are certain to reach a code path where document will be changed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+                            return doc;
+                        },
+                        equivalenceKey: "PreferConstCharOverConstUnitStringInStringBuilderAppend"),
+                    context.Diagnostics);
+            }
+        }
+
+        private static async Task<Document?> HandleStringLiteral(ILiteralOperation argumentLiteral, Document doc, SyntaxNode root, CancellationToken cancellationToken)
+        {
+            if (argumentLiteral.ConstantValue.HasValue && argumentLiteral.ConstantValue.Value is string unitString && unitString.Length == 1)
```
mavasani, line 57: Can we create a static helper method in the analyzer which takes a `ConstantValue` parameter that has this check? It can be shared between different cases within the analyzer and also between analyzer and code fixer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
mavasani: Basically, I am suggesting use the approach similar to the one taken here: https://github.com/dotnet/roslyn-analyzers/blob/6b16c9c25464f5f77d33da7a62f803857d4d3d6a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseAsSpanInsteadOfRangeIndexer.Fixer.cs#L51. Have all the logic to find the relevant operation with validations to be done before invoking `RegisterCodeFix`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+                            else if (operation is IArgumentOperation argumentOperation)
+                            {
+                                if (argumentOperation.Value is ILiteralOperation argumentLiteral)
+                                {
+                                    return await HandleStringLiteral(argumentLiteral, doc, root, cancellationToken).ConfigureAwait(false);
+                                }
+                            }
+                            return doc;
+                        },
+                        equivalenceKey: "PreferConstCharOverConstUnitStringInStringBuilderAppend"),
```
mavasani, line 50:
```suggestion
                        equivalenceKey: MicrosoftNetCoreAnalyzersResources.PreferConstCharOverConstUnitStringInStringBuilderMessage),
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            CancellationToken cancellationToken = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode expression)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        title: MicrosoftNetCoreAnalyzersResources.PreferConstCharOverConstUnitStringInStringBuilderMessage,
```
mavasani, line 32:
```suggestion
                        title: MicrosoftNetCoreAnalyzersResources.PreferConstCharOverConstUnitStringInStringBuilderTitle,
```

mavasani: Ah, I see that your diagnostic message is also generic, i.e. not specific to each flagged call site. It is fine to continue using  `PreferConstCharOverConstUnitStringInStringBuilderMessage` for your case.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            if (variableDeclarationOperation.Declarators.Length != 1)
+            {
+                return null;
+            }
+
+            DocumentEditor editor = await DocumentEditor.CreateAsync(doc, cancellationToken).ConfigureAwait(false);
+            SyntaxGenerator generator = editor.Generator;
+            ILocalSymbol currentSymbol = variableDeclaratorOperation.Symbol;
+            IVariableInitializerOperation variableInitializerOperation = OperationExtensions.GetVariableInitializer(variableDeclaratorOperation);
```
mavasani, line 91:
```suggestion
            var variableInitializerOperation = variableDeclaratorOperation.GetVariableInitializer();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
mavasani: Okay, so looking at the analyzer and code fix code it seems you have multiple declarations bailout as you want to avoid dealing with the complications of updating that code on the code fixer site. It is fine to not handle this case in the code fix, but I think you should still be flagging this case in the analyzer. I would recommend removing line 107 to 134 completely from the analyzer, but still have the bail out code in the code fixer - for multideclaration case, user has to fix his code manually.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,143 @@
...
+
+                    if (!invocationOperation.TargetMethod.Equals(appendStringMethod))
+                    {
+                        return;
+                    }
+
+                    ImmutableArray<IArgumentOperation> arguments = invocationOperation.Arguments;
+                    IArgumentOperation firstArgument = arguments[0];
+
+                    // We don't handle class fields. Only local declarations
```
mavasani, line 76: I think the analyzer can be simplified quite a bit here:

``` csharp
   if (!TryGetConstChar(arguments[0], out char constChar))
   {
      return;
   }

   if (arguments[0] is ILocalReferenceOperation localReference)
   {
      operationContext.ReportDiagnostic(localReference.Local.CreateDiagnostic(Rule));
   }
   else
   {
      operationContext.ReportDiagnostic(arguments[0].CreateDiagnostic(Rule));
   }
}

// internal so it can be accessed by the code fixer
internal static bool TryGetConstChar(IOperation operation, out char constChar)
{
   if (argument.ConstantValue.HasValue && argument.ConstantValue.Value is string unitString && unitString.Length == 1)
   {
      constChar = unitString[0];
      return true;
   }

   constChar = default;
   return false;
}   
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Ok, fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
pgovind: The `if` blocks here will always succeed because the analyzer verifies everything before calling the fixer like you suggested. I removed the `if` blocks now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
pgovind: With the new changes, this `if` block is used only in the analyzer now. Not sure it makes sense to put it into a helper anymore?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -12,6 +12,7 @@ CA1830 | Performance | Info | PreferTypedStringBuilderAppendOverloads, [Document
 CA1831 | Performance | Warning | UseAsSpanInsteadOfRangeIndexerAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1831)
 CA1832 | Performance | Info | UseAsSpanInsteadOfRangeIndexerAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1832)
 CA1833 | Performance | Info | UseAsSpanInsteadOfRangeIndexerAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1833)
+CA1834 | Performance | Info | PreferConstCharOverConstUnitStringAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1834)
```
Evangelink, line 15: Would be nice to create a ticket on https://github.com/MicrosoftDocs/visualstudio-docs/issues to add the documentation for this analyzer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1287,4 +1287,13 @@
...
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderDescription" xml:space="preserve">
+    <value>StringBuilder.Append(char) is more efficient than StringBuilder.Append(string) when the string is a single character. When calling Append with a constant, prefer using a constant char rather than a constant string containing one character.</value>
+  </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderMessage" xml:space="preserve">
+    <value>Use StringBuilder.Append(char) instead of StringBuilder.Append(string) when the input is a constant unit string.</value>
+  </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderTitle" xml:space="preserve">
+    <value>Consider using StringBuilder.Append(char) when applicable.</value>
```
Evangelink, line 1297: Suggestion: `StringBuilder.Append(char)` -> `'StringBuilder.Append(char)'`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1287,4 +1287,13 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderDescription" xml:space="preserve">
+    <value>StringBuilder.Append(char) is more efficient than StringBuilder.Append(string) when the string is a single character. When calling Append with a constant, prefer using a constant char rather than a constant string containing one character.</value>
+  </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderMessage" xml:space="preserve">
+    <value>Use StringBuilder.Append(char) instead of StringBuilder.Append(string) when the input is a constant unit string.</value>
```
Evangelink, line 1294: Suggestion: `StringBuilder.Append(char)` -> `'StringBuilder.Append(char)'` and `StringBuilder.Append(string)` -> `'StringBuilder.Append(string)'`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1287,4 +1287,13 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferConstCharOverConstUnitStringInStringBuilderDescription" xml:space="preserve">
+    <value>StringBuilder.Append(char) is more efficient than StringBuilder.Append(string) when the string is a single character. When calling Append with a constant, prefer using a constant char rather than a constant string containing one character.</value>
```
Evangelink, line 1291: Suggestion: `StringBuilder.Append(char)` -> `'StringBuilder.Append(char)'` and `StringBuilder.Append(string)` -> `'StringBuilder.Append(string)'`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(PreferConstCharOverConstUnitStringAnalyzer.RuleId);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            CancellationToken cancellationToken = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode expression)
```
Evangelink, line 29: [cosmetic] What about inverting this if to reduce nesting?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                                                                                      isPortedFxCopRule: false,
+                                                                                      isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
```
Evangelink, line 39: [cosmetic] I don't know if this shall be applied or not but I remember @sharwell was suggesting to simply use `context` everywhere.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                // Check that the object is a StringBuilder
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemTextStringBuilder, out INamedTypeSymbol? stringBuilderType))
+                {
+                    return;
+                }
+
+                IMethodSymbol appendStringMethod = stringBuilderType
+                    .GetMembers("Append")
+                    .OfType<IMethodSymbol>()
+                    .WhereAsArray(s =>
```
Evangelink, line 50: Is it worth using the `WhereAsArray` here? We call `FirstOrDefault` right after.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                }
+
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    var invocationOperation = (IInvocationOperation)operationContext.Operation;
+                    if (invocationOperation.Arguments.Length < 1)
+                    {
+                        return;
+                    }
+
```
Evangelink, line 66: [cosmetic] un-needed blank line.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                    else if (firstArgument.Value is ILiteralOperation literalArgumentValue)
+                    {
+                        operationContext.ReportDiagnostic(literalArgumentValue.CreateDiagnostic(Rule));
+                    }
+                }
+                , OperationKind.Invocation);
+            });
+        }
+
+        private static bool IsConstVar(IArgumentOperation argument)
```
Evangelink, line 95: [cosmetic][suggestion] Could be rewritten as a single return.

```csharp
=> argument.Value.ConstantValue.HasValue && argument.Value.ConstantValue.Value is string unitString && unitString.Length == 1
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -0,0 +1,405 @@
...
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringAnalyzer,
+    Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringFixer>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringAnalyzer,
+    Microsoft.NetCore.Analyzers.Runtime.PreferConstCharOverConstUnitStringFixer>;
+
+namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
+{
+    public class PreferConstCharOverConstUnitStringForStringBuilderAppendTests
```
Evangelink, line 14: I don't know if this is a rule but I have seen code-fix tests in separate files.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                // Check that the object is a StringBuilder
```
mavasani, line 41: Nit: Should either remove or update this comment:
```suggestion
                // Check that StringBuilder is defined in this compilation
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
mavasani: Agreed.
```suggestion
                    .FirstOrDefault(s =>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    var invocationOperation = (IInvocationOperation)operationContext.Operation;
+                    if (invocationOperation.Arguments.Length < 1)
+                    {
+                        return;
+                    }
+
+
```
mavasani, line 67:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                    if (!invocationOperation.TargetMethod.Equals(appendStringMethod))
+                    {
+                        return;
+                    }
+
+                    ImmutableArray<IArgumentOperation> arguments = invocationOperation.Arguments;
+                    IArgumentOperation firstArgument = arguments[0];
+
+                    // We don't handle class fields. Only local declarations
```
mavasani, line 76: Why? I think we can handle all constant values. It is fine if you want to report a diagnostic at a different location based on whether a local is used or not, but I think we should be able to flag any operation with a constant value.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                    if (!IsConstVar(firstArgument))
+                    {
+                        return;
+                    }
+
+                    if (firstArgument.Value is ILocalReferenceOperation argumentValue)
+                    {
+                        operationContext.ReportDiagnostic(argumentValue.CreateDiagnostic(Rule));
```
mavasani, line 84: Did we decide to not report this diagnostic on the local declaration? Looks fine to me, but just confirming since your earlier implementation did so. If we are not special casing the diagnostic location for locals, then I don't think you need to have the checks for `ILocalReferenceOperation` or `ILiteralOperation`. You can just write this as:

```csharp
                    if (IsConstVar(firstArgument))
                    {
                       operationContext.ReportDiagnostic(firstArgument.Value.CreateDiagnostic(Rule));
                    }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
mavasani: Agreed, also if this is a private static method, can we make it a static local function?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
mavasani: I would even consider just inlining the conditional expression from `IsConstVar` directly here and remove `IsConstVar` completely.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                SemanticModel semanticModel = await doc.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+                var operation = semanticModel.GetOperationWalkingUpParentChain(expression, cancellationToken);
```
mavasani, line 32: Why do we need to invoke `GetOperationWalkingUpParentChain`? Given you reported the diagnostic on `IArgumentOperation.Value`, can we instead use:

```csharp
var operation = semanticModel.GetOperation(expression, cancellationToken);
if (operation.Parent is IArgumentOperation argumentOperation)
{
```

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
carlossanlop: Thanks @Evangelink.
@mavasani sent us all detailed instructions today on how to document all of our new rules.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
carlossanlop: In my PR, I got the suggestion to consider adding separate strings for VB and for C#. Should we consider this suggestion for all these messages as well?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                    return;
+                }
+
+                IMethodSymbol appendStringMethod = stringBuilderType
+                    .GetMembers("Append")
+                    .OfType<IMethodSymbol>()
+                    .WhereAsArray(s =>
+                        s.Parameters.Length == 1 &&
+                        s.Parameters[0].Type.SpecialType == SpecialType.System_String)
+                    .FirstOrDefault();
```
carlossanlop, line 53: Suggesting removing FirstOrDefault here since it's suggested to call that two lines above:
```suggestion
                        s.Parameters[0].Type.SpecialType == SpecialType.System_String);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+                if (appendStringMethod is null)
+                {
+                    return;
+                }
+
+                compilationContext.RegisterOperationAction(operationContext =>
```
carlossanlop, line 59: I think the same suggestion from @Evangelink applies here about renaming these parameters to `context`:
```suggestion
                compilationContext.RegisterOperationAction(context =>
```

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
carlossanlop: @pgovind I just realized I'm using the same ruleID as you, but you submitted your PR first, so I'll bump up mine by one.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
mavasani: For the code fix, it might be fine to handle just the cases of `ILocalReferenceOperation` and `ILiteralOperation`, and let users manually fix any other case. For the analyzer, we should flag all operations with constant value that satisfy the value check.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
pgovind: I could, but other fixers have been written using the same style. Personally I don't have a strong preference for one or the other, but if the repo owners want it inverted, I'm happy to do it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Hmmm, but there are 2 `context`s though.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Ah, I missed deleting that. Good catch!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: In theory, yes, but this was just a decision we made. The analyzer is more meant to handle instances of `StringBuilder.Append("a")`. Handling local references was an easy addition. The problem with class fields is that we need to track everywhere that it is being used: For ex: There may be a local method that takes in a string, and 1 of it's invocations may be passing in a string class field. A code fix that changes the class field's type from string => char would now be a compiler error.

mavasani: @pgovind My suggestion is just for the analyzer. The fixer can still be more targeted and only handle literal operation and local reference operations. That should not prevent the analyzer from flagging field reference operations so user can manually fix it up.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: Done.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs
pgovind: But there are 2 `context` objects for 2 callbacks?

jeffhandley: Yeah, only the inner most context is appropriate, so by using `context` in each, it prevents accidental usage of the wrong context.

pgovind: Hmmmm, ok. Done.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
mavasani: We don't have any repo requirements here - its a matter of preference. Either way is fine.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+            Document doc = context.Document;
+            CancellationToken cancellationToken = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode expression)
+            {
+                SemanticModel semanticModel = await doc.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+                var operation = semanticModel.GetOperation(expression, cancellationToken);
+                if (operation is IArgumentOperation argumentOperation)
+                {
+                    IOperation? isLocalReferenceOperation = argumentOperation.Value as ILocalReferenceOperation;
```
mavasani, line 35:
```suggestion
                    var isLocalReferenceOperation = argumentOperation.Value as ILocalReferenceOperation;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+                    IOperation? isLiteralOperation = argumentOperation.Value as ILiteralOperation;
```
mavasani, line 36:
```suggestion
                    var isLiteralOperation = argumentOperation.Value as ILiteralOperation;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+                    if (isLocalReferenceOperation == null && isLiteralOperation == null)
+                    {
+                        return;
+                    }
+
+                    IVariableDeclaratorOperation? cachedVariableDeclaratorOperation = default;
+                    if (isLocalReferenceOperation != null)
+                    {
+                        ILocalReferenceOperation localReferenceOperation = (ILocalReferenceOperation)argumentOperation.Value;
```
mavasani, line 45: You will not need this additional cast anymore.

```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+                if (operation is IArgumentOperation argumentOperation)
+                {
+                    IOperation? isLocalReferenceOperation = argumentOperation.Value as ILocalReferenceOperation;
+                    IOperation? isLiteralOperation = argumentOperation.Value as ILiteralOperation;
+                    if (isLocalReferenceOperation == null && isLiteralOperation == null)
+                    {
+                        return;
+                    }
+
+                    IVariableDeclaratorOperation? cachedVariableDeclaratorOperation = default;
```
mavasani, line 42:
```suggestion
                    IVariableDeclaratorOperation? variableDeclaratorOperation = default;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+                    if (isLocalReferenceOperation != null)
+                    {
+                        ILocalReferenceOperation localReferenceOperation = (ILocalReferenceOperation)argumentOperation.Value;
+                        ILocalSymbol localArgumentDeclaration = localReferenceOperation.Local;
+                        SyntaxReference declaringSyntaxReference = localArgumentDeclaration.DeclaringSyntaxReferences.FirstOrDefault();
+                        if (declaringSyntaxReference is null)
+                        {
+                            return;
+                        }
+                        var variableDeclarator = semanticModel.GetOperationWalkingUpParentChain(declaringSyntaxReference.GetSyntax(cancellationToken), cancellationToken);
```
mavasani, line 52:
```suggestion
                        variableDeclaratorOperation = semanticModel.GetOperationWalkingUpParentChain(declaringSyntaxReference.GetSyntax(cancellationToken), cancellationToken) as IVariableDeclaratorOperation;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+
+                        if (variableDeclarator is IVariableDeclaratorOperation variableDeclaratorOperation)
```
mavasani, line 54:
```suggestion
                        if (variableDeclaratorOperation != null)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+                            }
+                        }
+                    }
+
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            title: MicrosoftNetCoreAnalyzersResources.PreferConstCharOverConstUnitStringInStringBuilderTitle,
+                            createChangedDocument: async c =>
+                            {
+                                if (argumentOperation.Value is ILiteralOperation argumentLiteral)
```
mavasani, line 81:
```suggestion
                                if (literalOperation != null)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+                                {
+                                    return await HandleStringLiteral(argumentLiteral, doc, root, cancellationToken).ConfigureAwait(false);
+                                }
+                                else
+                                {
+                                    return await HandleVariableDeclarator(cachedVariableDeclaratorOperation!, doc, root, cancellationToken).ConfigureAwait(false);
```
mavasani, line 87:
```suggestion
                                    RoslynDebug.Assert(variableDeclaratorOperation != null);
                                    return await HandleVariableDeclarator(variableDeclaratorOperation, doc, root, cancellationToken).ConfigureAwait(false);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private static async Task<Document?> HandleVariableDeclarator(IVariableDeclaratorOperation variableDeclaratorOperation, Document doc, SyntaxNode root, CancellationToken cancellationToken)
+        {
+            IVariableDeclarationOperation variableDeclarationOperation = (IVariableDeclarationOperation)variableDeclaratorOperation.Parent;
+            IVariableDeclarationGroupOperation variableGroupDeclarationOperation = (IVariableDeclarationGroupOperation)variableDeclarationOperation.Parent;
+
+            DocumentEditor editor = await DocumentEditor.CreateAsync(doc, cancellationToken).ConfigureAwait(false);
+            SyntaxGenerator generator = editor.Generator;
+            ILocalSymbol currentSymbol = variableDeclaratorOperation.Symbol;
+            var variableInitializerOperation = OperationExtensions.GetVariableInitializer(variableDeclaratorOperation);
```
mavasani, line 115:
```suggestion
            var variableInitializerOperation = variableDeclaratorOperation.GetVariableInitializer();
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -0,0 +1,405 @@
...
+using System.Text;
+ 
+namespace TestNamespace 
+{ 
+    class TestClass 
+    { 
+        private void TestMethod() 
+        { 
+            StringBuilder sb = new StringBuilder();
+            const string ch = ""ab"";
```
mavasani, line 171: Can you add a test with `null` initializer: `const string ch = null;`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -0,0 +1,405 @@
...
+using System.Text;
+ 
+namespace TestNamespace 
+{ 
+    class TestClass 
+    { 
+        private void TestMethod() 
+        { 
+            StringBuilder sb = new StringBuilder();
+            sb.Append([|"",""|]);
```
mavasani, line 279: Can you add no diagnostic tests for:
1. Multi character string: `sb.Append(""ab"");`
2. Null literal: `sb.Append(null);`
3. Interpolated string: `sb.Append($"{someLocal}");`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitString.Fixer.cs
pgovind: Can't do this. There'd be warnings/errors about using pattern matching instead

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
pgovind: Done.

pgovind:
Thanks for the reviews @mavasani  and @Evangelink

PR #3482, opened 2020-04-07
Disable analyzer by default in fxcop

Evangelink:
I have disabled in fxcop all newly added fxcop analyzers (in the unshipped file) for which the state is disabled.

Tagging @mavasani


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/AvoidExcessiveParametersOnGenericTypes.cs
```diff
@@ -26,7 +26,8 @@ public sealed class AvoidExcessiveParametersOnGenericTypes : DiagnosticAnalyzer
             RuleLevel.Disabled,
             s_localizableDescription,
             isPortedFxCopRule: true,
-            isDataflowRule: false);
+            isDataflowRule: false,
+            isEnabledByDefaultInFxCopAnalyzers: false);
```
mavasani, line 30: Thanks - one thing that might help in future is to run `msbuild /m /v:m /t:pack FxCopAnalyzers.sln` locally. This will auto-update the sarif and md files for FxCopAnalyzers package, which can be added as part of the PR that implements the new rule. These changes will show the default severity and enabled by default for the rules in FxCopAnalyzers package.

PR #3483, opened 2020-04-07
Update CA1806 to allow user defined methods

### LINKED ISSUE #3479: CA1806: Allow for user defined use method result
> In the meantime, I think that would be nice to tweak the code to handle all methods from System.Linq.Enumerable.
@mavasani Do you think that's also worth adding a user option to provide more methods?

Yes and yes 👍

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/issues/3363#issuecomment-609792745_

Evangelink:
Fix #3479


docs/Analyzer Configuration.md
```diff
@@ -512,3 +512,25 @@ Option Values: One or more fields of enum [Microsoft.CodeAnalysis.SymbolKind](ht
 Default Value: Namespace, NamedType, Method, Property, Event, Parameter
 
 Example: `dotnet_code_quality.CA1716.analyzed_symbol_kinds = Namespace, Property`
+
+### Additional use results methods
+Option Name: `additional_use_results_methods`
+
+Configurable Rules: [CA1806](https://docs.microsoft.com/en-us/visualstudio/code-quality/CA1806)
+
+Option Values: Names of additional methods (separated by '|') for CA1806.
```
Evangelink, line 521: Do we want to also allow types? Same question for namespaces? IMO types could make sense but I fail to see any point for namespaces.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -140,6 +150,13 @@ public override void Initialize(AnalysisContext analysisContext)
                 INamedTypeSymbol? xunitAssertType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitAssert);
                 INamedTypeSymbol? linqEnumerableType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable);
 
+                if (!(compilationContext.Compilation.SyntaxTrees.FirstOrDefault() is SyntaxTree tree))
```
mavasani, line 153: This approach should only be used if it is not possible to associate a specific syntax tree with the place where we need the option. In this case, that is not true as every operation block action has an associated syntax tree.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/DoNotIgnoreMethodResults.cs
```diff
@@ -140,6 +150,13 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                {
+                    return;
+                }
+
+                var userDefinedMethods = compilationContext.Options.GetAdditionalUseResultsMethodsOption(UserDefinedMethodRule, tree, compilationContext.Compilation, compilationContext.CancellationToken);
```
mavasani, line 158: I suggest moving this inside the `RegisterOperationAction` callback. Our options infrastructure ensures that we compute the option value only once per distinct editorconfig value set - i.e. it will hit the cache for all syntax trees that have the same editorconfig that applies to them.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1436,4 +1436,7 @@
   <data name="EventsShouldNotHaveBeforeOrAfterPrefixTitle" xml:space="preserve">
     <value>Events should not have 'Before' or 'After' prefix</value>
   </data>
+  <data name="DoNotIgnoreMethodResultsMessageUserDefinedMethod" xml:space="preserve">
+    <value>'{0}' calls '{1'} but does not use the value the method returns. This method is defined as a user-option. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</value>
```
mavasani, line 1440:
```suggestion
    <value>'{0}' calls '{1}' but does not use the value the method returns. This method is defined as a user-option. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</value>
```

mavasani: Wondering why the test framework did not fail the tests when 2 arguments are being provided for expected diagnostics?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/xlf/MicrosoftCodeQualityAnalyzersResources.cs.xlf
```diff
@@ -132,6 +132,11 @@
         <target state="translated">{0} volá {1}, ale nepoužívá hodnotu, kterou tato metoda vrací. O metodách Linq se ví, že nemají vedlejší účinky. Použijte výsledek v podmíněném příkazu, přiřaďte výsledek proměnné nebo ho předejte jako argument jiné metodě.</target>
         <note />
       </trans-unit>
+      <trans-unit id="DoNotIgnoreMethodResultsMessageUserDefinedMethod">
+        <source>'{0}' calls '{1'} but does not use the value the method returns. This method is defined as a user-option. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</source>
```
mavasani, line 136: @Evangelink You would need to rebuild locally to also update the xlf files :-)

Evangelink: 🤦

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: Yep that's weird.

Evangelink: Ok so the test framework is not failing when extra arguments are provided. `{1'}` is not parsed properly and so not considered as a placeholder. The `.WithArguments()` can take more arguments that required and won't fail. Might be worth improving this behavior.

mavasani: @sharwell for the suggestion. Possibly file a suggestion issue on roslyn-sdk repo for the testing library?

Evangelink: https://github.com/dotnet/roslyn-sdk/issues/516

PR #3485, opened 2020-04-08
Use "nullable" instead of "reference"

davkean:
This fires against reference and nullable value types.


src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
```diff
@@ -256,10 +256,10 @@
     <value>Fix numbered comments</value>
   </data>
   <data name="PreferNullLiteralDescription" xml:space="preserve">
-    <value>Use 'null' instead of 'default' for reference types.</value>
+    <value>Use 'null' instead of 'default' for nullable types.</value>
```
mavasani, line 259: Tag @sharwell

davkean:
@mavasani @sharwell This is just a string change, can we please take a look?

PR #3493, opened 2020-04-10
Add support for compiler provided analyzer config options

mavasani:
…onfigOptions

Our current analyzer configuration options support relies on users providing an additional file named `.editorconfig` with the options. This was added as a temporary workaround until .editorconfig based options were supported in CI by Roslyn compiler. Starting VS2019 16.3, roslyn compilers have built-in support for .editorconfig based analyzer options. This change adds support to our options infrastructure to support options provided by compiler provided analyzer config options in the analysis context. We still support the additional file based option configuration for backward compatibility, though it may be removed in the future.

This change requires us to move to Microsoft.CodeAnalysis 3.3.1, which is the first stable release with AnalyzerConfigOptions API.


mavasani:
Tagging @Evangelink

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionsShouldImplementGenericInterface.cs
```diff
@@ -88,24 +88,18 @@ public override void Initialize(AnalysisContext analysisContext)
                        interfaceToGenericInterfaceMapBuilder.Add(new KeyValuePair<INamedTypeSymbol, INamedTypeSymbol>(iEnumerableType, genericIEnumerableType));
                    }
 
-                   var userMap = context.Options.GetAdditionalRequiredGenericInterfaces(Rule, context.Compilation, context.CancellationToken);
```
mavasani, line 91: We cannot compute option values at compilation level anymore. These option values might differ for each document/folder, hence need to be computed for each symbol/syntax tree.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs
```diff
@@ -125,6 +125,11 @@ public override void Initialize(AnalysisContext analysisContext)
 
             analysisContext.RegisterCompilationAction(compilationContext =>
             {
+                if (!(compilationContext.Compilation.SyntaxTrees.FirstOrDefault() is SyntaxTree tree))
```
mavasani, line 128: Simpler approach as code metrics options are unlikely to differ between documents. We can change this in future if required.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/UseSecureCookiesASPNetCore.cs
```diff
@@ -179,6 +179,8 @@ public override void Initialize(AnalysisContext context)
                                         InterproceduralAnalysisConfiguration.Create(
                                             compilationAnalysisContext.Options,
                                             SupportedDiagnostics,
+                                            rootOperationsNeedingAnalysis.First().Item1.Syntax.SyntaxTree,
```
mavasani, line 182: @dotpaul - For option value computed in CompilationEndAction, we do not have a specific tree/document to query the options. I have taken a simple route of just using the first tree for computing option value. You would need to switch to a non-CompilationEndAction based analysis approach if you'd like to support differing options between trees in a compilation for such analyzers.

src/Utilities/Compiler/Options/AbstractCategorizedAnalyzerConfigOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.CodeAnalysis;
+
+namespace Analyzer.Utilities
+{
+    using static CategorizedAnalyzerConfigOptionsExtensions;
+
+    internal abstract class AbstractCategorizedAnalyzerConfigOptions : ICategorizedAnalyzerConfigOptions
```
mavasani, line 12: Most of the code here is existed code from `CategorizedAnalyzerConfigOptions.cs` just extracted to a separate base type.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -16,56 +16,142 @@ namespace Analyzer.Utilities
...
+        private static readonly ConditionalWeakTable<AnalyzerOptions, ICategorizedAnalyzerConfigOptions> s_cachedOptions
+            = new ConditionalWeakTable<AnalyzerOptions, ICategorizedAnalyzerConfigOptions>();
         private static readonly ImmutableHashSet<OutputKind> s_defaultOutputKinds =
             ImmutableHashSet.CreateRange(Enum.GetValues(typeof(OutputKind)).Cast<OutputKind>());
 
+        private static TValue GetOptionWithConflictHandling<TValue>(ISymbol symbol, Func<SyntaxTree, TValue> getOptionForTree, TValue conflictValue)
+        {
+            var value = getOptionForTree(symbol.Locations[0].SourceTree);
+
+            // Handle conflicts from option value in different partial declaration of the symbol.
```
mavasani, line 28: I have been in two minds of whether we should be doing conflict resolution for differing option values across partial declaration of a symbol. After putting some thought, I am going to remove the conflict resolution logic with following reasoning:

1. This is a highly unlikely scenario, especially given most of the non-generated type definitions will be in the same folder and likely have same options.
2. Conflict value chosen might still not be intuitive for the end user, as it will end up being the default value for most cases. The only real solution here is to report a user facing diagnostic about conflicting option values, but that needs much more wholesale repo wide changes, which seems too much work for a corner case scenario.
3. Performance: We will unnecessarily compute option values for multiple trees, which will end up being identical. Additionally, we will end up allocating a delegate for each invocation of this helper.

I will file a feature request for handling conflict resolution, in case we decide to support it in future.

mavasani:
Thanks @dotpaul and @genlu!

PR #3500, opened 2020-04-14
Instantiate ArgumentExceptions Correctly analyzer updates and fixer implementation

buyaa-n:
Fixes https://github.com/dotnet/runtime/issues/33763

- Updates made to analyzer:

1. Increased severity from IdeHidden_BulkConfigurable => IdeSuggestion
2. Don't warn when the enclosing method has no parameter
3. Do not show multiple diagnostics for reversed arguments, it looks ugly and reporting one issue from 2 sides, only showing the diagnosing having fixer:
image
4. Added api surface option for configuring method visibility
5. Do not warn if generic type parameter name used

- Implemented the ~~CSharp only version for fixer, please suggest how to make it more language ignostic way~~ language agnostic fixer


src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -16,4 +16,5 @@ CA2011 | Reliability | Info | AvoidInfiniteRecursion, [Documentation](https://do
 CA2012 | Reliability | Hidden | UseValueTasksCorrectlyAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2012)
 CA2013 | Reliability | Warning | DoNotUseReferenceEqualsWithValueTypesAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2013)
 CA2014 | Reliability | Warning | DoNotUseStackallocInLoopsAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2014)
-CA2015 | Reliability | Warning | DoNotDefineFinalizersForTypesDerivedFromMemoryManager, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2015)
\ No newline at end of file
+CA2208 | Usage | Info | InstantiateArgumentExceptionsCorrectlyAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2208)
```
mavasani, line 19: Interesting - didn't the code fix for updating this file suggest creating a new `## Changed Rules` table for CA2208? It is already part of [AnalyzerReleases.Shipped.md](https://github.com/dotnet/roslyn-analyzers/blob/62e23a714e3609c6bb9daac1dfc9fd3cc6a152e2/src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md), so I would have expected so. I will investigate why the code fix didn't do so.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -460,7 +460,7 @@
     <value>A call is made to the default (parameterless) constructor of an exception type that is or derives from ArgumentException, or an incorrect string argument is passed to a parameterized constructor of an exception type that is or derives from ArgumentException.</value>
   </data>
   <data name="InstantiateArgumentExceptionsCorrectlyMessageNoArguments" xml:space="preserve">
-    <value>Call the {0} constructor that contains a message and/or paramName parameter.</value>
+    <value>Please use overload of the {0} constructor that contains a message and/or paramName parameter.</value>
```
mavasani, line 463: We are generally more assertive in the diagnostic message.
```suggestion
    <value>Use overload of the '{0}' constructor that contains a message and/or paramName parameter.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1275,6 +1275,12 @@
   <data name="DoNotUseStackallocInLoopsMessage" xml:space="preserve">
     <value>Potential stack overflow. Move the stackalloc out of the loop.</value>
   </data>
+  <data name="InstantiateArgumentExceptionsCorrectlyChangeToTwoArgument" xml:space="preserve">
```
mavasani, line 1278: Can you suffix with `Message`, `Title` or `Description` for clarity how we use this resource string?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1275,6 +1275,12 @@
...
+    <value>Change to call the two argument constructor, pass null for the message.</value>
+  </data>
+  <data name="InstantiateArgumentExceptionsCorrectlyFlipArgumentOrder" xml:space="preserve">
+    <value>Swap the arguments order</value>
```
mavasani, line 1282: I think this feels too succinct, don't we need something more descriptive? Something that suggests what was wrong in the original code that needs the swap operation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
```diff
@@ -112,11 +115,25 @@ public override void Initialize(AnalysisContext analysisContext)
                         continue;
                     }
 
-                    CheckArgument(owningSymbol, creation, argument.Parameter, value, context);
+                    diagnostic = CheckArgument(owningSymbol, creation, argument.Parameter, value, context);
```
mavasani, line 118: Nit: can we consider `CheckArgument` itself to report the diagnostic instead of passing it back to the caller? It can be changed to return a bool, and we can break out from this loop when it returns false.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
```diff
@@ -125,26 +142,28 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                    rule = RuleIncorrectParameterName;
                 }
-
-                rule = RuleIncorrectParameterName;
             }
 
             if (rule != null)
             {
-                context.ReportDiagnostic(context.Operation.Syntax.CreateDiagnostic(rule, targetSymbol.Name, stringArgument, parameter.Name, creation.Type.Name));
+                return context.Operation.Syntax.CreateDiagnostic(rule, dictBuilder.ToImmutable(), targetSymbol.Name, stringArgument, parameter.Name, creation.Type.Name);
```
mavasani, line 163: Nit: Just a less verbose way to get the same functionality:
```suggestion
                return context.Operation.CreateDiagnostic(rule, dictBuilder.ToImmutable(), targetSymbol.Name, stringArgument, parameter.Name, creation.Type.Name);
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -13,5 +20,77 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
     [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
     public sealed class CSharpInstantiateArgumentExceptionsCorrectlyFixer : InstantiateArgumentExceptionsCorrectlyFixer
     {
+        protected override void PopulateCodeFix(CodeFixContext context, Diagnostic diagnostic, string paramPositionString, SyntaxNode node)
+        {
+            if (node is ObjectCreationExpressionSyntax creation)
```
mavasani, line 25: It seems like lot of this code can be moved to the base type `InstantiateArgumentExceptionsCorrectlyFixer` if you operate using `IOperation` APIs instead of syntax. You could likely also delete the C# and VB specific implementations, seal `InstantiateArgumentExceptionsCorrectlyFixer` and move `[ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic), Shared] to that type.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -13,5 +20,77 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
...
+                    codeAction = CodeAction.Create(
+                        title: MicrosoftNetCoreAnalyzersResources.InstantiateArgumentExceptionsCorrectlyFlipArgumentOrder,
+                        createChangedDocument: c => SwapArgumentsOrder(context.Document, creation, paramPosition, creation.ArgumentList.Arguments.Count, c),
+                        equivalenceKey: MicrosoftNetCoreAnalyzersResources.InstantiateArgumentExceptionsCorrectlyFlipArgumentOrder);
+                }
+                context.RegisterCodeFix(codeAction, diagnostic);
+            }
+        }
+
+        private static async Task<Document> SwapArgumentsOrder(Document document, ObjectCreationExpressionSyntax creation, int paramPosition, int argumentCount, CancellationToken token)
```
mavasani, line 47: Would this work if you had an `IObjectCreationOperation creation` here?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -13,5 +20,77 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
...
+        {
+            DocumentEditor editor = await DocumentEditor.CreateAsync(document, token).ConfigureAwait(false);
+            SyntaxNode parameter = AddNameOfIfLiteral(creation.ArgumentList.Arguments[paramPosition].Expression, editor.Generator);
+            SyntaxNode newCreation;
+            if (argumentCount == 2)
+            {
+                if (paramPosition == 0)
+                {
+                    newCreation = editor.Generator.ObjectCreationExpression(creation.Type, creation.ArgumentList.Arguments[1], parameter);
```
mavasani, line 56: this could potentially use `creation.Arguments[1].Syntax`

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -13,5 +20,77 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
...
+                        equivalenceKey: MicrosoftNetCoreAnalyzersResources.InstantiateArgumentExceptionsCorrectlyFlipArgumentOrder);
+                }
+                context.RegisterCodeFix(codeAction, diagnostic);
+            }
+        }
+
+        private static async Task<Document> SwapArgumentsOrder(Document document, ObjectCreationExpressionSyntax creation, int paramPosition, int argumentCount, CancellationToken token)
+        {
+            DocumentEditor editor = await DocumentEditor.CreateAsync(document, token).ConfigureAwait(false);
+            SyntaxNode parameter = AddNameOfIfLiteral(creation.ArgumentList.Arguments[paramPosition].Expression, editor.Generator);
```
mavasani, line 50: this could potentially use `creation.Arguments[paramPosition].Value.Syntax`

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -13,5 +20,77 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
...
+        private static async Task<Document> AddNullMessageToArgumentList(Document document, ObjectCreationExpressionSyntax creation, CancellationToken token)
+        {
+            DocumentEditor editor = await DocumentEditor.CreateAsync(document, token).ConfigureAwait(false);
+            SyntaxNode argument = AddNameOfIfLiteral(creation.ArgumentList.Arguments.First().Expression, editor.Generator);
+            SyntaxNode newCreation = editor.Generator.ObjectCreationExpression(creation.Type, editor.Generator.Argument(editor.Generator.NullLiteralExpression()), argument);
+            editor.ReplaceNode(creation, newCreation);
+            return editor.GetChangedDocument();
+        }
+
+        private static SyntaxNode AddNameOfIfLiteral(ExpressionSyntax expression, SyntaxGenerator generator)
```
mavasani, line 87: This should possibly be changed to `IOperation argumentValue` and the check below can be changed to `if (argumentValue is ILiteralOperation literal)`

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -13,5 +20,77 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
...
     public sealed class CSharpInstantiateArgumentExceptionsCorrectlyFixer : InstantiateArgumentExceptionsCorrectlyFixer
     {
+        protected override void PopulateCodeFix(CodeFixContext context, Diagnostic diagnostic, string paramPositionString, SyntaxNode node)
+        {
+            if (node is ObjectCreationExpressionSyntax creation)
+            {
+                int paramPosition = int.Parse(paramPositionString, CultureInfo.InvariantCulture);
+                CodeAction? codeAction = null;
+                if (creation.ArgumentList.Arguments.Count == 1)
+                { // Add null message
```
mavasani, line 30: Nit: we generally avoid a brace and comment on the same line. Suggest moving the comment to a new line.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
mavasani: Having scanned the file, I think you should be able to make the implementation completely language-agnostic by switching to IOperation. Just invoke `semanticModel.GetOperation(node, cancellationToken)` to go from syntax node to operation, and then change this check to `if (node is IObjectCreationOperation creation)`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Ah, so the new strings are just used in the code fix and not any diagnostic message or description? Consider adding a suffix `CodeFixTitle` to the name of the resource string, such as `InstantiateArgumentExceptionsCorrectlyFlipArgumentOrderCodeFixTitle`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Given these new resource strings are only used in offered code fix, and not any diagnostic, I think your current succinct wording looks fine to me - you can ignore this suggestion.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -1,29 +1,35 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Collections.Immutable;
+using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeFixes;
 
+
```
mavasani, line 9: Nit: revert unnecessary new line?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -106,17 +106,24 @@ End Sub
...
                         throw new System.ArgumentException(""first"");
                     }
                 }",
-                GetCSharpIncorrectMessageExpectedResult(6, 31, "Test", "first", "message", "ArgumentException"));
+                GetCSharpIncorrectMessageExpectedResult(6, 31, "Test", "first", "message", "ArgumentException"), @"
+                public class Class
+                {
+                    public void Test(string first)
+                    {
+                        throw new System.ArgumentException(null, nameof(first));
```
mavasani, line 124: Should we generate a named argument for 'null', say `message: null`? Alternatively, should we use a dummy message string to help inform users that they should likely provide an argument message, say `"message-to-be-filled" or some such (with better wording)?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
buyaa-n: Thanks, I will try that

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
buyaa-n: not sure, will try

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
buyaa-n: no idea, i got build error error saying something like "existing CA2208 updated but not listed in Unshipped.md" don't remember exact wording

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: Yes it is for codefix, will add that suffix

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
buyaa-n: The position of `parameterName` and `message` are different accrosss ArgumentException descendants, we want to diagnostic for `message` suppressed `parameterName` diagnostic, For ArgumentException descendants `message` is 2nd parameter and `parameterName` diagnostic would win if we report that diagnostic and break. Also we should trigger fixer only for `message` diagnostic

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
buyaa-n: `null` is suggested by in the issue CC @terrajobst @bartonjs

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
Evangelink: That's also funny because the ordering was broken. The code-fix was doing a good job at keeping the order...

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
bartonjs: > should we use a dummy message string?

`null` results in using the default ArgumentException message, so it's actually quite likely what the caller wants.

> Should we generate a named argument for 'null', say message: null? 

I wouldn't.  (Unless there's an editorconfig style value for this, that the analyzer can see... in which case we should match it).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
bartonjs: I recommend just reverting to the original because "Use" vs "Call" isn't strong enough to warrant a localization reset.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -324,29 +450,92 @@ End Sub
...
+                {
+                    public void Test(string paramName, string message)
+                    {
+                        throw new System.ArgumentException(paramName, message);
+                    }
+                }");
+        }
+
+        [Fact]
+        public async Task ArgumentException_NoArgumentss_ParentMethod_HasNoParameter_DoesNotWarn()
```
bartonjs, line 521:
```suggestion
        public async Task ArgumentException_NoArguments_ParentMethod_HasNoParameter_DoesNotWarn()
```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicInstantiateArgumentExceptionsCorrectly.Fixer.vb
```diff
@@ -13,5 +13,8 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
     Public NotInheritable Class BasicInstantiateArgumentExceptionsCorrectlyFixer
         Inherits InstantiateArgumentExceptionsCorrectlyFixer
 
+        Protected Overrides Sub PopulateCodeFix(context As CodeFixContext, diagnostic As Diagnostic, paramPositionString As String, node As SyntaxNode)
+            Throw New NotImplementedException()
```
bartonjs, line 17: I assume this is not an acceptable fixer :smile:.

Since the type is sealed (NotInheritable) and doesn't do anything; why do we even build it currently?

If it can go to IOperation I expect we're deleting it; but if we have to play with syntax nodes and we're not making it work I assume it's better to just delete it now, too.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -106,49 +106,148 @@ End Sub
...
-                End Class",
-                GetBasicIncorrectMessageExpectedResult(4, 31, "Test", "first", "message", "ArgumentException"));
+            await VerifyVB.VerifyCodeFixAsync(@"
+Public Class [MyClass]
+    Public Sub Test(first As String)
+        Throw New System.ArgumentException(""first"")
+    End Sub
+End Class",
+                GetBasicIncorrectMessageExpectedResult(4, 15, "Test", "first", "message", "ArgumentException"), @"
+Public Class [MyClass]
```
buyaa-n, line 135: Somehow indented sources not matching for VB tests, not sure why, will try to fix later

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: Reverted but localization state still updated to `state="needs-review-translation"`

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
buyaa-n: It did work after converting the SyntaxNode to IOperation using  SemanticModel

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
buyaa-n: it worked with little alterations

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInstantiateArgumentExceptionsCorrectly.Fixer.cs
mavasani: Yay!!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
```diff
@@ -121,13 +121,19 @@ public override void Initialize(AnalysisContext analysisContext)
...
+
                 if (diagnostic != null)
                 {
                     context.ReportDiagnostic(diagnostic);
                 }
             }
         }
 
+        private static bool MatchesConfiguredVisibility(ISymbol owningSymbol, OperationAnalysisContext context)
+        { // Should add for all rules?
```
bartonjs, line 133:
```suggestion
        {
            // Should add for all rules?
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -593,6 +593,111 @@ End Sub
...
+                        Throw New System.ArgumentException(""first is incorrect"", NameOf(TEnum))
+                    End Sub
+                End Class");
+        }
+
+
+        [Theory]
+        [InlineData("public", "dotnet_code_quality.api_surface = private", false)]
+        [InlineData("private", "dotnet_code_quality.api_surface = internal, public", false)]
+        [InlineData("public", "dotnet_code_quality.CA2208.api_surface = private, piblic", true)]
```
bartonjs, line 620: Is "piblic" an intentional typo?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
buyaa-n: Oh it was comment for myself, the `MatchesConfiguredVisibility(...)` method taking DiagnosticRule argument but we have 3 rules here, so I was wondering if I should check for all rules, but that was not needed, I have test for other rule, it was working, will remove

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
buyaa-n: huh, no, wondering how the test passed, maybe i did the typo after runnning tests, will check

buyaa-n: nothing wrong, public is default for SymbolVisibility enum, so anything non matching counted as public, anyways fixed typo,, thanks!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
buyaa-n: it seems not really depending from my code/fixer, Document editor might have issue with replacing VB node, @mavasani any suggestions?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
```diff
@@ -112,39 +115,56 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                    context.ReportDiagnostic(diagnostic);
                 }
             }
         }
-        private static void CheckArgument(
+
+        private static bool MatchesConfiguredVisibility(ISymbol owningSymbol, OperationAnalysisContext context)
+        {
+            return owningSymbol.MatchesConfiguredVisibility(context.Options, RuleIncorrectParameterName,
+                context.Compilation, context.CancellationToken, defaultRequiredVisibility: SymbolVisibilityGroup.All);
```
buyaa-n, line 135: I am wondering why the api surface configuration missing `protected` visibility is it fall into `public` or ...? CC @mavasani

mavasani: Do you want the rule to run on all methods by default or just public API surface? Your current approach will do the former, but I think you want the latter? If so, you should remove the explicit argument `defaultRequiredVisibility`, as the default argument for this parameter ensures only public API surface is analyzed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -368,6 +593,116 @@ End Sub
...
+                Public Class [MyClass]
+                    Public Sub Test(Of TEnum)(ByVal first As TEnum)
+                        Throw New System.ArgumentException(""first is incorrect"", NameOf(TEnum))
+                    End Sub
+                End Class");
+        }
+
+
+        [Theory]
+        [InlineData("public", "dotnet_code_quality.api_surface = private", false)]
```
mavasani, line 618: Please add a test where 2nd argument is empty, i.e. no end user configuration, and validate that diagnostic is reported if first argument is "public" or "protected", but not when it is "private" or "internal". Basically test that the default configuration is enabling analysis only for public API surface.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
mavasani: Alright, so I manually edited the descriptors for `InstantiateArgumentExceptionsCorrectlyAnalyzer` to use `RuleLevel.IdeSuggestions`, and the code fix does offer adding an entry for changed rule:

ReleaseTrackingCodeFix

Kindly revert the manual update to this file, and apply the code fix to do this correctly.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -1,29 +1,116 @@
...
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
-            // See https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md for more information on Fix All Providers
-            return WellKnownFixAllProviders.BatchFixer;
+            var diagnostic = context.Diagnostics.First();
+            string paramPositionString = diagnostic.Properties.GetValueOrDefault(InstantiateArgumentExceptionsCorrectlyAnalyzer.MessagePosition);
+            if (paramPositionString != null)
+            {
+                SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+                SyntaxNode node = root.FindNode(context.Span, getInnermostNodeForTie: true);
```
mavasani, line 34: add a null check for `node` and avoid calling PopulateCodeFixAsync

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
buyaa-n: I think it better to leave it as `All` by default as we can fine tune it using `api_surface` option, my plan was to allow all by default and set public/protected option for runtime repo, but protected option is missing which is why am asking about that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -1,29 +1,116 @@
...
+                if (paramPosition == 0)
+                {
+                    newCreation = editor.Generator.ObjectCreationExpression(creation.Type, creation.Arguments[1].Syntax, parameter);
+                }
+                else
+                {
+                    newCreation = editor.Generator.ObjectCreationExpression(creation.Type, parameter, creation.Arguments[0].Syntax);
+                }
+            }
+            else // 3 arguments
```
mavasani, line 83: Can we change the comment to an assert?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -1,29 +1,116 @@
...
+                else
+                {
+                    newCreation = editor.Generator.ObjectCreationExpression(creation.Type, parameter, creation.Arguments[1].Syntax, creation.Arguments[0].Syntax);
+                }
+            }
+            editor.ReplaceNode(creation.Syntax, newCreation);
+            return editor.GetChangedDocument();
+        }
 
+        private static async Task<Document> AddNullMessageToArgumentList(Document document, IObjectCreationOperation creation, CancellationToken token)
```
mavasani, line 98:
```suggestion
        private static async Task<Document> AddNullMessageToArgumentListAsync(Document document, IObjectCreationOperation creation, CancellationToken token)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.Fixer.cs
buyaa-n: sure

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
```diff
@@ -112,39 +115,56 @@ public override void Initialize(AnalysisContext analysisContext)
                         continue;
                     }
 
-                    CheckArgument(owningSymbol, creation, argument.Parameter, value, context);
+                    diagnostic = CheckArgument(owningSymbol, creation, argument.Parameter, value, context);
+                    if (diagnostic != null && !diagnostic.Properties.IsEmpty)
```
mavasani, line 119: Can you please add a comment here?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
mavasani: You should also be able to inline this method...

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -368,6 +593,116 @@ End Sub
...
 
+        [Fact]
+        public async Task ArgumentException_GenericParameterName_DoesNotWarn()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+                public class Class
+                {
+                    public void Test<TEnum>(TEnum first)
+                    {
+                        throw new System.ArgumentException(""first is incorrect"", nameof(TEnum));
```
mavasani, line 604: Shouldn't we also have a test that validates that `throw new System.ArgumentException(""TEnum"");` is fixed to `throw new System.ArgumentException(null, nameof(TEnum));`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
```diff
@@ -368,6 +593,116 @@ End Sub
...
+ End Class"
+                    },
+                    AdditionalFiles = { (".editorconfig", editorConfigText) }
+                },
+                MarkupOptions = MarkupOptions.UseFirstDescriptor
+            }.RunAsync();
+        }
+
+        [Fact]
+        public async Task EditorConfigConfiguredPublic_PrivateMethods_TriggeringOtherRules_DoesNotWarn()
```
mavasani, line 678: I think this test becomes redundant to take the approach suggested in https://github.com/dotnet/roslyn-analyzers/pull/3500/files#r413392484, which also validates that flipping the accessibility of the method from private to public on the same test generates diagnostics.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.Fixer.cs
```diff
@@ -1,29 +1,116 @@
...
-        public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
+        private static async Task PopulateCodeFixAsync(CodeFixContext context, Diagnostic diagnostic, string paramPositionString, SyntaxNode node)
         {
-            // Fixer not yet implemented.
-            return Task.CompletedTask;
+            SemanticModel model = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            var operation = model.GetOperation(node, context.CancellationToken);
+            if (operation is IObjectCreationOperation creation)
+            {
+                int paramPosition = int.Parse(paramPositionString, CultureInfo.InvariantCulture);
```
mavasani, line 45: Use `TryParse` with bail out on failure to be on safer side.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectly.cs
mavasani: > my plan was to allow all by default 

Ah, sorry, I misunderstood your desire. Your current code if fine then .

> protected option is missing which is why am asking about that

`SymbolVisibility` is actually computed by walking the containing symbol chain and computing overall visibility. So, a protected method within an internal type will have `SymbolVisibility.Internal`, while protected member within a public type will have `SymbolVisibility.Public`. It is basically effective member visibility.

See https://github.com/dotnet/roslyn-analyzers/blob/eb943221428db2173668b07204759cd64db134e5/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L249-L294

buyaa-n: Got it, thank you!

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
buyaa-n: Oh sorry, didn't noticed there were analyzer and fixer, its worked as needed, thanks!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
buyaa-n: Its not really testing that, the `MatchesConfiguredVisibility(...)` method takes DiagnosticDescriptor argument, for which i passed one of the 3 descriptors of the analyzer, i was wondering if i need ot call that method for each descriptor or it will work for all, and it is testing if the api surface option works for other descriptors even if i didn't checked that especially for those descriptors

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InstantiateArgumentExceptionsCorrectlyTests.cs
buyaa-n: As mentioned before our default visibility intended to be `All` so it would report diagnostic on private or internal, so added cases for empty confiuration that suceed for all (didn't added for internal as its different in VB)

PR #3508, opened 2020-04-15
Use only one notation for doc web links

Evangelink:


mavasani:
@Evangelink there are merge conflicts.

PR #3517, opened 2020-04-17
Fix build break

mavasani:
#3429 caused a build break as its PR validation completed before #3493 was merged in, which has a conflicting API change


Evangelink:
I might have a couple of other PRs in the same state.

mavasani:
@Evangelink - thanks, can you close and re-open those PRs, so we don't accidentally merged in till you fix and commit your change?

Evangelink:
Sure!

PR #3520, opened 2020-04-20
New analyzer: Prevent invocation of non-pure methods from readonly value-type fields

maxkoshevoi:
Changes associated with issue #3523

Recently I faced with unobvious behavior in C#. 

Take a look at this code:
```cs
class ListWrapper
{
    private readonly List<int>.Enumerator _enumerator;
 
    public ListWrapper(List<int> list)
    {
        _enumerator = list.GetEnumerator();
    }
 
    public void PrintNextElement()
    {
        _enumerator.MoveNext();
        Console.Write(_enumerator.Current);
    }
}

var wrapper = new ListWrapper(new List<int> { 1, 2, 3 });
wrapper.PrintNextElement();
wrapper.PrintNextElement();
wrapper.PrintNextElement();
```

Output will be: 000
But expected output is: 123

There will be no warnings or errors on compile and runtime.
Here is even post about it from jetbrains: https://www.jetbrains.com/help/resharper/ImpureMethodCallOnReadonlyValueField.html

**What is working**
If there is `readonly` field or property
- Warning if field's type is value type, method is not pure (not `readonly` and doesn't have `[Pure]` attribute) and we invoking or referencing it 
- No warning if field's type is reference type
- No warning if field's type is `readonly` 
- No warning if field's method is `readonly`
- No warning if field's method marked with `[Pure]` attribute
- No warning if invocation of field's method is directly in constructor body

**Not working**
- If method is referenced (e.g. assigned to delegate) inside a constructor, or invoked inside a function that is inside a constructor(lambda, local function or some other form of function), there will be a warning since I cannot guarantee whether method will be invoked from a constructor or just assigned to something in a constructor and invoked later.
But is seems like a very rare false positive
- For test `CSharpInvokingPureMethod` C# 8 is needed, but tests are executed using C# 7.3

image


sharwell:
I believe this is effectively a duplicate of #3420

maxkoshevoi:
> I believe this is effectively a duplicate of #3420

Seems not the case. Your analyzer doesn't produce warning through unit test or live testing: 

image
image

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidNonPureMethodCallOnReadonlyValueSymbol.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (parent != null && parent.Kind == OperationKind.ConstructorBody)
+            {
+                return;
+            }
+
+            var diagnostic = Diagnostic.Create(Rule, callOperation.Syntax.GetLocation(), callOperation.Instance.Syntax, callOperation.TargetMethod.Name);
+            context.ReportDiagnostic(diagnostic);
+        }
+
+        private static bool AnalizeMethod(IOperation instance, IMethodSymbol method)
```
Tyrrrz, line 92: Typo: Analyze

sharwell:
@maxkoshevoi Currently the non-copyable analyzer only treats value types as non-copyable if they are marked with an attribute. The analysis proposed here is the same analyzer, except it automatically gets enabled for any mutable value type.

maxkoshevoi:
> @maxkoshevoi Currently the non-copyable analyzer only treats value types as non-copyable if they are marked with an attribute. The analysis proposed here is the same analyzer, except it automatically gets enabled for any mutable value type.

@sharwell  If I understand correctly, your analyzer treat entire type as `[NonCopyable]` everywhere, whitch is not the goal of this analyzer. I don't prevent copying type, only modifying it and only if instance of the type is `readonly`.

In another words, how using `DoNotCopyValue` analyzer can I generate a warning here:
```cs
class ListWrapper
{
    private readonly List<int>.Enumerator _enumerator;
 
    public ListWrapper(List<int> list)
    {
        _enumerator = list.GetEnumerator();
    }
 
    public void PrintNextElement()
    {
        _enumerator.MoveNext();
        Console.Write(_enumerator.Current);
    }
}
```

Obviously, I don't have access to struct `List<int>.Enumerator`, and even if I had, it's still valid to call `MoveNext` in other circumstances.

sharwell:
Attempting to modify a `readonly` value type field is only a problem because you are modifying a copy. The following would be equally problematic:

```
class ListWrapper
{
    private List<int>.Enumerator _enumerator;
 
    public ListWrapper(List<int> list)
    {
        _enumerator = list.GetEnumerator();
    }
 
    public void PrintNextElement()
    {
        var enumerator = this._enumerator;
        enumerator.MoveNext();
        Console.Write(enumerator.Current);
    }
}
```

Tyrrrz:
@sharwell did you actually look at the linked article? It should answer your questions.

https://www.jetbrains.com/help/resharper/ImpureMethodCallOnReadonlyValueField.html

maxkoshevoi:
@sharwell my point is that it is really unobvious that I'm modifying a copy. 
It looks really strange that I'm invoking method that's located **inside** an instance of an object, it changes it's state, but the moment I return from this method, all state changes are gone. 
As far as I know in all other cases methods **inside** instances of structs or classes always working with original instance and not the copy, so behavior in this case is really unobvious.



> The following would be equally problematic:

No, it wouldn't. In your example `MoveNext` from `enumerator` variable will actually change is's state. Behavior that `var enumerator = this._enumerator;` will copy the struct instead of coping a link to `this._enumerator` is expected, so output `111` is also expected. If not, problem can quickly be found (since you can compare `this._enumerator.Current` and `enumerator.Current` after `MoveNext` is called).

TLDR: My point is, that in your example `MoveNext` will actually change state of `enumerator` where in my example, it will not (to be more specific - state changes won't be saved to instance on which `MoveNext` was called)

image


> it did successfully detect a problematic scenario

Yes, it did, but 
- Only when I added attribute to the type, which is not always possible (see `List<int>.Enumerator` example), and will apply to all instances of said type (which is not the intent here)
- It has false positive if you call method from within a constructor (there will be no coping)

image

sharwell:
> In another words, how using DoNotCopyValue analyzer can I generate a warning here:

In the code checked in today, it would not produce this warning. However, any expansion of this analyzer to detect `List<T>.Enumerator` as a non-copyable type would result in a diagnostic being reported for incorrect use of these methods.

sharwell:
> My point is that it is really unobvious that I'm modifying a copy.

It would be good to have an improved diagnostic message on Do Not Copy analyzer to indicate that a non-`readonly` operation is being performed on a copy of a mutable value type; the side effects of this operation will not be visible after the completion of the statement.

> No, it wouldn't. In your example MoveNext from enumerator variable will actually change is's state. Behavior that var enumerator = this._enumerator; will copy the struct instead of coping a link to this._enumerator is expected, so output 111 is also expected.

The problem would occur if `PrintNextElement` is called twice.

> It has false positive if you call method from within a constructor (there will be no coping)

There are many false positives in this analyzer. Its design intentionally assumes that for any case it isn't previously coded to recognize, it's possible the compiler has made a copy.

maxkoshevoi:
> It would be good to have an improved diagnostic message on Do Not Copy analyzer to indicate that a non-readonly operation is being performed on a copy of a mutable value type; the side effects of this operation will not be visible after the completion of the statement.

Exactly! That's what I'm trying to do =) It should be a thing that's turned on by default.

> The problem would occur if `PrintNextElement` is called twice.

I got that, but this behavior (copy by value) is something that value-type is defined by (so it's relatively known) and it cannot directly lead to problems.
On the other hand behavior in this analyzer is far rare and what's important, will 99% lead to problems. Since if method isn't `readonly` or `[Pure]`, it changes the state, and this state will defensively be lost. So, this analyzer only warns about places where there definitely be a problem.


On unrelated note. Do we have an analyzer to force extension methods for value types set first parameter as `in`? Cause if it's not, this causes exactly the same problem. And there are no downsides of using it (or am I missing something)

maxkoshevoi:
> However, any expansion of this analyzer to detect `List<T>.Enumerator` as a non-copyable type would result in a diagnostic being reported for incorrect use of these methods.

Fair point. But this analyzer is a little bit more specific. It allows coping of readonly value-type fields, and only disallows calling their unpure methods. So maybe it could be integrated into `DoNotCopyValue` analyze, but this will pollute it's purpose a little

sharwell:
The advantage of DoNotCopyValue is it forms the analysis foundation for anything that needs to warn for copying a value. The analysis proposed here is one specific case within this analysis:

* Warn if a `readonly` field is copied for the purpose of invoking a member which is not marked `readonly` or `[Pure]`

maxkoshevoi:
> The advantage of DoNotCopyValue is it forms the analysis foundation for anything that needs to warn for copying a value.

All methods and classes from `DoNotCopyValue` analyzer are `private`, so it's not foundation for anything really. If there was `OperationKind.Coping`, I would use it.

> the [indexer is not pure](https://github.com/dotnet/runtime/blob/faebda75e202d810b00ce8176c79363006f62113/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.Minimal.cs#L123-L126) (according to its signature).

Indexers are not included in this analyzer (though maybe they should be), but point taken. We should make this indexer `readonly` then. In addition to allowing this analyzer to work properly and making method header more clear, it potentially can even improve performance by allowing CLR to execute this indexer without coping `ImmutableArray<T>` instance

maxkoshevoi:
Also `DoNotCopyValue` doesn't detect `readonly` properties.

There should be a warning here:
image

sharwell:
> There should be two warnings here:

The analyzer should be reporting a warning during this test, just not on the two lines shown here. The warning should actually be here:

```
Mutable a { [|get|]; }
```

For both references to `a` within `Method`, the operation is on the return value from a method invocation (the method is `get_a()`). Methods are always assumed to own the values they return, so uses of them are treated as a value _move_ as opposed to a value _copy_.

sharwell:
@dotnet/dotnet-analyzers I'm not sure how to proceed with this one. Even though the replacement I'm proposing is still in its early stages, I'm concerned that the benefits provided by an interim approach wouldn't outweigh the drawbacks.

maxkoshevoi:
> The warning should actually be here

Just checked with ILSpy and it seems that `a` is not copied in `get_a()`, it's copied on method invocation (`a.A()`). Moreover it's copied regardless of whether property is read-only or not.
So this analyzer needs some modifications to fire when method is invoked even on mutable value type property.

Regarding the noise we could RuleLevel set  to Message instead of the Warning

sharwell:
> Just checked with ILSpy and it seems that a is not copied in get_a()

This is not correct. The copy occurs for the `ldfld` (the non-copying version of this instruction is `ldflda`).

> it's copied on method invocation (a.A())

This is a move, not a copy, so it's not problematic. During a move, the original value is "lost" so it's not possible to reach a state where a copy was mutated but the original is accessible with the pre-mutation value.

maxkoshevoi:
> This is a move, not a copy, so it's not problematic.

How does this work then? How can I increment `Num` field of property `str` (or access incremented version of `str` or something)?

image

sharwell:
The main change you would need is one of the following:

```
// Make str a field:
public Str str;

// Keep str a property, but make it return a reference:
private Str _str;
public ref Str str => ref _str;
```

maxkoshevoi:
Thanks for example of mutable value property

> Any auto-property will always copy its value during the return

I don't think it's copied during the return. As you can see in my screenshots it's copied during the method invocation. 
Anyway, copying is not the problem here, lost of altered state is. And the goal of this analyzer is to warn about unobvious lost of state changes (and not the copying itself).

sharwell:
> Anyway, copying is not the problem here, lost of altered state is.

The latter is _always_ caused by the former. Copying is not always problematic, but if a problem does occur then it was caused by operating on a copy. Detecting copies is the only way this analyzer can truly achieve its intended goal. Alternative approaches catch some but not all cases.

The primary advantage of intentionally restricted analysis to a subset of cases occurs when the restriction fully (or largely) prevents the analyzer from reporting false positives. The proposal above doesn't fall into this category because most of the false positives that occur with copy analysis can also appear for readonly field analysis.

> I don't think it's copied during the return. As you can see in my screenshots it's copied during the method invocation.

Values are only considered copied if they are not moves. In other words, after the memory is copied from location **a**&rarr;**b**, the location **a** must continue to be usable. This situation occurs for the `ldfld` instruction in `get_str()`, but it does not occur for the `stloc` (always a move because **a** is a stack location that gets popped as part of the instruction) or `ldflda` (loads by reference and not by value) instructions highlighted.

maxkoshevoi:
> Detecting copies is the only way this analyzer can truly achieve its intended goal.

True, but 
1) `Do Not Copy Value` is not reporting where it should (for example for properties) and reporting where it shouldn't (for example invoking method of readonly field within a constructor). Those cases are already handled by this analyzer (I'm not saying it's perfect though).
2) Even if it worked perfectly, it cannot be integrated into this analyzer in straightforward way since this would produce a lot of false positives (like you said "copying is not always problematic")
3) This analyzer needs to by out-out, and not opt-in (like `Do Not Copy Value` with it's attribute)

If w go back to square one the goal of this analyzer is to be companion to existing analyzers that prevent direct changes to state if state changes are not allowed.

Like this:

image

`str.Num = 42;` is reported as error by VS, so `str.Increment();` should also be reported. That's it, that's the whole purpose of this analyzer. 
It also should only report on method invocation (or method reference) since method is what doing the state change (for example, it shouldn't report here `Mutable a { [|get|]; }`)

mavasani (changes requested):

maxkoshevoi:
@sharwell hi, is there some news about this PR? You said that this analyzer could be implemented using #3420. Since that PR is merged now, could you please provide some details on how it could be done? Or maybe some general comments on existing code

Youssef1313:
@jmarolf This got closed unintentionally after the deletion of release/6.0.1xx-preview1 branch

PR #3511, opened 2020-04-15
Update NetCore resources

Evangelink:

PR #3521, opened 2020-04-20
Update all resources in MicrosoftNetCoreAnalyzersResources

Evangelink:
Replace #3511.

Tagging @dotpaul for review and @mavasani for info.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
@@ -127,25 +127,22 @@
     <value>Review SQL queries for security vulnerabilities</value>
   </data>
   <data name="DoNotCallToImmutableCollectionOnAnImmutableCollectionValueTitle" xml:space="preserve">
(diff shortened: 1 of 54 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -184,43 +181,43 @@
     <value>An operating system invoke method is defined and a method that has the equivalent functionality is located in the .NET Framework class library.</value>
   </data>
   <data name="UseManagedEquivalentsOfWin32ApiMessage" xml:space="preserve">
-    <value>Use managed equivalents of win32 api</value>
+    <value>Use managed equivalents of win32 api.</value>
```
mavasani, line 184: I thought we decided not to have a period for the title and message strings, only for description.

Evangelink: I thought we said '.' for messages too but you are right: https://github.com/dotnet/roslyn-analyzers/pull/3183#issuecomment-576724513 I'll do a new pass through the resources then xD

PR #3534, opened 2020-04-20
Use String.Contains instead of String.IndexOf where applicable

pgovind:
This is ready to go through a review now


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+                            }
+
+                            context.RegisterCodeFix(
+                                CodeAction.Create(
+                                    title: MicrosoftNetCoreAnalyzersResources.PreferStringContainsOverIndexOfTitle,
+                                    createChangedDocument: c => ReplaceBinaryOperationWithContains(doc, syntaxNode, root, argumentsForContainsInvocation, binaryOperation.Syntax, c, variableDeclarationGroupOperation, currentCulture),
+                                    equivalenceKey: "PreferStringContainsCurrentCultureOverIndexOfFixer"),
+                                context.Diagnostics);
+
+                            context.RegisterCodeFix(
```
pgovind, line 91: Is this the right way to register 2 code-fixes for this case? When I ran this analyzer locally, I didn't see both options pop-up in the UI. What am I missing?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,197 @@
...
+                        ParameterInfo.GetParameterInfo(charType),
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
+                if (stringAndComparisonTypeArgumentContainsMethod == null ||
+                    charAndComparisonTypeArgumentContainsMethod == null)
+                {
+                    return;
+                }
+
+                Dictionary<string, int> localSymbolsToNumberOfReferences = new Dictionary<string, int>();
+                compilationContext.RegisterOperationAction(referenceContext =>
```
pgovind, line 93: I have 2 actions registered here because I need to find all references to a local variable. For ex:
``` csharp
namespace TestNamespace 
{ 
    class TestClass 
    { 
        private void TestMethod() 
        { 
            const string str = ""This is a string"";
            int index = str.IndexOf(" + quotes + input + quotes + @");
            if (index == -1)
            {

            }
            // I need to know if index is used later
        } 
    } 
}

```
After the if block checks the value of `index`, I need to know if there are other references to `index` to understand if it is safe to remove the `int index = …` line. AFAIK, there is no way to do this with just the semantic model and analyze data flow. However, this approach doesn't seem to work either. When I debugged this code, I see that this callback is called 1st and stores `str`, then the binary operation callback below is called next, but it won't find the symbol `index` in the `localSymbolsToNumberOfReferences` dictionary, and then finally this callback is called to store `index`. The binary operation callback is not called after this, so it ends up being too late. Questions:
1) Is my approach right? If so, how do I ensure that the dictionary contains `index` when I need to look up the number of references.
2) If not, what is the right way to do this?

@mavasani

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,216 @@
...
+                }
+
+                compilation.RegisterOperationBlockStartAction(context =>
+                {
+                    ISymbol symbol = context.OwningSymbol;
+                    if (!(context.OwningSymbol is IMethodSymbol method))
+                    {
+                        return;
+                    }
+                    context.RegisterOperationAction(context =>
```
pgovind, line 102: This is still not the right way to implement this analyzer since it depends on the order in which the callbacks are executed. Is there a better way to find the number of references to a variable BEFORE the `OperationKind.Binary` callback is called? If not, I think we may have to reduce the scope of https://github.com/dotnet/runtime/issues/33785 to not handle local variables.

@mavasani

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+            Document doc = context.Document;
+            CancellationToken cancellationToken = context.CancellationToken;
+            var diagnostics = context.Diagnostics;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode expression)
+            {
+                SemanticModel semanticModel = await doc.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+                var compilation = semanticModel.Compilation;
+                var operation = semanticModel.GetOperation(expression, cancellationToken);
+                if (operation is IBinaryOperation binaryOperation)
```
Evangelink, line 37: You could revert this condition to avoid too much nesting.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.Fixer.cs
Evangelink: I think that the 2 titles need to differ.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                if (!compilation.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemString, out INamedTypeSymbol? stringType))
+                {
+                    return;
+                }
+                if (!compilation.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemChar, out INamedTypeSymbol? charType))
+                {
+                    return;
+                }
+                if (!compilation.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemStringComparison, out INamedTypeSymbol? stringComparisonType))
```
Evangelink, line 54:
```suggestion
                if (!compilation.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemString, out INamedTypeSymbol? stringType) ||
                !compilation.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemChar, out INamedTypeSymbol? charType) ||
                !compilation.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemStringComparison, out INamedTypeSymbol? stringComparisonType))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,216 @@
...
+                                                                                      RuleLevel.IdeSuggestion,
+                                                                                      s_localizableDescription,
+                                                                                      isPortedFxCopRule: false,
+                                                                                      isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private ImmutableArray<Location> _locations;
+
+        private readonly Dictionary<string, int> localSymbolsToNumberOfReferences = new Dictionary<string, int>();
```
Evangelink, line 38: This doesn't seem to be compatible with concurrent execution.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,216 @@
...
+                                var variableName = localReferenceOperation.Local;
+                                SyntaxReference declaration = variableName.DeclaringSyntaxReferences.FirstOrDefault();
+                                if (declaration is null)
+                                {
+                                    return;
+                                }
+
+                                var semanticModel = context.Operation.SemanticModel;
+                                SyntaxNode declarationNode = declaration.GetSyntax(context.CancellationToken);
+                                var operation = semanticModel.GetOperation(declarationNode, context.CancellationToken);
```
Evangelink, line 141: I might be wrong but this doesn't seem right to access syntax node then operation, I would expect you to be able to get this operation "directly".

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
mavasani: I am not sure I understand the problem, let's sync offline.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
pgovind: Any idea how? I'm not super familiar with the APIs here. I need to go from an `ILocalReferenceOperation` to it's `VariableDeclaratorOperation`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.Fixer.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        private static async Task<Document?> ReplaceBinaryOperationWithContains(Document document, SyntaxNode localReferenceOrParameterNode, SyntaxNode treeRoot, IEnumerable<SyntaxNode> argumentsForContainsInvocation, SyntaxNode binaryOperationSyntaxNode, CancellationToken cancellationToken, IVariableDeclarationGroupOperation? variableDeclarationGroupOperation = null, ISymbol? stringComparisonArgumentToContainsInvocation = null)
+        {
+            DocumentEditor editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            SyntaxEditor syntaxEditor = new SyntaxEditor(treeRoot, document.Project.Solution.Workspace);
+            SyntaxGenerator generator = editor.Generator;
+            var memberAccessExpression = generator.MemberAccessExpression(localReferenceOrParameterNode, "Contains");
+            if (stringComparisonArgumentToContainsInvocation != null)
+            {
+                var systemIdentifier = generator.IdentifierName("System");
```
pgovind, line 122: In other PRs, is it common to look at  the current namespaces in the file and include `System` only if it is missing?

PR #3537, opened 2020-04-22
Implement Prefer IsKind analyzer

### LINKED ISSUE #3496: Meta analyzer: Use IsKind instead of Kind
We should have an analyzer that offers the following conversions:

| Form | Improved form |
| --- | --- |
| `x.Kind() == SyntaxKind.Kind` | `x.IsKind(SyntaxKind.Kind)` |
| `x.Kind() != SyntaxKind.Kind` | `!x.IsKind(SyntaxKind.Kind)` |
| `x == default(SyntaxToken)` | `x.IsKind(SyntaxKind.None)` |

These apply for both `SyntaxToken` and `SyntaxNode`.

sharwell:
Closes #3496


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (operation.OperatorKind != BinaryOperatorKind.Equals
+                && operation.OperatorKind != BinaryOperatorKind.NotEquals)
+            {
+                return;
+            }
+
+            if (operation.LeftOperand.WalkDownConversion() is IInvocationOperation { TargetMethod: { Name: "Kind", ContainingType: var containingType } }
+                && containingTypeMap.TryGetValue(containingType, out _))
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Rule, operation.LeftOperand.Syntax.GetLocation()));
```
mavasani, line 79: Nit:
```suggestion
                context.ReportDiagnostic(operation.LeftOperand.CreateDiagnostic(Rule));
```

sharwell: Fixed

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
```diff
@@ -0,0 +1,57 @@
...
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(CSharpPreferIsKindFix))]
+    [Shared]
+    public sealed class CSharpPreferIsKindFix : PreferIsKindFix
+    {
+        protected override async Task<Document> ConvertKindToIsKindAsync(Document document, TextSpan sourceSpan, CancellationToken cancellationToken)
+        {
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            var binaryExpression = root.FindNode(sourceSpan, getInnermostNodeForTie: true).FirstAncestorOrSelf<BinaryExpressionSyntax>();
+
+            if (!(binaryExpression.Left is InvocationExpressionSyntax invocation))
```
mavasani, line 22:
```suggestion
            if (!(binaryExpression?.Left is InvocationExpressionSyntax invocation))
```

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
```diff
@@ -0,0 +1,423 @@
...
+class C
+{{
+    bool Method(SyntaxNode node)
+    {{
+        return [|node.Kind()|] {@operator} SyntaxKind.None;
+    }}
+}}
+";
+
+            var prefix = @operator switch
```
mavasani, line 36: Nit: can have the prefix be another test method parameter. This is fine as well though..

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -503,4 +503,16 @@
   <data name="EnableAnalyzerReleaseTrackingRuleDescription" xml:space="preserve">
     <value>Enabling release tracking for analyzer packages helps in tracking and documenting the analyzer diagnostics that ship and/or change with each analyzer release. See details at https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md.</value>
   </data>
+  <data name="PreferIsKindDescription" xml:space="preserve">
+    <value>Prefer IsKind</value>
```
mavasani, line 507: Can the description be improved to state why `IsKind` is preferred? What is the performance drawback from using `Kind()`

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -503,4 +503,16 @@
...
+  </data>
+  <data name="PreferIsKindMessage" xml:space="preserve">
+    <value>Prefer IsKind</value>
+  </data>
+  <data name="PreferIsKindTitle" xml:space="preserve">
+    <value>Prefer IsKind</value>
+  </data>
+  <data name="PreferIsKindFix" xml:space="preserve">
+    <value>Prefer IsKind</value>
```
mavasani, line 516: Should we just reuse `PreferIsKindTitle` for the code fix title as well?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink, line 507:
```suggestion
    <value>Prefer 'IsKind'.</value>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -503,4 +503,16 @@
   <data name="EnableAnalyzerReleaseTrackingRuleDescription" xml:space="preserve">
     <value>Enabling release tracking for analyzer packages helps in tracking and documenting the analyzer diagnostics that ship and/or change with each analyzer release. See details at https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md.</value>
   </data>
+  <data name="PreferIsKindDescription" xml:space="preserve">
+    <value>Prefer IsKind</value>
+  </data>
+  <data name="PreferIsKindMessage" xml:space="preserve">
+    <value>Prefer IsKind</value>
```
Evangelink, line 510:
```suggestion
    <value>Prefer 'IsKind'</value>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -503,4 +503,16 @@
...
+  </data>
+  <data name="PreferIsKindTitle" xml:space="preserve">
+    <value>Prefer IsKind</value>
```
Evangelink, line 513:
```suggestion
    <value>Prefer 'IsKind'</value>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink, line 516:
```suggestion
    <value>Prefer 'IsKind'</value>
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                // Kind() methods
+                //  Microsoft.CodeAnalysis.CSharp.CSharpExtensions.Kind
+                //  Microsoft.CodeAnalysis.VisualBasic.VisualBasicExtensions.Kind
+                //
+                // IsKind() methods
+                //  Microsoft.CodeAnalysis.CSharpExtensions.IsKind
+                //  Microsoft.CodeAnalysis.VisualBasicExtensions.IsKind
+
+                Dictionary<INamedTypeSymbol, INamedTypeSymbol> containingTypeMap = new Dictionary<INamedTypeSymbol, INamedTypeSymbol>();
```
Evangelink, line 47: Is it ok to use `Dictionary` with the analyzer being concurrent? Maybe there was a different explanation but I remember @mavasani requested me to use some other collection (can't remember which one precisely).

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: Shall we go for a message like `Replace this call with a call to 'IsKind'` or `Prefer calls to 'IsKind' over 'Kind()'`?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
sharwell: I'm trying out a new pattern where test _inputs_ are provided by parameters, but expected _outputs_ are provided by switch expressions.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
sharwell: `Dictionary<TKey, TValue>` is safe for concurrent reads as long as there are never any writes.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
mavasani: Unless some other analyzer author writes an analyzer with same ID that reports diagnostic at different location?

mavasani:
@sharwell What is the status on this PR? Is it still draft?

sharwell:
I was not able to consistently measure a performance improvement. In theory, code using `IsKind` should be faster than code using `Kind`.

Evangelink:
I am pretty sure I did saw performance improvement when using `IsKind` instead of `Kind()` but we have less and less syntax only analyzers now.

@sharwell Do you want to keep it as-is for now or maybe discuss it with people and move the PR foward?

CyrusNajmabadi:
This is valuable enough purely in that it provides consistency to the IDE codebase and removes the need for this topic to come up in PRs.  Even without strict perf data, that's good enough for me to take this.  Thanks!

sharwell:
The performance difference between the two comes down to this:

```
    L0000: add edx, 0xfffffffe
    L0003: cmp edx, 0x1ffe
    L0009: seta al
```

vs this (`IsKind`):

```
    L0000: test edx, edx
    L0002: setne al
```

Of course the full code paths are longer, but this is the part that differs between the two.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Microsoft.CodeAnalysis.CSharp.Analyzers.MetaAnalyzers.Fixers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(CSharpPreferIsKindFix))]
+    [Shared]
+    public sealed class CSharpPreferIsKindFix : PreferIsKindFix
+    {
+        protected override async Task<Document> ConvertKindToIsKindAsync(Document document, TextSpan sourceSpan, CancellationToken cancellationToken)
+        {
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            var binaryExpression = root.FindNode(sourceSpan, getInnermostNodeForTie: true).FirstAncestorOrSelf<BinaryExpressionSyntax>();
```
CyrusNajmabadi, line 20: always good to have a test where this is an argument to ensure the `getInnermostNodeForTie:` is warranted.  (if you already have that test, aces).

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
CyrusNajmabadi: IMO, i think it's valid for diagnostic IDEs to claim squatting rights.  other libraries should not produce these IDs, and fixers should not have to be resilient to someone trying to spoof the diagnsotic.

this does't hold for compiler errors of course, as they're free to change those will nilly.  but for bound analyzer/fixer libs, i would prefer that philosophy myself.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
```diff
@@ -0,0 +1,57 @@
...
+            if (negate)
+            {
+                newRoot = root.ReplaceNode(binaryExpression, SyntaxFactory.PrefixUnaryExpression(SyntaxKind.LogicalNotExpression, newInvocation.WithoutLeadingTrivia()).WithLeadingTrivia(newInvocation.GetLeadingTrivia()));
+            }
+            else
+            {
+                newRoot = root.ReplaceNode(binaryExpression, newInvocation);
+            }
+
+            return document.WithSyntaxRoot(newRoot);
```
CyrusNajmabadi, line 42: i didn't look deeply at the VB side, but this looked duplicative.  consider SyntaxGenerator, or pulling some logic up.  If there isn't enough helper code to do that, or the logic really needs to be tweaked enough for both, disregard.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+        private static void HandleBinaryOperation(OperationAnalysisContext context, Dictionary<INamedTypeSymbol, INamedTypeSymbol> containingTypeMap)
+        {
+            var operation = (IBinaryOperation)context.Operation;
+            if (operation.OperatorKind != BinaryOperatorKind.Equals
+                && operation.OperatorKind != BinaryOperatorKind.NotEquals)
+            {
+                return;
+            }
+
+            if (operation.LeftOperand.WalkDownConversion() is IInvocationOperation { TargetMethod: { Name: "Kind", ContainingType: var containingType } }
```
CyrusNajmabadi, line 76: maybe assert something aobut hte argument count?

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
sharwell: I wasn't able to find a case where it is needed.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
sharwell: Right now there isn't much exact duplicate code, so the details can be left for a follow up.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
sharwell: If we verify the parameter count, the pattern becomes:

```
IInvocationOperation { TargetMethod: { Name: "Kind", ContainingType: var containingType } and ({ ReducedFrom: null, Parameters: { Length: 1 } } or { ReducedFrom: not null, Parameters: { IsEmpty: true } }) }
```

I wrote the simpler form on the assumption that the symbols with this name inside these well known types only have the expected signatures.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
CyrusNajmabadi: Do this:

`Foo(x.Kind() == SyntaxKind.Whatever`

With this, getInnermost is needed as otherwise FindNode will find the ArgumentSyntax.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
CyrusNajmabadi: Wfm

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
CyrusNajmabadi: Argument vs expression is my go to case wherever I want to test that.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
sharwell: Even if it doesn't hold, you'd just get a gold bar and everything else keeps working.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/Fixers/CSharpPreferIsKindFix.cs
sharwell: This was my first thought, but the diagnostic is reported on the left child of the binary expression, not the full binary expression.

CyrusNajmabadi: ah, gotcha.  in that case, you're likely fine.

CyrusNajmabadi (approved):

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: ➡️ Fixing this along with https://github.com/dotnet/roslyn-analyzers/pull/3537#discussion_r413295551

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: ➡️ Fixing this along with https://github.com/dotnet/roslyn-analyzers/pull/3537#discussion_r413296639

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: ➡️ Fixing this along with https://github.com/dotnet/roslyn-analyzers/pull/3537#discussion_r413296639

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: ➡️ Fixing this along with https://github.com/dotnet/roslyn-analyzers/pull/3537#discussion_r413296639

sharwell: ➡️ Reworded the code fix message

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: Reworded the description to include much more detail.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+            description: s_localizableDescription,
+            helpLinkUri: null,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```
Youssef1313, line 35: Other analyzers in this category analyzes auto-generated code.

https://github.com/dotnet/roslyn-analyzers/blob/b148fa89c974a72c28184931767835709f34d1fa/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.cs#L18

For both cases, I'm not sure whether it should be analyzed or not.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            context.RegisterCompilationStartAction(context =>
```
Youssef1313, line 37: When I did this, I only did `RegisterOperationAction` and accessed the compilation from there. May I ask what's the difference between both approaches?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        private static void HandleBinaryOperation(OperationAnalysisContext context, Dictionary<INamedTypeSymbol, INamedTypeSymbol> containingTypeMap)
+        {
+            var operation = (IBinaryOperation)context.Operation;
+            if (operation.OperatorKind is not (BinaryOperatorKind.Equals or BinaryOperatorKind.NotEquals))
+            {
+                return;
+            }
+
+            if (operation.LeftOperand.WalkDownConversion() is IInvocationOperation { TargetMethod: { Name: "Kind", ContainingType: var containingType } }
```
Youssef1313, line 75: Consider reporting a diagnostic when the call is on the right side, e.g: `SyntaxKind.IfKeyword == token.Kind()`.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
```diff
@@ -0,0 +1,423 @@
...
+    Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers.PreferIsKindAnalyzer,
+    Microsoft.CodeAnalysis.VisualBasic.Analyzers.MetaAnalyzers.CodeFixes.BasicPreferIsKindFix>;
+
+namespace Microsoft.CodeAnalysis.Analyzers.UnitTests.MetaAnalyzers
+{
+    public class PreferIsKindAnalyzerTests
+    {
+        [Theory]
+        [InlineData("==")]
+        [InlineData("!=")]
```
Youssef1313, line 20: Consider testing `is` and `is not` operators as well. Probably with C# 9.0 patterns (`and`, `or`).

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
```diff
@@ -0,0 +1,423 @@
...
+";
+            var fixedSource =
+@"using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+class C
+{
+    bool Method(SyntaxNode node)
+    {
+        return !node.IsKind(SyntaxKind.None) && !node.IsKind(SyntaxKind.TrueKeyword);
```
Youssef1313, line 119: If there are overloads that take multiple kinds, it would look nicer to use them. But that's not a necessary change to do, especially if it will increase the complexity and decrease performance due its limited value.

I'd do that only if it's easy and cheap enough.

src/Microsoft.CodeAnalysis.Analyzers/VisualBasic/MetaAnalyzers/Fixers/BasicPreferIsKindFix.vb
```diff
@@ -0,0 +1,59 @@
...
+Imports Microsoft.CodeAnalysis.Text
+Imports Microsoft.CodeAnalysis.VisualBasic.Syntax
+
+Namespace Microsoft.CodeAnalysis.VisualBasic.Analyzers.MetaAnalyzers.CodeFixes
+    <ExportCodeFixProvider(LanguageNames.VisualBasic, Name:=NameOf(BasicPreferIsKindFix))>
+    <[Shared]>
+    Public NotInheritable Class BasicPreferIsKindFix
+        Inherits PreferIsKindFix
+
+        Protected Overrides Async Function ConvertKindToIsKindAsync(document As Document, sourceSpan As TextSpan, cancellationToken As CancellationToken) As Task(Of Document)
```
Youssef1313, line 16: Consider, if/when appropriate, extracting as much as possible in the base class.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
CyrusNajmabadi: i would not analyze auto-generated code.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
CyrusNajmabadi: the difference hre is that we can do an intiial check that decides if we should even be called back into later.  If we don't see what we expect in the compilation, we don't need to execute any work later.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
CyrusNajmabadi: enh. :)

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
CyrusNajmabadi: i actually think if you're doing `and` then this form is likely nicer to read :)

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
sharwell: For this case, it just doesn't matter enough.

src/Microsoft.CodeAnalysis.Analyzers/VisualBasic/MetaAnalyzers/Fixers/BasicPreferIsKindFix.vb
sharwell: ➡️ Also requested in https://github.com/dotnet/roslyn-analyzers/pull/3537#discussion_r555456764. I'm not a fan of `SyntaxGenerator` compared to `SyntaxFactory` due to its lack of fidelity, so there isn't much code eligible for sharing at this point.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
sharwell: Agree this case should have been handled.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
sharwell: ➡️ This approach ensures minimal overhead (no IOperation callbacks) in cases where the analyzer is installed in a project that doesn't reference Roslyn.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
sharwell: ➡️ Decided not to include this one for now because it's a relatively uncommon pattern and it adds substantial complexity to the code fix.

Youssef1313: > ➡️ Decided not to include this one for now because it's a relatively uncommon pattern and it adds substantial complexity to the code fixes.

Do you want to file an issue that tracks handling this later?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
sharwell: This analyzer is implemented for all Roslyn users, but those methods are not public APIs so I didn't include them in the analysis.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
sharwell: I'm excluding these for now since this analyzer package doesn't reference a new enough version of Microsoft.CodeAnalysis.Common to include `INegatedPatternOperation`.

PR #3539, opened 2020-04-22
Add tooling support to generate files in the repo for analyzer rules …

mavasani:
…that are missing documentation

Generating/updating these files is disabled by default as it needs querying the web and would make CI builds slower. These can be manually refreshed by passing in `/t:pack /p:GenerateAnalyzerRulesMissingDocumentationFile=true` to the msbuild command line for the solution

I will be filing separate issues to track adding documentation for CA rules.


mavasani:
Filed tracking issues to add missing documentation: https://github.com/dotnet/roslyn-analyzers/issues/3548, https://github.com/dotnet/roslyn-analyzers/issues/3549 and https://github.com/dotnet/roslyn-analyzers/issues/3550

PR #3542, opened 2020-04-22
Fix nullability warnings on DoNotCopyValue found with latest update

buyaa-n:
My local build failing with this warnings since pulled latest from master


buyaa-n:
> It sounds like you are building with an older version of the compiler

Not sure how I could fix that, anyways seems I should close this PR

mavasani:
@sharwell It seems like this is now showing up on the toolset that is enabled with Arcade: https://dev.azure.com/dnceng/public/_build/results?buildId=621338&view=logs&j=f0f0520d-467b-5d1d-a95d-4e3a2581fe8a&t=c140c00f-6c0a-5356-d099-4e047a5525ed. I am going to add suppressions for these violations and file a tracking issue to remove these suppressions once we move to Arcade with a newer toolset.

sharwell:
@mavasani We can just update the compiler toolset with a property

Issue #3528, opened 2020-04-20
New analyzer: Report on call to 'HasFlag' on enums not marked with 'FlagsAttribute'


Evangelink:
Add an analyzer that will report a diagnostic on a call of `HasFlag` method on an enum without the `FlagsAttribute` and suggest direct equality check instead.

```csharp
public class C
{
	public enum MyEnum { A, B, }

	public void Do(MyEnum m)
	{
		m.HasFlag(MyEnum.B); // Report here
	}
}
```

Evangelink:
Fixed by #3557

PR #3557, opened 2020-04-27
CA2248: Provide correct 'enum' argument to 'Enum.HasFlag'

### LINKED ISSUE #3529: New analyzer: Enum types on left and right side of the 'HasFlag' should be the same
Add an analyzer that will report a diagnostic on a call to the `HasFlag` method for which the enum type on which it is invoked is not the same as the one passed as argument.

```csharp
public class C
{
	[Flags]
	public enum MyEnum { A, B, }
	[Flags]
	public enum OtherEnum { A, }

	public void Do(MyEnum m)
	{
		m.HasFlag(OtherEnum.A); // Report here
	}
}
```

Evangelink:
Fix #3529


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCheckFlagFromDifferentEnum.cs
```diff
@@ -0,0 +1,53 @@
...
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Usage
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotCheckFlagFromDifferentEnum : DiagnosticAnalyzer
```
Evangelink, line 13: I don't really like this class name. Any suggestion?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCheckFlagFromDifferentEnum.cs
```diff
@@ -0,0 +1,53 @@
...
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCheckFlagFromDifferentEnumTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCheckFlagFromDifferentEnumMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCheckFlagFromDifferentEnumDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Usage,
+            RuleLevel.BuildWarning,
```
Evangelink, line 26: I don't think there could be much FP nor performance downside so I have enabled this rule as `BuildWarning` let me know if you prefer something else.

mavasani: No rule should be made a BuildWarning before creating a separate issue that has approval from dotnet-analyzers team. Please lower the severity here to IdeSuggestion at the max.

Evangelink:
Ping @mavasani

mavasani:
FYI as this is a new CA rule: Please see the updated **Documentation requirements** section in https://github.com/dotnet/roslyn-analyzers/blob/master/GuidelinesForNewRules.md.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/DoNotCheckFlagFromDifferentEnumTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    Public Sub Method(ByVal m As MyEnum)
+        m.HasFlag(MyEnum.A)
+    End Sub
+End Class
+");
+        }
+
+        [Fact]
+        public async Task CA2248_EnumTypesAreSameButNotFlag_NoDiagnostic()
```
mavasani, line 93: Can you also add a test where enum types are different by neither enum types have flags attribute to complete the test matrix?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1314,4 +1314,13 @@
   <data name="PreferConstCharOverConstUnitStringInStringBuilderTitle" xml:space="preserve">
     <value>Consider using 'StringBuilder.Append(char)' when applicable.</value>
   </data>
+  <data name="DoNotCheckFlagFromDifferentEnumDescription" xml:space="preserve">
+    <value>Calling the 'HasFlag' method on enum of some type to check the flag of another enum type will always return 'false'.</value>
```
mavasani, line 1318: I tried the below console app and it always throws an unhandled exception:

```csharp
using System;

public class C
{
    [Flags]
    public enum MyEnum { A, B, }
    [Flags]
    public enum OtherEnum { A, }

    {
        MyEnum a = MyEnum.A;
        Console.WriteLine(a.HasFlag(OtherEnum.A));
    }
}
```

```
Unhandled exception. System.ArgumentException: The argument type, 'C+OtherEnum', is not the same as the enum type 'C+MyEnum'.
   at System.Enum.HasFlag(Enum flag)
```

mavasani: We should fix the description.

mavasani: Something like: `'Enum.HasFlag' method expects the 'Enum' argument to be of the same type as the 'Enum' instance on which the method is invoked. If these are different 'Enum' types, an unhandled exception will be thrown at runtime.`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1314,4 +1314,13 @@
...
+  </data>
+  <data name="DoNotCheckFlagFromDifferentEnumMessage" xml:space="preserve">
+    <value>Change enum type either on left-hand side or right-hand side of the 'HasFlag' method call</value>
```
mavasani, line 1321:
```suggestion
    <value>'Enum' argument should be of type '{0}'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1314,4 +1314,13 @@
...
+  </data>
+  <data name="DoNotCheckFlagFromDifferentEnumTitle" xml:space="preserve">
+    <value>Flag enum type should be the same has containing enum type</value>
```
mavasani, line 1324:
```suggestion
    <value>Provide correct argument to 'Enum.HasFlag'</value>
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/DoNotCheckFlagFromDifferentEnumTests.cs
mavasani: It seems runtime throws an exception when comparing different non-flags enum types as well. So this rule should not have any dependence on whether the enum is marked as Flags or not (which seems to be the case in rule implementation, but not Title)

```csharp
using System;

public class C
{
    public enum MyEnum { A, B, }
    public enum OtherEnum { A, }

    static void Main()
    {
        Console.WriteLine(MyEnum.A.HasFlag(OtherEnum.A));
    }
}
```

Throws the same unhandled exception at runtime.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCheckFlagFromDifferentEnum.cs
mavasani: How about `ProvideCorrectArgumentToEnumHasFlag`?

Evangelink: I think I like `ProvideCorrectArgumentToEnumHasFlag` more as it gives more info about the context.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
Evangelink: Sorry I got confused by another ticket I have created: https://github.com/dotnet/roslyn-analyzers/issues/3528

To sum-up if the enum types on the arg and caller are different the method throws at runtime (whether or not the enums are marked with `Flags`) but if you call the method on a type which isn't marked with `Flags` the method always returns false.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/DoNotCheckFlagFromDifferentEnumTests.cs
Evangelink: Done! Thanks for catching-up this case.

Evangelink:
Doc ticket + PR are done and I have addressed the comment reviews you have left.

nschuessler:
We have found performance issues with `Enum.HasFlag` and recommend people just do the bitwise operator to check instead.  Consider should enum.hasflag just be discouraged?

mavasani:
> We have found performance issues with `Enum.HasFlag` and recommend people just do the bitwise operator to check instead `((e & MyEnum.Something) == MyEnum.Something)`.  Consider should enum.hasflag just be discouraged?

Tagging @stephentoub and @bartonjs for their views. Should `Enum.HasFlag` be completely discouraged due to performance reasons or do we feel the current analyzer implementation in this PR to flag incorrect usages of the API is preferable?

stephentoub:
> Should Enum.HasFlag be completely discouraged due to performance reasons

As for .NET Core 2.1, it's now recognized / optimized by the JIT into the equivalent bit flag checks, so it's fine to use.

Evangelink:
@mavasani Looks like we can move forward with this PR. I would suggest creating a ticket (same analyzer?) to suggest not using `HasFlag` prior to netcore 2.1, WDYT?

stephentoub:
> I would suggest creating a ticket (same analyzer?) to suggest not using HasFlag prior to netcore 2.1, WDYT?

I don't think it's worth adding an analyzer for performance issues already fixed in the platform several releases ago.

Evangelink:
@stephentoub I understand that this would increase the number of analyzers to support but I suspect quite a lot of users (especially in big companies) are still running code with non netcore target. And btw, I know .NET 5 is on its way :)

Anyways I'll let you decide what's worth :) In the meantime I will go open a PR to replace calls to `HasFlag` in some projects.

mavasani:
Agree with @stephentoub. We shouldn't be adding analyzer based on TFMs, discouraging an API based on issues in earlier shipped version. Such an analyzer should be authored in another custom analyzer package outside this repo.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterOperationAction(context =>
+            {
+                var invocation = (IInvocationOperation)context.Operation;
+
+                if (invocation.TargetMethod.Name == "HasFlag" &&
+                    invocation.TargetMethod.ContainingType.SpecialType == SpecialType.System_Enum &&
```
mavasani, line 43: Minor: Making this the first check would likely rule out majority of invocations and avoid unnecessary target method name comparisons.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1323,4 +1323,13 @@
   <data name="PreferConstCharOverConstUnitStringInStringBuilderTitle" xml:space="preserve">
     <value>Consider using 'StringBuilder.Append(char)' when applicable.</value>
   </data>
-</root>
+  <data name="ProvideCorrectArgumentToEnumHasFlagDescription" xml:space="preserve">
+    <value>'Enum.HasFlag' method expects the 'enum' argument to be of the same 'enum' type as the instance on which the method is invoked. If these are different 'enum' types, an unhandled exception will be thrown at runtime.</value>
+  </data>
+  <data name="ProvideCorrectArgumentToEnumHasFlagMessage" xml:space="preserve">
+    <value>The argument type, '{0}', is not the same as the enum type '{1}'</value>
```
mavasani, line 1330:
```suggestion
    <value>The argument type, '{0}', must be the same as the enum type '{1}'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
Evangelink: I'll also move the arguments length check before then!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                var invocation = (IInvocationOperation)context.Operation;
+
+                if (invocation.TargetMethod.ContainingType.SpecialType == SpecialType.System_Enum &&
+                    invocation.Arguments.Length == 1 &&
+                    invocation.TargetMethod.Name == "HasFlag" &&
+                    invocation.Arguments[0].Value is IConversionOperation conversion &&
+                    !invocation.Instance.Type.Equals(conversion.Operand.Type))
+                {
+                    context.ReportDiagnostic(invocation.CreateDiagnostic(Rule));
```
mavasani, line 48:
```suggestion
                    context.ReportDiagnostic(invocation.CreateDiagnostic(Rule, invocation.Arguments[0].Type.Name, invocation.Instance.Type.Name));
```

mavasani: Can we also add a non-markup based test to guard this?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            context.RegisterOperationAction(context =>
+            {
+                var invocation = (IInvocationOperation)context.Operation;
+
+                if (invocation.TargetMethod.ContainingType.SpecialType == SpecialType.System_Enum &&
+                    invocation.Arguments.Length == 1 &&
+                    invocation.TargetMethod.Name == "HasFlag" &&
+                    invocation.Arguments[0].Value is IConversionOperation conversion &&
+                    !invocation.Instance.Type.Equals(conversion.Operand.Type))
```
mavasani, line 46: Curious: Do we need to guard for null `Instance` for static method? https://github.com/dotnet/roslyn/blob/b61d9d2863eebfc76b032e5f8fcd0a918fb27147/src/Compilers/Core/Portable/Generated/Operations.Generated.cs#L631-L634

mavasani:
```suggestion
                    invocation.Instance != null &&
                    !invocation.Instance.Type.Equals(conversion.Operand.Type))
```

mavasani:
Ping @Evangelink

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
```diff
@@ -0,0 +1,54 @@
...
+                var invocation = (IInvocationOperation)context.Operation;
+
+                if (invocation.TargetMethod.ContainingType.SpecialType == SpecialType.System_Enum &&
+                    invocation.Arguments.Length == 1 &&
+                    invocation.Instance != null &&
+                    invocation.TargetMethod.Name == "HasFlag" &&
+                    invocation.Arguments[0].Value is IConversionOperation conversion &&
+                    !invocation.Instance.Type.Equals(conversion.Operand.Type))
+                {
+                    context.ReportDiagnostic(invocation.CreateDiagnostic(Rule, conversion.Operand.Type.Name, invocation.Instance.Type.Name));
```
mavasani, line 49: Can you try a test with `null` argument, i.e. `a.HasFlag(null)`? I think this will probably throw a null reference exception as `conversion.Operand.Type` is null. If so, can you add an additional null check for `conversion.Operand.Type` to the if condition?

Evangelink: Thanks! At runtime this code will cause an `ArgumentNullException`, do you want the rule to ignore this case? To report with `<null>/<Nothing>` instead of the type? Or do you prefer a new descriptor for this case?

mavasani: I think using a new descriptor is an overkill. I am fine with bailing out for this case OR reporting a diagnostic using `<null>/<Nothing>` as the first argument.

Issue #3525, opened 2020-04-20
Ideas for resources (resx) analyzers


Evangelink:
Add a new analyzer for roslyn-analyzers that will check for resources keys and values.

## Rules for keys

1. Key should end with `Title`, `Message`, `Description` or `CodeFixTitle`.

## Rules for values

1. If the value is linked to a title or message key, then it shouldn't end with a period '.'.

2. If the value is linked to a description, it should end with a period '.'.

3. Placeholders should be surrounded with single-quotes.

Tagging @mavasani for other ideas.

Evangelink:
Regarding keys, for rules with different messages, the current pattern is `MessageSomeText` and `MessageSomeOtherText`, could we use something else that would make it easier to have an automated detection? Like `Message_SomeText` or `SomeTextMessage`?

The analyzer could also be checking for `[Something]` or `[SomethingAttribute]` and report on it as this is a C# syntax.

mavasani:
@Evangelink - do you think the rule for keys would be possibly to enforce? I do foresee resource strings outside these 4 buckets? The core concern for me is that it would not be possible to add a source suppression for these cases as it is defined in a resx file.

mavasani:
@Evangelink - I'd say lets start with the analyzer that enforces what you suggested for rule values for now.

Evangelink:
@mavasani What about reporting on calling site instead of resource definition?

mavasani:
> What about reporting on calling site instead of resource definition?

Ah, that seems to be a good idea :) Please do add a good description, and if possible a code fix. I am presuming you would anyways need such an analyzer to require the resx file passed in as an additional file so you can have code fixes that automatically fix the issue instead of manual fixes.

mavasani:
@Evangelink Feel free to get started here. Note that analysis of resx files is a pretty desired bucket of rules. For example, finding duplicate resx strings, flagging unused resx strings, etc. would be super helpful as well.

Evangelink:
@mavasani It'd be nice if we could define the next steps for this analyzer, it seems to be pretty important.

mavasani:
Yeah, hoping to get to it this week...

PR #3566, opened 2020-04-28
[WIP] resx analyzer

Evangelink:
Contributes to #3525

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/Core/AnalyzerReleases.Unshipped.md
+++ b/src/Roslyn.Diagnostics.Analyzers/Core/AnalyzerReleases.Unshipped.md
@@ -6,6 +6,7 @@ RS0042 | RoslynDiagnosticsReliability | Warning | DoNotCopyValue
 RS0043 | RoslynDiagnosticsMaintainability | Warning | DoNotCallGetTestAccessor
 RS0044 | RoslynDiagnosticsMaintainability | Hidden | CreateTestAccessor
 RS0045 | RoslynDiagnosticsMaintainability | Hidden | ExposeMemberForTesting
+RS0047 | RoslynDiagnosticsDesign | Info | DefineResourceEntryCorrectly
 
 ### Changed Rules
 Rule ID | New Category | New Severity | Old Category | Old Severity | Notes
--- a/src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
@@ -1,0 +1,96 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Immutable;
+using System.IO;
+using System.Text.RegularExpressions;
+using System.Xml;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Roslyn.Diagnostics.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DefineResourceEntryCorrectly : DiagnosticAnalyzer
+    {
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.DefineResourceEntryCorrectlyTitle), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.DefineResourceEntryCorrectlyMessage), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.DefineResourceEntryCorrectlyDescription), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            RoslynDiagnosticIds.DefineResourceEntryCorrectlyRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.RoslynDiagnosticsDesign,
+            DiagnosticSeverity.Info,
+            isEnabledByDefault: true,
+            description: s_localizableDescription,
+            helpLinkUri: null,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private static readonly Regex s_dataNameRegex = new Regex("<data.*?name=\"(.*?)\"", RegexOptions.Compiled);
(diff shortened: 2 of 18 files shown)


src/Roslyn.Diagnostics.Analyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -6,6 +6,7 @@ RS0042 | RoslynDiagnosticsReliability | Warning | DoNotCopyValue
 RS0043 | RoslynDiagnosticsMaintainability | Warning | DoNotCallGetTestAccessor
 RS0044 | RoslynDiagnosticsMaintainability | Hidden | CreateTestAccessor
 RS0045 | RoslynDiagnosticsMaintainability | Hidden | ExposeMemberForTesting
+RS0047 | RoslynDiagnosticsDesign | Info | DefineResourceEntryCorrectly
```
Evangelink, line 9: https://github.com/dotnet/roslyn-analyzers/pull/3352 is taking `RS0046`

src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
```diff
@@ -0,0 +1,96 @@
...
+                    context.CancellationToken.ThrowIfCancellationRequested();
+
+                    var fileExtension = Path.GetExtension(file.Path);
+                    if (!fileExtension.Equals(".resx", StringComparison.OrdinalIgnoreCase))
+                    {
+                        continue;
+                    }
+
+                    var isDescription = false;
+                    var sourceText = file.GetText(context.CancellationToken);
```
Evangelink, line 57: I have tried a couple of resx/xml loader but we either don't have the right dll or this is failing because of the first line of the resx which doesn't seem to be xml compliant. Let me know if you know a better way that doing line by line reading.

src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
```diff
@@ -0,0 +1,96 @@
...
+            DiagnosticSeverity.Info,
+            isEnabledByDefault: true,
+            description: s_localizableDescription,
+            helpLinkUri: null,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private static readonly Regex s_dataNameRegex = new Regex("<data.*?name=\"(.*?)\"", RegexOptions.Compiled);
+        private static readonly Regex s_valueRegex = new Regex("<value>(.*?)</value>", RegexOptions.Compiled);
```
Evangelink, line 36: It's never 100% clear to me when `Compiled` is better.

Evangelink, line 36: FYI `.*?` allows a non-greedy version of `.*` this prevents unwanting characters! Besides, it's usually better in term of performances because it reduces the backtracking.

src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
```diff
@@ -0,0 +1,96 @@
...
+                        continue;
+                    }
+
+                    var isDescription = false;
+                    var sourceText = file.GetText(context.CancellationToken);
+                    foreach (var line in sourceText.Lines)
+                    {
+                        var text = line.ToString().Trim();
+
+                        var match = s_dataNameRegex.Match(text);
```
Evangelink, line 62: If we go for this line by line check, do we want to have a `StartsWith` check before the regex match?

Evangelink:
Tagging @mavasani and @sharwell to get your opinion on whether or not the direction looks good to you?!

src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                        match = s_valueRegex.Match(text);
+                        if (match.Success)
+                        {
+                            var endsWithPeriod = match.Groups[1].Value.EndsWith(".", s_ResxContentStringComparison);
+
+                            if (endsWithPeriod && !isDescription)
+                            {
+                                var linePositionSpan = sourceText.Lines.GetLinePositionSpan(line.Span);
+                                var location = Location.Create(file.Path, line.Span, linePositionSpan);
```
Evangelink, line 77: Do we want to report on the full line or only the `.`?

src/Roslyn.Diagnostics.Analyzers/Core/DefineResourceEntryCorrectly.cs
```diff
@@ -0,0 +1,96 @@
...
+                                context.ReportDiagnostic(Diagnostic.Create(Rule, location));
+                            }
+                            else if (isDescription && !endsWithPeriod)
+                            {
+                                var linePositionSpan = sourceText.Lines.GetLinePositionSpan(line.Span);
+                                var location = Location.Create(file.Path, line.Span, linePositionSpan);
```
Evangelink, line 83: Where do we want to report? Full line? The last character of the sentence?

mavasani:
@Evangelink Looking at the current implementation, I think my proposed idea of reporting diagnostics on resx files has few big drawbacks:

1. We cannot implement or offer a code fix. Code fixes are currently only supported for diagnostics on source file
2. CompilationAction: This diagnostic will not be reported in live analysis when you open a resx file. This is unfortunately the drawback of diagnostics with location in additional files, which I had not completely thought through.
3. False positives: We are assuming the relationship between between resource strings and their usage in analyzer based on just the resx name suffix, like Title, Description, etc. This makes it not useful as a generic analyzer for all analyzer authors.

I think we should take the following direction:

1. Lets target this analyzer should be added to 'Microsoft.CodeAnalysis.Analyzers', so it can be used by all analyzer authors.
2. Instead of directly reading resx and flagging the resource strings, let us try to approach this from the view of types deriving from DiagnosticAnalyzer, so we generate diagnostics in these source files. More specifically:
   1. Enhance `DiagnosticDescriptorCreationAnalyzer.AnalyzeTitle`, so if title is a constant string, and has any periods or newlines, we report a diagnostic that title is recommended to be a short string without periods or newlines.
   2. For cases where title argument is not a constant string, but a LocalizableString, we need to handle 2 cases:
      1. Argument is an object creation of type LocalizableResourceString. We can process the arguments to this creation and identify the name of resource string used for title (const string for the 1st argument) + name/type of the resources file (System.Type argument for the 3rd argument). We can map this to the resource string in resx file and then generate a similar diagnostic if it has periods or newlines.
      2. Argument is a field reference of type LocalizableString or LocalizableResourceString or const string. For const string, it should be easy to report the diagnostic. For LocalizableString or LocalizableResourceString, we need to actually analyze the field analyzer. We can setup this analysis by registering a SymbolStartAction for named types. We register an Operation action with OperationKind.FieldInitializer and use this callback to analyze each field initializer of type `LocalizableResourceString`, and build a map from field symbol to a tuple `(string resourceName, string resxFileName)`. In `DiagnosticDescriptorCreationAnalyzer.AnalyzeTitle` callback, we build a set of field symbols of type LocalizableString or LocalizableResourceString that are used as title arguments to descriptors. We register a SymbolEndAction which analyzes both these data structures to then link which resx string is used as title argument, and then report a similar diagnostic in this end action. NOTE: We should probably guard this analysis from executing only if at least one resx file was provided as an additional file, otherwise we cannot report any diagnostics here.
3. We can repeat similar analysis for Message and Description arguments as done for Title above, but with different validation: Message should not have any newline + either it should have no periods or two or more periods. Description should always have a period to end.
4. All the above diagnostics would be enabled by default, and also valid for any third party analyzer.
5. Additionally, we can add separate diagnostic descriptors for: Resource strings used for title, message or description should always end with "Title", "Message", "Description" respectively. I am in two minds if these diagnostics will be too noisy or not - we should maybe have these off by default and just enable them for our repo.
6. Code Fix: We should write a code fix that fixes all the above cases, with a FixAll operation. That is going to be the biggest benefit here.

Sorry for the long comment, we can discuss more offline for further ideas.

PR #3574, opened 2020-04-29
More details on tracking nullability

jcouv:
FYI @RussKie (thanks for the dogfooding and feedback)

### DIFF
--- a/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md
+++ b/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md
@@ -25,6 +25,12 @@ To enable support for [nullable reference types](https://docs.microsoft.com/dotn
 #nullable enable
 ```
 
+At that point, reference types in annotated code will need to be annotated with either a `?` (nullable) or a `!` (non-nullable). For instance, `C.AnnotatedMethod(string! nonNullableParameter, string? nullableParameter, int valueTypeParameter) -> void`.
+
+Any public API that uses an oblivious reference type (in `#nullable disable` or legacy code) will be tracked with a `~` marker. The marker lets you track how many public APIs still lack annotations. For instance, `~C.ObliviousMethod() -> string`.
+
+If you start with a fresh project or you reach 100% annotation on your public API on an existing project, the RS0041 warning ("Symbol uses some oblivious reference types") ensures that all public APIs remain annotated. If you are in the process of annotating an existing project, it is recommended to disable this warning (for instance in `.editorconfig` with `dotnet_diagnostic.RS0041.severity = none`) until you completely annotate the project.
+
 ## Conditional API Differences
 
 Sometimes APIs vary by compilation symbol such as target framework.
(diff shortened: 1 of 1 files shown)


src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md
```diff
@@ -25,6 +25,12 @@ To enable support for [nullable reference types](https://docs.microsoft.com/dotn
 #nullable enable
 ```
 
+At that point, reference types in annotated code will need to be annotated with either a `?` (nullable) or a `!` (non-nullable). For instance, `C.AnnotatedMethod(string! nonNullableParameter, string? nullableParameter, int valueTypeParameter) -> void`.
+
+Any public API that uses an oblivious reference type (in `#nullable disable` or legacy code) will be tracked with a `~` marker. The marker lets you track how many public APIs still lack annotations. For instance, `~C.ObliviousMethod() -> string`.
```
RussKie, line 30:
```suggestion
Any public API that haven't been annotated (i.e. uses an oblivious reference type) will be tracked with a `~` marker (irrespective of presence of `#nullable disable` in files). The marker lets you track how many public APIs still lack annotations. For instance, `~C.ObliviousMethod() -> string`.
```

src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md
```diff
@@ -25,6 +25,12 @@ To enable support for [nullable reference types](https://docs.microsoft.com/dotn
...
+
+If you start with a fresh project or you reach 100% annotation on your public API on an existing project, the RS0041 warning ("Symbol uses some oblivious reference types") ensures that all public APIs remain annotated. If you are in the process of annotating an existing project, it is recommended to disable this warning (for instance in `.editorconfig` with `dotnet_diagnostic.RS0041.severity = none`) until you completely annotate the project.
```
RussKie, line 32:
```suggestion
We recommend to enable [RS0041 warning](https://github.com/dotnet/roslyn-analyzers/blob/master/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md) if you start with a fresh project or your project has reached 100% annotation on your public API to ensure that all public APIs remain annotated. 
If you are in the process of annotating an existing project, we recommended to disable this warning until you complete the annotation. The rule can be disabled via `.editorconfig` with `dotnet_diagnostic.RS0041.severity = none`.
```

RussKie:
For the builds using Arcade SDK all rules are stored in `eng/Shipping.ruleset`, and changes to `.editorconfig` are ignored.
```diff
diff --git a/eng/Shipping.ruleset b/eng/Shipping.ruleset
index 8913c6738..8f2d0a82f 100644
--- a/eng/Shipping.ruleset
+++ b/eng/Shipping.ruleset
@@ -178,4 +178,7 @@
     <Rule Id="SA1649" Action="None" />
     <Rule Id="SA1651" Action="None" />
   </Rules>
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.PublicApiAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.PublicApiAnalyzers">
+    <Rule Id="RS0041" Action="None" />
+  </Rules>
 </RuleSet>
```

I have found it via Visual Studio:
image

Perhaps it is worth add this to the docs as well.

mavasani:
@RussKie For the scenario where .editorconfig configuration does not work, but ruleset based configuration works, can you file a separate issue at https://github.com/dotnet/roslyn/issues/new with simpler, standalone repro? Note that respecting the configuration setting is not controlled by each analyzer, but rather by the compiler.

mavasani:
@jcouv Feel free to merge whenever you are fine with the PR state.

PR #3578, opened 2020-04-29
Refactor tests to have only the expected IDs raising on each test

Evangelink:
The `DiagnosticDescriptorCreationAnalyzer` has a lot of descriptors but tests are grouped per set of descriptors so the idea of this PR is to simplify test readability and maintainability by ensuring only a sub-set of descriptor is enabled for each test.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
@@ -21,6 +21,8 @@ namespace Microsoft.CodeAnalysis.Analyzers.UnitTests.MetaAnalyzers
     {
         #region RS1007 (UseLocalizableStringsInDescriptorRuleId) and RS1015 (ProvideHelpUriInDescriptorRuleId)
 
+        private static readonly DiagnosticDescriptor[] RS1007_RS1015_Descriptors = new[] { DiagnosticDescriptorCreationAnalyzer.UseLocalizableStringsInDescriptorRule, DiagnosticDescriptorCreationAnalyzer.ProvideHelpUriInDescriptorRule, };
+
         [Fact]
         public async Task RS1007_RS1015_CSharp_VerifyDiagnostic()
         {
@@ -48,7 +50,7 @@ class MyAnalyzer : DiagnosticAnalyzer
     {
     }
 }",
-                new[] { DiagnosticDescriptorCreationAnalyzer.UseLocalizableStringsInDescriptorRule, DiagnosticDescriptorCreationAnalyzer.ProvideHelpUriInDescriptorRule, },
+                RS1007_RS1015_Descriptors,
                 GetCSharpRS1007ExpectedDiagnostic(11, 9),
                 GetCSharpRS1015ExpectedDiagnostic(11, 9));
         }
@@ -77,7 +79,7 @@ Class MyAnalyzer
     End Sub
 End Class
 ",
-                new[] { DiagnosticDescriptorCreationAnalyzer.UseLocalizableStringsInDescriptorRule, DiagnosticDescriptorCreationAnalyzer.ProvideHelpUriInDescriptorRule, },
+                RS1007_RS1015_Descriptors,
                 GetBasicRS1007ExpectedDiagnostic(10, 66),
                 GetBasicRS1015ExpectedDiagnostic(10, 70));
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
```diff
@@ -1094,19 +1062,8 @@ End Class
...
-                        GetBasicRS1028ResultAt(14, 73),
-                        GetBasicRS1028ResultAt(15, 71),
-                        GetBasicRS1028ResultAt(16, 71),
-                        GetBasicRS1028ResultAt(17, 71),
-                        GetBasicRS1028ResultAt(18, 71),
-                        GetBasicRS1028ResultAt(19, 71),
-                    }
                 },
-                SolutionTransforms = { WithoutEnableReleaseTrackingWarning }
+                SolutionTransforms = { (solution, projectId) => WithoutOtherDiagnosticDescriptorRules(solution, projectId, new[] { DiagnosticDescriptorCreationAnalyzer.DiagnosticIdMustBeInSpecifiedFormatRule, DiagnosticDescriptorCreationAnalyzer.UseCategoriesFromSpecifiedRangeRule, }) },
```
mavasani, line 1066: Can we avoid each test having to specify the array by creating helpers, similar to the one that previously existed?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
```diff
@@ -63,22 +64,22 @@ Imports Microsoft.CodeAnalysis.Diagnostics
 
 <DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)>
 Class MyAnalyzer
-	Inherits DiagnosticAnalyzer
```
Evangelink, line 66: I was sure I reverted this change before doing the commit... Let me know if you prefer not to have these included here, that might take me some time to revert each line.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
Evangelink: That's a good idea!

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
mavasani: It is fine given we are getting rid of tabs.

Evangelink:
@sharwell Here is my thought:

1/ I understand your comment and if we wanted to really represent user experience I would suggest to have all rules enabled for each tests (whatever the rule we are testing). This would be the closest to the end-user experience.

2/ Reading and adding tests to this class is pretty weird IMO because it doesn't feel like a unit test. It's planned to add more cases caught by this analyzer and we end-up with quite some issues unrelated to what is being tested.

sharwell:
> I understand your comment and if we wanted to really represent user experience I would suggest to have all rules enabled for each tests (whatever the rule we are testing)

Just to make sure I understand the starting state, is this what we are already doing?

> Reading and adding tests to this class is pretty weird IMO because it doesn't feel like a unit test

I never really found the distinction between "unit test" and "not unit test" valuable for long-term maintainability. Analyzer testing is structured around scenarios, where we examine the complete behavior of one analyzer in a specific context.

Let me go over the current test structure and see how this might fit.

mavasani:
@sharwell The primary issue here is that this analyzer reports a large number of different diagnostics, and @Evangelink is planning to extend it further. He is attempting to isolate the unit tests so they test the behavior only for the intended diagnostic.

@Evangelink If we go ahead with this PR, please do make sure that any expected diagnostics that you are removing from tests have a corresponding unit test validating that diagnostic. Basically, this PR should not lead to loss of test coverage.

Evangelink:
> 
> 
> > I understand your comment and if we wanted to really represent user experience I would suggest to have all rules enabled for each tests (whatever the rule we are testing)
> 
> Just to make sure I understand the starting state, is this what we are already doing?
> 

My point was more to say, why not going further with that and say that _ALL_ analyzers present in this repo analyze all tests in this repo. Basically, all CA and RS rules would be run on all tests. This way we get away from the "unit-test" behavior. Of course this would lead to a big PR as I expect a lot of issues to be raised but this way you have a bigger safety net.

I would still be in favor, and that's only my opinion, to have "unit-tests" just checking the rule you want to check and setting up an infrastructure where we would run a beta/nightly version of the analyzer against some well known codes on GitHub. Of course this test infrastructure is a bit heavy, especially if you want to start saving diffs of the issues found by each version so you can find if a fix is triggering new faults, is reducing/increasing FPs... Globally such infrastructure could help you to decide if a rule can be promoted to `Warning` because it has a low FP-rate.

Anyways I'll let you decide what you think is best. I am available for an offline chat if you think that could help.

Issue #1864, opened 2018-10-23
"Do not ignore method results" gives incorrect advice for Win32 APIs


davkean:
This rule is firing on Win32 APIs saying that the user needs to consume the "hresult"[sic] passing it to Marshal.ThrowExceptionForHR. This incorrect, these APIs are not COM-based APIs, they return various results from NULL, to FALSE, to a non-zero value to indicate a failure, of which can be retrieved using `Mashal.GetLastWin32Error()`

```
Severity	Code	Description	Project	Project Rank	Path	File	Line	Column	Category	Source	Suppression State	Tool
Warning	CA2010	Consume the hresult returned by method 'Friend Declare Auto Function SendMessage Lib "user32" (hwnd As System.IntPtr, msg As Integer, wParam As System.IntPtr, lParam As System.IntPtr) As System.IntPtr' and call Marshal.ThrowExceptionForHR.	Microsoft.VisualStudio.Editors	3	E:\project-system2\src\Microsoft.VisualStudio.Editors\ResourceEditor	E:\project-system2\src\Microsoft.VisualStudio.Editors\ResourceEditor\ResourceListView.vb	717	21	Reliability	IntelliSense	Active	Microsoft.NetCore.VisualBasic.Analyzers
Warning	CA2010	Consume the hresult returned by method 'Public Declare Auto Function SetWindowLong Lib "user32" (hWnd As System.IntPtr, Index As Integer, Value As System.IntPtr) As System.IntPtr' and call Marshal.ThrowExceptionForHR.	Microsoft.VisualStudio.AppDesigner	2	E:\project-system2\src\Microsoft.VisualStudio.AppDesigner\PropPageDesigner	E:\project-system2\src\Microsoft.VisualStudio.AppDesigner\PropPageDesigner\PropPageDesignerView.vb	701	17	Reliability	IntelliSense	Active	Microsoft.NetCore.VisualBasic.Analyzers
[...]

Issue #746, opened 2015-12-08
Fix DoNotIgnoreMethodResultsAnalyzer (CA1806 port) to handle PreserveSig methods


mavasani:
`DoNotIgnoreMethodResultsAnalyzer.IsHResultOrErrorCodeReturningMethod` does not completely match the FXCop implementation [here](http://index/UsageRules/R/88335f66dce7e0ad.html) - especially for ComImport types with methods marked with PreserveSig.

We need to tune this method to match the FxCop behavior once https://github.com/dotnet/roslyn/issues/7282 is addressed. I will also add a skip test `UnusedComImportPreserveSig` for this issue.

mavasani:
Also extend it to handle Pure attribute - @nguerrera can you please add more context on what needs to be done?

nguerrera:
Any method marked [Pure] indicates that it has no side effect. As such, it is always wrong to call it and not use the return value.

nguerrera:
https://msdn.microsoft.com/en-us/library/system.diagnostics.contracts.pureattribute(v=vs.110).aspx

mavasani:
This is blocked on dotnet/roslyn#7282, moving to Unknown milestone.

Youssef1313:
@mavasani The linked roslyn issue is now fixed. The issue can be addressed once you move to a version containing the new API.

PR #3581, opened 2020-04-29
Remove CA2010 (ConsumePreserveSigAnalyzer)

mavasani:
1. This analyzer has no unit tests
2. Its functionality is a subset of functionality provided for [CA1806: Do not ignore method results](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1806)
3. The functionality provided in this analyzer is not completely correct: See #1864. #746 tracks implementing this functionality correctly in CA1806 analyzer
4. This analyzer has no documentation: https://docs.microsoft.com/visualstudio/code-quality/ca2010 gives 404


src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.sarif
```diff
@@ -2457,6 +2457,25 @@
             ]
           }
         },
+        "CA1835": {
+          "id": "CA1835",
+          "shortDescription": "Prefer 'ReadOnlyMemory<Byte>' overload for 'WriteAsync'.",
```
mavasani, line 2462: @carlossanlop I think you may want to update the Title and Description for CA1835 rules to be shared, i.e. something like `Prefer 'ReadOnlyMemory<Byte>' overload for 'ReadAsync'/'WriteAsync'` and use it for both your descriptors.

mavasani:
Ping - trivial review deleting code.

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md
```diff
@@ -103,12 +103,12 @@ Rule ID | Title | Category | Enabled | Severity | CodeFix | Description |
...
 [CA1834](https://docs.microsoft.com/visualstudio/code-quality/ca1834) | Consider using 'StringBuilder.Append(char)' when applicable. | Performance | True | Warning | True | 'StringBuilder.Append(char)' is more efficient than 'StringBuilder.Append(string)' when the string is a single character. When calling 'Append' with a constant, prefer using a constant char rather than a constant string containing one character. |
+[CA1835](https://docs.microsoft.com/visualstudio/code-quality/ca1835) | Prefer 'ReadOnlyMemory<Byte>' overload for 'WriteAsync'. | Performance | True | Warning | False | All 'Stream' based classes provide a 'WriteAsync' overload that takes a 'ReadOnlyMemory<Byte>' as first argument. When possible, prefer the overload that takes a 'ReadOnlyMemory<Byte>', which is more efficient. |
```
stephentoub, line 106: Why is this being added in this PR?  Was it just missing erroneously?

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
```diff
@@ -17,7 +17,7 @@ Security: CA2100-CA2153, CA2300-CA2330, CA3000-CA3147, CA5300-CA5403
 Usage: CA1801, CA1806, CA1816, CA2200-CA2209, CA2211-CA2247
 Naming: CA1700-CA1726
 Maintainability: CA1500-CA1509
-Reliability: CA9999, CA2000-CA2015
+Reliability: CA9999, CA2000-CA2009, CA2011-CA2015
```
stephentoub, line 20: Are we ok reusing CA2010 for something else in the future?

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md
mavasani: This was me just running `/t:pack` locally, which runs the tool to auto-update the SARIF and md files to reflect the current state. I can revert these changes and have them done in a separate cleanup PR later on.

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
mavasani: Historically, we have avoided repurposing IDs given users might have existing suppressions/ruleset entries to turn these off. I am in two minds if we should make this change to indicate that we no longer use CA2010 OR keep the original range so we don't accidentally pick up CA2010. Maybe safer to go with the latter, let me revert this change.

mavasani:
Thanks - that made the PR much cleaner and only has deletions now.

PR #3584, opened 2020-04-30
Prefer IsEmpty over Count

Jozkee:
Fixes https://github.com/dotnet/runtime/issues/33801

This PR adds a rule to identify access to members called "Count" of type int/uint/long/ulong on a valid logical binary operation when compared to 0 or 1 and whose type also contains a visible boolean "IsEmpty" property.  

| Operation    | Replacement |
|--------------|-------------|
| Count == 0   | IsEmpty     |
| Count \!= 0  | \!IsEmpty   |
| Count < 0    | NA          |
| Count <= 0   | IsEmpty     |
| Count > 0    | \!IsEmpty   |
| Count >= 0   | NA          |
|              |             |
| Count == 1   | NA          |
| Count \!= 1  | NA          |
| Count < 1    | IsEmpty     |
| Count <= 1   | NA          |
| Count > 1    | NA          |
| Count >= 1   | \!IsEmpty   |
|              |             |
| 0 == Count   | IsEmpty     |
| 0 \!= Count  | \!IsEmpty   |
| 0 < Count    | \!IsEmpty   |
| 0 <= Count   | NA          |
| 0 > Count    | NA          |
| 0 >= Count   | IsEmpty     |
|              |             |
| 1 == Count   | NA          |
| 1 \!= Count  | NA          |
| 1 < Count    | NA          |
| 1 <= Count   | \!IsEmpty   |
| 1 > Count    | IsEmpty     |
| 1 >= Count   | NA          |


mavasani:
Few things to note:

1. CA1827-CA1829 were contributed very recently by a community member and we should feel free to tune its heuristic and/or make them more general, as appropriate.

2. The fact that CA1829 is implemented as a separate analyzer already causes bad user experience in certain cases where we flag both CA1827 and CA1829 on same code, leading to different end result for the user based on which fix they chose, instead of the most efficient one. I was planning to merge these analyzers to avoid multiple suggestions on same code.

3. New analysis added in this PR adds to these scenarios and makes for a bad user experience. For example, consider user code “.Count() != 0” when IsEmpty and Count property are both available. The most efficient check here is “!IsEmpty”, but we end up suggesting CA1827 and CA1829, and user either ends up with “.Any()” or “.Count != 0”. For the latter case, we end up with a follow-up violation from the analysis added here that would suggest “!IsEmpty”. It is unfortunate that user ended up with different end result in both cases, and the code generated by CA1829 code fix itself is always flagged by this rule.

4. I suggest we merge all these analyses into a single analyzer for a seamless user experience:
   1. At most one of these CA rules should fire for any given code, and that should be recommending the most efficient end result for the user
   2. Applying a code fix from any of the rules in this bucket should not lead to a follow-up violation from another rule in this bucket
IMO, this is only possible if we merge CA1827-CA1829 and the analysis added in this PR into a single analyzer.

jeffhandley:
Thanks @mavasani, I appreciate the feedback and recommendations.  We had previously looked at the "Use Any when applicable over Count" analyzer, [but our assessment at that time](https://github.com/dotnet/runtime/issues/33801#issuecomment-607510770) was that it was targeting different code and using different heuristics, but from your analysis here, I see how there is overlap in the experience.

Generalizing this group of analyzers into a single analyzer is intriguing. I think we'd need to do some more design work to figure out the right path forward on that though.

- We'd need to expand @Jozkee's mapping shown to capture the combined set of warnings and fixes
- The implementation would likely not be able to target specific types or interfaces, but instead target member names.
- We were already unsure if this could truly be considered a "Performance" category analyzer--I suspect that a general-purpose analyzer tackling all of these scenarios should be categorized as "Maintainability" since it's really about code readability/clarity.
- We'd have to figure out which (if any) of the diagnostic IDs to keep and what the downstream effects of that are

Jozkee:
I was scratching my head trying to see how we could merge the 4 rules all together in order to diagnose one rule at most as you mention and this is what I got.

First, I identified the rules and the types affected by them; the next table tries to illustrate that.
For example, CA1829 is only registered when the compilation contains `System.Linq.Enumerable` which extends `IEnumerable<T>`, hence, Applies To shows the type `IEnumerable<T>` and Applies When shows `System.Linq.Enumerable`, the same for the other 3 rules.

| Rule ID            | Applies to                        | Applies when                                                                                                                                                             | Description                                                       |
|--------------------|-----------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------|
| CA1827 <br> CA1828 | IEnumerable<T> <br> IQueryable<T> | Compilation contains: <br> System\.Linq\.Enumerable <br> EntityFrameworkQueryableExtensions <br> System\.Data\.Entity\.QueryableExtensions <br> System\.Linq\.Enumerable | Do not use Count\[Async\]\(\) when Any\[Async\]\(\) can be used\. |
| CA1829             | IEnumerable<T>                    | Compilation contains: <br> System\.Linq\.Enumerable                                                                                                                      | Use property Length/Count instead of Count\(\) when available\.          |
| CA1836             | All types                         | Always                                                                                                                                                                   | Prefer IsEmpty over Count property (and Count()?).                              |

Second, I tried to sketch a routine that could allow us to funnel into returning one rule following this principles:
1. What's the offending expression?
2. What's the most optimal expression that you can use to express the same?

And this is what I got. The comments below each condition says what are the possible rules that can be returned from that code branch.
Please disregard the fact that this is a mix of pseudo-code and C#
```
// Offending expression e.g: obj.Count() != 0
offendingExpression = context.Operation; 

if context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable)
// CA1827, 28, 29 & 36.
  if offendingExpression is (valid binary operation)* or (valid Equals invocation operation)**
    // CA1827, 28, 36. 
    IOperation operation = Get the non-constant side of the expression
    
    if operation is Count() method invocation
      // CA1827, 36
      if operation.ContainingType contains (valid IsEmpty property)***
        return CA1836
      else 
        return CA1827

    else if operation is CountAsync() method invocation
      return CA1828
    else if operation is Count property member accessor
      return CA1836

  else if offendingExpression is Count() method invocation
    return CA1829

if context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable)
// CACA1827, 28 & 36.
  if offendingExpression is (valid binary operation)* or (valid Equals invocation operation)**
    // CA1827, 28, 36. 
    IOperation operation = Determine the non-constant side of the expression.
    
    if operation is Count() method invocation
      // CA1827, 36
      if operation.ContainingType contains (valid IsEmpty property)***
        return CA1836
      else 
        return CA1827

    else if operation is CountAsync() method invocation
      return CA1828
    else if operation is Count property member accessor
      if operation.ContainingType contains (valid IsEmpty property)***
        return CA1836

// For all types
if offendingExpression is (valid binary operation)* or (valid Equals invocation operation)**
  IOperation operation = Determine the non-constant side of the expression.
  if operation is Count() method invocation OR operation is Count property member accessor
    if operation.ContainingType contains (valid IsEmpty property)***
      return CA1836
```

@mavasani @jeffhandley Would you consider this good enough? 
Would it be better to do it in this PR or in a separate PR?

@mavasani Should we re-arrange the sequence of Rule IDs so the 4 rules (CA1827-CA1829 & CA1836) show as consecutive numbers since they are somewhat related?

> Applying a code fix from any of the rules in this bucket should not lead to a follow-up violation from another rule in this bucket

I have the concern that this might keep happening even if we merge the rules.
Consider the following:
```cs
List<string> list = new List<string>();
bool isEmpty = list.Count() == 0;
```
As of today, this snippet triggers both rules CA1827 and CA1829 but I think that is because there are two offending expressions which are two separate analysis, one for `list.Count() == 0` and the other for `list.Count()`. 

Is there a way to query the current diagnosis results so we can either avoid emitting rule CA1829 if the expression is within the expression that triggered CA1827 OR remove diagnosis of CA1829 if CA1827 encompasses it?

mavasani:
Thanks @Jozkee - that seems to match pretty much what I had envisioned. You can ensure that we do not report duplicate diagnostics by writing separate analysis callbacks for method invocation and property reference as follows.

NOTE: I think CA1829 should be relaxed so semantics are equivalent to CA1836, i.e. flags Count() invocation if there is any valid property named Length/Count on containing type.

```
void Initialize(AnalysisContext context)
{
   context.RegisterCompilationStartAction(OnCompilationStart);
}

void OnCompilationStart(CompilationStartAnalysisContext context)
{
  var systemLinqEnumerable = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable)
  var enumerableCountMethod = systemLinqEnumerable?.GetMembers("Count").OfType<IMethodSymbol>().FirstOrDefault(m => m.Parameters.Length == 0)

  var queryableExtensionsType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDataEntityQueryableExtensions)
  var queryableCountAsyncMethod = queryableExtensionsType?.GetMembers("CountAsync").OfType<IMethodSymbol>().FirstOrDefault(m => m.Parameters.Length == 0)

  if (enumerableCountMethod != null or queryableCountAsyncMethod != null)
    compilationStartContext.RegisterOperationAction(
      c => AnalyzeMethodInvocation(c, enumerableCountMethod, queryableCountAsyncMethod),
      OperationKind.Invocation)  
  compilationStartContext.RegisterOperationAction(AnalyzePropertyReference, OperationKind.PropertyReference)
}

void AnalyzeMethodInvocation(OperationAnalysisContext context, IMethodSymbol enumerableCountMethod, IMethodSymbol queryableCountAsyncMethod)
{
  var invocation = (IInvocationOperation)context.Operation;
  var isCountMethodInvocation = invocation.TargetMethod.Equals(enumerableCountMethod)
  var isCountAsyncMethodInvocation = invocation.TargetMethod.Equals(queryableCountAsyncMethod)
  if (!isCountMethodInvocation and !isCountAsyncMethodInvocation)
    return

  // Define extension similar to WalkDownConversion
  var parentIgnoringConversions = invocation.WalkUpConversions();

  // Check if invocation is parented by an Offending expression e.g: obj.Count() != 0
  if parentIgnoringConversions is (valid binary operation)* or (valid Equals invocation operation)**
    if isCountMethodInvocation
      if invocation.ContainingType contains (valid IsEmpty property)***
        report CA1836
      else if invocation.ContainingType contains (valid Length/Count property)***
        report CA1829
      else
        report CA1827

    else if isCountAsyncMethodInvocation
      report CA1828

  else if invocation.ContainingType contains (valid Length/Count property)***
    report CA1829
}

void AnalyzePropertyReference(OperationAnalysisContext context)
{
  var propertyReference = (IPropertyReferenceOperation)context.Operation;
  if propertyReference.Name != "Count" and propertyReference.Name != "Length"
       return
  if Not propertyReference.ContainingType contains (valid IsEmpty property)***
       return

  var parentIgnoringConversions = propertyReference.WalkUpConversions();
  if parentIgnoringConversions is (valid binary operation)* or (valid Equals invocation operation)**
        report CA1836
}
```

mavasani:
> Should we re-arrange the sequence of Rule IDs so the 4 rules (CA1827-CA1829 & CA1836) show as consecutive numbers since they are somewhat related?

That is tricky, as we have already released an FxCop analyzers package with CA1827-CA1829. Users might have added source suppressions or ruleset/editorconfig entries to turn off the diagnostic, and we don't want them to be broken by us changing IDs of shipped rules. We have faced similar situations in the past, and come to terms with the fact that having non-sequential IDs is better then changing existing rule IDs from an end user perspective as it avoids breaking change for users.

mavasani:
@Jozkee Let me know if you have any other questions, we can discuss offline as well. Feel free to ping me anytime on teams.

Jozkee:
@mavasani thanks for replying! Few more questions:

> NOTE: I think CA1829 should be relaxed so semantics are equivalent to CA1836, i.e. flags Count() invocation if there is any valid property named Length/Count on containing type.

If we relax it, then the rule wouldn't be proper to `Microsoft.NetCore.Analyzers`, don't you think? 

As per the [documentation](https://github.com/dotnet/roslyn-analyzers/blob/master/GuidelinesForNewRules.md) (Section 4.ii):
> Analyzers related to pure code quality improvements, which are not specific to any API should go into Microsoft.CodeQuality.Analyzers. Analyzers specific to usage of a specific .NetCore/.NetStandard API should go into Microsoft.NetCore.Analyzers package.

However, since the rule already shipped, it wouldn't be wise to just move it to `Microsoft.CodeQuality.Analyzers`. Should we include rule CA1829 in both projects? 

Also, since we are planning on merging API specific rules (CA1827 and CA1828) with non-API specific (CA1829 if relaxed, and CA1836), in which project should those rules exist?

mavasani:
> If we relax it, then the rule wouldn't be proper to Microsoft.NetCore.Analyzers, don't you think

I would not worrying about the sub-package/folder that the rule belongs to - feel free to move the analyzer(s) around so all analysis for these related rules is in a single analyzer. End users will either install FxCop analyzers package, which includes both `Microsoft.NetCore.Analyzers` and `Microsoft.CodeQuality.Analyzers` OR get all the CA analyzers in .NET5 SDK. So, we should be fine.

mavasani:
> Also, since we are planning on merging API specific rules (CA1827 and CA1828) with non-API specific (CA1829 if relaxed, and CA1836), in which project should those rules exist?

I think it makes sense to keep them in `Microsoft.NetCore.Analyzers` as these are all API centric analyzers. We normally put API-agnostic analyzers that recommend pure code quality improvements in `Microsoft.CodeQuality.Analyzers`, say remove unnecessary parameter, make method static, etc. etc.

Jozkee:
I am almost finishing up with merging the analyses, and I found a couple of bugs in the current analyzer implementation for rule CA1827/28.

* `bool any = 0 >= list.Count();`
fixer suggests `bool any = list.Any();`
The condition will only be true when `Count = 0`, so `Any()` should be negated. 

* `bool any = 0 < list.Count();`
fixer suggests `bool any = !list.Any();`
This case is the flipped version of `Count > 0`; `Any()` should not be negated.

I will fix those two cases in the upcoming commits, I just wanted to point them out in this thread to keep track in case is needed, and also ask if there is a tracking issue for this that we may close after merging this, I doubt it since those look like very uncommon cases.

mavasani:
@Jozkee Thanks - awesome!

mavasani:
> also ask if there is a tracking issue for this that we may close after merging this, I doubt it since those look like very uncommon cases.

I don't think these are known issues.

Jozkee:
Recap:
* Removed the following classes  (the fixers still exist, though):
  * `DoNotUseCountWhenAnyCanBeUsedAnalyzer`
  * `UsePropertyInsteadOfCountMethodWhenAvailableAnalyzer`

* Created class `UseCountProperlyAnalyzer` (feel free to suggest a better name) which supports rules 1827, 28, 29 & 36 (added in this PR as well).

All tests are passing with the exception of a few ones related to relaxation of rule 1829, (will call those out individually as a review comment).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1323,4 +1323,13 @@
   <data name="PreferConstCharOverConstUnitStringInStringBuilderTitle" xml:space="preserve">
     <value>Consider using 'StringBuilder.Append(char)' when applicable.</value>
   </data>
-</root>
+  <data name="PreferIsEmptyOverCountDescription" xml:space="preserve">
+    <value>Prefer IsEmpty over Count when checking if your instance is empty.</value>
```
mavasani, line 1327:
```suggestion
    <value>Prefer 'IsEmpty' over 'Count' when checking if your instance is empty.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1323,4 +1323,13 @@
...
+  </data>
+  <data name="PreferIsEmptyOverCountMessage" xml:space="preserve">
+    <value>Prefer IsEmpty over Count when checking if your instance is empty.</value>
```
mavasani, line 1330:
```suggestion
    <value>Prefer 'IsEmpty' over 'Count' when checking if your instance is empty.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: I think we need to add more to the description text explaining the underlying reason behind the suggestion, i.e. what is the performance benefit from making the suggested change.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -577,6 +577,24 @@ public static SyntaxNode GetInstance(this IInvocationOperation invocationOperati
             return invocationOperation.IsExtensionMethodAndHasNoInstance() ? invocationOperation.Arguments[0].Value.Syntax : invocationOperation.Instance.Syntax;
         }
 
+        public static ITypeSymbol GetInstanceType(this IOperation operation)
+        {
+            IOperation instance = operation switch
+            {
+                IInvocationOperation invocation => invocation.IsExtensionMethodAndHasNoInstance() ?
+                invocation.Arguments[0].Value :
+                invocation.Instance,
```
mavasani, line 586: Can we extract this into a helper, say `public static IOperation GetInstance(this IInvocationOperation invocationOperation)`, rename the existing `GetInstance` extension that returns a `SyntaxNode` to `GetInstanceSyntax`: 
```csharp
public static SyntaxNode GetInstanceSyntax(this IInvocationOperation invocationOperation) => invocationOperation.GetInstance().Syntax;

public static IOperation GetInstance(this IInvocationOperation invocationOperation) => invocationOperation.IsExtensionMethodAndHasNoInstance() ? invocationOperation.Arguments[0].Value : invocationOperation.Instance;

```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
Jozkee: Here's a suggestion taken from the [`ConcurrentStack<T>`]() documentation:

_"For determining whether the type contains any items, prefer using 'IsEmpty' property rather than retrieving the number of items from the 'Count' property and comparing it to 0."_

cc @terrajobst

mavasani: I was expecting something such as “int comparison is more expensive then bool comparison”, in better wording though.

Jozkee: As mentioned in https://github.com/dotnet/runtime/issues/33801#issuecomment-617841271.

In some cases, like in concurrent collections, it is faster to use `IsEmpty` since the getter will stop at the first item, similar to the `Count()` vs `Any()` rule. 
In many other cases it just reads better. 

With that said, I don't think that we should say that comparing `bool` vs `int` is always faster, I don't know if it is significant enough to be mentioned, 

There's also cases where IsEmpty looks into Count:
```
public sealed partial class ImmutableHashSet<T>{
    public bool IsEmpty
    {
        get { return this.Count == 0; }
    }
}
```

I think that the purpose of the rule was to use `IsEmpty` when there was a clear intent of "check if the object contains or does not contain items".

Was your suggestion aiming to be brief? If that's the case, I don't think it applies here, since there is similar wording on `DoNotUseCountWhenAnyCanBeUsedDescription`
https://github.com/dotnet/roslyn-analyzers/blob/4d69cb3c3abd19f1d5f1308bcd4febf1432400d6/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx#L222-L224

mavasani: Thanks for the clarification! I think what you have is fine.

Jozkee:
@mavasani Can you grant me permission to add more reviewers or add below names yourself? 
cc @pgovind @buyaa-n @carlossanlop @bartonjs

mavasani:
@Jozkee you should have correct permissions now.

I took a quick skim over the analyzer code and it looks in pretty good shape to me. I will review the PR thoroughly later today or early tomorrow. Thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+
+            if (syncMethods.Count > 0 || asyncMethods.Count > 0)
+            {
+                context.RegisterOperationAction(operationContext => AnalyzeInvocationOperation(operationContext, syncMethods, asyncMethods), OperationKind.Invocation);
+            }
+
+            context.RegisterOperationAction(AnalyzePropertyReference, OperationKind.PropertyReference);
+        }
+
+        private static void AnalyzeInvocationOperation(OperationAnalysisContext context, HashSet<IMethodSymbol> syncMethods, HashSet<IMethodSymbol> asyncMethods)
```
mavasani, line 164: >HashSet<IMethodSymbol> syncMethods, HashSet<IMethodSymbol> asyncMethods [](start = 89, length = 71)

Can we use ImmutableHashSets here?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+        {
+            var invocationOperation = (IInvocationOperation)context.Operation;
+
+            // For C#, TargetMethod returns the actual extension method that we are looking for.
+            // For VB, TargetMethod shows as a member of the extended class, in order to get the static extension method, we need to call ReducedFrom.
+
+            // We use OriginalDefinition to normalize the method to its generic version. For VB we don't use it since I assume ReducedFrom already goes directly to the generic version.
+            IMethodSymbol originalDefinition = invocationOperation.Language == LanguageNames.VisualBasic ?
+                invocationOperation.TargetMethod.ReducedFrom :
+                invocationOperation.TargetMethod.OriginalDefinition;
```
mavasani, line 174: What if we use `IMethodSymbol originalDefinition = (invocationOperation.TargetMethod.ReducedFrom ?? invocationOperation.TargetMethod).OriginalDefinition` to avoid language specific checks?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            parentOperation = parentOperation.WalkUpParentheses();
+            parentOperation = parentOperation.WalkUpConversion();
+
+            // Analyze binary operation.
+            if (parentOperation is IBinaryOperation parentBinaryOperation)
+            {
+                AnalyzeParentBinaryOperation(context, invocationOperation, parentBinaryOperation,
+                    isAsync, originalDefinition.Name);
+            }
+            // Analize invocation operation, potentially obj.Count().Equals(0).
```
mavasani, line 210:
```suggestion
            // Analyze invocation operation, potentially obj.Count().Equals(0).
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            else if (parentOperation is IInvocationOperation parentInvocationOperation)
+            {
+                AnalyzeParentInvocationOperation(context, invocationOperation, parentInvocationOperation,
+                    isInstance: true, isAsync, originalDefinition.Name);
+            }
+            // Analize argument operation, potentially 0.Equals(obj.Count()).
```
mavasani, line 216:
```suggestion
            // Analyze argument operation, potentially 0.Equals(obj.Count()).
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            {
+                AnalyzeParentInvocationOperation(context, propertyReferenceOperation, parentInvocationOperation,
+                    isInstance: true, isAsync: false, methodName: null);
+            }
+            // Analize argument operation, potentially 0.Equals(obj.Count).
+            else if (parentOperation is IArgumentOperation argumentOperation)
+            {
+                AnalyzeParentInvocationOperation(context, propertyReferenceOperation, (IInvocationOperation)argumentOperation.Parent,
+                    isInstance: false, isAsync: false, methodName: null);
+            }
```
mavasani, line 254: Is it possible to share this code between this method and prior method?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            if (isAsync)
+            {
+                bool shouldNegateAny = !shouldNegateIsEmpty;
+                ReportCA1828(context, shouldNegateAny, useRightSideExpression ? OperationBinaryRight : OperationBinaryLeft, methodName!, parent);
+            }
+            else
+            {
+                ITypeSymbol type = operation.GetInstanceType();
+
+                if (TypeContainsVsibileProperty(context, type, IsEmpty, SpecialType.System_Boolean))
```
mavasani, line 280:
```suggestion
                if (type != null && TypeContainsVisibleProperty(context, type, IsEmpty, SpecialType.System_Boolean))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+                {
+                    ReportCA1836(context, useRightSideExpression, shouldNegateIsEmpty, parent);
+                }
+                else
```
mavasani, line 284: Why are we not checking `if (TypeContainsVisibleProperty(context, type, Count, SpecialType.System_Integer))` and `if (TypeContainsVisibleProperty(context, type, Length, SpecialType.System_Integer))` when in context of `Count()` callback and recommending use of `Count` or `Length` versus directly moving to recommending `Any()`? Isn't the former more efficient?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+                {
+                    bool shouldNegateAny = !shouldNegateIsEmpty;
+                    ReportCA1827(context, shouldNegateAny, useRightSideExpression ? OperationBinaryRight : OperationBinaryLeft, methodName!, parent);
```
mavasani, line 287: By sharing this method `AnalyzeParentBinaryOperation` between invocation and property reference callbacks, we now end up recommending `Any()`, when user has code that uses property `Count` or `Length`. This will certainly feel counter-intuitive to users who explicitly prefer to avoid LINQ and are using a much more efficient property check.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            {
+                ITypeSymbol type = operation.GetInstanceType();
+
+                if (TypeContainsVsibileProperty(context, type, IsEmpty, SpecialType.System_Boolean))
+                {
+                    ReportCA1836(context, !isInstance, shouldNegate: false, parent);
+                }
+                else
+                {
+                    ReportCA1827(context, shouldNegate: true, isInstance ? OperationEqualsInstance : OperationEqualsArgument, methodName!, parent);
```
mavasani, line 326: Same here - I don't think we want to ever report `CA1827` and recommend using LINQ when user's original code has a more efficient property access.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
mavasani, line 284: I think we should change this to:
```suggestion
                else if (methodName != null)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            }
+            else
+            {
+                ITypeSymbol type = operation.GetInstanceType();
+
+                if (TypeContainsVsibileProperty(context, type, IsEmpty, SpecialType.System_Boolean))
+                {
+                    ReportCA1836(context, !isInstance, shouldNegate: false, parent);
+                }
+                else
```
mavasani, line 324: Avoid reporting `CA1827` if user already has an efficient property access.
```suggestion
            else if (methodName != null)
```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -572,9 +572,26 @@ public static bool IsExtensionMethodAndHasNoInstance(this IInvocationOperation i
             return invocationOperation.TargetMethod.IsExtensionMethod && (invocationOperation.Language != LanguageNames.VisualBasic || invocationOperation.Instance == null);
         }
 
-        public static SyntaxNode GetInstance(this IInvocationOperation invocationOperation)
+        public static IOperation GetInstance(this IInvocationOperation invocationOperation)
+            => invocationOperation.IsExtensionMethodAndHasNoInstance() ? invocationOperation.Arguments[0].Value : invocationOperation.Instance;
+
+        public static SyntaxNode GetInstanceSyntax(this IInvocationOperation invocationOperation)
+            => invocationOperation.GetInstance().Syntax;
```
mavasani, line 579:
```suggestion
        public static SyntaxNode? GetInstanceSyntax(this IInvocationOperation invocationOperation)
            => invocationOperation.GetInstance()?.Syntax;
```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -572,9 +572,26 @@ public static bool IsExtensionMethodAndHasNoInstance(this IInvocationOperation i
             return invocationOperation.TargetMethod.IsExtensionMethod && (invocationOperation.Language != LanguageNames.VisualBasic || invocationOperation.Instance == null);
         }
 
-        public static SyntaxNode GetInstance(this IInvocationOperation invocationOperation)
+        public static IOperation GetInstance(this IInvocationOperation invocationOperation)
+            => invocationOperation.IsExtensionMethodAndHasNoInstance() ? invocationOperation.Arguments[0].Value : invocationOperation.Instance;
```
mavasani, line 576: `invocation.Instance` can be null
```suggestion
        public static IOperation? GetInstance(this IInvocationOperation invocationOperation)
            => invocationOperation.IsExtensionMethodAndHasNoInstance() ? invocationOperation.Arguments[0].Value : invocationOperation.Instance;
```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -572,9 +572,26 @@ public static bool IsExtensionMethodAndHasNoInstance(this IInvocationOperation i
...
+
+        public static SyntaxNode GetInstanceSyntax(this IInvocationOperation invocationOperation)
+            => invocationOperation.GetInstance().Syntax;
+
+        public static ITypeSymbol GetInstanceType(this IOperation operation)
```
mavasani, line 581: Instance can be null for static member references.
```suggestion
        public static ITypeSymbol? GetInstanceType(this IOperation operation)
```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -572,9 +572,26 @@ public static bool IsExtensionMethodAndHasNoInstance(this IInvocationOperation i
...
+                IInvocationOperation invocation => invocation.GetInstance(),
+
+                IPropertyReferenceOperation propertyReference => propertyReference.Instance,
+
+                _ => throw new NotImplementedException()
+            };
+
+            instance = instance.WalkDownConversion();
+
+            return instance.Type;
```
mavasani, line 594:
```suggestion
            return instance?.WalkDownConversion().Type;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            }
+
+            if (isAsync)
+            {
+                bool shouldNegateAny = !shouldNegateIsEmpty;
+                ReportCA1828(context, shouldNegateAny, useRightSideExpression ? OperationBinaryRight : OperationBinaryLeft, methodName!, parent);
+            }
+            else
+            {
+                ITypeSymbol type = operation.GetInstanceType();
```
mavasani, line 278:
```suggestion
                ITypeSymbol? type = operation.GetInstanceType();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+                return;
+            }
+
+            if (isAsync)
+            {
+                ReportCA1828(context, shouldNegate: true, isInstance ? OperationEqualsInstance : OperationEqualsArgument, methodName!, parent);
+            }
+            else
+            {
+                ITypeSymbol type = operation.GetInstanceType();
```
mavasani, line 318:
```suggestion
                ITypeSymbol? type = operation.GetInstanceType();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+
+                if (TypeContainsVsibileProperty(context, type, IsEmpty, SpecialType.System_Boolean))
```
mavasani, line 320:
```suggestion
                if (type != null && TypeContainsVisibleProperty(context, type, IsEmpty, SpecialType.System_Boolean))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+                else
+                {
+                    ReportCA1827(context, shouldNegate: true, isInstance ? OperationEqualsInstance : OperationEqualsArgument, methodName!, parent);
+                }
+            }
+        }
+
+        private static void AnalyzeCountInvocationOperation(OperationAnalysisContext context, IInvocationOperation invocationOperation)
+        {
+            ITypeSymbol type = invocationOperation.GetInstanceType();
```
mavasani, line 333:
```suggestion
            ITypeSymbol? type = invocationOperation.GetInstanceType();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+                return false;
+            }
+
+            IMethodSymbol methodSymbol = invocationOperation.TargetMethod;
+
+            return string.Equals(methodSymbol.Name, WellKnownMemberNames.ObjectEquals, StringComparison.Ordinal)
+                && IsInRangeInclusive((uint)methodSymbol.ContainingType.SpecialType, (uint)SpecialType.System_Int32, (uint)SpecialType.System_UInt64);
+        }
+
+        private static void AddIfNotNull(HashSet<IMethodSymbol> set, IEnumerable<IMethodSymbol>? others)
```
mavasani, line 433: Nit: make this a static local function for the single method where this is called?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+
+        private static bool TypeContainsVsibileProperty(OperationAnalysisContext context, ITypeSymbol type, string propertyName, SpecialType propertyType)
+            => TypeContainsVsibileProperty(context, type, propertyName, propertyType, propertyType);
+
+        private static bool TypeContainsVsibileProperty(OperationAnalysisContext context, ITypeSymbol type, string propertyName, SpecialType lowerBound, SpecialType upperBound)
+        {
+            if (type.GetMembers(propertyName).FirstOrDefault() is IPropertySymbol property)
+            {
+                return IsPropertyValidAndVisible(context, property, lowerBound, upperBound);
+            }
```
mavasani, line 545: Seems like the same code is used three times in this method. Can we extract to a static local function?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+            IsInRangeInclusive((uint)property.Type.SpecialType, (uint)lowerBound, (uint)upperBound) &&
+            property.GetMethod != null &&
+            context.Compilation.IsSymbolAccessibleWithin(property, context.ContainingSymbol.ContainingType) &&
+            context.Compilation.IsSymbolAccessibleWithin(property.GetMethod, context.ContainingSymbol.ContainingType);
+
+        private static bool TryGetZeroOrOneConstant(IOperation operation, out int constant)
+        {
+            constant = default;
+
+            if (operation?.Type?.SpecialType != SpecialType.System_Int32 &&
```
mavasani, line 586: Nit: Can we use a switch statement/expression here and simplify code?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,620 @@
...
+                    }
+                    break;
+                default:
+                    return false;
+            }
+
+            return true;
+        }
+
+        private static bool TypeContainsVsibileProperty(OperationAnalysisContext context, ITypeSymbol type, string propertyName, SpecialType propertyType)
```
mavasani, line 537: This seems like a very good candidate to become an extension method on `ITypeSymbol`:

```csharp
public static bool TypeContainsVisibleProperty(this ITypeSymbol type, string propertyName, SpecialType propertyType, INamedTypeSymbol contextSymbol)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
mavasani: Seems like this is moving existing code. Feel free to retain or change as appropriate.

Jozkee:
> When user code has Count(), why are we not preferring CA1829 (Use property Count/Length) over CA1827 (Use Any())?

I was thinking that the right thing to do was to replace the code with its most semantically correct version (`IsEmpty` or `Any()`), if it was a candidate (part of a valid binary or equals invocation operation); otherwise, just report replacing the invocation operation.

I do understand your point that a property check is better that an `Any()` method, although I think `Any()` reads better.

mavasani:
> I do understand your point that a property check is better that an Any() method, although I think Any() reads better.

I think given this is a performance rule, property access would be a better suggestion. I am also fine with having a user configurable option here to allow users to choose if they prefer performant code or readable code in this space. I will let you, @jeffhandley and the runtime team decide the preferred default semantics here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Jozkee: That's a good idea but now we will be mixing the purpose of the parameter, I can add a new parameter `bool allowLinq` to effectively avoid reporting Any() when the user was not even using Linq.

bartonjs:
> I do understand your point that a property check is better that an Any() method, although I think Any() reads better.

If there's an IsEmpty, it's almost certainly better.  We should add Any() => !IsEmpty and !Any() => IsEmpty into the list of things to correct.

(LINQ Any() is better than LINQ Count(), may be worse than implemented Count() and is almost definitely worse than property Count/Length/IsEmpty)

Jozkee:
> If there's an IsEmpty, it's almost certainly better. We should add Any() => !IsEmpty and !Any() => IsEmpty into the list of things to correct.

New rule for IsEmpty takes priority over the rule for Any() so that should be already addressed in this PR.

> (LINQ Any() is better than LINQ Count(), may be worse than implemented Count() and is almost definitely worse than property Count/Length/IsEmpty)

Better in the sense of performance? @bartonjs so you would suggest to place rule "Use Property Count/Length instead of LINQ Count()" above "Use LINQ Any() instead of LINQ Count()"?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Jozkee: What's the benefit of it?

bartonjs:
> Better in the sense of performance?

Yeah. Any() has to spin up an iterator, try to pump it, then throw it away. A property defined on the type is (assuming it follows guidelines) not going to do anywhere near that much work, or have GC impact.

> so you would suggest to place rule "Use Property Count/Length instead of LINQ Count()" above "Use LINQ Any() instead of LINQ Count()"?

Yep.  I think the best-to-worst order for "has stuff or not" is

1. IsEmpty property
1. Any() instance method
1. Count/Length instance property
1. Any() extension method
1. Count() instance method
1. Count() extension method

The Count() instance method could arguably float up one, but it's probably just trading one kind of complication with another vs LINQ Any() (otherwise it would have been a property).

Jozkee:
OK, I did the proper changes to prefer CA1829 over CA1827, except when the offending method contains a predicate e.g. `Count(_ => true)`, in that case, we can only suggest `Any(_ => true)`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,679 @@
...
+                    // If the invocation has a predicate we can only suggest CA1827; 
+                    // otherwise, we would loose the lambda if we suggest CA1829 or CA1836. 
+                    if (hasPredicate)
+                    {
+                        bool shouldNegateAny = !shouldNegateIsEmpty;
+                        ReportCA1827(context, shouldNegateAny, useRightSideExpression ? OperationBinaryRight : OperationBinaryLeft, methodName!, parent);
+                    }
+                    else
+                    {
+                        if (TypeContainsVisibileProperty(context, type, IsEmpty, SpecialType.System_Boolean))
```
mavasani, line 303: Typo
```suggestion
                        if (TypeContainsVisibleProperty(context, type, IsEmpty, SpecialType.System_Boolean))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -0,0 +1,679 @@
...
+                            ReportCA1827(context, shouldNegate: true, isInstance ? OperationEqualsInstance : OperationEqualsArgument, methodName!, parent);
+                        }
+                    }
+                }
+            }
+        }
+
+        private static void AnalyzeCountInvocationOperation(OperationAnalysisContext context, IInvocationOperation invocationOperation)
+        {
+            ITypeSymbol? type = invocationOperation.GetInstanceType();
```
mavasani, line 384: Should we just bail out here if type is null?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferIsEmptyOverCount.Fixer.cs
```diff
@@ -0,0 +1,68 @@
...
+    public abstract class PreferIsEmptyOverCountFixer : CodeFixProvider
+    {
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(UseCountProperlyAnalyzer.CA1836);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            SyntaxNode binaryExpression = root.FindNode(context.Span);
```
mavasani, line 26: Defensive check
```suggestion
            SyntaxNode binaryExpression = root.FindNode(context.Span);
            if (binaryExpression == null)
            {
                return;
            }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferIsEmptyOverCount.Fixer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            ImmutableDictionary<string, string> properties = context.Diagnostics[0].Properties;
```
mavasani, line 28:
```suggestion
            ImmutableDictionary<string, string> properties = context.Diagnostics[0].Properties;
            if (properties == null)
            {
                return;
            }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferIsEmptyOverCount.Fixer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            // Indicates weather the Count property is on the Right or Left side.
```
mavasani, line 30:
```suggestion
            // Indicates whether the Count property is on the Right or Left side.
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferIsEmptyOverCount.Fixer.cs
```diff
@@ -0,0 +1,68 @@
...
+                createChangedDocument: async cancellationToken =>
+                {
+                    DocumentEditor editor = await DocumentEditor.CreateAsync(context.Document, cancellationToken).ConfigureAwait(false);
+                    SyntaxGenerator generator = editor.Generator;
+                    // The Count property within the binary expression.
+                    SyntaxNode countAccessor = GetMemberAccessorFromBinary(binaryExpression, useRightSideExpression);
+                    // The object that the Count property belongs to OR null if countAccessor is not a MemberAccessExpressionSyntax.
+                    SyntaxNode? objectExpression = GetExpressionFromMemberAccessor(countAccessor);
+                    // The IsEmpty property meant to replace the binary expression.
+                    SyntaxNode isEmptyNode = objectExpression is null ?
```
mavasani, line 46: Nit: add newline before comments
```suggestion
                    SyntaxGenerator generator = editor.Generator;

                    // The Count property within the binary expression.
                    SyntaxNode countAccessor = GetMemberAccessorFromBinary(binaryExpression, useRightSideExpression);

                    // The object that the Count property belongs to OR null if countAccessor is not a MemberAccessExpressionSyntax.
                    SyntaxNode? objectExpression = GetExpressionFromMemberAccessor(countAccessor);

                    // The IsEmpty property meant to replace the binary expression.
                    SyntaxNode isEmptyNode = objectExpression is null ?
```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
Jozkee: I couldn't get a repro where this would return null, even with static members:
```cs
using System;
using System.Linq;
using System.Collections.Concurrent;

public class Test
{
    private static ConcurrentDictionary<string, string> _dictionary;

    public bool Foo => _dictionary.Count > 0;
}
```

mavasani: Probably not for your case, but given this is an extension method, some other analyzer might call it from the context of a static method invocation in future.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Jozkee: That's a good point, I don't think so, I think we could find some other way to get `type` when there is no instance, but as I said on https://github.com/dotnet/roslyn-analyzers/pull/3584#discussion_r427486672, I haven't been able to get null, even with static members.

Jozkee: OK, I was able to get null from GetInstanceType with the following snippet:
```cs
public class Test
{
    public static int Count {get;}
    public bool IsEmpty;
    public bool Foo => Count > 0;
}
```

So, we would be bailing just with static properties and static methods, static extension methods does still have an instance so those are fine.

Is it OK to not report on that case?

mavasani: I think it seems fine to bail out on static cases.

Jozkee:
@mavasani Thanks, just a few more changes to fix some bugs that I found addressing your feedback.
* `AnalyzePropertyReference` must only report CA1836, before it could incorrectly report CA1829 if you were using property Count and there was also property Length available.
* `AnalyzePropertyReference` should not report if the property name is not Count or Length, before this, it was incorrectly suggesting CA1836 for all properties.
* Refactored `AnalyzeParent*` methods to only indicate if the parent operation should be replaced; I also moved the logic to determine which rule we should report to separate methods, `DetermineReportForInvocationAnalysis` and `DetermineReportForPropertyReference` to correctly honor the statement I already made on the first bullet.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferIsEmptyOverCountTests.cs
```diff
@@ -0,0 +1,360 @@
...
+
+        [Theory]
+        [InlineData("(Count) > 0", "Not IsEmpty")]
+        [InlineData("Count > (0)", "Not IsEmpty")]
+        [InlineData("(Count) > (0)", "Not IsEmpty")]
+        [InlineData("(Me.Count) > 0", "Not IsEmpty")]
+        [InlineData("Me.Count > (0)", "Not IsEmpty")]
+        [InlineData("(Me.Count) > (0)", "Not IsEmpty")]
+        // TODO: Reduce suggested fix to avoid special casing here.
+        [InlineData("((Me).Count) > (0)", "Not (Me).IsEmpty")]
```
Jozkee, line 208: For VB only, I had to manually specify the "expected fixed code" since depending on the parenthesis and on having the `Me` keyword, the result may chnage.

Why is this not a problem for C#? If you see above test `CSharpTestFixOnParentheses`, I always get the most "reduced" version when firing the rule.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -209,174 +209,109 @@ static void AddIfNotNull(ImmutableHashSet<IMethodSymbol>.Builder set, IEnumerabl
...
+            DetermineReportForInvocationAnalysis(context, invocationOperation, parentOperation,
+                shouldReplaceParent, isAsync, useRightSide, shouldNegateIsEmpty, hasPredicate, originalDefinition.Name, operationKey);
         }
 
         private static void AnalyzePropertyReference(OperationAnalysisContext context)
         {
             var propertyReferenceOperation = (IPropertyReferenceOperation)context.Operation;
+
+            string propertyName = propertyReferenceOperation.Member.Name;
+            if (propertyName != Count && propertyName != Length)
```
mavasani, line 247: Nice!

Issue #3586, opened 2020-04-30
Update the Title and Description for CA1835 rules to be shared in both descriptors declared by this analyzer


mavasani:
@carlossanlop I think you may want to update the Title and Description for CA1835 rules to be shared, i.e. something like `Prefer 'ReadOnlyMemory<Byte>' overload for 'ReadAsync'/'WriteAsync'` and use it for both your descriptors.

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/pull/3581_

carlossanlop:
No, because the `ReadAsync` method takes a `Memory<byte>` (see [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=netcore-3.1#System_IO_Stream_ReadAsync_System_Memory_System_Byte__System_Threading_CancellationToken_)), while the `WriteAsync` method takes a `ReadOnlyMemory<byte>` (see [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.writeasync?view=netcore-3.1#System_IO_Stream_WriteAsync_System_ReadOnlyMemory_System_Byte__System_Threading_CancellationToken_)).

carlossanlop:
Does it cause problems to have two separate titles and descriptions? If yes, is there a way to have one with a `{0}` placeholder? That was one of the questions in my PR but IIRC the answer was that it's not possible.

mavasani:
So basically each rule ID has to be mapped to a single title and description string. Unlike, messages we don't have the luxury to dynamically create these strings or use multiple of them for the same ID.

mavasani:
Category, Title and Description are the core metadata for each rule ID. Messages are specific to each violation, which might have additional/differing context, hence allowed to have variations per reported diagnostic instance.

carlossanlop:
Ok I can make the title and description texts generic enough to cover both cases. I'll address this in my current fixer PR.

carlossanlop:
Fixed in #3592

PR #3592, opened 2020-05-01
Fixer: Prefer Memory overloads for Stream async Read/Write methods

carlossanlop:
Fixes dotnet/runtime#33790

### Summary
If the user invokes any of these methods from an object that is or inherits from Stream:

```cs
ReadAsync(byte[] buffer, int offset, int count)
ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
WriteAsync(byte[] buffer, int offset, int count)
WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
```

We should suggest to instead use:
```cs
ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken)
WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken)
```

Works for both C# and VB.
All unit tests are passing locally.

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -34,10 +34,7 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
                 return args.FirstOrDefault(argOperation =>
                 {
                     return argOperation.Syntax is ArgumentSyntax argNode &&
-                           argNode.NameColon != null &&
-                           argNode.NameColon.Name != null &&
-                           argNode.NameColon.Name.Identifier != null &&
-                           argNode.NameColon.Name.Identifier.ValueText == name;
+                           argNode.NameColon?.Name?.Identifier.ValueText == name;
                 });
             }
         }
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.NetCore.Analyzers.Runtime;
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class PreferStreamAsyncMemoryOverloadsCSharpFixer : PreferStreamAsyncMemoryOverloadsFixer
+    {
+        protected override IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed)
+        {
+            isNamed = false;
+
+            // The expected position is beyond the total arguments, so we don't expect to find the argument in the array
+            if (index >= args.Length)
+            {
+                return null;
+            }
+            // If the argument in the specified index does not have a name, then it is in its expected position
+            else if (args[index].Syntax is ArgumentSyntax argNode && argNode.NameColon == null)
+            {
+                return args[index];
+            }
+            // Otherwise, find it by name
+            else
+            {
+                isNamed = true;
+                return args.FirstOrDefault(argOperation =>
+                {
+                    return argOperation.Syntax is ArgumentSyntax argNode &&
+                           argNode.NameColon != null &&
+                           argNode.NameColon.Name != null &&
+                           argNode.NameColon.Name.Identifier != null &&
+                           argNode.NameColon.Name.Identifier.ValueText == name;
+                });
+            }
+        }
+    }
+}
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
@@ -1278,23 +1278,14 @@
   <data name="DoNotUseStackallocInLoopsMessage" xml:space="preserve">
     <value>Potential stack overflow. Move the stackalloc out of the loop.</value>
   </data>
-  <data name="PreferStreamReadAsyncMemoryOverloadsTitle" xml:space="preserve">
-    <value>Prefer 'Memory&lt;Byte&gt;' overload for 'ReadAsync'.</value>
+  <data name="PreferStreamAsyncMemoryOverloadsTitle" xml:space="preserve">
+    <value>Prefer the memory based overloads for 'ReadAsync' and 'WriteAsync'.</value>
   </data>
-  <data name="PreferStreamReadAsyncMemoryOverloadsDescription" xml:space="preserve">
-    <value>All 'Stream' based classes provide a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as first argument. When possible, prefer the overload that takes a 'Memory&lt;Byte&gt;', which is more efficient.</value>
+  <data name="PreferStreamAsyncMemoryOverloadsDescription" xml:space="preserve">
+    <value>All 'Stream' based classes provide a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as the first argument, and a 'WriteAsync' overload that takes a 'ReadOnlyMemory&lt;Byte&gt;' as the first argument. Whenever possible, prefer the memory based overloads, which are more efficient.</value>
   </data>
   <data name="PreferStreamAsyncMemoryOverloadsMessage" xml:space="preserve">
     <value>Change the '{0}' method call to the overload that receives a '{1}'.</value>
-  </data>
-  <data name="PreferStreamWriteAsyncMemoryOverloadsTitle" xml:space="preserve">
-    <value>Prefer 'ReadOnlyMemory&lt;Byte&gt;' overload for 'WriteAsync'.</value>
-  </data>
-  <data name="PreferStreamWriteAsyncMemoryOverloadsDescription" xml:space="preserve">
-    <value>All 'Stream' based classes provide a 'WriteAsync' overload that takes a 'ReadOnlyMemory&lt;Byte&gt;' as first argument. When possible, prefer the overload that takes a 'ReadOnlyMemory&lt;Byte&gt;', which is more efficient.</value>
-  </data>
-  <data name="PreferStreamWriteAsyncMemoryOverloadsMessage" xml:space="preserve">
-    <value>Change the 'WriteAsync' method call to the overload that receives a 'ReadOnlyMemory&lt;Byte&gt;'.</value>
   </data>
   <data name="InstantiateArgumentExceptionsCorrectlyChangeToTwoArgumentCodeFixTitle" xml:space="preserve">
     <value>Change to call the two argument constructor, pass null for the message.</value>
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Immutable;
 using System.Composition;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
@@ -61,6 +62,13 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 return;
             }
 
+            IArgumentOperation bufferArgumentOperation = invocation.Arguments.FirstOrDefault(a => a.Value.Type.Equals(SpecialType.System_Byte));
+            if (bufferArgumentOperation == null)
+            {
+                return;
+            }
+            IArgumentOperation offsetArgumentOperation = invocation.Arguments.FirstOrDefault(a => a.Value.Type.Equals(SpecialType.System_Int32));
+
             string title = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle;
 
             context.RegisterCodeFix(
@@ -79,9 +87,9 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             SyntaxNode instanceNode = invocation.Instance.Syntax;
 
             // Need the byte array object so we can invoke its AsMemory() method
+            
             SyntaxNode bufferInstanceNode = invocation.Arguments[0].Value.Syntax; // byte[] buffer
 
-            // These arguments are not modified, just moved inside AsMemory
             SyntaxNode offsetNode = invocation.Arguments[1].Syntax; // int offset
             SyntaxNode countNode = invocation.Arguments[2].Syntax;  // int count
 
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Immutable;
 using System.Threading;
 using System.Threading.Tasks;
+using Analyzer.Utilities;
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.Editing;
 using Microsoft.CodeAnalysis.Operations;
@@ -29,6 +30,9 @@ namespace Microsoft.NetCore.Analyzers.Runtime
     /// </summary>
     public abstract class PreferStreamAsyncMemoryOverloadsFixer : CodeFixProvider
     {
+        // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
+        protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
+
         public sealed override ImmutableArray<string> FixableDiagnosticIds =>
             ImmutableArray.Create(PreferStreamAsyncMemoryOverloads.RuleId);
 
@@ -80,24 +84,21 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             // No nullcheck for this, because there is an overload that may not contain it
             IArgumentOperation? cancellationTokenOperation = GetArgumentByPositionOrName(invocation.Arguments, 3, "cancellationToken", out bool isCancellationTokenNamed);
 
-            string title = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle;
+            string titleAndMethodName = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle + invocation.TargetMethod.Name;
 
-            Task<Document> fixInvocation = FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
+            Func<CancellationToken, Task<Document>> createChangedDocument = _ => FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
                                                          bufferOperation.Value.Syntax, isBufferNamed,
                                                          offsetOperation.Value.Syntax, isOffsetNamed,
                                                          countOperation.Value.Syntax, isCountNamed,
                                                          cancellationTokenOperation?.Value.Syntax, isCancellationTokenNamed);
 
             context.RegisterCodeFix(
-                CodeAction.Create(
-                    title: title,
-                    createChangedDocument: c => fixInvocation,
-                    equivalenceKey: title),
+                MyCodeAction.Create(
+                    title: titleAndMethodName,
+                    createChangedDocument,
+                    equivalenceKey: titleAndMethodName),
                 context.Diagnostics);
         }
-
-        // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
-        protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
 
         private static Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName,
             SyntaxNode bufferValueNode, bool isBufferNamed,
@@ -146,6 +147,15 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
             return Task.FromResult(doc.WithSyntaxRoot(newRoot));
+        }
+
+        // Needed for Telemetry (https://github.com/dotnet/roslyn-analyzers/issues/192) 
+        private class MyCodeAction : DocumentChangeAction
+        {
+            public MyCodeAction(string title, Func<CancellationToken, Task<Document>> createChangedDocument, string equivalenceKey)
+                : base(title, createChangedDocument, equivalenceKey)
+            {
+            }
         }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -84,7 +84,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             // No nullcheck for this, because there is an overload that may not contain it
             IArgumentOperation? cancellationTokenOperation = GetArgumentByPositionOrName(invocation.Arguments, 3, "cancellationToken", out bool isCancellationTokenNamed);
 
-            string titleAndMethodName = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle + invocation.TargetMethod.Name;
+            string title = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle;
 
             Func<CancellationToken, Task<Document>> createChangedDocument = _ => FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
                                                          bufferOperation.Value.Syntax, isBufferNamed,
@@ -92,12 +92,12 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                                                          countOperation.Value.Syntax, isCountNamed,
                                                          cancellationTokenOperation?.Value.Syntax, isCancellationTokenNamed);
 
-            context.RegisterCodeFix(
-                MyCodeAction.Create(
-                    title: titleAndMethodName,
+            var action = new MyCodeAction(
+                    title: title,
                     createChangedDocument,
-                    equivalenceKey: titleAndMethodName),
-                context.Diagnostics);
+                    equivalenceKey: title + invocation.TargetMethod.Name);
+
+            context.RegisterCodeFix(action, context.Diagnostics);
         }
 
         private static Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName,
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -86,18 +86,18 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             string title = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle;
 
-            Func<CancellationToken, Task<Document>> createChangedDocument = _ => FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
+            Task<Document> createChangedDocument(CancellationToken _) => FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
                                                          bufferOperation.Value.Syntax, isBufferNamed,
                                                          offsetOperation.Value.Syntax, isOffsetNamed,
                                                          countOperation.Value.Syntax, isCountNamed,
                                                          cancellationTokenOperation?.Value.Syntax, isCancellationTokenNamed);
 
-            var action = new MyCodeAction(
+            context.RegisterCodeFix(
+                new MyCodeAction(
                     title: title,
                     createChangedDocument,
-                    equivalenceKey: title + invocation.TargetMethod.Name);
-
-            context.RegisterCodeFix(action, context.Diagnostics);
+                    equivalenceKey: title + invocation.TargetMethod.Name),
+                context.Diagnostics);
         }
 
         private static Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName,
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,0 +1,100 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using System.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic), Shared]
+    public class PreferStreamAsyncMemoryOverloadsFixer : CodeFixProvider
+    {
+        public sealed override ImmutableArray<string> FixableDiagnosticIds =>
+            ImmutableArray.Create(PreferStreamAsyncMemoryOverloads.RuleId);
+
+        public sealed override FixAllProvider GetFixAllProvider() =>
+            WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            CancellationToken ct = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(ct).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode node)
+            {
+                SemanticModel model = await doc.GetSemanticModelAsync(ct).ConfigureAwait(false);
+                if (model.GetOperation(node, ct) is IInvocationOperation invocation)
+                {
+                    string methodName = invocation.TargetMethod.Name;
+
+                    string title;
+                    if (methodName == "ReadAsync")
+                    {
+                        title = MicrosoftNetCoreAnalyzersResources.PreferStreamReadAsyncMemoryOverloadsTitle;
+                    }
+                    else if (methodName == "WriteAsync")
+                    {
+                        title = MicrosoftNetCoreAnalyzersResources.PreferStreamWriteAsyncMemoryOverloadsTitle;
+                    }
+                    else
+                    {
+                        return;
+                    }
+
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            title: title,
+                            createChangedDocument: c => FixInvocation(doc, root, invocation, methodName),
+                            equivalenceKey: MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsMessage),
+                        context.Diagnostics);
+                }
+            }
+        }
+
+        private static Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName)
+        {
+            SyntaxGenerator generator = SyntaxGenerator.GetGenerator(doc);
+
+            // The stream object
+            SyntaxNode instanceNode = invocation.Instance.Syntax;
+
+            // Need the byte array object so we can invoke its AsMemory() method
+            SyntaxNode bufferInstanceNode = invocation.Arguments[0].Value.Syntax; // byte[] buffer
+
+            // These arguments are not modified, just moved inside AsMemory
+            SyntaxNode offsetNode = invocation.Arguments[1].Syntax; // int offset
+            SyntaxNode countNode = invocation.Arguments[2].Syntax;  // int count
+
+            // Generate an invocation of the AsMemory() methdo from the byte array object
+            SyntaxNode asMemoryExtensionMethodNode = generator.IdentifierName("AsMemory");
+            SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferInstanceNode, asMemoryExtensionMethodNode);
+            SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(asMemoryExpressionNode, offsetNode, countNode);
+
+            // Create a new async method call for the stream object, no arguments yet
+            SyntaxNode asyncMethodIdentifierNode = generator.IdentifierName(methodName);
+            SyntaxNode asyncMethodNode = generator.MemberAccessExpression(instanceNode, asyncMethodIdentifierNode);
+
+            // Add the arguments to the async method call, with or without CancellationToken
+            SyntaxNode newInvocationExpression;
+            if (invocation.Arguments.Length > 3)
+            {
+                newInvocationExpression = generator.InvocationExpression(
+                    asyncMethodNode, asMemoryInvocationNode,
+                    invocation.Arguments[3].Syntax /* CancellationToken */);
+            }
+            else
+            {
+                newInvocationExpression = generator.InvocationExpression(asyncMethodNode, asMemoryInvocationNode);
+            }
+
+            SyntaxNode newInvocation = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression);
+            return Task.FromResult(doc.WithSyntaxRoot(newInvocation));
+        }
+    }
+}
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Collections.Immutable;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
@@ -114,15 +113,20 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             // buffer.AsMemory(int start, int length)
             // offset should become start
             // count should become length
-            SyntaxNode namedStartNode = isOffsetNamed ? generator.Argument("start", RefKind.None, offsetValueNode) : offsetValueNode;
-            SyntaxNode namedLengthNode = isCountNamed ? generator.Argument("length", RefKind.None, countValueNode) : countValueNode;
+            SyntaxNode namedStartNode = isOffsetNamed ? generator.Argument(name: "start", RefKind.None, offsetValueNode) : offsetValueNode;
+            SyntaxNode namedLengthNode = isCountNamed ? generator.Argument(name: "length", RefKind.None, countValueNode) : countValueNode;
 
             // Generate an invocation of the AsMemory() method from the byte array object, using the correct named arguments
-            SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferValueNode, "AsMemory");
-            SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(asMemoryExpressionNode, namedStartNode, namedLengthNode);
+            SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferValueNode.WithoutTrivia(), memberName: "AsMemory");
+            SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(
+                asMemoryExpressionNode,
+                namedStartNode.WithTriviaFrom(offsetValueNode),
+                namedLengthNode.WithTriviaFrom(countValueNode));
 
             // Generate the new buffer argument, ensuring we include the argument name if the user originally indicated one
-            SyntaxNode namedBufferNode = isBufferNamed ? generator.Argument("buffer", RefKind.None, asMemoryInvocationNode) : asMemoryInvocationNode;
+            SyntaxNode namedBufferWithAsMemoryInvocationNode =
+                    (isBufferNamed ? generator.Argument(name: "buffer", RefKind.None, asMemoryInvocationNode) : asMemoryInvocationNode)
+                .WithTriviaFrom(bufferValueNode);
 
             // Create an async method call for the stream object with no arguments
             SyntaxNode asyncMethodNode = generator.MemberAccessExpression(streamInstanceNode, methodName);
@@ -131,84 +135,17 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             SyntaxNode[] nodeArguments;
             if (cancellationTokenValueNode != null)
             {
-                SyntaxNode namedCancellationTokenNode = isCancellationTokenNamed ? generator.Argument("cancellationToken", RefKind.None, cancellationTokenValueNode) : cancellationTokenValueNode;
-                nodeArguments = new SyntaxNode[] { namedBufferNode, namedCancellationTokenNode };
+                SyntaxNode namedCancellationTokenNode = isCancellationTokenNamed ? generator.Argument(name: "cancellationToken", RefKind.None, cancellationTokenValueNode) : cancellationTokenValueNode;
+                nodeArguments = new SyntaxNode[] { namedBufferWithAsMemoryInvocationNode, namedCancellationTokenNode.WithTriviaFrom(cancellationTokenValueNode) };
             }
             else
             {
-                nodeArguments = new SyntaxNode[] { namedBufferNode };
+                nodeArguments = new SyntaxNode[] { namedBufferWithAsMemoryInvocationNode };
             }
-            SyntaxNode newInvocationExpression = generator.InvocationExpression(asyncMethodNode, nodeArguments);
+            SyntaxNode newInvocationExpression = generator.InvocationExpression(asyncMethodNode, nodeArguments).WithTriviaFrom(streamInstanceNode);
 
-            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression);
+            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
             return Task.FromResult(doc.WithSyntaxRoot(newRoot));
-        }
-    }
-
-    [ExportCodeFixProvider(LanguageNames.CSharp)]
-    public sealed class PreferStreamAsyncMemoryOverloadsCSharpFixer : PreferStreamAsyncMemoryOverloadsFixer
-    {
-        protected override IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed)
-        {
-            isNamed = false;
-
-            // The expected position is beyond the total arguments, so we don't expect to find the argument in the array
-            if (index >= args.Length)
-            {
-                return null;
-            }
-            // If the argument in the specified index does not have a name, then it is in its expected position
-            else if (args[index].Syntax is CodeAnalysis.CSharp.Syntax.ArgumentSyntax argNode && argNode.NameColon == null)
-            {
-                return args[index];
-            }
-            // Otherwise, find it by name
-            else
-            {
-                isNamed = true;
-                return args.FirstOrDefault(argOperation =>
-                {
-                    return argOperation.Syntax is CodeAnalysis.CSharp.Syntax.ArgumentSyntax argNode &&
-                           argNode.NameColon != null &&
-                           argNode.NameColon.Name != null &&
-                           argNode.NameColon.Name.Identifier != null &&
-                           argNode.NameColon.Name.Identifier.ValueText == name;
-                });
-            }
-        }
-
-    }
-
-    [ExportCodeFixProvider(LanguageNames.VisualBasic)]
-    public sealed class PreferStreamAsyncMemoryOverloadsVisualBasicFixer : PreferStreamAsyncMemoryOverloadsFixer
-    {
-        protected override IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed)
-        {
-            isNamed = false;
-
-            // The expected position is beyond the total arguments, so we don't expect to find the argument in the array
-            if (index >= args.Length)
-            {
-                return null;
-            }
-            // If the argument in the specified index does not have a name, then it is in its expected position
-            else if (args[index].Syntax is CodeAnalysis.VisualBasic.Syntax.SimpleArgumentSyntax argNode && argNode.NameColonEquals == null)
-            {
-                return args[index];
-            }
-            // Otherwise, find it by name
-            else
-            {
-                isNamed = true;
-                return args.FirstOrDefault(argOperation =>
-                {
-                    return argOperation.Syntax is CodeAnalysis.VisualBasic.Syntax.SimpleArgumentSyntax simpleArgNode &&
-                           simpleArgNode.NameColonEquals != null &&
-                           simpleArgNode.NameColonEquals.Name != null &&
-                           simpleArgNode.NameColonEquals.Name.Identifier != null &&
-                           simpleArgNode.NameColonEquals.Name.Identifier.ValueText == name;
-                });
-            }
         }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Collections.Immutable;
-using System.Composition;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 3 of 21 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                    else
+                    {
+                        return;
+                    }
+
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            title: title,
+                            createChangedDocument: c => FixInvocation(doc, root, invocation, methodName),
```
carlossanlop, line 53: Should I take out anything from the `FixInvocation` method and put it before `context.RegisterCodeFix`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
```diff
@@ -139,21 +140,7 @@ private void AnalyzeCompilationStart(CompilationStartAnalysisContext context)
                 ParameterInfo.GetParameterInfo(cancellationTokenType)
             };
 
-            // Create the arrays with the exact parameter order of the desired methods
```
carlossanlop, line 142: I noticed these were unused in the analyzer code. The preferred method extraction did not require these arrays.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
```diff
@@ -252,7 +254,7 @@ private void AnalyzeCompilationStart(CompilationStartAnalysisContext context)
                         return;
                     }
 
-                    context.ReportDiagnostic(awaitOperation.Operation.CreateDiagnostic(rule, ruleMessageMethod, ruleMessagePreferredMethod));
+                    context.ReportDiagnostic(invocation.CreateDiagnostic(rule, ruleMessageMethod, ruleMessagePreferredMethod));
```
carlossanlop, line 257: I noticed the fixer was catching the whole await operation. Now the fixer will only receive the actual method invocation: `stream.ReadAsync(...)` or `stream.WriteAsync(...)`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
```diff
@@ -5,26 +5,34 @@
...
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
     Microsoft.NetCore.Analyzers.Runtime.PreferStreamAsyncMemoryOverloads,
-    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+    Microsoft.NetCore.Analyzers.Runtime.PreferStreamAsyncMemoryOverloadsFixer>;
 using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
     Microsoft.NetCore.Analyzers.Runtime.PreferStreamAsyncMemoryOverloads,
-    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+    Microsoft.NetCore.Analyzers.Runtime.PreferStreamAsyncMemoryOverloadsFixer>;
+
+#pragma warning disable CA1305 // Specify IFormatProvider in string.Format
```
carlossanlop, line 13: I found another unit test doing this. I hope it's ok.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -19,10 +20,10 @@ class C
 {
     public void M()
     {
-        using (FileStream fs = File.Open(""file.txt"", FileMode.Open))
+        using (FileStream s = File.Open(""file.txt"", FileMode.Open))
```
carlossanlop, line 23: Renamed the stream object to `s` in all the unit tests so that I could format the source code text the same way for all tests.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -199,10 +200,52 @@ class C
...
+using System.IO;
+using System.Threading;
+class C
+{
+    public async void M()
+    {
+        using (FileStream s = File.Open(""path.txt"", FileMode.Open))
+        {
+            byte[] buffer = new byte[s.Length];
+            await s.ReadAsync(buffer);
```
carlossanlop, line 227: Added these extra new unit tests: If the user calls `ReadAsync` with only a buffer (no offset and count), it gets automatically casted to a `Memory` object. Same with `WriteAsync` -> `ReadOnlyMemory`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,100 @@
...
+            // Need the byte array object so we can invoke its AsMemory() method
+            SyntaxNode bufferInstanceNode = invocation.Arguments[0].Value.Syntax; // byte[] buffer
+
+            // These arguments are not modified, just moved inside AsMemory
+            SyntaxNode offsetNode = invocation.Arguments[1].Syntax; // int offset
+            SyntaxNode countNode = invocation.Arguments[2].Syntax;  // int count
+
+            // Generate an invocation of the AsMemory() methdo from the byte array object
+            SyntaxNode asMemoryExtensionMethodNode = generator.IdentifierName("AsMemory");
+            SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferInstanceNode, asMemoryExtensionMethodNode);
```
Jozkee, line 76:
```suggestion
            SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferInstanceNode, "AsMemory");
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,100 @@
...
+            SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(asMemoryExpressionNode, offsetNode, countNode);
+
+            // Create a new async method call for the stream object, no arguments yet
+            SyntaxNode asyncMethodIdentifierNode = generator.IdentifierName(methodName);
+            SyntaxNode asyncMethodNode = generator.MemberAccessExpression(instanceNode, asyncMethodIdentifierNode);
```
Jozkee, line 81:
```suggestion
            SyntaxNode asyncMethodNode = generator.MemberAccessExpression(instanceNode, methodName);
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
Jozkee: I would pass the required `IFormatProvider` instead, unless there is a reason to not do it.

carlossanlop: It's not affecting the way the unit tests behave, so I considered it unimportant.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: There should be a Roslyn analyzer that suggest this 😄

carlossanlop:
I also need to resolve @mavasani 's request to group the description and title messages into one: https://github.com/dotnet/roslyn-analyzers/issues/3586

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public sealed override FixAllProvider GetFixAllProvider() =>
+            WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            CancellationToken ct = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(ct).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode node)
```
mavasani, line 29: In general, it is better to reduce nesting by writing this as

```csharp
if (!(root.FindNode(context.Span) is SyntaxNode node))
{
   return;
}

var model = await doc.GetSemanticModelAsync(ct).ConfigureAwait(false);
if (!(model.GetOperation(node, ct) is IInvocationOperation invocation))
{
   return;
}

...
...
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: Nope, this looks good. The only rule of thumb is that we make sure that once we register a code fix, it is indeed applicable and produces a changed document when the registered code action is applied by the user.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+                    else
+                    {
+                        return;
+                    }
+
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            title: title,
+                            createChangedDocument: c => FixInvocation(doc, root, invocation, methodName),
+                            equivalenceKey: MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsMessage),
```
mavasani, line 54: Using a common equivalence key between code actions means that when a user invokes a FixAll operation for "ReadAsync" violation, it will also fix all "WriteAsync" violations. Is that what we want? If so, then this is correct. Otherwise, you can use `equivalenceKey: title`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+            // Create a new async method call for the stream object, no arguments yet
+            SyntaxNode asyncMethodNode = generator.MemberAccessExpression(instanceNode, methodName);
+
+            // Add the arguments to the async method call, with or without CancellationToken
+            SyntaxNode newInvocationExpression;
+            if (invocation.Arguments.Length > 3)
+            {
+                newInvocationExpression = generator.InvocationExpression(
+                    asyncMethodNode, asMemoryInvocationNode,
+                    invocation.Arguments[3].Syntax /* CancellationToken */);
```
mavasani, line 87: Can arguments be more then 4? The above if check indicates it can be `4 or more`. I would recommend that if the only known case here is 4 arguments, then add a bail out in RegisterCodeFixesAsync when argument count `>4` before invoking RegisterCodeFix, and change the check above to `if (invocation.Arguments.Length == 4)`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
```diff
@@ -263,26 +265,30 @@ private void AnalyzeCompilationStart(CompilationStartAnalysisContext context)
...
             INamedTypeSymbol streamType,
+            out IInvocationOperation? actualInvocation,
             out IMethodSymbol? actualMethod)
         {
+            actualInvocation = null;
             actualMethod = null;
 
             // The await should have a known operation child, check its kind
-            if (!(awaitOperation.Operation is IInvocationOperation invocation))
+            if (!(awaitOperation.Operation is IInvocationOperation childOperation))
```
mavasani, line 275:
```suggestion
            if (!(awaitOperation.Operation is IInvocationOperation awaitedInvocation))
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
```diff
@@ -5,26 +5,34 @@
...
 
 namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
 {
     public class PreferStreamAsyncMemoryOverloadsTestBase
     {
         protected static Task AnalyzeCSAsync(string source, params DiagnosticResult[] expected) =>
-            AnalyzeCSForVersionAsync(source, ReferenceAssemblies.NetCore.NetCoreApp50, expected);
+            VerifyCSForVersionAsync(source, null, ReferenceAssemblies.NetCore.NetCoreApp50, expected);
+
+        protected static Task FixCSAsync(string originalSource, string fixedSource, params DiagnosticResult[] expected) =>
```
mavasani, line 22: Nit: for consistency with naming in other test files

```suggestion
        protected static Task CSharpVerifyCodeFixAsync(string originalSource, string fixedSource, params DiagnosticResult[] expected) =>
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -400,234 +443,223 @@ End Class
...
-        }
+        #endregion
 
-        [Fact]
-        public Task VB_Analyzer_Diagnostic_ConfigureAwait()
+        #region Helpers
+
+        private const string AsyncMethodName = "Read";
+
+        private Task GetCSReadAsyncDiagnostic(string source, string originalArgs, string fixedArgs, bool withConfigureAwait, int startLine, int startColumn, int endLine, int endColumn)
```
mavasani, line 641: This naming seems weird as it is not getting diagnostic, it is actually verifying the code fix. We should rename this to something like say `CSharpVerifyCodeFixAsync`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -400,234 +443,223 @@ End Class
...
-    Public Async Sub M()
-        Using fs As FileStream = File.Open(""file.txt"", FileMode.Open)
-            Dim buffer As Byte() = New Byte(fs.Length - 1) {}
-            Await fs.ReadAsync(buffer, 0, fs.Length).ConfigureAwait(False)
-        End Using
-    End Sub
-End Class
-            ", GetVBResult(9, 19, 9, 75));
+            return FixCSAsync(
+                GetFormattedSourceCode(source, AsyncMethodName, originalArgs, withConfigureAwait, LanguageNames.CSharp),
```
mavasani, line 644: None of our unit tests need to do any formatting of input/output source code. Why is this needed?

mavasani: Ah, so I read the implementation of `GetFormattedSourceCode` and the helper is not actually performing whitespace formatting, but producing input/output code formatted based on the arguments passed in here. Seems reasonable, but the name might confuse some readers.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -400,234 +443,223 @@ End Class
...
+                GetFormattedSourceCode(source, AsyncMethodName, fixedArgs, withConfigureAwait, LanguageNames.CSharp),
+                GetCSResult(startLine, startColumn, endLine, endColumn));
         }
 
-        [Fact]
-        public Task VB_Analyzer_Diagnostic_CancellationToken_ConfigureAwait()
+        private Task GetVBReadAsyncDiagnostic(string source, string originalArgs, string fixedArgs, bool withConfigureAwait, int startLine, int startColumn, int endLine, int endColumn)
```
mavasani, line 649:
```suggestion
        private Task VisualBasicVerifyCodeFixAsync(string source, string originalArgs, string fixedArgs, bool withConfigureAwait, int startLine, int startColumn, int endLine, int endColumn)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -400,234 +443,223 @@ End Class
...
+        [InlineData("New Byte(s.Length - 1) {}, 0, s.Length",
+                    "(New Byte(s.Length - 1) {}).AsMemory(0, s.Length)",
+                    true, 70)]
+        [InlineData("New Byte(s.Length - 1) {}, 0, s.Length, New CancellationToken()",
+                    "(New Byte(s.Length - 1) {}).AsMemory(0, s.Length), New CancellationToken()",
+                    false, 95)]
+        [InlineData("New Byte(s.Length - 1) {}, 0, s.Length, New CancellationToken()",
+                    "(New Byte(s.Length - 1) {}).AsMemory(0, s.Length), New CancellationToken()",
+                    true, 95)]
+        public Task VB_Analyzer_Diagnostic_InlineByteArray(string originalArgs, string fixedArgs, bool withConfigureAwait, int endColumn)
```
mavasani, line 618: Don't like hard coding the `endColumn` number.

mavasani: Can we change the code fix tests to be markup syntax based so there is no need for any magic numbers in tests?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -400,234 +443,223 @@ End Class
...
-            ", GetVBResult(9, 19, 9, 53));
-        }
-
-        [Fact]
-        public Task VB_Analyzer_Diagnostic_Inline()
-        {
-            return AnalyzeVBAsync(@"
+End Module
+            ";
+            return GetVBReadAsyncDiagnostic(source, originalArgs, fixedArgs, withConfigureAwait, 9, 19, 9, endColumn);
```
mavasani, line 602: Ideally, we can just get rid of the last 4 arguments here by switching to markup syntax for code fix tests.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic), Shared]
+    public class PreferStreamAsyncMemoryOverloadsFixer : CodeFixProvider
```
mavasani, line 16: Nit:

```suggestion
    public sealed class PreferStreamAsyncMemoryOverloadsFixer : CodeFixProvider
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani, line 29: I think you need to specify `getInnermostNodeForTie = true` for case where invocation itself is an argument to some other invocation. Good to validate with a unit test.

```suggestion
            if (root.FindNode(context.Span, getInnermostNodeForTie: true) is SyntaxNode node)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: I cannot not use markup syntax `[||]` or `{||}`. I get an error about multiple diagnostics being detected. 
This was a suggestion given in the Analyzer PR as well, and we opted to use column names instead.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: See my answer above.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
mavasani: Ah yes, sorry my bad that I forgot again. Ignore the suggestion about markup syntax.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+            SyntaxGenerator generator = SyntaxGenerator.GetGenerator(doc);
+
+            // The stream object
+            SyntaxNode instanceNode = invocation.Instance.Syntax;
+
+            // Need the byte array object so we can invoke its AsMemory() method
+            SyntaxNode bufferInstanceNode = invocation.Arguments[0].Value.Syntax; // byte[] buffer
+
+            // These arguments are not modified, just moved inside AsMemory
+            SyntaxNode offsetNode = invocation.Arguments[1].Syntax; // int offset
```
mavasani, line 71: Can we add unit tests where `offset` and `count` are specified in reverse order via named arguments? I think your implementation will handle these correctly and also maintain the named arguments for AsMemory invocation, but would be good to validate with unit test.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            CancellationToken ct = context.CancellationToken;
+            SyntaxNode root = await doc.GetSyntaxRootAsync(ct).ConfigureAwait(false);
+            if (root.FindNode(context.Span) is SyntaxNode node)
+            {
+                SemanticModel model = await doc.GetSemanticModelAsync(ct).ConfigureAwait(false);
+                if (model.GetOperation(node, ct) is IInvocationOperation invocation)
+                {
```
mavasani, line 33: Defensive check?

```suggestion
                if (model.GetOperation(node, ct) is IInvocationOperation invocation && invocation.Arguments.Length >= 3)
                {
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: Alternatively, can we simplify the code by completely removing the arguments length check? Something like:

```csharp
var argumentsBuilder = ImmutableArray<SyntaxNode>.CreateBuilder();
argumentsBuilder.Add(asyncMethodNode);
argumentsBuilder.Add(asMemoryInvocationNode);
foreach (var arg in invocation.Arguments.Skip(3))
{
   argumentsBuilder.Add(arg.Syntax);
}

var newInvocation = generator.InvocationExpression(argumentsBuilder.ToImmutable());
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+                newInvocationExpression = generator.InvocationExpression(
+                    asyncMethodNode, asMemoryInvocationNode,
+                    invocation.Arguments[3].Syntax /* CancellationToken */);
+            }
+            else
+            {
+                newInvocationExpression = generator.InvocationExpression(asyncMethodNode, asMemoryInvocationNode);
+            }
+
+            SyntaxNode newInvocation = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression);
```
mavasani, line 94:
```suggestion
            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,98 @@
...
+            return Task.FromResult(doc.WithSyntaxRoot(newInvocation));
```
mavasani, line 95:
```suggestion
            return Task.FromResult(doc.WithSyntaxRoot(newRoot));
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
```diff
@@ -36,27 +44,58 @@ protected static DiagnosticResult GetVBResultForRule(int startLine, int startCol
...
+            string configureAwait = string.Empty;
+
+            if (withConfigureAwait)
+            {
+                char booleanArgumentInitial = 'f';
+                if (language == LanguageNames.VisualBasic)
+                {
+                    booleanArgumentInitial = 'F';
+                }
+                configureAwait = $".ConfigureAwait({booleanArgumentInitial}alse)";
```
mavasani, line 58:
```suggestion
                var arg = LanguageNames.VisualBasic ? "False" : "false";
                configureAwait = $".ConfigureAwait({arg})";
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -400,234 +443,223 @@ End Class
...
-            return AnalyzeCSAsync(@"
+        #region C# - Diagnostic - Analyzer
+
+        [Theory]
+        [InlineData("buffer, 0, (int)s.Length",
+                    "buffer.AsMemory(0, (int)s.Length)",
+                    false, 56)]
+        [InlineData("buffer, 0, (int)s.Length",
+                    "buffer.AsMemory(0, (int)s.Length)",
+                    true, 56)]
```
mavasani, line 509: Does it make sense to add an additional test parameter `withAwait` and an additional `InlineData(withAwait: false)` to verify that no diagnostic is reported when `withAwait` is `false`? Or do you feel this is already adequately tested in prior tests to not warrant adding that scenario to this test? Anyway is fine to me, this is just a suggestion in case you feel it will solidify your tests..

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamWriteAsyncMemoryOverloadsTests.cs
```diff
@@ -385,378 +428,284 @@ End Class
...
+    {{
+        using (FileStream s = new FileStream(""path.txt"", FileMode.Create))
+        {{
+            byte[] buffer = {{ 0xBA, 0x5E, 0xBA, 0x11, 0xF0, 0x07, 0xBA, 0x11 }};
+            await {0};
+        }}
+    }}
+}}
+            ";
+            return GetCSWriteAsyncDiagnostic(source, originalArgs, fixedArgs, withConfigureAwait, 12, 19, 12, endColumn);
```
mavasani, line 521: Same suggestion here about naming. Either use `CSharpVerifyCodeFixAsync` or `VerifyCSharpCodeFixAsync` to clarify the actual method semantics.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Sure, but I'll make sure to avoid the if nesting, as suggested in the other comment.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: I would like to be cautious and only let the user apply Fix All for methods with the exact same name, not for both names.

carlossanlop: I'll change this to title, but we need to keep in mind that I will merge the two title messages and the two description messages, as requested in the other issue.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Sure, I can do that. Good catch.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: Note you can also just append the `invocation.TargetMethod.Name` to the key to ensure separate equivalence class based on method name.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: The undesired methods can have either 3 or 4 arguments:
```cs
Stream.WriteAsync(Byte[], Int32, Int32)
Stream.WriteAsync(Byte[], Int32, Int32, CancellationToken)
Stream.ReadAsync(Byte[], Int32, Int32)
Stream.ReadAsync(Byte[], Int32, Int32, CancellationToken)
```

carlossanlop: I'll use `invocation.Arguments.Length == 4`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: I think it's sufficiently tested in the tests that are not expected to detect a diagnostic.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: 😃

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: Let me change the name for clarity

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamWriteAsyncMemoryOverloadsTests.cs
carlossanlop: Dibe

carlossanlop: Done

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Done

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: @mavasani I added this test and it fails because the arguments get moved to the new `AsMemory` call with the exact same names that were being used in the ReadAsync/WriteAsync method, but the argument names are not the same in `AsMemory`:

ReadAsync/WriteAsync => int offset, int count
AsMemory => int start, int length

I think I should remove the name from the argument before adding it to the AsMemory call. I'm investigating how to do that. If you think I should do something differently, let me know, @mavasani.

mavasani: Can we just map the argument names, i.e. replace offset with start and count with length? Removing the name might not be desirable, especially if user prefers named arguments for constant values, which seem could be likely for these calls. You might have to add C# and VB specific code to check if argument syntax nodes have a [NameColon](http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Generated/Syntax.xml.Syntax.Generated.cs,2059) and [this](http://sourceroslyn.io/#Microsoft.CodeAnalysis.Workspaces/Editing/SyntaxGenerator.cs,1992) overload to create new argument syntax nodes.

mavasani: Basically, all arguments that had an explicit named argument should have a named argument in fixed code. All positional arguments without an explicit name should be retained as positional arguments.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,112 @@
...
+                newInvocationExpression = generator.InvocationExpression(
+                    asyncMethodNode, asMemoryInvocationNode,
+                    invocation.Arguments[3].Syntax /* CancellationToken */);
+            }
+            else
+            {
+                newInvocationExpression = generator.InvocationExpression(asyncMethodNode, asMemoryInvocationNode);
+            }
+
+            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression);
```
carlossanlop, line 108: I should also include the trivia.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Understood. I'll make sure to keep the names. Thanks for the help.

carlossanlop: I also discovered that my fixer assumes the user will place the arguments in the correct location. If the cancellationToken or the buffer are located where the offset or the count are expected to be located, that's the argument the fixer ends up putting inside the AsMemory call. I'll address this.

carlossanlop: The analyzer expects ReadAsync/WriteAsync to have its arguments in this order:
- `byte[], int, int`
- `byte[], int, int, CancellationToken`

But if the user calls the method with named arguments in the wrong order (with the int data types in the wrong order as well), then the analyzer does not throw a diagnostic. Should I fix this or should we narrow down the analyzer to only work when the user puts the arguments in the correct order, or at the very least, the two `int` arguments in the 2nd and 3rd position?

mavasani: We should fix the analyzer. Use of named or positional arguments should not change the analysis behavior. I thought the IOperation API for invocation.Arguments abstracts out the user specified order in the sense that Arguments[0] always corresponds to the argument for the first parameter regardless of whether it was specified as the first argument by the user or at some other point using named argument. Do you not observe this?

mavasani: I looked at your analyzer implementation and I cannot figure out where in AnalyzeInvocation you perform any argument related checks- you only compare target method of invocation with one of the analyzed methods. That is correct implementation, and I would expect it to handle the case of out of order named arguments.

carlossanlop: > Do you not observe this?

I do not. I wrote a small unit test that attempts to fix the code example above and the arguments are not being found in the correct order, which makes me think I should find each argument by name, in the array of arguments, and I should not assume they will be located in the expected position.

mavasani: You are right about the Operation tree shape. `invocation.Arguments` are in source order, not in the parameter order: https://github.com/dotnet/roslyn/blob/e704ca635bd6de70a0250e34c4567c7a28fa9f6d/src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_IArgument.cs#L145-L177

You would need to change your fixer to use `IArgumentOperation.Parameter.Ordinal` to identify the arguments corresponding to first two parameters (ordinal 0 and 1). You would need C# and VB specific implementations to identify if they are named arguments, and if so generate a named argument in fixed code. 

However, that still doesn't answer the question why your analyzer is not reporting a diagnostic for out of order named arguments case. It is not comparing any argument ordering, so I would expect the analyzer handle this case properly.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: Yes, and please do add unit test for the trivia case!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: > However, that still doesn't answer the question why your analyzer is not reporting a diagnostic for out of order named arguments case
You're correct. The analyzer does not have a problem, only the fixer. The unit test I wrote had the wrong `endColumn` specified, that was why the diagnostic was not calling the fixer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Will do!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: I fixed the named arguments case. Please take a look.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Still pending to address trivia. The latest commits were just for the named arguments.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Pending adding a unit test.

carlossanlop: Unit test added, and it passed! Please take a look.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+            }
+            SyntaxNode newInvocationExpression = generator.InvocationExpression(asyncMethodNode, nodeArguments);
+
+            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression);
+            return Task.FromResult(doc.WithSyntaxRoot(newRoot));
+        }
+    }
+
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class PreferStreamAsyncMemoryOverloadsCSharpFixer : PreferStreamAsyncMemoryOverloadsFixer
```
mavasani, line 149: The language specific implementation should be moved to `Microsoft.CodeAnalysis.CSharp.NetAnalyzers.csproj`. Implementing it here would lead to performance issues where a solution with VB only projects will cause C# Roslyn assemblies to get loaded in VS, which will be a performance hit from unwanted assembly loads.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            isNamed = false;
+
+            // The expected position is beyond the total arguments, so we don't expect to find the argument in the array
+            if (index >= args.Length)
+            {
+                return null;
+            }
+            // If the argument in the specified index does not have a name, then it is in its expected position
+            else if (args[index].Syntax is CodeAnalysis.CSharp.Syntax.ArgumentSyntax argNode && argNode.NameColon == null)
```
mavasani, line 161: Should be able simplify once this type is moved to C# layer and you can add `using Microsoft.CodeAnalysis.CSharp.Syntax;` at the top of the file.

```suggestion
            else if (args[index].Syntax is ArgumentSyntax argNode && argNode.NameColon == null)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+            else
+            {
+                isNamed = true;
+                return args.FirstOrDefault(argOperation =>
+                {
+                    return argOperation.Syntax is CodeAnalysis.CSharp.Syntax.ArgumentSyntax argNode &&
+                           argNode.NameColon != null &&
+                           argNode.NameColon.Name != null &&
+                           argNode.NameColon.Name.Identifier != null &&
+                           argNode.NameColon.Name.Identifier.ValueText == name;
```
mavasani, line 175:
```suggestion
                           argNode.NameColon?.Name.Identifier.ValueText == name;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+                });
+            }
+        }
+
+    }
+
+    [ExportCodeFixProvider(LanguageNames.VisualBasic)]
+    public sealed class PreferStreamAsyncMemoryOverloadsVisualBasicFixer : PreferStreamAsyncMemoryOverloadsFixer
```
mavasani, line 183: This one should move to VB layer: https://github.com/dotnet/roslyn-analyzers/tree/master/src/NetAnalyzers/VisualBasic

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+            {
+                return;
+            }
+
+            // No nullcheck for this, because there is an overload that may not contain it
+            IArgumentOperation? cancellationTokenOperation = GetArgumentByPositionOrName(invocation.Arguments, 3, "cancellationToken", out bool isCancellationTokenNamed);
+
+            string title = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle;
+
+            Task<Document> fixInvocation = FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
```
mavasani, line 86: Invoking this method here will actually do the work of computing the fixed document, even before the user applied the fix. We should make it lazy so it is done only when user applies the code fix.
```suggestion
            Func<CancellationToken, Task<Document>> createChangedDocument = _ => FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+                                                         bufferOperation.Value.Syntax, isBufferNamed,
+                                                         offsetOperation.Value.Syntax, isOffsetNamed,
+                                                         countOperation.Value.Syntax, isCountNamed,
+                                                         cancellationTokenOperation?.Value.Syntax, isCancellationTokenNamed);
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
```
mavasani, line 93: Interesting - I would have excepted this to flag [RS0005: DoNotUseCodeActionCreateRuleId](https://github.com/dotnet/roslyn-analyzers/blob/62863ff47591ccd19a7c543dda8b402978063752/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs#L11). I will investigate why that rule did not fire.

Meanwhile, switch to the pattern used here to ensure we get telemetry on this code action usage: https://github.com/dotnet/roslyn-analyzers/blob/79c2638f54b9e7ef0a3954cc838ec4c13ea63c1f/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ImplementSerializationConstructors.Fixer.cs#L102-L109

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+                    title: title,
+                    createChangedDocument: c => fixInvocation,
+                    equivalenceKey: title),
+                context.Diagnostics);
+        }
+
+        // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
+        protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
```
mavasani, line 101: Nit: Consider moving abstract method declarations before method implementations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,214 @@
...
+                                                         bufferOperation.Value.Syntax, isBufferNamed,
+                                                         offsetOperation.Value.Syntax, isOffsetNamed,
+                                                         countOperation.Value.Syntax, isCountNamed,
+                                                         cancellationTokenOperation?.Value.Syntax, isCancellationTokenNamed);
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    title: title,
+                    createChangedDocument: c => fixInvocation,
+                    equivalenceKey: title),
```
mavasani, line 96: Append method name to title to ensure that FixAll operation only fixes violations with matching method name.

```suggestion
                    equivalenceKey: title + invocation.TargetMethod.Name),
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
```diff
@@ -1,63 +1,286 @@
...
+        {
+            yield return new object[] { "buffer, 0, buffer.Length",
+                                        "buffer.AsMemory(0, buffer.Length)" };
+            yield return new object[] { "buffer, 0, buffer.Length, new CancellationToken()",
+                                        "buffer.AsMemory(0, buffer.Length), new CancellationToken()" };
+        }
+
+        public static IEnumerable<object[]> CSharpNamedArgumentsTestData()
+        {
+            // Normal argument order is: (byte[] buffer, int offset, int count)
```
mavasani, line 116: Nice extensive tests! Can we also add a test where some named arguments are followed by unnamed arguments? For example, a common case would be to only use named arguments for constant values.

```csharp
            yield return new object[] { "buffer, offset: 0, buffer.Length",
                                        "buffer.AsMemory(start: 0, buffer.Length)" };
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Trivia handling added. Please take a look.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
carlossanlop, line 1: @mavasani asked me on chat to move the C# and VB specific files to their appropriate projects.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Done

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Ah thanks. I always forget about this relatively new nullability feature.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Okay np.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: @mavasani do I need to do this for both the title and the equivalenceKey?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
carlossanlop: Done

carlossanlop: Added all the combinations of skipping names.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Okay so I appended the method name to both, and the tests passed. Let me know if this was the wrong thing to do so I only apply it to the equivalenceKey as in your code suggestion.

mavasani: Only to the equivalence key. Title is the user facing string that is shown to users in light bulb, and we want that to be the localized string without method name.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -80,25 +84,22 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
                                                          offsetOperation.Value.Syntax, isOffsetNamed,
                                                          countOperation.Value.Syntax, isCountNamed,
                                                          cancellationTokenOperation?.Value.Syntax, isCancellationTokenNamed);
 
             context.RegisterCodeFix(
-                CodeAction.Create(
-                    title: title,
-                    createChangedDocument: c => fixInvocation,
-                    equivalenceKey: title),
+                MyCodeAction.Create(
```
mavasani, line 96:
```suggestion
                new MyCodeAction(
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -80,25 +84,22 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                    title: titleAndMethodName,
```
mavasani, line 97:
```suggestion
                    title: title,
```

carlossanlop:
> LGTM. I have couple more suggestions:
> 
> 1. Use `new MyCodeAction` as `MyCodeAction.Create` will do the same thing as `CodeAction.Create`
> 2. Use localized title string without method name as the first argument for `title`.

Fixed, @mavasani. Thanks. I'll merge as soon as the build finishes.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1278,24 +1278,15 @@
   <data name="DoNotUseStackallocInLoopsMessage" xml:space="preserve">
     <value>Potential stack overflow. Move the stackalloc out of the loop.</value>
   </data>
-  <data name="PreferStreamReadAsyncMemoryOverloadsTitle" xml:space="preserve">
-    <value>Prefer 'Memory&lt;Byte&gt;' overload for 'ReadAsync'.</value>
+  <data name="PreferStreamAsyncMemoryOverloadsTitle" xml:space="preserve">
+    <value>Prefer the memory based overloads for 'ReadAsync' and 'WriteAsync'.</value>
```
stephentoub, line 1282:
```suggestion
    <value>Prefer the `Memory`-based overloads for 'ReadAsync' and 'WriteAsync'.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1278,24 +1278,15 @@
...
   </data>
-  <data name="PreferStreamReadAsyncMemoryOverloadsDescription" xml:space="preserve">
-    <value>All 'Stream' based classes provide a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as first argument. When possible, prefer the overload that takes a 'Memory&lt;Byte&gt;', which is more efficient.</value>
+  <data name="PreferStreamAsyncMemoryOverloadsDescription" xml:space="preserve">
+    <value>All 'Stream' based classes provide a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as the first argument, and a 'WriteAsync' overload that takes a 'ReadOnlyMemory&lt;Byte&gt;' as the first argument. Whenever possible, prefer the memory based overloads, which are more efficient.</value>
```
stephentoub, line 1285:
```suggestion
    <value>'Stream' has a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as the first argument, and a 'WriteAsync' overload that takes a 'ReadOnlyMemory&lt;Byte&gt;' as the first argument. Prefer calling the memory based overloads, which are more efficient.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: As suggested below, I should only do this for equivalenceKey, not for title. Reverted.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
carlossanlop: Thanks. Sounds better.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloadsTestBase.cs
carlossanlop: I just realized the unit tests failed for one of these tests. Good catch, @mavasani. I'm working on getting it fixed.

carlossanlop: Nevermind, it was not a unit test failure caused by an unhandled case. It was a typo in one of my new testdata yielded objects. I fixed it, they are passing.

PR #3594, opened 2020-05-01
Prefer string.Contains over string.IndexOf

pgovind:
Analyzer and code-fix for https://github.com/dotnet/runtime/issues/33785


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1287,4 +1287,19 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferStringContainsOverIndexOfDescription" xml:space="preserve">
+    <value>Calls to 'String.IndexOf' where the result is then just compared to -1 can instead be replaced by 'String.Contains'</value>
+  </data>
+  <data name="PreferStringContainsOverIndexOfMessage" xml:space="preserve">
+    <value>Use 'String.Contains' instead of 'String.IndexOf' to check for membership</value>
```
pgovind, line 1294: Change this: `check for membership` -> `to improve readability`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                                                                                      isPortedFxCopRule: false,
+                                                                                      isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilation =>
```
genlu, line 41: It's confusing to call a `CompilationStartAnalysisContext` instance `compilation`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
+                var charAndComparisonTypeArgumentContainsMethod = stringContainsMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(charType),
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
+
+                // Roslyn doesn't yet support "FindAllReferences" at a file/block level. So instead, find references to local variables in this block.
+                compilation.RegisterOperationBlockStartAction(context =>
+                {
+                    ISymbol symbol = context.OwningSymbol;
+                    if (!(context.OwningSymbol is IMethodSymbol method))
```
genlu, line 85: what about invocation inside an initializer?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                    {
+                        return;
+                    }
+
+                    List<Location>? leftOperandInvocationLocations = null;
+                    PooledConcurrentDictionary<string, int> variableNameToNumberOfReferences = PooledConcurrentDictionary<string, int>.GetInstance();
```
genlu, line 91: Pooled objects are not disposed.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -12,6 +12,7 @@ CA1830 | Performance | Info | PreferTypedStringBuilderAppendOverloads, [Document
 CA1831 | Performance | Warning | UseAsSpanInsteadOfRangeIndexerAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1831)
 CA1832 | Performance | Info | UseAsSpanInsteadOfRangeIndexerAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1832)
 CA1833 | Performance | Info | UseAsSpanInsteadOfRangeIndexerAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1833)
+CA1834 | Performance | Info | PreferStringContainsOverIndexOf, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1834)
```
carlossanlop, line 15: You might want to rebase your branch because CA1834 is already being used by your previous analyzer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PreferStringContainsOverIndexOfMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PreferStringContainsOverIndexOfDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                                      s_localizableTitle,
+                                                                                      s_localizableMessage,
+                                                                                      DiagnosticCategory.Performance,
+                                                                                      RuleLevel.IdeSuggestion,
```
carlossanlop, line 30: In the issue description, Immo categorized this rule as Performance. Is that still accurate?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
carlossanlop: I've seen in most other analyzers that this is called `context`.
Also, could you move the lambda expression contents into its own method? Makes it a bit easier to read.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                var stringArgumentIndexOfMethod = stringIndexOfMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(stringType));
+                var charArgumentIndexOfMethod = stringIndexOfMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(charType));
+                var stringAndComparisonTypeArgumentIndexOfMethod = stringIndexOfMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(stringType),
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
+                var charAndComparisonTypeArgumentIndexOfMethod = stringIndexOfMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(charType),
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
```
carlossanlop, line 66: Null check these?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                    .GetMembers("Contains")
+                    .OfType<IMethodSymbol>()
+                    .WhereAsArray(s =>
+                        s.Parameters.Length == 2);
+                var stringAndComparisonTypeArgumentContainsMethod = stringContainsMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(stringType),
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
+                var charAndComparisonTypeArgumentContainsMethod = stringContainsMethods.GetFirstOrDefaultMemberWithParameterInfos(
+                        ParameterInfo.GetParameterInfo(charType),
+                        ParameterInfo.GetParameterInfo(stringComparisonType));
```
carlossanlop, line 79: Null check these?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
carlossanlop: @genlu can you provide an example?

carlossanlop, line 85: You already saved it in a variable in the previous line:
```suggestion
                    if (!(symbol is IMethodSymbol method))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                        ILocalReferenceOperation localReference = (ILocalReferenceOperation)context.Operation;
+                        ILocalSymbol symbol = localReference.Local;
+                        variableNameToNumberOfReferences.AddOrUpdate(symbol.Name, 1, (name, addValue) =>
+                        {
+                            return addValue + 1;
+                        });
+                    },
+                    OperationKind.LocalReference);
+
+                    context.RegisterOperationAction(context =>
```
carlossanlop, line 105: Would you mind putting the contents of this lambda in its own method for readability?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                                if (InvocationOperationHasTargetMethod(leftInvocationOperation))
+                                {
+                                    leftOperandInvocationLocations = blockLocationList;
+                                }
+                            }
+                        }
+                    },
+                    OperationKind.Binary);
+
+                    context.RegisterOperationBlockEndAction(context =>
```
carlossanlop, line 176: Can you also put this lambda in its own method?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
```diff
@@ -0,0 +1,227 @@
...
+                                    continue;
+                                }
+
+                                context.ReportDiagnostic(variableNameAndLocation.Value.CreateDiagnostic(Rule));
+                            }
+                        }
+                    });
+                });
+
+                bool InvocationOperationHasTargetMethod(IInvocationOperation invocationOperation)
```
carlossanlop, line 204: Is there a benefit on having this method declared locally inside `Initialize`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
pgovind: Fixed!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
pgovind: Nope. I changed it to `Usage` now, since that seems to be the closest to Readability(which does not exist as a category)?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
pgovind: Don't have to actually, since I'm never dereferencing these.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
pgovind: These 2 are being null checked when they're values are needed. I can't do it upfront because they may be null in some versions of netstandard, but the method being tagged itself may not need these vars

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
pgovind: Yup, I put it there to re-use some vars declared above. I'll change it to a non-local method to improve readability

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
mavasani: `Design` is primarily used for API design guidelines. I would consider using `Usage` or `Maintainability` instead.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs
mavasani: I think Gen meant:
```csharp
class C
{
   public bool field = /*field initializer*/
}
```

Theoretically, any invocation can be present inside the field/property initializer, though it seems unlikely to have `string.IndexOf` invocations inside there.

```
 `public bool

File: src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs (hunk 4 above)
Comment (mavasani) [reply src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs:91]: Yeah, just add a Dispose call in the OperationBlockEndAction.

Comment (pgovind) [reply src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.cs:91]: You mean a `Free` call right? I added that and will push up that commit in a bit

File: src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx [hunk 12]
```diff
@@ -1314,6 +1314,15 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferStringContainsOverIndexOfDescription" xml:space="preserve">
+    <value>Calls to 'String.IndexOf' where the result is then just compared to -1 can instead be replaced by 'String.Contains'</value>
```
carlossanlop, line 1318: The wording here confused me the first couple of times I read it. Maybe it can be simplified?:
```suggestion
    <value>Calls to 'String.IndexOf' where the result is compared to -1 can be replaced by 'String.Contains'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+using Analyzer.Utilities.Extensions;
+using Analyzer.Utilities.PooledObjects;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// Prefer string.Contains over string.IndexOf when the result is compared to 1 or -1
```
carlossanlop, line 16: > to 1 or -1

If this is the case, then the resource message should also mention "1". It only mentions "-1" right now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOf.Fixer.cs
```diff
@@ -0,0 +1,144 @@
...
+                                {
+                                    containsInvocation = generator.InvocationExpression(containsExpression, indexOfMethodArguments[0].Syntax);
+                                }
+                                else
+                                {
+                                    containsInvocation = generator.InvocationExpression(containsExpression, indexOfMethodArguments[0].Syntax, indexOfMethodArguments[1].Syntax);
+                                }
+                            }
+                        }
+                        SyntaxNode newIfCondition = containsInvocation!;
```
pgovind, line 125: Bug here: containsInvocation can still be null here if the `StringComparison` argument is not a constant value but rather a variable

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                        ILocalSymbol symbol = localReference.Local;
+                        variableNameToNumberOfReferences.AddOrUpdate(symbol.Name, 1, (name, addValue) =>
+                        {
+                            return addValue + 1;
+                        });
+                    },
+                    OperationKind.LocalReference);
+
+                    context.RegisterOperationAction(HandleStringIndexOfMethod, OperationKind.Binary);
+
```
mavasani, line 105: Improve readability by explicitly partitioning the local functions from core code in the method.

```suggestion
                    context.RegisterOperationAction(HandleStringIndexOfMethod, OperationKind.Binary);

                    return;
                    
                    // Local functions
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                    context.RegisterOperationBlockEndAction(CheckInvocationOrNumberOfReferencesAndReportDiagnostic);
```
mavasani, line 106: Easier to map what callback each method maps to by using name that matches the action:

```suggestion
                    context.RegisterOperationBlockEndAction(OnOperationBlockEnd);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                        ILocalReferenceOperation localReference = (ILocalReferenceOperation)context.Operation;
+                        ILocalSymbol symbol = localReference.Local;
+                        variableNameToNumberOfReferences.AddOrUpdate(symbol.Name, 1, (name, addValue) =>
+                        {
+                            return addValue + 1;
+                        });
+                    },
+                    OperationKind.LocalReference);
+
+                    context.RegisterOperationAction(HandleStringIndexOfMethod, OperationKind.Binary);
```
mavasani, line 104:
```suggestion
                    context.RegisterOperationAction(AnalyzeBinaryOperation, OperationKind.Binary);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                    if (!(context.OwningSymbol is IMethodSymbol method))
+                    {
+                        return;
+                    }
+
+                    List<Location>? leftOperandInvocationLocations = null;
+                    PooledConcurrentDictionary<string, int> variableNameToNumberOfReferences = PooledConcurrentDictionary<string, int>.GetInstance();
+                    PooledConcurrentDictionary<string, ImmutableArray<Location>> variableNameToLocationsMap = PooledConcurrentDictionary<string, ImmutableArray<Location>>.GetInstance();
+
+                    context.RegisterOperationAction(context =>
```
mavasani, line 93: Can we consistent and define a local function even for this callback?
```suggestion
                    context.RegisterOperationAction(AnalyzeLocalReference, OperationKind.LocalReference);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                    {
+                        ILocalReferenceOperation localReference = (ILocalReferenceOperation)context.Operation;
+                        ILocalSymbol symbol = localReference.Local;
```
mavasani, line 96: Don't really need this local. You can inline to `localReference.Local.Name` below.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                        variableNameToNumberOfReferences.AddOrUpdate(symbol.Name, 1, (name, addValue) =>
+                        {
+                            return addValue + 1;
+                        });
```
mavasani, line 100:
```suggestion
                        variableNameToNumberOfReferences.AddOrUpdate(localReference.Local.Name, 1, (_, addValue) => addValue + 1);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+
+                // Roslyn doesn't yet support "FindAllReferences" at a file/block level. So instead, find references to local variables in this block.
+                context.RegisterOperationBlockStartAction(context =>
+                {
+                    if (!(context.OwningSymbol is IMethodSymbol method))
+                    {
+                        return;
+                    }
+
+                    List<Location>? leftOperandInvocationLocations = null;
```
mavasani, line 89: This is not a thread safe data structure to be used inside binary operation callback.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                    },
+                    OperationKind.LocalReference);
+
+                    context.RegisterOperationAction(HandleStringIndexOfMethod, OperationKind.Binary);
+
+                    context.RegisterOperationBlockEndAction(CheckInvocationOrNumberOfReferencesAndReportDiagnostic);
+
+                    void HandleStringIndexOfMethod(OperationAnalysisContext context)
+                    {
+                        IBinaryOperation blockOperation = (IBinaryOperation)context.Operation;
```
mavasani, line 110: So I think analyzing all Binary operations is not the correct approach here. You should instead register a callback for `OperationKind.Invocation`, check if the `invocation.TargetMethod` equals one of the methods that you care about, and if not bail out. If it is method of interest, you walk up the parent chain to find an ancestor BinaryOperation and then perform the checks that the other sub-tree of binary operation has a constant value `-1` or `0`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                        var operatorKind = blockOperation.OperatorKind;
+                        if (operatorKind != BinaryOperatorKind.Equals && operatorKind != BinaryOperatorKind.GreaterThanOrEqual)
+                        {
+                            return;
+                        }
+
+                        // Check that the right hand side is a -1(IUnary Operation) or 0(ILiteralOperation)
```
mavasani, line 117: You should instead check `operation.ConstantValue.HasValue && (`operation.ConstantValue.Value == -1 || `operation.ConstantValue.Value == 0)`. It doesn't really matter what operation the constant value corresponds to.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,231 @@
...
+                        var rightOperand = blockOperation.RightOperand;
+                        if ((rightOperand is IUnaryOperation unaryOperation || rightOperand is ILiteralOperation literalOperation) && rightOperand.ConstantValue.HasValue && rightOperand.ConstantValue.Value is int intValue && (intValue == -1 || intValue == 0))
```
mavasani, line 119: Adding so many nesting if statements makes the code harder to read. Instead invert the if condition and bail out when false to make the code more easier to read.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,810 @@
...
+            string csInput = @"
+namespace TestNamespace 
+{ 
+    class TestClass 
+    { 
+        private void TestMethod() 
+        { 
+            const string str = ""This is a string"";
+            int index = str.IndexOf(" + startQuote + input + endQuote + @");
+            if ([|index" + operatorKind + value + @"|])
```
mavasani, line 36: What if I had intermediate code that re-assigned either `str` or the argument passed to `IndexOf` call? As in,

```csharp
string str = "First string";
int index = str.IndexOf("First");
str = "Second string";
if (index != -1)
{
   // This code should execute
}
```

OR

```csharp
string str = "First string";
string subStringToFind = "First";
int index = str.IndexOf(subStringToFind);
subStringToFind = "Second";
if (index != -1)
{
   // This code should execute
}
```

I think your diagnostic/code fix will cause the user's code to break at runtime.

pgovind: Good catch. I'll fix these. I'm also addressing cases where `str` is a class field or a string literal.

mavasani:
Given the complexity of the analyzer here, lets design it by ignoring the code fix start with.

Lets aim at reporting analyzer diagnostics for following cases:
1. `IndexOf` call is parented by a binary operation that compares its result with constant value -1 or 0.
2. `IndexOf` call is parented by a local variable declaration + all local references for this local in the method block have a parenting binary operation, such that the binary operation compares the local value to a constant value -1 or 0.

I would recommend implementing the analyzer as follows:
1. Register OperationBlockStart callback
2. Register a nested callback on block start for `OperationKind.Invocation`:
   1. In the callback, make sure this is the `IndexOf` method call that you care about. Otherwise bail out.
   2. Check if invocation.Parent is a binary operation, whose binary operator is the one you care about and other side of compare is a constant value 0 or -1. If so, report a diagnostic and return from the callback.
   3. Otherwise, check if invocation.Parent is a variable declarator. If so, add the invocation as a potential violation to track, say in a dictionary `PooledConcurrentDictionary<ILocalSymbol, IInvocationOperation> potentialInvocationsToFlag`.
3. Register a nested callback on block start for `OperationKind.LocalReference`:
   1. In the callback check if localReference.Parent is a binary operation that compare local with 0 or -1. If so, add the local reference to a set, say `localsWithRequiredComparision`.
   2. Otherwise, add the local to a set, say `localsToBailOut`.
4. Register an operation block end nested callback on block start:
   1. For all locals in `localsWithRequiredComparision`, which are not present in `localsToBailOut`, try get value from `potentialInvocationsToFlag` and report a violation on the `IInvocationOperation` in this dictionary.

mavasani:
@pgovind Made slight modifications to my suggested approach to simplify further - feel free to ping me to discuss further.

pgovind:
Ok, I removed the fixer and modified the analyzer. This should be good to review now

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
pgovind: Good catch. Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                    PooledConcurrentDictionary<ISymbol, IInvocationOperation> variableNameToOperationsMap = PooledConcurrentDictionary<ISymbol, IInvocationOperation>.GetInstance();
+
+                    context.RegisterOperationAction(PopulateLocalReferencesSet, OperationKind.LocalReference);
+
+                    context.RegisterOperationAction(AnalyzeInvocationOperation, OperationKind.Invocation);
+
+                    context.RegisterOperationBlockEndAction(OnOperationBlockEnd);
+
+                    // Local Functions
+                    void PopulateLocalReferencesSet(OperationAnalysisContext context)
```
mavasani, line 109: Nit: explicit return for readability and separation of local functions

```suggestion
                    return;
                    
                    // Local Functions
                    void PopulateLocalReferencesSet(OperationAnalysisContext context)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                    charAndComparisonTypeArgumentContainsMethod == null ||
+                    charArgumentContainsMethod == null)
+                {
+                    return;
+                }
+
+                // Roslyn doesn't yet support "FindAllReferences" at a file/block level. So instead, find references to local int variables in this block.
+                context.RegisterOperationBlockStartAction(OnOperationBlockStart);
+
+                void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
```
mavasani, line 92: Nit:

```suggestion
                context.RegisterOperationBlockStartAction(OnOperationBlockStart);
                return;

                void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                    void PopulateLocalReferencesSet(OperationAnalysisContext context)
+                    {
+                        ILocalReferenceOperation localReference = (ILocalReferenceOperation)context.Operation;
+                        if (localReference.Local.Type.SpecialType != SpecialType.System_Int32)
+                        {
+                            return;
+                        }
+
+                        var parent = localReference.Parent;
+                        if (!(parent is IBinaryOperation binaryOperation))
```
mavasani, line 118: Do we want to also bail out when we have binary operations that indicate users do need to invoke IndexOf as opposed to Contains, say `index > 3`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                context.RegisterOperationBlockStartAction(OnOperationBlockStart);
+
+                void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
+                {
+                    if (!(context.OwningSymbol is IMethodSymbol method))
+                    {
+                        return;
+                    }
+
+                    PooledConcurrentSet<ISymbol> localsToBailOut = PooledConcurrentSet<ISymbol>.GetInstance();
```
mavasani, line 99:
```suggestion
                    PooledConcurrentSet<ILocalSymbol> localsToBailOut = PooledConcurrentSet<ILocalSymbol>.GetInstance();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                    PooledConcurrentDictionary<ISymbol, IInvocationOperation> variableNameToOperationsMap = PooledConcurrentDictionary<ISymbol, IInvocationOperation>.GetInstance();
```
mavasani, line 100:
```suggestion
                    PooledConcurrentDictionary<ILocalSymbol, IInvocationOperation> variableNameToOperationsMap = PooledConcurrentDictionary<ILocalSymbol, IInvocationOperation>.GetInstance();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                    void AnalyzeInvocationOperation(OperationAnalysisContext context)
+                    {
+                        var invocationOperation = (IInvocationOperation)context.Operation;
+                        if (!IsDesiredTargetMethod(invocationOperation.TargetMethod))
+                        {
+                            return;
+                        }
+
+                        var parent = invocationOperation.Parent;
+                        if (parent is IBinaryOperation blockOperation)
```
mavasani, line 133:
```suggestion
                        if (parent is IBinaryOperation binaryOperation)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                        {
+                            var operatorKind = blockOperation.OperatorKind;
+                            if (operatorKind != BinaryOperatorKind.Equals && operatorKind != BinaryOperatorKind.GreaterThanOrEqual)
+                            {
+                                return;
+                            }
+
+                            var rightOperand = blockOperation.RightOperand;
```
mavasani, line 141: I think you should handle both cases `index op constant` and `constant op index`, i.e.:

```csharp
var otherOperand = binaryOperation.LeftOperand is invocationOperation ? binaryOperation.RightOperand : binaryOperation.LeftOperand;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                                context.ReportDiagnostic(blockOperation.CreateDiagnostic(Rule));
+                            }
+                        }
+                        else if (parent is IVariableInitializerOperation variableInitializer)
+                        {
+                            if (variableInitializer.Parent is IVariableDeclaratorOperation variableDeclaratorOperation)
+                            {
+                                variableNameToOperationsMap.TryAdd(variableDeclaratorOperation.Symbol, invocationOperation);
+                            }
+                            else if (variableInitializer.Parent is IVariableDeclarationOperation variableDeclarationOperation)
```
mavasani, line 153: Defensive check?
```suggestion
                            else if (variableInitializer.Parent is IVariableDeclarationOperation variableDeclarationOperation &&
                                     variableDeclarationOperation.Declarators.Length == 1)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                            {
+                                variableNameToOperationsMap.TryAdd(variableDeclarationOperation.Declarators[0].Symbol, invocationOperation);
+                            }
+                        }
+                    }
+
+                    void OnOperationBlockEnd(OperationBlockAnalysisContext context)
+                    {
+                        foreach (KeyValuePair<ISymbol, IInvocationOperation> variableNameAndLocation in variableNameToOperationsMap)
```
mavasani, line 162:
```suggestion
                        foreach (var (local, invocation) in variableNameToOperationsMap)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
mavasani: Please also add a unit test for this example, as it will be a false positive to flag IndexOf calls when the index is used for a comparison with a positive integer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                context.RegisterOperationBlockStartAction(OnOperationBlockStart);
+
+                void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
+                {
+                    if (!(context.OwningSymbol is IMethodSymbol method))
+                    {
+                        return;
+                    }
+
+                    PooledConcurrentSet<ISymbol> localsToBailOut = PooledConcurrentSet<ISymbol>.GetInstance();
```
mavasani, line 99: I think it would be good to add an explanatory comment about the algorithm and data structures used here so it is easier to understand for someone else in future.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,707 @@
...
+{ 
+    class TestClass 
+    { 
+        private void TestMethod() 
+        { 
+            const string str = ""This is a string"";
+            int index = [|str.IndexOf(" + startQuote + input + endQuote + @")|];
+            if (index" + operatorKind + value + @")
+            {
+
```
mavasani, line 38:
```suggestion
```

mavasani: And similarly for other tests...

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,707 @@
...
+            const string str = ""This is a string"";
+            int index = str.IndexOf(" + quotes + input + quotes + sb.ToString() + @");
+            if (index == -1)
+            {
+
+            }
+        } 
+    } 
+}";
+            await VerifyCS.VerifyAnalyzerAsync(csInput);
```
mavasani, line 400: Prior tests all use `ReferenceAssemblies.NetCore.NetCoreApp50`. Why not here?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,707 @@
...
+                TestState = { Sources = { vbInput } },
+                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp50,
+            };
+            await testOrdinal_vb.RunAsync();
+        }
+
+        [Theory]
+        [InlineData("This", "This", false)]
+        [InlineData("a", "a", true)]
+        public async Task TestStringAndCharWithMultipleDiagnostics(string input, string fix, bool isCharTest)
```
mavasani, line 189: Consider adding test where local use is not immediately following the declaration, say:

```csharp
int index1 = [|str.IndexOf(" + startQuote + input + endQuote + @")|];
int index2 = [|str.IndexOf(" + startQuote + input + endQuote + @")|];
if (index1 == -1 || index2 == -1)
{
}
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,707 @@
...
+        } 
+    } 
+}";
+            await VerifyCS.VerifyAnalyzerAsync(csInput);
+        }
+
+        [Theory]
+        [InlineData("This", false)]
+        [InlineData("a", true)]
+        public async Task TestIndexWrittenTo(string input, bool isCharTest)
```
mavasani, line 406: Consider adding a test where write happens after read, i.e.

```csharp
const string str = ""This is a string"";
            int index = str.IndexOf(" + quotes + input + quotes + @");
            if (index == -1)
            {
            }
            index += 2;
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,707 @@
...
+        {
+            string csInput = @"
+namespace TestNamespace 
+{ 
+    class TestClass 
+    { 
+        private void TestMethod() 
+        { 
+            const string str = ""This is a string"";
+            int index = [|str.IndexOf('a', System.StringComparison.Ordinal)|], aa = 5;
```
mavasani, line 667: Also add a test where you have reverse order of declarators, i.e. `int aa = 5, index = [|str.IndexOf('a', System.StringComparison.Ordinal)|];`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
pgovind: We could. I don't see much benefit from it though. If anything, we'd end up using more memory in `localsToBailOut`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
pgovind: Yup yup, I meant to do this rename. Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
pgovind: Yup. I was going to ask you if `Declarators` would always have a length of 1 even if there were multiple declarators(it looks like it from my limited debugging), but adding the check is better.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
pgovind: Fixed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
pgovind: Changed the unit test here like you suggested

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
pgovind: Done. Also modified a unit test to have `(-1 == str.IndexOf)`

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -22,6 +22,7 @@ CA2013 | Reliability | Warning | DoNotUseReferenceEqualsWithValueTypesAnalyzer,
 CA2014 | Reliability | Warning | DoNotUseStackallocInLoopsAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2014)
 CA2015 | Reliability | Warning | DoNotDefineFinalizersForTypesDerivedFromMemoryManager, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2015)
 CA2247 | Usage | Warning | DoNotCreateTaskCompletionSourceWithWrongArguments, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2247)
+CA2248 | Usage | Info | PreferStringContainsOverIndexOf, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2247)
```
mavasani, line 25: Was this file manually changed? If so, please revert this and use the code fix. Also you may want to switch to using CA2249 given https://github.com/dotnet/roslyn-analyzers/pull/3557 and corresponding doc PR https://github.com/MicrosoftDocs/visualstudio-docs/pull/5247/files are already in progress.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
pgovind: Fixed now and added unit tests too

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
```diff
@@ -126,10 +126,21 @@ void PopulateLocalReferencesSet(OperationAnalysisContext context)
...
+                            var otherOperand = binaryOperation.LeftOperand is ILocalReferenceOperation ? binaryOperation.RightOperand : binaryOperation.LeftOperand;
+                            if (otherOperand.ConstantValue.HasValue && otherOperand.ConstantValue.Value is int intValue)
+                            {
+                                if ((operatorKind == BinaryOperatorKind.Equals && intValue < 0) ||
+                                    (operatorKind == BinaryOperatorKind.GreaterThanOrEqual && intValue == 0))
+                                {
+                                    // Do nothing. This is a valid case to the tagged in the analyzer
+                                    return;
+                                }
+                            }
```
mavasani, line 141: Can we add a static helper method for this code that can be reused even below?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfTests.cs
```diff
@@ -0,0 +1,838 @@
...
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.PreferStringContainsOverIndexOfAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.PreferStringContainsOverIndexOfAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+// The following pragmas be removed when the fixer is implemented. 
+#pragma warning disable IDE0060 // Remove unused parameter
```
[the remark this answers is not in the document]
mavasani: Sounds good!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStringContainsOverIndexOfAnalyzer.cs
pgovind: Yup.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1314,6 +1314,15 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferStringContainsOverIndexOfDescription" xml:space="preserve">
+    <value>Calls to 'String.IndexOf' where the result is compared to 0 or -1 can be replaced by 'String.Contains'</value>
+  </data>
+  <data name="PreferStringContainsOverIndexOfMessage" xml:space="preserve">
+    <value>Use 'String.Contains' instead of 'String.IndexOf' to improve readability</value>
```
mavasani, line 1321: Do we prefer using `String` instead of `string` in user facing message string? Just a question, will let you and the PM team decide.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1314,6 +1314,15 @@
   <data name="PreferTypedStringBuilderAppendOverloadsRemoveToString" xml:space="preserve">
     <value>Remove the ToString call</value>
   </data>
+  <data name="PreferStringContainsOverIndexOfDescription" xml:space="preserve">
+    <value>Calls to 'String.IndexOf' where the result is compared to 0 or -1 can be replaced by 'String.Contains'</value>
```
stephentoub, line 1318: > where the result is compared to 0

This isn't always true, e.g. an `== 0` or `> 0` check might not be replaceable.

AraHaan:
While I would like this rule, I would not like it if this rule is spawned when I use ``string.IndexOf()`` in ``string.Substring`` to create a substring in specific bounds of the string. I even use it in my WIP patch that when complete will render the exception documentation text in the IDE of VS.

However currently it is WIP because I have yet to find how the stuff colors the text under ``<see langword``, ``<see cref`` etc. If there was a method where I could pass in the entire ``<see ... />`` blocks in the xml docstrings and it would color it for me that would help make my WIP be complete. It would have to basically be a catch all though.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Yeah, the implementation is actually much more stricter then the wordings used here.

```csharp
                    static bool CheckOperatorKindAndOperand(IBinaryOperation binaryOperation, IOperation otherOperand)
                    {
                        var operatorKind = binaryOperation.OperatorKind;
                        if (otherOperand.ConstantValue.HasValue && otherOperand.ConstantValue.Value is int intValue)
                        {
                            if ((operatorKind == BinaryOperatorKind.Equals && intValue < 0) ||
                                (operatorKind == BinaryOperatorKind.GreaterThanOrEqual && intValue == 0))
                            {
                                // This is the only case we are targeting in this analyzer
                                return true;
                            }
                        }
                        return false;
                    }
```

The wording should be adjusted to be more generic like say `Calls to 'String.IndexOf' which are only used to check the presence or absence of a sub-string within a string can be replaced by 'String.Contains'`. Not sure if this is still the best wording, but we should indicate intent instead of implementation details in this string.

mavasani:
@AraHaan Please see https://github.com/dotnet/roslyn-analyzers/pull/3594/files#r421183989. The analyzer has a pretty strict implementation for the cases it fires. It should not fire if the computed index is used to compute a sub-string.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
pgovind: I'll go ahead and merge what I have now and change it in the fixer if we need to

PR #3618, opened 2020-05-08
Update the infrastructure for roslyn-analyzers

jmarolf:
This ie required to get this repo to publish to a channel


mavasani:
@jmarolf Who would be the best person to review this change? @tmat? I do not have much insight into any of the scripts/infrastructure being added here.

jmarolf:
yeah I think @tmat would be a good person to review this.

Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <AssetsDir>$(MSBuildThisFileDirectory)assets\</AssetsDir>
     <VSSDKTargetPlatformRegRootSuffix>RoslynDev</VSSDKTargetPlatformRegRootSuffix>
     <AutoGenerateAssemblyVersion>true</AutoGenerateAssemblyVersion>
+    <NoWarn>$(NoWarn);NU5119</NoWarn>
```
tmat, line 8: >NU5119 [](start = 22, length = 6)

Seems like there should be some property NoDefaultExcludes to avoid this warning  
[https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5119](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5119)

eng/Versions.props
```diff
@@ -3,24 +3,22 @@
...
     <CoverletVersion>2.7.0</CoverletVersion>
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
-
     <!-- Use the correct compiler version -->
     <UsingToolMicrosoftNetCompilers>true</UsingToolMicrosoftNetCompilers>
-    <MicrosoftNetCompilersToolsetVersion>3.6.0-3.final</MicrosoftNetCompilersToolsetVersion>
-
+    <!-- Dependencies from https://github.com/dotnet/roslyn -->
+    <MicrosoftNETCoreCompilersPackageVersion>3.7.0-2.20257.5</MicrosoftNETCoreCompilersPackageVersion>
+    <MicrosoftNetCompilersToolsetVersion>$(MicrosoftNETCoreCompilersPackageVersion)</MicrosoftNetCompilersToolsetVersion>
```
tmat, line 21: This version flows automatically from Arcade, so you shouldn't need to set it.
The latest compilers flow into Arcade and from there they flow into any repo using Arcade.

eng/common/SetupNugetSources.ps1
```diff
@@ -0,0 +1,143 @@
...
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
+. $PSScriptRoot\tools.ps1
+
+# Add source entry to PackageSources
+function AddPackageSource($sources, $SourceName, $SourceEndPoint, $creds, $Username, $Password) {
+    $packageSource = $sources.SelectSingleNode("add[@key='$SourceName']")
+    
+    if ($packageSource -eq $null)
+    {
```
Evangelink, line 40: [cosmetic][suggestion] You are mixing styles for the curly brace, it's be easier for maintenance to have only one. This way we don't have to wonder which style to use.

jmarolf: things in the `eng/common` folder are updated automatically from the [dotnet/arcade](https://github.com/dotnet/arcade) repo. These files should never be touched other than by automated tooling

PR #3622, opened 2020-05-12
Multitarget tests to netcoreapp3.1

sharwell:


src/Utilities/Compiler/Extensions/StringCompatExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+#if !NETCOREAPP
+
+namespace System
+{
+    internal static class StringCompatExtensions
+    {
+        public static bool Contains(this string str, string value, StringComparison comparisonType)
+        {
+            return str.IndexOf(value, comparisonType) >= 0;
```
Evangelink, line 11: I thought that wasn't good to use the `IndexOf` when we can do `Contains` (see https://github.com/dotnet/roslyn-analyzers/pull/3594).

Evangelink:
This looks so much simpler than what I tried...

src/Utilities/Compiler/Extensions/StringCompatExtensions.cs
sharwell: There is no `Contains` method that takes a `StringComparison`.

Evangelink: Oh yes of course! Sorry!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/MarkAllNonSerializableFieldsTests.cs
```diff
@@ -379,7 +379,10 @@ End Class
         [Fact]
         public async Task CA2235WithArrayType()
         {
-            await VerifyCS.VerifyAnalyzerAsync(@"
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.NetFramework.Net472.Default,
```
Evangelink, line 384: Does it mean that the test is not run under the netcoreapp3.1 context? If so wouldn't it be better to have a new property on the `AdditionalMetadataReferences` class with a value set conditionally to `Net472` or `NetCoreApp31`?

sharwell: This analyzer explicitly disables itself if the compilation does not target .NET Framework.

Evangelink: > 
> This analyzer explicitly disables itself if the compilation does not target .NET Framework.

Sorry I left the comment on the first occurrence but this was more of a generic comment for your PR. Does all analyzers for which you have explicitly set `ReferenceAssemblies = ReferenceAssemblies.NetFramework.Net472.Default,` do not work for netcore?

sharwell: Yes, that is correct. In most cases (numerically), the APIs in question do not exist in .NET Standard 2.0. In other cases, the analyzer is disabled for that context.

Evangelink: Thank you! I am/was surprised by the number of cases/analyzers not "working" for the netstandard part.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureCryptographicAlgorithmsTests.cs
```diff
@@ -1965,7 +1965,10 @@ End Namespace
         [Fact]
         public async Task CA5350RIPEMD160ManagedInMethodDeclaration()
         {
-            await VerifyCS.VerifyAnalyzerAsync(@"
+            await new VerifyCS.Test
```
mavasani, line 1968: @dotpaul for reviewing changes to security rule tests.

sharwell:
@dotpaul Can you review this?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureDeserializerJsonNetWithoutBinderTests.cs
```diff
@@ -674,32 +674,38 @@ T Method<T>(JsonReader jr)
         [Fact]
         public async Task TypeNameHandlingNoneBinderNonNull_JsonSerializer_Populate_NoDiagnostic()
         {
-            await VerifyCSharpWithJsonNetAsync(@"
+#if NETCOREAPP
+            var serializationNamespace = "";
+#else
+            var serializationNamespace = "using System.Runtime.Serialization;";
+#endif
```
dotpaul, line 681: What problems did you run into with System.Runtime.Serialization.SerializationBinder on .NET Core?

dotpaul:
Sorry, totally missed Manish tagging me earlier.

How are you running the .NET Core unit tests?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureDeserializerJsonNetWithoutBinderTests.cs
dotpaul: Okay, so I think Newtonsoft.Json 10.0.1 is weird. Should change AdditionalMetadataReferences.DefaultWithNewtonsoftJson to use version 12.0.3.

Then the test can be:
```csharp
        [Fact]
        public async Task TypeNameHandlingNoneBinderNonNull_JsonSerializer_Populate_NoDiagnostic()
        {
            await VerifyCSharpWithJsonNetAsync(@"
using System;
using Newtonsoft.Json;

class Blah
{
    private Func<System.Runtime.Serialization.SerializationBinder> SbGetter;

    object Method(JsonReader jr)
    {
        System.Runtime.Serialization.SerializationBinder sb = SbGetter();
        if (sb != null)
        {
            JsonSerializer serializer = new JsonSerializer();
            serializer.Binder = sb;
            object o = new object();
            serializer.Populate(jr, o);
            return o;
        }
        else
        {
            return null;
        }
    }
}");
        }
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureSettingsForJsonNetTests.cs
```diff
@@ -856,25 +857,31 @@ public class MyISerializationBinder : Newtonsoft.Json.Serialization.ISerializati
         [Fact]
         public async Task Secure_FieldInitialization_BinderSet_NoDiagnostic()
```
dotpaul, line 858: After changing AdditionalMetadataReferences.DefaultWithNewtonsoftJson to use version 12.0.3, the test can be:

```csharp
        [Fact]
        public async Task Secure_FieldInitialization_BinderSet_NoDiagnostic()
        {
            await VerifyCSharpWithJsonNetAsync(@"
using System;
using Newtonsoft.Json;

class Blah
{
    public static readonly JsonSerializerSettings Settings = 
        new JsonSerializerSettings()
        {
            TypeNameHandling = TypeNameHandling.Objects,
            Binder = new MyBinder(),
        };
}

public class MyBinder : System.Runtime.Serialization.SerializationBinder
{
    public override Type BindToType(string assemblyName, string typeName) => throw new NotImplementedException();
    public override void BindToName(Type serializedType, out string assemblyName, out string typeName) => throw new NotImplementedException();
}");
        }
```

dotpaul:
FYI @chloduan for XML tests (don't have to review)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureDeserializerJsonNetWithoutBinderTests.cs
sharwell: Should we still have a test for 10.x or is that no longer supported by the analyzers?

sharwell: The tests have been updated to test against both 10.0.1 and 12.0.1.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureSettingsForJsonNetTests.cs
sharwell: The tests have been updated to test against both 10.0.1 and 12.0.1.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DoNotUseInsecureDeserializerJsonNetWithoutBinderTests.cs
dotpaul: I'd just remove the 10.x reference. Not that it isn't supported; just has a funky interface in its .NET Standard library. Totally fine with keeping it too.

Issue #3642, opened 2020-05-19
UseValueTasksCorrectlyAnalyzer throws ArgumentNullException


DavidZidar:
#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

CA2012 UseValueTasksCorrectlyAnalyzer

#### Repro steps

```
using System;
using System.Threading.Tasks;

class C
{
    public void GuardThenAssignVariable()
    {
        if (false) throw new Exception();
        var vt = Helpers.ReturnsValueTask();
    }
}
```

#### Expected behavior
UseValueTasksCorrectly should not throw.

#### Actual behavior
An async method with a guard clause that throws an exception in combination with storing `ValueTask` in a variable causes UseValueTasksCorrectlyAnalyzer to throw `ArgumentNullException`.

It seems that `block.FallThroughSuccessor.Destination` is `null` when handling the fall through successor since it is a throw expression which results in the call to `seen.TryGetValue` throw
`ArgumentNullException` because the key is a null value.

stephentoub:
Thanks for reporting this.

DavidZidar:
No problem!

PR #3643, opened 2020-05-19
Fix handling of successor.Destination in UseValueTasksCorrectly

stephentoub:
I didn't realize Destination could be null: IntelliSense lied to me again ;-) ("'Destination' is not null here.)

Fixes https://github.com/dotnet/roslyn-analyzers/issues/3642


mavasani:
@sharwell we really need to fix the IDE wrongly indicating non-nullness for oblivious APIs

PR #3648, opened 2020-05-20
CA1043: Allow System.Index and System.Range for indexers

### LINKED ISSUE #3638: CA1043: Support System.Index and System.Range
#### Analyzer package

#### Package Version

v3.0.0

#### Diagnostic ID

[CA1043](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1043)

#### Repro steps

```csharp
class Test
{
    public object this[Index index] => null;
    public object[] this[Range range] => null;
}
```

#### Expected behavior
Allow to use `System.Index` and `System.Range` as indexer argument.

#### Actual behavior
Forces to use integral type or string.

Evangelink:
Fix #3638 for the original 2 points but there are still some open questions (they might deserve a separate thread).

The first commit of the PR is the actual fix, the second is a code refactoring to reduce nesting, improve performance by checking the `api_surface` later in the code and also by using `ImmutableHashSet` instead of `array` for the contains check. Please feel free to push a revert for the second commit if you prefer not to have it.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UseIntegralOrStringArgumentForIndexers.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UseIntegralOrStringArgumentForIndexers.cs
@@ -68,7 +68,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             });
         }
 
-        private void AnalyzeSymbol(SymbolAnalysisContext context, ImmutableHashSet<INamedTypeSymbol> allowedTypes)
+        private static void AnalyzeSymbol(SymbolAnalysisContext context, ImmutableHashSet<INamedTypeSymbol> allowedTypes)
         {
             var symbol = (IPropertySymbol)context.Symbol;
             if (!symbol.IsIndexer || symbol.IsOverride)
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UseIntegralOrStringArgumentForIndexers.cs
```diff
@@ -31,16 +31,17 @@ public sealed class UseIntegralOrStringArgumentForIndexersAnalyzer : DiagnosticA
...
-                        SpecialType.System_String,
-                        SpecialType.System_Int16,
-                        SpecialType.System_Int32,
-                        SpecialType.System_Int64,
-                        SpecialType.System_Object,
-                        SpecialType.System_UInt16,
-                        SpecialType.System_UInt32,
-                        SpecialType.System_UInt64
-                        };
+        private static readonly ImmutableHashSet<SpecialType> s_allowedSpecialTypes =
```
sharwell, line 34: 📝 It's a near certainty that this change went the opposite direction of what was intended. `ImmutableHashSet<T>` has poor locality and lookup performance (every element is stored in a different object in memory), while the one case where linear searches excel is small arrays of primitive types.

mavasani: Thanks @sharwell. @Evangelink would you be able to fix it up with a follow-up PR with a PERF comment on the field on the same lines?

Evangelink: @sharwell Thanks that's interesting! Is it a remark specific to the immutables or is it true also for regular collections? My point is if I have a small collection is it better to have array or `HashSet` to do a `Contains`?

Evangelink: See #3666

PR #3649, opened 2020-05-20
Remove unused well-known type names

Evangelink:
Remove unused well-known type names.


mavasani:
@Evangelink How did you identify these? Did you use some analyzer OR reviewed the CodeLens reference count on top of each field OR did a Find All References for each of them?

Evangelink:
I did a search and replace to transform the fields onto properties and then reviewed the CodeLens reference count. Out of curiosity is it possible to have CodeLens ref count for fields?

mavasani:
> Out of curiosity is it possible to have CodeLens ref count for fields?

I think there was some feature request on Roslyn for it, I can't find it now...

PR #3651, opened 2020-05-20
Update CA1814 to not report on interface member implementation

### LINKED ISSUE #3650: CA1814 should not report for interface implementation
#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

Example: [CA1814](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1814)

#### Expected behavior

No diagnostic.

#### Actual behavior

The rule reports for property and method declaration even when they are the implementation of an interface. In such case, we have to use the multi-dimensional array.

Evangelink:
Fix #3650

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
@@ -76,37 +76,39 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
         {
             var property = (IPropertySymbol)context.Symbol;
 
-            // Bail-out when override or interface implementation as it can only be fixed in the base type.
-            if (property.IsOverride || property.IsImplementationOfAnyInterfaceMember())
+            // Bail-out when override as it can only be fixed in the base type.
+            if (property.IsOverride)
             {
                 return;
             }
 
-            if (IsMultiDimensionalArray(property.Type))
+            if (IsMultiDimensionalArray(property.Type) &&
+                !property.IsImplementationOfAnyInterfaceMember())
             {
                 context.ReportDiagnostic(property.CreateDiagnostic(DefaultRule, property.Name));
             }
 
-            AnalyzeParameters(context, property.Parameters);
+            AnalyzeParameters(context, property, property.Parameters);
         }
 
         private static void AnalyzeMethod(SymbolAnalysisContext context)
         {
             var method = (IMethodSymbol)context.Symbol;
 
-            // Bail-out when override or interface implementation as it can only be fixed in the base type.
+            // Bail-out when override as it can only be fixed in the base type.
             // If its a getter\setter then we will report on the property instead so skip analyzing the method.
-            if (method.IsOverride || method.IsImplementationOfAnyInterfaceMember() || method.AssociatedSymbol != null)
+            if (method.IsOverride || method.AssociatedSymbol != null)
             {
                 return;
             }
 
-            if (IsMultiDimensionalArray(method.ReturnType))
+            if (IsMultiDimensionalArray(method.ReturnType) &&
+                !method.IsImplementationOfAnyInterfaceMember())
             {
                 context.ReportDiagnostic(method.CreateDiagnostic(ReturnRule, method.Name, method.ReturnType));
             }
 
-            AnalyzeParameters(context, method.Parameters);
+            AnalyzeParameters(context, method, method.Parameters);
         }
 
         private static void AnalyzeObjectCreation(OperationAnalysisContext context)
@@ -119,11 +121,12 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
             }
         }
 
-        private static void AnalyzeParameters(SymbolAnalysisContext context, ImmutableArray<IParameterSymbol> parameters)
+        private static void AnalyzeParameters(SymbolAnalysisContext context, ISymbol containingSymbol, ImmutableArray<IParameterSymbol> parameters)
         {
             foreach (IParameterSymbol parameter in parameters)
             {
-                if (IsMultiDimensionalArray(parameter.Type))
+                if (IsMultiDimensionalArray(parameter.Type) &&
+                    !containingSymbol.IsImplementationOfAnyInterfaceMember())
                 {
                     context.ReportDiagnostic(parameter.CreateDiagnostic(DefaultRule, parameter.Name));
                 }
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensionalTests.cs
```diff
@@ -213,6 +214,69 @@ End Class
...
+{
+    int[,] {|#0:MethodReturningMultidimensionalArray|}(int[,] {|#1:array|});
+}
+
+public class C : IC
+{
+    public int[,] MethodReturningMultidimensionalArray(int[,] array)
+        => null;
+}",
+                VerifyCS.Diagnostic(PreferJaggedArraysOverMultidimensionalAnalyzer.ReturnRule).WithLocation(0).WithArguments("MethodReturningMultidimensionalArray", "int[*,*]"),
```
Evangelink, line 231: @sharwell That's a real pleasure to use the combination markup syntax + diagnostic. It's a lot easier to read and review.

mavasani: Also tagging @carlossanlop who faced a similar issue with an analyzer. It is indeed possible to use markup for analyzer reporting multiple diagnostic descriptors with same ID, and this test is good example of it.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
```diff
@@ -76,33 +76,37 @@ private static void AnalyzeProperty(SymbolAnalysisContext context)
         {
             var property = (IPropertySymbol)context.Symbol;
 
-            // If its an override then don't report it as it can only be fixed in the base type.
-            if (!property.IsOverride)
+            // Bail-out when override or interface implementation as it can only be fixed in the base type.
+            if (property.IsOverride || property.IsImplementationOfAnyInterfaceMember())
```
mavasani, line 80: `IsImplementationOfAnyInterfaceMember` is an expensive check. Can we switch it to be done after `IsMultiDimensionalArray` check?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
```diff
@@ -76,33 +76,37 @@ private static void AnalyzeProperty(SymbolAnalysisContext context)
...
 
         private static void AnalyzeMethod(SymbolAnalysisContext context)
         {
             var method = (IMethodSymbol)context.Symbol;
 
-            // If its an override then don't report it as it can only be fixed in the base type.
+            // Bail-out when override or interface implementation as it can only be fixed in the base type.
             // If its a getter\setter then we will report on the property instead so skip analyzing the method.
-            if (!method.IsOverride && method.AssociatedSymbol == null)
+            if (method.IsOverride || method.IsImplementationOfAnyInterfaceMember() || method.AssociatedSymbol != null)
```
mavasani, line 99: Same here - please delay `IsImplementationOfAnyInterfaceMember` as much as possible.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
Evangelink: I am not sure the code will look nice because we want this check to be done whatever the result of `IsMultiDimensionalArray` is.

```csharp
if (property.IsOverride) return;

if (IsMutltiDim(property.Type) && !property.IsImpl()) report;

if (!property.IsImpl()) AnalyzeParams();
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
Evangelink: I could swap the last 2 parts of the if condition but regarding delaying more the check we fall-back to my previous comment.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
mavasani: You can refactor `AnalyzeParams` itself takes the property and has `if (IsMutltiDim(parameter.Type) && !containingProperty.IsImpl()) report;`

Evangelink: So we end-up doing twice the check which is expensive. I'll store the result in a variable and pass-it along.

Evangelink: I will push a commit with the change you are suggesting but it seems a lot slower as the expensive call is now done multiple times...

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/PreferJaggedArraysOverMultidimensional.cs
```diff
@@ -115,11 +121,12 @@ private static void AnalyzeObjectCreation(OperationAnalysisContext context)
...
         }
 
-        private static void AnalyzeParameters(SymbolAnalysisContext context, ImmutableArray<IParameterSymbol> parameters)
+        private static void AnalyzeParameters(SymbolAnalysisContext context, ISymbol containingSymbol, ImmutableArray<IParameterSymbol> parameters)
         {
             foreach (IParameterSymbol parameter in parameters)
             {
-                if (IsMultiDimensionalArray(parameter.Type))
+                if (IsMultiDimensionalArray(parameter.Type) &&
+                    !containingSymbol.IsImplementationOfAnyInterfaceMember())
```
Evangelink, line 129: This "expensive" call is now done for each parameters although the check could be done once. Besides the check was also done previously for the method or property on the calling method.

mavasani: I think that is fine - multidimensional arrays are rare so this is okay. Invoking this for every property/method regardless of the type would be more expensive.

PR #3652, opened 2020-05-20
Update CA2248 to report on call to 'HasFlag' on enums not marked with…

### LINKED ISSUE #3528: New analyzer: Report on call to 'HasFlag' on enums not marked with 'FlagsAttribute'

Evangelink:
… 'FlagsAttribute'

Fix #3528

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
@@ -15,40 +15,68 @@ namespace Microsoft.NetCore.Analyzers.Usage
         internal const string RuleId = "CA2248";
 
         private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.ProvideCorrectArgumentToEnumHasFlagTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.ProvideCorrectArgumentToEnumHasFlagMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessageDifferentType = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.ProvideCorrectArgumentToEnumHasFlagMessageDifferentType), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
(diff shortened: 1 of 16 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlagTests.cs
```diff
@@ -137,9 +140,10 @@ public enum OtherEnum { A, }
 
     public void Method(MyEnum m)
     {
-        [|m.HasFlag(OtherEnum.A)|];
+        {|#0:m.HasFlag(OtherEnum.A)|};
     }
-}");
+}",
+                VerifyCS.Diagnostic(ProvideCorrectArgumentToEnumHasFlag.DifferentTypeRule).WithLocation(0).WithArguments("OtherEnum", "MyEnum"));
```
Evangelink, line 146: I decided to report only one diagnostic here because when types are different we do not know which of the two types is the right one (it's most likely the one on which the method is called though) thus I wasn't sure on which type to base the check. Note that we could also go for 3 diagnostics in this case, the old one, one for the fact that `MyEnum` is not marked with `[Flags]` and one for the fact that `OtherEnum` is not marked with `[Flags]`. Let me know if you have any preference.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
```diff
@@ -15,40 +15,68 @@ public sealed class ProvideCorrectArgumentToEnumHasFlag : DiagnosticAnalyzer
...
+
+                    if (!invocation.Instance.Type.Equals(conversion.Operand.Type))
+                    {
+                        context.ReportDiagnostic(invocation.CreateDiagnostic(DifferentTypeRule, GetArgumentTypeName(conversion), invocation.Instance.Type.Name));
+                    }
+                    else if (flagsAttributeType != null && !invocation.Instance.Type.HasAttribute(flagsAttributeType))
+                    {
+                        context.ReportDiagnostic(invocation.CreateDiagnostic(NotFlagsRule, invocation.Instance.Type.Name));
+                    }
+                    else
```
mavasani, line 74: Can we remove this else block?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlagTests.cs
mavasani: Single diagnostic sounds reasonable.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
Evangelink: Sure! Old habit :)

mavasani:
@Evangelink Just checking if you have the doc PR for CA2248 in place, or plan to do so in near future. Please add this case as well.

Evangelink:
I'll update the [doc PR](https://github.com/MicrosoftDocs/visualstudio-docs/pull/5247/files) to include the change from this PR

PR #3653, opened 2020-05-20
CA2241: AutoDetectStringFormattingMethods

manfred-brands:
In the end this was very simple as the method to detect a Format method already existed.

We could completely delete the existing configuration and simply check if the called method has a format parameter followed by a params parameter. This can not be much more expensive than checking the configuration.

### DIFF
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -239,6 +239,16 @@ Examples:
 |`dotnet_code_quality.additional_string_formatting_methods = NS.MyType.MyFormat(ParamType)` | Matches specific method 'MyFormat' with given fully qualified signature
 |`dotnet_code_quality.additional_string_formatting_methods = NS1.MyType1.MyFormat1(ParamType)\|NS2.MyType2.MyFormat2(ParamType)` | Matches specific methods 'MyFormat1' and 'MyFormat2' with respective fully qualified signature
 
+Option Name: `try_determine_additional_string_formatting_methods_automatically`
+
+Configurable Rules: [CA2241](https://docs.microsoft.com/visualstudio/code-quality/ca2241)
+
+Option Values: Boolean values
+
+Default Value: `false`
+
+Example: `dotnet_code_quality.try_determine_additional_string_formatting_methods_automatically = true`
+
 ### Excluded symbol names
 
 Option Name: `excluded_symbol_names`
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -241,6 +241,9 @@ Examples:
 
 Option Name: `try_determine_additional_string_formatting_methods_automatically`
 
+Boolean option to enable heuristically detecting of additional string formatting methods
+A method is considered a string formatting method if it has a 'string format' parameter followed by a 'params object[]' parameter.
+
 Configurable Rules: [CA2241](https://docs.microsoft.com/visualstudio/code-quality/ca2241)
 
 Option Values: Boolean values
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
@@ -67,7 +67,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                     var stringFormat = (string)formatStringArgument.Value.ConstantValue.Value;
                     int expectedStringFormatArgumentCount = GetFormattingArguments(stringFormat);
 
-                    // explict parameter case
+                    // explicit parameter case
                     if (info.ExpectedStringFormatArgumentCount >= 0)
                     {
                         // __arglist is not supported here
@@ -351,6 +351,13 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 var additionalStringFormatMethodsOption = context.Options.GetAdditionalStringFormattingMethodsOption(Rule, context.Operation.Syntax.SyntaxTree, context.Compilation, context.CancellationToken);
                 if (additionalStringFormatMethodsOption.Contains(method.OriginalDefinition) &&
                     TryGetFormatInfo(method, out info))
+                {
+                    return info;
+                }
+
+                // Check if method could be a string formatting method.
+                // This is defined as a method with a 'format' argument followed by an 'params object[]'.
+                if (TryGetFormatInfo(method, out info))
                 {
                     return info;
                 }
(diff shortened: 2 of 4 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -355,6 +355,13 @@ public StringFormatInfo(Compilation compilation)
                     return info;
                 }
 
+                // Check if method could be a string formatting method.
+                // This is defined as a method with a 'format' argument followed by an 'params object[]'.
+                if (TryGetFormatInfo(method, out info))
```
mavasani, line 360: As far as I recall, this behavior was considered in the original implementation, but was not chosen due to following reasons:
1. Potential of being a breaking change for the rule.
2. Makes this a heuristic based rule. Current implementation of the rule is conservative to the known or user configured set of formatting methods. This makes it aggressive to fire on all methods which have similar signature.

Given that this rule is marked as `RuleLevel.BuildWarningCandidate`, I am reluctant to change the default behavior to make it so aggressive. At the minimum, we need an option to guard this new behavior so users can select the preferred level of aggressiveness.

@dotnet/roslyn-analysis for opinions.

manfred-brands:
I added a commit which makes this behaviour configurable.
In addition the format parameter is now verified to be a string.

mavasani:
Change looks good to me overall with the default behavior being unchanged. Please add relevant documentation. Also we probably want a better option name, but I cannot think of a good one. Any suggestions?

manfred-brands:
Renamed option to: try_determine_additional_string_formatting_methods_automatically. Should be clear.
Added documentation and addressed your other review comments.

docs/Analyzer Configuration.md
```diff
@@ -239,6 +239,16 @@ Examples:
 |`dotnet_code_quality.additional_string_formatting_methods = NS.MyType.MyFormat(ParamType)` | Matches specific method 'MyFormat' with given fully qualified signature
 |`dotnet_code_quality.additional_string_formatting_methods = NS1.MyType1.MyFormat1(ParamType)\|NS2.MyType2.MyFormat2(ParamType)` | Matches specific methods 'MyFormat1' and 'MyFormat2' with respective fully qualified signature
 
+Option Name: `try_determine_additional_string_formatting_methods_automatically`
```
mavasani, line 242: Can you please add a more detailed description here on what this option does?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -355,6 +355,17 @@ public StringFormatInfo(Compilation compilation)
                     return info;
                 }
 
+                // Check if the user configured automatic determination of formatting methods.
+                // If so, check if the method called has a 'string format' parameter followed by an params array.
+                bool defaultTryDetermineAdditionalStringFormattingMethodsAutomatically = false;
+                var canTryDetermineAdditionalStringFormattingMethodsAutomatically = context.Options.GetBoolOptionValue(EditorConfigOptionNames.TryDetermineAdditionalStringFormattingMethodsAutomatically,
+                        Rule, context.Operation.Syntax.SyntaxTree, context.Compilation, defaultTryDetermineAdditionalStringFormattingMethodsAutomatically, context.CancellationToken);
```
mavasani, line 362:
```suggestion
                var determineAdditionalStringFormattingMethodsAutomatically = context.Options.GetBoolOptionValue(EditorConfigOptionNames.TryDetermineAdditionalStringFormattingMethodsAutomatically,
                        Rule, context.Operation.Syntax.SyntaxTree, context.Compilation, defaultValue: false, context.CancellationToken);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -355,6 +355,17 @@ public StringFormatInfo(Compilation compilation)
...
+                if (canTryDetermineAdditionalStringFormattingMethodsAutomatically &&
+                    TryGetFormatInfo(method, out info) && info.ExpectedStringFormatArgumentCount == -1)
```
mavasani, line 364:
```suggestion
                if (determineAdditionalStringFormattingMethodsAutomatically  &&
                    TryGetFormatInfo(method, out info) &&
                    info.ExpectedStringFormatArgumentCount == -1)
```

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -67,6 +67,13 @@ internal static partial class EditorConfigOptionNames
         /// </summary>
         public const string AdditionalStringFormattingMethods = "additional_string_formatting_methods";
 
+        /// <summary>
+        /// String option to enable heuristically detecting of additional string formatting methods for CA2241 (https://docs.microsoft.com/visualstudio/code-quality/ca2241).
```
mavasani, line 71: This exact comment can be added to the `Analyzer Configuration.md` doc section.

manfred-brands: I noticed that the comment is wrong, it should read 'Boolean option'.

docs/Analyzer Configuration.md
manfred-brands: Done.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
manfred-brands: Done.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
manfred-brands: Done.

manfred-brands:
Changes made and pushed. Would you like for me to collapse these commits into one or will that confuse the automatic system?

manfred-brands:
Thanks. Raised pull request for documentation: https://github.com/MicrosoftDocs/visualstudio-docs/pull/5391

PR #3654, opened 2020-05-20
Update CA2213 to consider CloseAsync as disposable method

### LINKED ISSUE #3601: CA2213 does not recognize an IAsyncDisposable implementaton as sufficient.
#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

Example: [CA2213 ](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2213?view=vs-2019)

#### Repro steps

    internal class DbConnectionService : IDbConnectionService, IAsyncDisposable
    {
        private DbConnection? _dbConnection;
        // ...
        public async ValueTask DisposeAsync()
        {
            if (_dbConnection != null)
            {
                await _dbConnection.CloseAsync().ConfigureAwait(false);
                _dbConnection = null;
            }

            // ...
        }
    }

#### Expected behavior

No CA2213 reported

#### Actual behavior

CA2213

Evangelink:
Fix #3601


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposableFieldsShouldBeDisposedTests.cs
```diff
@@ -3205,5 +3205,75 @@ End Sub
...
+    Implements IAsyncDisposable
+
+    Private a As A
+
+    Public Async Function DisposeAsync() As ValueTask Implements IAsyncDisposable.DisposeAsync
+        If a IsNot Nothing Then
+            Await a.CloseAsync().ConfigureAwait(False)
+            a = Nothing
+        End If
+    End Function
```
Evangelink, line 3274: I can't find out how to have the following bit of code compiling for vbnet. Compiler is complaining that `DisposeAsync` can't be marked with `async` because it's not `Sub` or doesn't return `Task/Task(Of T)`.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -229,6 +229,9 @@ private static bool HasDisposeCloseMethodSignature(this IMethodSymbol method)
                 method.ReturnsVoid && method.Parameters.IsEmpty;
         }
 
+        private static bool HasDisposeCloseAsyncMethodSignature(this IMethodSymbol method, INamedTypeSymbol? taskType)
```
Evangelink, line 232: Mimic the logic used for other methods.

mavasani: Can CloseAsync return a ValueTask like DisposeAsync below?

mavasani: Also can you please add doc comment for consistency with other similar methods in this file?

Evangelink: I don't know if it can be of type `ValueTask`, the few examples I have seen are using `Task`. We can either keep only `Task` for now and revisit later or play on the side of the potential false-negative and handle `ValueTask` now.

PR #3677, opened 2020-05-25
Fix PreferStreamAsyncMemoryOverloadsMessage

### LINKED ISSUE #3663: CA1835 warning text doesn't make sense
image

The text says:
"Change the 'WriteAsync' method call to the overload that receives a 'System.IO.Stream.WriteAsync(System.ReadOnlyMemory<byte>, System.Threading.CancellationToken)'."
which doesn't make sense (the method doesn't receive a method :smile:).

cc: @carlossanlop

Youssef1313:
Fixes #3663


mavasani:
@carlossanlop

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1285,7 +1285,7 @@
     <value>'Stream' has a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as the first argument, and a 'WriteAsync' overload that takes a 'ReadOnlyMemory&lt;Byte&gt;' as the first argument. Prefer calling the memory based overloads, which are more efficient.</value>
   </data>
   <data name="PreferStreamAsyncMemoryOverloadsMessage" xml:space="preserve">
-    <value>Change the '{0}' method call to the overload that receives a '{1}'.</value>
+    <value>Change the '{0}' method call to the overload '{1}'.</value>
```
Evangelink, line 1288: Not relate to your change but it seems like a good opportunity to fix this. Message should not end with a period.

Evangelink:
It seems that you didn't fully rebuild the solution as some translation files do not appear as changed.

Youssef1313:
> 
> 
> It seems that you didn't fully rebuild the solution as some translation files do not appear as changed.

Thanks for letting me know that :)
I think it should work now.
I really appreciate your help and also @mavasani's help on my very first contributions here :)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
Youssef1313: @Evangelink I've removed the period.

Evangelink:
Might be worth tagging @stephentoub too.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1285,7 +1285,7 @@
     <value>'Stream' has a 'ReadAsync' overload that takes a 'Memory&lt;Byte&gt;' as the first argument, and a 'WriteAsync' overload that takes a 'ReadOnlyMemory&lt;Byte&gt;' as the first argument. Prefer calling the memory based overloads, which are more efficient.</value>
   </data>
   <data name="PreferStreamAsyncMemoryOverloadsMessage" xml:space="preserve">
-    <value>Change the '{0}' method call to the overload that receives a '{1}'.</value>
+    <value>Change the '{0}' method call to the overload '{1}'</value>
```
carlossanlop, line 1288: @Youssef1313 the original message had the intention to let the user know that the `Stream` method currently being analyzed has an overload that takes a memory parameter, and we are suggesting the user to use that instead.

So if the original message was: `Change the '{0}' method call to the overload that receives a '{1}'.`, then the possibilities after formatting the string were:

- `Change the 'ReadAsync' method call to the overload that receives a 'Memory<byte>'.`
- `Change the 'WriteAsync' method call to the overload that receives a 'ReadOnlyMemory<byte>'.`

Your change makes the formatted message to be confusing. It will read:

- `Change the 'ReadAsync' method call to the overload 'Memory<byte>'.`
- `Change the 'WriteAsync' method call to the overload 'ReadOnlyMemory<byte>'.`

Can you please revert this change?

stephentoub: @carlossanlop, did you see the issue this is fixing?

carlossanlop: Oh I see. Shouldn't we instead fix the second argument and make sure it's the name of the memory class being used by the overload?

carlossanlop: With the original string, the message would be much shorter.

Youssef1313:
@carlossanlop Thanks for review. I updated the branch. Will the message now look as expected?

Youssef1313:
Current build failures are due to unit tests. I'll fix that after @stephentoub or @carlossanlop give approval that the message looks correct.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
```diff
@@ -230,15 +230,15 @@ private void AnalyzeCompilationStart(CompilationStartAnalysisContext context)
                     {
                         rule = PreferStreamReadAsyncMemoryOverloadsRule;
                         ruleMessageMethod = undesiredReadAsyncMethod.Name;
-                        ruleMessagePreferredMethod = preferredReadAsyncMethod.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
+                        ruleMessagePreferredMethod = string.Join(", ", preferredReadAsyncMethod.GetParameters());
```
stephentoub, line 233: What does the rendered result message look like now?

Youssef1313: Before:

> Change the 'ReadAsync' method call to the overload that receives a 'System.IO.Stream.ReadAsync(System.Memory<byte>, System.Threading.CancellationToken)'.

Now:

> Change the 'ReadAsync' method call to the overload that receives a 'System.Memory<byte>, System.Threading.CancellationToken'.

stephentoub: Personally, I think it's strange to show the arguments listed like this without the context of the method; that's a bit unusual.  I'd have actually preferred it just updating the message text:
```
Change the 'ReadAsync' method call to use the 'ReadAsync(System.Memory, System.Threading.CancellationToken)' overload.
```
(ideally without the type name on the method).

Youssef1313: @stephentoub I agree that the message text itself needs to be updated, and to remove the type name on the method. I'll do that soon.

carlossanlop: Thank you @Youssef1313 for taking care of this. I replied in the other [outdated comment](https://github.com/dotnet/roslyn-analyzers/pull/3677#discussion_r432248458), I understand what the reported problem is, and thanks to both for pointing out my mistake and correcting me.
Stephen described accurately above what the original message was intending: offering the user the signature of the alternative method to use, including the parameter types, and I agree that that's what we should show.

Youssef1313: @stephentoub

I couldn't get it exactly to be:

> Change the 'ReadAsync' method call to use the 'ReadAsync(System.Memory, System.Threading.CancellationToken)' overload.

Would it be okay if it is:

"Stream.ReadAsync(Memory<byte>, CancellationToken)" instead of
"ReadAsync(System.Memory, System.Threading.CancellationToken)"

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
```diff
@@ -230,15 +230,15 @@ private void AnalyzeCompilationStart(CompilationStartAnalysisContext context)
...
                         ruleMessageMemoryType = memoryType.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
```
Youssef1313, line 234: @carlossanlop Do you know if this variable is redundant?

It's used in the following line:

`context.ReportDiagnostic(invocation.CreateDiagnostic(rule, ruleMessageMethod, ruleMessagePreferredMethod, ruleMessageMemoryType));`

As far as I can tell, the `ruleMessageMethod`, `ruleMessagePreferredMethod`, and `ruleMessageMemoryType` are to be injected (using string.Format or whatever similar) into the rule message. However, the rule message only has {0} and {1}, so it seems that `ruleMessageMemoryType` is redundant.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
carlossanlop: I think that's be good enough. The user would not lose context if we offer that message, because they can explore the additional overloads and see the full name of those parameter types.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.cs
carlossanlop: No, I think the redundant argument is the variable `ruleMessageMemoryType`, because it is supposed to represent either `Memory` or `ReadOnlyMemory`, but with the message format does not use that. We only use the method name `{0}` and the preferred method name `{1}` that includes the parameters:

> Change the '**ReadAsync**' method call to use the '**ReadAsync(Memory, CancellationToken)**' overload.

Youssef1313: > No, I think the redundant argument is the variable `ruleMessageMemoryType`

That's exactly what I meant. I've removed that variable.

mavasani:
@carlossanlop good to merge?

PR #3679, opened 2020-05-26
CA1816 shouldn't warn for IAsyncDisposable.DisposeAsync

### LINKED ISSUE #3675: CA1816 shouldn't warn for IAsyncDisposable.DisposeAsync
#### Analyzer package

[Microsoft.NetCore.Analyzers](https://www.nuget.org/packages/Microsoft.NetCore.Analyzers/)

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

[CA1816](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1816)

#### Repro steps

Create a class which implements [`IAsyncDisposable`](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) and calls [`GC.SuppressFinalize`](https://docs.microsoft.com/en-us/dotnet/api/system.gc.suppressfinalize) in [`DisposeAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable.disposeasync).  For example:

```csharp
namespace CA1816Example
{
    using System;
    using System.Threading.Tasks;

    class MyAsyncDisposable : IAsyncDisposable
    {
        [DllImport("example.dll")]
        private static extern int GetHandle();

        [DllImport("example.dll")]
        private static extern void FreeHandle(int handle);

        private readonly int handle;

        public MyAsyncDisposable()
        {
            this.handle = GetHandle();
        }

        ~MyAsyncDisposable()
        {
            FreeHandle(this.handle);
        }

        public async ValueTask DisposeAsync()
        {
            await Task.Run(() => FreeHandle(this.handle)).ConfigureAwait(false);
            GC.SuppressFinalize(this);
        }
    }
}
```

#### Expected behavior

CA1816 would not be produced.

#### Actual behavior

    MyAsyncDisposable.cs(30,13): warning CA1816: MyAsyncDisposable.DisposeAsync() calls GC.SuppressFinalize(object), a method that is typically only called within an implementation of 'IDisposable.Dispose'. Refer to the IDisposable pattern for more information.

#### Discussion

Although I haven't seen official guidance for [dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern) with `IAsyncDisposable`, it seems likely that classes which define a finalizer (e.g. to release native resources) would want to suppress finalization if the work has been completed by `DisposeAsync`, as in the above example, and therefore CA1816 shouldn't apply to use in this method.  Thoughts?

Youssef1313:
Fixes #3675

I'm not familiar with the code base here, so I'm not sure if that's a correct fix.

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -181,6 +181,7 @@ class MyAsyncDisposable : IAsyncDisposable
         public async Task AsyncDisposableWithFinalizer_Basic_NoDiagnostic()
         {
             var code = @"
+Imports System
 Imports System.Runtime.InteropServices
 Imports System.Threading.Tasks
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -178,6 +178,44 @@ class MyAsyncDisposable : IAsyncDisposable
         }
 
         [Fact]
+        public async Task AsyncDisposableWithFinalizer_Basic_NoDiagnostic()
+        {
+            var code = @"
+Imports System.Runtime.InteropServices
+Imports System.Threading.Tasks
+
+Class MyAsyncDisposable
+    Inherits IAsyncDisposable
+
+    <DllImport(""example.dll"")>
+    Private Shared Function GetHandle() As Integer
+
+    <DllImport(""example.dll"")>
+    Private Shared Sub FreeHandle(handle As Integer)
+
+    Private ReadOnly handle As Integer
+
+    Public Sub New()
+        Me.handle = GetHandle()
+    End Sub
+
+    Protected Overrides Sub Finalize()
+        FreeHandle(Me.handle)
+    End Sub
+
+    Public Async Function DisposeAsync() As ValueTask
+        Await Task.Run(Sub() FreeHandle(Me.handle)).ConfigureAwait(False)
+        GC.SuppressFinalize(Me)
+    End Function
+End Class";
+            await new VerifyVB.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
+                TestCode = code
+            }.RunAsync();
+        }
+        
+        [Fact]
         public async Task SealedDisposableWithoutFinalizer_CSharp_NoDiagnostic()
         {
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -3,6 +3,7 @@
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Testing;
+using Test.Utilities;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
     Microsoft.NetCore.Analyzers.Runtime.CallGCSuppressFinalizeCorrectlyAnalyzer,
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -169,7 +169,11 @@ class MyAsyncDisposable : IAsyncDisposable
         GC.SuppressFinalize(this);
     }
 }";
-            await VerifyCS.VerifyAnalyzerAsync(code);
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
+                TestCode = code
+            }.RunAsync();
         }
 
         [Fact]
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -136,6 +136,43 @@ End Class";
         }
 
         [Fact]
+        public async Task AsyncDisposableWithFinalizer_CSharp_NoDiagnostic()
+        {
+            var code = @"
+using System;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+class MyAsyncDisposable : IAsyncDisposable
+{
+    [DllImport(""example.dll"")]
+    private static extern int GetHandle();
+
+    [DllImport(""example.dll"")]
+    private static extern void FreeHandle(int handle);
+
+    private readonly int handle;
+
+    public MyAsyncDisposable()
+    {
+        this.handle = GetHandle();
+    }
+
+    ~MyAsyncDisposable()
+    {
+        FreeHandle(this.handle);
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        await Task.Run(() => FreeHandle(this.handle)).ConfigureAwait(false);
+        GC.SuppressFinalize(this);
+    }
+}";
+            await VerifyCS.VerifyAnalyzerAsync(code);
+        }
+
+        [Fact]
         public async Task SealedDisposableWithoutFinalizer_CSharp_NoDiagnostic()
         {
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -185,13 +185,15 @@ Imports System.Runtime.InteropServices
 Imports System.Threading.Tasks
 
 Class MyAsyncDisposable
-    Inherits IAsyncDisposable
+    Implements IAsyncDisposable
 
     <DllImport(""example.dll"")>
     Private Shared Function GetHandle() As Integer
+    End Function
 
     <DllImport(""example.dll"")>
     Private Shared Sub FreeHandle(handle As Integer)
+    End Sub
 
     Private ReadOnly handle As Integer
 
@@ -214,7 +216,7 @@ End Class";
                 TestCode = code
             }.RunAsync();
         }
-        
+
         [Fact]
         public async Task SealedDisposableWithoutFinalizer_CSharp_NoDiagnostic()
         {
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
@@ -178,47 +178,6 @@ class MyAsyncDisposable : IAsyncDisposable
         }
 
         [Fact]
-        public async Task AsyncDisposableWithFinalizer_Basic_NoDiagnostic()
-        {
-            var code = @"
-Imports System
-Imports System.Runtime.InteropServices
-Imports System.Threading.Tasks
-
-Class MyAsyncDisposable
-    Implements IAsyncDisposable
-
-    <DllImport(""example.dll"")>
-    Private Shared Function GetHandle() As Integer
-    End Function
-
-    <DllImport(""example.dll"")>
-    Private Shared Sub FreeHandle(handle As Integer)
-    End Sub
-
-    Private ReadOnly handle As Integer
-
-    Public Sub New()
-        Me.handle = GetHandle()
-    End Sub
-
-    Protected Overrides Sub Finalize()
-        FreeHandle(Me.handle)
-    End Sub
-
-    Public Async Function DisposeAsync() As ValueTask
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
```diff
@@ -135,6 +135,43 @@ End Sub
...
+    {
+        this.handle = GetHandle();
+    }
+
+    ~MyAsyncDisposable()
+    {
+        FreeHandle(this.handle);
+    }
+
+    public async ValueTask DisposeAsync()
```
Evangelink, line 166: @bartonjs Sorry to bother you but is there any guideline as the one for `Dispose` with the `virtual void Dispose(bool disposing)` pattern for the async version?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -159,7 +159,8 @@ public static bool IsImplementationOfInterfaceMethod(this IMethodSymbol method,
         public static bool IsDisposeImplementation(this IMethodSymbol method, Compilation compilation)
```
Evangelink, line 159: This helper method is used in a lot of places and I am not sure if this is the right idea to update it. For example, this means that the dispose pattern implementation will now raise issues when you implement `DisposeAsync` which is probably not what we want. IMO we should either create a new helper, use [GetDisposeMethodKind](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs#L280) or call [method.ContainingType.IsDisposable(iDisposable, iAsyncDisposable)](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs#L287) @mavasani WDYT?

Evangelink:
Hi @Youssef1313, 

Thank you for your PR and your effort! I have left some comments but please disregard them for the time being. @mavasani Will do the official review :)

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Youssef1313: Creating a new helper (`IsAsyncDisposeImplementation`) sounds good. But let me wait for @mavasani

Youssef1313:
The test fails as it can't find IAsyncDisposable interface. It should be available starting with .NET Core 3.0.

Under what .NET version does these tests run?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
bartonjs:
```C#
public async ValueTask DisposeAsync()
{
    await DisposeAsyncCore().ConfigureAwait(false);
    Dispose(false);
    GC.SuppressFinalize(this);
}

protected virtual ValueTask DisposeAsyncCore()
{
    // Async version of code inside `if (disposing)` blocks from Dispose(bool)
}
```

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
mavasani: Thanks, yes a new helper sounds good. Can you also audit the existing callers of `IsDisposeImplementation` to verify they are handling IAsyncDisposable where appropriate?

Youssef1313: I only found [this](https://github.com/dotnet/roslyn-analyzers/blob/75d6b9c80862070fc1b5efd0b41bd750152e3c0f/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotCallDangerousMethodsInDeserialization.cs#L214) that calls `IsDisposeImplementation`, but I don't know if it should be handling AsyncDispose or not.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
```diff
@@ -135,6 +135,47 @@ End Sub
...
+    {
+        await Task.Run(() => FreeHandle(this.handle)).ConfigureAwait(false);
+        GC.SuppressFinalize(this);
+    }
+}";
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
```
Youssef1313, line 174: Build fails on this line. @mavasani Can you help clarifying the right way to write this test?

I got this line from another test, not sure why it works there but not here.

Evangelink: You are missing the `using Test.Utilities;` and that should be ok.

Youssef1313: @Evangelink Thanks!
I've added it and waiting for the build to finish.

Youssef1313: @Evangelink, That `using` directive fixed the compilation error. But I can't get the test to pass.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -184,6 +193,27 @@ public static bool IsDisposeImplementation([NotNullWhen(returnValue: true)] this
...
+
+            if (method.IsOverride)
+            {
+                return method.OverriddenMethod.IsAsyncDisposeImplementation(iAsyncDisposable);
+            }
+
+            // Identify the implementor of IAsyncDisposable.Dispose in the given method's containing type and check
+            // if it is the given method.
+            return method.Parameters.Length == 0 &&
+                method.IsImplementationOfInterfaceMethod(null, iAsyncDisposable, "DisposeAsync");
```
Youssef1313, line 214: Is checking the method return type necessary here?
If so, please tell me how to write it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
Evangelink: @mavasani It might be worth having a separate issue and update the comply to dispose pattern to support this new pattern. WDYT?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Evangelink: I would say that's probably better to check the return type. You will need to do the following `method.ReturnType.Equals(valueTaskType)` and to get `valueTaskType` you will need to either add a parameter of type `INamedTypeSymbol` to this method declaration to delegate getting the type to the caller or you could mimic what's done on `IsAsyncDisposeImplementation` (i.e. `INamedTypeSymbol? valueTaskType= compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemValueTask);`).

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -184,6 +194,27 @@ public static bool IsDisposeImplementation([NotNullWhen(returnValue: true)] this
...
+        public static bool IsAsyncDisposeImplementation([NotNullWhen(returnValue: true)] this IMethodSymbol? method, [NotNullWhen(returnValue: true)] INamedTypeSymbol? iAsyncDisposable, INamedTypeSymbol? valueTaskType)
+        {
+            if (method == null)
+            {
+                return false;
+            }
+
+            if (method.IsOverride)
+            {
+                return method.OverriddenMethod.IsAsyncDisposeImplementation(iAsyncDisposable, valueTaskType);
```
mavasani, line 209: Overridden method may still be null for error cases.

mavasani: Ah, you do have a null check, ignore this suggestion.

Youssef1313:
@mavasani @Evangelink Is there anything left for this PR?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -184,6 +194,27 @@ public static bool IsDisposeImplementation([NotNullWhen(returnValue: true)] this
...
+        /// Checks if the given method implements "IAsyncDisposable.Dispose" or overrides an implementation of "IAsyncDisposable.Dispose".
+        /// </summary>
+        public static bool IsAsyncDisposeImplementation([NotNullWhen(returnValue: true)] this IMethodSymbol? method, [NotNullWhen(returnValue: true)] INamedTypeSymbol? iAsyncDisposable, INamedTypeSymbol? valueTaskType)
+        {
+            if (method == null)
+            {
+                return false;
+            }
+
+            if (method.IsOverride)
```
Evangelink, line 207: I see that this is a duplicate of what's in place for `Dispose` but I am not sure to understand what's the point of doing the check on the overriden method. AFAIK the override method has the same signature, isn't it?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -184,6 +194,27 @@ public static bool IsDisposeImplementation([NotNullWhen(returnValue: true)] this
...
+            {
+                return method.OverriddenMethod.IsAsyncDisposeImplementation(iAsyncDisposable, valueTaskType);
+            }
+
+            // Identify the implementor of IAsyncDisposable.Dispose in the given method's containing type and check
+            // if it is the given method.
+            return method.ReturnType.Equals(valueTaskType) && method.Parameters.Length == 0 &&
```
Evangelink, line 214: Are we sure that `ReturnType` is never null?

```suggestion
            return method.ReturnType != null && method.ReturnType.Equals(valueTaskType) && method.Parameters.Length == 0 &&
```

If you apply the change you can also add the `[NotNullWhen(returnValue: true)]` on the `valueTaskType` parameter.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
```diff
@@ -135,6 +136,47 @@ End Sub
             await VerifyVB.VerifyAnalyzerAsync(code);
         }
 
+        [Fact]
+        public async Task AsyncDisposableWithFinalizer_CSharp_NoDiagnostic()
```
Evangelink, line 140: What about doing the same test for vbnet?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Youssef1313: @Evangelink I'm not sure about that. I just duplicated the logic that already exists. @mavasani do you know?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Youssef1313: I can't think of any case that can make ReturnType to be null. Do you have any case in mind?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
Youssef1313: Sure.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Evangelink: Some error, also maybe a void type.

Youssef1313: @Evangelink if that's the case, it would be a problem with the existing implementation too right?

Evangelink: Yeah they are checking for `ReturnsVoid` :)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
Youssef1313: @Evangelink, ValueTask doesn't seem to be supported in vb.net


https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=netcore-3.1#remarks

> The use of the ValueTask type is supported starting with C# 7.0 and is not supported by any version of Visual Basic.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -184,6 +194,27 @@ public static bool IsDisposeImplementation([NotNullWhen(returnValue: true)] this
                 method.IsImplementationOfInterfaceMethod(null, iDisposable, "Dispose");
```
Youssef1313, line 194:
```
            return method.ReturnsVoid &&
                method.Parameters.Length == 0 &&
                method.IsImplementationOfInterfaceMethod(null, iDisposable, "Dispose");
```

@Evangelink Why there is no warning reported on the use of `.Length == 0` here?
I think it should report to use .IsEmpty. There is an existing analyzer for that. Analyzer bug? or I'm missing something?

Evangelink: Not sure why there is no diagnostic, I would also expect one, I will see if we have the same behavior in some test.

> Beside that point, is there anything left for the PR?

Let me do a new pass on the PR.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/CallGCSuppressFinalizeCorrectlyTests.cs
Evangelink: See #3909

Evangelink:
Ping @mavasani

Evangelink:
@mavasani Could you please review this PR when you have some time?

mavasani (approved):

mavasani:
Triggering a re-run for tests..

PR #3683, opened 2020-05-27
Delete MethodKindEx.cs

### LINKED ISSUE #3682: Is MethodKindEx class redundant?
I don't get the purpose of the [MethodKindEx](https://github.com/dotnet/roslyn-analyzers/blob/afb82254fa022b4624c794469bf45acd17dec3fd/src/Utilities/Compiler/Extensions/MethodKindEx.cs) class.

It has the following constant:

```csharp
public const MethodKind LocalFunction = (MethodKind)17;
```

Which will ALWAYS have the same value as `MethodKind.LocalFunction`, so why it exists? why not just use `MethodKind.LocalFunction` and simply delete `MethodKindEx`?

Youssef1313:
I'm not sure if what I'm saying in #3682 is correct. So, I'm opening this PR to check if it will build successfully and passes the tests. (Unfortunately, I can't get the build inside Visual Studio works locally)

Fixes #3682


src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -547,7 +547,7 @@ public static bool IsLambdaOrLocalFunction(this IMethodSymbol method)
             switch (method.MethodKind)
             {
                 case MethodKind.LambdaMethod:
-                case MethodKindEx.LocalFunction:
+                case MethodKind.LocalFunction:
```
Youssef1313, line 550: It's a question more than a contribution, but can someone explain why LocalFunction isn't accessible from MethodKind here?
Based on the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.methodkind?view=roslyn-dotnet), `MethodKind.LocalFunction` exists. and it was accessible in MethodKindEx, so, why not accessible here?

sharwell: In one of the projects that uses this file, the project intentionally compiles against an old version of Roslyn for maximum compatibility. `MethodKindEx` allows `LocalFunction` to be used if the _runtime_ environment is new enough, without eliminating the ability of older clients to retain full functionality of the analyzer.

Youssef1313: Thanks!

PR #3691, opened 2020-05-29
Port IRefactoringHelpers from dotnet/roslyn

sharwell:


Evangelink:
Would it be possible to expose the code from roslyn through some NuGet package? Or moving it to the sdk or a different common project? Alternatively, we could use [Paket](https://fsprojects.github.io/Paket/index.html) and use this [clone file from GitHub](https://fsprojects.github.io/Paket/github-dependencies.html) feature. I could help and knowing its community I am sure we could have plenty of help if needed.

mavasani:
@sharwell Can this PR now be closed?

PR #3692, opened 2020-05-29
Test accessor refactoring

sharwell:
Converts two refactorings from a hidden diagnostic to proper `CodeRefactoringProvider`.

The `IRefactoringHelpersService` from dotnet/roslyn is ported to Roslyn.Diagnostics.Analyzers, but not made available as part of any more general packages. Consideration for making this a more widely available service is an open discussion in #3691.


src/Roslyn.Diagnostics.Analyzers/Core/AbstractCreateTestAccessor`1.cs
```diff
@@ -1,70 +1,82 @@
...
+                CodeAction.Create(
+                    RoslynDiagnosticsAnalyzersResources.CreateTestAccessorMessage,
+                    cancellationToken => CreateTestAccessorAsync(context.Document, location.SourceSpan, cancellationToken),
+                    nameof(AbstractCreateTestAccessor<TTypeDeclarationSyntax>)));
         }
 
-        private async Task<Document> CreateTestAccessorAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        private async Task<TTypeDeclarationSyntax?> GetRelevantTypeFromHeaderAsync(CodeRefactoringContext context)
+        {
+            var type = await context.TryGetRelevantNodeAsync<TTypeDeclarationSyntax>(RefactoringHelpers).ConfigureAwait(false);
```
mavasani, line 53: Can we inline this call and remove `GetRelevantTypeFromHeaderAsync`?

sharwell: Done

PR #3694, opened 2020-05-30
Implement CodeFix for CA2246

### LINKED ISSUE #2527: [Proposal] Compiler warnings with unobvious assignment
`C#` is a beautiful and flexible language that will not allow a beginner to *"shoot off his leg"*... or it will?

-----------

## Issue

Let's imagine that we have the following class:

```csharp
class Foo
{
    public int Value;
    public Foo Next;
}
```

Then we want to do something like this:

```csharp
var a = new Foo();
var b = new Foo { Value = 1, Next = new Foo { Value = 2 } };
a.Next = a = b;
Console.WriteLine(a.Next == a);
```

What we'll see in the console's output?
Most of you're now thinking something like this:

> It's kinda simple!
> First we assign "b" to "a", so "a" and "b" are now the same
> Then we assign "a" (or "b", cause they're the same now) to the "Next" field of object "a" (or "b")
> So `a.Next == a` is `true`!

And I have some bad news for you, cause it's `false` :D
If you're surprised, then let me surprise you even more:

```csharp
a = b;
a.Next = a; // or b
Console.WriteLine(a.Next == a); // true
```

---------

## Explanation

When we have construction like this:
```csharp
a = b;
a.x = b;
```

As a result, we'll get this `IL`-code:
```asm
;store b into a
ldloc b
stloc a 

;store b into a.x
ldloc a
ldloc b 
stfld class Foo Foo::X
```

But if we want to simplify this, and we'll use the following construction (that should mean the same as the example above):
```csharp
a.x = a = b; 
```

`IL` will look like this:
```asm
ldloc a ;load "a" to the top of the stack
ldloc b ;load "b" to the top of the stack
dup     ;load "b" to the top of the stack
stloc a ;store "b" to "a"
stfld class Foo Foo::X ;store "b" to an object on top of the stack that is no longer "a"!
```
As we see, the assignment happens to the object we didn't expected (cause we know that assignment goes from right to left).

```csharp
var aClone = a;
a.x = a = b; 
Console.WriteLine(a.x == a); // false
Console.WriteLine(a.x == b); // false
Console.WriteLine(aClone.x == a); // true
Console.WriteLine(aClone.x == b); // true
```

--------------

## Proposal

I found this behaviour kinda unobvious. Especially for those who aren't familiar with the *"subtleties"* of `C#`

So I want to suggest introducing **compiler warnings** for those cases where the left side of an assignment contains a variable, the object inside of which changes somewhere in the right side of the expression, since it wouldn't be very cool to replace `a = b; a.x = b;` with a similar `a.x = a = b;` and spend a few sleepless nights in search of a bug :D

Youssef1313:
Fixes #2527


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            DocumentEditor editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            var parentAssignment = (Microsoft.CodeAnalysis.CSharp.Syntax.AssignmentExpressionSyntax)node.Parent;
+
+            var rightAssignment = (Microsoft.CodeAnalysis.CSharp.Syntax.AssignmentExpressionSyntax)parentAssignment.Right.WithTriviaFrom(parentAssignment);
+
+            var newSecondAssignment = parentAssignment.WithRight(rightAssignment.Right);
+
+            editor.ReplaceNode(parentAssignment, rightAssignment);
+            editor.InsertAfter(rightAssignment, newSecondAssignment);
```
Youssef1313, line 47: I'm experimenting to implement the codefix, but it's not working. I believe the problem is making multiple modifications to the DocumentEditor, but I don't know how to get it working.

Note: I know that my current implementation here need to be moved to a `CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer`, and make a parallel one for VB, and also mark the parent as abstract. I'll do that later when I get a first working code here.

@mavasani Can you help here? Is there something I should write in between the `ReplaceNode` and `InsertAfter` node? or there is another approach to do what I want?

Evangelink: I am not an expert (yet) with code-fixes but there is something I don't get with the logic you are putting in place here. In your example `a.Field = a = b` we have:
- `parentAssignment` = `a.Field = a = b`
- `rightAssignment` = `a = b`
- `newSecondAssignment` = `a.Field = b`

Now you replace `a.Field = a = b` with `a = b` and then you add right after `a.Field = b` but in the expected code-fix you have written `a.Field = b = a`.

IMO (note that I haven't tested) we should generate a new node of type equality (using SyntaxGenerator) with left `b` and right `a` taken from `rightAssignment` and then replace the code of `newSecondAssignment` to be `var newSecondAssignment = parentAssignment.WithRight(newEqualityNode)`. After that all you need to do is the replace.

Youssef1313: I couldn't understand the reasoning for my code not to work.

But let me try the path you referred to.

The method I find that best suite the situation in the SyntaxFactory class is `SyntaxFactory.List(IEnumerable<TNode>)`. (Did you mean that method?)

So, I give it the `rightAssignment` and the `newSecondAssignment`, it will return `SyntaxList<TNode>`. Now, what should I do with that returned value?, how to replace the parentAssignment with the returned `SyntaxList`?

Youssef1313: > I am not an expert (yet) with code-fixes

You're really much better than me :)
I'm just a beginner, trying to gain some experience with open-source projects.

Evangelink: Sorry I was too lightweight on my explanation. The `SyntaxFactory` allows you to create any new kind of syntax you could find in C#. In your example, you want to transform `a.Field = a = b` into `a.Field = b = a`, to do so you actually want to create a new assignment expression (`b = a`) out of the old assignment expression `a = b`). You should be able to do something like `SyntaxFactory.AssignmentExpression(SyntaxKind.SimpleAssignmentExpression, rightAssignment.Right, rightAssignment.Left)` which will return the new assignment. Then if you do the `parentAssignment.WithRight(reversedAssignment)` you will replace the right part of `a.Field = a = b` (i.e. `a = b`) with your new assignment `b = a` and so end-up with `a.Field = b = a`. 

Note that you will probably have to add `WithoutTrivia` on the new right part to avoid the space before the semicolon `b = a ;` and you might also have to add some trailing space on the new left part to avoid `b= a ;`.

I would also suggest to add one more test with `a.Field = a = b = c;` to ensure the code is working properly for a more complex scenario.

And also you might want to use safe-cast rather than cast to avoid having your codefix blowing in case of invalid code. It's possible to have invalid (not compiling) code.

Youssef1313: Ops, Sorry, My fault. The test I wrote isn't correct.
I want to transform `a.Field = a = b;` into **two statements**:

```csharp
a = b;
a.Field = b;
```

Youssef1313: Although the explanation was for the unintended behavior stated in my wrong test, but the explanation was useful anyway :)

Evangelink: Oh! I see! Hum I will need to pull your PR and make some tests because I don't know enough codefixes to see what could be wrong.

Evangelink: Looking at the test result, it seems that the new node is not part of a syntax list as requested by the [InsertAfter method](https://docs.microsoft.com/dotnet/api/microsoft.codeanalysis.editing.syntaxeditor.insertafter?view=roslyn-dotnet#Microsoft_CodeAnalysis_Editing_SyntaxEditor_InsertAfter_Microsoft_CodeAnalysis_SyntaxNode_Microsoft_CodeAnalysis_SyntaxNode_). I am not sure what shall be called to add a new instruction to the blocks so I will do a bit of debugging.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.Fixer.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.QualityGuidelines.AssigningSymbolAndItsMemberInSameStatement,
+    Microsoft.CodeQuality.Analyzers.QualityGuidelines.AssigningSymbolAndItsMemberInSameStatementFixer>;
+
+namespace Microsoft.CodeAnalysis.NetAnalyzers.UnitTests.Microsoft.CodeQuality.Analyzers.QualityGuidelines
+{
+    public class AssigningSymbolAndItsMemberInSameStatementFixerTests
```
sharwell, line 11: 💡 Don't separate the analyzer and code fix tests. Rather, update the tests in-place to call `VerifyCodeFixAsync` instead of `VerifyAnalyzerAsync`.

* If no code fix is offered, use `VerifyCodeFixAsync(code, code)`. This validates that no code fix is offered for the scenario.
* If the code fix is offered, use `VerifyCodeFixAsync(code, fixedCode)`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Evangelink: I have found something that seems to be working but it feels quite wrong so I suspect there is a better way. My solution:

1. Create 2 new `ExpressionStatementSyntax` using the `SyntaxFactory` to represent the 2 statements you want to add.
2. Find ancestor of type `BlockSyntax`
3. Take all statements of this block and replace the old statement with the 2 new
4. Call `WithStatements`
5. Call `ReplaceNode` to replace the block

@sharwell Any pointer on a better solution?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.Fixer.cs
Youssef1313: Thanks. I'll update that soon.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink Thanks for making some trials and getting to that solution.
I'll wait and see if @sharwell sees a better solution.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.Fixer.cs
Youssef1313: @sharwell FYI, there are some splitted tests in that same folder.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,64 @@
...
+            SyntaxNode node = root.FindNode(context.Span);
+
+            if (node == null)
+            {
+                return;
+            }
+
+            string title = MicrosoftCodeQualityAnalyzersResources.AssigningSymbolAndItsMemberInSameStatementTitle;
+            context.RegisterCodeFix(new MyCodeAction(title,
+                 async ct => await SplitAssignment(context.Document, node, ct).ConfigureAwait(false),
```
mavasani, line 32: I believe we should offer 3 code fixes here as suggested in https://github.com/dotnet/roslyn-analyzers/issues/2527#issuecomment-530772002. Otherwise, we might misdirect user into making a breaking change assuming the tooling is recommending this code fix, hence it should be the appropriate one. `CA2246` is certainly one case where we should either have all fixes or none.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: ping @sharwell

Evangelink: @mavasani If you have the chance to ping @sharwell in a meeting, that would help @Youssef1313 moving forward on this PR.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,91 @@
...
+                return document;
+            }
+
+            // a = b;
+            right = right.WithTriviaFrom(parentAssignment);
+
+            // a.x = b;
+            var firstEqualsLastAssignment = GetAssignmentWithRight(parentAssignment, rightOfRight);
+
+            root = root.ReplaceNode(parentAssignment, new[] { right, firstEqualsLastAssignment });
```
Youssef1313, line 71: @Evangelink I re-wrote this without using the `DocumentEditor`. But it's still not working:

This line is failing with:

> System.InvalidOperationException : The item specified is not the element of a list.

1. The exception is unclear from Roslyn's side.
2. Assuming that this exception means that the node isn't part of the tree, I can't see any reason for that.

Any idea?

Pinging @sharwell and @mavasani as well.

Youssef1313:
@Evangelink @mavasani This is ready for another look.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -112,17 +206,41 @@ public class Test
...
+                FixedState = { Sources = { fixedCode } },
+                CodeActionIndex = codeActionIndex,
+            }.RunAsync();
         }
 
-        [Fact]
-        public async Task CSharpReassignGlobalVariableAndReferToItsField()
+        [Theory]
+        [InlineData(0, "x.Property.Property = y;")]
+        [InlineData(1, "x.Property.Property = x;")]
```
Youssef1313, line 240:
```suggestion
        [InlineData(1, "x.Property.Property = x.Property;")]
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -29,14 +31,38 @@ public void Method()
         a.Field = a = b;
```
Youssef1313, line 31:
```suggestion
        [|a.Field|] = a = b;
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -47,17 +73,40 @@ public class Test
     public void Method()
     {
         C a = new C(), b = new C(), c;
-        a.Property = c = a = b;
+        [|a.Property|] = c = a = b;
```
Youssef1313, line 76: @mavasani What should be the codefix(es) for this?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -89,18 +161,40 @@ public class Test
     public void Method()
     {
         C a = new C(), b = new C();
-        a.Property.Property = a.Property = a = b;
+        [|a.Property.Property|] = [|a.Property|] = a = b;
```
Youssef1313, line 164: Not sure what should be the codefix(es) for this one too.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            var title = MicrosoftCodeQualityAnalyzersResources.AssigningSymbolAndItsMemberInSameStatementTitle;
+            context.RegisterCodeFix(new MyCodeAction(title,
+                 async ct => await SplitAssignmentFirstOption(context.Document, context.Span, ct).ConfigureAwait(false),
+                 equivalenceKey: title + "0"),
+            context.Diagnostics);
+            context.RegisterCodeFix(new MyCodeAction(title,
+                 async ct => await SplitAssignmentSecondOption(context.Document, context.Span, ct).ConfigureAwait(false),
+                 equivalenceKey: title + "1"),
+            context.Diagnostics);
```
Evangelink, line 47: I am wondering if:

1. We should have a different title for the 2 code-fixes

2. Transform the following into the a [nested code-fix](https://docs.microsoft.com/dotnet/api/microsoft.codeanalysis.codeactions.codeaction.create#Microsoft_CodeAnalysis_CodeActions_CodeAction_Create_System_String_System_Collections_Immutable_ImmutableArray_Microsoft_CodeAnalysis_CodeActions_CodeAction__System_Boolean_)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        private static async Task<Document> SplitAssignmentSecondOption(Document document, TextSpan span, CancellationToken cancellationToken)
+        {
+            // This method splits `a.x = a = b` to:
+            // a = b;
+            // a.x = a;
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+
+            // a.x = a = b;
+            var parentAssignment = root.FindNode(span).Parent;
```
Evangelink, line 85: Why don't you do the cast here? Instead of having a direct-cast directly in the various sub-method called.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
Evangelink: > // For something like `a.x = a = b;`, we offer 3 code fixes:
            // First:
            //     a = b;
            //     a.x = b;
            // Second:
            //     a = b;
            //     a.x = a;
            // Third: (Not currently implemented)
            //     var temp = a;
            //     a = b;
            //     temp.x = b;

So I would say that you should end-up with:

```csharp
a = b;
c = a;
a.Property = b;
```

and/or

```csharp
a = b;
c = a;
a.Property = c;
```

Basically we take the far left or far right member.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
Evangelink: Based upon my previous comment, I would say:

```csharp
a = b;
a.Property = a;
a.Property.Property = b;
```

and/or

```csharp
a = b;
a.Property = a;
a.Property.Property = a.Property;
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+            right = GetExpressionFromAssignment(right).WithTriviaFrom(parentAssignment.Parent);
+
+            // a.x = b;
+            var firstEqualsLastAssignment = GetExpressionFromAssignment(GetAssignmentWithRight(parentAssignment, rightOfRight));
+
+            root = root.ReplaceNode(parentAssignment.Parent, new[] { right, firstEqualsLastAssignment });
+            return document.WithSyntaxRoot(root);
+        }
+
+        private static async Task<Document> SplitAssignmentSecondOption(Document document, TextSpan span, CancellationToken cancellationToken)
```
Evangelink, line 77: It seems easy to share/extract/factorize the code of the 2 methods. As far as I can see the differences are:

1. `!TryGetAssignmentExpressionParts(right, out _, out var rightOfRight)` vs `!TryGetAssignmentExpressionParts(right, out var leftOfRight, out _)`

2. `var firstEqualsLastAssignment = GetExpressionFromAssignment(GetAssignmentWithRight(parentAssignment, rightOfRight));` vs `var firstEqualsSecondAssignment = GetExpressionFromAssignment(GetAssignmentWithRight(parentAssignment, leftOfRight));`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+            // a.x = a;
+            var firstEqualsSecondAssignment = GetExpressionFromAssignment(GetAssignmentWithRight(parentAssignment, leftOfRight));
+
+            root = root.ReplaceNode(parentAssignment.Parent, new[] { right, firstEqualsSecondAssignment });
+            return document.WithSyntaxRoot(root);
+        }
+
+        private static SyntaxNode GetAssignmentWithRight(SyntaxNode assignmentExpression, SyntaxNode newRight)
+        {
+            return ((AssignmentExpressionSyntax)assignmentExpression).WithRight((ExpressionSyntax)newRight);
```
Evangelink, line 105: [cosmetic] You could use expression-bodied.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        private static SyntaxNode GetExpressionFromAssignment(SyntaxNode assignmentExpression)
+        {
+            return SyntaxFactory.ExpressionStatement((AssignmentExpressionSyntax)assignmentExpression);
```
Evangelink, line 110: [cosmetic] You could use expression-bodied.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            if (assignmentExpression is AssignmentExpressionSyntax assignment)
+            {
+                left = assignment.Left;
+                right = assignment.Right;
+                return true;
+            }
+            left = null;
+            right = null;
+            return false;
```
Evangelink, line 123:
```suggestion
            var assignment = assignmentExpression as AssignmentExpressionSyntax;

			left = assignment?.Left;
			right = assignment?.Right;

			return assignment != null;
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        public override FixAllProvider GetFixAllProvider()
+        {
+            return WellKnownFixAllProviders.BatchFixer;
```
Evangelink, line 128: [cosmetic] You could use expression-bodied.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink Yes I should do that.
I was writing this first as an abstract base class for both C# and VB before I realize that it's not applicable to VB.

Youssef1313: I think I'll probably need to change the way this works as it currently fails when the assignment have 4 members.

Youssef1313:
@Evangelink I wasn't able to handle the case of 4 assignment members (`a.Property = c = a = b;`), so I've done some refactoring. (or *nearly* a re-write). It seems to work (with one problem with handling trivia). But I feel the implementation can be improved.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            var title = MicrosoftCodeQualityAnalyzersResources.AssigningSymbolAndItsMemberInSameStatementTitle;
+
+            var replacements1 = replacements.Concat(GetAssignmentExpressionStatement(assignment.Left, mostRightMember, leadingTrivia, trailingTrivia));
+            var replacements2 = replacements.Concat(GetAssignmentExpressionStatement(members, leadingTrivia, trailingTrivia));
+
+#pragma warning disable RS1010 // Provide an explicit value for EquivalenceKey - false positve 
+            var nestedCodeAction = CodeAction.Create(title, ImmutableArray.Create<CodeAction>(
+                new MyCodeAction($"{title} 1", ct => GetDocument(context.Document, root, assignment.Parent, replacements1)),
+                new MyCodeAction($"{title} 2", ct => GetDocument(context.Document, root, assignment.Parent, replacements2))
```
Evangelink, line 72: The `title` parameter is what get displayed so it feels off to have `XXX 1` and `XXX 2`.

Evangelink (commented):
I will need a second review (tomorrow) but there are some things off.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,122 @@
...
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+#nullable enable
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
```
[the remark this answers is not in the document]
Youssef1313: @Evangelink You are correct. But this is existing problem before this PR. The analyzer itself is also in the common project.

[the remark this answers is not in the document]
Youssef1313: I believe the common project shouldn't even allow the usage of these namespaces.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink I have this temporary until I can get with good titles. Do you have any suggestions? Not sure how to describe the code fixes.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Threading.Tasks;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+#nullable enable
```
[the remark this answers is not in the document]
Youssef1313: Doesn't seem to be required. Good catch.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatement.Fixer.cs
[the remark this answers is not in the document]
mavasani: Yes, it should be moved to C# layer.

src/NetAnalyzers/CSharp/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA2246 | Usage | Info | CSharpAssigningSymbolAndItsMemberInSameStatement, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2246)
```
Youssef1313, line 5: @mavasani, I got a warning to add it to unshipped analyzers. But this doesn't seem correct. It's already shipped. The warning occurred after moving it from Core layer to C# layer.

How to get things automated to move it from Unshipped to Shipped in this layer, and to delete  it from Shipped in the Core layer?

mavasani: @Youssef1313 Can you manually move the existing shipped file entry in core layer to the shipped files in language specific layers?

Youssef1313:
I've fixed the trivia handling. However, all tests that contains a comparison of 4 elements (e.g. `[|a.Property.Property|] = [|a.Property|] = a = b;`) are failing.

It looks like that after the first replacement, the second will fail as it tries to replace a node that now isn't part of the syntax tree.

I faced such a problem before and the solution was to use ReplaceNode which takes a lambda to compute replacement. But that solution is very hard to apply in this specific case. I'm out of ideas for how to fix these tests.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+                // For following statements, take only whitespace trivia.
+                trailingTrivia = trailingTrivia.Where(t => t.IsKind(SyntaxKind.WhitespaceTrivia) || t.IsKind(SyntaxKind.EndOfLineTrivia)).ToSyntaxTriviaList();
+            }
+
+            var title = MicrosoftCodeQualityAnalyzersResources.AssigningSymbolAndItsMemberInSameStatementTitle;
+
+            var replacements1 = replacements.Concat(GetAssignmentExpressionStatement(assignment.Left, mostRightMember, leadingTrivia, trailingTrivia));
+            var replacements2 = replacements.Concat(GetAssignmentExpressionStatement(members, leadingTrivia, trailingTrivia));
+
+#pragma warning disable RS1010 // Provide an explicit value for EquivalenceKey - false positve
```
Evangelink, line 70: Instead of having this pragma, could we bump the version of the analyzer we are using to rely on one where I have fixed this issue?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            var mostRightMember = members.Peek(); // Don't move this near its usage. The result will be different.
+            var leadingTrivia = assignment.Parent.GetLeadingTrivia();
+            var trailingTrivia = assignment.Parent.GetTrailingTrivia();
+            var replacements = new List<SyntaxNode>(members.Count - 1);
+
+            while (members.Count > 2)
+            {
+                replacements.Add(GetAssignmentExpressionStatement(members, leadingTrivia, trailingTrivia));
```
Evangelink, line 59: I am not sure which change to suggest (maybe the method name) but when reviewing it feels weird because we loop on `members.Count` and it's not really clear there is a change to this variable. At first I thought there was a "silly" bug.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+                // For the first statement, take all trivia.
+                // For following statements, take only whitespace trivia.
+                trailingTrivia = trailingTrivia.Where(t => t.IsKind(SyntaxKind.WhitespaceTrivia) || t.IsKind(SyntaxKind.EndOfLineTrivia)).ToSyntaxTriviaList();
+            }
+
+            var title = MicrosoftCodeQualityAnalyzersResources.AssigningSymbolAndItsMemberInSameStatementTitle;
+
+            var replacements1 = replacements.Concat(GetAssignmentExpressionStatement(assignment.Left, mostRightMember, leadingTrivia, trailingTrivia));
+            var replacements2 = replacements.Concat(GetAssignmentExpressionStatement(members, leadingTrivia, trailingTrivia));
```
Evangelink, line 68: Could you please use a meaningful name instead of XXX1 and XXX2? Maybe something like `finalAssignmentToLeft` and `finalAssignmentToRight`.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+
+#pragma warning disable RS1010 // Provide an explicit value for EquivalenceKey - false positve 
+            var nestedCodeAction = CodeAction.Create(title, ImmutableArray.Create<CodeAction>(
+                new MyCodeAction($"{title} 1", ct => GetDocument(context.Document, root, assignment.Parent, replacements1)),
+                new MyCodeAction($"{title} 2", ct => GetDocument(context.Document, root, assignment.Parent, replacements2))
```
Evangelink, line 73: Same name suggestion here. We will currently see something like:

```
title
|-> title 1
|-> title 2
```

in the lightbulb, that's not really helpful.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+                ), isInlinable: false);
+#pragma warning restore RS1010
+
+            context.RegisterCodeFix(nestedCodeAction, context.Diagnostics);
+        }
+
+        /// <summary>
+        /// If the assignment expression is:  a = b = c = d
+        /// Return a stack containing a, b, c, d with `a` at the bottom and `d` at the top.
```
Evangelink, line 82:
```suggestion
        /// Return a stack containing a, b, c, d with 'a' at the bottom and 'd' at the top.
```

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// </summary>
+        private static Stack<ExpressionSyntax> GetAssignmentMembers(AssignmentExpressionSyntax node)
```
Evangelink, line 84: What about `SplitAssignments`?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
```diff
@@ -0,0 +1,123 @@
...
+            return GetAssignmentExpressionStatement(left, right, leadingTrivia, trailingTrivia);
+        }
+
+        private static ExpressionStatementSyntax GetAssignmentExpressionStatement(ExpressionSyntax left, ExpressionSyntax right, SyntaxTriviaList leadingTrivia, SyntaxTriviaList trailingTrivia)
+            => SyntaxFactory.ExpressionStatement(
+                SyntaxFactory.AssignmentExpression(SyntaxKind.SimpleAssignmentExpression, left, right))
+                    .WithLeadingTrivia(leadingTrivia).WithTrailingTrivia(trailingTrivia);
+
+        private static Task<Document> GetDocument(Document document, SyntaxNode root, SyntaxNode oldNode, IEnumerable<SyntaxNode> replacements)
+            => Task.FromResult(document.WithSyntaxRoot(root.ReplaceNode(oldNode, replacements)));
```
Evangelink, line 110: I am not sure but maybe using the `DocumentEditor` would help with solving the issue you are facing.

src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md
```diff
@@ -153,7 +153,6 @@ CA2242 | Usage | Info | TestForNaNCorrectlyAnalyzer, [Documentation](https://doc
 CA2243 | Usage | Disabled | AttributeStringLiteralsShouldParseCorrectlyAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2243)
 CA2244 | Usage | Info | AvoidDuplicateElementInitialization, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2244)
 CA2245 | Usage | Info | AvoidPropertySelfAssignment, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2245)
-CA2246 | Usage | Info | AssigningSymbolAndItsMemberInSameStatement, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2246)
```
Evangelink, line 156: Not sure about this change.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -26,17 +27,41 @@ public class Test
...
+        C a = new C(), b = new C();
+        a = b;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 53:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -47,17 +72,41 @@ public class Test
...
+        a = b;
+        c = a;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 99:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -68,17 +117,40 @@ public class Test
...
+        C a = new C(), b = new C();
+        a.Property = b;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 143:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -89,18 +161,41 @@ public class Test
...
+        a = b;
+        a.Property = a;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 188:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -112,17 +207,41 @@ public class Test
...
+    {{
+        x = y;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 234:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -134,17 +253,41 @@ public class Test
...
+    {{
+        x.Property = y;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 280:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -156,18 +299,42 @@ public class Test
...
+        x = y;
+        x.Property = x;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 327:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -180,12 +347,35 @@ public class Test
...
```
Evangelink, line 376:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AssigningSymbolAndItsMemberInSameStatementTests.cs
```diff
@@ -262,11 +454,32 @@ public class Test
...
+        C a = new C();
+        b = a;
+        {fix}
+    }}
+}}
+";
+            await new VerifyCS.Test
+            {
+                TestState = { Sources = { code } },
+                FixedState = { Sources = { fixedCode } },
```
Evangelink, line 480:
```suggestion
                TestCode = code,
                FixedCode = fixedCode,
```

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink Sure. Do you know which version got the fix?
We will also need to remove the suppression in:

https://github.com/dotnet/roslyn-analyzers/blob/91b109f2f69496b76e7a740a4547a9d9b9a12c0b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/MarkAttributesWithAttributeUsage.Fixer.cs#L56

src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md
Youssef1313: @Evangelink See https://github.com/dotnet/roslyn-analyzers/pull/3694#discussion_r502807776

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink If I used the same wording you suggested in https://github.com/dotnet/roslyn-analyzers/pull/3694#discussion_r503160986 as titles (e.g. `Split assignments with final assignment to the left variable`), do you think it's good to the end-user?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink To me, I'd interpret `SplitAssignments` such that it returns a collection of the new assignment statements. While it returns a collection of each member in the assignment.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Evangelink: Just made a test and it seems that the latest available pre-release (i.e. 3.3.1-beta1.20505.3) does not contain the fix so let's move on as-is and we will do a separate PR to get rid of both pragmas.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Evangelink: I think so. @mavasani any opinion?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Evangelink: You are right. I think I am a bit bothered by `Members` because it's more `Expressions` but it probably doesn't matter much.

Youssef1313: I'm okay with renaming to `GetAssignmentExpressions` if you think it's clearer.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAssigningSymbolAndItsMemberInSameStatement.Fixer.cs
Youssef1313: @Evangelink Can you elaborate more on how I should use `DocumentEditor` to fix this?

PR #3696, opened 2020-05-30
Fix 'Default struct constructor missing from public API'

### LINKED ISSUE #2196: Default struct constructor missing from public API
#### Analyzer package

DotNetAnalyzers.PublicApiAnalyzer

#### Package Version

Example: 2.9.0-beta1.19162.2+321600d1

#### Repro steps

1. Create a public struct with a private constructor with one or more parameters
1. Add all missing APIs to PublicAPI.Unshipped.txt

#### Expected behavior

The default constructor for the public struct appears in PublicAPI.Unshipped.txt.

#### Actual behavior

The default constructor is not included in PublicAPI.Unshipped.txt.

Youssef1313:
Fixes #2196

@sharwell , I just ported the solution you referred to in the issue.


src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -180,13 +180,13 @@ private void OnSymbolActionCore(ISymbol symbol, Action<Diagnostic> reportDiagnos
...
                 {
                     var namedType = (INamedTypeSymbol)symbol;
-                    if (namedType.InstanceConstructors.Length == 1 &&
-                        (namedType.TypeKind == TypeKind.Class || namedType.TypeKind == TypeKind.Struct))
+                    if ((namedType.InstanceConstructors.Length == 1 && namedType.TypeKind == TypeKind.Class)
+                        || namedType.TypeKind == TypeKind.Struct)
                     {
-                        var instanceConstructor = namedType.InstanceConstructors[0];
-                        if (instanceConstructor.IsImplicitlyDeclared)
+                        var implicitConstructor = namedType.InstanceConstructors.SingleOrDefault(x => x.IsImplicitlyDeclared);
```
Evangelink, line 186: My understanding is that we want to avoid as much as possible to avoid ending-up in situations where an analyzer throws so it might be better to either do a `FirstOrDefault` and disregard the case where we have multiple entries or do a `Where` and bail-out if we have too many items.

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1224,6 +1224,51 @@ public struct Nested { }
                 );
         }
 
+        [Fact]
+        public async Task ShippedTextWithMissingImplicitStructConstructorAsync()
```
Evangelink, line 1228: [cosmetic] We usually don't use the `Async` suffix for tests

```suggestion
        public async Task ShippedTextWithMissingImplicitStructConstructor()
```

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1224,6 +1224,51 @@ public struct Nested { }
...
+}
+";
+
+            var shippedText = @"
+C
+C.C(int value) -> void";
+            var unshippedText = string.Empty;
+
+            // Test0.cs(2,15): warning RS0016: Symbol 'implicit constructor for C' is not part of the declared API.
+            string arg = string.Format(PublicApiAnalyzerResources.PublicImplicitConstructorErrorMessageName, "C");
```
Evangelink, line 1265: I am surprised that the CI didn't fail here. We should only test for arguments not for messages in the tests.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
Youssef1313: I'd wait for @sharwell on this as this PR just a mirror on his PR in the old repo.

mavasani: +1 - It is almost always wrong to use `Single` or `SingleOrDefault` in analyzers/fixers when attempting to find matching symbols. My preference would be to handle this gracefully as suggested by @Evangelink

Youssef1313:
```suggestion
                        var implicitConstructor = namedType.InstanceConstructors.FirstOrDefault(x => x.IsImplicitlyDeclared);
```

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
Evangelink: Nevermind my comment was wrong. The resource is building the `implicit constructor for {0}` part.

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1224,6 +1224,51 @@ public struct Nested { }
...
+public struct C
+{
+}
+";
+
+            var shippedText = @"
+C";
+            var unshippedText = string.Empty;
+
+            // Test0.cs(2,15): warning RS0016: Symbol 'implicit constructor for C' is not part of the declared API.
```
mavasani, line 1240: Why is there no code fix for this diagnostic? All `RS0016` must have a corresponding code fix with FixAll support, otherwise they become very onerous.

Youssef1313:
> 
> 
> Please make sure the new violations have a corresponding code fix. If they already do, please add unit tests to verify code fix application.

Sorry for missing this for long. Codefix tests are added.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -182,13 +182,13 @@ private void OnSymbolActionCore(ISymbol symbol, Action<Diagnostic> reportDiagnos
                 if (symbol.Kind == SymbolKind.NamedType)
                 {
                     var namedType = (INamedTypeSymbol)symbol;
-                    if (namedType.InstanceConstructors.Length == 1 &&
-                        (namedType.TypeKind == TypeKind.Class || namedType.TypeKind == TypeKind.Struct))
+                    if ((namedType.InstanceConstructors.Length == 1 && namedType.TypeKind == TypeKind.Class)
```
Evangelink, line 185: [cosmetic] I would move the `TypeKind` check first just so that when reading we have the 2 typekind check aligned.

Evangelink (approved):
LGTM

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
Youssef1313:
```suggestion
                    if ((namedType.TypeKind == TypeKind.Class && namedType.InstanceConstructors.Length == 1)
```

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1295,6 +1295,43 @@ public struct Nested { }
...
+            await VerifyCSharpAdditionalFileFixAsync(source, shippedText, unshippedText, fixedUnshippedText);
+        }
+
+        [Fact]
+        public async Task ShippedTextWithMissingImplicitStructConstructorWithOtherOverloadsAsync()
+        {
+            var source = @"
+public struct {|RS0016:C|}
+{
+    public C(int value)
```
mavasani, line 1320: Can you also add a test matching the repro steps in the issue being fixed?

mavasani (approved):

Youssef1313:
@mavasani The upload to codecov failed the CI. I thought this was fixed in #4255?

mavasani:
Yes, seems like my PR did not fix the issue.

sharwell:
> Yes, seems like my PR did not fix the issue.

I wouldn't expect it to. The error status is controlled here:
https://github.com/dotnet/roslyn-analyzers/blob/5f3a4ce2c0fc01544b7f93077e0fafad947f62b6/eng/CodeCoverage.proj#L29-L30

Issue #3701, opened 2020-06-01
CA1829 not reporting for ImmutableArray


Youssef1313:
#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

[CA1829](https://docs.microsoft.com/visualstudio/code-quality/ca1829)

#### Repro steps

![image](https://user-images.githubusercontent.com/31348972/83401967-3427bd00-a406-11ea-80fa-ca05c42ad511.png)

#### Expected behavior

CA1829 is reported in both methods.

#### Actual behavior

CA1829 is reported only in the `int[]` case, but not in `ImmutableArray<int>` case.

Youssef1313:
It seems that there are unit tests for ImmutableArray, so I'm not sure  what's going on that it's not reported in my case

mavasani:
I think this might be fixed now- @Jozkee

Youssef1313:
@mavasani Did you start working on @Jozkee's issue? Because I started trying with it too 😄

mavasani:
@Jozkee re-implemented the analyzer recently, so i’d let him take it up.

Jozkee:
>It seems that there are unit tests for ImmutableArray

I run those tests you mention but those work fine and they seem to assert for the issue that you are reporting. 

I also tried creating a new project, install Microsoft.CodeAnalysis.FxCopAnalyzers v 3.0.0 and I was unable to duplicate this issue locally on my VS.

@mavasani is there something else we can do to verify that we are doing the correct steps?

mavasani:
@Youssef1313 Can you please try this with the latest package from here: https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet5&package=Microsoft.CodeAnalysis.FxCopAnalyzers&protocolType=NuGet&version=3.3.0-beta1.20302.1

Youssef1313:
It works. Closing.

PR #3702, opened 2020-06-01
Handle ArgumentSyntax in CA1829 fixer

### LINKED ISSUE #3700: CA1829: when Count() is used as an argument Analyzer reports successfully but Fixer fails
#### Analyzer package
Microsoft.CodeAnalysis.NetAnalyzers

#### Package Version
Latest bits

#### Diagnostic ID
[CA1829](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1829)

#### Repro steps
```cs
using System;
using System.Linq;
namespace Test
{
    class C
    {
        global::System.Collections.Concurrent.ConcurrentBag<int> GetData() => default;
        void M()
        {
            var b = 1.Equals(GetData().Count());
        }
    }
}

```
#### Behavior
Analyzer reports `GetData().Count()` but fixer fails, it does not bring up the suggestion.

Youssef1313:
From my basic analysis, the problem is that the fixer expects the node to be `InvocationExpressionSyntax`, otherwise, it won't attempt to fix anything.
My change checks if the node is `ArgumentSyntax`, then get its expression, and expect that expression to be `InvocationExpressionSyntax`.

This PR fixes the problem. However, I don't know if this is the correct way to fix it, or there is a more deeper issue.

Fixes #3700

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
@@ -49,6 +49,7 @@ namespace Microsoft.NetCore.Analyzers.Performance
         {
             var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
             var node = root.FindNode(context.Span, getInnerMostNodeForTie: true);
+
             if (node is object &&
                 context.Diagnostics[0].Properties.TryGetValue(UseCountProperlyAnalyzer.PropertyNameKey, out var propertyName) &&
                 propertyName is object &&
@@ -73,8 +74,6 @@ namespace Microsoft.NetCore.Analyzers.Performance
             SyntaxNode invocationNode,
             [NotNullWhen(returnValue: true)] out SyntaxNode? memberAccessNode,
             [NotNullWhen(returnValue: true)] out SyntaxNode? nameNode);
-
-        protected abstract SyntaxNode TryGetInvocationNodeFromArgumentSyntax(SyntaxNode node);
 
         /// <summary>
         /// Implements the <see cref="CodeAction"/> for replacing the use of <see cref="System.Linq.Enumerable.Count{TSource}(System.Collections.Generic.IEnumerable{TSource})"/> 
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
@@ -49,14 +49,13 @@ namespace Microsoft.NetCore.Analyzers.Performance
         {
             var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
             var node = root.FindNode(context.Span, getInnerMostNodeForTie: true);
-            var invocationNode = TryGetInvocationNodeFromArgumentSyntax(node);
             if (node is object &&
                 context.Diagnostics[0].Properties.TryGetValue(UseCountProperlyAnalyzer.PropertyNameKey, out var propertyName) &&
                 propertyName is object &&
-                TryGetExpression(invocationNode, out var expressionNode, out var nameNode))
+                TryGetExpression(node, out var expressionNode, out var nameNode))
             {
                 context.RegisterCodeFix(
-                    new UsePropertyInsteadOfCountMethodWhenAvailableCodeAction(context.Document, invocationNode, expressionNode, nameNode, propertyName),
+                    new UsePropertyInsteadOfCountMethodWhenAvailableCodeAction(context.Document, node, expressionNode, nameNode, propertyName),
                     context.Diagnostics);
             }
         }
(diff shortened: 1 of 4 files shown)


Youssef1313:
@mavasani, I just saw [your comment](https://github.com/dotnet/roslyn-analyzers/issues/3701#issuecomment-636817173) in #3701. Feel free to close this PR if it won't be needed after @Jozkee has re-implemented the analyzer, or keep it open if the new implementation of the analyzer will still need this.

mavasani:
I will let @Jozkee drive this PR/issue

Jozkee:
> Feel free to close this PR if it won't be needed after @Jozkee has re-implemented the analyze

@Youssef1313 This PR is still good, https://github.com/dotnet/roslyn-analyzers/pull/3584 merges the analyzers but left intact the existing fixers. 
Let me take a look.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -17,6 +17,46 @@ namespace Microsoft.NetCore.Analyzers.Performance.UnitTests
...
+    public static void M()
+    {{
+        var b = 1.Equals(GetData().Length);
+    }}
+}}
+" ,
+                    },
+                },
+            }.RunAsync();
+
```
Jozkee, line 59: Can you also add the same test for VB?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -17,6 +17,46 @@ namespace Microsoft.NetCore.Analyzers.Performance.UnitTests
...
+                    Sources=
+                    {
+                $@"using System;
+using System.Linq;
+public static class C
+{{
+    public static System.Collections.Immutable.ImmutableArray<int> GetData() => default;
+    public static void M()
+    {{
+        var b = 1.Equals({{|{UseCountProperlyAnalyzer.CA1829}:GetData().Count()|}});
```
Jozkee, line 35: Can you add tests where you have parenthesis wrapping `GetData().Count()` and in between?
* `(GetData().Count())`
* `(GetData()).Count()`

And the same for VB.

Jozkee:
Overall looks good however I left a few suggesting for you to consider.
Thanks @Youssef1313.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -49,14 +49,14 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            var invocationNode = TryGetInvocationNodeFromArgumentSyntax(node);
             if (node is object &&
                 context.Diagnostics[0].Properties.TryGetValue(UseCountProperlyAnalyzer.PropertyNameKey, out var propertyName) &&
                 propertyName is object &&
-                TryGetExpression(node, out var expressionNode, out var nameNode))
+                TryGetExpression(invocationNode, out var expressionNode, out var nameNode))
             {
                 context.RegisterCodeFix(
-                    new UsePropertyInsteadOfCountMethodWhenAvailableCodeAction(context.Document, node, expressionNode, nameNode, propertyName),
+                    new UsePropertyInsteadOfCountMethodWhenAvailableCodeAction(context.Document, invocationNode, expressionNode, nameNode, propertyName),
```
Youssef1313, line 59: > Instead of calling a new abstract could you just inline `TryGetInvocationNodeFromArgumentSyntax` on `TryGetExpression`?

Thanks @Jozkee for review. Doing so will cause a problem in this line, I want the correct node to be sent to the `UsePropertyInsteadOfCountMethodWhenAvailableCodeAction` ctor.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -49,14 +49,14 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
             var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
             var node = root.FindNode(context.Span);
```
mavasani, line 51:
```suggestion
            var node = root.FindNode(context.Span, getInnerMostNodeForTie: true);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailable.Fixer.cs
```diff
@@ -75,6 +75,8 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
             [NotNullWhen(returnValue: true)] out SyntaxNode? memberAccessNode,
             [NotNullWhen(returnValue: true)] out SyntaxNode? nameNode);
 
+        protected abstract SyntaxNode TryGetInvocationNodeFromArgumentSyntax(SyntaxNode node);
```
mavasani, line 78: You should not need this method.

PR #3706, opened 2020-06-02
Implement IdentifiersShouldNotContainUnderscoresFixer

### LINKED ISSUE #444: Port FxCop rule CA1707: IdentifiersShouldNotContainUnderscores
**Title:** Identifiers should not contain underscores

**Description:**

By convention, identifier names do not contain the underscore (_) character. This rule checks namespaces, types, members, and parameters.

**Dependency:** None

**Notes:**

See if StyleCop has implemented it

Youssef1313:
Fixes #444


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +25,33 @@ public sealed override FixAllProvider GetFixAllProvider()
...
+            context.RegisterCodeFix(new MyCodeAction(title,
+                                        async ct => await RenameWithoutUnderscores(context.Document, symbol, ct).ConfigureAwait(false),
+                                        equivalenceKey: title),
+                                    context.Diagnostics);
+        }
+
+        private static async Task<Document> RenameWithoutUnderscores(Document document, ISymbol symbol, CancellationToken cancellationToken)
+        {
+            Solution newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, symbol, symbol.Name.Replace("_", ""), null, cancellationToken).ConfigureAwait(false);
+            return newSolution.GetDocument(document.Id);
```
Youssef1313, line 46: This is written just like [this](https://github.com/dotnet/roslyn-analyzers/blob/0cbdd68e14722e911d50eee64ee95823b3ffa09c/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ParameterNamesShouldMatchBaseDeclaration.Fixer.cs#L59).

But I'm getting a null-reference warning:

image

Why a warning is emitted here but not there? I can't get the difference.

mavasani: GetDocument can return a null value if the document ID does not belong to the solution. You should add a `!` suppression here. We should consider adding GetRequiredXXX extension methods to our repo similar to the ones in Roslyn: https://github.com/dotnet/roslyn/blob/bc2d6a2597d07a574be4e292f9dbd9c8f975fd7f/src/Workspaces/SharedUtilitiesAndExtensions/Workspace/Core/Extensions/ISolutionExtensions.cs#L36-L68

Youssef1313: @mavasani Thanks. I didn't see the `!` there 😆 

That makes sense now.

Youssef1313: Should I add the extension methods in this PR? or leave it as another task?

A small question: why the `ISolutionExtensions` is made internal in roslyn? I can't see an obvious problem with exposing these extensions to be public.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -35,15 +35,15 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
-                                        async ct => await RenameWithoutUnderscores(context.Document, symbol, ct).ConfigureAwait(false),
+                                        async ct => await RenameWithoutUnderscoresAsync(context.Document, symbol, ct).ConfigureAwait(false),
                                         equivalenceKey: title),
                                     context.Diagnostics);
         }
 
-        private static async Task<Document> RenameWithoutUnderscores(Document document, ISymbol symbol, CancellationToken cancellationToken)
+        private static async Task<Document> RenameWithoutUnderscoresAsync(Document document, ISymbol symbol, CancellationToken cancellationToken)
         {
             Solution newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, symbol, symbol.Name.Replace("_", ""), null, cancellationToken).ConfigureAwait(false);
```
mavasani, line 45: Not sure if `symbol.Name.Replace("_", "")` is a good suggestion. For example, if I had something like `This_is_a_method`, it is likely, I want `ThisIsAMethod` or something else, but certainly not `Thisisamethod`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +25,33 @@ public sealed override FixAllProvider GetFixAllProvider()
             return WellKnownFixAllProviders.BatchFixer;
```
mavasani, line 25: We should not support a FixAll here unless we are sure we can offer a good replacement for all violations.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
mavasani: I think there is possible a pending work item to make these public with API review approval. @CyrusNajmabadi do you know?

mavasani: > Should I add the extension methods in this PR? or leave it as another task?

A separate PR might get in quicker.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: @mavasani
Does something like this makes sense?
It keeps all the letter as-is except that all letters after `_` will be capitalized.

```csharp
        private static string GetNewName(string name)
        {
            var builder = new StringBuilder();
            bool isNextUpperCase = false;
            foreach (char c in name)
            {
                if (c == '_')
                {
                    isNextUpperCase = true;
                    continue;
                }
                builder.Append(isNextUpperCase ? char.ToUpper(c, CultureInfo.InvariantCulture) : c);
                isNextUpperCase = false;
            }
            return builder.ToString();
        }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: @mavasani So, what should I return here?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
mavasani: Sounds reasonable - might want to rename `isNextUpperCase` to `isPreviousUnderscore` for clarity.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
mavasani: I think your approach in https://github.com/dotnet/roslyn-analyzers/pull/3706#discussion_r434098613 might be good enough to leave this as-is and enable a FixAll to make similar replacements everywhere.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +25,33 @@ public sealed override FixAllProvider GetFixAllProvider()
...
         {
-            // Fixer not yet implemented.
-            return Task.CompletedTask;
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var model = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
 
+            var node = root.FindNode(context.Span);
+            var symbol = model.GetDeclaredSymbol(node, context.CancellationToken);
+
+            string title = MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldNotContainUnderscoresTitle;
```
mavasani, line 36: We should use a different and more succint title for code fix, like say `Remove underscores`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
CyrusNajmabadi: > I think there is possible a pending work item to make these public with API review approval. @CyrusNajmabadi do you know?

Don't know what we're planning there.  It's a pretty awful situation all around.  i'd prefer we just break these to throw.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: @mavasani, I followed the structure from another CodeFix, it used the same title.

image

Is there something different here to use a different title unlike the other code fixes?

Should I add the fix title to the resources or just as private const ?

mavasani: > Is there something different here to use a different title unlike the other code fixes?

Fix title should ideally indicate the action being performed, rather then highlight the underlying issue. We also try to keep it smaller as it shows up in the light bulb.

> Should I add the fix title to the resources or just as private const ?

Yes, please add it the resources. All user facing strings should be added to resources file. Thanks!

Youssef1313:
@mavasani Any thing left except unit test?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +27,50 @@ public sealed override FixAllProvider GetFixAllProvider()
...
-        public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
-            // Fixer not yet implemented.
-            return Task.CompletedTask;
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var model = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
 
+            var node = root.FindNode(context.Span);
+            var symbol = model.GetDeclaredSymbol(node, context.CancellationToken);
```
mavasani, line 36: Null check and bail out for `node`.

mavasani: Also null check and bail out for `symbol`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +27,50 @@ public sealed override FixAllProvider GetFixAllProvider()
             return WellKnownFixAllProviders.BatchFixer;
         }
```
mavasani, line 29: I think you don't need C# and VB sub-types of `IdentifiersShouldNotContainUnderscoresFixer`. You can just make `IdentifiersShouldNotContainUnderscoresFixer` sealed and add the C# and VB export code fix attribute to it.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: I added null checks, but not sure what "bail out" exactly means.

Youssef1313:
I think there is something wrong in the way I wrote the test causing it to fail.

@mavasani, Can you check please?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -1,29 +1,81 @@
...
-        public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
-            // Fixer not yet implemented.
-            return Task.CompletedTask;
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var model = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
 
+            var node = root.FindNode(context.Span);
+            if (node == null) return;
```
mavasani, line 38:
```suggestion
            if (node == null)
            {
               return;
            }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -1,29 +1,81 @@
...
+
+            var symbol = model.GetDeclaredSymbol(node, context.CancellationToken);
+            if (symbol == null) return;
```
mavasani, line 41:
```suggestion
            if (symbol == null)
            {
               return;
            }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
mavasani: > not sure what "bail out" exactly means.

Meant that return on null, just like you have implemented :)

mavasani:
> I think there is something wrong in the way I wrote the test causing it to fail.

CA1707 analyzer uses multiple descriptors with same rule ID (see [here](https://github.com/dotnet/roslyn-analyzers/blob/0cbdd68e14722e911d50eee64ee95823b3ffa09c/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.cs#L44-L109)), but different messages. So the test framework cannot implicitly deduce which descriptor/message is expected from markup. You need to explicitly provide which descriptor to use, see an example test below: https://github.com/dotnet/roslyn-analyzers/blob/03d8e7dbfc82a440aa6b08792693c1a0ea89fe64/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlagTests.cs#L17-L36

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -1065,10 +1065,10 @@ End Sub
         public async Task CA1707_ClassNameHasUnderscores_Rename()
         {
             await VerifyCS.VerifyCodeFixAsync(@"
-public class [|My_class|]
+public class My_class
 {
 }
-", @"
+", GetCA1707CSharpResultAt(2, 14, SymbolKind.NamedType, "My_class"), @"
```
mavasani, line 1071: I would instead recommend something like below so we can avoid hard coding magic column/line numbers: https://github.com/dotnet/roslyn-analyzers/blob/03d8e7dbfc82a440aa6b08792693c1a0ea89fe64/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlagTests.cs#L17-L36

Youssef1313: @mavasani Thanks.
Does the "0" after `{|#` represents the index of the specific reported diagnostic?

Youssef1313: I got a test failure when I tried to it that way.

image

I've used `#2` because the `TypeRule` has index 2  in the analyzer.
I've tried also `#0` and the test also failed.

@mavasani

mavasani: No, 0 is just a constant indicating tagged location, so `WithLocation(0)` works. I believe you need:
```csharp
VerifyCS.Diagnostic(IdentifiersShouldNotContainUnderscoresAnalyzer.TypeRule).WithLocation(0).WithArguments(...));
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -1,29 +1,87 @@
...
+            }
+
+            var symbol = model.GetDeclaredSymbol(node, context.CancellationToken);
+            if (symbol == null)
+            {
+                return;
+            }
+
+            string title = MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldNotContainUnderscoresFixTitle;
+            context.RegisterCodeFix(new MyCodeAction(title,
```
Evangelink, line 50: IMO we should do the same as what I am trying to do for the remove `Opt` suffix PR and check that the new symbol name does not already exists before registering the code fix. See https://github.com/dotnet/roslyn-analyzers/pull/3352/files#diff-6fee323e5832f875afec52f98a70ba86R48

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -1,29 +1,87 @@
...
+                    isPreviousUnderscore = true;
+                    continue;
+                }
+                builder.Append(isPreviousUnderscore ? char.ToUpper(c, CultureInfo.InvariantCulture) : c);
+                isPreviousUnderscore = false;
+            }
+            return builder.ToString();
+        }
+
+        private class MyCodeAction : DocumentChangeAction
```
Evangelink, line 79: Please change the base class to be `SolutionChangeAction` as in https://github.com/dotnet/roslyn-analyzers/pull/3352/files#diff-6fee323e5832f875afec52f98a70ba86R65

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -1,29 +1,87 @@
...
+                                        async ct => await RenameWithoutUnderscoresAsync(context.Document, symbol, ct).ConfigureAwait(false),
+                                        equivalenceKey: title),
+                                    context.Diagnostics);
+        }
+
+        private static async Task<Document> RenameWithoutUnderscoresAsync(Document document, ISymbol symbol, CancellationToken cancellationToken)
+        {
+            Solution newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, symbol, GetNewName(symbol.Name), null, cancellationToken).ConfigureAwait(false);
+            return newSolution.GetDocument(document.Id)!;
+        }
```
Evangelink, line 60: You can change this to be a `Task<Solution>` like in https://github.com/dotnet/roslyn-analyzers/pull/3352/files#diff-6fee323e5832f875afec52f98a70ba86R60-R62

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1448,4 +1448,7 @@
   <data name="DoNotIgnoreMethodResultsMessageUserDefinedMethod" xml:space="preserve">
     <value>'{0}' calls '{1}' but does not use the value the method returns. This method is defined as a user-option. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</value>
   </data>
+  <data name="IdentifiersShouldNotContainUnderscoresFixTitle" xml:space="preserve">
```
Evangelink, line 1451:
```suggestion
  <data name="IdentifiersShouldNotContainUnderscoresCodeFixTitle" xml:space="preserve">
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: Great point! @Evangelink.
Symbol name clashing didn't come to my mind 😄 
I'll work on that.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: Sounds reasonable.

Youssef1313:
@Evangelink Thanks for your review. 💯

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -1,29 +1,93 @@
...
+            string title = MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldNotContainUnderscoresCodeFixTitle;
+            context.RegisterCodeFix(new MyCodeAction(title,
+                                        async ct => await RenameWithoutUnderscoresAsync(context.Document, symbol, newName, ct).ConfigureAwait(false),
+                                        equivalenceKey: title),
+                                    context.Diagnostics);
+        }
+
+        private static async Task<Solution> RenameWithoutUnderscoresAsync(Document document, ISymbol symbol, string newName, CancellationToken cancellationToken)
+        {
+            return await Renamer.RenameSymbolAsync(document.Project.Solution, symbol, newName, null, cancellationToken).ConfigureAwait(false);
```
mavasani, line 65: Can we just inline this call and avoid an await?

Youssef1313:
@mavasani, I've added an equivalent VB test to the test I previously added.
But have no idea about what kind of more cases you'd like to see?

Youssef1313:
The current test failure is due to the identifier being a reserved word after renaming. My_class is renaming to MyClass, which is invalid due to [MyClass](https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/me-my-mybase-and-myclass#myclass) being a VB keyword.

/cc @Evangelink Mentioning as your PR (https://github.com/dotnet/roslyn-analyzers/pull/3352) was based on renaming too. Did you handle such case?

Evangelink:
@Youssef1313 I am not handling this case so I have the same issue as you do. I would have expected the symbol renamer to handle this case. I can see a couple of implementations to workaround that but let's wait for @mavasani's feedback.

Youssef1313:
Sorry for the delay on this one. I've updated it locally to put `@newName` (C#) or `[newName]` (VB) when the new name becomes invalid (because it's a keyword). But the test fails (it doesn't enter the fixer at all) with the following error:

```
The markup location '#0' was not found in the input.
```

However, I see that the input has `{|#0:My_class|}`, not sure why it's not found.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -13,5 +14,15 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.ApiDesignGuidelines
     [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
     public sealed class CSharpIdentifiersShouldNotContainUnderscoresFixer : IdentifiersShouldNotContainUnderscoresFixer
     {
+        protected override string GetNewName(string name)
+        {
+            string result = RemoveUnderscores(name);
+            if (result.Length == 0) return string.Empty;
+            if (!SyntaxFacts.IsValidIdentifier(result))
```
mavasani, line 21:
```suggestion
            if (result.Length == 0)
                return string.Empty;

            if (!SyntaxFacts.IsValidIdentifier(result))
```

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -13,5 +14,15 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.ApiDesignGuidelines
...
+            {
+                return $"@{result}";
+            }
+            return result;
```
mavasani, line 25:
```suggestion

            return result;
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +27,67 @@ public sealed override FixAllProvider GetFixAllProvider()
...
+
+            // Make sure there is no symbol with the same name already exists.
+            if (!model.LookupSymbols(context.Span.Start, symbol.ContainingType, newName).IsEmpty)
+            {
+                return;
+            }
+
+            string title = MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldNotContainUnderscoresCodeFixTitle;
+            context.RegisterCodeFix(new MyCodeAction(title,
+                                        async ct => await Renamer.RenameSymbolAsync(context.Document.Project.Solution, symbol, newName, null, ct).ConfigureAwait(false),
```
mavasani, line 61:
```suggestion
                                        ct => Renamer.RenameSymbolAsync(context.Document.Project.Solution, symbol, newName, null, ct),
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -19,11 +27,67 @@ public sealed override FixAllProvider GetFixAllProvider()
...
+                    isPreviousUnderscore = true;
+                    continue;
+                }
+                builder.Append(isPreviousUnderscore ? char.ToUpper(c, CultureInfo.InvariantCulture) : c);
+                isPreviousUnderscore = false;
+            }
+            return builder.ToString();
+        }
+
+        protected abstract string GetNewName(string name);
```
mavasani, line 83: Please move this abstract method near the top of the file, ideally just below the constructor.

src/NetAnalyzers/VisualBasic/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/BasicIdentifiersShouldNotContainUnderscores.Fixer.vb
```diff
@@ -13,5 +14,13 @@ Namespace Microsoft.CodeQuality.VisualBasic.Analyzers.ApiDesignGuidelines
     Public NotInheritable Class BasicIdentifiersShouldNotContainUnderscoresFixer
         Inherits IdentifiersShouldNotContainUnderscoresFixer
 
+        Protected Overrides Function GetNewName(name As String) As String
+            Dim result = RemoveUnderscores(name)
+            If result.Length = 0 Then Return String.Empty
+            If Not SyntaxFacts.IsValidIdentifier(result) Then
```
mavasani, line 20:
```suggestion
            If result.Length = 0 Then
                Return String.Empty
            End If

            If Not SyntaxFacts.IsValidIdentifier(result) Then
```

src/NetAnalyzers/VisualBasic/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/BasicIdentifiersShouldNotContainUnderscores.Fixer.vb
```diff
@@ -13,5 +14,13 @@ Namespace Microsoft.CodeQuality.VisualBasic.Analyzers.ApiDesignGuidelines
...
+                Return $"[{result}]"
+            End If
+            Return result
```
mavasani, line 23:
```suggestion

            Return result
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -1060,6 +1060,32 @@ End Sub
 
         #endregion
 
+        #region "CodeFix Tests"
+        [Fact]
+        public async Task CA1707_ClassNameHasUnderscores_Rename()
```
mavasani, line 1065: Can you add some more tests based on the checks that you perform in the fixer - for example empty length check after RemoveUnderscores, `SyntaxFacts.IsValidIdentifier` check, `LookupSymbols` check.

Evangelink: Also it might be worth having one test for each of the supported symbol types:  `Namespace`, `NamedType`, `Method`, `Property`, `Field`, `Event`. The easiest might be to update all existing tests and add for each the expected code fix.

Evangelink: Also the field case is actually interesting because I think that the code-fix should not result in an upper-case for the first character when the field was having a leading underscore (e.g. `_user` should not result in `User` but in `user`).

Evangelink: Oh also for fields, there might be a question for the static fields, some conventions use the `s_` prefix and you probably don't want to rename with an upper-case `S` (e.g. `s_myList` should not give `sMyList`).

Youssef1313: @Evangelink, I think `_user` and `s_myList` cases shouldn't be even reported from the analyzer side (I haven't checked if they're currently reported or not).

Youssef1313:
@Evangelink @mavasani Any idea on the current test failure?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -13,17 +20,73 @@ public abstract class IdentifiersShouldNotContainUnderscoresFixer : CodeFixProvi
     {
         public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray<string>.Empty;
```
Evangelink, line 21: @Youssef1313 This is your problem! Took me a while to find out what was broken xD

```suggestion
        public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(IdentifiersShouldNotContainUnderscoresAnalyzer.RuleId);
```

Evangelink: @mavasani What about having an analyzer that reports on empty list of diagnostic IDs (for analyzers and fixers)?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Evangelink: I confirm that `_user` and `s_user` cases don't report any diagnostic. I would still add the few tests suggested by @mavasani and also I would consider this part of my first comment:

> The easiest might be to update all existing tests and add for each the expected code fix.

Youssef1313:
The test for renaming Namespace is failing.

The issue is that `var node = root.FindNode(context.Span);` returns an `IdentifierNameSyntax`, not a `MemberDeclarationSyntax`. This causes the call to `model.GetDeclaredSymbol` to return `null`.

One possible solution would be to check the node parent after `GetDeclaredSymbol` returns null. It doesn't seem like an ideal solution to me.

I think the analyzer itself should use AdditionalLocations to report the correct span for MemberDeclarationSyntax, but I'm unsure of how to do so either. Any suggestions?

@Evangelink @mavasani

Youssef1313:
Just noticied that `CA1707_ForFields_CSharp` test is also failing.
The actual behavior is probably correct, but I don't know how to test for it.
The issue is a name that is only an underscore `_`, it generates a diagnostic. But as the code fixer can't do anything with it. The fixed code still has diagnostic. This should be fixed manually. How to test against this?

@Evangelink @mavasani

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -19,7 +19,7 @@ public class IdentifiersShouldNotContainUnderscoresTests
     {
         #region CSharp Tests
         [Fact]
-        public async Task CA1707_ForAssembly_CSharp()
+        public async Task CA1707_ForAssembly_CSharp() // TODO: How to test the code fixer for this?
```
Evangelink, line 22: I have made a couple of tests and I couldn't find anything to access the projects/assemblies afterward.

Tagging @sharwell as he might have some input/ideas.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -107,17 +123,71 @@ public class UnderScoreInNameButNotExternallyVisible_
...
+    }
+}
+");
         }
 
         [Fact, WorkItem(1432, "https://github.com/dotnet/roslyn-analyzers/issues/1432")]
         public async Task CA1707_ForFields_CSharp()
         {
-            await VerifyCS.VerifyAnalyzerAsync(@"
+            await VerifyCS.VerifyCodeFixAsync(@"
```
Evangelink, line 154: You can replace this block with:

```csharp
await new VerifyCS.Test
            {
                TestState =
                {
                    Sources =
                    { @"
public class DoesNotMatter
{
        public const int {|#0:ConstField_|} = 5;
        public static readonly int {|#1:StaticReadOnlyField_|} = 5;

        // No diagnostics for the below
        private string InstanceField_;
        private static string StaticField_;
        public string _field;
        protected string Another_field;
}

public enum DoesNotMatterEnum
{
    {|#2:_EnumWithUnderscore|},
    {|#3:_|}
}

public class C
{
    internal class C2
    {
        public const int ConstField_ = 5;
    }
}
",
                    },
                    ExpectedDiagnostics =
                    {
                        VerifyCS.Diagnostic(IdentifiersShouldNotContainUnderscoresAnalyzer.MemberRule).WithLocation(0).WithArguments("DoesNotMatter.ConstField_"),
                        VerifyCS.Diagnostic(IdentifiersShouldNotContainUnderscoresAnalyzer.MemberRule).WithLocation(1).WithArguments("DoesNotMatter.StaticReadOnlyField_"),
                        VerifyCS.Diagnostic(IdentifiersShouldNotContainUnderscoresAnalyzer.MemberRule).WithLocation(2).WithArguments("DoesNotMatterEnum._EnumWithUnderscore"),
                        VerifyCS.Diagnostic(IdentifiersShouldNotContainUnderscoresAnalyzer.MemberRule).WithLocation(3).WithArguments("DoesNotMatterEnum._"),
                    }
                },
                FixedState =
                {
                    Sources =
                    {
                         @"
public class DoesNotMatter
{
        public const int ConstField = 5;
        public static readonly int StaticReadOnlyField = 5;

        // No diagnostics for the below
        private string InstanceField_;
        private static string StaticField_;
        public string _field;
        protected string Another_field;
}

public enum DoesNotMatterEnum
{
    EnumWithUnderscore,
    {|#0:_|}
}

public class C
{
    internal class C2
    {
        public const int ConstField_ = 5;
    }
}
",
                    },
                    ExpectedDiagnostics =
                    {
                        VerifyCS.Diagnostic(IdentifiersShouldNotContainUnderscoresAnalyzer.MemberRule).WithLocation(0).WithArguments("DoesNotMatterEnum._"),
                    },
                },
            }.RunAsync();
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -61,10 +61,10 @@ public class DoesNotMatter
         [Fact]
         public async Task CA1707_ForNamespace_CSharp()
```
Evangelink, line 62: Code fix doesn't work here. The problem seems to come from https://github.com/dotnet/roslyn-analyzers/pull/3706/files#diff-e68d732a091f5743e18f7949e45cdf47R42. In the case of the namespace that's not `GetDeclaredSymbol` but `GetSymbolInfo` that shall be called here.

Note sure if that's better to call one or the other based-upon the kind/type of the parent node or to always call `GetSymbolInfo` if `GetDeclaredSymbol` is null.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -13,5 +14,18 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.ApiDesignGuidelines
     [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
     public sealed class CSharpIdentifiersShouldNotContainUnderscoresFixer : IdentifiersShouldNotContainUnderscoresFixer
     {
+        protected override string GetNewName(string name)
+        {
+            string result = RemoveUnderscores(name);
+            if (result.Length == 0)
+                return string.Empty;
+
+            if (!SyntaxFacts.IsValidIdentifier(result))
```
Evangelink, line 23: From the tests you have added it seems that this isn't checking that the identifier is not matching a reserved keyword so you will need to tweak it. That's the reason some tests (e.g. CA1707_ForMemberparameters_CSharp) are failing.

Evangelink:
@Youssef1313 Let me know if I haven't provided enough info for you to move forward on this PR.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: @Evangelink Shouldn't the identifier after removing the underscores become invalid if and only if it's a reserved word?

Youssef1313: I'm unsure, but I *think* the "ForMemberparameters" test is failing because the call to `RenameSymbolAsync` on an interface parameter doesn't rename the parameter in an implementing class. Is that the expected behavior from `RenameSymbolAsync`?

Evangelink: Not sure to understand your comment. In one of your test, you have a parameter `int_`, when reaching this line it is now `int` so I was expecting this method to return false resulting in `@int` which isn't the case.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Youssef1313: @Evangelink 
The [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.modelextensions.getdeclaredsymbol?view=roslyn-dotnet) for GetDeclaredSymbol states the following about the node parameter:

> A syntax node that is a declaration. This can be any type derived from MemberDeclarationSyntax, TypeDeclarationSyntax, EnumDeclarationSyntax, **NamespaceDeclarationSyntax**, ParameterSyntax, TypeParameterSyntax, or the alias part of a UsingDirectiveSyntax

Is that a doc error? or I'm missing something?

I'll  wait for @mavasani to determine whether to call the correct method based on the parent node, or always call `GetDeclaredSymbol` first, then `GetSymbolInfo` if the result was null.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
Evangelink: > 
> 
> I'm unsure, but I _think_ the "ForMemberparameters" test is failing because the call to `RenameSymbolAsync` on an interface parameter doesn't rename the parameter in an implementing class. Is that the expected behavior from `RenameSymbolAsync`?

I have seen this behavior in some other PR, that's not what I was expecting but yes it is the case.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -107,17 +123,71 @@ public class UnderScoreInNameButNotExternallyVisible_
...
+        private string InstanceField_;
+        private static string StaticField_;
+        public string _field;
+        protected string Another_field;
+}
+
+public enum DoesNotMatterEnum
+{
+    {|#2:_EnumWithUnderscore|},
+    {|#3:_|}
```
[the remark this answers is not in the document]
Youssef1313: I thought about that at first, but preferred to keep the warning in this case without a code fix.

But after a second thought, it would probably be okay if we don't report on this case. It's really a very edge case. Almost no one will write something like that 😄 

@mavasani, would you be okay if we don't report on this case?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
Youssef1313: > you have a parameter int_, when reaching this line it is now int so I was expecting this method to return false resulting in @int which isn't the case.

```csharp
public class implementI : I
{
    public void M(int int_)
```

@Evangelink You mean the above parameter right?

The problem with it not being changed to `@int` is because the analyzer don't report for it.
The analyzer reported for the parameter in the interface `I`, which was indeed renamed to @int

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Evangelink: Thanks for the doc! I haven't looked at it and you are right. The thing is that `node` is of type `IdentifierNameSyntax` and not `NamespaceDelcarationSyntax`, if you do `GeDeclaredSymbol(node.Parent)` that's ok.

Evangelink: Alright so it's actually a little more complex, sometimes node is `IdentifierNameSyntax` and sometimes it is not (`PropertyDeclarationSyntax`). If it's an identifier you need to access the parent otherwise you should not. You will need to create an abstract method and override it in sub-classes to get the right node.

Evangelink:
@Youssef1313 Do you mind if I push a commit on your PR?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
Evangelink: My bad I thought I was on a different context. You are right the only "problem" is linked to the missing rename for implementations.

Youssef1313:
@Evangelink You can push anytime :)

Evangelink:
@Youssef1313 Don't bother! I thought I found something to unlock but it's not the case...

To sum-up my local changes:
- Update the few broken tests because of the rename not renaming the override/implementations. It'd be nice to create a ticket (I couldn't find any I was supposed to have created).

- Create a new method to retrieve the declaration node (for example `protected abstract SyntaxNode GetDeclarationNode(SyntaxNode node);`), the implementation is rather simple.

C#:
```csharp
protected override SyntaxNode GetDeclarationNode(SyntaxNode node)
    => node.IsKind(SyntaxKind.IdentifierName)
        ? node.Parent
        : node;
```

VB.NET:
```vbnet
Protected Overrides Function GetDeclarationNode(node As SyntaxNode) As SyntaxNode
    If node.IsKind(SyntaxKind.IdentifierName) Then
        Return node.Parent
    Else
        Return node
    End If
End Function
```

- Update the field test (`CA1707_ForFields_CSharp`) as explained here https://github.com/dotnet/roslyn-analyzers/pull/3706#discussion_r467611566

Youssef1313:
@Evangelink Thanks for the great help you provided here. I opened an issue in dotnet/roslyn for the problem with `RenameSymbolAsync`. I'll do your suggestions very soon and push.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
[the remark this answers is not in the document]
mavasani: Yes, I don't really care if we report a diagnostic on `_` or not.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -140,33 +223,36 @@ internal class C2
...
         [Fact, WorkItem(1432, "https://github.com/dotnet/roslyn-analyzers/issues/1432")]
         public async Task CA1707_ForMethods_CSharp()
         {
-            await VerifyCS.VerifyAnalyzerAsync(@"
+            await VerifyCS.VerifyCodeFixAsync(@"
 public class DoesNotMatter
 {
-    public void PublicM1_() { }
+    public void {|#0:PublicM1_|}() { }
     private void PrivateM2_() { } // No diagnostic
```
mavasani, line 242: Interesting - I presume this is due to the fact that this rule only runs on public API surface by default? If so, we should have a follow-up PR to make this rule support [api_surface](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#analyzed-api-surface) option. Implementation would be as simple as invoking `symbol.MatchesConfiguredVisibility()` extension method instead of `symbol.IsExternallyVisible()`

Youssef1313:
What should the expected behavior be for a case like the following (which is far away from a real scenario)?

```csharp
public class My_Class { }
public class My__Class { }
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
```diff
@@ -767,81 +1158,185 @@ End Get
...
     ' No diagnostic
     Public Event E_ As System.EventHandler Implements I1.E_
-    Public Event E2_ As System.EventHandler
+    Public Event {|#3:E2_|} As System.EventHandler
 End Class
 
 Public Class Derives
     Inherits ImplementI1
+
+    Public Shadows Event {|#4:E2_|} As System.EventHandler
```
Youssef1313, line 1264: Looks like this is not getting renamed for some reason.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Evangelink: I have just checked the rule and we already use `MatchesConfiguredVisiblity`, I will close the ticket.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Youssef1313, line 1264: When renaming `E2_` in `Derives` class, the new name "E2" is detected as a clashing, most probably because `E2` already exists in`ImplementI1`. However, we indeed want to rename it. It's marked with Shadows and is meant to shadow the member in `ImplementI1`.

My opinion is that the analyzer don't report for the "Shadows" case, and RenameSymbolAsync gets it updated when https://github.com/dotnet/roslyn/issues/46663 is fixed.

@mavasani @Evangelink What are your opinions?

Youssef1313, line 1264: When renaming `E2_` in `Derives` class, the new name "E2" is detected as a clashing, most probably because `E2` already exists in`ImplementI1`. However, we indeed want to rename it. It's marked with Shadows and is meant to shadow the member in `ImplementI1`.

My opinion is that the analyzer don't report for the "Shadows" case, and RenameSymbolAsync gets it updated when https://github.com/dotnet/roslyn/issues/46663 is fixed.

@mavasani @Evangelink What are your opinions?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
Evangelink: If I am not mistaken this comment is now resolved in line 56, @Youssef1313 could you confirm?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Evangelink: Ping @sharwell

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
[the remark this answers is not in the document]
Evangelink: On the second thought I also think reporting or not are both valid options.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Evangelink: It seems that there is a compiler warning when there is a member shadowing a non-existing member so I think that's fine if the code-fix misses renaming this case.

Tagging @mavasani for confirmation.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
mavasani: I don't know if there is an easy way to do this - also I am not sure how anyone will invoke this fix...

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
mavasani: Yes, I am fine to keep it as-is for now to reduce the scope of this PR. If required, this can be tackled in a follow-up issue/PR.

mavasani:
@Youssef1313 are you able to fix the tests here? Anything else remaining for this PR?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CSharpIdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -13,5 +14,23 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.ApiDesignGuidelines
     [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
     public sealed class CSharpIdentifiersShouldNotContainUnderscoresFixer : IdentifiersShouldNotContainUnderscoresFixer
     {
+        protected override string GetNewName(string name)
+        {
+            string result = RemoveUnderscores(name);
+            if (result.Length == 0)
+                return string.Empty;
```
Evangelink, line 21:
```suggestion
            {
                return string.Empty;
            }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -11,19 +18,76 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
...
+            {
+                if (c == '_')
+                {
+                    isPreviousUnderscore = true;
+                    continue;
+                }
+                builder.Append(isPreviousUnderscore ? char.ToUpper(c, CultureInfo.InvariantCulture) : c);
+                isPreviousUnderscore = false;
+            }
+            return builder.ToString();
```
Evangelink, line 82:
```suggestion

            return builder.ToString();
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscores.Fixer.cs
```diff
@@ -11,19 +18,76 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
...
+            var builder = new StringBuilder();
+            bool isPreviousUnderscore = false;
+            foreach (char c in name)
+            {
+                if (c == '_')
+                {
+                    isPreviousUnderscore = true;
+                    continue;
+                }
+                builder.Append(isPreviousUnderscore ? char.ToUpper(c, CultureInfo.InvariantCulture) : c);
```
Evangelink, line 79:
```suggestion

                builder.Append(isPreviousUnderscore ? char.ToUpper(c, CultureInfo.InvariantCulture) : c);
```

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -813,7 +813,7 @@ By convention, identifier names do not contain the underscore (_) character. Thi
 |Category|Naming|
 |Enabled|True|
 |Severity|Hidden|
-|CodeFix|False|
+|CodeFix|True|
```
Evangelink, line 816: Out of curiosity, is it a manual change?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainUnderscoresTests.cs
Youssef1313: @Evangelink I believe there is a feature request in roslyn to apply codefixes from command-line. When it's implemented, there should be a way to apply this codefix.

Evangelink:
@Youssef1313 Could you fix this little conflict and the few open comments?

mavasani (approved):

mavasani:
Thanks @Youssef1313 and sorry for the delay in getting this merged.

Youssef1313:
> Thanks @Youssef1313 and sorry for the delay in getting this merged.

No problem.

PR #3641, opened 2020-05-18
Analyzer + Fixer: Forward cancellation token to method invocations

carlossanlop:
CA2016: Forward CancellationToken to methods.

Fixes https://github.com/dotnet/runtime/issues/33774

### Summary

The rule should try to identify places where a `CancellationToken` (ct) should have been passed but wasn't. For example, in a method that takes a ct, a method is called that has an overload that takes a ct but a shorter overload that doesn't take a ct was used instead.

### Conditions for positive cases
   - ~~The node to analyze is an awaited invocation, or is the origin invocation of a `ConfigureAwait`.~~
   - The parent method signature has a ct parameter.
   - The invocation is not receiving a ct argument.
   - The invocation method either:
       - Only has one method version, but the signature has one optional ct.
       - Has a method overload with the exact same arguments in the same order, plus one ct parameter at the end.
   - A normal method, a lambda or a nested method receives a ct, ~~and there's an awaited invocation inside it~~.
        
### Conditions for negative cases
   - The parent method signature does not have a ct parameter.
   - ~~The node to analyze is not an awaited invocation.~~
   - The ~~awaited~~ invocation is already receiving a ct argument.
   - The invocation method does not have an overload with the exact same arguments that also receives a ct.
   - The user is using the overload that takes the ct, but is already passing a value. Can be `default` or `default(CancellationToken)` in C#, `Nothing` in VB, or `CancellationToken.None`, or the token from a `CancellationTokenSource`, or something else.
   - If the overload that receives a ct receives more than one ct, do not suggest it.

### Pending to address
- [x] Case where `CancellationToken` is renamed among the usings directives (is it called an alias?).
- [x] Trivia
- [x] Ensure we have unit tests verifying false negatives with the cases not handled below (potential future improvements).
- [X] Appropriate strings to message, title and description.

### Potential future improvements
   - Finding an overload with one ct parameter, but it is not located in the last position. For example, some `ContinueWith` [overloads ](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.continuewith?view=netcore-3.1#System_Threading_Tasks_Task_ContinueWith_System_Action_System_Threading_Tasks_Task__System_Threading_CancellationToken_System_Threading_Tasks_TaskContinuationOptions_System_Threading_Tasks_TaskScheduler_) have the ct in the middle.
   - Passing a named ct in a different position than the last.
[...]

PR #3712, opened 2020-06-04
Use CancellationToken whenever possible

Youssef1313:
By manually going through the code, I found there are places where CancellationToken can be used but not actually used. There are more places than this PR, I opened this to get a review on whether I should continue on that or not.


mavasani:
@Youssef1313 I think it is better to wait for @carlossanlop to merge in https://github.com/dotnet/roslyn-analyzers/pull/3641 and use the added analyzer/fixer to drive and enforce these.

Youssef1313:
The new analyzer will find **some** occurrences, not all.
So, I think I should wait, and after the analyzer is shipped and fixes what it can, I start on the non-captured occurrences in this PR.

What do you think? @mavasani

mavasani:
@Youssef1313 Thanks, that sounds good. I can even merge this PR if you'd like, and you can create a new PR after the analyzer/fixer is merged. Whatever you prefer.

Youssef1313:
I'm okay in both cases :)

mavasani:
I'll merge given this is ready! :)

PR #3714, opened 2020-06-05
Fix CodeMetrics source "LineCount" metric to not include leading and …

mavasani:
…trailing empty newlines.

Fixes customer feedback from https://developercommunity.visualstudio.com/content/problem/1014036/code-metrics-lines-of-source-code-off-by-one.html

### DIFF
--- a/src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
+++ b/src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
@@ -86,11 +86,42 @@ namespace Microsoft.CodeAnalysis.CodeMetrics
                     // Declaration on a single line, we count it as a separate line.
                     delta = 1;
                 }
+                else
+                {
+                    // Ensure that we do not count the leading and trailing emtpy new lines.
+                    var additionalNewLines = Math.Max(0, GetNewlineCount(declSyntax.GetLeadingTrivia(), leading: true) + GetNewlineCount(declSyntax.GetTrailingTrivia(), leading: false) - 1);
+                    delta -= additionalNewLines;
+                }
 
                 linesOfCode += delta;
             }
 
             return linesOfCode;
+
+            static int GetNewlineCount(SyntaxTriviaList trivialList, bool leading)
+            {
+                var triviaParts = trivialList.ToFullString().Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None).ToImmutableArray();
+                return GetNewlineCount(triviaParts, leading);
+
+                static int GetNewlineCount(ImmutableArray<string> triviaParts, bool leading)
+                {
+                    var index = leading ? 0 : triviaParts.Length - 1;
+                    var loopCondition = leading ? LoopConditionForLeading : (Func<int, int, bool>)LoopConditionForTrailing;
+                    var incrementOrDecrement = leading ? 1 : -1;
+                    var count = 0;
+                    for (; loopCondition(index, triviaParts.Length); index += incrementOrDecrement, count++)
+                    {
+                        var trivia = triviaParts[index].Trim();
+                        if (trivia.Length != 0)
+                            break;
+                    }
+
+                    return count;
+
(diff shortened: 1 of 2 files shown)


src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
```diff
@@ -86,11 +86,42 @@ internal static async Task<long> GetLinesOfCodeAsync(ImmutableArray<SyntaxRefere
...
+                    for (; loopCondition(index, triviaParts.Length); index += incrementOrDecrement, count++)
+                    {
+                        var trivia = triviaParts[index].Trim();
+                        if (trivia.Length != 0)
+                            break;
+                    }
+
+                    return count;
+
+                    static bool LoopConditionForLeading(int index, int length) => index < length - 1;
```
genlu, line 121: This is a local function to a local function to a local function :D

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
```diff
@@ -86,11 +86,42 @@ internal static async Task<long> GetLinesOfCodeAsync(ImmutableArray<SyntaxRefere
...
+                static int GetNewlineCount(ImmutableArray<string> triviaParts, bool leading)
+                {
+                    var index = leading ? 0 : triviaParts.Length - 1;
+                    var loopCondition = leading ? LoopConditionForLeading : (Func<int, int, bool>)LoopConditionForTrailing;
+                    var incrementOrDecrement = leading ? 1 : -1;
+                    var count = 0;
+                    for (; loopCondition(index, triviaParts.Length); index += incrementOrDecrement, count++)
+                    {
+                        var trivia = triviaParts[index].Trim();
+                        if (trivia.Length != 0)
```
dotpaul, line 115: >Length [](start = 35, length = 6)

Do comments count as syntax trivia? Didn't see any unit tests with comments in the test code.

mavasani: Yes, good point, let me add a test.

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
mavasani: Yep and all static - I like such static nesting :)

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
genlu, line 115: `if (!string.IsNullOrWhiteSpace(triviaParts[index])) {} `?

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
```diff
@@ -86,11 +86,42 @@ internal static async Task<long> GetLinesOfCodeAsync(ImmutableArray<SyntaxRefere
...
+                            break;
```
genlu, line 116: A couple of questions:
1. Looks like we count comments as part of "Lines"?
2. Do we want to exclude empty newlines in between comments too?

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
mavasani: Yep, helped simplify the loop. Thanks!

src/Utilities/Compiler/CodeMetrics/MetricsHelper.cs
mavasani: I thought about it, but I think it is reasonable to count the start of any non-whitespace trivia as a line belonging to next declaration.

dotpaul: Whatev the answer is, should have a test with blank lines between comments.

mavasani: Sure, added test. We can go both ways here, but this seems to be a rare case, so we can revisit if we get feedback this is problematic.

PR #3729, opened 2020-06-09
Fix UseCountProperly analyzers to avoid invalid cast exception

### LINKED ISSUE #3724: Invalid cast exception in UseCountProperlyAnalyzer
#### Analyzer package

Used the latest `Microsoft.CodeAnalysis.NetAnalyzers` from https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json

#### Package Version

3.3.0-beta1.20305.4 (Latest) OR build the latest source from Roslyn analyzers repo

#### Exception

```
Severity	Code	Description	File	Suppression State	Project	Line	Source	Detail Description
Error	AD0001	Analyzer 'Microsoft.NetCore.Analyzers.Performance.UseCountProperlyAnalyzer' threw an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type 'Microsoft.CodeAnalysis.Operations.CSharpLazyPropertyReferenceOperation' to type 'Microsoft.CodeAnalysis.Operations.IInvocationOperation'.'.
Exception occurred with following context:
Compilation: Microsoft.CodeAnalysis
IOperation: PropertyReference
SyntaxTree: c:\roslyn\src\Compilers\Core\Portable\FileSystem\PathUtilities.cs
SyntaxNode: parentPath.Length [MemberAccessExpressionSyntax]@[20641..20658) (568,110)-(568,127)

System.InvalidCastException: Unable to cast object of type 'Microsoft.CodeAnalysis.Operations.CSharpLazyPropertyReferenceOperation' to type 'Microsoft.CodeAnalysis.Operations.IInvocationOperation'.
   at Microsoft.NetCore.Analyzers.Performance.UseCountProperlyAnalyzer.AnalyzePropertyReference(OperationAnalysisContext context)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteOperationAction>b__59_0(ValueTuple`2 data)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info)
-----

Suppress the following diagnostics to disable this analyzer: CA1827, CA1828, CA1829, CA1836		Active	Microsoft.CodeAnalysis (netcoreapp3.1)	1	IntelliSense	Analyzer 'Microsoft.NetCore.Analyzers.Performance.UseCountProperlyAnalyzer' threw the following exception:
'Exception occurred with following context:
Compilation: Microsoft.CodeAnalysis
IOperation: PropertyReference
SyntaxTree: c:\roslyn\src\Compilers\Core\Portable\FileSystem\PathUtilities.cs
SyntaxNode: parentPath.Length [MemberAccessExpressionSyntax]@[20641..20658) (568,110)-(568,127)

System.InvalidCastException: Unable to cast object of type 'Microsoft.CodeAnalysis.Operations.CSharpLazyPropertyReferenceOperation' to type 'Microsoft.CodeAnalysis.Operations.IInvocationOperation'.
   at Microsoft.NetCore.Analyzers.Performance.UseCountProperlyAnalyzer.AnalyzePropertyReference(OperationAnalysisContext context)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteOperationAction>b__59_0(ValueTuple`2 data)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info)
-----
'.
Error	AD0001	Analyzer 'Microsoft.NetCore.Analyzers.Performance.UseCountProperlyAnalyzer' threw an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type 'Microsoft.CodeAnalysis.Operations.CSharpLazyObjectCreationOperation' to type 'Microsoft.CodeAnalysis.Operations.IInvocationOperation'.'.
Exception occurred with following context:
Compilation: Microsoft.CodeAnalysis
IOperation: PropertyReference
SyntaxTree: c:\roslyn\src\Compilers\Core\Portable\Syntax\SyntaxDiffer.cs
SyntaxNode: after!.GetText().Length [MemberAccessExpressionSyntax]@[3399..3422) (82,47)-(82,70)

System.InvalidCastException: Unable to cast object of type 'Microsoft.CodeAnalysis.Operations.CSharpLazyObjectCreationOperation' to type 'Microsoft.CodeAnalysis.Operations.IInvocationOperation'.
   at Microsoft.NetCore.Analyzers.Performance.UseCountProperlyAnalyzer.AnalyzePropertyReference(OperationAnalysisContext context)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteOperationAction>b__59_0(ValueTuple`2 data)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info)
-----

Suppress the following diagnostics to disable this analyzer: CA1827, CA1828, CA1829, CA1836		Active	Microsoft.CodeAnalysis (netcoreapp3.1)	1	IntelliSense
```

Evangelink:
Fix #3724 and might avoid https://github.com/dotnet/roslyn/pull/44961


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -403,9 +405,9 @@ private static void DetermineReportForInvocationAnalysis(OperationAnalysisContex
             if (!shouldReplaceParent)
             {
                 // Parent expression doesn't met the criteria to be replaced; fallback on analysis for standalone call to Count().
-                if (!isAsync && !hasPredicate)
+                if (!isAsync && !hasPredicate && operation is IInvocationOperation invocation)
                 {
-                    AnalyzeCountInvocationOperation(context, (IInvocationOperation)operation);
+                    AnalyzeCountInvocationOperation(context, invocation);
```
Evangelink, line 410: This change and the one on line 408 are not related to the issue but there was nothing ensuring we couldn't reach this line with a non invocation operation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -228,10 +228,11 @@ static void AddIfNotNull(ImmutableHashSet<IMethodSymbol>.Builder set, IEnumerabl
                 operationKey = OperationEqualsInstance;
             }
             // Analyze argument operation, potentially 0.Equals(obj.Count()).
-            else if (parentOperation is IArgumentOperation argumentOperation)
+            else if (parentOperation is IArgumentOperation argumentOperation &&
+                argumentOperation.GetAncestor<IInvocationOperation>(OperationKind.Invocation) is { } invocation)
```
mavasani, line 232: This would not work correctly if say this is an argument passed to an IObjectCreationOperation, which in turn is passed as an argument to an invocation.

mavasani: We should instead directly check if parent is an invocation.

Evangelink: You are right! Besides, as we are on an argument operation there is no need to walk-up multiple nodes.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -646,5 +646,22 @@ public class SomeClass
...
+public class C
+{
+    public static bool IsChildPath(string parentPath, string childPath)
+    {
+        return (IsDirectorySeparator(childPath[parentPath.Length]) || IsDirectorySeparator(childPath[{|CA1829:parentPath.Count()|}]));
+    }
+
+    public static bool IsDirectorySeparator(char c) => false;
+}");
+        }
```
Youssef1313, line 665: Consider adding parallel VB test

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Text;
```
mavasani, line 6: Is this necessary?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -228,10 +228,11 @@ static void AddIfNotNull(ImmutableHashSet<IMethodSymbol>.Builder set, IEnumerabl
                 operationKey = OperationEqualsInstance;
             }
             // Analyze argument operation, potentially 0.Equals(obj.Count()).
-            else if (parentOperation is IArgumentOperation argumentOperation)
+            else if (parentOperation is IArgumentOperation argumentOperation &&
+                argumentOperation.Parent is IInvocationOperation invocation)
```
Jozkee, line 232: What happens if there is a conversion or parenthesis in the argument? e.g: `0.Equals((uint)obj.Count())`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Jozkee: Since this is only called from `AnalyzeInvocationOperation` which is the method that ensures that `operation` is `IInvocationOperation`, I was assuming this would always be true.

Instead of jumping through IInvocationOperation -> IOperation -> IInvocationOperation, I would rather change the parameter type to IInvocationOperation to avoid this cast but then you will have to use explicit calls on the methods where `operation` is passed as a parameter.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -646,5 +647,37 @@ public class SomeClass
                     .WithLocation(8, 23)
                     .WithArguments(nameof(IReadOnlyCollection<int>.Count)));
         }
+
+        [Fact, WorkItem(3724, "https://github.com/dotnet/roslyn-analyzers/issues/3724")]
+        public static async Task PropertyAccessParentIsNotAlwaysDirectlyTheInvocation()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
```
Jozkee, line 654: Shouldn't this also receive a `fixedSource` argument?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Jozkee: nit: 
```suggestion
                argumentOperation.Parent is IInvocationOperation invocationOperation)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Evangelink: I focused on the issue but you are right I will add the fix.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Evangelink: Nope! I was trying to add a case (with stringbuilder) that would cause my old code to fail but couldn't really make it right.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Evangelink: I will make some test and update the code accordingly. Thanks!

mavasani: > What happens if there is a conversion or parenthesis in the argument?

IArgumentOperation is still the highest node: https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Generated/Operations.Generated.cs#L646

Evangelink: That's interesting, we have a diagnostic raised suggesting to use `.Any()` instead of `Count()` while we should not in this case. `0.Equals((uint)obj.Count())`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -646,5 +647,37 @@ public class SomeClass
...
+
+    public static bool IsDirectorySeparator(char c) => false;
+}");
+
+            await VerifyVB.VerifyAnalyzerAsync(@"
+Imports System.Linq
+
+Public Class C
+    Public Shared Function IsChildPath(ByVal parentPath As String, ByVal childPath As String) As Boolean
+        Return (IsDirectorySeparator(childPath(parentPath.Length)) OrElse IsDirectorySeparator(childPath(parentPath.Count())))
```
Jozkee, line 673:
```suggestion
        Return (IsDirectorySeparator(childPath(parentPath.Length)) OrElse IsDirectorySeparator(childPath({|CA1829:parentPath.Count()|})))
```

This seems to be the reason of CI failure.

Evangelink: Got it! I added the vbnet code without running the tests to ensure everything was alright 🤦

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
Jozkee: Can you share your scenario? I can't repro it.
```cs
[Fact]
public Task CSharpTest()
    => VerifyCS.VerifyAnalyzerAsync(
    @"using System.Linq;

    class C
    {
        private string _str;
        public int GetLength() => 0.Equals((uint)_str.Count()) ? 
            0 : _str.Length;
    }", VerifyCS.Diagnostic(UseCountProperlyAnalyzer.CA1829).WithLocation(6, 46).WithArguments("Length"));
}
```

Jozkee: @Evangelink Are you referring to test `PropertyAccessParentIsNotAlwaysDirectlyTheInvocation`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -646,5 +647,75 @@ public class SomeClass
...
+using System.Collections.Generic;
+using System.Linq;
+
+public class C
+{
+    public static bool IsChildPath(string parentPath, string childPath, IEnumerable<string> obj)
+    {
+        return (IsDirectorySeparator(childPath[parentPath.Length]) ||
+            IsDirectorySeparator(childPath[{|CA1829:parentPath.Count()|}])) ||
+            0.Equals((uint)obj.Count());
```
Jozkee, line 666: You should remove this here and in VB tests, it causes diagnosis of CA1827 that this test does not expect.

The reason for CA1827 being returned and not CA1829 is that `IEnumerable<T>` does not contain a `Count` or `Length` property.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
```diff
@@ -646,5 +647,75 @@ public class SomeClass
...
+
+Public Class C
+    Public Shared Function IsChildPath(ByVal parentPath As String, ByVal childPath As String, ByVal obj As IEnumerable(Of String)) As Boolean
+        Return (IsDirectorySeparator(childPath(parentPath.Length)) OrElse IsDirectorySeparator(childPath({|CA1829:parentPath.Count()|})) OrElse 0.Equals(CUInt(obj.Count()))
+    End Function
+
+    Public Shared Function IsDirectorySeparator(ByVal c As Char) As Boolean
+        Return False
+    End Function
+End Class
```
Youssef1313, line 702: I've a PR in roslyn for an analyzer that removes ByVal. I'm not aware if this test is run on the newer Roslyn version and the new analyzer flagged ByVal to be removed will that cause this test to fail due to reported diagnostics that isn't expected?

Consider removing ByVal if it can cause a problem in the future, or ignore this suggestion if it won't affect the unit test.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Evangelink: This is actually the test failing that I was talking about. `CA1827` suggests to use `.Any()` here but it shouldn't! I cannot use `Any` when I actually need the count value.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Evangelink: Except if the rule is on by default and is a `CS` or `BC` rule it shouldn't impact here. But thank you for the input I will make the change.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Jozkee: I don't get why it shouldn't. As I said earlier, there is no `Count` property in `IEnumerable` nor `IEnumerable<T>`.

Evangelink: Oh wait I am stupid it is actually valid!

Jozkee: Given that `obj` is an `IEnumerable<T>`, I think there is no need to test for `0.Equals(obj.Count())` since that would report CA1827 and these tests focus primarily on rule CA1829.

Up to you if you want to keep it, just make sure to fix the CI issues.

Evangelink: I have removed the test but I think we should create a separate ticket to fix the code/fixer but this test case because it's not handled properly.

Example:
```csharp
[Fact]
public Task ZeroEqualsUIntCastCount_WithPredicate_Diagnostic_CA1827()
	=> VerifyCodeFixAsync<UseCountProperlyAnalyzer, CSharpDoNotUseCountWhenAnyCanBeUsedFixer>(
		testSource: _sourceProvider.GetCodeWithExpression(
			_sourceProvider.GetEqualsUintCastTargetExpressionInvocationCode(0, withPredicate: true, _sourceProvider.MethodName),
			_sourceProvider.ExtensionsNamespace),
		CSharpCodeFixVerifier<UseCountProperlyAnalyzer, CSharpDoNotUseCountWhenAnyCanBeUsedFixer>
		.Diagnostic(UseCountProperlyAnalyzer.s_rule_CA1827)
		.WithArguments("Count")
		.WithSpan(10, 21, 10, 63),
		fixedSource: _sourceProvider.GetCodeWithExpression(
			_sourceProvider.GetFixedExpressionCode(withPredicate: true, negate: true),
			_sourceProvider.ExtensionsNamespace));
```
	
```csharp		
internal string GetEqualsUintCastTargetExpressionInvocationCode(int value, bool withPredicate, string methodName)
	=> $@"{value}.Equals({GetUIntCastCode(GetTargetExpressionCode(withPredicate, methodName))})";

internal abstract string GetUIntCastCode(string expression);
```

```csharp
internal override string GetUIntCastCode(string expression)
	=> $"(uint){expression}";
```

```csharp
internal override string GetUIntCastCode(string expression)
	=> $"CUInt({expression})";
```

Evangelink: And btw, IMO, this test infra is a lot harder to use and reason about that a plain string test.

mavasani:
@Evangelink can you please fix the build? @Jozkee can you review the latest changes?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Evangelink, line 6:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -575,10 +577,10 @@ private static bool IsRightCountComparison(IBinaryOperation binaryOperation, out
             return true;
         }
 
-        private static bool TypeContainsVisibileProperty(OperationAnalysisContext context, ITypeSymbol type, string propertyName, SpecialType propertyType)
-            => TypeContainsVisibileProperty(context, type, propertyName, propertyType, propertyType);
+        private static bool TypeContainsVisibleProperty(OperationAnalysisContext context, ITypeSymbol type, string propertyName, SpecialType propertyType)
```
Evangelink, line 580: Fix typo in method name.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UsePropertyInsteadOfCountMethodWhenAvailableTests.cs
Evangelink: > Could you please file the issue?

Issue #3743

> I found it useful to test the same scenarios for diferent types and for both VB and C# languages since it would avoid test duplication.

I agree that being able to reuse the code with various type is useful but the current infra makes it really hard (for me) to read and debug. We could use strings with interpolated strings where we would inject the type but keeping the overall test string easily readable. Again that's only my opinion and the goal is not to be mean or give a bad criticism.

Evangelink:
@Jozkee, @mavasani Thank you for the reviews and feedback!

mavasani:
Thanks a lot @Evangelink!

PR #3731, opened 2020-06-09
Defensive null check to prevent possible null exception

### LINKED ISSUE #3730: CA1069 throws AD0001
#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

Example: [CA1069](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1069)

#### Repro steps

I had this issue while doing some refactoring in an enum I was creating by removing the value associated to the field. I couldn't reproduce the issue but thought it might still be worth reporting it.

```
Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.EnumShouldNotHaveDuplicatedValues' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.
Parameter name: key'
```

Youssef1313:
Fixes #3730

cc: @mavasani @Evangelink

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
@@ -87,7 +87,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                         .OfType<IFieldReferenceOperation>()
                         .Where(fro => fro.Field.ContainingType.Equals(enumSymbol))
                         .Any();
-                    if (onlyReferencesOneField)
+                    if (onlyReferencesOneField || fieldInitializerOperation.Value.ConstantValue.Value == null)
                     {
                         return;
                     }
@@ -109,7 +109,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     {
                         var fieldSyntax = field.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax();
 
-                        if (fieldSyntax != null)
+                        if (fieldSyntax != null && value != null)
                         {
                             membersByValue.AddOrUpdate(value,
                                 new ConcurrentBag<(SyntaxNode, string)> { (fieldSyntax, field.Name) },


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -87,7 +87,7 @@ public sealed override void Initialize(AnalysisContext context)
                         .OfType<IFieldReferenceOperation>()
                         .Where(fro => fro.Field.ContainingType.Equals(enumSymbol))
                         .Any();
-                    if (onlyReferencesOneField)
+                    if (onlyReferencesOneField || fieldInitializerOperation.Value.ConstantValue.Value == null)
```
mavasani, line 90: Can this be moved the line above where we check `!fieldInitializerOperation.Value.ConstantValue.HasValue`

mavasani: Also tagging @Evangelink for review

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -73,7 +73,8 @@ public sealed override void Initialize(AnalysisContext context)
                     var fieldInitializerOperation = (IFieldInitializerOperation)oc.Operation;
 
                     if (fieldInitializerOperation.InitializedFields.Length != 1 ||
-                        !fieldInitializerOperation.Value.ConstantValue.HasValue)
+                        !fieldInitializerOperation.Value.ConstantValue.HasValue ||
+                        fieldInitializerOperation.Value.ConstantValue.Value == null)
```
Evangelink, line 77: I am wondering if it's possible to be true here and then false when we reach the line 96. If so, we might want to move (or duplicate) those 2 checks just before the`AddOrUpdate`.

mavasani: @Evangelink isn’t it the same field initializer operation used at both places? When do you think the constant value can change?

PR #3733, opened 2020-06-09
Forward CancellationToken

Youssef1313:
This fixes the cases that can't be detected by @carlossanlop's analyzer.
I've left comments in code for things I'm in doubt with.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
@@ -206,7 +206,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                         {
                             var semanticModel = argument.SemanticModel;
 
-                            var symbol = semanticModel.GetSymbolInfo(argument.Value.Syntax, oaContext.CancellationToken).Symbol; // Does it matter here to use csaContext.CancellationToken or oaContext.CancellationToken?
+                            var symbol = semanticModel.GetSymbolInfo(argument.Value.Syntax, oaContext.CancellationToken).Symbol;
 
                             if (symbol != null &&
                                 (symbol.Equals(currentUICultureProperty) ||
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
@@ -206,7 +206,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                         {
                             var semanticModel = argument.SemanticModel;
 
-                            var symbol = semanticModel.GetSymbolInfo(argument.Value.Syntax).Symbol;
+                            var symbol = semanticModel.GetSymbolInfo(argument.Value.Syntax, oaContext.CancellationToken).Symbol; // Does it matter here to use csaContext.CancellationToken or oaContext.CancellationToken?
 
                             if (symbol != null &&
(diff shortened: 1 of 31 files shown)


mavasani:
The token is not a parameter, but a property on the context type which is the parameter.

mavasani:
I think we can enhance this analyzer in future to consider looking for properties of type CancellationToken on (struct) parameter types and have either a separate diagnostic ID or an editorconfig option to control this richer functionality. What do you think @carlossanlop?

src/NetAnalyzers/Core/Microsoft.NetFramework.Analyzers/Helpers/SyntaxNodeHelper.cs
```diff
@@ -123,17 +123,17 @@ public IEnumerable<IMethodSymbol> GetCalleeMethodSymbols(SyntaxNode? node, Seman
             return GetDeclaredSymbol(node, semanticModel) ?? GetReferencedSymbol(node, semanticModel);
         }
 
-        public static ISymbol? GetDeclaredSymbol(SyntaxNode? node, SemanticModel semanticModel)
+        public static ISymbol? GetDeclaredSymbol(SyntaxNode? node, SemanticModel semanticModel) // TODO: Should I add an optional CancellationToken argument here to this public method?
```
mavasani, line 126: I don't think we want to spend any energy on analyzers in `Microsoft.NetFramework.Analyzers` project. Can you revert changes in this file and other files in NetFramework projects?

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -305,7 +305,7 @@ public static bool IsOperationNoneRoot(this IOperation operation)
             if (operation.Instance == null &&
                 operation.Property.ContainingType.IsAnonymousType)
             {
-                var declarationSyntax = operation.Property.ContainingType.DeclaringSyntaxReferences[0].GetSyntax();
+                var declarationSyntax = operation.Property.ContainingType.DeclaringSyntaxReferences[0].GetSyntax(); // TODO: Should the method take CancellationToken and pass it to GetSyntax() ?
```
mavasani, line 308: Possibly, but I would leave that out of this PR and revert the change in this file.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -206,7 +206,7 @@ public override void Initialize(AnalysisContext analysisContext)
                         {
                             var semanticModel = argument.SemanticModel;
 
-                            var symbol = semanticModel.GetSymbolInfo(argument.Value.Syntax).Symbol;
+                            var symbol = semanticModel.GetSymbolInfo(argument.Value.Syntax, oaContext.CancellationToken).Symbol; // Does it matter here to use csaContext.CancellationToken or oaContext.CancellationToken?
```
Youssef1313, line 209: @mavasani, This is the only comment left now.

mavasani: Please always use innermost context's cancellation token. Analyzer driver is free to invoke each analyzer callback with a different cancellation token, and in fact it does happen in the IDE execution. I have seen unhandled operation cancelled exceptions from analyzers not using the correct cancellation token.

carlossanlop:
> I think we can enhance this analyzer in future to consider looking for properties of type CancellationToken on (struct) parameter types and have either a separate diagnostic ID or an editorconfig option to control this richer functionality. What do you think @carlossanlop?

I prefer the option of having a separate diagnostic ID.

Youssef1313:
@mavasani @carlossanlop Anything left for this?

PR #3751, opened 2020-06-15
Update CA1305 to not report on invariant format specifiers

### LINKED ISSUE #3507: Should CA1305 detect invariant formats and not trigger?
#### Analyzer package

#### Package Version

v2.9.8 (latest release)

#### Diagnostic ID

Example: [CA1305](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1305?view=vs-2017#when-to-suppress-warnings)

#### Repro steps

`DateTimeOffset.Now.ToString("o")`

#### Expected behavior

CA1305 should not warn as "o" isn't culture specific:

> Because the "O" or "o" standard format specifier conforms to an international standard, the formatting or parsing operation that uses the specifier always uses the invariant culture and the Gregorian calendar.

#### Actual behavior

CA1305 is triggered and the suppression suggestion applies:

> It is safe to suppress a warning from this rule when it is certain that the default format is the correct choice, and where code maintainability is not an important development priority.

#### See also
This was previously also reported in https://github.com/dotnet/roslyn-analyzers/issues/1949, however the possibility of just _not triggering_ wasn't discussed. Instead of blindly adding `CultureInfo.InvariantCulture` everywhere or when suppressing false positives becomes the norm, the analyzer's added value decreases.

Evangelink:
Handle DateTime.ToString, DateTimeOffset.ToString and TimeSpan.ToString format specifiers

Fix #3507


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnly.cs
```diff
@@ -164,21 +164,12 @@ public KnownTypes(WellKnownTypeProvider wellKnownTypeProvider)
             private static ImmutableHashSet<INamedTypeSymbol> GetIImmutableInterfaces(WellKnownTypeProvider wellKnownTypeProvider)
             {
                 var builder = ImmutableHashSet.CreateBuilder<INamedTypeSymbol>();
-                AddIfNotNull(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsImmutableIImmutableDictionary));
-                AddIfNotNull(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsImmutableIImmutableList));
-                AddIfNotNull(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsImmutableIImmutableQueue));
-                AddIfNotNull(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsImmutableIImmutableSet));
-                AddIfNotNull(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsImmutableIImmutableStack));
+                builder.AddIfNotNull(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsImmutableIImmutableDictionary));
```
Evangelink, line 167: Extracted as helper.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -249,5 +256,42 @@ private static ParameterInfo GetParameterInfo(INamedTypeSymbol type, bool isArra
...
+            if ((dateTimeType != null && targetMethod.ContainingType.Equals(dateTimeType)) ||
+                (dateTimeOffsetType != null && targetMethod.ContainingType.Equals(dateTimeOffsetType)))
+            {
+                return format == "o" || format == "O" || format == "r" || format == "R" || format == "s" || format == "u";
+            }
+
+            if (timeSpanType != null && targetMethod.ContainingType.Equals(timeSpanType))
+            {
+                return format == "c";
+            }
```
Evangelink, line 292: Let me know if you prefer moving this block inside the previous `if`.

Evangelink:
Ping @mavasani

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -249,5 +256,42 @@ private static ParameterInfo GetParameterInfo(INamedTypeSymbol type, bool isArra
...
+                !(invocationOperation.Arguments[0].Value.ConstantValue.Value is string format))
+            {
+                return false;
+            }
+
+            // Handle invariant format specifiers, see https://github.com/dotnet/roslyn-analyzers/issues/3507
+            if ((dateTimeType != null && targetMethod.ContainingType.Equals(dateTimeType)) ||
+                (dateTimeOffsetType != null && targetMethod.ContainingType.Equals(dateTimeOffsetType)))
+            {
+                return format == "o" || format == "O" || format == "r" || format == "R" || format == "s" || format == "u";
```
mavasani, line 286: Nit: write it as a switch expression or define a static readonly hashset/array and do a contains check?

Evangelink:
@mavasani I have addressed your suggestion.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -249,5 +258,42 @@ private static ParameterInfo GetParameterInfo(INamedTypeSymbol type, bool isArra
...
+            }
+
+            if (invariantToStringTypes.Contains(targetMethod.ContainingType))
+            {
+                return true;
+            }
+
+            if (invocationOperation.Arguments.Length != 1 ||
+                !invocationOperation.Arguments[0].Value.ConstantValue.HasValue ||
+                !(invocationOperation.Arguments[0].Value.ConstantValue.Value is string format))
```
mavasani, line 279: `Arguments[0]` does not mean it is the argument for the first parameter. If named arguments are used, such as you specify the argument to second non-first parameter first, then it will be that argument. You want to find argument that corresponds the first parameter here, and would need to iterate through all arguments till to you find one with matching parameter. Can you define an extension method `IArgumentOperation FindArgumentForParameterAtIndex(this ImmutableArray<IArgumentOperation> arguments, int parameterIndex)` and use it here?

mavasani: Ah actually, `invocationOperation.Arguments.Length != 1` check ensures you are good. Ignore my suggestion.

PR #3754, opened 2020-06-16
Remove hack in DisposeAnalysis

### LINKED ISSUE #2152: Ensure DisposeAnalysis does not flag using declarations
Due to https://github.com/dotnet/roslyn/issues/32100, there will no IOperation indicating using declarations in IOperation tree or ControlFlowGraph. We need to make sure we bail out successfully for dispose analysis for using declarations.

Evangelink:
Fix #2152

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs
@@ -13,7 +13,6 @@ using Microsoft.CodeAnalysis.FlowAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DisposeAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PointsToAnalysis;
-using Microsoft.CodeAnalysis.Operations;
 
 namespace Microsoft.NetCore.Analyzers.Runtime
 {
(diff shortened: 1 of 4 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
```diff
@@ -12218,5 +12218,40 @@ End Sub
 End Class",
                 GetBasicResultAt(12, 18, "New A()"));
         }
+
+        [Fact]
+        public async Task Dispose_UsingDeclaration_NoDiagnostic()
```
Evangelink, line 12223: If I got things right from the ticket on roslyn's side these are the extra cases what used to be causing issues.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs
```diff
@@ -141,13 +141,6 @@ public override void Initialize(AnalysisContext context)
                                 return;
                             }
 
-                            if (disposeAnalysisResult.ControlFlowGraph.OriginalOperation.HasAnyOperationDescendant(o => o.Kind == OperationKind.None && !(o.Parent is INameOfOperation)))
```
mavasani, line 144: Can we add a test for nameof case or already exists from when this workaround was added? I think it is the latter, but just confirming.

Evangelink: There is already a test (see #3309) but thanks for making me check!

PR #3755, opened 2020-06-16
Add initial prototype for global control flow graph state based dataflow analysis

mavasani:
Our current DFA framework supports tracking and analyzing dataflow states for entities and/or locations on different control flow paths. However, quite a few analyses do not require tracking such fine grained state, they only require tracking a CFG wide state which changes based on different operations encountered in the CFG. For example, platform checker analyzer described in https://github.com/dotnet/designs/pull/110 requires the analysis to track which OS platforms and versions are valid on any given control flow path.

This PR adds an initial DFA for such a CFG wide state based analysis. It also adds the skeleton analyzer for platform checker analysis (not MEF exported yet) to validate the DFA. There are bunch of follow-up items here:
1. Replace the skeleton analyzer with the real platform checker analyzer
2. Add extensive analyzer tests
3. Naming for added DFA: I have named it as "FlightEnabledAnalysis" based on the term that was used in the original feature request. I plan to rename the analysis and associated types to something more generic in a follow-up change
4. Refactor the DFA to extract out common logic with extensibility points so that more such analyses can be trivially written on top of it.
5. Add comments and doc comments to DFA

The primary goal of sending this PR before the DFA related cleanup items is to unblock @buyaa-n to allow parallel work on items 1. and 2. above for the actual analyzer + unit tests.


RoslynAnalyzers.sln
```diff
@@ -166,7 +166,6 @@ Global
 		src\Utilities\Workspaces\Workspaces.Utilities.projitems*{1c62a2f6-7a44-4030-9d1a-7e7a966736e5}*SharedItemsImports = 5
 		src\Utilities\Compiler\Analyzer.Utilities.projitems*{2a4af3a4-307b-4252-888e-e6546244585a}*SharedItemsImports = 5
 		src\Utilities\Workspaces\Workspaces.Utilities.projitems*{2a4af3a4-307b-4252-888e-e6546244585a}*SharedItemsImports = 5
-		src\Utilities\Compiler\Analyzer.Utilities.projitems*{2a520f3e-8c5d-43fe-aa03-fe5e3c5f23d1}*SharedItemsImports = 5
```
mavasani, line 169: @Evangelink Even I am seeing these now :) Probably best to just update the sln?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/RuntimePlatformCheckAnalyzerTests.cs
```diff
@@ -0,0 +1,785 @@
...
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.InteropServices.RuntimePlatformCheckAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices.UnitTests
+{
+    public class RuntimePlatformCheckAnalyzerTests
+    {
+        private readonly string PlatformCheckApiSource = @"
```
mavasani, line 14: @terrajobst @buyaa-n You may want to review the tests here, which are primarily added to validate and demonstrate the DFA support added here. It uses a dummy skeleton analyzer that just creates a diagnostic with a string representation of the available DFA state at all method invocations. This would be replaced be real analyzer checks for platform specific APIs + real diagnostics.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/FlightEnabledAnalysis/FlightEnabledAnalysis.FlightEnabledDataFlowOperationVisitor.cs
```diff
@@ -0,0 +1,211 @@
...
+
+    internal partial class FlightEnabledAnalysis : ForwardDataFlowAnalysis<FlightEnabledAnalysisData, FlightEnabledAnalysisContext, FlightEnabledAnalysisResult, FlightEnabledBlockAnalysisResult, FlightEnabledAbstractValue>
+    {
+        /// <summary>
+        /// Operation visitor to flow the FlightEnabled values across a given statement in a basic block.
+        /// </summary>
+        private sealed class FlightEnabledDataFlowOperationVisitor
+            : AnalysisEntityDataFlowOperationVisitor<FlightEnabledAnalysisData, FlightEnabledAnalysisContext, FlightEnabledAnalysisResult, FlightEnabledAbstractValue>
+        {
+            private readonly AnalysisEntity _globalEntity;
```
mavasani, line 25: This is the global entity storing CFG wide state, which gets updated for every visited operation in the visitor.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysisData.cs
```diff
@@ -5,6 +5,8 @@
 
 namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
 {
+    // TODO: We should delete this type and use the following instead:
+    //      using TaintedDataAnalysisData = DictionaryAnalysisData<AnalysisEntity, TaintedDataAbstractValue>;
```
mavasani, line 9: @dotpaul I have moved PredicateAnalysis based code to a separate sub-type `PredicateAnalysisEntityDataFlowOperationVisitor`. Given that tainted analysis does not use predicated data, you should be able to delete `TaintedDataAnalysisData` completely and just use this using.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/AnalysisEntityDataFlowOperationVisitor.cs
```diff
@@ -463,43 +465,6 @@ private ImmutableHashSet<AnalysisEntity> GetChildAnalysisEntities(Func<AnalysisE
 
         #endregion
 
-        #region Predicate analysis
```
mavasani, line 466: All code just moved to new sub-type.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/SetAbstractDomain.cs
```diff
@@ -54,11 +54,11 @@ private static ImmutableHashSet<T> MergeOrIntersect(ImmutableHashSet<T> value1,
         {
             if (value1.IsEmpty)
             {
-                return value2;
+                return merge ? value2 : value1;
```
mavasani, line 57: @dotpaul Wonder how this bug was not caught before :) I have a unit test for this now.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/FlightEnabledAnalysis/FlightEnabledAbstractValue.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        public FlightEnabledAbstractValue(FlightEnabledAbstractValue parent1, FlightEnabledAbstractValue parent2)
+            : this(ImmutableHashSet<string>.Empty,
+                   ImmutableHashSet.Create(parent1, parent2),
+                   height: Math.Max(parent1.Height, parent2.Height) + 1,
+                   FlightEnabledAbstractValueKind.Known)
+        {
+        }
+
+        public ImmutableHashSet<string> EnabledFlights { get; }
```
mavasani, line 56: I plan to make the state here extensible instead of using just strings, so we can write additional analyses on top of the added DFA.

RoslynAnalyzers.sln
Evangelink: I am not alone \o/ ! I don't know about the update... I have seen that we end-up with less projects referenced than in some other/older branches so I am wondering if something is broken.

buyaa-n:
> The primary goal of sending this PR before the DFA related cleanup items is to unblock @buyaa-n to allow parallel work on items 1. and 2. above for the actual analyzer + unit tests.

Thank you @mavasani 🙏, I will start working on it

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/SetAbstractDomain.cs
dotpaul: Also fix in 2.9.x?

mavasani: Yes, sure. Would you prefer I retarget the PR to 2.9.x for other changes or are fine with me just cherry-picking this change?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/FlightEnabledAnalysis/FlightEnabledAnalysis.FlightEnabledDataFlowOperationVisitor.cs
dotpaul:
```suggestion
            // This is the global entity storing CFG wide state, which gets updated for every visited operation in the visitor.
            private readonly AnalysisEntity _globalEntity;
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/SetAbstractDomain.cs
dotpaul: I suppose retarget to 2.9.x so that DFA stays as similar as possible between branches.

mavasani: Sure, let me give that a shot.

mavasani: @dotpaul I tried, but there are quite a few API changes from 2.9.x to master, and lot of nullability errors too. I'll instead create a separate PR to port the changes to base DFA types to 2.9.x

mavasani:
@dotpaul Any further feedback? I'll create a separate PR targeting 2.9.x for DFA framework changes. Thanks!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/RuntimePlatformCheckAnalyzerTests.cs
```diff
@@ -0,0 +1,785 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+class Test
+{
+    void M1()
+    {
+        {|#0:M2()|};    // Platform checks:''
+
+        {|#1:Debug.Assert(RuntimeInformationHelper.IsOSPlatformOrLater(OSPlatform.Windows, 3, 0, 0, 0))|};  // Platform checks:'IsOSPlatformOrLater;Windows;3.0.0.0'
```
terrajobst, line 570: @mavasani `Debug.Assert` is good one that I didn't consider. For control flow, are you special casing `Debug.Assert` or are you handling all methods marked with `[DoesNotReturn]`, `[DoesNotReturnIf]`? What about `Environment.FailFast`, `Environment.Exit` etc?
//cc @buyaa-n

dotpaul:
> @dotpaul Any further feedback? I'll create a separate PR targeting 2.9.x for DFA framework changes. Thanks!

Not sure if I completely followed the new DFA, but everything else looks good to me.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/RuntimePlatformCheckAnalyzerTests.cs
mavasani: Right now I just handled Assert methods in Debug and Contracts namespace (latter was already handled for a different analyzer). It would be trivial to enhance this with functionality to handle more cases - @buyaa-n probably good to make a note of these cases in the TODO list.

buyaa-n: Yes, there is many test cases i didn't think about, it even supports InterproceduralAnalysis

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/RuntimePlatformCheckAnalyzerTests.cs
```diff
@@ -0,0 +1,785 @@
...
+        {|#0:M2()|};    // Platform checks:''
+
+        if(RuntimeInformationHelper.IsOSPlatformOrLater(OSPlatform.Windows, 1) &&
+           RuntimeInformationHelper.IsOSPlatformOrLater(OSPlatform.Linux, 1))
+        {
+            {|#1:M2()|};        // Platform checks:'IsOSPlatformOrLater;Linux;1 && IsOSPlatformOrLater;Windows;1'
+        }
+        else
+        {
+            {|#2:M2()|};        // Platform checks:''
```
mavasani, line 318: This seems like a bug in the DFA - this should detect negation of checks above. I'll try to fix it in follow-up.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/RuntimePlatformCheckAnalyzerTests.cs
buyaa-n: > Right now I just handled Assert methods in Debug and Contracts namespace (latter was already handled for a different analyzer). It would be trivial to enhance this with functionality to handle more cases - @buyaa-n probably good to make a note of these cases in the TODO list.

Sure i will add it in my TODO list

mavasani: > it even supports InterproceduralAnalysis

Yes, that comes for free for any analysis based on the repo's framework. However, I have chosen the default to no interprocedural analysis for this DFA analysis for perf reasons. Users can turn it on through editorconfig options if they are making these platform checks in helpers.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/RuntimePlatformCheckAnalyzerTests.cs
buyaa-n: What you think it should to detect by negation of checks above? I can't say because there AND logic used, for me this evaluation is righ

mavasani: I was thinking it should have `!IsOSPlatformOrLater;Linux;1 || !IsOSPlatformOrLater;Windows;1`, but you are right, that is not any valuable data.

mavasani:
Thanks @dotpaul. I am going to merge this in to unblock @buyaa-n, and continue working on pending DFA related items.

PR #3757, opened 2020-06-16
Add non-generic TCS support to CA2247

stephentoub:
A non-generic TaskCompletionSource was added in .NET 5.  Add checking for it in addition to the generic in CA2247.

### DIFF
--- a/src/Utilities/Compiler/WellKnownTypeNames.cs
+++ b/src/Utilities/Compiler/WellKnownTypeNames.cs
@@ -282,7 +282,8 @@ namespace Analyzer.Utilities
         public const string SystemThreadingTasksGenericTask = "System.Threading.Tasks.Task`1";
         public const string SystemThreadingTasksGenericValueTask = "System.Threading.Tasks.ValueTask`1";
         public const string SystemThreadingTasksTask = "System.Threading.Tasks.Task";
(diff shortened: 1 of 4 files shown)


src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -282,7 +282,8 @@ internal static class WellKnownTypeNames
         public const string SystemThreadingTasksGenericTask = "System.Threading.Tasks.Task`1";
         public const string SystemThreadingTasksGenericValueTask = "System.Threading.Tasks.ValueTask`1";
         public const string SystemThreadingTasksTask = "System.Threading.Tasks.Task";
-        public const string SystemThreadingTasksTaskCompletionSource = "System.Threading.Tasks.TaskCompletionSource`1";
+        public const string SystemThreadingTasksTaskCompletionSource = "System.Threading.Tasks.TaskCompletionSource";
+        public const string SystemThreadingTasksTaskCompletionSourceGeneric = "System.Threading.Tasks.TaskCompletionSource`1";
```
Evangelink, line 286: Looking at the other definitions, I think it would be better to use `SystemThreadingTasksGenericTaskCompletionSource`.

stephentoub: Renamed.

PR #3759, opened 2020-06-17
Add support to read in MSBuild property values required by analyzers

mavasani:
Using the new global analyzerconfig feature added by the compiler team, analyzers and end users can now thread in any required MSBuild property values as analyzer options. This change adds the following support:

1. Define a new type `MSBuildPropertyOptionNames` with the names of properties that required to be threaded to analyzers
2. Update the targets file dropped in with our analyzer packages to have the properties defined in the above type to be threaded to the analyzers
3. Add extension method `GetMSBuildPropertyValue` to `AnalyzerOptions` fetch option value inside our analyzers:
```csharp
var optionValue = context.Options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.TargetFramework, ...);
```

### DIFF
--- a/src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
+++ b/src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Analyzer.Utilities
+{
+    /// <summary>
+    /// MSBuild property names that are required to be threaded as analyzer config options.
+    /// </summary>
+    internal static partial class MSBuildPropertyOptionNames
+    {
+        public const string TargetFramework = "TargetFramework";
+    }
+}
--- a/src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
+++ b/src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
@@ -8,5 +8,6 @@ namespace Analyzer.Utilities
     internal static partial class MSBuildPropertyOptionNames
     {
         public const string TargetFramework = "TargetFramework";
+        public const string TargetPlatformMinVersion = "TargetPlatformMinVersion";
     }
 }
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -22,7 +22,7 @@
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
     <MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>$(MicrosoftCodeAnalysisVersion)</MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>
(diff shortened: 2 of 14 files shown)


src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Analyzer.Utilities
+{
+    /// <summary>
+    /// MSBuild property names that are required to be threaded as analyzer config options.
+    /// </summary>
+    internal static partial class MSBuildPropertyOptionNames
+    {
+        public const string TargetFramework = "TargetFramework";
```
mavasani, line 10: @buyaa-n I verified that TargetFramework property is available to analyzers with the newly added `GetMSBuildPropertyValue` API.

For unit tests, you can take approach similar to the below test:
https://github.com/dotnet/roslyn-analyzers/blob/0cbdd68e14722e911d50eee64ee95823b3ffa09c/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs#L891-L908
https://github.com/dotnet/roslyn-analyzers/blob/0cbdd68e14722e911d50eee64ee95823b3ffa09c/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs#L942-L958

You will need to ensure that the analyzer config document text that you pass into the test has following contents, where `property_name` and `property_value` are defined by your test input:

```ini
is_global = true
build_property.<%property_name%> = <%property_value%>
```

We should likely add some test helper so all tests don't have to write up such verbose test case.

buyaa-n, line 10: Another property we will need is `TargetPlatformMinVersion`

buyaa-n: > For unit tests, you can take approach similar to the below test:

Thank you, i will give it try after this PR is merged, before when i played around it i had tested using .editconfig as additional file, which is worked after some update in the options population logic

mavasani: Sure, I have pushed a commit for it. Feel free to add any required properties to this type, and the next package should ensure they are available to the analyzers.

For `TargetPlatformMinVersion`, I notice it is actually not set for default C# classlib project. Is that expected?

image

buyaa-n: Yes it is optional, so i wouldn't expect it defaulted

mavasani: @buyaa-n I decided to do this while I was working in this space - I pushed another commit that should allow you to write unit tests such as:

```csharp
            var test = new VerifyCS.Test()
            {
                TestState =
                {
                    Sources = {source}
                },

                AnalyzerConfigDocument = "build_property.TargetFramework = net452"
            };

            await test.RunAsync();
```

mavasani:
Thanks @buyaa-n!

PR #3764, opened 2020-06-22
Use nameof where appropriate

sharwell:
This change codifies our pattern of property names in code matching property names in the project file.

### DIFF
--- a/src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
+++ b/src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
@@ -5,9 +5,9 @@ namespace Analyzer.Utilities
     /// <summary>
     /// MSBuild property names that are required to be threaded as analyzer config options.
     /// </summary>
-    internal static partial class MSBuildPropertyOptionNames
+    internal static class MSBuildPropertyOptionNames
     {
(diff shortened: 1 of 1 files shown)


src/Utilities/Compiler/Options/MSBuildPropertyOptionNames.cs
```diff
@@ -5,9 +5,9 @@ namespace Analyzer.Utilities
     /// <summary>
     /// MSBuild property names that are required to be threaded as analyzer config options.
     /// </summary>
-    internal static partial class MSBuildPropertyOptionNames
+    internal static class MSBuildPropertyOptionNames
     {
-        public const string TargetFramework = "TargetFramework";
-        public const string TargetPlatformMinVersion = "TargetPlatformMinVersion";
+        public const string TargetFramework = nameof(TargetFramework);
```
mavasani, line 10: @sharwell I explicitly avoided using nameof here so this does not break if someone decides to rename this field to say, `TargetFrameworkPropertyName`. I would be much more comfortable with this change after @buyaa-n adds analyzer functionality that consumes this property + unit tests guarding it.

buyaa-n: Ok, i will add a unit test

Issue #3674, opened 2020-05-25
Add analyzer that warns about redundant fields initialization


marek-safar:
There are cases where C# compiler is not able to optimize out redundant fields initialization. 

```c#
class C
{
    static object o = new object ();
    static bool b = false;
}
```

The simple example is extracted from https://github.com/dotnet/roslyn/issues/43652

mavasani:
This is already implemented as IDE0052 for  private fields that are written but not read. It currently only handles instance fields but can potentially be enhanced. It is also available for CI enforcement in CodeStyle NuGet package.

Issue #627, opened 2015-12-01
Port FxCop rule CA1805: DoNotInitializeUnnecessarily


lgolding:
**Title:** Do not initialize unnecessarily

**Description:**

The common language runtime initializes all fields to their default values before running the constructor. In most cases, initializing a field to its default value in a constructor is redundant, which degrades performance and adds to maintenance costs. One case where it is not redundant occurs when the constructor calls another constructor of the same class or a base class constructor and that constructor initializes the field to a non-default value. In this case, changing the value of the field back to its default value can be appropriate.

**Dependency:** None

**Notes:**

lgolding:
This is no longer necessary (and in fact, it hasn't shipped since Dev10, per @michaelcfanning), because the JITter now eliminates redundant initializations for reference and primitive value types (and sometimes but not always for user-defined value types).

xtqqczze:
This was added in https://github.com/dotnet/roslyn-analyzers/pull/3772, label `Resolution-Won't Fix` can be removed.

PR #3772, opened 2020-06-22
Add CA1805 Do not initialize unnecessarily

stephentoub:
I wanted this for dotnet/runtime, and discovered it used to be an fxcop rule but it explicitly wasn't ported.  So rather than introduce a new ID, I just reused the fxcop one.  Let me know if I should go a different way, or if we really don't want this for some reason.

Fixes https://github.com/dotnet/roslyn-analyzers/issues/3674
cc: @mavasni, @marek-safar, @jeffhandley


mavasani:
@stephentoub The original Won't Fix/Port reasoning seems to be https://github.com/dotnet/roslyn-analyzers/issues/627#issuecomment-164177880. I am not sure if that is valid or not, but given the reasoning is actually based on runtime semantics, I think runtime team is best positioned to make a call. Searching in VS repo, it seems CA1805 was removed even from legacy FxCop package few years ago and the doc page for it also seems non-existent. I am fine with re-use of the ID.

stephentoub:
> I am not sure if that is valid or not

It's valid but incomplete.  The JIT will successfully eliminate the duplication when generating asm in most cases, but the initialization often still contributes to IL assembly size (which is what caused @marek-safar to open the cited issue I believe), as the C# compiler won't eliminate the initialization in all situations, e.g. if any of the initialization isn't to a default value, it'll all be output:
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgZgAIS6BhOgbyrs7oAcpsA3AIYYYDUkjrYAdhjoB9IbgCuMUnQC8dFMQDcHLrwHDRhcZJnzFK4hroAGPZS48+QkWInAIEXJcHKYek0AM38AZxhHZ0M3EzMAeWAAKxgwWQV/FRRbKSVcXCiDV2MPOgBxJWwAEz8AgFZbKphQvIxCzn1OWgZsgFkACgBKNk66AF8qMaA==
On top of that, it's just unnecessary code to maintain.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -0,0 +1,104 @@
...
+                                long l => l == 0,
+                                float f => f == 0f,
+                                double d => d == 0,
+                                _ => false
+                            };
+
+                        // Special-case `null!`/`default!` to not warn about it, as it's often used to suppress nullable warnings on fields.
+                        static bool IsNullSuppressed(IOperation op) =>
+                            op.Syntax?.Parent is PostfixUnaryExpressionSyntax parent &&
+                            parent.Kind() == CodeAnalysis.CSharp.SyntaxKind.SuppressNullableWarningExpression;
```
mavasani, line 97: Will the JIT ensure that the constant value gets inlined here? We should ensure we do not load C# related Roslyn assemblies from a VB project.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                foreach (var field in init.InitializedFields)
+                {
+                    // If a field uses a value known to be its default, warn.
+                    if (!field.IsConst && init.Value != null && UsesKnownDefaultValue(init, field))
+                    {
+                        context.ReportDiagnostic(init.CreateDiagnostic(DefaultRule, field.Name));
+                    }
+
+                    static bool UsesKnownDefaultValue(IFieldInitializerOperation init, IFieldSymbol field)
```
mavasani, line 46: Nit: Can the local function be moved outside the loop?

```csharp
                foreach (var field in init.InitializedFields)
                {
                    // If a field uses a value known to be its default, warn.
                    if (!field.IsConst && init.Value != null && UsesKnownDefaultValue(init, field))
                    {
                        context.ReportDiagnostic(init.CreateDiagnostic(DefaultRule, field.Name));
                    }
                }

                return;

                static bool UsesKnownDefaultValue(IFieldInitializerOperation init, IFieldSymbol field)
                {
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -0,0 +1,104 @@
...
+            analysisContext.RegisterOperationAction(context =>
+            {
+                var init = (IFieldInitializerOperation)context.Operation;
+
+                foreach (var field in init.InitializedFields)
+                {
+                    // If a field uses a value known to be its default, warn.
+                    if (!field.IsConst && init.Value != null && UsesKnownDefaultValue(init, field))
+                    {
+                        context.ReportDiagnostic(init.CreateDiagnostic(DefaultRule, field.Name));
```
mavasani, line 43: Will this generate duplicate diagnostics for VB where an initializer might apply to multiple fields? https://github.com/dotnet/roslyn/blob/730e63429a6691f076f75556a61ef759b0d96743/src/Compilers/VisualBasic/Test/IOperation/IOperation/IOperationTests_ISymbolInitializer.vb#L175-L188

I am wondering if we should just process the first field symbol as the modifiers and initializers would be shared by all fields for such a declaration.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.Fixer.cs
```diff
@@ -0,0 +1,57 @@
...
+            //     private string _value = null;
+            // the node will be for the `= null;` portion.
+            if (root.FindNode(context.Span) is SyntaxNode node)
+            {
+                string title = MicrosoftCodeQualityAnalyzersResources.DoNotInitializeUnnecessarilyFix;
+                context.RegisterCodeFix(
+                    new MyCodeAction(title,
+                    async ct =>
+                    {
+                        // Simply delete the node.
```
mavasani, line 38: VB `As` clause declarations?
```vb
Dim valType As New ValueType
```

I think you would need to write a VB specific sub-type of the fixer that handles it.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarilyTests.cs
```diff
@@ -0,0 +1,169 @@
...
+    protected internal static sbyte SByteValueStatic     [|= 0|];
+    protected static int Int16ValueStatic [|= 0|];
+    private protected static uint UInt16ValueStatic [|= 0|];
+    public static int Int32ValueStatic [|= 0|];
+    internal static uint UInt32ValueStatic [|= 0|];
+    public static long Int64ValueStatic [|= 0|];
+    public static ulong UInt64ValueStatic [|= 0|];
+    static float s_someFloatStatic [|= 0f|];
+    static readonly double s_someDoubleStatic [|= 0|];
+    public static System.DateTime CurrentTimeStatic [|= default|];
```
mavasani, line 103: Add tests for:

1. Declarations with multiple fields, where:
    1. Both have required initializers
    2. One of them has required initializer
    3. Neither have required initializers
    4. Fields with shared initializer (VB As New clause seems to be the only one I think)

2. Simple VB test for sanity

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -0,0 +1,104 @@
...
+                            };
+
+                        // Special-case `null!`/`default!` to not warn about it, as it's often used to suppress nullable warnings on fields.
+                        static bool IsNullSuppressed(IOperation op) =>
+                            op.Syntax?.Parent is PostfixUnaryExpressionSyntax parent &&
+                            parent.Kind() == CodeAnalysis.CSharp.SyntaxKind.SuppressNullableWarningExpression;
+                    }
+                }
+            },
+            OperationKind.FieldInitializer);
```
mavasani, line 101: Should this rule also apply to property and local symbol initializers?

mavasani: Actually not local symbol initializers due to definite assignment requirements, but seems reasonable to include property initializers?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
stephentoub: > Will the JIT ensure that the constant value gets inlined here?

Won't even make it to the JIT; the C# compiler just emits the value being compared against as a ldc.i4, e.g.
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgZgAIS6BhOgbyrs4fuAglzoBZABQARAIYBPAPIAzAOowYAazoB3JcoCUbDl32EA7Os10AvGboSZCzQDpFAEwB2MAM6OpAbj2cAvlR+QA

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
stephentoub: Sure

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
stephentoub: > I am wondering if we should just process the first field symbol

Ok

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.Fixer.cs
stephentoub: I'll make the fixer C# only for now.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
stephentoub: Makes sense.  Added.

stephentoub:
Thanks for the good feedback, @mavasani. Addressed.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.Fixer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
```
mavasani, line 11: Given the fixer is now C# specific and uses C# syntax APIs, we should move it to the C# layer: https://github.com/dotnet/roslyn-analyzers/tree/0fcf66bca3d6748be1fa4df89b4778a79c90b080/src/NetAnalyzers/CSharp

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1215,10 +1215,13 @@
...
+    <value>The common language runtime initializes all fields to their default values before running the constructor. In most cases, initializing a field to its default value in a constructor is redundant, which adds to maintenance costs and may degrade performance (such as with increased assembly size). One case where it is not redundant occurs when the constructor calls another constructor of the same class or a base class constructor and that constructor initializes the field to a non-default value. In this case, changing the value of the field back to its default value can be appropriate.</value>
   </data>
   <data name="DoNotInitializeUnnecessarilyMessage" xml:space="preserve">
-    <value>Do not initialize unnecessarily</value>
+    <value>Member '{0}' is explicitly initialized to its default value.</value>
+  </data>
+  <data name="DoNotInitializeUnnecessarilyFix" xml:space="preserve">
+    <value>Remove the unnecessary assignment.</value>
```
Evangelink, line 1224: I might be mistaken but I think we decided not to have trailing period for code fix titles, am I right @mavasani ?

stephentoub: As it's currently inconsistent, I'm going to go ahead and merge as-is.  If the right answer is to remove the periods, we should have a cleanup PR that removes them from all of the messages consistently.

PR #3774, opened 2020-06-23
Add a suppressor for 'typeof(T).FullName'

sharwell:


src/Roslyn.Diagnostics.Analyzers/Core/NamedTypeFullNameNotNullSuppressor.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (!(diagnostic.Location.SourceTree is { } tree))
+                {
+                    continue;
+                }
+
+                var root = tree.GetRoot(context.CancellationToken);
+                var node = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+                var semanticModel = context.GetSemanticModel(tree);
+                var operation = semanticModel.GetOperation(node, context.CancellationToken);
+                if (operation is IPropertyReferenceOperation { Property: { Name: nameof(Type.FullName) }, Instance: ITypeOfOperation { } })
```
mavasani, line 46: Is there a plan to change the compiler to detect this and avoid firing false positives? Why would such a suppression be specific to consumers of Roslyn.Diagnostics.Analyzers as opposed to everyone that enables nullable reference type features.

mavasani: Tagging @jcouv

sharwell: > Is there a plan to change the compiler to detect this and avoid firing false positives?

Not that I'm aware of. According to the signatures in the API, the compiler is correct in reporting the diagnostic.

> Why would such a suppression be specific to consumers of Roslyn.Diagnostics.Analyzers as opposed to everyone that enables nullable reference type features?

It's an experiment in using suppressions to target common patterns that can't be covered by the NRT feature itself. If it proves valuable and other users ask for it, we could move it to one of the shared packages.

stephentoub: > According to the signatures in the API, the compiler is correct in reporting the diagnostic.

And we lack the expressivity necessary to differentiate these cases in the API. FullName can return null, it just won't with this calling pattern.

mavasani: Ah, thanks for the explanations. I still hope all customers eventually get benefit of these suppressions by default, not just the ones using Roslyn.Diagnostics.Analyzers.

src/Roslyn.Diagnostics.Analyzers/Core/NamedTypeFullNameNotNullSuppressor.cs
```diff
@@ -0,0 +1,64 @@
...
+using System;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Roslyn.Diagnostics.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class NamedTypeFullNameNotNullSuppressor : DiagnosticSuppressor
```
mavasani, line 12: Probably a good time to start building a test framework for suppressors :-)

src/Roslyn.Diagnostics.Analyzers/Core/NamedTypeFullNameNotNullSuppressor.cs
sharwell: @stephentoub How should we track proposals to roll these out more generally?

src/Roslyn.Diagnostics.Analyzers/Core/NamedTypeFullNameNotNullSuppressor.cs
sharwell: Testing is blocked on https://github.com/dotnet/roslyn/issues/41584

jaredpar:
I strongly believe we should not be doing this as a Roslyn analyzer, or even an analyzer in general. There is nothing contextual about this particular nullable warning: it applies to 100% of C# code. If this should be suppressed then it should be done in the compiler, the API annotation should be changed or we should do nothing. 

Having an analyzer that suppresses this warning just makes nullability harder to understand for customers. It means that in one environment a piece of nullable code will produce a set of warnings and in another it will not. This can leave customers scratching their head trying to figure out what is wrong with nullable, why aren't the annotations working here? What indication do they have that **special** nullability rules apply here?

src/Roslyn.Diagnostics.Analyzers/Core/NamedTypeFullNameNotNullSuppressor.cs
jaredpar: > Is there a plan to change the compiler to detect this and avoid firing false positives?

Hard to have a plan when we're not really involved in the discussions.

mavasani:
@jaredpar Agree with your stance. Note that this was done as a temporary experiment to see if the experience is reasonable and then do one of the following (after discussions with compiler and runtime teams):
1. Revert this suppressor and move the logic into the compiler, possibly with annotations on the APIs to aid analysis.
2. If that is not feasible, move this suppressor into a package that is easily available for everyone enabling nullability.

1. is obviously the preferred approach. @sharwell’s attempt to enable this suppressor hit another roadblock in terms of potential local build time regression for Roslyn.sln: https://github.com/dotnet/roslyn/pull/45759#discussion_r451137473, which is definitely undesirable and likely renders any suppressor based approach undesirable on multiple fronts. We should consider reverting this suppressor and work towards approach 1.

jaredpar:
I think there is still the issue of whether or not this should be suppressed at all. As @tannergooding pointed out in [the roslyn issue](https://github.com/dotnet/roslyn/issues/45774) there are many reflection APIs that have this exact behavior: almost never return `null` but can in very specific circumstances. Why do we want to special case this API but none of the others?

I'm also skeptical this is the type of change the compiler should be making on its own. The compiler relies on the APIs for pretty much all of it's nullability information and then provides a standard analysis on top of that information. This suppression would be the compiler specifically overriding the annotations and essentially saying "we know better". What happens in the future if the runtime chooses to extend the situations where `null` can be returned? We'd be in a bit of a pickle then.

There are cases where the compiler has applied special casing to APIs in runtime, most notable of which is `IEquatable<T>`. Those are limited and generally done as a mutual decision between the two teams that it's the best for the ecosystem. So for me the next step would be getting buy off from the runtime team that they feel this is a contract a) they want to support and b) has sufficient use cases in real code that we should look to special casing it.

PR #3778, opened 2020-06-24
Fix CA1805 handling of nullable attributes

stephentoub:
Or, rather, punt.  If fields are annotated with nullable attributes, it's possible you end up in a situation like:
```C#
[AllowNull] private T _value = default;
```
where for whatever reason (https://github.com/dotnet/roslyn/issues/37138), the compiler won't warn on the above, but will warn on:
```C#
[AllowNull] private T _value;
```

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
@@ -37,7 +37,11 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
             {
                 var init = (IFieldInitializerOperation)context.Operation;
                 IFieldSymbol? field = init.InitializedFields.FirstOrDefault();
-                if (field != null && !field.IsConst && init.Value != null && UsesKnownDefaultValue(init.Value, field.Type))
+                if (field != null &&
+                    !field.IsConst &&
+                    init.Value != null &&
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -37,7 +37,11 @@ public override void Initialize(AnalysisContext analysisContext)
             {
                 var init = (IFieldInitializerOperation)context.Operation;
                 IFieldSymbol? field = init.InitializedFields.FirstOrDefault();
-                if (field != null && !field.IsConst && init.Value != null && UsesKnownDefaultValue(init.Value, field.Type))
+                if (field != null &&
+                    !field.IsConst &&
+                    init.Value != null &&
+                    field.GetAttributes().Length == 0 && // in case of attributes that impact nullability analysis
```
mavasani, line 43: We can consider to bail out only if attribute has "Null" within its name? I am fine with current conservative approach as well.

stephentoub: Hadn't concerned the naming aspect.  It's a good suggestion.  At the same time, I think the conservative place is a good place to start here: worst case, we miss a few.  And the rule is only "info" by default, which I think (correct me if I'm wrong) gives more flexibility in the future should we find more opportunities to warn.

mavasani: Yep, starting conservative is fine.

PR #3780, opened 2020-06-24
Fix CA1062 for untracked interprocedural calls

### LINKED ISSUE #2919: CA1062 forgets non-null status from ValidatedNotNull after call to function containing delegate
#### Analyzer package

#### Package Version

v2.9.6 (Latest)

#### Diagnostic ID

[CA1062](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1062-validate-arguments-of-public-methods?view=vs-2019)

#### Visual Studio version

16.3.3 (latest)

#### Repro steps

In the following minimized repro, accessing `foo` in `Test()` after the call to `Bar()` causes a false positive of CA1062 despite `Bar()` not having any way to access `foo`:
```c#
using System;

namespace ReproCA1062
{
    public static class Repro
    {
        public static void Test(string foo)
        {
            NotNull(foo, nameof(foo));
            int x = foo.Length; // no warning
            Bar();
            x = foo.Length; // CA1062 on foo
        }

        public static void Bar()
        {
            Action<int> hmm = x => { };
            hmm(0);
        }

        public static void NotNull([ValidatedNotNull] object param, string paramName)
        {
            if (param == null)
            {
            }
        }
    }

    [System.AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = true)]
    public sealed class ValidatedNotNullAttribute : Attribute { }
}
```

Note that the bug does not appear if I use a local null-check:
```c#
public static void Test(string foo)
{
    if (foo == null)
    {
        throw new ArgumentNullException(nameof(foo));
    }
    int x = foo.Length; // no warning
    Bar();
    x = foo.Length; // no warning
}
```

mavasani:
Fixes #2919

Switch parameter validation analysis from pessimistic to optimistic - if we encounter a call that cannot be analyzed interprocedurally, we don't reset analysis data, which would nullify all the prior validations in the method.

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
@@ -4577,7 +4577,7 @@ public class C<T> where T : class
     {
         o = o ?? new object();
         AllocateSlow(s);
-        var x = s.Length;
+        var x = [|s|].Length;
     }
 
     private T AllocateSlow(string s)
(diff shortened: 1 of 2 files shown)


mavasani:
Ping.. trivial review.

mavasani:
Thanks @dotpaul!

PR #3781, opened 2020-06-25
Handle IsPattern null checks for CA1062

### LINKED ISSUE #3049: CA1062 - False alarm for switch pattern matching
Migrated from [#3847](https://github.com/MicrosoftDocs/visualstudio-docs/issues/3847)

The following example triggers CA1062 mistakenly:

```csharp
[SuppressMessage("Design", "CA1062:Validate arguments of public methods", Justification = "False alarm: see the null case")]
public static int IndexOf<T>(this IEnumerable<T> source, T element)
{
    switch (source)
    {
        case null:
            throw new ArgumentNullException(nameof(source), Res.ArgumentNull);
        case IList<T> genericList:
            return genericList.IndexOf(element);
        case IList list:
            return list.IndexOf(element);
        default:
            var comparer = ComparerHelper<T>.EqualityComparer;
            var index = 0;
            foreach (T item in source)
            {
                if (comparer.Equals(item, element))
                    return index;
                index++;
            }

            return -1;
    }
}
```

mavasani:
Fixes #3049


mavasani:
Ping.. trivial review.

PR #3783, opened 2020-06-25
Fix dataflow predicate analysis for pattern operations

### LINKED ISSUE #2185: Handle DiscardPattern operation kind in flow-analysis package
IDiscardPatternOperation, OperationKind.DiscardPattern and IDeclarationPatternOperation.MatchedType were added in Microsoft.CodeAnalysis 3.0. We had to disable few asserts in flow analysis package to allow for these in features/dataflow branch targeting 2.9.0 bits. This issue tracks re-enabling those asserts, building Roslyn.sln to come up with minimal repro for these asserts and fix and add unit tests.

### LINKED ISSUE #3716: CA1062 False positive
#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

[CA1062](https://docs.microsoft.com/visualstudio/code-quality/ca1062)

#### Repro steps

I've a small abstract class that represents a chess piece, with the following Equals implementation.

image

#### Expected behavior

If I'm not missing something, there shouldn't be a warning.

1. Assume the passed `obj` is null.
1. `obj is Piece p` should be false.
1. `!(obj is Piece p)` should be true.
1. The right of the `||` operator won't be evaluated.

#### Actual behavior

A warning is issued.

mavasani:
Fixes #3716

Also handle the new discard and recursive patterns correctly.
Fixes #2185

### DIFF
--- a/src/Utilities/Compiler/Extensions/IOperationExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IOperationExtensions.cs
@@ -512,6 +512,10 @@ namespace Analyzer.Utilities.Extensions
         {
             return pattern switch
             {
+#if CODEANALYSIS_V3_OR_BETTER
+                IDeclarationPatternOperation declarationPattern => declarationPattern.MatchedType,
+                IRecursivePatternOperation recursivePattern => recursivePattern.MatchedType,
+#else
                 IDeclarationPatternOperation declarationPattern => declarationPattern.DeclaredSymbol switch
                 {
                     ILocalSymbol local => local.Type,
@@ -520,7 +524,7 @@ namespace Analyzer.Utilities.Extensions
 
                     _ => null,
                 },
-
+#endif
                 IConstantPatternOperation constantPattern => constantPattern.Value.Type,
 
                 _ => null,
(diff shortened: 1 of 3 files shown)


src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -512,6 +512,10 @@ public static bool IsWithinLambdaOrLocalFunction(this IOperation operation)
         {
             return pattern switch
             {
+#if CODEANALYSIS_V3_OR_BETTER
```
dotpaul, line 515: Did you mean to target 2.9.x?

mavasani: Ah yes, thanks!

mavasani: Actually, no. These new IOperation APIs have been introduced in Microsoft.CodeAnalysis v3.3

dotpaul: Maybe it should be named CODEANALYSIS_V3_3_OR_BETTER? Dunno where that's defined.

mavasani: Sure, I can clean that up in a separate PR once this goes in. That works?

mavasani: @dotpaul Any further feedback?

mavasani:
Thanks @dotpaul

PR #3785, opened 2020-06-25
Fix CA1805 assignment of default to nullable

stephentoub:
One more case found from dotnet/runtime.  Code like:
```C#
private int? _value = 0;
```
shouldn't warn, but it currently does.

This is the last issue discovered in running the rule over dotnet/runtime:
https://github.com/dotnet/runtime/pull/38410


stephentoub:
Thanks, @mavasani.

PR #3788, opened 2020-06-26
Improve DFA performance

### LINKED ISSUE #3738: Build time > 10 minutes for a simple project when code analysis is on.
#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

Build time > 10 minutes for a simple project when code analysis is on.

#### Repro steps

1. Clone the repository from following URL
2. [Code Analysis Build Issue Repro](https://github.com/himalpatel81/CodeAnalysisIssue)
3. Open the solution in Visual Studio and Compile. (building the project outside Visual studio also gives the same result)
4. Notice the build time. It will be more than 10 minutes.

As per the update given on [this Link](https://github.com/dotnet/roslyn/issues/44777), the data flow analysis takes all the time ([this method](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowAnalysis.cs#L35) with all its recursive paths).

#### Expected behavior
It should compile the project in appropriate time.
#### Actual behavior
It takes more than 10 minutes to build a single project with a single class and a single method.

mavasani:
Fixes #3738 - build time drops for the repro from 10+ minutes to few seconds.

Add PointsToAnalysisKind enum to allow partial/complete points to analysis. Complete analysis mode is same as implemented today. Partial analysis mode skips tracking points to values for fields and properties, which is one of the primary cause of large points to data maps and slow DFA execution, and would be used by majority of non-security rules that do not need tracking fields/properties. Security rules need complete analysis and should be unaffected. We would still need to investigate in performance improvements for complete analysis mode, but the hope is this change should still improve the overall performance for the entire analyzers package and address all the performance reports for the same.

**Real world testing:** I am going to test the package with these changes against couple of repos + also verify if it addresses recent performance bugs filed on the repo for slow DFA execution.


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidDeadConditionalCode.cs
```diff
@@ -90,7 +91,9 @@ public override void Initialize(AnalysisContext context)
                             var cfg = operationBlockContext.GetControlFlowGraph(operationRoot);
                             var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(operationBlockContext.Compilation);
                             var valueContentAnalysisResult = ValueContentAnalysis.TryGetOrComputeResult(cfg, owningSymbol, wellKnownTypeProvider,
-                                    operationBlockContext.Options, AlwaysTrueFalseOrNullRule, operationBlockContext.CancellationToken,
+                                    operationBlockContext.Options, AlwaysTrueFalseOrNullRule,
+                                    PointsToAnalysisKind.Complete,
```
mavasani, line 95: This can potentially be switched to partial analysis in future, if required. Currently, this rule is disabled by default.

PR #3789, opened 2020-06-26
Improve DFA performance

mavasani:
This is the backport of #3788 to 2.9.x to help avoid ongoing merge conflict pain, with following differences:

1. Switch all DFA rules to PointsToAnalysisKind.Complete
2. No new tests, because tests added in #3788 rely on new test framework and test helpers added in master.

In future, if required, we can backport the performance fixes to the DFA rules that are switched to partial analysis in master with #3788.


mavasani:
@dotpaul can you review? Tests are now green!

Issue #3786, opened 2020-06-25
CA2016 bugs - Diagnostics triggered in edge cases


carlossanlop:
The [dotnet/runtime PR](https://github.com/dotnet/runtime/pull/37607) to fix CA2016 in the repo source code found some edge cases that need to be fixed:

### Diagnostic triggered unexpectedly when there is an extra middle parameter

See https://github.com/dotnet/runtime/pull/37607/files#r445874631

The user was originally callin `TryAdd(T item)`, and the analyzer suggested to change it to use the one that takes a token. The expected behavior was that the user would not get a suggestion because the number of parameters differs by more than one between the two overloads.

```cs
        public bool TryAdd(T item)
        {...}
        public bool TryAdd(T item, TimeSpan timeout)
        {...}
        public bool TryAdd(T item, int millisecondsTimeout)
        {...}
        public bool TryAdd(T item, int millisecondsTimeout, CancellationToken cancellationToken)
        {...}
```

Similar thing happened in that same file with `TryTake`.

---

### Diagnostic triggered when an invocation inside a lambda that does not pass a ct, but the parent method does

See https://github.com/dotnet/runtime/pull/37607/files#r446107092

If the invocation that accepts an ct lives inside a lambda that does not pass a ct, do not offer a diagnostic.

```cs
using System;
using System.Threading;
public static class Extensions
{
    public static void Extension(this bool b, Action<int> action) { }
    public static void MyMethod(this int i, CancellationToken c = default) { }
}
class C
{
    public void M(CancellationToken ct)
    {
        bool b = false;
        b.Extension((j) =>
        {
            j.MyMethod();
        });
    }
}
```

---

### Static local method should not get a ct passed

```cs
using System;
using System.Threading;
class C
{
    public static void MyMethod(int i, CancellationToken c = default) {}
    public void M(CancellationToken ct)
    {
        LocalStaticMethod();
        static void LocalStaticMethod()
        {
            MyMethod(5);
        }
    }
}
```

PR #3791, opened 2020-06-26
Fix CA2016 bugs - diagnostic triggered in edge cases

carlossanlop:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/3786

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
@@ -519,6 +519,36 @@ class C
             return CS8VerifyAnalyzerAsync(originalCode);
         }
 
+        [Fact]
+        [WorkItem(3786, "https://github.com/dotnet/roslyn-analyzers/issues/3786")]
+        public Task CS_AnalyzerOnlyDiagnostic_LocalMethod_InsideOf_StaticLocalMethod_TokenInTopParent()
+        {
+            // Local static functions are available in C# >= 8.0
+            // The user should fix convert the static local method into a non-static local method,
+            // or pass `default` or `CancellationToken.None` manually
+            string originalCode = @"
+using System;
+using System.Threading;
+class C
+{
+    public static void MyMethod(int i, CancellationToken c = default) {}
+    public void M(CancellationToken ct)
+    {
+        LocalStaticMethod();
+        static void LocalStaticMethod()
+        {
+            LocalMethod();
+            void LocalMethod()
+            {
+                [|MyMethod|](5);
+            }
+        }
+    }
+}
+            ";
+            return CS8VerifyAnalyzerAsync(originalCode);
+        }
+
         #endregion
 
         #region Diagnostics with fix = C#
@@ -2149,6 +2179,54 @@ class C
 }
             ";
             return VerifyCS.VerifyCodeFixAsync(originalCode, fixedCode);
+        }
+
+        [Fact]
+        [WorkItem(3786, "https://github.com/dotnet/roslyn-analyzers/issues/3786")]
+        public Task CS_Diagnostic_LocalMethod_InsideOf_StaticLocalMethodPassingToken()
+        {
+            // Local static functions are available in C# >= 8.0
+            string originalCode = @"
+using System;
+using System.Threading;
+class C
+{
+    public static void MyMethod(int i, CancellationToken c = default) {}
+    public void M(CancellationToken c)
+    {
+        LocalStaticMethod(c);
+        static void LocalStaticMethod(CancellationToken ct)
+        {
+            LocalMethod();
+            void LocalMethod()
+            {
+                [|MyMethod|](5);
+            }
+        }
+    }
+}
+            ";
+            string fixedCode = @"
+using System;
+using System.Threading;
+class C
+{
+    public static void MyMethod(int i, CancellationToken c = default) {}
+    public void M(CancellationToken c)
+    {
+        LocalStaticMethod(c);
+        static void LocalStaticMethod(CancellationToken ct)
+        {
+            LocalMethod();
+            void LocalMethod()
+            {
+                MyMethod(5, ct);
+            }
+        }
+    }
+}
+            ";
+            return CS8VerifyCodeFixAsync(originalCode, fixedCode);
         }
 
         #endregion
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
@@ -3,7 +3,9 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Testing;
+using Test.Utilities;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
     Microsoft.NetCore.CSharp.Analyzers.Runtime.CSharpForwardCancellationTokenToInvocationsAnalyzer,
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -1881,61 +1934,6 @@ class O
             return CS8VerifyCodeFixAsync(originalCode, fixedCode);
         }
 
-        [Fact]
-        public Task CS_Diagnostic_LambdaAndExtensionMethod()
```
carlossanlop, line 1885: This test method was converted into a "NoDiagnostic" method.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -2002,63 +2000,94 @@ async void M(CancellationToken ct)
             return VerifyCS.VerifyCodeFixAsync(originalCode, fixedCode);
         }
 
-        [Theory]
-        [InlineData("CancellationToken c", "", "", "")]
-        [InlineData("", "CancellationToken c", "", "")]
-        [InlineData("", "", "<CancellationToken>", "c")]
-        public Task CS_Diagnostic_MultiNesting(string topMethodParam, string localMethodParam, string extensionTypeParam, string extensionArg)
+        [Fact]
```
carlossanlop, line 2003: I split the "MultiNesting" methods into two separate test methods:
- One that verifies the ct being passed from the parent local method.
- One that verifies the ct being passed from the top method, skipping the local method.

The case for the lambda is no longer offering a diagnostic.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -3988,31 +3994,22 @@ End Class
             return VerifyVB.VerifyCodeFixAsync(originalCode, fixedCode);
         }
 
-        [Theory]
-        [InlineData("c As CancellationToken", "", "")]
-        [InlineData("", "(Of CancellationToken)", "c")]
-        public Task VB_Diagnostic_MultiNesting(string topMethodParam, string extensionTypeParam, string extensionArg)
+        [Fact]
+        public Task VB_Diagnostic_MultiNesting_TopMethod()
```
carlossanlop, line 3998: Local methods are not available in VB, but the "MultiNesting" method was still relevant because lock block bodies exist there, so we need to look for the parent method ct.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -357,6 +357,59 @@ public static class Extensions
             return CS8VerifyAnalyzerAsync(originalCode);
         }
 
+        [Fact]
```
mavasani, line 360: Nit: Add `WorkItem` attribute for all added tests?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -2002,63 +2000,94 @@ async void M(CancellationToken ct)
...
             bool b = false;
-            b.Extension(({extensionArg}) =>
+            lock (lockingObject)
             {{
-                lock (lockingObject)
-                {{
-                    [|TokenMethod|]();
-                }}
-            }});
+                TokenMethod(c);
```
mavasani, line 2041: Do we have a test where this is a static local function? I presume your code fix should bail out for that, and we only get a diagnostic?

carlossanlop: That's a very good edge case.
Let me check. If I don't find such test, I'll add it.
I agree, it should bail out. If it doesn't, I'll fix it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -357,6 +357,59 @@ public static class Extensions
...
+    public static void MyMethod(this int i, CancellationToken c = default)
+    {
+    }
+}
+class C
+{
+    public void M(CancellationToken ct)
+    {
+        bool b = false;
+        b.Extension((j) =>
```
AArnott, line 403: Can you add an anonymous delegate version of this test? lambdas and delegates should almost always be treated equivalently, and I think this case is one of those.

carlossanlop: Done. For both C# and VB.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
carlossanlop: Done

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
carlossanlop: I didn't have a test for that, so I added it. The analyzer wasn't able to handle this case, so I fixed it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -357,6 +359,107 @@ public static class Extensions
...
+using System.Threading;
+class C
+{
+    public static void MyMethod(int i, CancellationToken c = default) {}
+    public void M(CancellationToken ct)
+    {
+        LocalStaticMethod();
+        static void LocalStaticMethod()
+        {
+            MyMethod(5);
```
mavasani, line 456: This should generate a diagnostic. We should not be excluding these.

mavasani: Please also add a test where immediate outer local function is not static but its container is a static local function. We should generate a diagnostic even for that case:

```csharp
public static void MyMethod(int i, CancellationToken c = default) {}
    public void M(CancellationToken ct)
    {
        LocalStaticMethod();
        static void LocalStaticMethod()
        {
            LocalMethod();
            void LocalMethod()
            {
                MyMethod(5); // This should generate a diagnostic, but no code fix
            }
        }
}
```

mavasani: Basically, refactoring a large method to have local functions (nested or non-nested ones, static or non-static ones) should not alter the diagnostics reported by this analyzer. Local functions are primarily used to refactor code for readability, unlike lambdas which are passed to other methods and need to ensure no captures to avoid allocations.

carlossanlop: @mavasani Can you explain why having a static local function should trigger a diagnostic? According to the [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/static-local-functions), the token would not not reachable inside the static local method:

> _"A local function declared static cannot capture state from the enclosing scope. As a result, locals, parameters, and this from the enclosing scope are not available within a static local function."._

I tested the above code in a console app, and I get errors if I attempt to consume the token in any of the two cases:

image

My understanding was that suggesting a diagnostic without offering a fix was for cases that were valid but the fixer had limitations preventing it from offering a fix. In this case, we would be offering a diagnostic for a case that is unsupported/invalid.

Can you please tell me if I'm misunderstanding?

I can add the additional unit test you requested, but I would like to clarify this first.

mavasani: > the token would not not reachable inside the static local method:

Yes, I don't expect a code fix for this case. User would need to either convert it to a non-static local function OR thread in a new cancellation token parameter to the static local function. Both of these are absolutely acceptable changes given it leads to correctly passing in a cancellation token to a cancellable API invoked from the static local method.

carlossanlop: I see. Thanks for the explanation.
I will update the analyzer so it offers a diagnostic, and will work on ensuring the fixer bails out when it detects this case.

carlossanlop: Addressed, @mavasani. PTAL whenever you get a chance.

PR #3792, opened 2020-06-27
Update arcade (2.9.x branch)

sharwell:


eng/Version.Details.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Dependencies>
+  <ProductDependencies>
+    <Dependency Name="Microsoft.NETCore.Compilers" Version="3.7.0-3.20269.11">
```
sharwell, line 4: ❔ Should this also be set back to 3.3.1-beta3-final?

mavasani: @dotpaul to decide. I think it seems reasonable to move to latest compiler if nullable warnings are easy to fix.

dotpaul: I'd say do whatever is easiest for this PR to upgrade arcade, assuming it's not hard to move to the latest compilers in a separate PR.

mavasani:
Thanks a lot @sharwell!

PR #3797, opened 2020-06-30
Convert RulesetToEditorConfigConverter to use top-level statements

sharwell:


src/Tools/RulesetToEditorconfigConverter/Source/Program.cs
```diff
@@ -1,42 +1,37 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+#pragma warning disable IDE0051 // Remove unused private members
```
mavasani, line 3: Isn't this bug fixed yet?

src/Tools/RulesetToEditorconfigConverter/Source/Program.cs
```diff
@@ -1,42 +1,37 @@
...
+#pragma warning disable CA1812 // Avoid uninstantiated internal classes
```
mavasani, line 4: Can you file an issue for this CA1812 false positive?

src/Tools/RulesetToEditorconfigConverter/Source/Program.cs
sharwell: It was still getting reported in 16.7 Preview 3. I'm assuming that dotnet/roslyn#44848 will allow this to be cleaned up when the IDE is updated.

sharwell: Looks like https://github.com/dotnet/roslyn/issues/45496 might not be fixed yet.

mavasani:
@sharwell Can you resolve conflicts and fix build errors?

sharwell:
@mavasani Last time around build errors appeared to be caused by CI using an older compiler. Has that been fixed?

mavasani:
> Last time around build errors appeared to be caused by CI using an older compiler. Has that been fixed?

We did move to a newer compiler recently:
https://github.com/dotnet/roslyn-analyzers/blob/4da462a69dec7e5ce877dcb8fecb84f94ffe7bee/eng/Versions.props#L27

mavasani:
@sharwell - we already converted another project to use top level statements in this repo. So this PR should be unblocked.

sharwell:
@mavasani this is now updated

PR #3815, opened 2020-07-03
Update RS0046: ignore interface implementation and overrides

### LINKED ISSUE #3814: RS0046 should ignore implementations with matching names
#### Analyzer package

Roslyn.Diagnostics.Analyzers

#### Package Version

3.3.0-beta1.20262.5

#### Diagnostic ID

RS0046

#### Repro steps

```csharp
abstract class Base
{
  public abstract void Method(TextSpan? selectionOpt);
}

class Derived : Base
{
  public override void Method(TextSpan? selectionOpt) { }
}
```

#### Expected behavior

RS0046 is reported on `Base.Method`.

#### Actual behavior

RS0046 is reported on both `Base.Method` and `Derived.Method`.

#### Additional details

* This exclusion should apply to both interface implementations and overriding members.
* The exclusion only applies when the implementation parameter name is the _same_ as the base parameter name. If the parameter names do not match, the implementation will report a warning if applicable.

Evangelink:
Fix #3814

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
@@ -66,17 +66,34 @@ namespace Roslyn.Diagnostics.CSharp.Analyzers
             }
 
             var symbol = semanticModel.GetDeclaredSymbol(identifier.Parent, cancellationToken);
-            if (symbol?.GetMemberOrLocalOrParameterType()?.NullableAnnotation() != Analyzer.Utilities.Lightup.NullableAnnotation.Annotated)
-            {
-                return;
-            }
 
-            if (symbol.Kind != SymbolKind.Parameter ||
-                symbol.ContainingSymbol == null ||
-                (!symbol.ContainingSymbol.IsOverride && !symbol.ContainingSymbol.IsImplementationOfAnyInterfaceMember()))
+
+            if (ShouldReport(symbol))
             {
                 reportAction(identifier.CreateDiagnostic(Rule));
             }
+        }
+
+        private static bool ShouldReport(ISymbol symbol)
+        {
+            if (symbol?.GetMemberOrLocalOrParameterType()?.NullableAnnotation() != Analyzer.Utilities.Lightup.NullableAnnotation.Annotated)
+            {
+                // Not in a nullable context, bail-out
+                return false;
+            }
+
+            if (symbol.Kind != SymbolKind.Parameter ||
+                symbol.ContainingSymbol == null)
+            {
+                return true;
+            }
+
+            if (!symbol.ContainingSymbol.IsOverride && !symbol.ContainingSymbol.IsImplementationOfAnyInterfaceMember())
+            {
+                return true;
+            }
+
+            return false;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Roslyn.Diagnostics.Analyzers/UnitTests/CSharpAvoidOptSuffixForNullableEnableCodeTests.cs
```diff
@@ -264,5 +265,99 @@ public void Method1({|CS0246:Class2|}? s)
...
+#nullable enable
+
+public interface ISomething
+{
+    void Method1(string? s);
+}
+
+public class Class1 : ISomething
+{
+    public void Method1(string? sOpt)
```
Evangelink, line 302: Is it possible to force the renaming of implementation/overrides?

Evangelink, line 302: Would it be better to still report if the name is different from the parent? In this context that would mean that after the fix, we would start reporting on `sOpt`as the interface definition is now `s`. On this note, do you have any analyzer that reports when a parameter name is changed between definitions (interface says s but class says sOpt)? If not, would you be interested in such analyzer?

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpAvoidOptSuffixForNullableEnableCode.cs
```diff
@@ -66,7 +66,14 @@ private static void ReportOnInvalidIdentifier(SyntaxToken identifier, SemanticMo
...
             var symbol = semanticModel.GetDeclaredSymbol(identifier.Parent, cancellationToken);
-            if (symbol?.GetMemberOrLocalOrParameterType()?.NullableAnnotation() == Analyzer.Utilities.Lightup.NullableAnnotation.Annotated)
+            if (symbol?.GetMemberOrLocalOrParameterType()?.NullableAnnotation() != Analyzer.Utilities.Lightup.NullableAnnotation.Annotated)
+            {
+                return;
+            }
+
+            if (symbol.Kind != SymbolKind.Parameter ||
+                symbol.ContainingSymbol == null ||
+                (!symbol.ContainingSymbol.IsOverride && !symbol.ContainingSymbol.IsImplementationOfAnyInterfaceMember()))
```
mavasani, line 76: Nit: At this point it seems it would be more readable to write this with a static local function `ShouldReport(ISymbol)` with explicit return false for parameters whose containing member is an implementaiton.

src/Roslyn.Diagnostics.Analyzers/UnitTests/CSharpAvoidOptSuffixForNullableEnableCodeTests.cs
mavasani: Oh sorry, that seems to be only for the overloads. I don't think functionality exists for implementation/overrides.

Evangelink: @mavasani Any opinion on this?

mavasani: >  On this note, do you have any analyzer that reports when a parameter name is changed between definitions (interface says s but class says sOpt)? If not, would you be interested in such analyzer?

I am not sure, @sharwell are you aware if we have such a naming rule in IDE code style naming analyzers? I am assuming this will likely go into Roslyn repo as a naming analyzer.

PR #3838, opened 2020-07-08
Add CA1837 Use Environment.ProcessId

stephentoub:
New API being added for .NET 5 in https://github.com/dotnet/runtime/pull/38908
cc: @bartonjs


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdFixer.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            SemanticModel model = await doc.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            SyntaxNode root = await doc.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            if (root.FindNode(context.Span) is SyntaxNode node &&
```
mavasani, line 31: I believe this will return false when the diagnostic is on an argument: ‘SomeMethod(Process.GetCurrentProcess().Id);’. Can you please also add a unit test for this case?

```suggestion
            if (root.FindNode(context.Span, getInnermostNodeForTie: true) is SyntaxNode node &&
```

stephentoub: There is a unit test for it:
https://github.com/dotnet/roslyn-analyzers/pull/3838/files#diff-426f1f0417289c59e42bd1ebdc2718e6R85
It's why I added:
https://github.com/dotnet/roslyn-analyzers/pull/3838/files#diff-c0c60dc8311eeaa510154fcb87e26c2eR43
Maybe I can delete that check if I follow your getInnermostNodeForTie suggestion?

mavasani: Ah, I see you handle this case below and also have a unit test. You may be able to simplify the fixer code by still passing the above flag and not have to deal with IArgumentOperation below.

mavasani: If you make that change, ‘IOperation operation’ can be changed to ‘IPropertyReferenceOperation operation’

stephentoub: Cool, that worked. Thanks.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdFixer.cs
```diff
@@ -0,0 +1,61 @@
...
+                model.GetOperation(node, context.CancellationToken) is IPropertyReferenceOperation operation)
+            {
+                string title = MicrosoftNetCoreAnalyzersResources.UseEnvironmentProcessIdFix;
+                context.RegisterCodeFix(
+                    new MyCodeAction(title,
+                    async cancellationToken =>
+                    {
+                        DocumentEditor editor = await DocumentEditor.CreateAsync(doc, cancellationToken).ConfigureAwait(false);
+                        editor.ReplaceNode(node,
+                            editor.Generator.MemberAccessExpression(
```
mavasani, line 42: You probably want “.WithTriviaFrom(node)”. For example, if it is a non-first argument with a leading newline trivia. Might be good to add a test.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        public static int ProcessId => 0;
+    }
+}
+
+class C
+{
+    int M()
+    {
+        int pid = Environment.ProcessId;
```
mavasani, line 108: Do we care about the case where Process.GetProcess() is saved into a local/parameter, followed only by ID property references in the operation block? Might be common case given invocations is involved within the expression. If so, I think we can implement it with a simple DFA using the recently added Global control flow state based analysis. I can prototype it as a follow up PR as a proof of concept and hopefully demonstrate it should now be easy to write custom, performant DFA in the repo.

stephentoub: Personally I don't think it's worth it, but if you're looking to kick the tires of the global control flow state based analysis and want to try it out on this, cool :)

If someone was actually doing this the right way, they'd have written:
```C#
int pid;
using (Process p = Process.GetCurrentProcess())
    pid = p.Id;
```
or something like that, since Process is disposable (and finalizable).

mavasani: Sounds good. I will wait for next opportunity 😀

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessId.cs
```diff
@@ -0,0 +1,68 @@
...
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcess, out var processType) ||
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemEnvironment, out var environmentType))
+                {
+                    return;
+                }
+
+                // And only if we have Process.GetCurrentProcess, Process.Id, and Environment.ProcessId.
+                var processMembers = processType.GetMembers();
+                var processGetCurrentProcessSymbol = processMembers.FirstOrDefault(m => m.Name == "GetCurrentProcess");
+                var processIddSymbol = processMembers.FirstOrDefault(m => m.Name == "Id");
```
bartonjs, line 46:
```suggestion
                var processIdSymbol = processMembers.FirstOrDefault(m => m.Name == "Id");
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        Use(Environment.ProcessId);
+        return Environment.ProcessId;
+    }
+
+    void Use(int pid) {}
+}
+");
+        }
+    }
```
bartonjs, line 117: If you want an interesting "with trivia" replacement, perhaps

```C#
int pid = System.
    Diagnostics.
        Process.
            // Static invocation
            GetCurrentProcess().
                Id;
```

I have no idea what the <strike>right</strike> expected replacement is.

Yeah, my interesting trivia isn't all that interesting.  Waterfall with a comment.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1353,6 +1353,18 @@
   <data name="PreferIsEmptyOverCountTitle" xml:space="preserve">
     <value>Prefer IsEmpty over Count</value>
   </data>
+  <data name="UseEnvironmentProcessIdDescription" xml:space="preserve">
+    <value>Environment.ProcessId is simpler and faster than Process.GetCurrentProcess().Id.</value>
+  </data>
+  <data name="UseEnvironmentProcessIdMessage" xml:space="preserve">
+    <value>Use Environment.ProcessId instead of Process.GetCurrentProcess().Id.</value>
```
Evangelink, line 1360: Could you remove the trailing period for message and titles?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1353,6 +1353,18 @@
   <data name="PreferIsEmptyOverCountTitle" xml:space="preserve">
     <value>Prefer IsEmpty over Count</value>
   </data>
+  <data name="UseEnvironmentProcessIdDescription" xml:space="preserve">
+    <value>Environment.ProcessId is simpler and faster than Process.GetCurrentProcess().Id.</value>
```
Evangelink, line 1357:
```suggestion
    <value>'Environment.ProcessId' is simpler and faster than 'Process.GetCurrentProcess().Id'.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
Evangelink, line 1360:
```suggestion
    <value>Use 'Environment.ProcessId' instead of 'Process.GetCurrentProcess().Id'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1353,6 +1353,18 @@
...
+  </data>
+  <data name="UseEnvironmentProcessIdMessage" xml:space="preserve">
+    <value>Use Environment.ProcessId instead of Process.GetCurrentProcess().Id.</value>
+  </data>
+  <data name="UseEnvironmentProcessIdTitle" xml:space="preserve">
+    <value>Use Environment.ProcessId.</value>
```
Evangelink, line 1363:
```suggestion
    <value>Use 'Environment.ProcessId'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1353,6 +1353,18 @@
...
+  </data>
+  <data name="UseEnvironmentProcessIdFix" xml:space="preserve">
+    <value>Use Environment.ProcessId.</value>
```
Evangelink, line 1366:
```suggestion
    <value>Use 'Environment.ProcessId'</value>
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
stephentoub: Thanks.  My gut is any comments in the middle of such an expression should disappear with the expression, as the purpose of the fixer is to replace the whole thing, and if Process is going away and GetCurrentProcess is going away and Id is going away, it's really unlikely any comment would remain relevant.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdFixer.cs
stephentoub: Thanks.  Didn't know about that.  Added.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
stephentoub: There are tons of ending periods in the resx.  If we really don't want them, is someone going to go through and clean them all up soon?
cc: @mavasani

mavasani: Yes, we need to write an analyzer/code fix to cleanup and strictly enforce any desired patterns. I'll try to complete the analyzer/fixer that @Evangelink started in this space.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdFixer.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document doc = context.Document;
+            SemanticModel model = await doc.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            SyntaxNode root = await doc.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            if (root.FindNode(context.Span, getInnermostNodeForTie: true) is SyntaxNode node &&
```
Youssef1313, line 31: Curious when this condition can be false?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.UseEnvironmentProcessId,
+    Microsoft.NetCore.Analyzers.Runtime.UseEnvironmentProcessIdFixer>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Tasks.DoNotCreateTaskCompletionSourceWithWrongArguments,
+    Microsoft.NetCore.Analyzers.Tasks.DoNotCreateTaskCompletionSourceWithWrongArgumentsFixer>;
```
Youssef1313, line 10: This needs to be updated. And probably it's worth adding tests for vb.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdFixer.cs
stephentoub: If FindNode returns null.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
stephentoub: Updated to what?

Youssef1313: To the analyzer and code fix under test, which is use env process id.
@stephentoub

stephentoub: Ah, yes :)

PR #3844, opened 2020-07-09
Avoid capturing a mutable list in code action application lambda

### LINKED ISSUE #3824: Suggested code fix is being applied 3 times
The CA2016 rule detects methods that can take a CancellationToken (either because it's an optional parameter or there's a method overload) and suggests to pass it, if the parent method defines a token parameter.

Today I was testing the consumption of the latest Microsoft.CodeAnalysis.NetAnalyzers nuget package for the dotnet/runtime repo, and noticed that if I manually applied the suggestions for that rule, the token would get added 3 times instead of one.

[Here ](https://github.com/dotnet/roslyn-analyzers/pull/3791) is the latest PR that fixed some bugs in that analyzer. Notice the unit tests passed, which means the token was applied only once in all cases that accepted a code fix, as expected. 

@mavasani asked me to manually verify if by changing [this ](https://github.com/dotnet/roslyn-analyzers/blob/e993332bbf9820cd76147948655810eb2545c377/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Fixer.cs#L31) and [this ](https://github.com/dotnet/roslyn-analyzers/blob/master/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Fixer.cs#L115) from `List<SyntaxNode>` to `ImmutableArray<SyntaxNode>` the issue would go away, but unfortunately I still see it.

Have there been any recent changes in the Code Fixer logic that could be causing this behavior?

Here's a screenshot of my repro. I made sure to install the latest nuget package version. Notice that the suggestion preview is **correct**:

image

But if I apply the code fix, this is the result:

image

mavasani:
Fixes #3824

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpForwardCancellationTokenToInvocations.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpForwardCancellationTokenToInvocations.Fixer.cs
@@ -1,8 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeFixes;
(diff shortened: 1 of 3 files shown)


sharwell:
Can we add a test for this?

mavasani:
> Can we add a test for this?

I wasn't sure how. It only seems to repro in VS IDE, possibly getting mutated when preview is invoked?

mavasani:
@carlossanlop @sharwell can you review? If you have a suggestion on how to write a test, please let me know.

mavasani:
Thanks!

carlossanlop:
Thanks for taking care of this, @mavasani.

Issue #3793, opened 2020-06-28
RS0041 reported in Resources.Designer.cs


JohanLarsson:
```
Severity	Code	Description	Project	File	Line	Category	Suppression State
Warning	RS0041	Symbol 'ResourceManager.get' uses some oblivious reference types.	Gu.Localization (netstandard2.0)	C:\Git\_GuOrg\Gu.Localization\Gu.Localization\Properties\Resources.Designer.cs	40	Build	N/A
```

mavasani:
Tagging @jcouv

Youssef1313:
Looks like it's intentional based on [this](https://github.com/dotnet/roslyn-analyzers/blob/2bc4ddee57716be74aca26d1508391767b05a33d/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs#L189) comment.

jcouv:
From discussion with @sharwell it is good that we complain about public API issues both in user code and in generated code, so that all issues with the public API surface are visible.

sharwell:
On that note, I do need to update [ResxSourceGenerator](https://github.com/dotnet/roslyn-sdk/pull/552) to add the `#nullable enable` line.

bitbonk:
It is still not clear how this problem can be solved at all. So I created https://github.com/dotnet/roslyn-analyzers/issues/5521

bitbonk:
@sharwell Since this is now closed, do you maybe have a hint on how to solve the problem?

Issue #3672, opened 2020-05-24
Stack overflow when constraining on IComparable<T>


rsking-mondo:
#### Analyzer package

[Microsoft.CodeAnalysis.PublicApiAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.PublicApiAnalyzers)

#### Package Version

v3.0.0 (Latest)

#### Repro steps

1. Install Microsoft.CodeAnalysis.PublicApiAnalyzers
2. Enable nullable
3. Create generic method, that constrains where T : System.IComparable<T>
4. Add #nullable enable to PublicAPI.Shipped.txt
5. run `dotnet build` using .NET Core SDK 3.1.300

#### Expected behavior

Code compiles

#### Actual behavior

```cmd
C:\Program Files\dotnet\sdk\3.1.300\Roslyn\Microsoft.CSharp.Core.targets(59,5): error : Stack overflow.
```

rsking-mondo:
[TestComparable.zip](https://github.com/dotnet/roslyn-analyzers/files/4674647/TestComparable.zip)
Attached is a sample project showing this issue.

mavasani:
@jcouv

jcouv:
Fixed. Thanks for reporting this @rsking-mondo

PR #3847, opened 2020-07-10
Fix unbounded recursion

jcouv:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/3672 (stack overflow on `where T : IComparable<T>` in public API analyzer when nullability analysis is enabled).
Closes https://github.com/dotnet/roslyn-analyzers/issues/3793 (adding a test)


jcouv:
@mavasani @sharwell for review. Thanks

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -832,7 +832,16 @@ public override bool VisitNamedType(INamedTypeSymbol symbol)
...
                     {
-                        if (Instance.Visit(typeArgument))
+                        // type parameters will already have been checked on the type defining them, so we can just do a shallow check
+                        if (typeArgument.TypeKind == TypeKind.TypeParameter)
+                        {
+                            if (typeArgument.IsReferenceType &&
+                                typeArgument.NullableAnnotation() == NullableAnnotation.None)
+                            {
+                                return true;
+                            }
```
sharwell, line 842: ❔ Why no `else return false`?

jcouv: We need to check the following type arguments (there may be more than one). Added a test to cover that scenario.

jcouv:
@sharwell Good to go?

PR #3853, opened 2020-07-14
Update CA1031: Allow excluded_symbol_names user option

### LINKED ISSUE #3211: CA1031 should ignore "safe" methods
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

CA1031

#### Repro steps

Note: this is strict a simple example, not something you might actually write.

```csharp
int TryConvert ( string value )
{
   try
   {
      return Int32.Parse(value);
   } catch (Exception e)
   {   
   };
   return 0;
}
```
1. Compile the above code and you, correctly, get the warning that you shouldn't catch generic exception

#### Expected behavior

The issue with this warning is that there are cases where methods are "safe". Safe methods are methods that are attempting to do something and the expectation is that they can fail but that it isn't an error that should be bubbled up. 

Let's ignore for purposes of this discussion that some exceptions (like `OutOfMemoryException`) cannot be correctly handled by the exception. Because C# doesn't have an easy way to say "handle all exceptions except - other than exception filters) it is difficult to properly build a series of try-catch blocks that would correctly handle all the exceptions you might care about and the few you don't, short of an exception filter. Since the purpose of this warning is to just let you know something might be wrong it makes sense to be able to ignore it when it doesn't make sense. 

The problem is that the existing suppression syntax (outside a global suppression) is either a verbose code analysis attribute or 2 pragmas. It would be nice if certain methods could be considered "safe" out of the box and therefore be ignored by this warning. Off the top of my head I'd argue that `Safe-` and `Try-` methods would qualify but I could easily see other methods that might as well. I'd like to see either these 2 types of methods auto-excluded and/or a special attribute be added (or perhaps there is one already) such that a method can be marked safe for this type of error. Again, the existing analysis attribute just seems too wordy and the pragma clutters up things. I believe this was done for the validating null argument rule so something similar would be nice here.

#### Actual behavior

Currently you have to suppress this warning in "safe" methods even though most of the time such exceptions are silently logged/eaten/etc.

Evangelink:
In addition to providing the excluded_symbol_names options the support for wildcard symbol names is also enabled.

Fix #3211

Note that I still need to:

- [ ] Update the doc website

- [ ] Update https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#excluded-symbol-names


src/Utilities/Compiler/DoNotCatchGeneralUnlessRethrown.cs
```diff
@@ -78,6 +84,18 @@ public override void Initialize(AnalysisContext analysisContext)
...
                             compilationStartAnalysisContext.Options, compilationStartAnalysisContext.CancellationToken);
 
+                    bool IsExcludedSymbol(SyntaxTree tree)
+                    {
+                        if (!_allowExcludedSymbolNames)
+                        {
+                            return false;
+                        }
+
+                        var excludedSymbols = operationBlockAnalysisContext.Options.GetExcludedSymbolNamesWithValueOption(SupportedDiagnostics[0], tree,
```
mavasani, line 94: We should instead be invoking the following extension method: https://github.com/dotnet/roslyn-analyzers/blob/641b5e8b6ddf8ccf095f5ac58a346ed1de546a18/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L196-L204

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotCatchGeneralExceptionTypesTests.cs
```diff
@@ -701,5 +701,70 @@ End Sub
...
+        // No configuration - validate no diagnostics in default configuration
+        [InlineData("")]
+        // Check with exact method signature
+        [InlineData("dotnet_code_quality.CA1031.excluded_symbol_names = M:SomeNamespace.Test.M1(System.String)")]
+        [InlineData("dotnet_code_quality.excluded_symbol_names = M:SomeNamespace.Test.M1(System.String)")]
+        // Check with wildcard method signature
+        [InlineData("dotnet_code_quality.CA1031.excluded_symbol_names = M:SomeNamespace.Test.M*")]
+        [InlineData("dotnet_code_quality.excluded_symbol_names = M:SomeNamespace.Test.M*")]
+        // Check with wildcard type signature
+        [InlineData("dotnet_code_quality.CA1031.excluded_symbol_names = T:SomeNamespace.Tes*")]
```
mavasani, line 715: Interesting, not sure how the non-method tests are passing, given you check `excludedSymbols.Contains(method);`. Can you remind me if this option value leads to type symbol and all its member symbols being included in the resultant symbol set returned when option value is parsed?

mavasani: Ah, so I presume the `.Contains` works up the containing symbol chain. So, specifying a non-wildcard value like `[InlineData("dotnet_code_quality.CA1031.excluded_symbol_names = T:SomeNamespace.Test")]`, will cause the test to fail? We should ensure it is not such. I this also indicates that we are doing duplicate work in `ISymbolExtensions.IsConfiguredToSkipAnalysis`, that method has a loop walking up the containing symbol chain, and does a contains check for each symbol in this chain, which will in turn execute the wildcard based containing symbol chain walk.

Evangelink: The wildcard walks-up the containing chain while the non-wildcard doesn't. In the example of `[InlineData("dotnet_code_quality.CA1031.excluded_symbol_names = T:SomeNamespace.Test")]` yes the test will fail (i.e. the diagnostic will still be reported). I'll check what's going-on when I use `IsConfiguredToSkipAnalysis` and get back to you.

src/Utilities/Compiler/DoNotCatchGeneralUnlessRethrown.cs
Evangelink: That's not the first time I am using the wrong method... I am wondering if we should hide `GetExcludedSymbolNamesWithValueOption` has a local function of the `IsConfiguredToSkipAnalysis` to avoid such issue. We could also create an analyzer but that seems overkill compared to hiding the function.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotCatchGeneralExceptionTypesTests.cs
Evangelink: Yeah \o/ So using the `IsConfiguredToSkipAnalysis` helper allows this walk-up check! Now that I see that helper I think that my wildcard implementation is actually way to complex has I was already trying to do the walk-up chain while the helper already does a walk-up callback. I don't feel doing this now but I will leave a comment on the ticket to revisit the algo.

src/Utilities/Compiler/DoNotCatchGeneralUnlessRethrown.cs
Evangelink: I guess we need to create a ticket for the disposable rules: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs#L108 I have once again used the wrong helper.

Evangelink:
Ping @mavasani

mavasani:
@Evangelink Do you want to tackle the TODOs you mentioned in the PR description in this PR or a follow-up?

Evangelink:
I'll do a follow-up tomorrow.

PR #3867, opened 2020-07-15
Update RS1024 to cover more comparisons

### LINKED ISSUE #2493: Roslyn analyzer for comparing symbols
As part of https://github.com/dotnet/roslyn/issues/35933, symbol equality is going to change to only include things that were in the language in c# 7.3, and we're going to start requiring using explicit comparators when comparing `ITypeSymbol`s. This is going to affect a few different scenarios:

1. Straight comparisons of symbols should use a comparator so that they are explicit about what type of comparison they want.
2. Dictionaries should use a comparator, as by default symbols with differing nested nullabilities will be considered equal for the purposes of being keys.
3. Anything that calls `GetHashCode()` will likely be doing the wrong thing in some cases.

Evangelink:
Fix #2493


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -31,7 +32,27 @@ public class CompareSymbolsCorrectlyAnalyzer : DiagnosticAnalyzer
             description: s_localizableDescription,
             customTags: WellKnownDiagnosticTags.Telemetry);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+        public static readonly DiagnosticDescriptor GetHashCodeRule = new DiagnosticDescriptor(
```
Evangelink, line 35: I have changed only the description for this descriptor, let me know if you would like a different message too.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -31,7 +32,27 @@ public class CompareSymbolsCorrectlyAnalyzer : DiagnosticAnalyzer
...
+            DiagnosticIds.CompareSymbolsCorrectlyRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.MicrosoftCodeAnalysisCorrectness,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: s_localizableDescriptionGetHashCode,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public static readonly DiagnosticDescriptor CollectionRule = new DiagnosticDescriptor(
```
Evangelink, line 45: There is currently no difference between this descriptor and the `EqualityDescriptor`, the only goal is to prevent registering the code-fix for the wrong case.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -31,7 +32,27 @@ public class CompareSymbolsCorrectlyAnalyzer : DiagnosticAnalyzer
...
+            DiagnosticIds.CompareSymbolsCorrectlyRuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.MicrosoftCodeAnalysisCorrectness,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: s_localizableDescription,
+            customTags: WellKnownDiagnosticTags.Telemetry);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(EqualityRule);
```
Evangelink, line 55: I haven't put the other descriptors here because otherwise I have to change all use of the markup syntax in the tests. @sharwell Maybe the check shall be relaxed to allow markup syntax when the different descriptors registered have the same ID, WDYT?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -202,6 +271,54 @@ private static bool IsExplicitCastToObject(IOperation operation)
...
+            {
+                AddOrUpdate(nameof(Enumerable.Contains), enumerableType);
+                AddOrUpdate(nameof(Enumerable.Distinct), enumerableType);
+                AddOrUpdate(nameof(Enumerable.GroupBy), enumerableType);
+                AddOrUpdate(nameof(Enumerable.GroupJoin), enumerableType);
+                AddOrUpdate(nameof(Enumerable.Intersect), enumerableType);
+                AddOrUpdate(nameof(Enumerable.Join), enumerableType);
+                AddOrUpdate(nameof(Enumerable.SequenceEqual), enumerableType);
+                AddOrUpdate(nameof(Enumerable.ToDictionary), enumerableType);
+                AddOrUpdate("ToHashSet", enumerableType);
```
Evangelink, line 304: Looks like `ToHashSet` is not on the old framework.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
```diff
@@ -538,5 +543,228 @@ class TestClass {{
...
+                    {
+                        @"
+Imports System
+Imports System.Collections.Immutable
+Imports System.Collections.Generic
+Imports System.Linq
+Imports Microsoft.CodeAnalysis
+
+Public Class C
+    Public Sub MethodWithDiagnostics(kvps As IEnumerable(Of KeyValuePair(Of ISymbol, Integer)), symbols As IEnumerable(Of ISymbol), symbol As ISymbol)
```
Evangelink, line 718: For the cases where we would expect a diagnostic and where we don't have it (i.e. x4, x8, x9, x10, x13, x15 and x18), there is something weird with how the method is resolved.

For example, for [ToImmutableHashSet](https://docs.microsoft.com/dotnet/api/system.collections.immutable.immutablehashset.toimmutablehashset), the docs considers the method as being generic while what we have is not a generic method:

image

Because of that we don't consider the method as acting on `ISymbol` so we bail-out.

Evangelink: Ping @mavasani

mavasani: @Evangelink Can you look at `method.OriginalDefinition`?

Evangelink: It doesn't work. I found a workaround but I cannot tell whether this is a bug to report or not.

Evangelink:
Ping @mavasani

mavasani:
Tagging @ryzngard for review

Evangelink:
Ping @ryzngard (do you have some time for review)?

ryzngard:
@Evangelink yes! Sorry for the delayed review, I'll get to it today.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
ryzngard: I think description change is fine for this

ryzngard: Oh, I see we have the same rule ID. FWIW I don't think it hurts to separate these out into different ID. It provides more clarity even if the message is the same.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
ryzngard: I think we should make different ID for the checks. Even if messaging remains the same, it will help correlate the rule being triggered.

ryzngard (approved):

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
Evangelink: It's never really clear when we want to split the IDs or just use different descriptors. @mavasani could you let me know what's your preference here?

mavasani: We use the following rule of thumb:
1. If the only difference is in diagnostic message, use multiple descriptors with same ID but different message.
2. If we expect user to want to bulk configure and/or suppress issues from each bucket separately, split into multiple IDs.

Evangelink: @ryzngard Based on the previous comment, I don't really think we expect different bulk configuration so I would say it's better to keep the same ID but let me know if you would still prefer to have a different one.

mavasani:
Filed https://github.com/dotnet/roslyn-analyzers/issues/4301 for Andrew's suggestion.

PR #3868, opened 2020-07-15
Standardize naming in WellKnownTypeNames

Evangelink:
Add suffix to the names matching the number of generic type arguments.

Also fix some broken ordering in the constants.

PR #3872, opened 2020-07-16
Standardize naming in WellKnownTypeNames

Evangelink:
Relates to #3868


src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -252,7 +252,7 @@ internal static class WellKnownTypeNames
         public const string SystemThreadingCancellationToken = "System.Threading.CancellationToken";
         public const string SystemThreadingInterlocked = "System.Threading.Interlocked";
         public const string SystemThreadingMonitor = "System.Threading.Monitor";
-        public const string SystemThreadingTasksGenericTask = "System.Threading.Tasks.Task`1";
+        public const string SystemThreadingTasksTasks = "System.Threading.Tasks.Task`1";
```
mavasani, line 255: This should be `SystemThreadingTasksTask1`
```suggestion
        public const string SystemThreadingTasksTask1 = "System.Threading.Tasks.Task`1";
```

mavasani:
@Evangelink I think you would need to rename even the use sites :-)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
```diff
@@ -89,7 +89,7 @@ private static void AnalyzeOperation(OperationAnalysisContext context, Immutable
         private static bool TryGetTaskTypes(Compilation compilation, out ImmutableArray<INamedTypeSymbol> taskTypes)
         {
             INamedTypeSymbol? taskType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTask);
-            INamedTypeSymbol? taskOfTType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksGenericTask);
+            INamedTypeSymbol? taskOfTType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTasks);
```
sharwell, line 92: 😕 I don't think this one is right

Evangelink:
Sorry guys, I have blindly applied the rename suggestion! It's alright now.

PR #3875, opened 2020-07-16
Do not report CA1812 (Make static) on public APIs for web projects

### LINKED ISSUE #3835: Disable CA1822 on ASP.Net Startup Classes etc
Classes like `ConfigureServices` are being marked with `CA1822` as a fix. If the method is static it will not work unless I am missing something.

This also occurs in controller methods which don't access any instance data. (Again won't work if you apply the fix)

mavasani:
Fixes #3835

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -292,25 +292,6 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                 }
             }
 
-            // Web attributes
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebServicesWebMethodAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpGetAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpPostAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpPutAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpDeleteAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpPatchAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpHeadAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebMvcHttpOptionsAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemWebHttpRouteAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpDeleteAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpGetAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpPostAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpPutAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpHeadAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpPatchAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcHttpOptionsAttribute));
-            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcRouteAttribute));
-
             // MSTest attributes
             Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestInitializeAttribute));
             Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftVisualStudioTestToolsUnitTestingTestMethodAttribute));
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -50,8 +50,10 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                 // Get the list of all method' attributes for which the rule shall not be triggered.
                 ImmutableArray<INamedTypeSymbol> skippedAttributes = GetSkippedAttributes(wellKnownTypeProvider);
 
+                var isWebProject = compilationContext.Compilation.IsWebProject(compilationContext.Options, compilationContext.CancellationToken);
+
                 compilationContext.RegisterSymbolStartAction(
-                    symbolStartContext => OnSymbolStart(symbolStartContext, wellKnownTypeProvider, skippedAttributes),
+                    symbolStartContext => OnSymbolStart(symbolStartContext, wellKnownTypeProvider, skippedAttributes, isWebProject),
                     SymbolKind.NamedType);
             });
 
@@ -60,7 +62,8 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
             static void OnSymbolStart(
                 SymbolStartAnalysisContext symbolStartContext,
                 WellKnownTypeProvider wellKnownTypeProvider,
-                ImmutableArray<INamedTypeSymbol> skippedAttributes)
+                ImmutableArray<INamedTypeSymbol> skippedAttributes,
+                bool isWebProject)
             {
                 // Since property/event accessors cannot be marked static themselves and the associated symbol (property/event)
                 // has to be marked static, we want to report the diagnostic on the property/event.
@@ -94,7 +97,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                     }
 
                     // Don't run any other check for this method if it isn't a valid analysis context
-                    if (!ShouldAnalyze(methodSymbol, wellKnownTypeProvider, skippedAttributes))
+                    if (!ShouldAnalyze(methodSymbol, wellKnownTypeProvider, skippedAttributes, isWebProject))
                     {
                         return;
                     }
@@ -182,7 +185,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
             }
         }
 
-        private static bool ShouldAnalyze(IMethodSymbol methodSymbol, WellKnownTypeProvider wellKnownTypeProvider, ImmutableArray<INamedTypeSymbol> skippedAttributes)
+        private static bool ShouldAnalyze(IMethodSymbol methodSymbol, WellKnownTypeProvider wellKnownTypeProvider, ImmutableArray<INamedTypeSymbol> skippedAttributes, bool isWebProject)
         {
             // Modifiers that we don't care about
             if (methodSymbol.IsStatic || methodSymbol.IsOverride || methodSymbol.IsVirtual ||
@@ -191,7 +194,15 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                 return false;
             }
 
-            if (IsSkippedMethod(methodSymbol, wellKnownTypeProvider))
+            // Do not analyze constructors and finalizers.
+            if (methodSymbol.IsConstructor() || methodSymbol.IsFinalizer())
+            {
+                return false;
+            }
+
+            // Do not analyze public APIs for web projects
+            // See https://github.com/dotnet/roslyn-analyzers/issues/3835 for details.
+            if (isWebProject && methodSymbol.IsExternallyVisible())
             {
                 return false;
             }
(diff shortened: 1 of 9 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -308,24 +319,6 @@ void Add(INamedTypeSymbol? symbol)
             return builder?.ToImmutable() ?? ImmutableArray<INamedTypeSymbol>.Empty;
         }
 
-        private static bool IsSkippedMethod(IMethodSymbol methodSymbol, WellKnownTypeProvider wellKnownTypeProvider)
```
Evangelink, line 311: Is there any need to keep all the web attribute exclusions (see https://github.com/dotnet/roslyn-analyzers/pull/3875/files#diff-b71c5f4c8942242dec4740ebf6efda72R282-R298) now that we don't report on public APIs for web projects?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -1016,9 +1016,18 @@ public class C : System.Web.HttpApplication
     // The following controller methods can't be made static
     protected void Application_Start() { }
     protected void Application_End() { }
-}");
+}";
 
-            await VerifyVB.VerifyAnalyzerAsync(@"
+            const string editorConfigText = "build_property.UsingMicrosoftNETSdkWeb = true";
+            var csTest = new VerifyCS.Test()
```
Evangelink, line 1022: Any reason you have split the test and the `RunAsync`? You could write something like

```csharp
await new VerifyCS.Test
{
	TestCode = csSource,
	AnalyzerConfigDocument = editorConfigText,
}.RunAsync();
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -1016,9 +1016,18 @@ public class C : System.Web.HttpApplication
...
+            {
+                TestCode = csSource,
+                AnalyzerConfigDocument = editorConfigText
```
Evangelink, line 1025: [minor] @sharwell likes to have the trailing comma here :)

```suggestion
                AnalyzerConfigDocument = editorConfigText,
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
Evangelink: In case you apply this comment, note that it applies to the various tests you have changed.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
Evangelink: If you decide to apply this comment, note that you will have to do the change for the various tests you have changed.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -1016,9 +1016,18 @@ public class C : System.Web.HttpApplication
     // The following controller methods can't be made static
     protected void Application_Start() { }
     protected void Application_End() { }
-}");
+}";
 
-            await VerifyVB.VerifyAnalyzerAsync(@"
+            const string editorConfigText = "build_property.UsingMicrosoftNETSdkWeb = true";
```
Evangelink, line 1021: Just to confirm I got things right, we don't expect user to set this in their `.editorconfig` file? This is a msbuild property that is present for web apps, right?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -191,7 +194,15 @@ private static bool ShouldAnalyze(IMethodSymbol methodSymbol, WellKnownTypeProvi
...
-            if (IsSkippedMethod(methodSymbol, wellKnownTypeProvider))
+            // Do not analyze constructors and finalizers.
+            if (methodSymbol.IsConstructor() || methodSymbol.IsFinalizer())
+            {
+                return false;
+            }
+
+            // Do not analyze public APIs for web projects
+            // See https://github.com/dotnet/roslyn-analyzers/issues/3835 for details.
+            if (isWebProject && methodSymbol.IsExternallyVisible())
```
Evangelink, line 205: Shall we really skip `protected` members? I don't know exactly if they expect the member to be `public` for the reflection or if `protected` is ok.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
```diff
@@ -59,5 +61,15 @@ internal static bool TargetsDotNetFramework(this Compilation compilation)
...
         }
+
+        /// <summary>
+        /// Gets a value indicating whether the project of the compilation is a Web SDK project,
+        /// based on MSBuild property <see cref="MSBuildPropertyOptionNames.UsingMicrosoftNETSdkWeb"/>.
+        /// </summary>
+        internal static bool IsWebProject(this Compilation compilation, AnalyzerOptions options, CancellationToken cancellationToken)
+        {
+            var usingMicrosoftNETSdkWeb = options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.UsingMicrosoftNETSdkWeb, compilation, cancellationToken);
+            return usingMicrosoftNETSdkWeb != null && bool.TryParse(usingMicrosoftNETSdkWeb, out var result) && result;
```
Evangelink, line 72: Do you need to potentially trim the string option before doing a `TryParse`? Also as you are then checking the fact this is true, isn't it better to replace the full line with `string.Equals(usingMicrosoftNETSdkWeb?.Trim(), "true", StringComparison.OrdinalIgnoreCase)`?

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
```diff
@@ -59,5 +61,15 @@ internal static bool TargetsDotNetFramework(this Compilation compilation)
...
             return false;
         }
+
+        /// <summary>
+        /// Gets a value indicating whether the project of the compilation is a Web SDK project,
+        /// based on MSBuild property <see cref="MSBuildPropertyOptionNames.UsingMicrosoftNETSdkWeb"/>.
+        /// </summary>
+        internal static bool IsWebProject(this Compilation compilation, AnalyzerOptions options, CancellationToken cancellationToken)
+        {
+            var usingMicrosoftNETSdkWeb = options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.UsingMicrosoftNETSdkWeb, compilation, cancellationToken);
```
Evangelink, line 71: [cosmetic] the variable name reads to me as if it's a boolean while it's a string, maybe `usingMicrosoftNETSdkWebPropertyValue` is clearer (not sure because it's really long...).

src/Utilities/Compiler/Options/AbstractCategorizedAnalyzerConfigOptions.cs
```diff
@@ -88,13 +89,23 @@ public bool TryGetOptionValue<T>(string optionName, OptionKind kind, DiagnosticD
...
-                TryGetGeneralOptionValue(optionKeyPrefix, out optionValue))
+
+            T optionValue;
+            if (rule != null)
+            {
+                if (TryGetSpecificOptionValue(rule.Id, optionKeyPrefix, out optionValue) ||
+                    TryGetSpecificOptionValue(rule.Category, optionKeyPrefix, out optionValue) ||
+                    TryGetAnySpecificOptionValue(rule.CustomTags, optionKeyPrefix, out optionValue))
+                {
+
```
Evangelink, line 103:
```suggestion
```

src/Utilities/Compiler/Options/AbstractCategorizedAnalyzerConfigOptions.cs
```diff
@@ -88,13 +89,23 @@ public bool TryGetOptionValue<T>(string optionName, OptionKind kind, DiagnosticD
...
-        private (bool found, object? value) ComputeOptionValue<T>(string optionName, OptionKind kind, DiagnosticDescriptor rule, TryParseValue<T> tryParseValue)
+        private (bool found, object? value) ComputeOptionValue<T>(string optionName, OptionKind kind, DiagnosticDescriptor? rule, TryParseValue<T> tryParseValue)
         {
             var optionKeyPrefix = MapOptionKindToKeyPrefix(kind);
-            if (TryGetSpecificOptionValue(rule.Id, optionKeyPrefix, out var optionValue) ||
-                TryGetSpecificOptionValue(rule.Category, optionKeyPrefix, out optionValue) ||
-                TryGetAnySpecificOptionValue(rule.CustomTags, optionKeyPrefix, out optionValue) ||
-                TryGetGeneralOptionValue(optionKeyPrefix, out optionValue))
+
+            T optionValue;
```
Evangelink, line 96: Out of curiosity, is it to reduce allocation that you declare the variable before rather than doing `out var`?

src/Utilities/Compiler/Options/AbstractCategorizedAnalyzerConfigOptions.cs
```diff
@@ -88,13 +89,23 @@ public bool TryGetOptionValue<T>(string optionName, OptionKind kind, DiagnosticD
...
+            if (rule != null)
+            {
+                if (TryGetSpecificOptionValue(rule.Id, optionKeyPrefix, out optionValue) ||
```
Evangelink, line 99: [suggestion] could be merged with outer if.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
Evangelink, line 71: Is this property valid only for aspnet core or also regular aspnet? If it's valid only for aspnet core, do we really want to get rid of the old exclusions? That would mean raising FPs for "old" projects.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
mavasani: Yes, users do not have to do anything.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
mavasani: I am not sure, but I made is as conservative as possible, given protected members in public types are public APIs.

Evangelink:
Without forcing you to add too many tests, it might be nice to add a couple of tests for some of the cases mentioned here https://github.com/dotnet/roslyn-analyzers/issues/3835#issuecomment-657591877

src/Utilities/Compiler/Options/AbstractCategorizedAnalyzerConfigOptions.cs
mavasani: No, the var is again needed in a separate if check below, hence it must be scoped to the whole method.

mavasani:
> Without forcing you to add too many tests, it might be nice to add a couple of tests for some of the cases mentioned here

I am not sure if that provides any valuable test coverage addition. The analyzer itself is not detecting any special patterns, except for bailing out on externally visible types when a certain option is provided - basically if you set that option manually in .editorconfig, it will skip externally visible types for that project. It is not checking for any semantics around the symbols or compilation, so there will not be real value add.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
mavasani: It will be valid for all SDK based aspnet projects. @pranavkm Is there any MSBuild property that is set for regular aspnet projects that we can use?

Evangelink:
It just provides a minimal safety-net in case we change something in the rule later on but I agree value is minimal here. It could be valuable to have some "integration" test though (but that's not in the scope of this PR).

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: It seems that `Application_Start` could be `protected` so it's mandatory anyways.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
mavasani: Hmm @pranavkm for advice. Can these web attributes be applied on non-public methods? Unless we are sure, I'd prefer to keep the attribute based exclusions.

src/Utilities/Compiler/Options/AbstractCategorizedAnalyzerConfigOptions.cs
mavasani: That required adding parentheses around the second checks and was making the code clumsy :-)

mavasani:
>  It could be valuable to have some "integration" test though (but that's not in the scope of this PR).

Yes, that would be valuable, and agree out of the scope of this PR.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
mavasani: Awaiting an object creation with initializer with an async method invocation seems too chatty to me :-). I prefer to keep object creation with initializer as its own statement.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
mavasani: Sure, will change.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
mavasani: Seems like other tests in this file follow the pattern you suggested. I will switch the pattern in new tests to make it consistent.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
pranavkm: > Is there any MSBuild property that is set for non-SDK based aspnet projects that we can use?

Not that I know of.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
pranavkm: The link is bugged (I can't see where it's meant to go), but in general MVC (5.0 and Core) will only discover public types and methods.

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
mavasani: I looked around and the only way I could identify is by looking at `ProjectTypeGuids` property.

```csharp
public const string WebAppProjectGuidString = "349C5851-65DF-11DA-9384-00065B846F21";
public const string WebSiteProjectGuidString = "E24C65DC-7377-472B-9ABA-BC803B73C61A";
```

I'll can add these property value checks unless we find a better way.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: @pranavkm Here is the link on master: https://github.com/dotnet/roslyn-analyzers/blob/master/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs#L270-L287

mavasani:
@Evangelink I have added support for non-SDK projects and also added support for `api_surface` option for the analyzer. I have left the attribute checks as I am not sure if they are intended only for public API surface or not.

mavasani:
@Evangelink would you mind reviewing the latest commit when you get time? Thanks!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
pranavkm: Those are required to be on public types or methods. Feel free to remove these exclusions.

Issue #3870, opened 2020-07-15
Visual Studio version 16.6 onwards is extremely slow and hangs too often!


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/1109183/visual-studio-version-166-onwards-is-extremely-slo.html)._

---
[regression] [worked-in:16.4.2]
This feedback is very hard to pin down on one instance but I am seeing that for big solutions (~100 projects) version 16.6 and above hang too often. The IDE just freezes or becomes painfully slow. I often have to kill the process and restart. This is more frequent when adding a new project reference requires nuget restore to run on the project. I have tried preview versions too and they might be margnially better but general perf has degraded way too much to not have been noticed by others.

---
#### Roi Chen [MSFT] on 7/10/2020, 02:38 AM: 

<p>Thank you for taking the time to log this issue!<br>
In order for us to investigate further, please provide more information by recording your repro steps via the Visual Studio Feedback Tool (Help -&gt; Send Feedback -&gt; Report A Problem). The feedback tool will ensure that we collect the needed information for you without worrying about what to provide (recording, dump file or ETL trace). Since this issue is now marked as Need More Info, that workflow is enabled in the Feedback Tool – <a target='_blank' href="https://docs.microsoft.com/en-us/visualstudio/ide/how-to-report-a-problem-with-visual-studio-2017?view=vs-2017#when-further-information-is-needed-need-more-info">full instructions</a><br>
In our <a target='_blank' href="https://aka.ms/vsfeedbackguidelines/#writing-a-good-bug-report-or-feature-suggestion">guidelines</a>, you can get tips on how to provide clear and simple reproducible steps.<br>
We look forward to hearing from you!</p>

#### Ramjot Singh on 7/10/2020, 11:52 PM: 

I am adding records and traces of a session where VS got extremely slow. Every simple action would take seconds.
#### Ramjot Singh on 7/14/2020, 06:37 PM: 

<p><a rel="user" href="https://developercommunity.visualstudio.com/users/18031/d62bfa40-d617-6c9d-ad68-8760dcad8e39.html" nodeid="18031">@Roi Chen [MSFT]</a> Do you need anymore info? Or were the provided traces enough?</p>

#### Feedback Bot on 7/15/2020, 09:24 AM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

[...]

PR #3881, opened 2020-07-17
WellKnownTypeProvider optimizations attempt #A

dotpaul:
Attempt to address #3870 without making things worse.

Idea was to avoid (number of tainted data source types + maybe sink / sanitizer types) * (number of referenced assemblies) = (number of calls to `IAssemblySymbol.GetTypeByMetadataName()`), by checking `IAssemblySymbol.TypeNames` first. But realizing `TypeNames` seemed to allocate more than unnecessary `IAssemblySymbol.GetTypeByMetadataName()` calls. So I'm settling on checking `NamespaceNames` instead.

What do you think of avoiding the `Compilation.GetTypeByMetadataName()` call entirely? That seemed to account for a bunch of allocations.

```
Callees of the TaintedDataSymbolMap constructor from the reported trace:

Name | Inc % | Inc
-- | -- | --
Microsoft.NetCore.Analyzers!Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis.TaintedDataSymbolMap`1[System.__Canon]..ctor(class Analyzer.Utilities.WellKnownTypeProvider,class System.Collections.Generic.IEnumerable`1) | 74.0 | 21,036,554,240.000
+ Microsoft.NetCore.Analyzers!WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName | 74.0 | 21,035,767,808.000
\|+ mscorlib.ni!System.Collections.Concurrent.ConcurrentDictionary`2[System.__Canon,System.__Canon].GetOrAdd(System.__Canon, System.Func`2) | 74.0 | 21,031,440,384.000
\|\|+ Microsoft.NetCore.Analyzers!Analyzer.Utilities.WellKnownTypeProvider.<TryGetOrCreateTypeByMetadataName>b__7_0(class System.String) | 73.9 | 21,015,066,624.000
\|\|\|+ microsoft.codeanalysis.csharp.ni!IAssemblySymbol.GetTypeByMetadataName | 58.7 | 16,677,484,544.000
\|\|\|+ microsoft.codeanalysis.ni!Compilation.GetTypeByMetadataName | 14.6 | 4,153,982,464.000
\|\|\|+ microsoft.codeanalysis.csharp.ni!IModuleSymbol.get_ReferencedAssemblySymbols | 0.6 | 179,895,840
\|\|\|+ microsoft.codeanalysis.csharp.ni!IAssemblySymbol.get_Modules | 0.0 | 3,703,220
```

I tested against a project with 999 referenced assemblies, and a handful of duplicate types to trigger the fallback logic.

Still need to see if I can get the reported codebase, and other real world code.

### DIFF
--- a/src/Utilities/Compiler/WellKnownTypeProvider.cs
+++ b/src/Utilities/Compiler/WellKnownTypeProvider.cs
@@ -61,6 +61,7 @@ namespace Analyzer.Utilities
         /// </summary>
         private readonly ConcurrentDictionary<string, INamedTypeSymbol?> _fullNameToTypeMap;
 
+#if !NETSTANDARD1_3 // Assuming we're on .NET Standard 2.0 or later, cache the type names that are probably compile time constants.
         /// <summary>
         /// Static cache of full type names (with namespaces) to namespace name parts,
         /// so we can query <see cref="IAssemblySymbol.NamespaceNames"/>.
@@ -74,6 +75,7 @@ namespace Analyzer.Utilities
         /// </remarks>
         private static readonly ConcurrentDictionary<string, string[]> _fullTypeNameToNamespaceNames =
             new ConcurrentDictionary<string, string[]>(StringComparer.Ordinal);
+#endif
 
         /// <summary>
         /// Attempts to get the type by the full type name.
--- a/src/Utilities/Compiler/WellKnownTypeProvider.cs
+++ b/src/Utilities/Compiler/WellKnownTypeProvider.cs
@@ -2,6 +2,8 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
@@ -39,12 +41,27 @@ namespace Analyzer.Utilities
         private readonly ConcurrentDictionary<string, INamedTypeSymbol?> _fullNameToTypeMap;
 
         /// <summary>
+        /// Global cache of full type names (with namespaces) to namespace name parts and simple type name (without namespace),
+        /// so we can query <see cref="IAssemblySymbol.NamespaceNames"/> and <see cref="IAssemblySymbol.TypeNames"/>.
+        /// </summary>
+        /// <remarks>
+        /// Example: "System.Collections.Generic.List`1" => ( [ "System", "Collections", "Generic" ], "List" )
+        /// 
+        /// https://github.com/dotnet/roslyn/blob/9e786147b8cb884af454db081bb747a5bd36a086/src/Compilers/CSharp/Portable/Symbols/AssemblySymbol.cs#L455
+        /// suggests the TypeNames collection can be checked to avoid expensive operations.
+        /// </remarks>
+        private static readonly ConcurrentDictionary<string, (ImmutableHashSet<string> NamespaceNames, string SimpleTypeName)> _fullTypeNameToSimpleInfo =
+            new ConcurrentDictionary<string, (ImmutableHashSet<string> NamespaceNames, string SimpleTypeName)>(StringComparer.Ordinal);
+
+        /// <summary>
         /// Attempts to get the type by the full type name.
         /// </summary>
         /// <param name="fullTypeName">Namespace + type name, e.g. "System.Exception".</param>
         /// <param name="namedTypeSymbol">Named type symbol, if any.</param>
         /// <returns>True if found in the compilation, false otherwise.</returns>
-        public bool TryGetOrCreateTypeByMetadataName(string fullTypeName, [NotNullWhen(returnValue: true)] out INamedTypeSymbol? namedTypeSymbol)
+        public bool TryGetOrCreateTypeByMetadataName(
+            string fullTypeName,
+            [NotNullWhen(returnValue: true)] out INamedTypeSymbol? namedTypeSymbol)
         {
             namedTypeSymbol = _fullNameToTypeMap.GetOrAdd(
                 fullTypeName,
@@ -58,13 +75,53 @@ namespace Analyzer.Utilities
                     var type = Compilation.GetTypeByMetadataName(fullyQualifiedMetadataName);
 #pragma warning restore RS0030 // Do not used banned APIs
 
-                    type ??= Compilation.Assembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
+                    // sharwell says: Suppose you reference assembly A with public API X.Y, and you reference assembly B with
+                    // internal API X.Y. Even though you can use X.Y from assembly A, compilation.GetTypeByMetadataName will 
+                    // fail outright because it finds two types with the same name.
+
+                    ImmutableHashSet<string>? namespaceNames = null;
+                    string? typeName = null;
                     if (type is null)
                     {
+#if NETSTANDARD1_3 // Probably in 2.9.x branch; just cache everything.
+                        (namespaceNames, typeName) = _fullTypeNameToSimpleInfo.GetOrAdd(
+                            fullTypeName,
+                            GetSimpleNameInfoFromFullTypeName);
+#else // Assuming we're on .NET Standard 2.0 or later, cache type names which are probably compile time constants.
+                        if (string.IsInterned(fullTypeName) != null)
+                        {
+                            (namespaceNames, typeName) = _fullTypeNameToSimpleInfo.GetOrAdd(
+                                fullTypeName,
+                                GetSimpleNameInfoFromFullTypeName);
+                        }
+                        else
+                        {
+                            (namespaceNames, typeName) = GetSimpleNameInfoFromFullTypeName(fullTypeName);
+                        }
+#endif
+
+                        if (IsSubsetOfCollection(namespaceNames, Compilation.Assembly.NamespaceNames)
+                            && Compilation.Assembly.TypeNames.Contains(typeName))
+                        {
+                            type = Compilation.Assembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
+                        }
+                    }
+
+                    if (type is null)
+                    {
+                        RoslynDebug.Assert(namespaceNames != null);
+                        RoslynDebug.Assert(typeName != null);
+
                         foreach (var module in Compilation.Assembly.Modules)
                         {
                             foreach (var referencedAssembly in module.ReferencedAssemblySymbols)
                             {
+                                if (!IsSubsetOfCollection(namespaceNames, referencedAssembly.NamespaceNames!)
+                                    || !referencedAssembly.TypeNames.Contains(typeName))
+                                {
+                                    continue;
+                                }
+
                                 var currentType = referencedAssembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
                                 if (currentType is null)
                                 {
@@ -125,6 +182,52 @@ namespace Analyzer.Utilities
                 && typeSymbol is INamedTypeSymbol namedTypeSymbol
                 && namedTypeSymbol.TypeArguments.Length == 1
                 && typeArgumentPredicate(namedTypeSymbol.TypeArguments[0]);
+        }
+
+        private static (ImmutableHashSet<string> NamespaceNames, string SimpleTypeName) GetSimpleNameInfoFromFullTypeName(string fullTypeName)
+        {
+            int plusIndex = fullTypeName.LastIndexOf('+');   // For nested types.
+            int dotIndex = fullTypeName.LastIndexOf('.');
+            int backTickIndex = fullTypeName.LastIndexOf('`');
+
+            int typeStartIndex = Math.Max(dotIndex, plusIndex) + 1;   // Exclude the '+' or '.'; LastIndexOf() returns -1 if not found.
+            int typeEndIndex = backTickIndex >= 0 && backTickIndex > typeStartIndex ? backTickIndex : fullTypeName.Length;
+
+            ImmutableHashSet<string> namespaceNames;
+            if (dotIndex >= 0)
+            {
+                namespaceNames = ImmutableHashSet.Create(
+                    StringComparer.Ordinal,
+                    fullTypeName
+                        .Substring(0, dotIndex >= 0 ? dotIndex : fullTypeName.Length)
+                        .Split('.'));
+            }
+            else
+            {
+                namespaceNames = ImmutableHashSet<string>.Empty;
+            }
+
+            return (namespaceNames, fullTypeName[typeStartIndex..typeEndIndex]);
+        }
+
+        // This would live in ImmutableHashSetsExtensions.cs, but Microsoft.CodeAnalysis.PublicApiAnalyzers.csproj doesn't get
+        // it.
+        private static bool IsSubsetOfCollection<T>(ImmutableHashSet<T> set1, ICollection<T> set2)
+        {
+            if (set1.Count > set2.Count)
+            {
+                return false;
+            }
+
+            foreach (T item in set1)
+            {
+                if (!set2.Contains(item))
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
     }
 }
--- a/src/Utilities/Compiler/WellKnownTypeProvider.cs
+++ b/src/Utilities/Compiler/WellKnownTypeProvider.cs
@@ -5,6 +5,8 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Threading;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 
@@ -22,6 +24,15 @@ namespace Analyzer.Utilities
         {
             Compilation = compilation;
             _fullNameToTypeMap = new ConcurrentDictionary<string, INamedTypeSymbol?>(StringComparer.Ordinal);
+            _referencedAssemblies = new Lazy<ImmutableHashSet<IAssemblySymbol>>(
+                () =>
+                {
+                    return ImmutableHashSet.Create<IAssemblySymbol>(
+                        Compilation.Assembly.Modules
+                            .SelectMany(m => m.ReferencedAssemblySymbols)
+                            .ToArray());
+                },
+                LazyThreadSafetyMode.ExecutionAndPublication);
         }
 
         public static WellKnownTypeProvider GetOrCreate(Compilation compilation)
@@ -34,6 +45,16 @@ namespace Analyzer.Utilities
         }
 
         public Compilation Compilation { get; }
+
+        /// <summary>
+        /// All the referenced assembly symbols.
+        /// </summary>
+        /// <remarks>
+        /// Seems to be less memory intensive than:
+        /// foreach (Compilation.Assembly.Modules)
+        ///     foreach (Module.ReferencedAssemblySymbols)
+        /// </remarks>
+        private readonly Lazy<ImmutableHashSet<IAssemblySymbol>> _referencedAssemblies;
 
         /// <summary>
         /// Mapping of full name to <see cref="INamedTypeSymbol"/>.
@@ -50,8 +71,8 @@ namespace Analyzer.Utilities
         /// https://github.com/dotnet/roslyn/blob/9e786147b8cb884af454db081bb747a5bd36a086/src/Compilers/CSharp/Portable/Symbols/AssemblySymbol.cs#L455
         /// suggests the TypeNames collection can be checked to avoid expensive operations.
         /// </remarks>
-        private static readonly ConcurrentDictionary<string, (ImmutableHashSet<string> NamespaceNames, string SimpleTypeName)> _fullTypeNameToSimpleInfo =
-            new ConcurrentDictionary<string, (ImmutableHashSet<string> NamespaceNames, string SimpleTypeName)>(StringComparer.Ordinal);
+        private static readonly ConcurrentDictionary<string, (string[] NamespaceNames, string SimpleTypeName)> _fullTypeNameToSimpleInfo =
+            new ConcurrentDictionary<string, (string[] NamespaceNames, string SimpleTypeName)>(StringComparer.Ordinal);
 
         /// <summary>
         /// Attempts to get the type by the full type name.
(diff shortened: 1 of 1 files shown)


src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -20,6 +24,15 @@ private WellKnownTypeProvider(Compilation compilation)
         {
             Compilation = compilation;
             _fullNameToTypeMap = new ConcurrentDictionary<string, INamedTypeSymbol?>(StringComparer.Ordinal);
+            _referencedAssemblies = new Lazy<ImmutableHashSet<IAssemblySymbol>>(
+                () =>
+                {
+                    return ImmutableHashSet.Create<IAssemblySymbol>(
```
dotpaul, line 30: I was thinking, can different modules in the same compilation reference the same assembly? That might have made the original fallback logic not work, since we'd visit the same assembly more than once.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -126,5 +205,45 @@ internal bool IsTaskOfType([NotNullWhen(returnValue: true)] ITypeSymbol? typeSym
                 && namedTypeSymbol.TypeArguments.Length == 1
                 && typeArgumentPredicate(namedTypeSymbol.TypeArguments[0]);
         }
+
+        private static (string[] NamespaceNames, string SimpleTypeName) GetSimpleNameInfoFromFullTypeName(string fullTypeName)
```
dotpaul, line 209: I'm sure there are weird cases I'm not aware of for this.

mavasani:
I think before going into the specifics of this PR, I would like to step back on the current implementation of GetOrCreateTypeByMetadataName, more specifically the additional processing done on top of `compilation.GetTypeByMetadataName` here https://github.com/dotnet/roslyn-analyzers/blob/f248a4cd228c3c2442784ee1936397244a1d5d84/src/Utilities/Compiler/WellKnownTypeProvider.cs#L61-L93

1. @paulomorgado was the first one to identify the need for any additional processing on top of `compilation.GetTypeByMetadataName`, when he saw that the API returns null for `ImmutableArray`, when executed _in context of  unit tests_, though the API worked fine when the analyzer was executed on real projects.
2. @sharwell generalized the solution in https://github.com/dotnet/roslyn-analyzers/commit/7049a4b63b6c558ab877a8cdfab5da29b3c751e1 so our helper method `GetOrCreateTypeByMetadataName` handles multiple type definitions in separate reference assemblies, where only one is _visible_ inside the compilation.

I locally reverted https://github.com/dotnet/roslyn-analyzers/commit/7049a4b63b6c558ab877a8cdfab5da29b3c751e1 on latest master sources (basically delete all the code from Line 61-93 in WellKnownTypeProvider.cs) and all unit tests passed. This indicates that the original issue that was seen by @paulomorgado no longer exists in latest test framework. Additionally, I don't believe we really need to handle the case of _multiple type definitions in separate reference assemblies, where only one is visible inside the compilation_, that is not common, certainly not for the well known types we care about in this repo. If at all this functionality is deemed important, it should be implemented within Roslyn, either in the existing `compilation.GetTypeByMetadataName` API or a new API on `Compilation`, as a Roslyn feature.

I would like to propose that we revert https://github.com/dotnet/roslyn-analyzers/commit/7049a4b63b6c558ab877a8cdfab5da29b3c751e1 to address this performance issue. @dotpaul, can you quickly measure the perf impact on your test solution after the revert on top of master sources?

mavasani:
Another thing to note from https://github.com/dotnet/roslyn-analyzers/issues/3870#issuecomment-658938338. All the large number of allocations are coming from the `IAssemblySymbol.GetTypeByMetadataName` in this additional code. Invoking `Compilation.GetTypeByMetadataName` is not showing up on traces, indicating the implementation for the security analyzers as well as `Compilation.GetTypeByMetadataName` is just fine in terms of performance.

Let us not try to complicate the well known type detection logic in this repo for a seemingly corner case at the risk of introducing further performance and/or functional regressions. Any functionality or performance improvements desired in this space should be proposed and implemented directly within Roslyn.

sharwell:
@mavasani We definitely rely on the fallback behavior in edge cases, so removing it entirely isn't viable. If we were only using these analyzers in one project, we could leverage information about that project, but these are much more broadly used.

sharwell:
> I locally reverted 7049a4b on latest master sources ... and all unit tests passed

This is because our test cases are much more restricted with dotnet/roslyn-sdk#329. The scenarios where this applies in production code are typically lightup scenarios. One particularly-common case is the nullability attributes.

mavasani:
> We definitely rely on the fallback behavior in edge cases, so removing it entirely isn't viable. 

I don't see any tests. Can you point out what analyzers/well known types rely on this behavior? If it is specific to some particular analyzer package, can we `#if` the code to only execute for that package OR special case/hard code such ambiguous well known type names in this method and execute the additional code only for such queries? I don't see any need for introducing this performance overhead for 100% of cases when 1% of cases seem to be requiring it.

mavasani:
>  The scenarios where this applies in production code are typically lightup scenarios.

None of the CA rules use lightup. Again, there is no test case or a repro-able manual scenario where I can validate your assertion. Without either of these, we should revert the change, and then add it back for targeted scenarios or packages.

sharwell:
> I don't see any need for introducing this performance overhead for 100% of cases when 1% of cases seem to be requiring it.

The performance overhead is generally negligible. We use this helper _everywhere_ but _only one_ analyzer has a resulting performance problem.

> None of the CA rules use lightup.

I meant API porting scenarios. For example, a project using https://github.com/tunnelvisionlabs/dotnet-httpclient35 instead of https://www.nuget.org/packages/System.Net.Http/.

The specific case where this is useful is APIs that are not provided by all runtimes but could be substituted by a user. The implementation follows the logic that the compiler itself uses in answering the question "is this API available", which is exactly the test we're trying to perform.

sharwell:
💭 It may be possible to implement the fallback using our own resolution strategy based on `INamespaceSymbol.GetMembers(string)`. If this method can return multiple items with the same name (e.g. when they are present in multiple assemblies), then it should work for our purposes.

src/Utilities/Compiler/WellKnownTypeProvider.cs
sharwell: It's an interesting point, but I think in most cases a compilation only has one module.

mavasani:
> The performance overhead is generally negligible.

This is not true. Performance overhead has shown to be significant for any analyzer that request large number of well known types.

> We use this helper everywhere but only one analyzer has a resulting performance problem.

This is the whole bucket of taint analysis based security analyzers, not just one analyzer.

> The specific case where this is useful is APIs that are not provided by all runtimes but could be substituted by a user. 

Sure, but that does not explain the scenario where APIs exist in multiple reference assemblies. No one should be shipping a reference assembly with a non-private type duplicating a well-known system type. I feel we are hurting the performance for an unlikely scenario.

> The implementation follows the logic that the compiler itself uses in answering the question "is this API available", which is exactly the test we're trying to perform.

I don't think individual analyzers should attempt to inject custom well-known type searching logic. If at all, this functionality should be added to the Roslyn APIs. There is nothing special about our analyzers that need custom logic over on top of other clients of `compilation.GetTypeByMetadataName`.

I am also wondering if there has been any real user request or feedback about the scenario you are trying to justify. Analyzer packages from this repo and others have relied on `compilation.GetTypeByMetadataName` for a very long time, and I don't recall seeing a single user feedback about this.

> It may be possible to implement the fallback using our own resolution strategy based on INamespaceSymbol.GetMembers(string). If this method can return multiple items with the same name (e.g. when they are present in multiple assemblies), then it should work for our purposes.

I think we should open an API feature request in Roslyn explaining the problem, the scenarios that are affected and potential approaches to solve the problem.

sharwell:
> Performance overhead has shown to be significant for any analyzer that request large number of well known types.

The impact of this method in the context of an analyzer project is limited to two specific cases:

1. The analyzer requests a type which is present in multiple locations
1. The analyzer requests a type which is not present

The first case is the one where this algorithm is required, so there's not much we can do there. The second one is the case we're looking to optimize.

> No one should be shipping a reference assembly with a non-private type duplicating a well-known system type.

The problem does not occur for strictly non-private types. The problem occurs when an assembly defines an `internal` type and there is no IVT that gives access to the current compilation.

> I think we should open an API feature request in Roslyn explaining the problem, the scenarios that are affected and potential approaches to solve the problem.

Yes, this would be helpful.

mavasani:
> The impact of this method in the context of an analyzer project is limited to two specific cases:
- The analyzer requests a type which is present in multiple locations
- The analyzer requests a type which is not present
The first case is the one where this algorithm is required, so there's not much we can do there. The second one is the case we're looking to optimize.

This does not seem to be a scenario specific to our analyzers. In fact it does not even seem to be specific to analyzers, it applies to all clients of `compilation.GetTypeByMetadataName`. If you strongly feel the scenario is as important as you suggest, it certainly needs to be implemented in Roslyn. Deferring will most likely lead to it never getting implemented in Roslyn and we will hit some other perf issues in analyzers in future.

mavasani:
> The problem occurs when an assembly defines an internal type and there is no IVT that gives access to the current compilation.

Can you point to any one shipped and supported assembly/package that contains a duplicate for any type defined in https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/WellKnownTypeNames.cs that needs us to support this custom well-known type detection?

dotpaul:
I also agree that the Roslyn API is the best place to address the duplicate type issues, as well as any allocation optimizations. It's pretty reasonable for a user of the Roslyn API to expect `Compilation.GetTypeByMetadataName()` to resolve types in the same way the compiler does.

> Another thing to note from #3870 (comment). All the large number of allocations are coming from the IAssemblySymbol.GetTypeByMetadataName in this additional code. Invoking Compilation.GetTypeByMetadataName is not showing up on traces, indicating the implementation for the security analyzers as well as Compilation.GetTypeByMetadataName is just fine in terms of performance.

Looking at inclusive allocations from WellKnownTypeProvider, `Compilation.GetTypeByMetadataName()` shows up:

```
From the reported trace
Callees of WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName value factory lambda
GC Heap Alloc Ignore Free (Coarse Sampling)

    Name                                                                                                                               	Inc %	              Inc	  Inc Ct
    Microsoft.NetCore.Analyzers!Analyzer.Utilities.WellKnownTypeProvider.<TryGetOrCreateTypeByMetadataName>b__7_0(class System.String)	 77.5	22,029,457,408.000	 201,139
--> + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.GetTypeByMetadataName                                                           	 59.6	16,938,093,568.000	 154,642
--> + microsoft.codeanalysis.ni!Compilation.GetTypeByMetadataName                                                                      	 17.3	 4,905,042,432.000	  44,796
    + microsoft.codeanalysis.csharp.ni!IModuleSymbol.get_ReferencedAssemblySymbols                                                     	  0.6	       182,556,256	   1,666
    + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.get_Modules                                                                     	  0.0	         3,809,716	      35
```

(Did you know copying and pasting from PerfView into Visual Studio Code gets you reasonably formatted plain text? That's neat.)

> @dotpaul, can you quickly measure the perf impact on your test solution after the revert on top of master sources?

Sure, here's my results (take with a few shakes of salt, since it's a pretty artificial scenario):

```
From my artificial testing with ~3.3.0-beta2.finalish code
Callees of WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName value factory lambda
GC Heap Alloc Ignore Free

    Name                                                                                                                                                                   	Inc %	          Inc	  Inc Ct
    microsoft.codeanalysis.netanalyzers!Analyzer.Utilities.WellKnownTypeProvider.<TryGetOrCreateTypeByMetadataName>b__7_0(class System.String)                            	 56.9	   30,248,124	 277,836
--> + microsoft.codeanalysis.ni!Compilation.GetTypeByMetadataName                                                                                                          	 49.0	   26,030,980	 225,533
--> + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.GetTypeByMetadataName                                                                                               	  7.5	    3,990,240	  51,204
    + microsoft.codeanalysis.csharp.ni!IModuleSymbol.get_ReferencedAssemblySymbols                                                                                         	  0.4	      225,304	   1,066
    + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.get_Modules                                                                                                         	  0.0	        1,344	      24
    + microsoft.codeanalysis.netanalyzers!ISymbolExtensions.GetResultantVisibility                                                                                         	  0.0	          144	       6
    + clr!ThePreStub                                                                                                                                                       	  0.0	           64	       1
--> + microsoft.codeanalysis.csharp.ni!CSharpCompilation.get_CommonAssembly                                                                                                	  0.0	           24	       1
    + microsoft.codeanalysis.csharp.ni!Microsoft.CodeAnalysis.CSharp.Symbols.PublicModel.AssemblySymbol+<Microsoft-CodeAnalysis-IAssemblySymbol-get_Modules>d__5.MoveNext()	  0.0	           24	       1



From my artificial testing without fallback logic (commented out lines 61-93 in WellKnownTypeProvider.cs)
Callees of WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName value factory lambda
GC Heap Alloc Ignore Free

    Name                                                                                                                                       	Inc %	          Inc	  Inc Ct
    microsoft.codeanalysis.netanalyzers!Analyzer.Utilities.WellKnownTypeProvider.<TryGetOrCreateTypeByMetadataName>b__7_0(class System.String)	 53.0	   25,883,500	 224,307
--> + microsoft.codeanalysis.ni!Compilation.GetTypeByMetadataName                                                                              	 53.0	   25,883,500	 224,307



From my artificial testing of this initial PR
Callees of WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName value factory lambda
GC Heap Alloc Ignore Free

    Name                                                                                                                                                                                    	Inc %	         Inc	 Inc Ct
    microsoft.codeanalysis.netanalyzers!Analyzer.Utilities.WellKnownTypeProvider+<>c__DisplayClass9_0.<TryGetOrCreateTypeByMetadataName>b__0(class System.String)                                19.9      5,924,190     77,681
--> + microsoft.codeanalysis.csharp.ni!CSharpCompilation.get_CommonAssembly                                                                                                                 	 10.0	   2,990,218	 38,928
    |+ microsoft.codeanalysis.csharp.ni!Microsoft.CodeAnalysis.CSharp.CSharpCompilation+ReferenceManager.CreateSourceAssemblyForCompilation(Microsoft.CodeAnalysis.CSharp.CSharpCompilation)	 10.0	   2,990,194	 38,927
    |+ microsoft.codeanalysis.csharp.ni!Microsoft.CodeAnalysis.CSharp.Symbols.SymbolExtensions.GetPublicSymbol[System.__Canon](Microsoft.CodeAnalysis.CSharp.Symbol)                        	  0.0	          24	      1
--> + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.GetTypeByMetadataName                                                                                                                	  5.2	   1,547,016	 14,977
--> + microsoft.codeanalysis.csharp!MetadataOrSourceAssemblySymbol.get_NamespaceNames                                                                                                       	  3.9	   1,150,480	 20,093
    + mscorlib.ni!System.Lazy`1[System.__Canon].LazyInitValue()                                                                                                                             	  0.4	     133,672	  2,112
    + microsoft.codeanalysis.netanalyzers!WellKnownTypeProvider.IsSubsetOfCollection                                                                                                        	  0.2	      67,104	  1,053
    + mscorlib!System.Collections.Concurrent.ConcurrentDictionary`2[System.__Canon,System.ValueTuple`2[System.__Canon,System.__Canon]].GetOrAdd(!0,class System.Func`2)                     	  0.1	      31,976	    461
    + system.collections.immutable.ni!System.Collections.Immutable.ImmutableHashSet`1[System.__Canon].GetEnumerator()                                                                       	  0.0	       2,948	     28
    + microsoft.codeanalysis.netanalyzers!ISymbolExtensions.GetResultantVisibility                                                                                                          	  0.0	         576	     24
    + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.GivesAccessTo                                                                                                                        	  0.0	          72	      3
    + clr!JIT_New                                                                                                                                                                           	  0.0	          64	      1
    + clr!ThePreStub                                                                                                                                                                        	  0.0	          64	      1
```

(I don't get why `CSharpCompilation+ReferenceManager.CreateSourceAssemblyForCompilation()` gets called in the last case, but not previously.)

mavasani:
Nice! So, it seems the optimization helps even reducing the allocations on `Compilation.GetTypeByMetadataName`. That convinces me of @sharwell's plan to retain the optimization here, while at the same time opening an issue on Roslyn to take these performance optimization into `Compilation.GetTypeByMetadataName` itself.

src/Utilities/Compiler/WellKnownTypeProvider.cs
dotpaul: @mavasani @sharwell I'll change this to only try to get the namespace names, but know of any good unit tests in the compiler for parsing out namespace names from a fully qualified type name?

dotpaul:
Ah, I did find `CSharpCompilation+ReferenceManager.CreateSourceAssemblyForCompilation()`:

```
From my artificial testing with ~3.3.0-beta2.finalish code
Callees of WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName value factory lambda
GC Heap Alloc Ignore Free

    Name                                                                                                                                                                                     	Inc %	          Inc	  Inc Ct
    microsoft.codeanalysis.netanalyzers!Analyzer.Utilities.WellKnownTypeProvider.<TryGetOrCreateTypeByMetadataName>b__7_0(class System.String)                                              	 56.9	   30,248,124	 277,836
    + microsoft.codeanalysis.ni!Compilation.GetTypeByMetadataName                                                                                                                            	 49.0	   26,030,980	 225,533
    |+ microsoft.codeanalysis.csharp.ni!CSharpCompilation.CommonGetTypeByMetadataName                                                                                                        	 49.0	   26,025,700	 225,401
    ||+ microsoft.codeanalysis.csharp.ni!AssemblySymbol.GetTypeByMetadataName                                                                                                                	 43.3	   23,026,576	 186,293
--> ||+ microsoft.codeanalysis.csharp.ni!Microsoft.CodeAnalysis.CSharp.CSharpCompilation+ReferenceManager.CreateSourceAssemblyForCompilation(Microsoft.CodeAnalysis.CSharp.CSharpCompilation)	  5.6	    2,991,084	  38,940
    ||+ microsoft.codeanalysis.csharp.ni!Microsoft.CodeAnalysis.CSharp.Symbols.SymbolExtensions.GetPublicSymbol[System.__Canon](Microsoft.CodeAnalysis.CSharp.Symbol)                        	  0.0	        8,040	     168
    |+ microsoft.codeanalysis.ni!Microsoft.CodeAnalysis.ConcurrentCache`2[System.__Canon,System.__Canon].TryAdd(System.__Canon, System.__Canon)                                              	  0.0	        5,280	     132
    + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.GetTypeByMetadataName                                                                                                                 	  7.5	    3,990,240	  51,204
    + microsoft.codeanalysis.csharp.ni!IModuleSymbol.get_ReferencedAssemblySymbols                                                                                                           	  0.4	      225,304	   1,066
    + microsoft.codeanalysis.csharp.ni!IAssemblySymbol.get_Modules                                                                                                                           	  0.0	        1,344	      24
    + microsoft.codeanalysis.netanalyzers!ISymbolExtensions.GetResultantVisibility                                                                                                           	  0.0	          144	       6
    + clr!ThePreStub                                                                                                                                                                         	  0.0	           64	       1
    + microsoft.codeanalysis.csharp.ni!CSharpCompilation.get_CommonAssembly                                                                                                                  	  0.0	           24	       1
    + microsoft.codeanalysis.csharp.ni!Microsoft.CodeAnalysis.CSharp.Symbols.PublicModel.AssemblySymbol+<Microsoft-CodeAnalysis-IAssemblySymbol-get_Modules>d__5.MoveNext()                  	  0.0	           24	       1
```

Also, to attack the other part of the equation (number of tainted data source types + maybe sink / sanitizer types), I'll see if I can only check for less types to determining whether to run the analyzer at all. That'll help in the case where no types are available. Manish also had an idea to defer resolving types until later, but that'll be more involved. In any case, if either of those things happen, they'll be in separate PRs.

src/Utilities/Compiler/WellKnownTypeProvider.cs
dotpaul: I guess this still might have the same issue if we get back different IAssemblySymbol objects for the same assembly for some reason, but I'm not going to worry about it. 😄

src/Utilities/Compiler/WellKnownTypeProvider.cs
dotpaul: I think what I have now is good enough.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -126,5 +191,71 @@ internal bool IsTaskOfType([NotNullWhen(returnValue: true)] ITypeSymbol? typeSym
...
+                switch (fullTypeName[i])
+                {
+                    case '.':
+                        lastDotIndex = i;
+                        dotCount++;
+                        break;
+
+                    // Dunno if Compilation.GetTypeByMetadataName() or IAssemblySymbol.GetTypeByMetadataName() will ever 
+                    // support something like "System.Collections.Generic.List<System.String>", but if they do, we'll be ready.
+                    case '`':
```
mavasani, line 210: This list makes the code seem fragile. Should we consider the reverse where we process all known characters, such as valid identifier character and `.`, and bail out on everything else? Probably something like this for the former: http://sourceroslyn.io/#Microsoft.CodeAnalysis/InternalUtilities/UnicodeCharacterUtilities.cs,51

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -126,5 +191,71 @@ internal bool IsTaskOfType([NotNullWhen(returnValue: true)] ITypeSymbol? typeSym
...
+            ExitLoop:
+            if (dotCount == 0)
+            {
+                return Array.Empty<string>();
+            }
+
+            string[] namespaceNames = new string[dotCount];
+            int namespaceIndex = 0;
+            int prevStartIndex = 0;
+            for (int i = 0; i <= lastDotIndex; i++)
```
mavasani, line 231: We can potentially avoid the second loop if we use and update an array builder for namespaceNames and a string builder for current namespace name in the prior loop, promoting the current namespace name into `namespaceNames` whenever a `.` is encountered.

src/Utilities/Compiler/WellKnownTypeProvider.cs
dotpaul: Sure, I'll do that. I was copying http://sourceroslyn.io/#Microsoft.CodeAnalysis/MetadataReader/MetadataHelpers.cs,77 before, but I agree known characters will be more robust in the future.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -126,5 +193,117 @@ internal bool IsTaskOfType([NotNullWhen(returnValue: true)] ITypeSymbol? typeSym
...
+                // FormattingChar
+                case UnicodeCategory.Format:
+                    return true;
+
+                default:
+                    return false;
+            }
+        }
+
+        private static bool IsSubsetOfCollection<T>(ImmutableArray<T> set1, ICollection<T> set2)
```
sharwell, line 291: 😕 I'm confused. It seems like `set1` would be something like this for `System.Diagnostics.Debug`:

```
ImmutableArray.Create("System", "Diagnostics")
```

But `set2` would contain things like

```
System
System.Diagnostics
...
```

Am I missing something here?

dotpaul: If an assembly has:

```
System
System.Diagnostics
System.Threading
System.Threading.Tasks
```

Then its `IAssemblySymbol.NamespaceNames` would have:

```
System
Diagnostics
Threading
Tasks
```

sharwell: Well that is horrifying

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -33,62 +48,114 @@ static WellKnownTypeProvider CreateWellKnownTypeProvider(Compilation compilation
...
-                    type ??= Compilation.Assembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
                     if (type is null)
                     {
-                        foreach (var module in Compilation.Assembly.Modules)
+                        RoslynDebug.Assert(namespaceNames != null);
+
+                        foreach (IAssemblySymbol? referencedAssembly in _referencedAssemblies.Value)
                         {
-                            foreach (var referencedAssembly in module.ReferencedAssemblySymbols)
+                            if (!IsSubsetOfCollection(namespaceNames, referencedAssembly.NamespaceNames!))
```
sharwell, line 131:
```suggestion
                            if (!IsSubsetOfCollection(namespaceNames, referencedAssembly.NamespaceNames))
```

src/Utilities/Compiler/WellKnownTypeProvider.cs
dotpaul: ¯\_(ツ)_/¯

dotpaul:
Thank you @mavasani and @sharwell !

PR #3882, opened 2020-07-20
Update CA1812 to not report on vbnet static-like holder type

### LINKED ISSUE #1878: "Avoid uninstantiated internal classes" provides fires on and provides nonsensical suggestions for VB static holder types
VB doesn't have the concept of a "static" class; they have modules which are very similar but not the same, in that module's members are automatically imported into every class that imports the namespace it belows too.

Below is a very common pattern and is similar to the bug filed in https://github.com/dotnet/roslyn/issues/30882.

#### Repro steps

1

``` VB
    ''' <summary>
    ''' Utilities related to DTE projects and project items
    ''' </summary>
    ''' <remarks></remarks>
    Friend NotInheritable Class DTEUtils

        Private Const PROJECTPROPERTY_BUILDACTION As String = "BuildAction"

        ''' <summary>
        ''' This is a shared class - disallow instantiation.
        ''' </summary>
        ''' <remarks></remarks>
        Private Sub New()
        End Sub

        ''' <summary>
        ''' Retrieves the given project item's property, if it exists, else Nothing
        ''' </summary>
        ''' <param name="PropertyName">The name of the property to retrieve.</param>
        ''' <returns></returns>
        ''' <remarks></remarks>
        Public Shared Function GetProjectItemProperty(ProjectItem As ProjectItem, PropertyName As String) As [Property]
            If ProjectItem.Properties Is Nothing Then
                Return Nothing
            End If

            For Each Prop As [Property] In ProjectItem.Properties
                If Prop.Name.Equals(PropertyName, StringComparison.OrdinalIgnoreCase) Then
                    Return Prop
                End If
            Next

            Return Nothing
        End Function
   End Class
```

#### Expected behavior
No warning

#### Actual behavior

```
Severity	Code	Description	Project	Project Rank	File	Line	Category	Suppression State	Tool	Detail Description
Warning	CA1812	DTEUtils is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it static (Shared in Visual Basic).	Microsoft.VisualStudio.AppDesigner	2	E:\project-system2\src\Microsoft.VisualStudio.AppDesigner\Common\DTEUtils.vb	13	Performance	Active	Microsoft.CodeQuality.Analyzers	An instance of an assembly-level type is not created by code in the assembly.

```

Evangelink:
Fix #1878


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/StaticHolderTypes.cs
```diff
@@ -73,13 +73,18 @@ public override void Initialize(AnalysisContext context)
         private static void AnalyzeSymbol(SymbolAnalysisContext context)
         {
             var symbol = (INamedTypeSymbol)context.Symbol;
+
             if (!symbol.IsStatic &&
```
Evangelink, line 77: I have reordered the checks to postpone more expensive checks to later in the code. I am not sure if `IsStaticHolderType` is more expensive than `MatchesConfiguredVisibility` or not so feel free to tell me if I shall swap them.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/StaticHolderTypes.cs
```diff
@@ -73,13 +73,18 @@ public override void Initialize(AnalysisContext context)
...
-                symbol.MatchesConfiguredVisibility(context.Options, Rule, context.Compilation, context.CancellationToken) &&
+                !symbol.IsAbstract &&
+                !IsSealedAndVisualBasic(symbol) &&
```
Evangelink, line 79: In vb.net the rule recommends to mark the class as "sealed" (`NotInheritable`) because `static` doesn't exist, so if the class is already marked as such, bail-out before doing more complex process. Note that the change is required because before the `IsStaticHolderType` was returning false for a vbnet static-like type holder causing `CA1812` to report FPs.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1511,6 +1511,87 @@ public void M()
...
+
+    Public Shared Sub M()
+    End Sub
+
+    Public Shared Property P As Integer
+
+    Public Shared Event ThresholdReached As EventHandler
+End Class
+
+Friend Class C4
```
Evangelink, line 1557: The static-like type holder considers this type as correct (although it will report `CA1052` to suggest adding `NotInheritable`) so we don't report `CA1812`.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1511,6 +1511,87 @@ public void M()
...
+    End Sub
+
+    Public Shared Function GetSomething(o As Object) As String
+        Return o.ToString()
+    End Function
+End Class
+
+Friend NotInheritable Class C5
+
+    Public Sub New()
```
Evangelink, line 1569: The helper is not considering the default constructor visibility and there was no note on if there was a technical motivation to do so, hence I have decided not to change the behavior. We could revisit if we decide we want to have `CA1812` on this type.

Evangelink: @sharwell You wrote the helper, was there any technical motivation?

Evangelink:
Ping @mavasani

src/Utilities/Compiler/Extensions/INamedTypeSymbolExtensions.cs
```diff
@@ -139,8 +139,10 @@ public static bool IsStaticHolderType(this INamedTypeSymbol symbol)
                 return false;
             }
 
-            // Sealed objects are presumed to be non-static holder types
-            if (symbol.IsSealed)
+            // Sealed objects are presumed to be non-static holder types for C#.
+            // In VB.NET the type cannot be static and guidelines favor having a sealed (NotInheritable) type
+            //  to act as static holder type.
+            if (symbol.IsSealed && symbol.Language == LanguageNames.CSharp)
```
mavasani, line 145: Do you need to do changes to both these extension method and the analyzer? Can we not just do one of them?

Evangelink: I need both of them because the helper is a little misleading, it's actually `IsStaticHolderLikeType` (or something like this). The helper checks that the member looks like a static holder but disregard the fact it is static to allow for the rule to report.

So in the end, the change here is just to prevent judging the type as not static-holder like when it's sealed. And the change in `StaticHolderTypes.cs` is made to ensure that if the class is sealed and it's vbnet we don't suggest to mark the class as static.

PR #3884, opened 2020-07-20
Update CA1822 to not report on auto-property

### LINKED ISSUE #2834: CA1822 fires on a get-only property that has an attribute on the getter
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

[CA1822](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1822-mark-members-as-static)

#### Repro steps

1. Create a new console application in .NET Framework or .NET Core
2. Add the following code to a new file called Example.cs
```
using System.Diagnostics;

namespace ConsoleApp
{
    public sealed class Example
    {
        public Example(string value) => Value = value;

        public string Value
        {
            [DebuggerStepThrough]
            get;
        }

        public override string ToString() => Value;
    }
}
```

#### Expected behavior
CA1822 should not trigger on Example.Value.

#### Actual behavior
Receive a C1822 diagnostic that Value does not access instance data.

Evangelink:
Fix #2834


Evangelink:
Ping @mavasani

mavasani:
Would it be reasonable to skip all auto-properties, regardless of whether or not they have attributes?

Evangelink:
Yes! That's actually what we are doing, the PR title was incorrect. Basically, there is no callback for auto-properties without attributes so they are already excluded. This PR handles the case where there is a callback for an auto-property with an attribute.

Evangelink:
Ping @mavasani

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -1178,6 +1178,48 @@ End Function
             }.RunAsync();
         }
 
+        [Fact]
+        public async Task Properties_NoDiagnostic()
```
mavasani, line 1182:
```suggestion
        public async Task AutoProperties_NoDiagnostic()
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -1178,6 +1178,48 @@ End Function
...
         }
 
+        [Fact]
+        public async Task AutoProperties_NoDiagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+public class C1
```
mavasani, line 1188: Do we have a test to verify the following property is flagged:

```csharp
using System.Diagnostics;

public class C1
{
    private static int s_field;
    public int P1
    {
        get { return s_field; }
        set { s_field = value; }
    }
}
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -254,6 +255,17 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
...
+            // Auto-properties (readonly, get, set) marked with an attribute have a block context callback while they don't
+            // if there is no attribute (see https://github.com/dotnet/roslyn/issues/46132).
+            // We consider that auto-property have the intent to always be instance members so we want to workaround this issue
+            // by bailing-out if this is an accessor, there are some operations in the block but they are all "none" operations.
+            if (methodSymbol.IsAccessorMethod()
+                && !blockOperations.IsEmpty
+                && !blockOperations.Where(op => op.Kind != OperationKind.None).Any())
+            {
+                return false;
+            }
```
mavasani, line 267: Will this break if https://github.com/dotnet/roslyn/issues/46132 is fixed?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -254,6 +255,17 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
                 return false;
             }
 
+            // Auto-properties (readonly, get, set) marked with an attribute have a block context callback while they don't
```
mavasani, line 258: Can this added code block be moved to an extension method?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -254,6 +255,17 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
...
+            // if there is no attribute (see https://github.com/dotnet/roslyn/issues/46132).
+            // We consider that auto-property have the intent to always be instance members so we want to workaround this issue
+            // by bailing-out if this is an accessor, there are some operations in the block but they are all "none" operations.
+            if (methodSymbol.IsAccessorMethod()
```
mavasani, line 262: Can we file a Roslyn feature request to expose `IMethodSymbol.IsAutoProperty` as a public API?

mavasani:
@Evangelink I can merge once you are able to add test + move added code to extension method. Thanks!

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
Evangelink: We only have the setter here: https://github.com/dotnet/roslyn-analyzers/blob/master/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs#L173

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: I am not 100% sure but IMO it shouldn't break. Current problem is that we don't have the callback for the other cases so:

- Solution 1: We no longer have any callback, then this logic becomes useless. I don't think Roslyn team will go in this direction.

- Solution 2: We now have callbacks also for auto-properties, I would assume the following snippet to remain valid and to also work for attributeless auto-properties.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: Sure

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: See https://github.com/dotnet/roslyn/issues/46682

src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
```diff
@@ -63,6 +64,19 @@ static bool IsSingleStatementBody(IBlockOperation body)
...
             return false;
         }
+
+        /// <summary>
+        /// Auto-properties (readonly, get, set) marked with an attribute have a block context callback while they don't if
+        /// there is no attribute (see https://github.com/dotnet/roslyn/issues/46132).
+        /// Auto-properties only have 'OperationKind.None' operations.
+        /// </summary>
+#pragma warning disable RS1012 // Start action has no registered actions.
+        public static bool IsAutoProperty(this OperationBlockStartAnalysisContext context)
```
mavasani, line 74: Lets change this to completely symbol based logic as discussed in https://github.com/dotnet/roslyn/issues/46682. Given we will not have access to access to new API added in https://github.com/dotnet/roslyn/commit/83ecbec0733d65f892f48c090120a923512b4652, we can look for all the field symbols on the property's containing type and check if any field's associated symbol is the property symbol.

```csharp
public static bool IsAutoPropertyAccessor(this IMethodSymbol method)
    => methodSymbol.IsPropertyAccessor() && methodSymbol.AssociatedSymbol is IPropertySymbol propertySymbol && methodSymbol.ContainingType.GetMembers().OfType<IFieldSymbol>.Any(f => propertySymbol.Equals(f.AssociatedSymbol);
```

mavasani: I feel this would be much more stable logic then relying on operation tree shape and root operation node kinds. Please make sure to move this method over to `IMethodSymbolExtensions`.

Evangelink: Thank you @mavasani that's indeed a lot clearer and a more robust approach.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -254,7 +254,10 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
                 return false;
             }
 
-            return true;
+            // Auto-properties (readonly, get, set) marked with an attribute have a block context callback while they don't
+            // if there is no attribute (see https://github.com/dotnet/roslyn/issues/46132).
+            // We consider that auto-property have the intent to always be instance members so we want to workaround this issue.
+            return !methodSymbol.IsAutoPropertyAccessor();
```
mavasani, line 260:
```suggestion
            // We consider that auto-property have the intent to always be instance members, so we explicitly skip them.
            return !methodSymbol.IsAutoPropertyAccessor();
```

src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
```diff
@@ -2,6 +2,7 @@
 
 #if HAS_IOPERATION
 
+using System.Linq;
```
mavasani, line 5: Revert?

Evangelink:
Sorry I broke some tests, give me a couple of minutes xD

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: Isn't it better to keep track of the ticket? Because depending on how this is implemented later on, it's possible we will no longer have the callback (or on the opposite we always have the callback).

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -1178,6 +1178,95 @@ End Function
...
+
+    public string P2 { get; }
+
+    public string P3
+    {
+        [DebuggerStepThrough]
+        get;
+    }
+
+    public string P4
```
Evangelink, line 1229: With the new implementation we will start reporting for `P4` as there is no implicitly declared field for this property (most likely because of the error). I am fine with either keeping the error or working around it by also allowing not to find an associated field for the property. Let me know your preference and I will update the code.

mavasani: Yeah, we don't care if there is a compiler error here, its fine to adjust the test to expect a diagnostic here.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
mavasani: Hmm do you still feel any operation tree shape based approach can be better then current symbol based detection approach? IMO, even if the callback was made, it seems more reasonable to not rely on operation tree shape for such detection. I'll let you make a call, feel free to ignore this suggestion.

src/Utilities/Compiler/Extensions/IPropertySymbolExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace Analyzer.Utilities.Extensions
+{
+    public static class IPropertySymbolExtensions
```
mavasani, line 8:
```suggestion
    internal static class IPropertySymbolExtensions
```

nietras:
In VS2022 running Code Cleanup with 
```
dotnet_diagnostic.CA1822.severity = error
```
I saw a lot of properties being re-written to static still, should this not have been fixed with this PR?
```csharp
public int Value => 128;
```
rewritten to:
```csharp
public static int Value => 128;
```
in WPF view models or similar this will silently break a lot of code.

Evangelink:
Hi @nietras, please open a separate issue to discuss your issue. And to answer your question, this PR was focused on auto-properties not on readonly properties returning a constant value. Let's continue the discussion on the ticket.

PR #3889, opened 2020-07-22
Fix typos on properties in .props files

Evangelink:


Evangelink:
@sharwell Is it ok if I merge once this is green? Shall I target branch `2.9.x` instead?

sharwell:
This one seems fine for the master branch (not likely to conflict with work done separately in 2.9.x). It looks good to merge once the build passes.

Evangelink:
Interesting, looks like I don't have the "merge" right. @sharwell If you have a few minutes to merge the PR that would be great :)

PR #3893, opened 2020-07-22
Update excluded symbols API to prevent misuse

### LINKED ISSUE #3890: Dispose dataflow exclusion is not using the right API
Dispose dataflow is not using the right API to decide whether a symbol is excluded or not: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs#L107-L111

Instead the following API shall be used: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L199

As a consequence, the containing symbols will not be considered resulting in some exclusions being ignored.

Evangelink:
Fix #3890


src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -193,59 +193,6 @@ public static ImmutableArray<IParameterSymbol> GetParameters(this ISymbol? symbo
                 allowedVisibilities.Contains(symbol.GetResultantVisibility());
         }
 
-        /// <summary>
-        /// Returns true if the given source symbol has been configured to be excluded from analysis by options.
-        /// </summary>
-        public static bool IsConfiguredToSkipAnalysis(
```
Evangelink, line 199: Moved to `AnalyzerOptionsExtensions.cs`.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -272,41 +273,79 @@ static bool TryParseValue(string value, out string result)
...
+                        }
+                    }
+                }
+
+                symbol = symbol.ContainingSymbol;
+            }
+
+            return false;
+
+            static SymbolNamesWithValueOption<Unit> GetExcludedSymbolNamesWithValueOption(
```
Evangelink, line 329: Refactored as local function to ensure it's not called directly. This will prevent misuse of the API as I did.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -272,41 +273,79 @@ static bool TryParseValue(string value, out string result)
...
+                AnalyzerOptions options,
+                DiagnosticDescriptor rule,
+                ISymbol symbol,
+                Compilation compilation,
+                CancellationToken cancellationToken)
+                => TryGetSyntaxTreeForOption(symbol, out var tree)
+                    ? options.GetSymbolNamesWithValueOption<Unit>(EditorConfigOptionNames.ExcludedSymbolNames, rule, tree, compilation, cancellationToken)
+                    : SymbolNamesWithValueOption<Unit>.Empty;
+
+            static SymbolNamesWithValueOption<Unit> GetExcludedTypeNamesWithDerivedTypesOption(
```
Evangelink, line 339: Refactored as local function to ensure it's not called directly. This will prevent misuse of the API.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
```diff
@@ -12002,12 +12003,17 @@ class Test
...
-        [Fact, WorkItem(3085, "https://github.com/dotnet/roslyn-analyzers/issues/3085")]
-        public async Task LocalInvocationOfAnExcludedType_NoDiagnostic()
+        [Theory, WorkItem(3085, "https://github.com/dotnet/roslyn-analyzers/issues/3085")]
+        [InlineData("")]
+        [InlineData("dotnet_code_quality.CA2000.excluded_symbol_names = T:MyNamespace.A")]
+        [InlineData("dotnet_code_quality.excluded_symbol_names = T:MyNamespace.A")]
+        [InlineData("dotnet_code_quality.CA2000.excluded_symbol_names = N:MyNamespace")]
+        [InlineData("dotnet_code_quality.excluded_symbol_names = N:MyNamespace")]
+        [InlineData("dotnet_code_quality.CA2000.excluded_type_names_with_derived_types = T:MyNamespace.A")]
+        [InlineData("dotnet_code_quality.excluded_type_names_with_derived_types = T:MyNamespace.A")]
```
Evangelink, line 12013: These tests are failing without the change of API call.

docs/Analyzer Configuration.md
```diff
@@ -325,7 +325,48 @@ Additionally, all the dataflow analysis based rules can be configured with a sin
 
 Option Name: `excluded_type_names_with_derived_types`
 
-Configurable Rules: [CA1303](https://docs.microsoft.com/visualstudio/code-quality/ca1303)
+Configurable Rules:
```
Evangelink, line 328: Let me know if you want me to revert this change and do a separate PR for it. Basically, except for `CA2000` before this PR, this user option was already available for all the rules listed in `excluded_symbol_names` but the doc wasn't correct.

Side note, I have seen that most of the website docs are also missing this and need to be updated too.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.cs
```diff
@@ -82,7 +83,7 @@ private DisposeAnalysis(DisposeAnalysisDomain analysisDomain, DisposeDataFlowOpe
             bool exceptionPathsAnalysis,
             PointsToAnalysisKind pointsToAnalysisKind,
             bool performCopyAnalysis,
-            SymbolNamesWithValueOption<Unit> excludedSymbols,
+            Func<ISymbol, bool> isConfiguredToSkipAnalysis,
```
Evangelink, line 86: Let me know if passing a `Func` is a performance issue. Also, let me know if you prefer a delegate to have something maybe more explicit.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -272,41 +273,79 @@ static bool TryParseValue(string value, out string result)
...
-            ? options.GetExcludedSymbolNamesWithValueOption(rule, tree, compilation, cancellationToken)
-            : SymbolNamesWithValueOption<Unit>.Empty;
 
-        public static SymbolNamesWithValueOption<Unit> GetExcludedSymbolNamesWithValueOption(
-            this AnalyzerOptions options,
+        /// <summary>
+        /// Returns true if the given source symbol has been configured to be excluded from analysis by options.
+        /// </summary>
+        public static bool IsConfiguredToSkipAnalysis(
+            this ISymbol symbol,
```
mavasani, line 281: It seems weird to have extension method with symbol as the 'this' parameter in this file. Instead a better approach would be to revert the change to `ISymbolExtensions` and define new extension methods in this file with `this AnalyzerOptions options` which just calls back into the existing extension method on `ISymbol`.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.cs
mavasani: Should be fine.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
Evangelink: That's a good remark. We do have a couple of helpers in this class that take a symbol as extra argument so I will try a different signature. If you still think that's no good I will refactor and use your suggested solution.

Evangelink:
Ping @mavasani

Evangelink:
@mavasani I am pinging you here too (let me know if you don't have time).

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
mavasani: @Evangelink Can we also move `MatchesConfiguredVisibility` extension method above to the same place? This way all options related methods are consistently extension methods on AnalyzerOptions instead of ISymbol.

mavasani: Note `MatchesConfiguredVisibility` has 2 overloads.

mavasani (approved):
LGTM, thanks and sorry for the delay! I can merge once you are able to address the last suggestion.

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
Evangelink: Done

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿
```
mavasani, line 1: @Evangelink unnecessary newline at the start?

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -1,4 +1,5 @@
...
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
mavasani, line 2:
```suggestion
// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```

mavasani (approved):

PR #3898, opened 2020-07-23
Add codefix for CA1018

### LINKED ISSUE #655: Port FxCop rule CA1018: MarkAttributesWithAttributeUsage
**Title:** Mark attributes with AttributeUsageAttribute

**Description:**

When you define a custom attribute, mark it by using AttributeUsageAttribute to indicate where in the source code the custom attribute can be applied. The meaning and intended usage of an attribute will determine its valid locations in code.

**Dependency:** None

**Notes:**

Evangelink:
Fix #655


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/MarkAttributesWithAttributeUsage.Fixer.cs
```diff
@@ -0,0 +1,63 @@
...
+            var nodeToFix = root.FindNode(context.Span);
+            if (nodeToFix == null)
+            {
+                return;
+            }
+
+            var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            if (semanticModel == null ||
+                !semanticModel.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemAttributeUsageAttribute, out var attributeUsageAttributeType) ||
+                !semanticModel.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemAttributeTargets, out var attributeTargetsType))
```
Evangelink, line 36: I have decided to check for the type before registering the codefix. I have seen this approach and one where access was done from within the registered codefix, please let me know your preference.

mavasani: That seems fine.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1451,4 +1451,7 @@
   <data name="DoNotIgnoreMethodResultsMessageUserDefinedMethod" xml:space="preserve">
     <value>'{0}' calls '{1}' but does not use the value the method returns. This method is defined as a user-option. Use the result in a conditional statement, assign the result to a variable, or pass it as an argument to another method.</value>
   </data>
+  <data name="MarkAttributesWithAttributeUsageCodeFix" xml:space="preserve">
+    <value>Apply attribute 'AttributeUsageAttribute(AttributeTargets.All)'</value>
```
mavasani, line 1455: Instead of presenting user with a single option, it might be better to create a code with nested actions (see second parameter for http://sourceroslyn.io/#Microsoft.CodeAnalysis.Workspaces/CodeActions/CodeAction.cs,357). Top level code action would read `Apply AttributeUsageAttribute` and we will have a nested code action for each field of `AttributeTargets`, and the nested titles can just read `AttributeTargets.All`, `AttributeTargets.Class`, etc. `AttributeTargets.All` should be the first nested action though. WDYT?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/MarkAttributesWithAttributeUsage.Fixer.cs
```diff
@@ -0,0 +1,63 @@
...
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var nodeToFix = root.FindNode(context.Span);
+            if (nodeToFix == null)
+            {
+                return;
+            }
+
+            var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            if (semanticModel == null ||
+                !semanticModel.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemAttributeUsageAttribute, out var attributeUsageAttributeType) ||
```
mavasani, line 35: `semanticModel` cannot be null
```suggestion
            if (!semanticModel.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemAttributeUsageAttribute, out var attributeUsageAttributeType) ||
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: Ok, I guess we don't want to offer the combination options, do we?

mavasani: I think those will be too many, isn't it? If we feel there are some common ones, then it would be reasonable to show them.

Evangelink: I agree for both suggestions/remarks.

Evangelink: Not sure how to test the various sub-cases. Also I am not sure if the sub-action needs the full name (parent + current)?

mavasani: I think the test framework takes a CodeFixIndex or some such index argument, let me check.

mavasani: Sub-action name can just be the value, i.e. `AtributeTargets.All`, `Attributes.Class`, etc.

mavasani: https://github.com/dotnet/roslyn-analyzers/blob/ad82c9b0c5ad408e9cf84f831941e2c0f5152799/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/InterfaceMethodsShouldBeCallableByChildTypesTests.Fixer.cs#L370

Evangelink: I have tried with index and equivalency key but I can't manage to have the test resulting in the good output.

mavasani, line 1455:
```suggestion
    <value>Apply 'AttributeUsageAttribute'</value>
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/MarkAttributesWithAttributeUsageTests.cs
```diff
@@ -1,30 +1,81 @@
...
+        [InlineData(AttributeTargets.Property)]
+        [InlineData(AttributeTargets.ReturnValue)]
+        [InlineData(AttributeTargets.Struct)]
+        public async Task TestSimpleAttributeClass(AttributeTargets attributeTarget)
         {
-            await VerifyCS.VerifyAnalyzerAsync(@"
+            var attributeTargetsValue = "AttributeTargets." + attributeTarget.ToString();
+
+            await new VerifyCS.Test
+            {
```
mavasani, line 41: I think you also need to provide a `CodeActionIndex` value here based on the index of the `attributeTarget` field into the enumerated members of `AttributeTargets`

Evangelink: I have tried a combination of both but it's only working for the `All` value. Basically anything else but `CodeActionIndex = 0` results in no codefix. Any value for the equivalency key gives no codefix and if I combine both properties then I have errors saying that the combination isn't good.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: Alright, let me try to investigate.

mavasani:
@Evangelink I have pushed commits to fix the tests.

mavasani (approved):

mavasani:
@Evangelink I'll wait for you to take another look/make any changes before merging the PR.

Evangelink:
LGTM

PR #3907, opened 2020-07-24
Update to Microsoft.CodeAnalysis.Testing 1.0.1-beta1.20374.2

sharwell:
* Infer Fix All in Document iteration counts
* Infer generated code analysis configuration
* Use markup keys for locations in additional files when the same test uses markup keys for locations in source

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATaskTests.Fixer.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATaskTests.Fixer.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Testing;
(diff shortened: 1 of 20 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATaskTests.Fixer.cs
```diff
@@ -252,7 +253,20 @@ public async Task M()
             {
                 TestState = { Sources = { code } },
                 FixedState = { Sources = { fixedCode } },
-                BatchFixedState = { Sources = { fixAllCode } },
+                BatchFixedState =
+                {
+                    Sources = { fixAllCode },
+                    ExpectedDiagnostics =
```
Evangelink, line 259: Just wondering if we should open a ticket for this? It seems rather weird to have fixer introducing compiler error. WDYT?

sharwell: It's definitely a bug. It impacts both the C# and the VB fixer.

Evangelink: See #3910

PR #3913, opened 2020-07-27
Enable RS0037 and fix related issues

### LINKED ISSUE #3894: Enable RS0037 and fix related issues
@mavasani What about _just_ keeping the RS0037 suppression, and remove it in a follow-up PR?

_Originally posted by @sharwell in https://github.com/dotnet/roslyn-analyzers/pull/3888#issuecomment-662547329_

Evangelink:
Fix #3894 

Tagging @jcouv


jcouv:
Thanks
You could either fix the handful of oblivious APIs that remain, or suppress the warning for ` Symbol 'PredicateAnalysisEntityDataFlowOperationVisitor<TAnalysisData, TAnalysisContext, TAnalysisResult, TAbstractAnalysisValue>' uses some oblivious reference types.` (and related) and filing a follow-up issue.

Evangelink:
@jcouv the point is that I cannot find how to fix the `oblivious API`, there is no code-fix offered and when looking at the definitions, the only thing I could find would be a missing `!` on the generic definitions but adding it causes another rule (public api declared but not found) to trigger.

jcouv:
@Evangelink Those complaints about oblivious APIs must be suppressed at the moment (typically via editorconfig), as roslyn-analyzers is using netstandard libraries which are all oblivious.
For instance `IEquatable` is referenced from:
```
#region Assembly netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
// C:\Users\jcouv\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\netstandard.dll
#endregion

namespace System
{
    public interface IEquatable<T>
```

Evangelink:
@mavasani Any preference between `.editorconfig` and ruleset here?

mavasani:
> Any preference between .editorconfig and ruleset here?

.editorconfig is preferred. If needed at a specific file/folder level, .editorconfig is the only option.

Evangelink:
Maybe I got something wrong but my understanding is that we want to purely and simply disable the rule for all projects in the repo.

PR #3914, opened 2020-07-27
Add resource string analyzer/code fix for DiagnosticDescriptor arguments

### LINKED ISSUE #3525: Ideas for resources (resx) analyzers

mavasani:
Closes #3525
Adds a code fix for RS1031, RS1032 and RS1033
Additionally, if the resx file with analyzer resource strings is provided as an additional file to the analyzer, also validates the strings in the resx file.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
@@ -4,17 +4,25 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Text;
+using System.Threading;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
 using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
 using Microsoft.CodeAnalysis.ReleaseTracking;
+using Microsoft.CodeAnalysis.Text;
 
 namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
 {
+    using PooledLocalizabeStringsConcurrentDictionary = PooledConcurrentDictionary<INamedTypeSymbol, PooledConcurrentSet<(IFieldSymbol field, IArgumentOperation argument)>>;
+    using PooledResourcesDataValueConcurrentDictionary = PooledConcurrentDictionary<string, ImmutableDictionary<string, (string value, Location location)>>;
+    using PooledFieldToResourceNameAndFileNameConcurrentDictionary = PooledConcurrentDictionary<IFieldSymbol, (string nameOfResource, string resourceFileName)>;
+
     [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
     public sealed partial class DiagnosticDescriptorCreationAnalyzer : DiagnosticAnalyzer
     {
@@ -25,6 +33,9 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
         private const string IsEnabledByDefaultParameterName = "isEnabledByDefault";
         private const string DefaultSeverityParameterName = "defaultSeverity";
         private const string RuleLevelParameterName = "ruleLevel";
+
+        internal const string DefineDescriptorArgumentCorrectlyFixValue = nameof(DefineDescriptorArgumentCorrectlyFixValue);
+        private const string DefineDescriptorArgumentCorrectlyFixAdditionalDocumentLocationInfo = nameof(DefineDescriptorArgumentCorrectlyFixAdditionalDocumentLocationInfo);
 
         private static readonly ImmutableHashSet<string> CADiagnosticIdAllowedAssemblies = ImmutableHashSet.Create(
             StringComparer.Ordinal,
@@ -165,7 +176,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
             s_localizableDefineDiagnosticTitleCorrectlyMessage,
             DiagnosticCategory.MicrosoftCodeAnalysisDesign,
             DiagnosticSeverity.Warning,
-            isEnabledByDefault: false,
+            isEnabledByDefault: true,
             customTags: WellKnownDiagnosticTags.Telemetry);
 
         /// <summary>
@@ -177,7 +188,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
             s_localizableDefineDiagnosticMessageCorrectlyMessage,
             DiagnosticCategory.MicrosoftCodeAnalysisDesign,
             DiagnosticSeverity.Warning,
-            isEnabledByDefault: false,
+            isEnabledByDefault: true,
             customTags: WellKnownDiagnosticTags.Telemetry);
 
         /// <summary>
@@ -189,7 +200,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
             s_localizableDefineDiagnosticDescriptionCorrectlyMessage,
             DiagnosticCategory.MicrosoftCodeAnalysisDesign,
             DiagnosticSeverity.Warning,
-            isEnabledByDefault: false,
+            isEnabledByDefault: true,
             customTags: WellKnownDiagnosticTags.Telemetry);
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
@@ -225,8 +236,9 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
 
             context.RegisterCompilationStartAction(compilationContext =>
             {
-                INamedTypeSymbol? diagnosticDescriptorType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisDiagnosticDescriptor);
-                if (diagnosticDescriptorType == null)
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisDiagnosticDescriptor, out var diagnosticDescriptorType) ||
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisLocalizableString, out var localizableResourceType) ||
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisLocalizableResourceString, out var localizableResourceStringType))
                 {
                     return;
                 }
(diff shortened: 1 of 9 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -323,15 +392,37 @@ public override void Initialize(AnalysisContext context)
...
+            }
         }
 
         private static bool TryGetDescriptorCreateMethodAndArguments(
             IFieldInitializerOperation fieldInitializer,
             INamedTypeSymbol diagnosticDescriptorType,
-            out IMethodSymbol creationMethod,
-            out ImmutableArray<IArgumentOperation> creationArguments)
+            [NotNullWhen(returnValue: true)] out IMethodSymbol? creationMethod,
+            [NotNullWhen(returnValue: true)] out ImmutableArray<IArgumentOperation> creationArguments)
```
Evangelink, line 425: Not sure to understand what `NotNullWhen(true)` brings on this parameter.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -356,93 +447,427 @@ bool IsCreateHelper(IMethodSymbol method)
...
+            DiagnosticDescriptor descriptor,
+            IArgumentOperation argumentOperation,
+            string fixValue,
+            Location fixLocation,
+            Action<Diagnostic> reportDiagnostic)
+        {
+            // Additional location in an additional document does not seem to be preserved
+            // from analyzer to code fix due to a Roslyn bug.
+            // We workaround this bug by passing additional document file path and location span as strings.
+            // TODO: File a Roslyn bug.
```
Evangelink, line 528: Open TODO without link.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -356,93 +447,427 @@ bool IsCreateHelper(IMethodSymbol method)
...
             }
+            else
+            {
+                var span = fixLocation.SourceSpan;
+                var locationInfo = $"{span.Start};{span.Length};{fixLocation.GetLineSpan().Path}";
+                properties = properties.Add(DefineDescriptorArgumentCorrectlyFixAdditionalDocumentLocationInfo, locationInfo);
+            }
+
+            var location = argumentOperation.Syntax.GetLocation();
+            var diagnostic = Diagnostic.Create(descriptor, location, additionalLocations: additionalLocations, properties: properties);
```
Evangelink, line 544: Assuming the overload exists.

```suggestion
            var diagnostic = argumentOperation.CreateDiagnostic(descriptor, additionalLocations: additionalLocations, properties: properties);
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -356,93 +447,427 @@ bool IsCreateHelper(IMethodSymbol method)
...
+        {
+            Debug.Assert(diagnostic.Id == DiagnosticIds.DefineDiagnosticTitleCorrectlyRuleId ||
+                diagnostic.Id == DiagnosticIds.DefineDiagnosticMessageCorrectlyRuleId ||
+                diagnostic.Id == DiagnosticIds.DefineDiagnosticDescriptionCorrectlyRuleId);
 
-            if (IsMultiSentences(title) || EndsWithPeriod(title) || ContainsLineReturn(title))
+            if (diagnostic.Properties.TryGetValue(DefineDescriptorArgumentCorrectlyFixAdditionalDocumentLocationInfo, out var locationInfo))
             {
-                operationAnalysisContext.ReportDiagnostic(titleArgument.CreateDiagnostic(DefineDiagnosticTitleCorrectlyRule));
+                var index = locationInfo.IndexOf(';');
```
Evangelink, line 558: [question] Maybe this is better for performance but wouldn't it be better to do a split and work with the strings rather than doing a locate, tryparse repeat approach?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_ResourceStringsFormat.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        private static ImmutableDictionary<string, (string value, Location location)> GetOrCreateResourceMap(
+            AnalyzerOptions options,
+            string resourceFileName,
+            PooledResourcesDataValueConcurrentDictionary resourceMap,
+            CancellationToken cancellationToken)
+        {
+            Debug.Assert(HasResxAdditionalFiles(options));
+
+            resourceFileName += ".resx";
```
Evangelink, line 41: Might be worth adding an assertion that the string doesn't already end in `.resx`, don't you think?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_ResourceStringsFormat.cs
```diff
@@ -0,0 +1,104 @@
...
+                        if (indexOfDataElement < 0 ||
+                            !valueElementStr.StartsWith(valueTagPrefix, StringComparison.OrdinalIgnoreCase) ||
+                            !valueElementStr.EndsWith(valueTagSuffix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            continue;
+                        }
+
+                        var indexOfValue = indexOfDataElement +
+                                dataElementStr.IndexOf(valueElementStr, StringComparison.Ordinal) +
+                                valueTagPrefix.Length;
```
Evangelink, line 89:
```suggestion
                            dataElementStr.IndexOf(valueElementStr, StringComparison.Ordinal) +
                            valueTagPrefix.Length;
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_ResourceStringsFormat.cs
```diff
@@ -0,0 +1,104 @@
...
+                const string valueTagPrefix = @"<value>";
+                const string valueTagSuffix = @"</value>";
+                var builder = ImmutableDictionary.CreateBuilder<string, (string value, Location location)>();
+                var sourceText = file.GetText(cancellationToken);
+                var sourceTextStr = sourceText.ToString();
+                var parsedDocument = XDocument.Parse(sourceTextStr, LoadOptions.PreserveWhitespace);
+                foreach (var dataElement in parsedDocument.Descendants("data"))
+                {
+                    if (dataElement.Attribute("name")?.Value is { } name &&
+                        dataElement.Elements("value").FirstOrDefault() is { } valueElement)
```
Evangelink, line 75: [minor] You could revert the condition to reduce nesting.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/DefineDiagnosticDescriptorArgumentsCorrectlyFix.cs
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    continue;
+                }
+
+                var codeFixTitle = diagnostic.Id switch
+                {
+                    DiagnosticIds.DefineDiagnosticTitleCorrectlyRuleId => CodeAnalysisDiagnosticsResources.DefineDiagnosticTitleCorrectlyTitle,
+                    DiagnosticIds.DefineDiagnosticMessageCorrectlyRuleId => CodeAnalysisDiagnosticsResources.DefineDiagnosticMessageCorrectlyTitle,
+                    DiagnosticIds.DefineDiagnosticDescriptionCorrectlyRuleId => CodeAnalysisDiagnosticsResources.DefineDiagnosticDescriptionCorrectlyTitle,
+                    _ => throw new NotImplementedException()
```
Evangelink, line 55: I would probably use `InvalidOperationException` instead and add some explicit message containing the diagnostic ID in case we end up in this "impossible" state.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -764,6 +764,24 @@ public static bool IsSetMethodInvocation(this IPropertyReferenceOperation operat
...
+
+            foreach (var argument in arguments)
+            {
+                if (argument.Parameter.Ordinal == parameterIndex)
+                {
+                    return argument;
+                }
+            }
+
+            throw new NotImplementedException();
```
Evangelink, line 783: I would use `InvalidOperationException` to represent invalid state and I would also include the index in the message.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: I thought it ensures the array is not default?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: Yep, I need to file a Roslyn issue.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: Yes, thanks. Will change.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: I am not sure if file names/paths on different OS can allow ';' character. Your approach also seems reasonable.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: https://github.com/dotnet/roslyn/issues/46377

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: I added new overloads and made this change.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
mavasani: Done.

mavasani:
@Evangelink Thanks for the great feedback. I have resolved your comments.

mavasani:
Thanks @Evangelink!

mavasani:
Yep, that is correct. I was hoping to get to it in a day or two. If you are ready to get to it earlier, then go for it.

PR #3916, opened 2020-07-28
[PublicApiAnalyzer] Add support for creating shipped and unshiped tex…

### LINKED ISSUE #2622: PublicApiAnalyzer should be capable of creating text files
Currently the Public API Analyzer only works if you manually create `PublicAPI.Shipped.txt` and `PublicAPI.Unshipped.txt` in your project, and set their item types to `AdditionalFiles`.

This creates an unnecessary hurdle to adoption of the analyzer. See https://github.com/dotnet/roslyn-analyzers/issues/2580#issuecomment-501926424 for an example.

It should create these files lazily within the project when needed, such as when using a code fix to add a public API.

mavasani:
…t files with code fixes

Closes #2622


src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
```diff
@@ -195,13 +206,11 @@ public override void Initialize(AnalysisContext context)
...
             var additionalFiles = compilationContext.Options.AdditionalFiles;
+            var errors = new List<Diagnostic>();
 
-            if (!TryGetApiData(additionalFiles, compilationContext.CancellationToken, out ApiData shippedData, out ApiData unshippedData))
-            {
-                return;
-            }
-
-            if (!ValidateApiFiles(shippedData, unshippedData, out List<Diagnostic> errors))
+            // TODO: Switch to "RegisterAdditionalFileAction" available in Microsoft.CodeAnalysis "3.8.x" to report additional file diagnostics.
```
sharwell, line 211: ❔ File an issue for this?

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -379,6 +379,18 @@ static string GetPackageSpecificContents(string packageName)
...
+                {
+                    return @"
+
+  <!-- Workaround for https://github.com/dotnet/roslyn/issues/4655 -->
+  <ItemGroup Condition=""Exists('$(MSBuildProjectDirectory)\PublicAPI.Shipped.txt')"" >
+	<AdditionalFiles Include=""PublicAPI.Shipped.txt"" />
+  </ItemGroup>
+  <ItemGroup Condition=""Exists('$(MSBuildProjectDirectory)\PublicAPI.Unshipped.txt')"" >
+	<AdditionalFiles Include=""PublicAPI.Unshipped.txt"" />
+  </ItemGroup>";
```
sharwell, line 392: This is going to cause a problem for projects that already include these files.

mavasani: I will change the condition to also check if it is not already an additional file.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
mavasani: https://github.com/dotnet/roslyn-analyzers/issues/3918

dougbu:
This looks great @mavasani❕ What will the gesture be in VS or VS Code to add missing `PublicApi.*.txt` files❔ Might be fine if that led to warnings (which our build will turn into errors) when we move to the new version of the package. But, I'd like to plan ahead.

dotnet/aspnetcode currently references Microsoft.CodeAnalysis.PublicApiAnalyzers in all of our implementation projects but doesn't add `@(AdditionalFiles)` items unless `PublicApi.*.txt` files are present already. See https://github.com/dotnet/aspnetcore/blob/master/eng/targets/CSharp.Common.targets#L17-L31

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
mavasani: I verified that the additional files get de-duped somewhere in the pipeline. Analyzer only sees unique additional files. For example, I added the following to a project and verified under the debugger that unique additional files are present in analyzer options:
```xml
	<ItemGroup>
		<AdditionalFiles Include="PublicAPI.Shipped.txt" />
		<AdditionalFiles Include="PublicAPI.Shipped.txt" />
	</ItemGroup>
	<ItemGroup>
		<AdditionalFiles Include="PublicAPI.Unshipped.txt" />
		<AdditionalFiles Include="PublicAPI.Unshipped.txt" />
	</ItemGroup>
```

mavasani:
@dougbu Missing API files will be treated as equivalent to empty API files. You will see an RS0016 for every public API in the project as soon as a package reference is added to the `Microsoft.CodeAnalysis.PublicApiAnalyzer`, and the first time you invoke a code fix for adding any public API, both shipped and unshipped files will be created in your project directory and also marked as additional files.

mavasani:
@dougbu After you consume the package with this change, you can delete these lines: https://github.com/dotnet/aspnetcore/blob/0437117cfb5bc91a34878fb962076bb05f4b8273/eng/targets/CSharp.Common.targets#L23-L30

dougbu:
Thanks again @mavasani. Please let us know when 3.3.0-beta3.final or whatever is out w/ this improvement.

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
dougbu: Adding items to an item group excludes duplicates by default though I'm not positive that is true both inside and outside a target. @rainersigwald❔

See also https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-items?view=vs-2019#BKMK_KeepDuplicates

dougbu: Could this cause problems in projects with code like we have at https://github.com/dotnet/aspnetcore/blob/0437117cfb5bc91a34878fb962076bb05f4b8273/eng/targets/CSharp.Common.targets#L27-L30 i.e. when exactly is this program run and what's the default `*.targets` file it'll create❔

mavasani: This is added to the targets file that is included with the NuGet package:
image

image

rainersigwald: Adding items creates duplicates by default both inside and outside targets; only inside targets with that attribute does it filter.

dougbu: Oops. Thanks @rainersigwald❕

mavasani:
@sharwell Any other feedback?

mavasani:
Ping for reviews.. This helps unblock aspnetcore repo from adopting the public API analyzer package.

mavasani:
Merging this to unblock, I will address any further feedback with a follow-up PR.

mavasani:
@dougbu You can now use https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet5&package=Microsoft.CodeAnalysis.PublicApiAnalyzers&protocolType=NuGet&version=3.3.0-beta3.20381.3&view=overview and get the code fix to create public API files across your solution.

dougbu:
This PR leaves the user in an odd state, giving both RS0016 and RS0037 errors when the `PublicApi.*.txt` files do not exist. Suggest the RS0016 fixer should include the required `#nullable enable` line to avoid fixing the missing files twice.

PR #3921, opened 2020-07-29
Add an analyzer for PublishSingleFile

agocke:
When publishing single-file applications, certain APIs are very likely
to return incorrect results. Most notably, asking for the Location of
any Assembly loaded from the single-file will return an empty string,
instead of a file path. To minimize suprise, this analyzer flags all
usages of the APIs most likely to break when used in single-file when
the PublishSingleFile property is set to true, and the
IncludeAllContentForSelfExtractProperty is not true (which provides the
legacy API behavior).


agocke:
@vitek-karas @sbomer @mateoatr @tlakollo @samsp-msft Can you all take a look at this and tell me what you think? There are certainly some instances with potential false positives (as I included in my test case), but it seems like this mistake is so easy to make and this scenario is so new that it would be better to produce extra warnings up front.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1422,4 +1422,13 @@
   <data name="DoNotUseOutAttributeStringPInvokeParametersTitle" xml:space="preserve">
     <value>Do not use 'OutAttribute' on string parameters for P/Invokes</value>
   </data>
+  <data name="AvoidAssemblyLocationInSingleFileDescription" xml:space="preserve">
+    <value>Accessing Assembly location or file path is not valid when publishing as a single-file.</value>
```
vitek-karas, line 1426: I don't like the word "valid" - it makes it sound like something completely wrong. Maybe just explain what it does:
"Property 'Assembly.Location' returns empty string when publishing the app as a single-file."

But I'm not a native speaker, so maybe your version actually sounds better...

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,149 @@
...
+                HashSet<IPropertySymbol> properties = new HashSet<IPropertySymbol>(SymbolEqualityComparer.Default);
+                HashSet<IMethodSymbol> methods = new HashSet<IMethodSymbol>(SymbolEqualityComparer.Default);
+
+                var assemblyType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemReflectionAssembly);
+                if (assemblyType is not null)
+                {
+                    // properties
+                    addIfNotNull(properties, tryGetSingleSymbol<IPropertySymbol>(assemblyType.GetMembers("Location")));
+                    addIfNotNull(properties, tryGetSingleSymbol<IPropertySymbol>(assemblyType.GetMembers("CodeBase")));
+                    addIfNotNull(properties, tryGetSingleSymbol<IPropertySymbol>(assemblyType.GetMembers("EscapedCodeBase")));
```
vitek-karas, line 73: These are now obsolete (https://github.com/dotnet/runtime/pull/39261) - not sure what's the UX - will I get two warnings for the same thing?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
vitek-karas: Actually maybe we should include the fact that the assembly has to be part of the app:
"Property `Assembly.Location` returns empty string for any assembly from application when published as single-file"

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
agocke: Probably. I didn't know we were going to obsolete them, so I can remove this.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
samsp-msft: Saying it returns empty string is better as it's then a condition the app can check against.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1422,4 +1422,13 @@
   <data name="DoNotUseOutAttributeStringPInvokeParametersTitle" xml:space="preserve">
     <value>Do not use 'OutAttribute' on string parameters for P/Invokes</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="AvoidAssemblyLocationInSingleFileDescription" xml:space="preserve">
+    <value>Accessing Assembly location or file path is not valid when publishing as a single-file.</value>
+  </data>
+  <data name="AvoidAssemblyLocationInSingleFileTitle" xml:space="preserve">
+    <value>Avoid '{0}'`' when publishing as a single-file</value>
```
Evangelink, line 1429: I don't know if I did a mistake on my suggestion or what happened but there is a problem with the title.

```suggestion
    <value>Avoid '{0}' when publishing as a single-file</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+
```
Evangelink, line 3:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -0,0 +1,169 @@
...
+    {
+        var a = Assembly.GetExecutingAssembly();
+        _ = a.Location;
+        _ = a.CodeBase;
+        _ = a.EscapedCodeBase;
+    }
+}";
+            return VerifyDiagnosticsAsync(src,
+                // /0/Test0.cs(8,13): warning CA3000: Avoid `System.Reflection.Assembly.Location` when publishing as a single-file. Assemblies inside a single-file bundle do not have a file or file path. If the path to the app directory is needed, consider calling System.AppContext.BaseDirectory.
+                VerifyCS.Diagnostic().WithSpan(8, 13, 8, 23).WithArguments("System.Reflection.Assembly.Location"),
```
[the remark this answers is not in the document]
agocke: This style of diagnostics verification is what Roslyn uses for all the compiler tests. Why invent something different here? I don't see the value making the source code harder to read.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
agocke: In that case we'll probably need individual diagnostic rules for each API to describe its behavior.

agocke: Updated

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1422,4 +1422,13 @@
...
-</root>
\ No newline at end of file
+  <data name="AvoidAssemblyLocationInSingleFileTitle" xml:space="preserve">
+    <value>Avoid using accessing Assembly file path when publishing as a single-file</value>
+  </data>
+  <data name="AvoidAssemblyLocationInSingleFileMessage" xml:space="preserve">
+    <value>'{0}' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.</value>
+  </data>
+  <data name="AvoidAssemblyGetFilesInSingleFile" xml:space="preserve">
+    <value>Assemblies embedded in a single-file app cannot have additional files in the manifest.</value>
```
vitek-karas, line 1432: Maybe it would be better to similarly to Location describe the actual behavior. In this case the API is going to throw an exception (it already does for in-memory assemblies outside of single-file).

agocke: The issue at https://github.com/dotnet/runtime/issues/38405 seems to imply the expected behavior is to return null or empty array.

sbomer:
Is the plan to share warning codes with a future single-file analysis done by the linker? If so we should make sure to reserve `IL3000/IL3001` in mono/linker to avoid conflicts - or if not we might want to use a different prefix here.

agocke:
@sbomer Yeah this was basically a set up for that exact idea. That we could move warnings back and forth between the linker and this analyzer framework by re-using error codes. I agree, it seems like a good idea to reserve those codes in the linker. Do you know how I should do that?

sbomer:
I filed https://github.com/mono/linker/issues/1393 to document this - I think it's enough for now. If we wanted to we could add a check to ensure we don't use those warning numbers.

tlakollo:
So I'm assuming this means 1XXX for linker **and single file** errors(?), 2XXX for linker warnings 3XXX for single file warnings

agocke:
@tlakollo  Yup, that was the idea.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -39,6 +39,9 @@ CA2354 | Security | Disabled | DataSetDataTableInIFormatterSerializableObjectGra
 CA2355 | Security | Disabled | DataSetDataTableInSerializableObjectGraphAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2355)
 CA2356 | Security | Disabled | DataSetDataTableInWebSerializableObjectGraphAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2356)
 
+IL3000 | Publish  | Warning  | DoNotUseAssemblyLocationInSingleFile, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/il3000)
+IL3001 | Publish  | Warning  | DoNotUseAssemblyGetFilesInSingleFile, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/il3001)
```
mavasani, line 43: Consider adding a new range row to https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt. This is an additional file which is used by a meta-analyzer on the repo that enforces few things:

1. No two analyzers share the same diagnostic ID in an analyzer package
2. There are no gaps introduced in the diagnostic ID range
3. Diagnostic ID selected for a specific category falls in the allowed range for that category

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
vitek-karas: I do agree that having one for each is lot of work - on the other hand it might give us opportunity to include some actionable feedback in the messages (if there's any). For example for `Assembly.Location` we may point users to `AppDomain.CurrentDomain.BaseDirectory` as a reasonable replacement for certain cases.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
vitek-karas: We'll probably have to update that issue. More recent discussions:
* `Assembly.CodeBase`: https://github.com/dotnet/runtime/issues/40087 - the agreement is to actually throw `PlatformNotSupported` - it's "safer" and we're obsoleting these anyway, so making them work feels contraproductive.
* Module.FullyQualifiedName https://github.com/dotnet/runtime/issues/40103 - should return `<Unknown>`. This will make it work the same as memory-loaded assemblies. This also means that `GetFile`/`GetFiles` will throw (as it does in non-single-file for memory-loaded assemblies).

agocke: What about `AssemblyName.CodeBase`? AFAIK that's not even obsolete

vitek-karas: I personally would expect `Assembly.CodeBase` and `AssemblyName.CodeBase` to behave the same. There has been discussion on it, but it sort of disappeared: https://github.com/dotnet/runtime/issues/31127
I'll start it again.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,158 @@
...
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Publish
+{
+    /// <summary>
+    /// CA3000: Do not use Assembly.Location in single-file publish
```
mavasani, line 15: Fix ID in doc comment?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,158 @@
...
+                    {
+                        return;
+                    }
+
+                    operationContext.ReportDiagnostic(invocation.CreateDiagnostic(GetFilesRule, targetMethod));
+                }, OperationKind.Invocation);
+
+                return;
+
+                static bool contains<T, TComp>(List<T> list, T elem, TComp comparer)
```
mavasani, line 115: We follow the Roslyn IDE conventions in this repo for local function names
```suggestion
                static bool Contains<T, TComp>(List<T> list, T elem, TComp comparer)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,158 @@
...
+                    return;
+                }
+                var includesAllContent = context.Options.GetMSBuildPropertyValue(
+                    MSBuildPropertyOptionNames.IncludeAllContentForSelfExtract, compilation, context.CancellationToken);
+                if (string.Equals(includesAllContent?.Trim(), "true", StringComparison.OrdinalIgnoreCase))
+                {
+                    return;
+                }
+
+                var properties = new List<IPropertySymbol>();
```
mavasani, line 70: Will this data structure be thread-safe given we will get concurrent callbacks?

mavasani: Might be best to use an ImmutableHashSet/ImmutableArray for these.

mavasani: If you decide to use ImmutableHashSet, then you can use this extension method: https://github.com/dotnet/roslyn-analyzers/blob/194d1443c4f5901afd4142587f88c35b06b37654/src/Utilities/Compiler/Extensions/ImmutableHashSetExtensions.cs#L95

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
agocke: yuck :)

agocke: I'll do it, but I'd like to register my objection as the designer and author of local functions 😄

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1422,4 +1422,13 @@
...
-</root>
\ No newline at end of file
+  <data name="AvoidAssemblyLocationInSingleFileTitle" xml:space="preserve">
+    <value>Avoid using accessing Assembly file path when publishing as a single-file</value>
+  </data>
+  <data name="AvoidAssemblyLocationInSingleFileMessage" xml:space="preserve">
+    <value>'{0}' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.</value>
+  </data>
+  <data name="AvoidAssemblyGetFilesInSingleFile" xml:space="preserve">
+    <value>'{0}' will throw for assemblies embedded in a single-file app.</value>
```
mavasani, line 1465: This resource string is used as both the title and message for `IL3001`. Title string should not have any string format arguments, so you probably want to use separate resource strings.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
agocke: This should be safe given that threading model. ImmutableHashSet is probably a bad idea for lists of < 8 elements. I can move to ImmutableArray instead.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,158 @@
...
+                    {
+                        if (comparer.Equals(e, elem))
+                        {
+                            return true;
+                        }
+                    }
+                    return false;
+                }
+
+                static TSymbol? tryGetSingleSymbol<TSymbol>(ImmutableArray<ISymbol> members) where TSymbol : class, ISymbol
```
mavasani, line 128:
```suggestion
                static TSymbol? TryGetSingleSymbol<TSymbol>(ImmutableArray<ISymbol> members) where TSymbol : class, ISymbol
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFile.cs
```diff
@@ -0,0 +1,158 @@
...
+                            else
+                            {
+                                return null;
+                            }
+                        }
+                    }
+                    return candidate;
+                }
+
+                static void addIfNotNull<TSymbol>(List<TSymbol> properties, TSymbol? p) where TSymbol : class, ISymbol
```
mavasani, line 148:
```suggestion
                static void AddIfNotNull<TSymbol>(List<TSymbol> properties, TSymbol? p) where TSymbol : class, ISymbol
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
agocke: Actually I meant to re-use AvoidAssemblyLocationInSingleFileTitle, since the title actually makes sense for both. good catch.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -43,6 +43,9 @@ CA2361 | Security | Disabled | DoNotUseDataSetReadXml, [Documentation](https://d
 CA2362 | Security | Disabled | DataSetDataTableInSerializableTypeAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2362)
 CA2363 | Security | Disabled | DataSetDataTableInSerializableTypeAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2363)
 
+IL3000 | Publish  | Warning  | DoNotUseAssemblyLocationInSingleFile, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/il3000)
```
mavasani, line 46: Do we want to create issue(s) on https://github.com/MicrosoftDocs/visualstudio-docs/issues to add documentation for the rules at https://docs.microsoft.com/visualstudio/code-quality/il3000 and https://docs.microsoft.com/visualstudio/code-quality/il3001?

agocke:
@mavasani You can merge this if you want, I don't have commit privileges for this repo

mavasani:
> You can merge this if you want, I don't have commit privileges for this repo

Interesting, I though all roslyn team members have the permissions. Let me explicitly add you.

mavasani:
@agocke Can you try again?

agocke:
@mavasani I'm not a Roslyn team member any more 😄

carlossanlop:
@agocke FYI, the change made to the file `src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md` added a space before the two new table rows, and markdown is not including them in the table:

https://github.com/dotnet/roslyn-analyzers/blob/master/src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md

agocke:
Woops, fixed here https://github.com/dotnet/roslyn-analyzers/pull/3944

PR #3922, opened 2020-07-29
Enable analysis of DiagnosticDescriptor arguments

Evangelink:
Follow-up of #3914 and relates to #3525


src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -142,52 +142,52 @@
...
-    <value>Specify at least one SyntaxKind of interest when registering a syntax node analyzer action.</value>
+    <value>Specify at least one SyntaxKind of interest when registering a syntax node analyzer action</value>
   </data>
   <data name="MissingOperationKindArgumentToRegisterActionMessage" xml:space="preserve">
-    <value>Specify at least one OperationKind of interest when registering an operation analyzer action.</value>
+    <value>Specify at least one OperationKind of interest when registering an operation analyzer action</value>
   </data>
   <data name="MissingKindArgumentToRegisterActionTitle" xml:space="preserve">
-    <value>Missing kind argument when registering an analyzer action.</value>
+    <value>Missing kind argument when registering an analyzer actionMissing kind argument when registering an analyzer actionMissing kind argument when registering an analyzer action</value>
```
mavasani, line 154: Did the code fix mess this up? Can you file an issue if you have a repro?

src/Directory.Build.targets
```diff
@@ -9,6 +9,9 @@
   -->
   <ItemGroup>
     <EmbeddedResource Update="**\*.resx" GenerateSource="true" />
+
+    <!-- Also pass the resx files as additional files to enable entry validation -->
+    <AdditionalFiles Include="**\*.resx" />
```
Evangelink, line 14: The order seems to matter for the rendering inside VS but not for the behavior. Let me know if you prefer it to appear as `EmbeddedResource` in VS.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -351,7 +350,7 @@
     <value>Compare symbols correctly</value>
   </data>
   <data name="ConfigureGeneratedCodeAnalysisDescription" xml:space="preserve">
```
mavasani, line 352: We should delete this resource string and update the analyzer to not pass in a description. It has no useful information over the title/message.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink, line 154: I missed this one and will fix it. There is a problem with the fixer when doing the fix-all solution which seems to be fixing the same entry multiple time by doing a concat rather than replace.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -197,32 +197,31 @@
 1. Define a new scope for the registered start action, possibly with a private nested type for analyzing each code unit.
 2. If required, define and initialize state in the start action.
 3. Register at least one non-end action that refers to this state in the start action. If no such action is necessary, consider replacing the start action with a non-start action. For example, a CodeBlockStartAction with no registered actions or only a registered CodeBlockEndAction should be replaced with a CodeBlockAction.
-4. If required, register an end action to report diagnostics based on the final state.
-</value>
+4. If required, register an end action to report diagnostics based on the final state.</value>
```
Evangelink, line 200: Do we want to have a special allowance for this scenario (i.e. blank line)?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -360,7 +359,7 @@
     <value>Configure generated code analysis</value>
   </data>
   <data name="EnableConcurrentExecutionDescription" xml:space="preserve">
-    <value>Enable concurrent execution</value>
+    <value>Enable concurrent execution.</value>
```
mavasani, line 362: Same here - please delete this resource string and update analyzer.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -510,13 +509,13 @@
     <value>Define diagnostic title correctly</value>
   </data>
   <data name="DefineDiagnosticMessageCorrectlyMessage" xml:space="preserve">
-    <value>The diagnostic message should not contain any line return character and should either be a single sentence without a trailing period or a multi-sentences with a trailing period.</value>
+    <value>The diagnostic message should not contain any line return character and should either be a single sentence without a trailing period or a multi-sentences with a trailing period</value>
```
mavasani, line 512: Oh the irony! :-)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectPrefix.cs
```diff
@@ -16,7 +16,7 @@ public sealed class IdentifiersShouldHaveCorrectPrefixAnalyzer : DiagnosticAnaly
         public const string RuleId = "CA1715";
 
         private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectPrefixTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
-        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectPrefixTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.IdentifiersShouldHaveCorrectPrefixDescription), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
```
Evangelink, line 19: I think this could be another analyzer we write and that checks a match between variable suffix and resource entry suffix.

mavasani, line 19: Nice!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -184,13 +184,13 @@
     <value>Use generic event handler instances</value>
   </data>
   <data name="UseGenericEventHandlerInstancesForDelegateMessage" xml:space="preserve">
-    <value>Remove '{0}' and replace its usage with a generic EventHandler, for e.g. EventHandler&lt;T&gt;, where T is a valid EventArgs</value>
+    <value>Remove '{0}' and replace its usage with a generic EventHandler, for e.g. EventHandler&lt;T&gt;, where T is a valid EventArgs.</value>
```
Evangelink, line 187: This is a false-positive, our naive algo is considering `. ` as new sentence separator which isn't the case here. Let me know if you want me to revert this change. Also let me know if we shall create a ticket?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -184,13 +184,13 @@
...
   </data>
   <data name="UseGenericEventHandlerInstancesForDelegateDescription" xml:space="preserve">
     <value>A type contains a delegate that returns void, whose signature contains two parameters (the first an object and the second a type that is assignable to EventArgs), and the containing assembly targets Microsoft .NET Framework?2.0.</value>
   </data>
   <data name="UseGenericEventHandlerInstancesForEventMessage" xml:space="preserve">
-    <value>Change the event '{0}' to replace the type '{1}' with a generic EventHandler, for e.g. EventHandler&lt;T&gt;, where T is a valid EventArgs</value>
+    <value>Change the event '{0}' to replace the type '{1}' with a generic EventHandler, for e.g. EventHandler&lt;T&gt;, where T is a valid EventArgs.</value>
```
Evangelink, line 193: Same here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -361,25 +361,25 @@
...
   <data name="AvoidZeroLengthArrayAllocationsTitle" xml:space="preserve">
-    <value>Avoid zero-length array allocations.</value>
+    <value>Avoid zero-length array allocations</value>
   </data>
   <data name="AvoidZeroLengthArrayAllocationsMessage" xml:space="preserve">
     <value>Avoid unnecessary zero-length array allocations.  Use {0} instead.</value>
   </data>
   <data name="DoNotUseEnumerableMethodsOnIndexableCollectionsInsteadUseTheCollectionDirectlyTitle" xml:space="preserve">
-    <value>Do not use Enumerable methods on indexable collections. Instead use the collection directly</value>
+    <value>Do not use Enumerable methods on indexable collections</value>
```
Evangelink, line 376: I have checked and this is the expected behavior of your code-fix.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -997,7 +997,7 @@
     <value>The Range-based indexer on string values produces a copy of requested portion of the string. This copy is usually unnecessary when it is implicitly used as a ReadOnlySpan or ReadOnlyMemory value. Use the AsSpan method to avoid the unnecessary copy.</value>
   </data>
   <data name="UseAsSpanInsteadOfRangeIndexerMessage" xml:space="preserve">
-    <value>Use '{0}' instead of the '{1}'-based indexer on '{2}' to avoid creating unnecessary data copies.</value>
+    <value>Use '{0}' instead of the '{1}'-based indexer on '{2}' to avoid creating unnecessary data copiesUse '{0}' instead of the '{1}'-based indexer on '{2}' to avoid creating unnecessary data copiesUse '{0}' instead of the '{1}'-based indexer on '{2}' to avoid creating unnecessary data copies</value>
```
Evangelink, line 1000: I will fix this duplication.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -208,13 +208,13 @@
     <value>The default value of an uninitialized enumeration, just as other value types, is zero. A nonflags-attributed enumeration should define a member by using the value of zero so that the default value is a valid value of the enumeration. If an enumeration that has the FlagsAttribute attribute applied defines a zero-valued member, its name should be ""None"" to indicate that no values have been set in the enumeration.</value>
   </data>
   <data name="EnumsShouldHaveZeroValueMessageFlagsRename" xml:space="preserve">
```
mavasani, line 210: Rename resource name to end with `Message`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -208,13 +208,13 @@
...
-    <value>In enum {0}, change the name of {1} to 'None'.</value>
+    <value>In enum {0}, change the name of {1} to 'None'</value>
   </data>
   <data name="EnumsShouldHaveZeroValueMessageFlagsMultipleZeros" xml:space="preserve">
-    <value>Remove all members that have the value zero from {0} except for one member that is named 'None'.</value>
+    <value>Remove all members that have the value zero from {0} except for one member that is named 'None'</value>
   </data>
```
mavasani, line 215: Rename resource name to end with `Message`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -208,13 +208,13 @@
...
   <data name="EnumsShouldHaveZeroValueMessageNotFlagsNoZeroValue" xml:space="preserve">
```
mavasani, line 216: Rename resource name to end with `Message`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -274,13 +274,13 @@
     <value>Attributes can define mandatory arguments that must be specified when you apply the attribute to a target. These are also known as positional arguments because they are supplied to attribute constructors as positional parameters. For every mandatory argument, the attribute should also provide a corresponding read-only property so that the value of the argument can be retrieved at execution time. Attributes can also define optional arguments, which are also known as named arguments. These arguments are supplied to attribute constructors by name and should have a corresponding read/write property.</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageDefault" xml:space="preserve">
```
mavasani, line 276: Seems like lot of message resource strings in this file have "Message" within the resource name instead of at the end. Not sure if it is possible to rename these to move Message to the end without making the name meaningless.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: I prefer to keep both resources so it's clear what can be updated

mavasani: How is it valuable over the diagnostic message?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: Ahaha we did the comments at the same time :) I don't really have an exact repro but it does happen when doing a fix-all.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: You guys are too fast, I was adding my notes xD

We have a LOT of resources where the title, message and description are the same. I am fine with both approach but going on Sam's direction would make our analyzers simpler. Otherwise we will have to have special cases that might lead to false negative.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
sharwell: I'd change `e.g.` to `example` here

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectPrefix.cs
Evangelink: See my other comment :)

src/Directory.Build.targets
mavasani: It seems fine. I was thinking more of the approach here: https://github.com/dotnet/roslyn/blob/master/docs/analyzers/Using%20Additional%20Files.md#passing-a-group-of-files, but either way is okay. I am also wondering if this entry (regardless of the one we choose) should be directly appended to the below string which gets added to `Microsoft.CodeAnalysis.Analyzers.targets`, so all analyzer authors see the new rules light up automatically. https://github.com/dotnet/roslyn-analyzers/blob/be84346550158fb6933154ffaae6ae2a2e93c198/src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs#L372-L380

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
mavasani: I am not sure if that is beneficial, current change looks fine.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: This was one of the cases I was explaining in the ticket. Currently when we have a rule where the various descriptors share the same title and description but differ by the message we use the following pattern: <RuleName><Message><DescriptorSpecificSuffix>. It would be easier for the analyzer if we were able to use a different pattern, for example <RuleName><Message_><DescriptorSpecificSuffix>.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: > This was one of the cases I was explaining in the ticket. Currently when we have a rule where the various descriptors share the same title and description but differ by the message we use the following pattern: . It would be easier for the analyzer if we were able to use a different pattern, for example <Message_>.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectPrefix.cs
mavasani: Possible, but that analyzer should possibly be disabled by default, as most analyzer authors do not mandate specific naming suffix. WDYT?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: Seems like a corner case, I'd let you make a call if you feel this is important enough to address.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Yes, this looks to be the correct change.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Can you also file an issue to fix the code fix?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: Hum actually nevermind I got your suggestion wrong. Yes +1 to remove the description when it's the same at the message.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectPrefix.cs
Evangelink: I was more thinking about a Roslyn analyzer so something that apply for this repo and maybe roslyn. WDYT?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: I think the fix would be to change the below to use a `HashSet` instead of a `List` and make `FixInfo` implement `IEquatable<FixInfo`: https://github.com/dotnet/roslyn-analyzers/blob/be84346550158fb6933154ffaae6ae2a2e93c198/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/DefineDiagnosticDescriptorArgumentsCorrectlyFix.CustomFixAllProvider.cs#L97

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectPrefix.cs
mavasani: Roslyn repo has a different format for resource strings - it expects the resource name to match the string value with `_` replacing the whitespace character between words. It would apply only to this repo.

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
```diff
@@ -363,4 +363,10 @@
   <data name="NamedTypeFullNameNotNullSuppressorJustification" xml:space="preserve">
     <value>The 'Type.FullName' property is never null when using the syntax 'typeof(T).FullName'</value>
   </data>
+  <data name="BlankLinesDescription" xml:space="preserve">
+    <value>Avoid multiple blank lines.</value>
```
mavasani, line 367: Either delete this resource + change the descriptor to pass null description OR make this string more informative over the message.

src/Directory.Build.targets
mavasani: @Evangelink You can revert this change once https://github.com/dotnet/roslyn-analyzers/pull/3924 goes in and you move to a newer analyzer package.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
mavasani: https://github.com/dotnet/roslyn-analyzers/pull/3923 for the fix

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: Sounds good. I think you don't need to make this change as part of this PR. I am fine with even not changing the resource name until we actually have an analyzer that checks name and strictly enforces it.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: > I'd change e.g. to example here

Yeah, that sounds reasonable.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: https://github.com/dotnet/roslyn-analyzers/pull/3923

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
mavasani: Ignore this comment for this PR.

mavasani:
> Can you please file follow-up issues and fix the unwanted resource concats?

@Evangelink Can you fix the resource concats? Once https://github.com/dotnet/roslyn-analyzers/pull/3923 and https://github.com/dotnet/roslyn-analyzers/pull/3924 are merged, you can update to the new package and this PR should be good to merge. Lets try to get all these 3 PRs in to avoid repeated merge conflicts with open PRs that violate these rules.

src/Directory.Build.targets
mavasani: This can now be reverted.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectPrefix.cs
Evangelink: Let's do nothing for now then.

Evangelink:
I have the same error locally about `RS0016` and `RS0037` but this doesn't make sense as I am only changing resource values...

mavasani:
@Evangelink The issue is at this line: https://github.com/dotnet/roslyn-analyzers/blob/31fe7d4b7dbf85609426f49861ca58f0d2d04b3b/src/Directory.Build.props#L17

We are enabling the PublicApiAnalyzers for the entire repo, though we only intend to track public API for a single project: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.csproj

Prior to https://github.com/dotnet/roslyn-analyzers/pull/3916, public API analyzer would bail out if it did not see any existing shipped or unshipped public API files for the project. With that PR, the analyzer no longer bails out but assumes empty public API surface for missing files.

Correct fix here is to remove the package reference to PublicApiAnalyzers at `src/Directory.Build.props` and instead add it directly to https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.csproj

Evangelink:
Thanks Manish!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1020,7 +1020,7 @@
     <value>An enumeration's values should not start with the type name of the enumeration.</value>
   </data>
   <data name="DoNotPrefixEnumValuesWithTypeNameMessage" xml:space="preserve">
-    <value>Do not prefix enum values with the name of the enum type '{0}'. </value>
+    <value>Do not prefix enum values with the name of the enum type '{0}'. .</value>
```
mavasani, line 1023: This should be fixed. We should update the analyzer to flag leading and trailing spaces for all 3 kind of strings.

Evangelink: See #3958

Evangelink:
@mavasani Actually I now remember why we added the PublicAPI analyzer directly in the `Directory.Build.props`, it was a suggestion from @sharwell as there is a transitive dependency from `Roslyn.Diagnostics.Analyzers` and without the explicit reference we do have issues in the command line but not within VS. So I do need to keep the reference in the file but we do need to disable the rule for the other projects. Let me know your preference (editorconfig or ruleset) to enable the rule only on the project we want.

eng/Analyzers_ShippingRules.ruleset
```diff
@@ -26,8 +26,22 @@
...
+    <Rule Id="RS0017" Action="None" />
+    <Rule Id="RS0022" Action="None" />
+    <Rule Id="RS0024" Action="None" />
+    <Rule Id="RS0025" Action="None" />
+    <Rule Id="RS0026" Action="None" />
+    <Rule Id="RS0027" Action="None" />
+    <Rule Id="RS0036" Action="None" />
+    <Rule Id="RS0037" Action="None" />
     <Rule Id="RS0041" Action="None" />
+    <Rule Id="RS0048" Action="None" />
```
mavasani, line 45: This is not maintainable. Please revert these changes and instead wait for https://github.com/dotnet/roslyn-analyzers/pull/3960

src/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.ruleset
```diff
@@ -0,0 +1,20 @@
...
+    <Rule Id="RS0025" Action="Warning" />
+    <Rule Id="RS0026" Action="Warning" />
+    <Rule Id="RS0027" Action="Warning" />
+    <Rule Id="RS0036" Action="Warning" />
+    <Rule Id="RS0037" Action="Warning" />
+    <!-- TODO: Remove this exclusion and fix related issues, tracked by https://github.com/dotnet/roslyn-analyzers/issues/3915 -->
+    <Rule Id="RS0041" Action="None" />
+    <Rule Id="RS0048" Action="Warning" />
+  </Rules>
+</RuleSet>
```
mavasani, line 20: Revert - this is not maintainable.

mavasani:
@Evangelink Would it be possible to disable RS0046 for this PR and instead create a separate follow-up PR to enable it and apply https://github.com/dotnet/roslyn-analyzers/pull/3922/commits/af7af67fa0111c24c68adfd87f9f2c116d9b8d05?

src/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.ruleset
Evangelink: Done

eng/Analyzers_ShippingRules.ruleset
Evangelink: Done

Evangelink:
@mavasani That shall be ok. Once this is merged, I will create a follow-up PR with the fixes for `RS0046`.

src/Directory.Build.props
```diff
@@ -11,12 +11,16 @@
   </PropertyGroup>
 
   <!-- Test runner configuration -->
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)Common\Test\App.config" CopyToOutputDirectory="PreserveNewest" Condition="'$(IsUnitTestProject)' == 'true'" />
+  </ItemGroup>
+
+  <!-- Code analyzers -->
```
Evangelink, line 18: The various analyzers were put under a "Test runner configuration" section so I moved them in a separate group.

src/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.csproj
```diff
@@ -17,6 +17,7 @@
   <Import Project="..\Utilities\Compiler\Analyzer.Utilities.projitems" Label="Shared" />
   <Import Project="..\Utilities\FlowAnalysis\FlowAnalysis.Utilities.projitems" Label="Shared" />
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)" />
```
Evangelink, line 20: Un-needed, I will remove it.

src/Directory.Build.props
```diff
@@ -11,11 +11,16 @@
...
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="$(MicrosoftCodeAnalysisFXCopAnalyzersVersion)" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="$(MicrosoftCodeAnalysisAnalyzersVersion)" />    
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="$(MicrosoftCodeAnalysisAnalyzersVersion)" />
+    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)" />
```
mavasani, line 22: Do we even need this entry? We should get them via package reference to ‘Roslyn.Diagnostics.Analyzers’ correct?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1296,7 +1296,7 @@
     <value>When an exception is raised in a finally clause, the new exception hides the active exception. This makes the original error difficult to detect and debug.</value>
   </data>
   <data name="DoNotRaiseExceptionsInExceptionClausesMessageFinally" xml:space="preserve">
-    <value>Do not raise an exception from within a finally clause. </value>
+    <value>Do not raise an exception from within a finally clause. .</value>
```
mavasani, line 1299: Can we fix this string?

src/Directory.Build.props
Evangelink: Yes! I don't know if the behavior is intended or if there is a bug but with only the transitive dependency the diagnostics get raised during command line build but not within VS. Adding the analyzer explicitly allows for the mechanism to work properly in VS.

mavasani:
@Evangelink I can merge once you fix the last resource string.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: Crap I reverted this change too... This PR is endless xD

mavasani:
Once this PR is merged, we should retrigger CI on any open PR with green state and if it changes or adds a resource string in violation.

mavasani:
Yay!!

PR #3924, opened 2020-07-30
Add all EmbeddedResource files with '.resx' extension as analyzer add…

mavasani:
…itional files for Microsoft.CodeAnalysis.Analyzers package

This enables resource string validation meta-analyzers by default for all analyzer authors


mavasani:
Thank you @Evangelink!

PR #3926, opened 2020-07-30
Don't track parameter as a source even if it was tainted if it doesn't flow into a sink argument

JarLob:
The added test was reporting an injection twice before the fix, because both arguments are tainted. However the only one of the arguments flows into a sink parameter. The example with `Process.Start(string fileName, string arguments, string userName, SecureString password, string domain)` is a little artificial since tainted data flowing into `userName` may be bad, however the function defines only `fileName` and `arguments` as sinks, so it demonstrates the issue.


mavasani:
@dotpaul for review

dotpaul:
Thank you @JarLob ! Would you mind targeting the 2.9.x branch rather than the master branch for this?

dotpaul:
Yeah, might be easiest to create a new branch off of 2.9.x, cherry-pick your two commits, and open a new PR.

Issue #3827, opened 2020-07-07
CA1835 code fixer should add a `using System` directive if it is not written


Tan90909090:
#### Analyzer package
[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version
v3.3.0-beta1.final (Latest pre-release)

#### Diagnostic ID
[CA1835](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1835)

#### Repro steps
1. Write a following code:
```cs
using System.IO;
using System.Threading.Tasks;

public class C
{
    async ValueTask FooAsync(Stream stream, byte[] buffer)
    {
        await stream.WriteAsync(buffer, 0, buffer.Length) // CA1835 occurs here
            .ConfigureAwait(false);
    }

    static void Main() { }
}
```

2. Apply `Prefer the 'Memory'-based overloads for 'ReadAsync' and 'WriteAsync'`

#### Expected behavior
The code-fixer adds a `using System;` line because a `AsMemory` extension method belongs to `System.MemoryExtensions`.

#### Actual behavior
The code-fixer does not add a `using System;` line:
```cs
using System.IO;
using System.Threading.Tasks;

public class C
{
    async ValueTask FooAsync(Stream stream, byte[] buffer)
    {
        await stream.WriteAsync(buffer.AsMemory(0, buffer.Length)) // CA1835 occurs here
            .ConfigureAwait(false);
    }

    static void Main() { }
}
```
This code has a `CS1061: 'byte[]' does not contain a definition for 'AsMemory' and no accessible extension method 'AsMemory' accepting a first argument of type 'byte[]' could be found (are you missing a using directive or an assembly reference?)`.

mavasani:
@carlossanlop

carlossanlop:
Thanks @Tan90909090 
@mavasani I'll take a look as soon as I get some bandwidth.

PR #3931, opened 2020-07-31
CA1835 fix - Automatically add using/imports System if not yet included

carlossanlop:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/3827

If rule CA1835 finds a diagnostic and applies a fix, but the `System` namespace has not yet been added (which contains the `AsMemory()` method for byte arrays), then the compilation can fail.
With this change, the `System` namespace should be added if the user has not added it yet.

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -40,9 +40,9 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
             }
         }
 
-        protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> imports)
+        protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> importList)
         {
-            foreach (SyntaxNode import in imports)
+            foreach (SyntaxNode import in importList)
             {
                 if (import is UsingDirectiveSyntax usingDirective &&
                     usingDirective.Name is IdentifierNameSyntax identifierName &&
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -44,9 +44,7 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
         {
             foreach (SyntaxNode import in importList)
             {
-                if (import is UsingDirectiveSyntax usingDirective &&
-                    usingDirective.Name is IdentifierNameSyntax identifierName &&
-                    identifierName.Identifier.Text == "System")
+                if (import is UsingDirectiveSyntax { Name: IdentifierNameSyntax { Identifier: { Text: nameof(System) } } })
                 {
                     return true;
                 }
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
@@ -37,6 +38,20 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
                            argNode.NameColon?.Name?.Identifier.ValueText == name;
                 });
             }
+        }
+
+        protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> imports)
+        {
+            foreach (SyntaxNode import in imports)
+            {
+                if (import is UsingDirectiveSyntax usingDirective &&
+                    usingDirective.Name is IdentifierNameSyntax identifierName &&
+                    identifierName.Identifier.Text == "System")
+                {
+                    return true;
+                }
+            }
+            return false;
         }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading;
 using System.Threading.Tasks;
@@ -32,6 +33,9 @@ namespace Microsoft.NetCore.Analyzers.Runtime
     {
         // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
         protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
+
+        // 
+        protected abstract bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> imports);
 
         public sealed override ImmutableArray<string> FixableDiagnosticIds =>
             ImmutableArray.Create(PreferStreamAsyncMemoryOverloads.RuleId);
@@ -100,7 +104,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 context.Diagnostics);
         }
 
-        private static Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName,
+        private Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName,
             SyntaxNode bufferValueNode, bool isBufferNamed,
             SyntaxNode offsetValueNode, bool isOffsetNamed,
             SyntaxNode countValueNode, bool isCountNamed,
@@ -119,6 +123,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             // Generate an invocation of the AsMemory() method from the byte array object, using the correct named arguments
             SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferValueNode.WithoutTrivia(), memberName: "AsMemory");
+
             SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(
                 asMemoryExpressionNode,
                 namedStartNode.WithTriviaFrom(offsetValueNode),
@@ -145,7 +150,11 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             }
             SyntaxNode newInvocationExpression = generator.InvocationExpression(asyncMethodNode, nodeArguments).WithTriviaFrom(streamInstanceNode);
 
-            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
+            bool containsSystemImport = IsSystemNamespaceImported(generator.GetNamespaceImports(root));
+
+            SyntaxNode rootWithImports = containsSystemImport ? root : generator.AddNamespaceImports(root, generator.NamespaceImportDeclaration("System").WithAddImportsAnnotation());
+            SyntaxNode newRoot = generator.ReplaceNode(rootWithImports, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
+
             return Task.FromResult(doc.WithSyntaxRoot(newRoot));
         }
 
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -34,8 +34,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime
         // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
         protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
 
-        // 
-        protected abstract bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> imports);
+        // Verifies if a namespace has already been added to the usings/imports list.
+        protected abstract bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> importList);
 
         public sealed override ImmutableArray<string> FixableDiagnosticIds =>
             ImmutableArray.Create(PreferStreamAsyncMemoryOverloads.RuleId);
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -123,7 +123,6 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             // Generate an invocation of the AsMemory() method from the byte array object, using the correct named arguments
             SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferValueNode.WithoutTrivia(), memberName: "AsMemory");
-
             SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(
                 asMemoryExpressionNode,
                 namedStartNode.WithTriviaFrom(offsetValueNode),
@@ -152,10 +151,11 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             bool containsSystemImport = IsSystemNamespaceImported(generator.GetNamespaceImports(root));
 
-            SyntaxNode rootWithImports = containsSystemImport ? root : generator.AddNamespaceImports(root, generator.NamespaceImportDeclaration("System").WithAddImportsAnnotation());
-            SyntaxNode newRoot = generator.ReplaceNode(rootWithImports, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
+            // The invocation needs to be replaced before adding the import/using, it won't work the other way around
+            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
+            SyntaxNode newRootWithImports = containsSystemImport ? newRoot : generator.AddNamespaceImports(newRoot, generator.NamespaceImportDeclaration("System"));
 
-            return Task.FromResult(doc.WithSyntaxRoot(newRoot));
+            return Task.FromResult(doc.WithSyntaxRoot(newRootWithImports));
         }
 
         // Needed for Telemetry (https://github.com/dotnet/roslyn-analyzers/issues/192) 
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -153,12 +153,12 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             // The invocation needs to be replaced before adding the import/using, it won't work the other way around
             SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
-            SyntaxNode newRootWithImports = containsSystemImport ? newRoot : generator.AddNamespaceImports(newRoot, generator.NamespaceImportDeclaration("System"));
+            SyntaxNode newRootWithImports = containsSystemImport ? newRoot : generator.AddNamespaceImports(newRoot, generator.NamespaceImportDeclaration(nameof(System)));
 
             return Task.FromResult(doc.WithSyntaxRoot(newRootWithImports));
         }
 
-        // Needed for Telemetry (https://github.com/dotnet/roslyn-analyzers/issues/192) 
+        // Needed for Telemetry (https://github.com/dotnet/roslyn-analyzers/issues/192)
         private class MyCodeAction : DocumentChangeAction
         {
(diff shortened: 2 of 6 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -578,6 +576,43 @@ public static IEnumerable<object[]> CSharpInlineByteArrayTestData()
                                         "(new byte[s.Length]).AsMemory(0, (int)s.Length), new CancellationToken()" };
         }
 
+        [Fact]
+        public Task CS_Fixer_Diagnostic_EnsureSystemNamespaceAutoAdded()
```
carlossanlop, line 580: @mavasani I am seeing the following error when I run any of my new unit tests:


<details>
  <summary>Error</summary>

  ```
  Message: 
      Context: Iterative code fix application
      The upper limit for the number of code fix iterations was exceeded
      Expected: True
      Actual:   False
    Stack Trace: 
      XUnitVerifier.True(Boolean assert, String message) line 62
      <FixEachAnalyzerDiagnosticAsync>d__43.MoveNext() line 538
      ExceptionDispatchInfo.Throw()
      <VerifyFixAsync>d__42.MoveNext() line 434
      TaskAwaiter.ThrowForNonSuccess(Task task)
      TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
      ConfiguredTaskAwaiter.GetResult()
      <VerifyFixAsync>d__40.MoveNext() line 371
      TaskAwaiter.ThrowForNonSuccess(Task task)
      TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
      ConfiguredTaskAwaiter.GetResult()
      <RunAsync>d__38.MoveNext() line 247
      TaskAwaiter.ThrowForNonSuccess(Task task)
      TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
      TaskAwaiter.ThrowForNonSuccess(Task task)
      TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
      TaskAwaiter.ThrowForNonSuccess(Task task)
      TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  ```
</details>

I am not sure what the root cause is. When I step into the fixer code with the debugger, and put a breakpoint right after the new root is generated, the code seems to be correct - the `System` namespace gets added where the unit tests expect it.

Any ideas what could be causing the failure?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -38,5 +39,19 @@ public sealed class CSharpPreferStreamAsyncMemoryOverloadsFixer : PreferStreamAs
                 });
             }
         }
+
+        protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> imports)
```
buyaa-n, line 43: Seems this is only overriden for C#, but i see VB tests added, it can work without overriding?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -33,6 +34,9 @@ public abstract class PreferStreamAsyncMemoryOverloadsFixer : CodeFixProvider
         // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
         protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
 
+        //
```
buyaa-n, line 37: Forgot to leave the comment?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
buyaa-n: Never mind it was down below

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Huh weird, I thought I added a comment on top of this method. Thanks for catching.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: @Evangelink @sharwell may also know what's causing this error.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani, line 43: @sharwell Does the code fixer actually need to check this? I though `AddImportAnnotation` will automatically no-op if it sees an existing using directive with same import.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
```diff
@@ -578,6 +576,43 @@ public static IEnumerable<object[]> CSharpInlineByteArrayTestData()
...
+        {
+            byte[] buffer = new byte[s.Length];
+            await s.ReadAsync(new byte[s.Length], 0, (int)s.Length);
+        }
+    }
+}";
+            string fixedCode = @"
+using System.IO;
+using System.Threading;
+using System;
```
mavasani, line 600: @carlossanlop I would have expected the code fix to have added the using in sorted order. Can you check the behavior when you apply the fix in VS with existing usings?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -145,7 +150,11 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
             }
             SyntaxNode newInvocationExpression = generator.InvocationExpression(asyncMethodNode, nodeArguments).WithTriviaFrom(streamInstanceNode);
 
-            SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
+            bool containsSystemImport = IsSystemNamespaceImported(generator.GetNamespaceImports(root));
+
+            SyntaxNode rootWithImports = containsSystemImport ? root : generator.AddNamespaceImports(root, generator.NamespaceImportDeclaration("System").WithAddImportsAnnotation());
```
carlossanlop, line 155: Is `WithAddImportAnnotation` necessary here? I was trying to see if that would solve the unit test failure, and decided to leave it so I could ask here in the review. Should I remove it?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -145,7 +150,11 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            SyntaxNode newRoot = generator.ReplaceNode(rootWithImports, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
```
carlossanlop, line 156: If I put a breakpoint here, the `newRoot` contains the `using System;` line added at the end of the list of imports, as expected, yet the unit test fail with the error I shared in my other comment.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: I was thinking the other way. `WithAddImportAnnotation` should be the supported way to add an imports instead of explicitly requiring to add a namespace import using the generator. I think @sharwell is the better person to answer this question.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: I had to add this check because otherwise, the import would be added at least 2 times. At least that's the behavior I saw in the unit tests, since the test source code gets analyzed more than once for each test.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
mavasani: This indicates that either the diagnostic was not resolved after the required iterations or a new diagnostic was produced. Are you able to debug into the testing library to figure out?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: @mavasani It does not get added at the beginning, I was also expecting that.
`generator.AddNamespaceImports` added it at the end. I verified this when debugging the new fixer code.

src/Utilities/Workspaces/SyntaxNodeExtensions.cs
```diff
@@ -24,6 +24,11 @@ internal static class SyntaxNodeExtensions
             }
         }
 
+        /// <summary>
+        /// Annotates a syntax node representing a type so that any missing imports get automatically added. Does not work in any other kinds of nodes.
+        /// </summary>
+        /// <param name="syntaxNode">The type node to annotate.</param>
+        /// <returns>The annotated type node.</returns>
         public static SyntaxNode WithAddImportsAnnotation(this SyntaxNode syntaxNode)
```
carlossanlop, line 32: @sharwell I documented this method we discussed yesterday, so people know it only works with Types. Let me know if you'd like to change the wording.

mavasani: I believe we should file a Roslyn bug here. Roslyn should throw an exception with an actionable message when an annotation is added to an unsupported node, instead of silently ignoring it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
sharwell: > Is WithAddImportAnnotation necessary here?

No, it doesn't make sense to use this annotation on using directives.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
sharwell: Try setting the `FixOne` option, and fixing any remaining issues with the test. You'll likely see that after the code fix is applied, problems are still appearing in the code.

src/Utilities/Workspaces/SyntaxNodeExtensions.cs
carlossanlop: Done: https://github.com/dotnet/roslyn/issues/46481

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -33,6 +34,9 @@ public abstract class PreferStreamAsyncMemoryOverloadsFixer : CodeFixProvider
         // Checks if the argument in the specified index has a name. If it doesn't, returns that arguments. If it does, then looks for the argument using the specified name, and returns it, or null if not found.
         protected abstract IArgumentOperation? GetArgumentByPositionOrName(ImmutableArray<IArgumentOperation> args, int index, string name, out bool isNamed);
 
+        // 
+        protected abstract bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> imports);
```
carlossanlop, line 38: Need to rename the parameter `imports` because VB does not like it (reserved keyword, even though it starts with uppercase I). Strangely, VS did not complain when I used it to build the solution.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: Thanks, @sharwell.

Note to self - Found a `FixOne` usage example here:
[roslyn/src/EditorFeatures/CSharpTest/Diagnostics/MakeMethodSynchronous/MakeMethodSynchronousTests.cs#L458](https://github.com/dotnet/roslyn/blob/8a7bfd860e0544078ddbe6a0192f5d9ca0edac41/src/EditorFeatures/CSharpTest/Diagnostics/MakeMethodSynchronous/MakeMethodSynchronousTests.cs#L458)
```cs
            await new VerifyCS.Test
            {
                TestCode = source,
                FixedState =
                {
                    Sources = { expected },
                    MarkupHandling = MarkupMode.Allow,
                },
                CodeFixTestBehaviors = CodeFixTestBehaviors.FixOne,
            }.RunAsync();
```

carlossanlop:
@mavasani @sharwell sorry for the delay, I was looking into other things.

I added `FixOne` and ran the test. It seems the fixer is indeed adding the missing import, but then the test fails because the `ReadAsync` line was not fixed. I suppose that's expected, since only one thing was changed (adding the import).

<details>
  <summary>Code I used</summary>

  ```cs
        [Fact]
        public Task CS_Fixer_Diagnostic_EnsureSystemNamespaceAutoAdded()
        {
            string originalCode = @"
  using System.IO;
  using System.Threading;

  class C
  {
    public async void M()
    {
        using (FileStream s = File.Open(""path.txt"", FileMode.Open))
        {
            byte[] buffer = new byte[s.Length];
            await s.ReadAsync(new byte[s.Length], 0, (int)s.Length);
        }
    }
  }";
            string fixedCode = @"
  using System.IO;
  using System.Threading;
  using System;

  class C
  {
    public async void M()
    {
        using (FileStream s = File.Open(""path.txt"", FileMode.Open))
        {
            byte[] buffer = new byte[s.Length];
            await s.ReadAsync((new byte[s.Length]).AsMemory(0, (int)s.Length));
        }
    }
  }";
            var test = new Test.Utilities.CSharpCodeFixVerifier<
    PreferStreamAsyncMemoryOverloads,
    CSharp.Analyzers.Runtime.CSharpPreferStreamAsyncMemoryOverloadsFixer>.Test
            {
                TestCode = originalCode,
                FixedCode = fixedCode,
                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp50,
                CodeFixTestBehaviors = CodeFixTestBehaviors.FixOne
            };
            test.ExpectedDiagnostics.Add(GetCSharpResult(12, 19, 12, 68));
            return test.RunAsync();

            // return CSharpVerifyExpectedCodeFixDiagnosticsAsync(originalCode, fixedCode, GetCSharpResult(12, 19, 12, 68));
        }
  ```
</details>

<details>
  <summary>Test error message</summary>

  ```
    Context: Iterative code fix application
    content of '/0/Test0.cs' did not match. Diff shown with expected as baseline:
     
     using System.IO;
     using System.Threading;
     using System;
     
     class C
     {
         public async void M()
         {
             using (FileStream s = File.Open("path.txt", FileMode.Open))
             {
                 byte[] buffer = new byte[s.Length];
    -            await s.ReadAsync((new byte[s.Length]).AsMemory(0, (int)s.Length));
    +            await s.ReadAsync(new byte[s.Length], 0, (int)s.Length);
             }
         }
     }
    
    Expected: True
    Actual:   False
  ```
</details>

So I thought maybe I could add the expected diagnostic for the addition of the import, and removing the `FixOne` restriction, but the test fails saying only one diagnostic was found (the one for `ReadAsync`), because adding a missing import does not count as a diagnostic.

<details>
  <summary>Updated code</summary>

  ```cs
            var test = new Test.Utilities.CSharpCodeFixVerifier<
    PreferStreamAsyncMemoryOverloads,
    CSharp.Analyzers.Runtime.CSharpPreferStreamAsyncMemoryOverloadsFixer>.Test
            {
                TestCode = originalCode,
                FixedCode = fixedCode,
                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp50,
                // CodeFixTestBehaviors = CodeFixTestBehaviors.FixOne
            };
            test.ExpectedDiagnostics.Add(GetCSharpResult(4, 1, 4, 13));
            test.ExpectedDiagnostics.Add(GetCSharpResult(12, 19, 12, 68));
            return test.RunAsync();
  ```
</details>

<details>
  <summary>New error</summary>

  ```
 Message: 
    Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessageException : Context: Diagnostics of test state
    Mismatch between number of diagnostics returned, expected "2" actual "1"
    
    Diagnostics:
    // /0/Test0.cs(12,19): info CA1835: Change the 'ReadAsync' method call to use the 'Stream.ReadAsync(Memory<byte>, CancellationToken)' overload
    VerifyCS.Diagnostic(PreferStreamAsyncMemoryOverloads.CA1835).WithSpan(12, 19, 12, 68).WithArguments("ReadAsync", "Stream.ReadAsync(Memory<byte>, CancellationToken)"),
    
    
    Assert.Equal() Failure
    Expected: 2
    Actual:   1
  ```
</details>

I'm not sure I understand what's causing this. Any ideas?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
carlossanlop: Found the root cause: I had to replace the invocation node _before_ adding the using/import.
If I add the using/import first, then it seems the reference to the original location of the invocation to update, gets lost (the code does not get properly updated), so attempting to replace the invocation fails silently.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferStreamReadAsyncMemoryOverloadsTests.cs
mavasani: Tagging @CyrusNajmabadi for insights here. How does code fixer ensure that a new using import that it wants to is sorted? Otherwise, I feel like you have a follow-up bug coming your way when someone complains the code fix broke the sort order of their imports.

CyrusNajmabadi: > Tagging @CyrusNajmabadi for insights here. How does code fixer ensure that a new using import that it wants to is sorted? 

The internal roslyn system for adding imports will add it in a manner that respects all the editor options here (sorting, spacing, etc.).

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -38,5 +39,19 @@ public sealed class CSharpPreferStreamAsyncMemoryOverloadsFixer : PreferStreamAs
...
             }
         }
+
+        protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> importList)
+        {
+            foreach (SyntaxNode import in importList)
+            {
+                if (import is UsingDirectiveSyntax usingDirective &&
+                    usingDirective.Name is IdentifierNameSyntax identifierName &&
+                    identifierName.Identifier.Text == "System")
```
CyrusNajmabadi, line 49:
```suggestion
                if (import is UsingDirectiveSyntax { Name: IdentifierNameSyntax { Identifier: { Text: "System" } } })
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -145,8 +149,13 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
             }
             SyntaxNode newInvocationExpression = generator.InvocationExpression(asyncMethodNode, nodeArguments).WithTriviaFrom(streamInstanceNode);
 
+            bool containsSystemImport = IsSystemNamespaceImported(generator.GetNamespaceImports(root));
+
+            // The invocation needs to be replaced before adding the import/using, it won't work the other way around
             SyntaxNode newRoot = generator.ReplaceNode(root, invocation.Syntax, newInvocationExpression.WithTriviaFrom(invocation.Syntax));
-            return Task.FromResult(doc.WithSyntaxRoot(newRoot));
+            SyntaxNode newRootWithImports = containsSystemImport ? newRoot : generator.AddNamespaceImports(newRoot, generator.NamespaceImportDeclaration("System"));
```
CyrusNajmabadi, line 156:
```suggestion
            SyntaxNode newRootWithImports = containsSystemImport ? newRoot : generator.AddNamespaceImports(newRoot, generator.NamespaceImportDeclaration(nameof(System)));
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Nice suggestion @CyrusNajmabadi, let's use the new stuff 😊.
I guess it shouls also be `nameof(System)` here, based on your other comment, right?

mavasani:
Should this PR target https://github.com/dotnet/roslyn-analyzers/tree/release/dotnet5-rc1?

carlossanlop:
@mavasani it would make sense to submit it it for RC1. Would it get to the SDK on time so that we can see it consumed in dotnet/runtime?

mavasani:
> so that we can see it consumed in dotnet/runtime

I don't believe dotnet/runtime uses the analyzers/fixers that ship with the SDK. The repo directly references `Microsoft.CodeAnalysis.NetAnalyzers` nuget package, so you should be able to just bump up the referenced version in the repo to get the fix.

carlossanlop:
Ah ok thanks.
So I should submit a backport to the RC1 branch, which I'll do after merging this change (I assume it's the procedure for this repo, otherwise please correct me).

mavasani:
> So I should submit a backport to the RC1 branch, which I'll do after merging this change (I assume it's the procedure for this repo, otherwise please correct me).

That is if you want the fix to be in RC1. Otherwise, I believe it will flow into RC2. @jmarolf to confirm

PR #3932, opened 2020-07-31
Fix for implicit conversion from String to RawSqlString

JarLob:
The proposed fix makes the taint analyzer treat implicit conversions as transferring taint.

The added sink and test demonstrate that before the fix an implicit conversion of `string` to `RawSqlString` loses the taint and there is no warning. The implicit conversion is defined in a library the user doesn't control, but is recognized as "UserDefined" (maybe this is the bug?). Ideally there should be a way to define the taint transfer rule. There are taint transfer methods already, but it looks like they are for taint transfer between arguments. Let me know if I missed the way to define the rule. It looked easier to modify `VisitConversion`.

To demonstrate the issue I've changed the .NET framework version referenced in the compiled tests, because the `FromSql` overload is available only in EntityFrameworkCore 2.0, but it is not compatible with .NET 4.6. However I didn't touch the .NET version used to build analyzers. Looks like all tests are passing.

There was already a test for implicit conversions `FlowAnalysis_ImplicitUserDefinedConversionsInInitializer_Diagnostic`, however I'm not sure I'm following its logic. Does it check that taint tracking loses the information about the constantness? I have flipped the logic adjusting to the new behavior.

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
@@ -179,6 +179,24 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
                 }
             }
 
+            public override TaintedDataAbstractValue VisitConversion(IConversionOperation operation, object? argument)
+            {
+                var operandValue = Visit(operation.Operand, argument);
+
+                if (!operation.Conversion.Exists)
+                {
+                    return ValueDomain.UnknownOrMayBeValue;
+                }
+
+                if (operation.Conversion.IsImplicit)
+                {
+                    return operandValue;
+                }
+
+                // Conservative for error code and user defined operator.
+                return !operation.Conversion.IsUserDefined ? operandValue : ValueDomain.UnknownOrMayBeValue;
+            }
+
             protected override TaintedDataAbstractValue ComputeAnalysisValueForReferenceOperation(IOperation operation, TaintedDataAbstractValue defaultValue)
             {
                 // If the property reference itself is a tainted data source
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
@@ -181,7 +181,7 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
 
             public override TaintedDataAbstractValue VisitConversion(IConversionOperation operation, object? argument)
             {
-                var operandValue = Visit(operation.Operand, argument);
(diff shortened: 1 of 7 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -2757,8 +2757,14 @@ public override TAbstractAnalysisValue VisitConversion(IConversionOperation oper
         {
             var operandValue = Visit(operation.Operand, argument);
 
+            if (!operation.Conversion.Exists)
+                return ValueDomain.UnknownOrMayBeValue;
+
+            if (operation.Conversion.IsImplicit)
```
dotpaul, line 2763: This new logic makes sense for TaintedDataAnalysis, but not so sure about everything else. @mavasani, what do you think?

It might be better to override `VisitConversion()` in `TaintedDataOperationVisitor`.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -2757,8 +2757,14 @@ public override TAbstractAnalysisValue VisitConversion(IConversionOperation oper
         {
             var operandValue = Visit(operation.Operand, argument);
 
+            if (!operation.Conversion.Exists)
+                return ValueDomain.UnknownOrMayBeValue;
```
dotpaul, line 2761:
```suggestion
            {
                return ValueDomain.UnknownOrMayBeValue;
            }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -2757,8 +2757,14 @@ public override TAbstractAnalysisValue VisitConversion(IConversionOperation oper
...
+
+            if (operation.Conversion.IsImplicit)
+                return operandValue;
```
dotpaul, line 2764:
```suggestion
            {
                return operandValue;
            }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
mavasani: Agreed. Not sure we can fallback to operand value for user defined implicit conversions for all analyses. Can we move the change to `TaintedDataOperationVisitor`?

JarLob: Good point

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
```diff
@@ -179,6 +179,24 @@ public override TaintedDataAbstractValue DefaultVisit(IOperation operation, obje
                 }
             }
 
+            public override TaintedDataAbstractValue VisitConversion(IConversionOperation operation, object? argument)
+            {
+                var operandValue = Visit(operation.Operand, argument);
```
dotpaul, line 184: Nit: prefer explict types in this file

```suggestion
                TaintedDataAbstractValue operandValue = Visit(operation.Operand, argument);
```

PR #3938, opened 2020-08-02
Allow tainting function arguments

JarLob:
This pull request allows tainting method arguments itself that is a common case in MVC pattern, like:
```cs
public class MyController : Controller
{
    public string DoSomething(string input) // << input is tainted argument
    {
        new SqlCommand(input);
        return null;
    }
}
```

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMapExtensions.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMapExtensions.cs
@@ -131,6 +131,24 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
         }
 
         /// <summary>
+        /// Determines if the given parameter is a tainted data source.
+        /// </summary>
+        /// <param name="sourceSymbolMap"></param>
+        /// <param name="parameterSymbol"></param>
+        /// <returns></returns>
+        public static bool IsSourceParameter(this TaintedDataSymbolMap<SourceInfo> sourceSymbolMap, IParameterSymbol parameterSymbol)
+        {
+            ISymbol containingSymbol = parameterSymbol.ContainingSymbol;
+            foreach (SourceInfo sourceInfo in sourceSymbolMap.GetInfosForType(containingSymbol.ContainingType))
+            {
+                if (sourceInfo.TaintedArguments.Any(match => match(parameterSymbol)))
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
         /// Determines if the given array can be a tainted data source when its elements are all constant.
         /// </summary>
         /// <param name="sourceSymbolMap"></param>
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMapExtensions.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMapExtensions.cs
@@ -136,13 +136,15 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
         /// <param name="sourceSymbolMap"></param>
         /// <param name="parameterSymbol"></param>
         /// <returns></returns>
-        public static bool IsSourceParameter(this TaintedDataSymbolMap<SourceInfo> sourceSymbolMap, IParameterSymbol parameterSymbol)
+        public static bool IsSourceParameter(this TaintedDataSymbolMap<SourceInfo> sourceSymbolMap, IParameterSymbol parameterSymbol, WellKnownTypeProvider wellKnownTypeProvider)
         {
             ISymbol containingSymbol = parameterSymbol.ContainingSymbol;
             foreach (SourceInfo sourceInfo in sourceSymbolMap.GetInfosForType(containingSymbol.ContainingType))
             {
-                if (sourceInfo.TaintedArguments.Any(match => match(parameterSymbol)))
+                if (sourceInfo.TaintedArguments.Any(match => match(parameterSymbol, wellKnownTypeProvider)))
+                {
                     return true;
+                }
             }
 
             return false;
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using System.Linq;
 using Analyzer.Utilities.Extensions;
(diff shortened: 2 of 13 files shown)


dotpaul:
Thank you again, @JarLob ! Could you target the 2.9.x branch also? For the time being, we're still releasing 2.9.x packages with security rule fixes and additions.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMapExtensions.cs
```diff
@@ -130,6 +130,24 @@ public static bool IsSourceProperty(this TaintedDataSymbolMap<SourceInfo> source
...
+        /// <param name="sourceSymbolMap"></param>
+        /// <param name="parameterSymbol"></param>
+        /// <returns></returns>
+        public static bool IsSourceParameter(this TaintedDataSymbolMap<SourceInfo> sourceSymbolMap, IParameterSymbol parameterSymbol)
+        {
+            ISymbol containingSymbol = parameterSymbol.ContainingSymbol;
+            foreach (SourceInfo sourceInfo in sourceSymbolMap.GetInfosForType(containingSymbol.ContainingType))
+            {
+                if (sourceInfo.TaintedArguments.Any(match => match(parameterSymbol)))
+                    return true;
```
dotpaul, line 145:
```suggestion
                {
                    return true;
                }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -19,6 +21,21 @@ static WebInputSources()
...
             var sourceInfosBuilder = PooledHashSet<SourceInfo>.GetInstance();
 
+            sourceInfosBuilder.AddSourceInfo(
+                WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerBase,
+                 new ParameterMatcher[]{
+                    (parameter) => {
+                        ISymbol containingSymbol = parameter.ContainingSymbol;
+
+                        if (containingSymbol.DeclaredAccessibility != Accessibility.Public)
+                            return false;
```
dotpaul, line 31:
```suggestion
                        {
                            return false;
                        }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -19,6 +21,21 @@ static WebInputSources()
         {
             var sourceInfosBuilder = PooledHashSet<SourceInfo>.GetInstance();
 
+            sourceInfosBuilder.AddSourceInfo(
+                WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerBase,
+                 new ParameterMatcher[]{
+                    (parameter) => {
+                        ISymbol containingSymbol = parameter.ContainingSymbol;
+
+                        if (containingSymbol.DeclaredAccessibility != Accessibility.Public)
```
dotpaul, line 30:
```suggestion
                        if (containingSymbol.DeclaredAccessibility != Accessibility.Public
                            || containingSymbol.IsConstructor())
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -19,6 +21,21 @@ static WebInputSources()
...
+                            return false;
+
+                        if (containingSymbol.IsConstructor())
```
dotpaul, line 33: (Then this `if` could be removed)

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -19,6 +21,21 @@ static WebInputSources()
         {
             var sourceInfosBuilder = PooledHashSet<SourceInfo>.GetInstance();
 
+            sourceInfosBuilder.AddSourceInfo(
+                WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerBase,
+                 new ParameterMatcher[]{
+                    (parameter) => {
+                        ISymbol containingSymbol = parameter.ContainingSymbol;
+
```
dotpaul, line 29: Definitely should take into account `[NonAction]` attributes, and possibly `[NonController]` attributes. We can't have the `INamedTypeSymbol`s at this point, so may have to do something like look at the `MetadataName` of `AttributeData.AttributeClass`, and walk up their `ContainingNamespace`s checking their `MetadataName`s.

https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/actions?view=aspnetcore-3.1 suggests that controllers don't necessarily have to derive from `ControllerBase`, but that's probably the most common case.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
dotpaul: Also, please add unit tests for these cases.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
JarLob: Do you mean `return !containingSymbol.IsConstructor();`?

JarLob: Ah, I see
```cs
if (containingSymbol.DeclaredAccessibility != Accessibility.Public
                            || containingSymbol.IsConstructor())
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
JarLob: Love the document, is there something like this for the old asp.net mvc?

JarLob:
> ASP.NET (non-Core) MVC controllers would be neat too (but don't feel obligated to add 'em).

Every time I need to add a new reference in unit tests it is a pain. I'll you guys implement it ;)

JarLob:
> Thank you again, @JarLob ! Could you target the 2.9.x branch also? For the time being, we're still releasing 2.9.x packages with security rule fixes and additions.

As the last time? Close this PR open a new one? Or commit in both branches?

JarLob:
Since you are here. Is there a way to define a `return` from such method as a `sink`? Like when it returns a non html encoded string.

dotpaul:
> As the last time? Close this PR open a new one? Or commit in both branches?

Close this PR, open a new one. Or rebase and force push to your fork, and retarget this PR. Definitely only submit a PR to the 2.9.x branch.

> Every time I need to add a new reference in unit tests it is a pain. I'll let you guys implement it ;)

Fair enough. 😄 

> Since you are here. Is there a way to define a `return` from such method as a `sink`? Like when it returns a non html encoded string.

Can you provide an example?

JarLob:
> Can you provide an example?

```cs
public class MyController : Controller
{
    public string DoSomething(string input) 
    {           
        return input; // << input is tainted, treat return as sink
    }
}
```

dotpaul:
Closing in favor of #3966. Continuing discussion of return values as sinks in #3990.

PR #3939, opened 2020-08-02
Support generics types for sinks

JarLob:
Add support for generics in sink type name like `DbSet<TEntity>.FromSql()`.


dotpaul:
Thank you @JarLob ! Looks good to me. Same ask to target the 2.9.x branch.

PR #3944, opened 2020-08-04
Fix AnalyzerReleases.Unshipped.md

agocke:
I put the publish rules in their own table because they're a bit different from normal analyzers


src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -43,6 +43,9 @@ CA2361 | Security | Disabled | DoNotUseDataSetReadXml, [Documentation](https://d
 CA2362 | Security | Disabled | DataSetDataTableInSerializableTypeAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2362)
 CA2363 | Security | Disabled | DataSetDataTableInSerializableTypeAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2363)
 
+### Publish Rules
```
mavasani, line 46: The analyzer that reads this additional file expects some fixed set of table headers. For now, it needs to go under New Rules. Also if custom headers are needed, we would need to update the analyzer. Can you revert all changes to this file and just use the code fix to update it?

PR #3945, opened 2020-08-04
Fix false positive for PointsToAnalysisKind.PartialWithoutTrackingFie…

### LINKED ISSUE #3873: CA2000 FP with ? or ?? operators on a return;
#### Analyzer package

#### Package Version

v3.3.0-beta2.final

#### Diagnostic ID

CA2000: Dispose objects before losing scope

#### Repro steps

Using VS2019 16.6.4, targetting .net framework 4.7.2

CA2000 seems to firing incorrectly on usages where a ?? or ? test is made, suggesting that it's branch analysis doesn't detect that the object is referenced in the branch, eg:
```
using System;
using System.Diagnostics;

namespace CA2000Test
{
    class Program
    {
        {
        }

        public IDisposable TestCa2000(string path, bool boolA, bool boolB)
        {
            if (boolA)
            {
                IDisposable entry = null;
                if (Dictionary.ContainsKey(path))
                {
                    entry = Dictionary[path];
                }

                return entry ?? new Process();
            }

            return !boolB ? new Process() : Dictionary[path];
        }

        public Dictionary<string, IDisposable> Dictionary { get; set; }
    }
}
```

The returns trigger a CA2000 warning, unless there's something very subtle it's trying to tell me, it appears to be a false positive.

#### Expected behavior
No warning

#### Actual behavior
Program.cs(26,29,26,42): warning CA2000: Call System.IDisposable.Dispose on object created by 'new Process()' before all references to it are out of scope.
Program.cs(23,33,23,46): warning CA2000: Call System.IDisposable.Dispose on object created by 'new Process()' before all references to it are out of scope.

mavasani:
…ldsAndProperties

Fixes #3873
This regressed with my recent change to add PointsToAnalysisKind for partial analysis support. We no longer track points to values for fields and properties for `PointsToAnalysisKind.PartialWithoutTrackingFieldsAndProperties`. However, we should still use the default points to value for such entities for the analysis, instead of unknown points to value to ensure that merging values coming from tracked and untracked entities is correct

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
@@ -12327,5 +12327,72 @@ public class Class1
                 LanguageVersion = CSharpLanguageVersion.CSharp8,
             }.RunAsync();
         }
+
+        [Fact, WorkItem(3873, "https://github.com/dotnet/roslyn-analyzers/issues/3873")]
+        public async Task Dispose_ConditionalControlFlow_NoDiagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


mavasani:
Thank you @dotpaul

PR #3953, opened 2020-08-05
Remove CA2363; not sure if we'd run into it

dotpaul:
On second thought, this rule is probably unnecessary.


dotpaul:
Thank you @mavasani !

PR #3956, opened 2020-08-06
Disable CA2109 by default, since Code Access Security isn't a security boundary

dotpaul:


dotpaul:
Thank you @mavasani !

PR #3597, opened 2020-05-02
Refactor issue templates

Evangelink:
Suggest some issue templates that would ease filling specific issues.

This is more of a draft version so feel free to suggest anything you would like / you prefer.

PR #3959, opened 2020-08-06
Exclude non-code files from build

Youssef1313:
The build was triggered on https://github.com/dotnet/roslyn-analyzers/pull/3597 while it doesn't touch any code files.

This should fix that.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,7 @@ pr:
     exclude:
       - docs/*
       - .github/*
+      - **/*.md
 
 jobs:
 - job: Windows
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -5,8 +5,16 @@ trigger:
...
-- 2.9.x
+  branches:
+    include:
+    - master
+    - 2.9.x
+  paths:
+    exclude:
+      - docs/*
+      - .github/*
+      - *.md
```
Youssef1313, line 16: I'm unsure if the `*.md` will work as expected or not. Can someone confirm?

sharwell:
Blocked on https://github.com/dotnet/roslyn/pull/46590

Youssef1313:
@sharwell Should I close this?

sharwell:
Yeah it doesn't look like we'll be able to resolve the blocking issues here in the short term.

PR #3960, opened 2020-08-06
Add option for public API analyzer to bail out on missing API files

mavasani:
https://github.com/dotnet/roslyn-analyzers/pull/3916 added support to public API analyzer to treat missing API files as empty files and report diagnostics. This can be a breaking change for consumers where they relied on old semantics and installed this package or dependent package `Roslyn.Diagnostics.Analyzers` for the whole repo. This change adds an option to fall back the analyzer to old semantics.

### DIFF
--- a/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
+++ b/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
@@ -6,6 +6,8 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using System.Threading;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Text;
@@ -29,6 +31,11 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
         internal const string InvalidReasonShippedCantHaveRemoved = "The shipped API file can't have removed members";
         internal const string InvalidReasonMisplacedNullableEnable = "The '#nullable enable' marker can only appear as the first line in the shipped API file";
         internal const string PublicApiIsShippedPropertyBagKey = "PublicAPIIsShipped";
+
+        /// <summary>
+        /// Boolean option to configure if public API analyzer should bail out silently if public API files are missing.
+        /// </summary>
+        private const string BailOnMissingPublicApiFilesEditorConfigOptionName = "dotnet_public_api_analyzer.require_api_files";
 
         internal static readonly DiagnosticDescriptor DeclareNewApiRule = new DiagnosticDescriptor(
             id: DiagnosticIds.DeclarePublicApiRuleId,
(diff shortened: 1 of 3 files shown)


mavasani:
@Evangelink once this gets merged in, you can update your PR that moves us to newer dogfooding package to set this editorconfig option for our repo.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
```diff
@@ -30,6 +32,11 @@ public sealed partial class DeclarePublicApiAnalyzer : DiagnosticAnalyzer
         internal const string InvalidReasonMisplacedNullableEnable = "The '#nullable enable' marker can only appear as the first line in the shipped API file";
         internal const string PublicApiIsShippedPropertyBagKey = "PublicAPIIsShipped";
 
+        /// <summary>
+        /// Boolean option to configure if public API analyzer should bail out silently if public API files are missing.
+        /// </summary>
+        private const string BailOnMissingPublicApiFilesEditorConfigOptionName = "dotnet_public_api_analyzer.require_api_files";
```
Evangelink, line 38: This should go into `EditorConfigOptionNames`, shouldn't it?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
```diff
@@ -299,6 +313,56 @@ private static bool TryGetApiData(ImmutableArray<AdditionalText> additionalTexts
             return true;
         }
 
+        private static bool TryGetEditorConfigOptionForMissingFiles(AnalyzerOptions analyzerOptions, Compilation compilation, out bool optionValue)
```
Evangelink, line 316: Can't you use/reuse the logic inside `AnalyzerOptionsExtensions` here?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
mavasani: I would like to keep `EditorConfigOptionNames` reserved for the core CA rules package. Another reason for not putting it with other options is that public API analyzer is based off a very old Roslyn version (1.2.1) for backwards compatibility, so it cannot compute use `AnalyzerOptionsExtensions` or related APIs for decoding options. This makes it distinct enough to just keep it isolated to this file and package.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
mavasani: Nope, public API analyzer is built on top of a very old Roslyn API for back compat reasons and we cannot reference the APIs from `AnalyzerOptionsExtensions`, which need Microsoft.CodeAnalysis version 3.x or higher

https://github.com/dotnet/roslyn-analyzers/blob/ed0bbcac3280784cb701db3a44263a27337ec599/src/PublicApiAnalyzers/Core/Analyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.csproj#L10

mavasani:
Thank you @Evangelink!

PR #3961, opened 2020-08-06
Ensure that we do not add duplicate public API files for multi-tfm pr…

### LINKED ISSUE #3952: RS0016 fix-all can result in multiple files being created
Doing a fix-all of `RS0016` on #3922 has lead to a couple of `PublicAPI.Unshipped1.txt` files being created along with the `PublicAPI.Unshipped.txt` file.

mavasani:
…ojects

Fixes #3952

I am not sure how to add a unit test for multi-tfm projects.


mavasani:
FYI: @dougbu

src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
```diff
@@ -295,10 +295,14 @@ protected override async Task<Solution> GetChangedSolutionAsync(CancellationToke
                     newSolution = newSolution.WithAdditionalDocumentText(pair.Key, pair.Value);
                 }
 
+                using var uniqueProjectPaths = PooledHashSet<string>.GetInstance();
                 foreach (KeyValuePair<ProjectId, SourceText> pair in addedPublicSurfaceAreaText)
                 {
                     var project = newSolution.GetProject(pair.Key);
-                    newSolution = AddPublicApiFiles(project, pair.Value);
+                    if (project.FilePath != null && uniqueProjectPaths.Add(project.FilePath))
```
Evangelink, line 302: Is it possible to somehow create a test reproducing the issue so that it's not regressed later on? If not, could you add a note in the src?

src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
```diff
@@ -295,10 +295,14 @@ protected override async Task<Solution> GetChangedSolutionAsync(CancellationToke
                     newSolution = newSolution.WithAdditionalDocumentText(pair.Key, pair.Value);
                 }
 
+                using var uniqueProjectPaths = PooledHashSet<string>.GetInstance();
```
mavasani, line 298:
```suggestion
                // NOTE: We need to avoid creating duplicate files for multi-tfm projects. See https://github.com/dotnet/roslyn-analyzers/issues/3952.
                using var uniqueProjectPaths = PooledHashSet<string>.GetInstance();
```

src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
mavasani: Good idea, I added a comment. I don't know of a good way of creating a unit test with multi-tfm targets.

mavasani:
Thanks @Evangelink

PR #3951, opened 2020-08-05
Support sanitizers that write the sanitized value into output parameter

JarLob:
This allows defining sanitizers like `void HtmlEncode (string s, System.IO.TextWriter output)` that is neither returning a value nor is an object constructor call.

PR #3964, opened 2020-08-06
Support sanitizers that write the sanitized value into output parameter

JarLob:
#3951 PR retargeted to 2.9.x branch

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
@@ -332,10 +332,8 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
                         }
                         else
                         {
-#pragma warning disable CS8602 // Dereference of a possibly null reference.
-                            // sanitizedParameterPairs is not null when isSanitizingMethod is true as per IsSanitizingMethod declaration
+                            RoslynDebug.Assert(sanitizedParameterPairs != null, "sanitizedParameterPairs is not null when isSanitizingMethod is true as per IsSanitizingMethod declaration");
                             foreach ((string ifTaintedParameter, string thenSanitizedTarget) in sanitizedParameterPairs)
-#pragma warning restore CS8602 // Dereference of a possibly null reference.
                             {
                                 if (thenSanitizedTarget == TaintedTargetValue.Return)
                                 {
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
@@ -239,13 +239,15 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
                 PooledHashSet<string>? taintedParameterNamesCached = null;
                 try
                 {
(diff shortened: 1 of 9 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
```diff
@@ -298,11 +321,47 @@ public override TaintedDataAbstractValue VisitObjectCreation(IObjectCreationOper
...
+                            // it was either sanitizing constructor or
+                            // the short form or registering sanitizer method by just the name
+                            result = TaintedDataAbstractValue.NotTainted;
+                        }
+                        else
+                        {
+#pragma warning disable CS8602 // Dereference of a possibly null reference.
+                            // sanitizedParameterPairs is not null when isSanitizingMethod is true as per IsSanitizingMethod declaration
+                            foreach ((string ifTaintedParameter, string thenSanitizedTarget) in sanitizedParameterPairs)
+#pragma warning restore CS8602 // Dereference of a possibly null reference.
```
Evangelink, line 338:
```suggestion
                            RoslynDebug.Assert(sanitizedParameterPairs != null, "sanitizedParameterPairs is not null when isSanitizingMethod is true as per IsSanitizingMethod declaration");
                            foreach ((string ifTaintedParameter, string thenSanitizedTarget) in sanitizedParameterPairs)
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -19,6 +19,23 @@ static WebInputSources()
...
+                WellKnownTypeNames.SystemWebHttpServerUtility,
+                isInterface: false,
+                taintedProperties: null,
+                taintedMethodsNeedsPointsToAnalysis: null,
+                taintedMethodsNeedsValueContentAnalysis: null,
+                transferMethods: new (MethodMatcher, (string, string)[])[]{
+                    (
+                        (methodName, arguments) =>
+                            methodName == "HtmlEncode" &&
+                            arguments.Count() == 2,
```
Evangelink, line 32:
```suggestion
                            arguments.Length == 2,
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
```diff
@@ -298,11 +321,47 @@ public override TaintedDataAbstractValue VisitObjectCreation(IObjectCreationOper
                             }
                         }
                     }
+
+                    if (isSanitizingMethod)
+                    {
+                        if (!sanitizedParameterPairs.Any())
```
Evangelink, line 327: You could probably use `Count` or `Length` here instead.

dotpaul:
@JarLob, do you want to address any of the other comments before I press merge?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
JarLob: I recall previously there was a ban for LINQ in analyzers to avoid hidden allocations. Otherwise what is the benefit of using `Length` over `Any`?

Evangelink: I also thought there was an analyzer to suggest `Count` instead of `Any`, I will need to look if it was disabled or if there as some special cases. `Count`/`Length` is a property access so it's faster and you don't have any allocation.

PR #3966, opened 2020-08-06
Allow tainting function arguments

JarLob:
https://github.com/dotnet/roslyn-analyzers/pull/3938 retargeted to 2.9.x branch


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,57 @@ internal static class WebInputSources
         /// </summary>
         public static ImmutableHashSet<SourceInfo> SourceInfos { get; }
 
+        private static readonly BoundedCacheWithFactory<Compilation, ConcurrentDictionary<ISymbol, bool>> s_sourceArgumentCacheByCompilation = new BoundedCacheWithFactory<Compilation, ConcurrentDictionary<ISymbol, bool>>();
```
dotpaul, line 19: Instead of caching by the `IMethodSymbol` of the potential action method, I think it'd be better to cache the `INamedTypeSymbol` of the potential controller. That'd be a smaller cache and get more cache hits.

Then it's probably okay to check if the `IMethodSymbol` is an action method multiple times without caching, because at that point at least we'd know we're inside a controller.

Also add doc comments for what all the keys and values are.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,57 @@ internal static class WebInputSources
...
         /// </summary>
         static WebInputSources()
         {
             var sourceInfosBuilder = PooledHashSet<SourceInfo>.GetInstance();
 
+            sourceInfosBuilder.AddSourceInfo(
+                WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerBase,
+                WellKnownTypeNames.SystemObject, // checking all System.Object derived types is expensive, so it first check if MicrosoftAspNetCoreMvcControllerBase is resolvable
+                 new ParameterMatcher[]{
+                    (parameter, wellKnownTypeProvider) => {
```
dotpaul, line 32: More logic: https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/dependency-injection?view=aspnetcore-3.1#action-injection-with-fromservices suggests we should ignore parameters with `[FromServices]` attributes

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/SourceTriggeredTaintedDataAnalyzerBase.cs
```diff
@@ -138,6 +138,20 @@ public override void Initialize(AnalysisContext context)
                                 },
                                 OperationKind.PropertyReference);
 
+                            operationBlockStartContext.RegisterOperationAction(
```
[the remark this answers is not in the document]
JarLob: Isn't it already done? https://github.com/dotnet/roslyn-analyzers/pull/3966/files#diff-9ec6d351914231938bd424bba8ea86a8R57

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
JarLob: I have added a second cache to address full recompilation too. Let me know if you think it is overkill.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,81 @@ internal static class WebInputSources
...
+        /// <summary>
+        /// Cached information if the specified symbol is a Asp.Net Core Controller: (compilation) -> ((class symbol) -> (is Controller))
+        /// </summary>
+        private static readonly BoundedCacheWithFactory<Compilation, ConcurrentDictionary<ISymbol, bool>> s_classIsControllerByCompilation =
+            new BoundedCacheWithFactory<Compilation, ConcurrentDictionary<ISymbol, bool>>();
+
+        /// <summary>
+        /// Cached information if the specified symbol is a Asp.Net Core Action: (compilation) -> ((method symbol) -> (is Action))
+        /// </summary>
+        private static readonly BoundedCacheWithFactory<Compilation, ConcurrentDictionary<ISymbol, bool>> s_methodIsActionByCompilation =
```
dotpaul, line 28: > I have added a second cache to address full recompilation too. Let me know if you think it is overkill.

What did you mean by a second cache?

I'm still skeptical of caching by method. I'd prefer not to have a cache grow for each method in code unless that's absolutely necessary. I'd be okay if the methodCache only had entries for methods within a controller. Of course, it's hard to say for sure what's better without data from performance testing and trying different approaches.

I was thinking more along the lines of:
- If `parameter.ContainingSymbol` isn't an `IMethodSymbol`, or `parameter.ContainingSymbol.ContainingSymbol` isn't a `INamedTypeSymbol`, then return false.
- For the `INamedTypeSymbol` of the type, check the classCache if it's a controller or not.
- If not found in the cache:
  - Determine if the type is a controller, add to classCache.
- If type is not a controller, return false.
- If the method isn't action, return false.
- If the parameter has `[FromServices]`, return false.
- Return true.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,81 @@ internal static class WebInputSources
...
+                            return isAction;
+
+                        if (!(parameter.ContainingSymbol is IMethodSymbol methodSymbol))
+                        {
+                            methodCache.TryAdd(parameter.ContainingSymbol, false);
+                            return false;
+                        }
+
+                        var classCache = s_classIsControllerByCompilation.GetOrCreateValue(wellKnownTypeProvider.Compilation, (compilation) => new ConcurrentDictionary<ISymbol, bool>());
+                        if (!classCache.TryGetValue(methodSymbol.ContainingSymbol, out bool isController))
```
dotpaul, line 56: I think the `out bool isController` value needs to be checked. Should probably add a unit test that has a non-controller with multiple methods referencing parameters.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,81 @@ internal static class WebInputSources
...
+                        {
+                            if (!(methodSymbol.ContainingSymbol is INamedTypeSymbol typeSymbol)
+                                || (!typeSymbol.GetBaseTypesAndThis().Any(x => x.Name.EndsWith("Controller", System.StringComparison.Ordinal))
+                                    && (!wellKnownTypeProvider.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerAttribute, out var controllerAttributeTypeSymbol)
```
dotpaul, line 60: If we return false if we can't find a type from `WellKnownTypeProvider`, might as check that up front. Or can make `DependencyFullTypeName` an `ImmutableArray`.

JarLob: I think I lost you here. Is the suggestion to move all `WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName` up?

JarLob:
> Oops, sorry, I meant to say: see if sourceInfoSymbolMap has any SourceInfos that have TaintedArguments, before calling operationBlockStartContext.RegisterOperationAction(..., OperationKind.ParameterReference).

I'm not sure how to implement it... The `sourceInfoSymbolMap` is a `TaintedDataSymbolMap<SourceInfo>` object. I wanted to implement `HasParameterMatcher` similar to existing `IsEmpty`, but the class is only aware of `ITaintedDataInfo`. Should I add a new method to the interface?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
JarLob: Good catch!

JarLob:
Please notice that with the current lock-free implementation, multiple threads may run the same "is controller" path until the result is cached. However I'm not sure if waiting until the first thread calculates it is better.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,72 @@ internal static class WebInputSources
...
+                        if (!(parameter.ContainingSymbol is IMethodSymbol methodSymbol)
+                            || !(methodSymbol.ContainingSymbol is INamedTypeSymbol typeSymbol))
+                        {
+                            return false;
+                        }
+
+                        var classCache = s_classIsControllerByCompilation.GetOrCreateValue(wellKnownTypeProvider.Compilation, (compilation) => new ConcurrentDictionary<ISymbol, bool>());
+                        if (!classCache.TryGetValue(methodSymbol.ContainingSymbol, out bool isController))
+                        {
+                            if ((!typeSymbol.GetBaseTypesAndThis().Any(x => x.Name.EndsWith("Controller", System.StringComparison.Ordinal))
```
dotpaul, line 47: Nit: Could only add to `s_classIsControllerByCompilation` when the entry isn't found:

```csharp
                            if ((!typeSymbol.GetBaseTypesAndThis().Any(x => x.Name.EndsWith("Controller", System.StringComparison.Ordinal))
                                && (!wellKnownTypeProvider.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerAttribute, out var controllerAttributeTypeSymbol)
                                    || !typeSymbol.HasDerivedTypeAttribute(controllerAttributeTypeSymbol)))
                                || !wellKnownTypeProvider.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftAspNetCoreMvcNonControllerAttribute, out var nonControllerAttributeTypeSymbol)
                                || typeSymbol.HasDerivedTypeAttribute(nonControllerAttributeTypeSymbol))
                            {
                                isController = false;
                            }
                            else
                            {
                                isController = true;
                            }

                            classCache.TryAdd(typeSymbol, isController);
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,72 @@ internal static class WebInputSources
         /// </summary>
         public static ImmutableHashSet<SourceInfo> SourceInfos { get; }
 
+        /// <summary>
+        /// Cached information if the specified symbol is a Asp.Net Core Controller: (compilation) -> ((class symbol) -> (is Controller))
+        /// </summary>
+        private static readonly BoundedCacheWithFactory<Compilation, ConcurrentDictionary<ISymbol, bool>> s_classIsControllerByCompilation =
```
dotpaul, line 22: Nit: Could make the type more specific, since we're already checking if `methodSymbol.ContainingType` is an `INamedTypeSymbol`:

```suggestion
        private static readonly BoundedCacheWithFactory<Compilation, ConcurrentDictionary<INamedTypeSymbol, bool>> s_classIsControllerByCompilation =
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,13 +16,72 @@ internal static class WebInputSources
...
+
+                            isController = true;
+                        }
+
+                        if (!isController)
+                        {
+                            return false;
+                        }
+
+                        classCache.TryAdd(methodSymbol.ContainingSymbol, true);
```
dotpaul, line 65: Then if the cache entry is added above, won't need to add it here.

```suggestion
```

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -678,6 +678,76 @@ public static bool HasAttribute(this ISymbol symbol, [NotNullWhen(returnValue: t
...
+        {
+            if (attribute == null)
+            {
+                return false;
+            }
+
+            while (symbol != null)
+            {
+                if (symbol.GetAttributes().Any(attr => attr.AttributeClass.Equals(attribute)))
+                    return true;
```
dotpaul, line 705:
```suggestion
                {
                    return true;
                }
```

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -678,6 +678,76 @@ public static bool HasAttribute(this ISymbol symbol, [NotNullWhen(returnValue: t
...
+
+                if (symbol.BaseType == null)
+                    return false;
```
dotpaul, line 708:
```suggestion
                {
                    return false;
                }
```

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -678,6 +678,76 @@ public static bool HasAttribute(this ISymbol symbol, [NotNullWhen(returnValue: t
...
+        {
+            if (attribute == null)
+            {
+                return false;
+            }
+
+            while (symbol != null)
+            {
+                if (symbol.GetAttributes().Any(attr => attr.AttributeClass.Equals(attribute)))
+                    return true;
```
dotpaul, line 740:
```suggestion
                {
                    return true;
                }
```

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -678,6 +678,76 @@ public static bool HasAttribute(this ISymbol symbol, [NotNullWhen(returnValue: t
...
+
+                if (symbol.OverriddenMethod == null)
+                    return false;
```
dotpaul, line 743:
```suggestion
                {
                    return false;
                }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/ITaintedDataInfo.cs
```diff
@@ -14,11 +16,21 @@ internal interface ITaintedDataInfo
         /// </summary>
         string FullTypeName { get; }
 
+        /// <summary>
+        /// Qualified names of the optional dependency types.
+        /// </summary>
+        ImmutableArray<string>? DependencyFullTypeNames { get; }
```
dotpaul, line 22: `ImmutableArray<T>` is a struct, so would rather not box it.

```suggestion
        ImmutableArray<string> DependencyFullTypeNames { get; }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/SanitizerInfo.cs
```diff
@@ -69,6 +70,16 @@ internal sealed class SanitizerInfo : ITaintedDataInfo, IEquatable<SanitizerInfo
...
 
+        /// <summary>
+        /// Indicates that <see cref="OperationKind.ParameterReference"/> is required.
+        /// </summary>
+        public bool RequiresParameterReferenceAnalysis => false;
+
+        /// <summary>
+        /// Qualified names of the optional dependency types.
+        /// </summary>
+        public ImmutableArray<string>? DependencyFullTypeNames => null;
```
dotpaul, line 81:
```suggestion
        public ImmutableArray<string> DependencyFullTypeNames => ImmutableArray<string>.Empty;
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/SinkInfo.cs
```diff
@@ -57,6 +58,16 @@ public SinkInfo(string fullTypeName, ImmutableHashSet<SinkKind> sinkKinds, bool
...
 
+        /// <summary>
+        /// Indicates that <see cref="OperationKind.ParameterReference"/> is required.
+        /// </summary>
+        public bool RequiresParameterReferenceAnalysis => false;
+
+        /// <summary>
+        /// Qualified names of the optional dependency types.
+        /// </summary>
+        public ImmutableArray<string>? DependencyFullTypeNames => null;
```
dotpaul, line 69:
```suggestion
        public ImmutableArray<string> DependencyFullTypeNames => ImmutableArray<string>.Empty;
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/SourceInfo.cs
```diff
@@ -31,27 +33,36 @@ internal class SourceInfo : ITaintedDataInfo, IEquatable<SourceInfo>
...
             FullTypeName = fullTypeName ?? throw new ArgumentNullException(nameof(fullTypeName));
             IsInterface = isInterface;
             TaintedProperties = taintedProperties ?? throw new ArgumentNullException(nameof(taintedProperties));
+            TaintedArguments = taintedArguments ?? throw new ArgumentNullException(nameof(taintedArguments));
             TaintedMethods = taintedMethods ?? throw new ArgumentNullException(nameof(taintedMethods));
             TaintedMethodsNeedsPointsToAnalysis = taintedMethodsNeedsPointsToAnalysis ?? throw new ArgumentNullException(nameof(taintedMethodsNeedsPointsToAnalysis));
             TaintedMethodsNeedsValueContentAnalysis = taintedMethodsNeedsValueContentAnalysis ?? throw new ArgumentNullException(nameof(taintedMethodsNeedsValueContentAnalysis));
             TransferMethods = transferMethods ?? throw new ArgumentNullException(nameof(transferMethods));
             TaintConstantArray = taintConstantArray;
+            DependencyFullTypeNames = dependencyFullTypeNames;
```
dotpaul, line 53:
```suggestion
            DependencyFullTypeNames = dependencyFullTypeNames ?? ImmutableArray<string>.Empty;
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/SourceInfo.cs
```diff
@@ -31,27 +33,36 @@ internal class SourceInfo : ITaintedDataInfo, IEquatable<SourceInfo>
...
 
         /// <summary>
         /// Full type name of the...type (namespace + type).
         /// </summary>
         public string FullTypeName { get; }
 
+        /// <summary>
+        /// Full type names of the optional dependency/referenced types that should be resolved.
+        /// </summary>
+        public ImmutableArray<string>? DependencyFullTypeNames { get; }
```
dotpaul, line 64:
```suggestion
        public ImmutableArray<string> DependencyFullTypeNames { get; }
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMap.cs
```diff
@@ -16,6 +16,23 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
     internal class TaintedDataSymbolMap<TInfo> : IEquatable<TaintedDataSymbolMap<TInfo>?>
         where TInfo : ITaintedDataInfo
     {
+        private static bool TryResolveDependencies(TInfo info, WellKnownTypeProvider wellKnownTypeProvider)
+        {
+            if (info.DependencyFullTypeNames == null)
```
dotpaul, line 21:
```suggestion
            if (!info.DependencyFullTypeNames.IsEmpty)
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMap.cs
```diff
@@ -16,6 +16,23 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
...
+            {
+                return true;
+            }
+
+            foreach (var dependency in info.DependencyFullTypeNames)
```
dotpaul, line 26:
```suggestion
            foreach (string dependency in info.DependencyFullTypeNames)
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
```diff
@@ -12,6 +16,18 @@ internal static class WebInputSources
         /// </summary>
         public static ImmutableHashSet<SourceInfo> SourceInfos { get; }
 
+        /// <summary>
+        /// Cached information if the specified symbol is a Asp.Net Core Controller: (compilation) -> ((class symbol) -> (is Controller))
+        /// </summary>
+        private static readonly BoundedCacheWithFactory<Compilation, ConcurrentDictionary<INamedTypeSymbol, bool>> s_classIsControllerByCompilation =
+            new BoundedCacheWithFactory<Compilation, ConcurrentDictionary<INamedTypeSymbol, bool>>();
+
+        private static readonly ImmutableArray<string> s_dependencyFullTypeNames = ImmutableArray.Create<string>(WellKnownTypeNames.MicrosoftAspNetCoreMvcControllerBase,
```
dotpaul, line 25: Nit: The source infos are initialized in a static constructor anyway, so could do without a static field here.

dotpaul: (The static field being `s_dependencyFullTypeNames`)

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMap.cs
JarLob: Needs to be inverted.

JarLob: Since it is used in foreach later I wonder if the check is needed at all.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
JarLob: Same applies to `s_classIsControllerByCompilation`?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
```diff
@@ -115,6 +115,16 @@ protected override void ApplyMissingCurrentAnalysisDataForUnhandledExceptionData
                 base.ApplyMissingCurrentAnalysisDataForUnhandledExceptionData(dataAtException.CoreAnalysisData, CurrentAnalysisData.CoreAnalysisData, throwBranchWithExceptionType);
             }
 
+            protected override TaintedDataAbstractValue GetDefaultValueForParameterOnEntry(IParameterSymbol parameter, AnalysisEntity analysisEntity)
+            {
+                if (this.DataFlowAnalysisContext.SourceInfos.IsSourceParameter(parameter, WellKnownTypeProvider))
+                {
+                    return TaintedDataAbstractValue.CreateTainted(parameter, parameter.DeclaringSyntaxReferences[0].GetSyntax(), this.OwningSymbol);
```
JarLob, line 122: `parameter.DeclaringSyntaxReferences[0].GetSyntax()` - terrible hack?

dotpaul: @mavasani, do you have any better ideas here? We want to track the location of the origin of the tainted data (in this case, a method parameter).

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/WebInputSources.cs
dotpaul: Nah, the cache is fine.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataSymbolMap.cs
dotpaul: Oh, whoops, sorry for the incorrect `!`.

Good call, don't need the check with the foreach.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataAnalysis.TaintedDataOperationVisitor.cs
mavasani: Given that you eventually need to store the location, I don't see a good alternative: https://github.com/dotnet/roslyn-analyzers/blob/8bc7b0bf842d3fee3ee8595e783828f8c4ee448e/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/SymbolAccess.cs#L39-L42

For PointsToAbstractValue, we need similar special casing for parameters as there is no associated operation. We handle that by creating a special abstract location: https://github.com/dotnet/roslyn-analyzers/blob/8bc7b0bf842d3fee3ee8595e783828f8c4ee448e/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/AbstractLocation.cs#L63-L64

Not sure you can do that here as you seem to really need to store the location in your symbol access type.

dotpaul: I think this approach is fine for now.

If we run into weirdness with it, maybe we can do some special casing like:

1. In `GetDefaultValueForParameterOnEntry()`, we somehow mark the abstract value as needing its syntax location needs to be updated.
2. In `ComputeAnalysisValueForReferenceOperation()` (or somewhere), we see if it's an `IParameterReferenceOperation`, and if its current abstract value is still tainted and marked for its syntax location needs to be updated, return a new abstract value with the `IParameterReferenceOperation`'s syntax location.

PR #3972, opened 2020-08-07
CA1024: Do not report on tasks return types

### LINKED ISSUE #3877: Update CA1024 to not report on get method return tasks
#### Analyzer package

#### Package Version

v3.3.0-beta1.final (Latest pre-release)

#### Diagnostic ID

Example: [CA1024](https://docs.microsoft.com/visualstudio/code-quality/ca1024)

#### Repro steps

Over simplified repro:

```csharp
using System.Threading.Tasks;
public class Something
{
    public Task GetContent()
    {
        return null;
    }
}
```

#### Expected behavior

No diagnostic reported. The docs states that 

> Properties should behave as if they are fields; if the method cannot, it should not be changed to a property.

having `Task`/`ValueTask` return type usually imply there will be some async/await going-on and I don't think it's a good practice for a property.

#### Actual behavior

Diagnostic is reported.

Evangelink:
Fix #3877

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
@@ -42,47 +42,63 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             analysisContext.EnableConcurrentExecution();
             analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
-            analysisContext.RegisterOperationBlockStartAction(context =>
+            analysisContext.RegisterCompilationStartAction(context =>
             {
+                var taskTypesBuilder = ImmutableHashSet.CreateBuilder<ITypeSymbol>();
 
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
```diff
@@ -42,47 +42,63 @@ public override void Initialize(AnalysisContext analysisContext)
...
-                    if (!hasInvocations)
+                    // A few additional checks to reduce the noise for this diagnostic:
+                    // Ensure that the method is non-generic, non-virtual/override, has no overloads and doesn't have special names: 'GetHashCode' or 'GetEnumerator'.
+                    // Also avoid generating this diagnostic if the method body has any invocation expressions.
+                    // Also avoid implicit interface implementation (explicit are handled through the member accessibility)
+                    if (methodSymbol.IsGenericMethod ||
+                        methodSymbol.IsVirtual ||
+                        methodSymbol.IsOverride ||
+                        methodSymbol.Name == GetHashCodeName ||
+                        methodSymbol.Name == GetEnumeratorName ||
```
Evangelink, line 81: I think these checks shall be done before the `MatchesConfiguredVisibility`.

PR #3973, opened 2020-08-07
Update resource analyzers to flag leading and trailing whitespaces

### LINKED ISSUE #3958: Update resources analyzers to flag leading and trailing spaces
This should be fixed. We should update the analyzer to flag leading and trailing spaces for all 3 kind of strings.

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/pull/3922_

Evangelink:
Fix #3958


src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs
```diff
@@ -3553,6 +3722,177 @@ End Class
...
         }
 
+        [Fact, WorkItem(3958, "https://github.com/dotnet/roslyn-analyzers/issues/3958")]
+        public async Task RS1032_LeadingOrTrailingWhitespaces_Diagnostic()
+        {
+            var additionalFileName = "Resources.resx";
+            var additionalFileText = @"
+<root>
+  <data name=""AnalyzerMessage1"" xml:space=""preserve"">
+    <value>Message with trailing whitespace </value>
```
mavasani, line 3732: Can you add another scenario where trailing spaces are after a period, similar to the case caught in the original PR, say `Message with trailing whitespaces after a period.   `? I have a feeling your current logic might be need to be tweaked to move whitespace fixing part above multi-line fixing part.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs
```diff
@@ -605,6 +610,8 @@ private static void AnalyzeMessageCore(string message, IArgumentOperation argume
                     fixedMessage = endsWithPeriod ? RemoveTrailingPeriod(fixedMessage) : fixedMessage + ".";
                 }
 
+                fixedMessage = hasLeadingOrTrailingWhitespaces ? RemoveLeadingAndTrailingWhitespaces(fixedMessage) : fixedMessage;
```
mavasani, line 613: I think this might have to move ahead to just after we fix line returns but before the above if statement.

Issue #3954, opened 2020-08-05
Change CA1836 from diagnosing all types to only the ones proven to be worthy


Jozkee:
Specifically types in `System.Collections.Concurrent`.

[It was proven](https://github.com/dotnet/runtime/pull/39890#discussion_r460098557) that there are .NET types (`ImmutableArray<T>`) that by using the suggested fix, would suffer a perf. regression and there are also internal types [`XmlQueryType`](https://github.com/dotnet/runtime/pull/39890#discussion_r459966637) that have an `IsEmpty` property that is semantically different to `Count == 0`; there are also a some scenarios in the wild that backup the latter concern.

- https://github.com/edisontung/playscript-mono/blob/87f9d7c0f81741d0076c70f22ce2de73f872c53e/mcs/class/System/System.Text.RegularExpressions/interval.cs#L79
- https://github.com/akil03/Kwizz/blob/bce8b10a14bd6793eaf32eb28566826c2018fb9e/Assets/UI/DoozyUI/Scripts/ScriptableObjects/NamesDatabase.cs
- https://github.com/Mauru/MauronAlpha.Csharp/blob/eb8303e248c459795501571a53804fc3d06dbf03/MauronAlpha.TextProcessing/Units/Word.cs

Another point in favor of this turn is the recent change done to exclude Memory/Span types https://github.com/dotnet/roslyn-analyzers/issues/3810.

With that said, we should turn the rule from diagnose for all types that contain a `Count` or `Length` property and an `IsEmpty` property  to diagnose only for Concurrent types and then go from there adding up more types that are proven to be beneficent.

cc @jeffhandley @bartonjs @stephentoub @jkotas @mavasani @sharwell

mavasani:
@Jozkee If you don't mind, can you add a [configurable option](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#supported-editorconfig-options) for this rule which will fallback to current logic and ignore this allow list? It should be very simple to add such an option, for example see this commit: https://github.com/dotnet/roslyn-analyzers/commit/6d6f99588fdfd80de599d71627652e9fa3de3e1c#diff-1b5cdc4937a8cdfc79545945ad6c0783

mavasani:
@Jozkee if adding an option is leading to too much work for you, I can do the work of adding the option as follow-up PR as well. Feel free to skip that part for your original PR. Thanks!

jeffhandley:
[...]

PR #3978, opened 2020-08-07
CA1836 - Switch to allow list that only specifies concurrent collections

Jozkee:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/3954

cc @jeffhandley


Jozkee:
I will proceed to run the rule in dotnet/runtime and open up a PR with the proper fixes.
I don't think we need to update documentation after this change.

mavasani:
> I don't think we need to update documentation after this change.

I think we should consider mentioning the allow list on which the rule operates in the documentation, similar to https://docs.microsoft.com/visualstudio/code-quality/ca1826?view=vs-2019#rule-description

PR #3981, opened 2020-08-09
Fix NRE in BasicDoNotUseCountWhenAnyCanBeUsedFixer

### LINKED ISSUE #3749: 'BasicDoNotUseCountWhenAnyCanBeUsedFixer' encountered an error
#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

[CA1827](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1827)

#### Repro steps

image

#### StackTrace

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at Microsoft.CodeAnalysis.SyntaxNodeExtensions.WithTrailingTrivia[TSyntax](TSyntax node,IEnumerable`1 trivia)
   at async Microsoft.NetCore.Analyzers.Performance.DoNotUseCountWhenAnyCanBeUsedFixer.DoNotUseCountWhenAnyCanBeUsedCodeAction.GetChangedDocumentAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.<>c__DisplayClass11_0.<GetPreviewAsync>b__0(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Extensions.IExtensionManagerExtensions.PerformFunctionAsync[T](<Unknown Parameters>)
   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)
```

Note: The fix message mentioned AnyAsync and CountAsync, while my method isn't async at all.

image

Note2: I haven't tested with the latest master.

Youssef1313:
Fixes #3749

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
@@ -457,7 +457,7 @@ Module C
     End Function
     Sub M()
         Dim b As Boolean
-
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(IntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(IntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(UIntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(LongZero)|}}
@@ -502,6 +502,7 @@ Module C
     Sub M()
         Dim b As Boolean
 
+        b = Not GetData().Any()
         b = Not GetData().Any()
         b = Not GetData().Any()
         b = Not GetData().Any()
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
@@ -457,6 +457,7 @@ Module C
     End Function
     Sub M()
         Dim b As Boolean
+
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(IntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(IntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(UIntegerZero)|}}
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
@@ -992,6 +992,7 @@ Namespace System.Data.Entity
         Async Sub M()
             Dim b As Boolean
 
+            b = {{|{this.Verifier.DiagnosticId}:(Await GetData().CountAsync).Equals(IntegerZero)|}}
             b = {{|{this.Verifier.DiagnosticId}:(Await GetData().CountAsync()).Equals(IntegerZero)|}}
             b = {{|{this.Verifier.DiagnosticId}:(Await GetData().CountAsync()).Equals(UIntegerZero)|}}
             b = {{|{this.Verifier.DiagnosticId}:(Await GetData().CountAsync()).Equals(LongZero)|}}
@@ -1049,6 +1050,7 @@ Namespace System.Data.Entity
         Async Sub M()
             Dim b As Boolean
 
+            b = Not Await GetData().AnyAsync()
             b = Not Await GetData().AnyAsync()
             b = Not Await GetData().AnyAsync()
             b = Not Await GetData().AnyAsync()
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
@@ -460,28 +460,47 @@ Module C
 
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(IntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(IntegerZero)|}}
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(UIntegerZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(UIntegerZero)|}}
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(LongZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(LongZero)|}}
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(ULongZero)|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count().Equals(ULongZero)|}}
 
+        b = {{|{this.Verifier.DiagnosticId}:IntegerZero.Equals(GetData().Count)|}}
         b = {{|{this.Verifier.DiagnosticId}:IntegerZero.Equals(GetData().Count())|}}
+        b = {{|{this.Verifier.DiagnosticId}:UIntegerZero.Equals(GetData().Count)|}}
         b = {{|{this.Verifier.DiagnosticId}:UIntegerZero.Equals(GetData().Count())|}}
+        b = {{|{this.Verifier.DiagnosticId}:LongZero.Equals(GetData().Count)|}}
         b = {{|{this.Verifier.DiagnosticId}:LongZero.Equals(GetData().Count())|}}
+        b = {{|{this.Verifier.DiagnosticId}:ULongZero.Equals(GetData().Count)|}}
         b = {{|{this.Verifier.DiagnosticId}:ULongZero.Equals(GetData().Count())|}}
 
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count = IntegerZero|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count() = IntegerZero|}}
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count >= IntegerOne|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count() >= IntegerOne|}}
+        b = {{|{this.Verifier.DiagnosticId}:IntegerZero = GetData().Count|}}
         b = {{|{this.Verifier.DiagnosticId}:IntegerZero = GetData().Count()|}}
+        b = {{|{this.Verifier.DiagnosticId}:IntegerOne > GetData().Count|}}
         b = {{|{this.Verifier.DiagnosticId}:IntegerOne > GetData().Count()|}}
 
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count = UIntegerZero|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count() = UIntegerZero|}}
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count >= UIntegerOne|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count() >= UIntegerOne|}}
+        b = {{|{this.Verifier.DiagnosticId}:UIntegerZero = GetData().Count|}}
         b = {{|{this.Verifier.DiagnosticId}:UIntegerZero = GetData().Count()|}}
+        b = {{|{this.Verifier.DiagnosticId}:UIntegerOne > GetData().Count|}}
         b = {{|{this.Verifier.DiagnosticId}:UIntegerOne > GetData().Count()|}}
 
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count = LongZero|}}
         b = {{|{this.Verifier.DiagnosticId}:GetData().Count() = LongZero|}}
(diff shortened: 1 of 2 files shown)


Youssef1313:
This PR just mirrors the original fix made by @mavasani in 2404ec0 to the VB fixer, and added a test to confirm if it works or not.

Youssef1313:
After mirroring 2404ec0, the NRE is gone, but the fixer isn't working. I'll debug that soon and try to figure out where things goes wrong.

Youssef1313:
I think the async case might still be broken. I can't think of a good solution.
I'm thinking of refactoring the whole fixer to use semantic analysis instead of syntax analysis, that would be ideal IMO, but very tedious.

Any ideas @mavasani @Evangelink?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
```diff
@@ -104,7 +104,7 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
 
         End Function
 
-        Private Shared Sub GetExpressionAndInvocationArguments(sourceExpression As ExpressionSyntax, isAsync As Boolean, ByRef expression As SyntaxNode, ByRef arguments As IEnumerable(Of SyntaxNode))
+        Private Shared Function TryGetExpressionAndInvocationArguments(sourceExpression As ExpressionSyntax, isAsync As Boolean, ByRef expression As SyntaxNode, ByRef arguments As IEnumerable(Of SyntaxNode)) As Boolean
```
Evangelink, line 107: I am not good with vbnet and I have noticed the code was already the same before but isn't it expected to have the `Out` attribute for the last 2 parameters?

Evangelink (commented):
> I think the async case might still be broken. I can't think of a good solution.
I'm thinking of refactoring the whole fixer to use semantic analysis instead of syntax analysis, that would be ideal IMO, but very tedious.

Not sure to understand which case you are referring to, could you please add a test even if failing?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
Youssef1313: I think OutAttribute is usually used for interop scenarios. But it won't be necessary here.
However, I'm not 100% sure.

Youssef1313:
@Evangelink I added the failing test for the async case

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
```diff
@@ -1051,6 +1054,7 @@ End Function
             b = Not Await GetData().AnyAsync()
```
Evangelink, line 1054: That remains as (Await GetData().CountAsync).Equals(IntegerZero)

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
```diff
@@ -133,18 +133,25 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
...
+                expression = DirectCast(invocationExpression.Expression, MemberAccessExpressionSyntax).Expression
+                arguments = invocationExpression.ArgumentList.ChildNodes()
+                Return True
 
             End If
 
-            expression = DirectCast(invocationExpression.Expression, MemberAccessExpressionSyntax).Expression
-            arguments = invocationExpression.ArgumentList.ChildNodes()
+            Dim memberAccessExpression = TryCast(sourceExpression, MemberAccessExpressionSyntax)
+            If memberAccessExpression IsNot Nothing Then ' This case happens for something like: x = GetData().Count where Count method is called without parentheses.
```
Evangelink, line 145: There shall be a similar logic on line 126 (inside of the ` If isAsync Then` block)

mavasani:
@Jozkee for review

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
```diff
@@ -458,6 +458,7 @@ End Function
     Sub M()
         Dim b As Boolean
 
+        b = {{|{this.Verifier.DiagnosticId}:GetData().Count.Equals(IntegerZero)|}}
```
Jozkee, line 461: Can you add this non parentheses scenario for `0.Equals(GetData().Count)`, `GetData().Count = 0` and  `0 = GetData().Count` cases below?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
```diff
@@ -990,6 +992,7 @@ End Function
         Async Sub M()
             Dim b As Boolean
 
+            b = {{|{this.Verifier.DiagnosticId}:(Await GetData().CountAsync).Equals(IntegerZero)|}}
```
Jozkee, line 995: Same as my previous comment, could you please add this case on the other expressions?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
```diff
@@ -38,13 +38,13 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
...
-                            GetExpressionAndInvocationArguments(
+                            If TryGetExpressionAndInvocationArguments(
                                 sourceExpression:=member.Expression,
                                 isAsync:=isAsync,
                                 expression:=expression,
-                                arguments:=arguments)
-
-                            Return True
+                                arguments:=arguments) Then
+                                Return True
```
Jozkee, line 46: You can `Return TryGetExpressionAndInvocationArguments` instead, same for the other cases.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
Jozkee, line 145: Move the comment to a new line on top of line 144 so its clear that this is a special scenario.

jozkee (changes requested):
Thanks for the PR, could you please fix the Async scenario that is currently failing? 
I left a suggestion next to @Evangelink's comment on how you can fix it.

I also left a few other comments related to to improve tests and some style suggestions.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
Jozkee: Correct, that's why `Async` scenarios were not working.

Instead of duplicating the code on the async case, you can refactor the conditions to something like this:
```vb
' Extract source expression from Async expression if needed.
If isAsync Then

    Dim awaitExpressionSyntax = TryCast(sourceExpression, AwaitExpressionSyntax)

    If Not awaitExpressionSyntax Is Nothing Then
        sourceExpression = awaitExpressionSyntax.Expression
    End If

End If

' Try to get source expression as InvocationExpression.
Dim invocationExpression As InvocationExpressionSyntax = TryCast(sourceExpression, InvocationExpressionSyntax)

If invocationExpression IsNot Nothing Then

    expression = DirectCast(invocationExpression.Expression, MemberAccessExpressionSyntax).Expression
    arguments = invocationExpression.ArgumentList.ChildNodes()
    Return True

End If

'  Try to get source expression as MemberAccessExpression which is only legal on VB.
Dim memberAccessExpression = TryCast(sourceExpression, MemberAccessExpressionSyntax)

If memberAccessExpression IsNot Nothing Then
    expression = memberAccessExpression.Expression
    arguments = Enumerable.Empty(Of SyntaxNode)()
    Return True
End If
```

mavasani:
@Youssef1313 would you be able to resolve the review comments?

mavasani:
@Jozkee can you please take another look?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
Jozkee: That is not the case for the parentheses scenario `(Await GetData().CountAsync()).Equals(IntegerZero)` or am I missing something?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
Jozkee, line 46:
```suggestion
                                Return TryGetExpressionAndInvocationArguments(
                                    sourceExpression:=member.Expression,
                                    isAsync:=isAsync,
                                    expression:=expression,
                                    arguments:=arguments)
```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
```diff
@@ -115,36 +108,41 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
 
             End While
 
-            Dim invocationExpression As InvocationExpressionSyntax = Nothing
-
             If isAsync Then
 
                 Dim awaitExpressionSyntax = TryCast(sourceExpression, AwaitExpressionSyntax)
 
                 If Not awaitExpressionSyntax Is Nothing Then
```
Jozkee, line 115: Nit:
```suggestion
                If awaitExpressionSyntax IsNot Nothing Then
```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
```diff
@@ -115,36 +108,41 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
...
+                Return True
 
             End If
 
-            expression = DirectCast(invocationExpression.Expression, MemberAccessExpressionSyntax).Expression
-            arguments = invocationExpression.ArgumentList.ChildNodes()
+            '  Try to get source expression as MemberAccessExpression which is only legal on VB.
+            Dim memberAccessExpression = TryCast(sourceExpression, MemberAccessExpressionSyntax)
+
+            If memberAccessExpression IsNot Nothing Then ' This case happens for something like: x = GetData().Count where Count method is called without parentheses.
```
Jozkee, line 136: Nit: 
```suggestion
             ' This case happens for something like: x = GetData().Count where Count method is called without parentheses.
            If memberAccessExpression IsNot Nothing Then
```

jozkee (approved):
Otherwise; lgtm, thanks.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicDoNotUseCountWhenAnyCanBeUsedFixer.vb
Youssef1313: These are already "IsNot" in master, They was fixed after I created this branch.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/DoNotUseCountWhenAnyCanBeUsedTests.Tests.cs
Youssef1313: @Jozkee I couldn't get what you mean

Jozkee: I was replying to @Evangelink's suggestion; what I got from his comment was that `(Await GetData().CountAsync).Equals(IntegerZero)` (CountAsync without parentheses) should not be caught by the analyzer and I said that I disagree given that the analyzer already catches `(Await GetData().CountAsync()).Equals(IntegerZero)` (CountAsync with parentheses).

We should be consistent.

Youssef1313: @Jozkee The comment from @Evangelink was before I had fixed the async case. He was pointing to the test failure that was expecting `Not Await GetData().AnyAsync()`, but the actual was `(Await GetData().CountAsync).Equals(IntegerZero)`

Jozkee: So, that's what I was missing :)
Thanks for clarifying.

PR #3998, opened 2020-08-12
Update Versions.props

jmarolf:
Changing the version number to match the .NET release number and have the pre-release version label match the .NET preview label (rc1).


jmarolf:
> See my other PR which does this.

Can you point me at that? its unclear what markdown files are documentation and what are infrastructure in the repo

PR #4001, opened 2020-08-13
Use EffectiveAnalysisLevel instead of AnalysisLevel for warning wave mapping

mavasani:
Relates to https://github.com/dotnet/sdk/pull/12898. This allows us to avoid dealing with special values for AnalysisLevel such as 'none', 'preview', 'latest', etc.

Verified that we now correctly handle values such as `5`, `5.0`, `5.0.0.0` and so on to map to `AnalysisLevel_5.editorconfig` for warning wave.

### DIFF
--- a/src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
+++ b/src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
@@ -371,8 +371,8 @@ $@"<Project>{GetCommonContents(packageName)}{GetPackageSpecificContents(packageN
                     if (packageName == "Microsoft.CodeAnalysis.NetAnalyzers")
                     {
                         return $@"
-      <!-- Default '{packageVersionPropName}' to 'AnalysisLevel' -->
-      <{packageVersionPropName} Condition=""'$({packageVersionPropName})' == '' and $(AnalysisLevel) != ''"">$(AnalysisLevel)</{packageVersionPropName}>
+      <!-- Default '{packageVersionPropName}' to 'EffectiveAnalysisLevel' with trimmed trailing '.0' -->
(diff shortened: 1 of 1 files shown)


src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -371,8 +371,8 @@ static string GetPropertyStringForSettingDefaultPropertyValue(string packageName
                     if (packageName == "Microsoft.CodeAnalysis.NetAnalyzers")
                     {
                         return $@"
-      <!-- Default '{packageVersionPropName}' to 'AnalysisLevel' -->
-      <{packageVersionPropName} Condition=""'$({packageVersionPropName})' == '' and $(AnalysisLevel) != ''"">$(AnalysisLevel)</{packageVersionPropName}>
+      <!-- Default '{packageVersionPropName}' to 'EffectiveAnalysisLevel' with trimmed trailing '.0' -->
+      <{packageVersionPropName} Condition=""'$({packageVersionPropName})' == '' and $(EffectiveAnalysisLevel) != ''"">$([System.Text.RegularExpressions.Regex]::Replace($(EffectiveAnalysisLevel), '(.0)*$', ''))</{packageVersionPropName}>
```
jmarolf, line 375: If the goal is to remove the trailing 0s couldn't you just do `[System.Int32]::Parse($(EffectiveAnalysisLevel), [System.Globalization.NumberStyles]::Any)`

mavasani: Will that not throw an exception if a user has invalid value for `EffectiveAnalysisLevel`, say some random non-numerical string? I thought this was more safer.

jmarolf: I believe a regex that just strips trailing zeros will be safer. I am torn between failing fast for unknown inputs and being as resilient as possible.

PR #4005, opened 2020-08-13
Create CryptographicHardwareIntrinsicsAnalyzer.cs

### LINKED ISSUE #3646: Suggestion: New analyzer rule that warns when using cryptographic hardware intrinsics
In .NET Core 3.0, we added support for x86 hardware intrinsics, including for AES-NI (via the type [`System.Runtime.Intrinsics.X86.Aes`](https://apisof.net/catalog/System.Runtime.Intrinsics.X86.Aes)). We added these types because they may have valid use cases for a very limited set of callers. However, the number of people who should be calling these APIs is __extremely__ limited. There exist numerous pitfalls for these APIs - far too numerous for me to list here - that aren't immediately apparent. And even expert developers who are intimately familiar with cryptographic algorithms are likely to get tripped up by these.

We should introduce an analyzer which looks for references to this type and which says "hey, you really should be using the [`System.Security.Cryptography.Aes`](https://apisof.net/catalog/System.Security.Cryptography.Aes) type instead."

__Category:__ Security

Youssef1313:
@mavasani @GrabYourPitchforks, I need an initial review on this.

Fixes #3646


mavasani:
Tagging @dotpaul as this is a security rule.

dotpaul:
Thank you @Youssef1313 ! Could you please target the **2.9.x** branch for this PR? For the time being, we're shipping new security rules and updates in v2.9 releases as well.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/CryptographicHardwareIntrinsicsAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(
+                compilationStartContext =>
+                {
+                    INamedTypeSymbol symbol = compilationStartContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeIntrinsicsX86Aes);
+
+                    compilationStartContext.RegisterOperationAction(
+                        context => AnalyzeObjectCreation(context, tooGenericExceptionSymbols, reservedExceptionSymbols),
+                        OperationKind.ObjectCreation);
```
Youssef1313, line 47: Should I use `OperationKind.Invocation` here instead? I'm unsure.

dotpaul: Yes. `System.Runtime.Intrinsics.X86.Aes` looks like a static class, i.e. it can't be instantiated, so you wouldn't see ObjectCreation operations.

Try playing around the `System.Runtime.Intrinsics.X86.Aes` API, and then you'll have an idea of what unit tests to write.

dotpaul: Also, since it's an abstract class, we should also make sure no one is deriving from it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/CryptographicHardwareIntrinsicsTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        public async Task InvokingDecryptLast_CS_Diagnostic()
+        {
+            var test = new VerifyCS.Test
+            {
+                ReferenceAssemblies = new ReferenceAssemblies(
+                    "netcoreapp3.1",
+                    new PackageIdentity(
+                        "Microsoft.NETCore.App.Ref",
+                        "3.1.0"),
+                    Path.Combine("ref", "netcoreapp3.1")),
```
Youssef1313, line 35: :warning: I had to do it that way because `ReferenceAssemblies.NetCore.NetCoreApp31` doesn't exist. It's probably because an older version of roslyn SDK being used in 2.9.x branch.

@mavasani, Do you want to update the sdk version instead? If so, please inform me where the version should be updated.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/CryptographicHardwareIntrinsicsAnalyzer.cs
GrabYourPitchforks: We shouldn't need to worry about deriving from it. All of the intrinsics are abstract classes with internal ctors.

GrabYourPitchforks:
Don't forget these classes as well:

* System.Runtime.Intrinsics.Arm.Aes
* System.Runtime.Intrinsics.Arm.Sha1
* System.Runtime.Intrinsics.Arm.Sha256

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/CryptographicHardwareIntrinsicsAnalyzer.cs
dotpaul: Oh okay, cool.

Youssef1313:
Closing as the discussion in the issue has moved towards using EditorBrowsable for these APIs.

PR #3917, opened 2020-07-28
Platform compatibility analyzer

buyaa-n:
## Customer Impact

Customers could get a warning if they referenced platform-specific APIs

## Testing
Unit tests, manual tests, and dogfooding within the `dotnet/runtime` repo on local builds.

## Risk
Medium
 
### Description: 
We have added new attributes to annotate APIs being platform-specific; `SupportedOSPlatform` and `UnsupportedOSPlatform`.
Based on them we want a Roslyn analyzer that informs developers when they use platform-specific APIs from call sites where the API might not be available.

The semantics of these attributes are updated to become as follows:
- An API that doesn't have any of these attributes is considered supported by all platforms.
- If either `[SupportedOSPlatform` or `[UnsupportedOSPlatform]` attributes are present, we group all attributes by OS platform identifier:
  - Allow list. If the lowest version for each OS platform is a `[SupportedOSPlatform]` attribute, the API is considered to only be supported by the listed platforms and unsupported by all other platforms.
  - Deny list. If the lowest version for each OS platform is a `[UnsupportedOSPlatform]` attribute, then the API is considered to only be unsupported by the listed platforms and supported by all other platforms.
  - Inconsistent list. If for some platforms the lowest version attribute is `[SupportedOSPlatform]` while for others it is `[UnsupportedOSPlatform]`
- Both attributes can be instantiated without version numbers. This means the version number is assumed to be 0.0. This simplifies guard clauses, see examples below for more details.

#### Sample 1: Deny list. 
The API is unsupported only on Windows, then added support from version 10.0.1903, then removed from version 10.0.2004:
```csharp
[UnsupportedOSPlatform("windows")]
[SupportedOSPlatform("windows10.0.1903")]
[UnsupportedOSPlatform("windows10.0.2004")]
public void DoesNotWorkOnWindows();
```
#### Sample 2: Allow list: 
Api only allowed on ios12.0 - 14.0 and on ipados from version 13.0
```csharp
[SupportedOSPlatform("ios12.0")]
[UnsupportedOSPlatform("ios14.0")]
[SupportedOSPlatform("ipados13.0")]
public void OnlyWorksOniOSAndIpadOs();
```

Related design doc https://github.com/dotnet/designs/blob/master/accepted/2020/platform-checks/platform-checks.md,
https://github.com/dotnet/designs/blob/master/accepted/2020/platform-exclusion/platform-exclusion.md

Runtime issue: dotnet/runtime#37359 

### Finding Diagnostics

[...]

PR #4026, opened 2020-08-16
Workaround for https://github.com/dotnet/roslyn/issues/46859 

mavasani:
**Strongly recommended to review ignoring white space diffs: https://github.com/dotnet/roslyn-analyzers/pull/4026/files?diff=unified&w=1**

See https://github.com/dotnet/roslyn/issues/46859 and https://github.com/dotnet/roslyn-analyzers/pull/3917#issuecomment-674477914 for details. We might end up with corrup pooled data structures in presence of cancellation, so work around the Roslyn bug by not returning pooled object back to the pool in presence of cancellation.

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
-using System.Threading;
 using Analyzer.Utilities.Extensions;
 using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.CopyAnalysis;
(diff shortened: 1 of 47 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Threading;
```
mavasani, line 7:
```suggestion
```

dotpaul:
Could this be done in 2.9.x?

mavasani:
I am planning to actually fix the analyzer driver bug and close this PR out. Even after these changes, we see similar asserts cropping up.

dotpaul:
Ah okay. I guess this issue is mostly about the IDE, right? So if people run into this, the solution will be to upgrade their Visual Studio?

mavasani:
Yes, no build impact. Only affects IDE typing scenario.

mavasani:
Actually, turns out this is not an analyzer driver bug. Analyzer is expected to not throw cancellation exception or free resources on potentially cancellable code as driver will make the same callback next time. I am going to close this PR and open one targeting 2.9.x

PR #4035, opened 2020-08-18
Do not report CA1307 for known non-string comparisons

### LINKED ISSUE #2581: False positive in CA1307 for string.IndexOf(char)
#### Analyzer package

#### Package Version

v2.9.3

#### Diagnostic ID

[CA1307](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1307-specify-stringcomparison): Specify StringComparison

#### Repro steps

Compile the following code against .NET Core 2.2:

    class C
    {
        int M(string s) => s.IndexOf('?');
    }

#### Expected behavior

No warning.

#### Actual behavior

Warning: The behavior of 'string.IndexOf(char)' could vary based on the current user's locale settings. Replace this call in 'C.M(string)' with a call to 'string.IndexOf(char, System.StringComparison)'.

The premisse of the diagnostic, that string.indexOf(char) could vary is incorrect. The [method documentation](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=netcore-2.2#System_String_IndexOf_System_Char_) clearly states:

> This method performs an ordinal (culture-insensitive) search, where a character is considered equivalent to another character only if their Unicode scalar values are the same.

Note also that replacing the `s.IndexOf('?')` by `s.IndexOf('?', StringComparison.Ordinal)` is undesirable, as the method is less efficient.

mavasani:
Fixes #2581
When we know that a comparison is between a string type receiver and a non-string type parameter OR that the method is not a string comparison method, do not report CA1307.

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparisonTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparisonTests.cs
@@ -2,6 +2,7 @@
 
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Testing;
+using Test.Utilities;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
     Microsoft.NetCore.Analyzers.Runtime.SpecifyStringComparisonAnalyzer,
@@ -55,7 +56,7 @@ GetCA1307CSharpResultsAt(14, 18, $"string.Compare({StringArgType}, int, {StringA
         }
 
         [Fact]
-        public async Task CA1307_StringWithTests_CSharp()
+        public async Task CA1307_StringWithStringTests_CSharp()
         {
             await VerifyCS.VerifyAnalyzerAsync(@"
 using System;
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparisonTests.cs
```diff
@@ -79,8 +80,25 @@ public bool StringWith()
...
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+using System.Globalization;
+
+public class StringComparisonTests
+{
+    public bool StringWith(string strA, char chA, char chB)
+    {
+        var x = strA.EndsWith(chA);
```
dotpaul, line 94: Oh, the `EndsWith(char)` and `StartsWith(char)` overloads don't exist in .NET Framework 4.7.2.

mavasani: Ah thanks, do we know if we have a way to skip tests for a specific target framework?

dotpaul: #if NETCOREAPP 😄 ?

mavasani: Thanks!

mavasani:
Thanks!

PR #4036, opened 2020-08-18
[RC1] CA1835 fix - Automatically add using/imports System if not yet included

carlossanlop:
This is a backport of https://github.com/dotnet/roslyn-analyzers/pull/3931

If rule CA1835 finds a diagnostic and applies a fix, but the System namespace has not yet been added (which contains the `AsMemory()` method for byte arrays), then the compilation can fail.
With this change, the `System` namespace should be automatically added when the user applies the fix, if the namespace has not been added it yet.


jmarolf:
NOTE: dual-checkins are not required in this repo. Anything merged into RC1 will be merged into the master branch.

carlossanlop:
@jmarolf I already merged it to master. Do I need to do something different here?
Also, what is the process to get this approved? Who should approve it?

jmarolf:
@carlossanlop I am fine approving and merging this in. Tactics does not need to approve things until next week [according to the schedule](https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/6727/NET-5-Schedule).

> I already merged it to master. Do I need to do something different here?

You did everything correctly, I just wanted to point out that in the future we can just merge changes into the place we want a fix to _ship_ in and not worry about how those fixes flow to future releases as that is automatic.

PR #4037, opened 2020-08-18
Ensure that we handle cancellation in returning pooled objects to the pool

mavasani:
**Strongly recommended to review ignoring white space diffs: https://github.com/dotnet/roslyn-analyzers/pull/4037/files?diff=unified&w=1**

If an analyzer callback throws operation cancelled exception ,it is known to have cancelled its work without side effects. An action which throws OperationCancelledException after cancellation is requested is considered to have not been invoked, i.e. if work resumes for analyzing the same compilation, then another call needs to be made for this callback. Analyzers are responsible for ensuring cancelled callbacks do not have side effects. Once a side effect has been applied by the analyzer callback, the callback should ensure that the remainder of the operation is not cancelled, otherwise state corruption could occur.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheck.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheck.cs
@@ -235,8 +235,8 @@ namespace Microsoft.NetCore.Analyzers.Security
                             }
                             finally
                             {
-                                rootOperationsNeedingAnalysis.Free();
-                                allResults?.Free();
+                                rootOperationsNeedingAnalysis.Free(compilationAnalysisContext.CancellationToken);
+                                allResults?.Free(compilationAnalysisContext.CancellationToken);
                             }
                         });
                 });
--- a/src/Utilities/Compiler/PooledObjects/PooledDictionary.cs
+++ b/src/Utilities/Compiler/PooledObjects/PooledDictionary.cs
@@ -5,6 +5,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Threading;
 
 #pragma warning disable CA1000 // Do not declare static members on generic types
 
@@ -22,7 +23,7 @@ namespace Analyzer.Utilities.PooledObjects
             _pool = pool;
         }
 
-        public void Dispose() => Free();
+        public void Dispose() => Free(CancellationToken.None);
 
         public ImmutableDictionary<K, V> ToImmutableDictionaryAndFree()
         {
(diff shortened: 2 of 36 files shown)


src/Utilities/Compiler/PooledObjects/PooledDictionary.cs
```diff
@@ -37,14 +38,21 @@ private PooledDictionary(ObjectPool<PooledDictionary<K, V>>? pool, IEqualityComp
...
             }
 
-            _pool?.Free(this);
+            _pool?.Free(this, CancellationToken.None);
             return result;
         }
 
-        public void Free()
+        public void Free(CancellationToken cancellationToken)
         {
```
dotpaul, line 46: We should probably mention that people should use this overload if the PooledDictionary is allocated in one analyzer action but freed in a different analyzer action.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheck.cs
```diff
@@ -235,8 +235,8 @@ public override void Initialize(AnalysisContext context)
                             }
                             finally
                             {
-                                rootOperationsNeedingAnalysis.Free();
-                                allResults?.Free();
+                                rootOperationsNeedingAnalysis.Free(compilationAnalysisContext.CancellationToken);
+                                allResults?.Free(compilationAnalysisContext.CancellationToken);
```
sharwell, line 239: 📝 This file is potentially problematic.

* The last time we check the cancellation token is in the call to `PropertySetAnalysis.BatchGetOrComputeHazardousUsages`
* If cancellation occurs after the last check, we will proceed to call `ReportDiagnostic` and finish without throwing an exception, but won't return the objects to the pool

Technically this isn't going to produce corrupt results, but it does mean the pattern doesn't account for cases where we finish successfully.

sharwell, line 239: `allResults` is local to this method, so it can always be returned when the method completes.

```suggestion
                                allResults?.Free();
```

mavasani: Yeah, I am not too concerned about cancellation firing after analysis but before diagnostic reporting. Bigger concern here is to avoid corruption of data structures.

mavasani:
@sharwell I would like to merge this PR to get us to a better state. We can do any further improvements in a follow-up PR.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotDisableHttpClientCRLCheck.cs
mavasani: I kept it consistent with the other free call in the finally, just feels more cleaner.

mavasani: Yes, something that can be handled in a follow-up PR. I don't see any user visible impact for this as IDE and compiler will de-dupe identical diagnostics.

src/Utilities/Compiler/PooledObjects/PooledDictionary.cs
mavasani: I'll do this as a follow-up. I want to try get this change into master so it does not miss the RC2 snap.

PR #4043, opened 2020-08-19
Separate out CA1307 into two rules

### LINKED ISSUE #2581: False positive in CA1307 for string.IndexOf(char)

mavasani:
Fixes #2581, again!

1. `CA1310`: Correctness rule that only flags string compare methods that are known to use culture specific string comparison by default
2. `CA1307`: Maintainability/Readability rule that flags the remainder methods that have an overload with an additional StringComparison parameter at the end. This rule does not care about the default string comparison used by the API.

### DIFF
--- a/src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md
+++ b/src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md
@@ -61,6 +61,7 @@ CA1306 | Globalization | Disabled | SetLocaleForDataTypesAnalyzer, [Documentatio
 CA1307 | Globalization | Hidden | SpecifyStringComparisonAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1307)
 CA1308 | Globalization | Disabled | NormalizeStringsToUppercaseAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1308)
 CA1309 | Globalization | Hidden | UseOrdinalStringComparisonAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1309)
+CA1310 | Globalization | Hidden | SpecifyStringComparisonAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1310)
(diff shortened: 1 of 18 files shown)


mavasani:
@GrabYourPitchforks @KrisVandermotten

mavasani:
Thanks @dotpaul

mavasani:
Filed https://github.com/MicrosoftDocs/visualstudio-docs/issues/5754 for adding docs.

GrabYourPitchforks:
Doesn't this mean that CA1307 will now always fire on `string.GetHashCode()`, `string.IndexOf(char)`, and similar? I fear this will end up introducing significant noise into the system.

mavasani:
Yes, if you only care about correctness, you should disable CA1307 and just enable CA1310. CA1307 would most likely be used like a refactoring/hidden/suggestion available in IDE for users where they want to opportunistically keep moving to API overloads with StringComparison parameter.

KrisVandermotten:
Looks good, thanks!

PR #4047, opened 2020-08-20
Swap rule IDs CA1307 and CA1310 and change default enable/severities

mavasani:


mavasani:
@KrisVandermotten

KrisVandermotten:
If I understand correctly, what used to be the behavior of CA1307 in previous releases is now CA1310, and CA1307 has new behavior. Is that correct?

What what the rationale behind this swap?

Would it not make more sense the keep the old behavior under the old number, i.e. CA1307, and give the new number to the new behavior?

What about teams that have this line in their .editorconfig?

    dotnet_diagnostic.CA1307.severity = none

The would not get the new behavior, which obviously was not their intent when they added that line a long time ago.

Or did I miss something?

mavasani:
@KrisVandermotten It was suggested that it was better to have the noisy/readability one as a new rule instead of one which has already been used. I am fine if we feel this swap is not needed - I can then undo most of this change and instead just disable CA1307 by default. @dotpaul any concerns with that?

KrisVandermotten:
> I am fine if we feel this swap is not needed - I can then undo most of this change and instead just disable CA1307 by default.

That would be fine with me. Basically, it means that CA1307 remains more or less unchanged, compared to previous releases, except for a change in the wording of its diagnostic message, and that it would be disabled by default. Any override of the previous severity level would remain in place.

On the other hand, there would be a new rule with new behavior, and it would be enabled by default. Sounds good to me.

dotpaul:
> @KrisVandermotten It was suggested that it was better to have the noisy/readability one as a new rule instead of one which has already been used. I am fine if we feel this swap is not needed - I can then undo most of this change and instead just disable CA1307 by default. @dotpaul any concerns with that?

No concerns from me, especially if people find it more intuitive to have CA1307 being the "noisy" one and disabled by default, and the new rule being enabled by default.

mavasani:
Sounds good. I'll close this PR and open a new one.

PR #4057, opened 2020-08-22
Fix IndexOutOfRangeException

### LINKED ISSUE #4052: warning AD0001: Analyzer 'Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.IdentifiersShouldNotContainTypeNames' threw an exception of type 'System.IndexOutOfRangeException'
Thanks @mavasani 

See these warnings as well 😄 

```
2>C:\Users\david\source\repos\WebApplication2\APIPlayground\Program.cs(27,1,27,6): warning CA1812: <Program>$ is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it static (Shared in Visual Basic).
2>CSC : warning AD0001: Analyzer 'Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.IdentifiersShouldNotContainTypeNames' threw an exception of type 'System.IndexOutOfRangeException' with message 'Index was outside the bounds of the array.'.
```

_Originally posted by @davidfowl in https://github.com/dotnet/aspnetcore/issues/25077#issuecomment-678035542_

Youssef1313:
Fixes #4052
Not very sure this is the correct fix. I just prevented accessing `Locations[0]` if `Locations` doesn't have any items.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainTypeNames.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainTypeNames.cs
@@ -127,7 +127,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             }
 
             var identifier = symbol.Name;
-            if (s_typeNames.Contains(identifier))
+            if (s_typeNames.Contains(identifier) && symbol.Locations.Length > 0)
             {
                 Diagnostic diagnostic = Diagnostic.Create(Rule, symbol.Locations[0], identifier);
                 context.ReportDiagnostic(diagnostic);


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldNotContainTypeNames.cs
```diff
@@ -127,11 +127,11 @@ private static void AnalyzeSymbol(ISymbol symbol, SymbolAnalysisContext context)
             }
 
             var identifier = symbol.Name;
-            if (s_typeNames.Contains(identifier))
+            if (s_typeNames.Contains(identifier) && symbol.Locations.Length > 0)
```
mavasani, line 130: The exception occurs even before this is in the call `MatchesConfiguredVisibility` above. I have the fix with a unit test added in first commit of https://github.com/dotnet/roslyn-analyzers/pull/4066.

PR #4069, opened 2020-08-26
Fix assert in presence of explicit cast of flow capture

mavasani:


mavasani:
Tagging @JarLob

JarLob:
Still asserts, although in the other place:
```cs
        private object foo(string str)
        {
            int? intVal = int.TryParse(str, out var outInt) ? outInt : (int?)null;
            return (intVal.HasValue ? (object)intVal.Value : (object)str);
        }
```

mavasani:
Thanks @JarLob, should be fixed now.

@dotpaul can you please review? Very trivial change...

dotpaul (approved):
:ship:

mavasani:
Thank you @dotpaul!

Issue #4072, opened 2020-08-27
[Platform Compat Analyzer] Attributes are not respected on event add/remove methods


jeffhandley:
### Describe the bug

Diagnostics are not raised when code registers or unregisters event handlers and the `add` or `remove` methods have `[SupportedOSPlatform]` or `[UnsupportedOSPlatform]` attributes applied.  When the attribute is applied to the `event` itself, diagnostics are raised as expected.

### Steps To Reproduce

```csharp
        [SupportedOSPlatform("windows")] // References to this event DO report diagnostics as expected
        public static event EventHandler WindowsOnlyEvent
        {
            add { }
            remove { }
        }

        public static event EventHandler WindowsOnlyEventAdd
        {
            [SupportedOSPlatform("windows")] // References to register an event handler DO NOT report diagnostics as expected
            add { }
            remove { }
        }

        public static event EventHandler WindowsOnlyEventRemove
        {
            add { }

            [SupportedOSPlatform("windows")] // References to unregister an event handler DO NOT report diagnostics as expected
            remove { }
        }
```
### Expected behavior

Annotations on event `add` / `remove` methods should be respected and should result in diagnostics when referenced in code that is reachable in unsupported platforms.

### Actual behavior

No diagnostic is raised.

### Analyzer

**Platform Compatibility Analyzer** (local RC1 build)

**Diagnostic ID**: [CA1416](https://docs.microsoft.com/visualstudio/code-quality/ca1416)

## Additional context

We do not have any occurrences of this pattern for the APIs we've marked with attributes, and I'm not certain this was expected to work.  We should consult with @terrajobst to determine if this should be fixed.

terrajobst:
This isn't as bad as #4071 (I'm not aware of a single event we attributed) but it's good hygiene. We'd expect this to "just work".

buyaa-n:
> On second thought, while we haven't attributed events, we have attributed types, some of which have events. I'm now wondering whether we correctly warn for SomeType.SomEvent += somethod.

Yes we are, it was not working only for attributes applied to the `add/remove` accessors

Issue #4071, opened 2020-08-27
[Platform Compat Analyzer] Attributes are not respected on property getters/setters


jeffhandley:
### Describe the bug

Calling property getters or setters that are annotated with `[SupportedOSPlatform]` or `[UnsupportedOSPlatform]` does not raise diagnostics. When the attribute is applied to the property itself, diagnostics are raised as expected.

### Steps To Reproduce

```csharp
        [SupportedOSPlatform("windows")]
        public static bool WindowsOnlyProperty { get; set; } // References to this property DO report diagnostics as expected

        public static bool WindowsOnlyPropertyGetter
        {
            [SupportedOSPlatform("windows")] // References to this property getter DO NOT report diagnostics as expected
            get { return true; }
            set { }
        }

        public static bool WindowsOnlyPropertySetter
        {
            get { return true; }

            [SupportedOSPlatform("windows")] // References to this property setter DO NOT report diagnostics as expected
            set { }
        }
```
### Expected behavior

Annotations on property getters and setters should be respected and should result in diagnostics when referenced in code that is reachable in unsupported platforms.

### Actual behavior

No diagnostic is raised when referencing property getters or setters that are reachable with unsupported platforms.

### Analyzer

**Platform Compatibility Analyzer** (local RC1 build)

**Diagnostic ID**: [CA1416](https://docs.microsoft.com/visualstudio/code-quality/ca1416)

## Additional context

Some properties on [`System.Console`](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Console/src/System/Console.cs) have been annotated using this approach of applying the attribute to property setters, while leaving the property getters supported on all platforms.  This bug was discovered when trying to access the `Console.WindowHeight` property setter and the `Console.CursorVisible` property getter with code reachable in platforms other than Windows.

We should attempt to fix this in RC1 with the introduction of the analyzer.

/cc @jmarolf

terrajobst:
Oh... I haven't seen this! Yeah, this needs to be fixed for RTM (unlike #4072). We applied a ton of attributes to just getters/setters.

PR #4074, opened 2020-08-27
Support for Property getter/setter and event add/remove annotations

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4071 **Attributes are not respected on property getters/setters**
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4072 **Attributes are not respected on event add/remove methods**

Copied [IOperation Extensions method](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/OperationExtensions.cs#L25) from roslyn and used it for determining if Read or Write operation taken place

CC @jmarolf @mavasani @jeffhandley @terrajobst


src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -801,6 +801,331 @@ public static bool IsSetMethodInvocation(this IPropertyReferenceOperation operat
 
             return parameterOrderedArguments.ToImmutableArray();
         }
+
+        // Copied from roslyn https://github.com/dotnet/roslyn/blob/master/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/OperationExtensions.cs#L25
```
buyaa-n, line 805: Copied the whole implementation as it can be useful for existing and future analyzers

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -801,6 +801,331 @@ public static bool IsSetMethodInvocation(this IPropertyReferenceOperation operat
...
+
+#if CODEANALYSIS_V3_OR_BETTER
```
buyaa-n, line 807: Some operations used here newer or so not recognised for **BannedAPiAnalyzers**:
```
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(854,26): error CS0246: The type or namespace name 'IRecursivePatternOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\Microsoft.CodeAnalysis.BannedApiAnalyzers\Core\Microsoft.CodeAnalysis.BannedApiAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(865,26): error CS0246: The type or namespace name 'ISwitchExpressionArmOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\Microsoft.CodeAnalysis.BannedApiAnalyzers\Core\Microsoft.CodeAnalysis.BannedApiAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(883,26): error CS0246: The type or namespace name 'IPropertySubpatternOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\Microsoft.CodeAnalysis.BannedApiAnalyzers\Core\Microsoft.CodeAnalysis.BannedApiAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(962,42): error CS0246: The type or namespace name 'IReDimClauseOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\Microsoft.CodeAnalysis.BannedApiAnalyzers\Core\Microsoft.CodeAnalysis.BannedApiAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(965,56): error CS0246: The type or namespace name 'IReDimOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\Microsoft.CodeAnalysis.BannedApiAnalyzers\Core\Microsoft.CodeAnalysis.BannedApiAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(1062,22): error CS0246: The type or namespace name 'ICoalesceAssignmentOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\Microsoft.CodeAnalysis.BannedApiAnalyzers\Core\Microsoft.CodeAnalysis.BannedApiAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(854,26): error CS0246: The type or namespace name 'IRecursivePatternOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\PerformanceSensitiveAnalyzers\Core\Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(865,26): error CS0246: The type or namespace name 'ISwitchExpressionArmOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\PerformanceSensitiveAnalyzers\Core\Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(883,26): error CS0246: The type or namespace name 'IPropertySubpatternOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\PerformanceSensitiveAnalyzers\Core\Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(962,42): error CS0246: The type or namespace name 'IReDimClauseOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\PerformanceSensitiveAnalyzers\Core\Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(965,56): error CS0246: The type or namespace name 'IReDimOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\PerformanceSensitiveAnalyzers\Core\Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.csproj]
D:\dotnet\roslyn-analyzers\src\Utilities\Compiler\Extensions\IOperationExtensions.cs(1062,22): error CS0246: The type or namespace name 'ICoalesceAssignmentOperation' could not be found (are you missing a using directive or an assembly reference?) [D:\dotnet\roslyn-analyzers\src\PerformanceSensitiveAnalyzers\Core\Microsoft.CodeAnalysis.PerformanceSensitiveAnalyzers.csproj]
```

jmarolf:
I would wait for @mavasani to review this before merging

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -801,6 +801,331 @@ public static bool IsSetMethodInvocation(this IPropertyReferenceOperation operat
...
+
+                default:
+                    return false;
+            }
+        }
+#endif
+    }
+
+    [Flags]
+    internal enum ValueUsageInfos
```
mavasani, line 1074: Nit: Can we have the name match the Roslyn name, i.e. `ValueUsageInfo`? Also please move it to a separate file outside Extensions folder.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
```diff
@@ -180,6 +180,46 @@ public int M3(int option)
             await VerifyAnalyzerAsyncCs(source);
         }
 
+        [Fact]
```
mavasani, line 183: Nit: add WorkItem attribute?

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
buyaa-n: There is this warning `Flags enums should have plural names`, i will suppress it then

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -808,6 +818,12 @@ private static PlatformAttributes CopyAllAttributes(PlatformAttributes copyTo, P
 
                 AddPlatformAttributes(symbol.GetAttributes(), ref attributes);
 
+                if (symbol is IMethodSymbol method && (method.MethodKind == MethodKind.PropertyGet || method.MethodKind == MethodKind.PropertySet))
```
mavasani, line 821: What is the containing symbol for this get/set method? I thought it would have been the property symbol, but seems like it was the containing type?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
buyaa-n: sure

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
mavasani, line 821:
```suggestion
                if (symbol.IsAccessorMethod())
```

mavasani: This should also handle the event case...

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -490,10 +490,20 @@ private static void ReportDiagnostics(IOperation operation, SmallDictionary<stri
                 IInvocationOperation iOperation => iOperation.TargetMethod,
                 IObjectCreationOperation cOperation => cOperation.Constructor,
                 IFieldReferenceOperation fOperation => IsWithinConditionalOperation(fOperation) ? null : fOperation.Field,
+                IPropertyReferenceOperation pOperation => GetPropertyAccessor(pOperation.Property, operation),
```
mavasani, line 493: Will you be handling `IEventReferenceOperation` in a separate PR? It should be similar change to this one I believe..

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: > This should also handle the event case...

Yeah i am just started looking that bug, looks very similar

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
mavasani: Great, thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: Yeah, I see it is similar, my plan is: raise PR for property first as it is more important and if time allows add Event accessors into it while it is under review

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
```diff
@@ -180,6 +180,46 @@ public int M3(int option)
...
+    public static bool WindowsOnlyPropertySetter
+    {
+        get { return true; }
+        [SupportedOSPlatform(""windows"")]
+        set { }
+    }
+
+    public void M1()
+    {
+        WindowsOnlyPropertyGetter = true;
```
mavasani, line 207: Can you please add a test where the reference is a read + write, i.e. say `WindowsOnlyPropertyGetter += value;` and `WindowsOnlyPropertySetter += value;`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: yeah it was not, containing symbol was the containing class

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
buyaa-n: Good point

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -490,10 +490,20 @@ private static void ReportDiagnostics(IOperation operation, SmallDictionary<stri
...
                 IMemberReferenceOperation mOperation => mOperation.Member,
                 _ => null,
             };
 
+        private static ISymbol GetPropertyAccessor(IPropertySymbol property, IOperation operation)
+            => operation.GetValueUsageInfo(property.ContainingSymbol) switch
```
mavasani, line 499: I believe the `ValueUsageInfo.ReadWrite` should be handled to return both the getter and setter, which would mean `ISymbol? GetOperationSymbol` should be changed to `IEnumerable<ISymbol>? GetOperationSymbols`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
mavasani: Can you please also add a test where you pass property as ref, in, out argument to a method call?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -490,10 +490,20 @@ private static void ReportDiagnostics(IOperation operation, SmallDictionary<stri
...
+            {
+                ValueUsageInfos.Read => property.GetMethod,
```
mavasani, line 501: Using a switch statement to compare against a single field may not work correctly when the return flags value has multiple bits set, for example `ValueUsageInfo.ReadableReference` (`in` param), `ValueUsageInfo.WritableReference` (`out` param) and `ValueUsageInfo.ReadableWritableReference`  (`ref` param). Instead it would be better to write this method as:
```csharp
private static IEnumerable<IMethodSymbol> GetReferencedAccessorSymbols(IPropertySymbol property, IPropertyReferenceOperation operation)
{
    var valueUsageInfo = operation.GetValueUsageInfo(property.ContainingSymbol);
    if (valueUsageInfo & ValueUsageInfo.Read != 0 || valueUsageInfo & ValueUsageInfo.Name != 0)
    {
        yield return property.GetMethod;
    }

    if (valueUsageInfo & ValueUsageInfo.Write != 0)
    {
        yield return property.SetMethod;
    }
}
`

```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
mavasani: Seems like a false positive. I think `Info` means plural, though probably it is not a legitimate word so maybe its an acceptable false positive :-)? Suppressing should be fine.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
mavasani: Thanks, your change seems good.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
mavasani: Great, thanks.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
buyaa-n: > Can you please also add tests where you pass property as ref, in, out argument to a method call?

I tried, got compiler error CS0206: `A property or indexer may not be passed as an out or ref parameter, so i am not handling the case ValueUsageInfo.Reference at all

mavasani: Ah, thanks! Would you be able to add comment on the same lines in this method?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: Well, the calls site and the entire code path is not expecting multiple symbols. and i thought `ValueUsageInfo.ReadWrite` will be covered by separate Read/Write as `ReadWrite = Read | Write` but now i think it might not, will check

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
mavasani: I mean a comment in your analyzer where it processes value usage info.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
mavasani: Yeah, I think `WindowsOnlyPropertySetter += value;` will not be flagged with current implementation, when it should be.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
mavasani: You may also want to copy over https://github.com/dotnet/roslyn/blob/755c3b186fba57b676b5db41d7dce09b600f1e42/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/ValueUsageInfo.cs#L67-L80 to make the analyzer code cleaner :-)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: >  if (symbol is IMethodSymbol method && (method.MethodKind == MethodKind.PropertyGet || method.MethodKind == MethodKind.PropertySet)) => if (symbol.IsAccessorMethod())

`symbol.IsAccessorMethod()` is neat, but the result of `is` expression `method` is used within if block, so am gonna keep this as is

mavasani: You can add an extension method even on IMethodSymbol that gets called from ISymbolExtensions. It will also handle events for you.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -808,21 +850,28 @@ private static PlatformAttributes CopyAllAttributes(PlatformAttributes copyTo, P
...
 
             static bool AddPlatformAttributes(ImmutableArray<AttributeData> immediateAttributes, [NotNullWhen(true)] ref SmallDictionary<string, PlatformAttributes>? attributes)
             {
+                bool added = false;
                 foreach (AttributeData attribute in immediateAttributes)
                 {
                     if (s_osPlatformAttributes.Contains(attribute.AttributeClass.Name))
                     {
-                        TryAddValidAttribute(ref attributes, attribute);
+                        added |= TryAddValidAttribute(ref attributes, attribute);
```
mavasani, line 871: Does this skip calls to TryAddValidAttribute if added is already true? I always get confused with semantics of such compound or operations. Might be simpler to just write this as an if whose body sets added to true unconditionally.

mavasani: Btw, I did not see implementation for TryAddValidAttribute change in this PR. Do we need the changes to AddPlatformAttribues?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
```diff
@@ -180,6 +181,48 @@ public int M3(int option)
...
+        var s = [|WindowsOnlyPropertyGetter|];
+        [|WindowsOnlyPropertyGetter|] |= true;
+        [|WindowsOnlyPropertySetter|] &= false;
+        [|WindowsOnlyPropertySetter|] = false;
+        s = WindowsOnlyPropertySetter;
+        M2([|WindowsOnlyPropertyGetter|]);
+        var name = nameof([|WindowsOnlyPropertyGetter|]);
```
mavasani, line 215: This doesn’t really invoke the getter. Maybe we should use the property symbol itself for this case?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: The value returned from this function is not really used anywhere, I added this change to resolve the symbols returned in Enumerable, just `attributes != null` is not really express if any attribute added. But later choose another approach for Enumerable. But returning correct value will be helpful later when we refactor the logic to use immutable attributes,

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzerTests.cs
buyaa-n: Well the property `WindowsOnlyPropertyGetter` has attribute only on the getter, not on the property itself, so as it is warning here i would assume it is invoking getter, no?

buyaa-n: Actually you are right, it's warning here because i make it 😄, will update

jeffhandley:
What gets emitted in the diagnostic for the name of the API?  Ideally, we would match the behavior of the `Obsolete` attribute's precedent.  Here's what it does.  Does the helper method we're using to get the display name already handle this by chance?

**In C#**
image
> Test.WindowsOnlyPropertyGetter.get is obsolete

image
> Test.WindowsOnlyPropertySetter.set is obsolete

**In VB**
image
> 'Get' accessor of 'Public Shared Overloads Property WindowsOnlyGetter as Boolean' is obsolete

image
> 'Set' accessor of 'Public Shared Overloads Property WindowsOnlySetter as Boolean' is obsolete

buyaa-n:
> What gets emitted in the diagnostic for the name of the API? Ideally, we would match the behavior of the `Obsolete` attribute's precedent. Here's what it does. Does the helper method we're using to get the display name already handle this by chance?

Display name is handling only C# similarly (well we are using c# option `SymbolDisplayFormat.CSharpShortErrorMessageFormat`), i guess we need to split that logic into a separate version for VB

mavasani:
I actually prefer our current approach of using the similar display name for VB and C#, we strive very hard to not have different error messages between languages for same issue. I’d say it is in fact a compiler issue that messages are not similar for obsolete diagnostics between languages.

jeffhandley:
Ok, cool. So long as it's deliberate and the property accessor names are friendly like `Obsolete`'s messages. Thanks.

mavasani:
Note that it is fine to use VB error message format for VB and C# error message format for C#. You can tweak that part easily in your ReportDiagnostics method by switching on IOperation.LanguageName. Let’s still try to use the same resources string for C# and VB so the message appears similar to user, but just with correct VB and C# terminology.

buyaa-n:
JFYI: Messages looks in C# like:
`'Test.WindowsOnlyPropertyGetter.get' is supported on 'windows'`
`'Test.WindowsOnlyPropertySetter.set' is supported on 'windows'`

In VB:
`'Test.WindowsOnlyPropertyGetter()' is supported on 'windows'`
`'Test.WindowsOnlyPropertySetter(Boolean)' is supported on 'windows'`

mavasani:
@buyaa-n Above messages are when both languages use `ToDisplayString(CSharpErrorMessageFormat)`? Or after change for each language to use different error message format?

buyaa-n:
> @buyaa-n Above messages are when both languages use `ToDisplayString(CSharpErrorMessageFormat)`? Or after change for each language to use different error message format?

we don't have any path for VB, so they both using C# `SymbolDisplayFormat.CSharpShortErrorMessageFormat`

mavasani:
Thanks, can you try changing it to use C#/VB short error message format based on [IOperation.Language](http://sourceroslyn.io/#Microsoft.CodeAnalysis/Operations/IOperation.cs,52) and see the message?

buyaa-n:
> Thanks, can you try changing it to use C#/VB short error message format based on [IOperation.Language](http://sourceroslyn.io/#Microsoft.CodeAnalysis/Operations/IOperation.cs,52) and see the message?

Sure, about to do that

buyaa-n:
Now VB looks like
`'Public Shared Property Get WindowsOnlyPropertyGetter() As Boolean' is supported on 'windows'`
`'Public Shared Property Set WindowsOnlyPropertySetter(value As Boolean)' is supported on 'windows'`
Pretty close to the Obsolete's, thanks @mavasani

jeffhandley:
I recommend we go ahead and merge this in and then have @jmarolf proceed with getting the new build integrated into the SDK.

buyaa-n:
> I recommend we go ahead and merge this in and then have @jmarolf proceed with getting the new build integrated into the SDK.

The bug fix for Event accessor is coming, I am about to push i guess we want ti fixed

buyaa-n:
Event accessors added, @mavasani final round review, please 🙏

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -494,6 +526,43 @@ private static void ReportDiagnostics(IOperation operation, SmallDictionary<stri
                 _ => null,
             };
 
+        private static IEnumerable<ISymbol> GetPropertyAccessors(IPropertySymbol property, IOperation operation)
```
mavasani, line 529:
```suggestion
        private static IEnumerable<ISymbol> GetPropertyAccessors(IPropertySymbol property, IPropertyOperation operation)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -494,6 +526,43 @@ private static void ReportDiagnostics(IOperation operation, SmallDictionary<stri
...
+
+        private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operation)
+        {
+            if (operation.Parent is IEventAssignmentOperation eventAssignment)
+            {
+                if (eventAssignment.Adds)
+                    return iEvent.AddMethod;
+                else
+                    return iEvent.RemoveMethod;
+            }
```
mavasani, line 562: Do we need to handle `RaiseMethod` as well? http://sourceroslyn.io/#Microsoft.CodeAnalysis/Symbols/IEventSymbol.cs,49.

mavasani: Oh well it is VB only :-) http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Symbols/PublicModel/EventSymbol.cs,63. I am fine with not handling it here and just file a follow-up tracking issue.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -801,6 +801,348 @@ public static bool IsSetMethodInvocation(this IPropertyReferenceOperation operat
...
+                default:
+                    return false;
+            }
+        }
+#endif
+    }
+
+    [Flags]
+#pragma warning disable CA1714 // Flags enums should have plural names
+    internal enum ValueUsageInfo
```
mavasani, line 1075: Nit: consider moving `ValueUsageInfo` and `ValueUsageInfoExtensions` to a separate file.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: From the caller site it is coming as IOperation and used as is, this change will need unnecessary casting

jeffhandley:
I've done manual testing with this branch and property getters and setters as well as event add/remove methods are all working as expected. :100:

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: An issue created https://github.com/dotnet/roslyn-analyzers/issues/4089

PR #4080, opened 2020-08-28
use latest sdk

jmarolf:


global.json
```diff
@@ -1,13 +1,13 @@
 {
   "tools": {
-    "dotnet": "3.1.200",
+    "dotnet": "5.0.100-preview.8.20417.9",
     "vs": {
       "version": "16.6"
     },
     "xcopy-msbuild": "16.6.5-alpha1"
   },
   "sdk": {
```
Youssef1313, line 9: In a similar roslyn PR, https://github.com/dotnet/roslyn/pull/46038/files, `"allowPrerelease": true,` was added. Is it necessary here?

global.json
```diff
@@ -1,13 +1,13 @@
 {
   "tools": {
-    "dotnet": "3.1.200",
+    "dotnet": "5.0.100-preview.8.20417.9",
     "vs": {
       "version": "16.6"
```
Youssef1313, line 5: Not sure if this will fix the current build failure. If not, ignore the suggestion.


```suggestion
      "version": "16.8"
```

global.json
jmarolf: you are correct!

global.json
```diff
@@ -1,13 +1,14 @@
 {
   "tools": {
-    "dotnet": "3.1.200",
+    "dotnet": "5.0.100-preview.8.20417.9",
     "vs": {
-      "version": "16.6"
+      "version": "16.9"
```
mavasani, line 5: Sorry, I don't understand but what does VS 16.9 version mean, that is the next version?

jmarolf: stolen from [here](https://github.com/dotnet/roslyn/pull/47165#discussion_r478573904) I wanted to see if this passed the build first. I am not convinced this is actually what we want. Will need to confirm this doesn't break external contributors.

Youssef1313:
eng/Verions.props *may* need updates (Not very sure though).

global.json
```diff
@@ -1,13 +1,14 @@
 {
   "tools": {
-    "dotnet": "3.1.200",
+    "dotnet": "5.0.100-preview.8.20417.9",
```
riarenas, line 3: The build is failing due to missing a 3.1 runtime. I tested that 

```suggestion
    "dotnet": "5.0.100-preview.8.20417.9",
    "runtimes": {
      "dotnet": [
        "3.1.7"
      ]
    },
```
fixed this locally for me.

global.json
jmarolf: alright, talked to the arcade folks. It just needs to be newer that preview 6. I am going to update us to be on preview 7

PR #4081, opened 2020-08-28
[Proposal] Promote some rules which have code fixes to IDE suggestions/refactorings

mavasani:
Suggestions:
1. CA1012 (Abstract types should not have public constructors) It was disabled as during our triage we were not sure about the rule semantics - the message did not have the term 'public', while the implementation did, and now has been fixed.
2. CA2008 (Do not create tasks without passing a TaskScheduler): This rule was supposed to be deprecated in favor of VS threading analyzers, but since then we have decided that important threading analyzers will be merged into this package as CA rules. This rule is no longer deprecated.
3. CA1027 (Mark enums with FlagsAttribute) and CA2217 (Do not mark enums with FlagsAttribute) - We considered promoting the rules to warnings, but delayed that as it might be too aggressive. They should certainly be useful IDE suggestions with code fixes.
4. CA1032 (Implement standard exception constructors): This rule was disabled pending confirmation that its behavior matches .NET API recommendations. Confirmed it matches https://docs.microsoft.com/dotnet/api/system.exception?view=netcore-3.1#Custom
5. CA1064 (Exceptions should be public): Seems like a good suggestion. I can convert it to hidden diagnostic/refactoring if we have concerns.
6. CA1802 (Use literals where appropriate): Only concern was this may change public API surface, but that concern applies to all design rules. I have found this rule useful with no false positives on all repos.
7. CA1813 (Seal attribute types for improved performance): Seems like a good suggestion. Not sure why one would want to declare a non-abstract, non-sealed attribute type.
8. CA1820 (Test for empty strings using string length): This is a good suggestion with no false positives. Disabling the rule just due to the fact that test code may not care about it is not valid:
   1. We want to suggest good code quality even for test code.
   2. This applies to all performance rules, not just this one. If you want to write non-performant test code, you should explicitly disable 'category-performance' for all test projects.

Refactorings:
1. CA1052 (Static holder classes should be marked static): There is concern about false positives when type is used as a generic type argument, but seems like a good refactoring to offer on non-static types that only have static members.
2. CA1066 (Implement IEquatable when overriding Object.Equals): We felt this rule can have false positives. However, seems like a good refactoring to offer for struct types to implement IEquatable.
3. CA2007 (Do not directly await a Task in libraries. Append ConfigureAwait(false) to the task): No longer deprecated in favor of VS threading analyzers. It is a refactoring rather then a suggestion as the rule is not applicable to all project types.


mavasani:
Closing this PR as we have internally decided to take a conservative route in enabling more CA analyzers in the IDE for .NET5 and use the feedback to iterate on the defaults for .NET6.

PR #4084, opened 2020-08-28
Add a test case for verbatim string

Youssef1313:


Evangelink:
Have you faced any issue or are you simply trying to improve the coverage?

Youssef1313:
@Evangelink I was facing a problem, but then realized that this rule was actually disabled in my case 😄 
So there is actually no problem with verbatim string in the analyzer. Let me close as this won't really affect the coverage.

PR #4092, opened 2020-08-28
Report a warning when user has a package reference to an older versio…

### LINKED ISSUE #3977: Report a warning when user has a package reference to an older version of .NET SDK analyzers then built-in ones in the .NET SDK
Follow-up to https://github.com/dotnet/roslyn-analyzers/pull/3974. That PR ensures that users can override/de-couple built-in .NET analyzers with an explicit analyzer NuGet package reference. We should build on that to warn users when they move to an SDK with newer version of analyzers then the one they have installed with the explicit NuGet package reference.

> I’d say the SDK should have a static property that indicates which version of the package it has ingested. When the package is installed, it should compare its own version number with that. If it’s equal or higher, no warning. If its lower, issue a warning like this:
WARNING: The .NET SDK has newer analyzers then what is provided by version x.y.z of this package. Update or remove this package reference.

mavasani:
…n of .NET SDK analyzers then built-in ones in the .NET SDK

Fixes #3977

Verified the change locally by applying the targets and props file changes even in the SDK version of these files. There needs to be an additional change in the SDK to unconditionally import the SDK props, instead of conditioning it on EnableNETAnalyzers. This is required to ensure the `_NETAnalyzersSDKAssemblyVersion` is correctly set in the SDK to compare against the version in the NuGet package. I will submit a separate PR to the SDK repo for this change.

image

### DIFF
--- a/src/Tools/GenerateDocumentationAndConfigFiles/CommonPropertyNames.cs
+++ b/src/Tools/GenerateDocumentationAndConfigFiles/CommonPropertyNames.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace GenerateDocumentationAndConfigFiles
+{
+    public static class CommonPropertyNames
+    {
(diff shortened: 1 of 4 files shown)


mavasani:
Created https://github.com/dotnet/sdk/pull/13224 for the SDK change to get this to work.

PR #4104, opened 2020-08-31
Add support for AnalysisMode

### LINKED ISSUE #4100: Question: what's the suggested usage between NetAnalyzers and FxCopAnalyzers?
After reading the [post](https://devblogs.microsoft.com/dotnet/automatically-find-latent-bugs-in-your-code-with-net-5/) in devblog, I've got some questions:

- Is it supported to use the two analyzer packages at the same time?
image
After upgrading to .NET 5 Preview 8, there is a warning icon in IDE for FxCopAnalyzers package. I'm afraid there's some conflict between the two packages.

- What's the suggested way to **aggressively opt-in** more **up-to-date analysis**?
FxCopAnalyzers and NetAnalyzers carries many analyzers, and some of them are old and not suitable for .NET 5. The default severity of FxCopAnalyzers looks too aggressive, where NetAnalyzers looks too loose. It's unclear that which rule set is suggested.
I mentioned https://github.com/dotnet/roslyn-analyzers/issues/4002#issuecomment-673561776 . But there are also other analyzers defaults to a different severity level like CA1305.

mavasani:
Closes #4100

**Proposal**
This new property would allow users to switch between following 3 analysis modes:

1. **Default mode**: Analyzers with enabled by default and severity values that we ship today. It will continue to be the default analysis mode.
2. **OptIn or AllDisabled or Conservative mode**: All analyzers are disabled by default. Only the ones that user explicitly enables in their editorconfig/ruleset will be enabled with the explicitly configured severity.
3. **OptOut or AllEnabled or Aggressive mode**: Majority of analyzers get enabled as warnings, similar to the current FxCop analyzers package. We still want some security rules and no longer recommended rules to be disabled, but for all practical purposes, we can consider this to be the most aggressive analysis mode for the package.

The property should work amicably with AnalysisLevel, i.e. if user sets AnalysisLevel to 5.0, the aggressive, all-enabled mode only turns on rules shipped in or before 5.0 to warnings. Rules shipping after 5.0 will still be turned off by default.

**Targeted scenarios**

1. **User requests**: We already have a few user requests to enable aggressive analysis mode, similar to FxCop analyzers:
   1. https://github.com/dotnet/roslyn-analyzers/issues/4100: This was requested from a customer from Jon’s blog.
   2. https://github.com/dotnet/roslyn/issues/47046: This was requested from a customer trying out our bulk configuration editorconfig functionality to switch NetAnalyzers to be as aggressive as FxCopAnalyzers
This feature enables users who are super passionate about analyzers and code quality to easily switch to an opt-out, aggressive, analysis mode, which seems fantastic to me! At the same time, it helps users who like complete opt-in model with explicit control over each rule that is enabled in their configuration file.

2. **Path to deprecation for FxCopAnalyzers package**: Currently, we ship identical CA rules in FxCopAnalyzers and NetAnalyzers package/SDK. The only difference is default severity and enabled by default values for rules. Adding this AnalysisMode knob means existing FxCop analyzer customers can easily migrate to the .NET SDK based analyzers and get equivalent functionality and this allows us to deprecate this package soon. This will make the overall code quality analyzers story much cleaner from an end user perspective.


src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -141,13 +141,19 @@ public static int Main(string[] args)
...
+                    foreach (var analysisMode in Enum.GetValues(typeof(AnalysisMode)))
+                    {
+                        CreateEditorconfig(
                         outputDir,
-                        $"AnalysisLevel_{GetNormalizedVersionStringForEditorconfigFileNameSuffix(version)}.editorconfig",
-                        $"Rules from '{version}' release",
-                        $"Rules with default severity and enabled state from '{version}' release. Rules that are first released in a version later then '{version}' are disabled.",
+                        $"AnalysisLevel_{analysisLevelVersionString}_{analysisMode}.editorconfig",
+                        $"Rules from '{version}' release with '{analysisMode}' analysis mode",
+                        $"Rules with enabled by default state from '{version}' release with '{analysisMode}' analysis mode. Rules that are first released in a version later then '{version}' are disabled.",
```
gewarren, line 152:
```suggestion
                        $"Rules with enabled-by-default state from '{version}' release with '{analysisMode}' analysis mode. Rules that are first released in a version later than '{version}' are disabled.",
```

gewarren (approved):

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -521,5 +554,12 @@ public Assembly LoadFromPath(string fullPath)
                 return Assembly.LoadFrom(fullPath);
             }
         }
+
+        private enum AnalysisMode
+        {
+            Default,
+            AllDisabled,
+            AllEnabled,
```
mavasani, line 562: @gewarren @mikadumont Any feedback on the customer facing values for `AnalysisMode` property? Other options that come to my mind are:

`Default, AllDisabledByDefault, AddEnabledByDefault`

OR

`Default, OptOut, OptIn`

OR

`Default, Conservative, Aggressive`

any other suggestions?

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -521,5 +554,12 @@ public Assembly LoadFromPath(string fullPath)
                 return Assembly.LoadFrom(fullPath);
             }
         }
+
+        private enum AnalysisMode
```
mavasani, line 558:
```suggestion
        // NOTE: **Do not** change the names of the fields for this enum - that would be a breaking change for user visible property setting for `AnalysisMode` property in MSBuild project file.
        private enum AnalysisMode
```

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -521,5 +554,13 @@ public Assembly LoadFromPath(string fullPath)
                 return Assembly.LoadFrom(fullPath);
             }
         }
+
+        // NOTE: **Do not** change the names of the fields for this enum - that would be a breaking change for user visible property setting for `AnalysisMode` property in MSBuild project file.
+        private enum AnalysisMode
+        {
+            Default,
+            AllDisabled,
+            AllEnabled,
```
mavasani, line 563: Changing the value to more correctly reflect the mode:
```suggestion
            AllDisabledByDefault,
            AllEnabledByDefault,
```

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -270,14 +279,27 @@ bool AddRule(DiagnosticDescriptor rule)
                         return true;
                     }
 
-                    (bool isEnabledByDefault, DiagnosticSeverity effectiveSeverity) GetEnabledByDefaultAndSeverity(DiagnosticDescriptor rule)
+                    (bool isEnabledByDefault, DiagnosticSeverity effectiveSeverity) GetEnabledByDefaultAndSeverity(DiagnosticDescriptor rule, AnalysisMode analysisMode)
                     {
+                        if (analysisMode == AnalysisMode.AllDisabled)
```
mavasani, line 284:
```suggestion
                        if (analysisMode == AnalysisMode.AllDisabledByDefault)
```

src/Tools/GenerateGlobalAnalyzerConfigs/Program.cs
```diff
@@ -270,14 +279,27 @@ bool AddRule(DiagnosticDescriptor rule)
...
+                        {
+                            return (isEnabledByDefault: false, DiagnosticSeverity.Warning);
+                        }
+
                         var isEnabledByDefault = rule.IsEnabledByDefault;
                         var effectiveSeverity = rule.DefaultSeverity;
 
+                        var isEnabledRuleInAggressiveMode = analysisMode == AnalysisMode.AllEnabled &&
```
mavasani, line 292:
```suggestion
                        var isEnabledRuleInAggressiveMode = analysisMode == AnalysisMode.AllEnabledByDefault &&
```

mavasani:
This was approved offline by the PM, dev and doc team.

PR #4107, opened 2020-08-31
Fix up GetControlFlowGraph extension method for property initializers

### LINKED ISSUE #4105: Platform compatibility analyzer does not work for property initializer syntax
```C#
class Program
{
    public static object Property { get; } = OperatingSystem.IsWindows() ? null : new Foo();
}

[UnsupportedOSPlatform("windows")]
class Foo
{
}
```

This fails with

```
Program.cs(9,87,9,96): warning CA1416: 'Foo' is unsupported on 'windows'
```

SDK version: 5.0.100-rc.1.20431.3

mavasani:
Fixes #4105
Verified added unit tests for NoDiagnostic cases fail before the fix.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/UseContainerLevelAccessPolicy.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/UseContainerLevelAccessPolicy.cs
@@ -137,8 +137,7 @@ namespace Microsoft.NetCore.Analyzers.Security
 
                                 if (argumentOperation != null)
                                 {
-                                    var cfg = invocationOperation.GetTopmostParentBlock()?.GetEnclosingControlFlowGraph();
-                                    if (cfg != null)
+                                    if (invocationOperation.TryGetEnclosingControlFlowGraph(out var cfg))
                                     {
                                         var interproceduralAnalysisConfig = InterproceduralAnalysisConfiguration.Create(
                                                                                 operationAnalysisContext.Options,
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
@@ -173,13 +173,13 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         return;
                     }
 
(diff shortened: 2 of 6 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/UseContainerLevelAccessPolicy.cs
```diff
@@ -137,8 +137,7 @@ public override void Initialize(AnalysisContext context)
 
                                 if (argumentOperation != null)
                                 {
-                                    var cfg = invocationOperation.GetTopmostParentBlock()?.GetEnclosingControlFlowGraph();
-                                    if (cfg != null)
+                                    if (invocationOperation.TryGetEnclosingControlFlowGraph(out var cfg))
```
mavasani, line 140: @dotpaul We may want to back port the product changes to 2.9.x branch. You probably want to add unit tests for the two security analyzers changed here as they would have not worked in presence of property/field initializers prior to this change. I'll let you make a call if this is a critical enough fix that you'd prefer to backport, especially given you are probably better suited to write corresponding regression tests for security analyzers.

dotpaul: I'll try to write a couple of tests tonight.

mavasani:
Thanks, I am going to merge this in master to unblock this bug.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/UseContainerLevelAccessPolicy.cs
dotpaul: #4109

dotpaul: @mavasani As long as it's straightforward, I'd prefer any DFA improvements to be backported to 2.9.x. If it's not straightforward, I guess it depends. 😃

jeffhandley:
@mavasani Is there a package version that this fix is already available in that @pranavkm could use to consume the fix in dotnet/aspnetcore#25421 instead of [keeping a workaround](https://github.com/dotnet/aspnetcore/pull/25421/files#r482026796) `#pragma warning disable` in place?

mavasani:
I think we actually want https://github.com/dotnet/roslyn-analyzers/pull/4124 to be merged for @pranavkm's request. I am going to merge that in soon and point out a package with the fix.

PR #4113, opened 2020-09-01
Handle analysis for escaped lambdas and local functions

### LINKED ISSUE #4090: Platform compat analyzer: flow analysis not working for the arguments of lambda expression
Flow analysis not working for the arguments of lambda expression
```csharp
class Test
{
    void M1()
    {
        M3(wo =>
        {
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 wo.M2(); // should not warn
        });           
    }
    public static void M3(Action<WindowsOnly> windowsAction) { }
}
```
Same for the result saved in local:
```
bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```

_Originally posted by @eerhardt in https://github.com/dotnet/runtime/pull/41209#discussion_r476445352_

mavasani:
Fixes #4090

Dataflow analysis now tracks lambda and local functions that are escaped via different means like passed as delegate argument, return value, assigned to ref/out parameter, assigned to field/property, etc. At the end of analysis, we invoke a standalone analysis for these escaped lambdas and local functions and this is available explicitly from the DataflowAnalysisResult for analyzers to consume.

This change fixes up the PlatformCompatibilityAnalyzer to consume these standalone results, which is sufficient to fix the critical issue #4090. I will have a follow-up PR so all flow based anayzers fallback to these standalone analysis results when querying for any operation inside a local function or a lambda.


src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -553,6 +581,40 @@ private void OnStartExitBlockAnalysis(BasicBlock exitBlock)
...
+        private void PerformStandaloneLambdaOrLocalFunctionAnalysisOnExit()
+        {
+            // First append the escaped local functions and lambdas from points to result, if any.
+            if (DataFlowAnalysisContext.PointsToAnalysisResult is { } pointsToAnalysisResult)
+            {
+                _escapedLocalFunctions.AddRange(pointsToAnalysisResult.EscapedLocalFunctions);
+                _escapedLambdas.AddRange(pointsToAnalysisResult.EscapedLambdas);
+            }
+
+            // Perform standalone analysis for local functions that escaped via conversion to delegates.
```
mavasani, line 593:
```suggestion
            // Perform standalone analysis for local functions that escaped.
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -553,6 +581,40 @@ private void OnStartExitBlockAnalysis(BasicBlock exitBlock)
...
+                foreach (var localFunction in DataFlowAnalysisContext.ControlFlowGraph.LocalFunctions)
+                {
+                    if (_escapedLocalFunctions.Contains(localFunction))
+                    {
+                        PerformStandaloneLocalFunctionInterproceduralAnalysis(localFunction);
+                    }
+                }
+            }
+
+            // Perform standalone analysis for lambdas that escaped via conversion to delegates.
```
mavasani, line 605:
```suggestion
            // Perform standalone analysis for lambdas that escaped.
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -46,13 +49,24 @@ public abstract class DataFlowOperationVisitor<TAnalysisData, TAnalysisContext,
...
+        private readonly HashSet<IFlowAnonymousFunctionOperation> _visitedLambdas;
         private readonly HashSet<IOperation>? _returnValueOperations;
         private ImmutableDictionary<IParameterSymbol, AnalysisEntity>? _lazyParameterEntities;
         private ImmutableHashSet<IMethodSymbol>? _lazyContractCheckMethods;
         private TAnalysisData? _currentAnalysisData;
         private BasicBlock? _currentBasicBlock;
         private int _recursionDepth;
 
+        /// <summary>
+        /// Local functions that escaped analysis scope via conversion to delegates passed to non-analyzed calls.
```
mavasani, line 61:
```suggestion
        /// Local functions that escaped from this method.
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -46,13 +49,24 @@ public abstract class DataFlowOperationVisitor<TAnalysisData, TAnalysisContext,
...
+        /// </summary>
+        private readonly HashSet<IMethodSymbol> _escapedLocalFunctions;
+
+        /// <summary>
+        /// Lambda methods that escaped analysis scope via conversion to delegates passed to non-analyzed calls.
```
mavasani, line 66:
```suggestion
        /// Lambda methods that escaped from this method.
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -2423,6 +2472,158 @@ IParameterSymbol GetMappedParameterForArgument(IArgumentOperation argumentOperat
...
+            IMethodSymbol invokedMethod,
+            PointsToAnalysisResult? pointsToAnalysisResult,
+            CopyAnalysisResult? copyAnalysisResult,
+            ValueContentAnalysisResult? valueContentAnalysisResult)
+        {
+            var invocationInstance = (AnalysisEntityFactory.ThisOrMeInstance, ThisOrMePointsToAbstractValue);
+            var thisOrMeInstance = invocationInstance;
+            var pointsToValues = pointsToAnalysisResult?[cfg.GetEntry()].Data;
+            var copyValues = copyAnalysisResult?[cfg.GetEntry()].Data;
+            var valueContentValues = valueContentAnalysisResult?[cfg.GetEntry()].Data;
```
mavasani, line 2608: Ah, all of this seems dead code, let me remove and also remove the corresponding analysis result parameters.
```suggestion
```

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -2935,77 +3136,109 @@ private TAbstractAnalysisValue VisitInvocation_NonLambdaOrDelegateOrLocalFunctio
...
+            PooledHashSet<IFlowAnonymousFunctionOperation> lambdaTargets)
+        => ResolveLambdaOrDelegateOrLocalFunctionTargetsCore(operation: null, invocationTarget, methodTargetsOptBuilder, lambdaTargets);
+
+        private bool ResolveLambdaOrDelegateOrLocalFunctionTargets(
+            IOperation operation,
+            PooledHashSet<(IMethodSymbol method, IOperation? instance)> methodTargetsOptBuilder,
+            PooledHashSet<IFlowAnonymousFunctionOperation> lambdaTargets)
+        => ResolveLambdaOrDelegateOrLocalFunctionTargetsCore(operation, invocationTarget: null, methodTargetsOptBuilder, lambdaTargets);
+
+        private bool ResolveLambdaOrDelegateOrLocalFunctionTargetsCore(
```
mavasani, line 3163: This is almost entirely existing code, just refactored so we can use it both for context sensitive interprocedural analysis and computing escaped lambda/local functions targets.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -220,26 +220,36 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
...
-                var results = analysisResult.TryGetInterproceduralResults();
-                if (results != null)
+                return false;
+            }
+
+            var results = analysisResult.TryGetInterproceduralResults();
+            if (!results.Any())
+            {
+                // Do not flag code inside unused lambda/local functions.
+                return true;
```
buyaa-n, line 235: Here we are suppressing warning for "unused lambda/local" methods, i am OK with this approach but @jeffhandley @terrajobst might have other opinions for this

mavasani: Thanks @buyaa-n. 
Note that we are not suppressing for unused lambda/local functions, we don't even analyze them. As soon as there is _some_ reference to the lambda/local function, we will analyze it. This is because we normally analyze these in context of the callers (for explicit invocations) or standalone (for escape cases where lambda/local function can no longer be tracked and may be invoked by other means outside the scope of the analysis). If this scenario is important, it would be easy to enhance the analysis in a follow-up PR, but I felt that is not an important scenario as users will likely not have unused lambda/local functions - compiler also generates a warning for it I believe.

mavasani: Users would also get diagnostics for unused lambda/local function:
image

terrajobst: Not analyzing unused local functions feels odd to me. Whether something is used or not usually doesn't remove the requirement for it to be correct. For example, the compiler still binds the body, requires definitive assignment rules etc. I don't see why analyzers would be any different from that.

mavasani: Hmm analogy with binding seems a good case that users would expect diagnostics while they code up a local function. Let me fix this up by tracking this information during flow analysis.

mavasani: @buyaa-n @terrajobst this should be fixed now. Buyaa, the latest commit https://github.com/dotnet/roslyn-analyzers/pull/4113/commits/b4bd3f4912373e2f06b23d2530384b3dd19c67be also fixes couple other analyzer bugs I found while adding tests, please review the analyzer changes. Thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -220,26 +220,34 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
...
+            Debug.Assert(results.Any(), "Expected at least one analysis result for lambda/local function");
+
+            foreach (var localResult in results)
+            {
+                Debug.Assert(localResult.ControlFlowGraph.OriginalOperation == containingLambdaOrLocalFunctionOperation);
+
+                var localValue = localResult[platformSpecificOperation.Kind, platformSpecificOperation.Syntax];
+
+                // Value must be known and guarded in all analysis contexts.
+                // NOTE: IsKnownValueGuarded mutates the input values, so we pass in cloned values
```
mavasani, line 241: @buyaa-n I think we should prioritize this cleanup item to remove side effects from `IsKnownValueGuarded` and make `PlatformAttributes` type immutable. This will very likely lead to more subtle bugs, like the one I fixed in the latest commit.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -498,9 +531,12 @@ void ProcessLocalOrParameter(ISymbol symbol)
...
 
-        public static bool IsWithinLambdaOrLocalFunction(this IOperation operation)
-            => operation.GetAncestor<IAnonymousFunctionOperation>(OperationKind.AnonymousFunction) != null ||
-               operation.GetAncestor<ILocalFunctionOperation>(OperationKind.LocalFunction) != null;
+        public static bool IsWithinLambdaOrLocalFunction(this IOperation operation, [NotNullWhen(true)] out IOperation? containingLambdaOrLocalFunctionOperation)
+        {
+            var kinds = ImmutableArray.Create(OperationKind.AnonymousFunction, OperationKind.LocalFunction);
+            containingLambdaOrLocalFunctionOperation = operation.GetAncestor(kinds);
+            return containingLambdaOrLocalFunctionOperation != null;
+        }
```
mavasani, line 539: Needed to handle nested lambda/local functions correctly.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -250,6 +250,39 @@ public static bool IsOperationNoneRoot(this IOperation operation)
             }
         }
 
+        /// <summary>
+        /// Gets the first ancestor of this operation with:
+        ///  1. Any OperationKind from the specified <paramref name="ancestorKinds"/>.
+        ///  2. If <paramref name="predicate"/> is non-null, it succeeds for the ancestor.
+        /// Returns null if there is no such ancestor.
+        /// </summary>
+        public static IOperation? GetAncestor(this IOperation root, ImmutableArray<OperationKind> ancestorKinds, Func<IOperation, bool>? predicate = null)
```
mavasani, line 259: This is mostly a clone of previous method, except our search looks for any kind from an array of potential ancestor kinds instead of just one.

mavasani:
@dotpaul @buyaa-n This should now be ready for review.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: Agree with making immutable entirely, but for the current context updating the attributes for the current context was intentional. Like the guard method might only suppress the part of attributes and we might still want to warn for the not suppressed attribute

mavasani: The issue with that is if there are 2 calls to a local function, where first one is guarded but second is not, then we do not warn. That is fixed now though.

buyaa-n: I see, as it is returning false when the value is not guarded (`!IsKnownValueGuarded`) i think it still can fail in case the order of the results of local functions are switched, no?

mavasani: I'll add a test, but I think given all calls to `IsKnownValueGuarded` now operate on a copy it shouldn't matter. Mutation does make this harder to reason about, so the best way to verify would be through a unit test.

buyaa-n: As it is only for looping through lambda/local function results I am good with working with a copy

mavasani: @buyaa-n added test to verify it works for either order now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: Thank you @mavasani that is great

buyaa-n (approved):
Changes for platform compat analyzer looks good, thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
buyaa-n: Thanks

dotpaul (approved):
DFA stuff looks good to me. :ship:

mavasani:
Thanks @buyaa-n @dotpaul

Issue #4002, opened 2020-08-13
New rule: Extra type checking / type casting


nschuessler:
### Describe the problem you are trying to solve

I have found that there is a lot of duplicate type checking and casting in code and wrote an analyzer and fix provider to address it. This is a common pattern:

```
object o = new StringBuilder();
if (o is StringBuilder)
{
    StringBuilder sb = o as StringBuilder;
}
```

This could be reduced in a better way to:

```csharp
object o = new StringBuilder();
StringBuilder sb = o as StringBuilder;
if (sb != null)
{
     // Use sb
}
```

This is closely related to excessive casting in that the `as` operator in the above could be replaced by a cast.
This is a contrived example but similar to code I've found:

```csharp
public void InefficientTypeChecking()
{
    object sb = new StringBuilder();
    object encoding = System.Text.Encoding.UTF8;

    int length = ((StringBuilder)sb).Length;
    StringBuilder tempBuilder = sb as StringBuilder;

    // Only one cast of this type should be ignored.
    Encoding tempEncoding = encoding as Encoding;

    if (Environment.TickCount > 0)
    {
        ((StringBuilder)sb).Append("Howdy");
    }
    else
    {
        ((StringBuilder)sb).Append("Hi");
    }

    this.StringBuildingFunction((StringBuilder)sb);
}
```

It could be better written using a temporary variable to save the result of the cast:

```csharp
public void InefficientTypeChecking()
{
    object sb = new StringBuilder();
    object encoding = System.Text.Encoding.UTF8;
    var var1 = (StringBuilder)sb;
    int length = var1.Length;
    StringBuilder tempBuilder = var1;

    // Only one cast of this type should be ignored.
    Encoding tempEncoding = encoding as Encoding;

    if (Environment.TickCount > 0)
    {
        var1.Append("Howdy");
    }
    else
    {
        var1.Append("Hi");
    }

    this.StringBuildingFunction(var1);
}
```

Although fixes to this rule are not likely to yield a big performance gain, this is a simple use case and implementation of flow-analysis combined with a pattern of coordinating and fixing duplicated code between an analyzer and a fix provider that can be used as a template for later analyzers that have bigger impact on performance.

[...]

PR #4114, opened 2020-09-02
Draft: ExtraTypeCheckingAnalyzer

nschuessler:
First rough cut on integrating new analyzer into the repository. Issue: [4002](https://github.com/dotnet/roslyn-analyzers/issues/4002)

Known issues:
- Tests not ported
- Diagnostic codes not selected yet.
- Have a few warnings to fix (around `DiagnosticDescriptorHelper.Create`)
- Did not update the public API files yet.
- I suspect unused / empty overrides of base classes will need to be removed.
- Didn't spend a lot of time sealing, and checking public/internal/privates
- I attempted to make reusable `AbstractValue` and `AbstractValueKind` classes ... is it even possible?
- There will be a discussion on how to support multiple diagnostic rendering configurations (Text/Markdown) ==> editor.config?
- There will probably be a discussion on how/if/should diagnostic.Properties can be customized.
- My implementation of the `ExtraTypeCheckingAnalysisData` is passing extra data out of the Visitor class .... and probably pushing the boundaries intended by the API.
- How does the help URL get customized (i.e. documentation about the diagnostics)
- There is a lot of null checking that I didn't have to have before -- just had asserts.
- The `<inheritdoc />` come from quieting `StyleCop.Analyzers.1118` which we use internally, expect that is not desired here.
Mention @mavasani  so he can see the review.

Q: Is the `WellKnowTypeProvider` available now outside the repository (so we don't have to use `GetMetadataTypeByName` on `Compilation` ?


nschuessler:
@mavasani if you could take a peek when you get a chance to give high level guidance.
Thx.

Issue #4117, opened 2020-09-02
AD0001 in PlatformCompatibilityAnalyzer


mavasani:
```
CSC(0,0): error AD0001: Analyzer 'Microsoft.NetCore.Analyzers.InteropServices.PlatformCompatabilityAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
```

@mavasani The build showing this error after updating. Is it a roslyn-analyzers bug given that it doesn't happen before updating?

_Originally posted by @Youssef1313 in https://github.com/dotnet/roslyn/pull/47375#issuecomment-685801130_

mavasani:
@buyaa-n I will investigate with changes from my current PR and see if that resolves this issue.

mavasani:
@buyaa-n call stack for NRE in the analyzer is below:

```
CSC : error AD0001: Analyzer 'Microsoft.NetCore.Analyzers.InteropServices.PlatformCompatabilityAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'. [c:\roslyn\src\Dependencies\PooledObjects\Microsoft.CodeAnalysis.PooledObjects.Package.csproj]
  Exception occurred with following context: (TaskId:47)
  Compilation: Microsoft.CodeAnalysis.PooledObjects.Package (TaskId:47)
  System.NullReferenceException: Object reference not set to an instance of an object. (TaskId:47)
  at Microsoft.NetCore.Analyzers.InteropServices.PlatformCompatabilityAnalyzer.<Initialize>g__GetOperatingSystemGuardMethods|19_1(IMethodSymbol runtimeIsOSPlatformMethod, INamedTypeSymbol operatingSystemType) (TaskId:47)
  at Microsoft.NetCore.Analyzers.InteropServices.PlatformCompatabilityAnalyzer.<Initialize>b__19_0(CompilationStartAnalysisContext context) (TaskId:47)
  at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info) (TaskId:47)
```

mavasani:
[...]

PR #4121, opened 2020-09-02
Fix NRE found in platform compat analyzer

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4117

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
@@ -98,10 +98,12 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 var typeName = WellKnownTypeNames.SystemOperatingSystem;
 
                 // TODO: remove 'typeName + "Helper"' after tests able to consume the real new APIs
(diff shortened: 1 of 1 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -97,10 +97,11 @@ public override void Initialize(AnalysisContext context)
             {
                 var typeName = WellKnownTypeNames.SystemOperatingSystem;
 
-                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(typeName + "Helper", out var operatingSystemType))
+                // TODO: remove 'typeName + "Helper"' after tests able to consume the real new APIs
+                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(typeName + "Helper", out var operatingSystemType) &&
+                    !context.Compilation.TryGetOrCreateTypeByMetadataName(typeName, out operatingSystemType))
```
mavasani, line 102: Is it possible to write a unit test? I think probably a test without including any OS mock helpers will hit?

buyaa-n: The type will exist even without Mock helpers or attributes, only if we can add test for the .Net version where this type does not exist, not sure what version that would be, might be netfx

mavasani (approved):

PR #4124, opened 2020-09-03
Enable value content tracking for OSPlatform.Create calls

### LINKED ISSUE #4119: Platform compat analyzer guard method nowt working when OSPlatform.Create(string) used
### Describe the bug
```csharp
class Program
{
    {
        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Create("windows")))
        {
            Foo(); // not supposed to warn
        }
    }

    [UnsupportedOSPlatform("windows")]
    public static void Foo() {}
}
```
### Expected behavior
Not warn
### Actual behavior
Warning

I think we want to warn it in RC2

mavasani:
Fixes #4119


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.OperationVisitor.cs
```diff
@@ -36,9 +37,19 @@ private sealed class OperationVisitor : GlobalFlowStateDataFlowOperationVisitor
 
                 if (_platformCheckMethods.Contains(method.OriginalDefinition))
                 {
-                    return PlatformMethodValue.TryDecode(method, visitedArguments, DataFlowAnalysisContext.ValueContentAnalysisResult, _osPlatformType, out var platformInfo) ?
-                        new GlobalFlowStateAnalysisValueSet(platformInfo) :
-                        GlobalFlowStateAnalysisValueSet.Unknown;
+                    using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                    if (PlatformMethodValue.TryDecode(method, visitedArguments, DataFlowAnalysisContext.ValueContentAnalysisResult, _osPlatformType, infosBuilder))
```
mavasani, line 41: TryDecode can return multiple platform infos now via dataflow. Unit test has been added for this case.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysis.cs
```diff
@@ -38,15 +39,18 @@ private GlobalFlowStateAnalysis(GlobalFlowStateAnalysisDomain analysisDomain, Gl
             out ValueContentAnalysisResult? valueContentAnalysisResult,
             InterproceduralAnalysisKind interproceduralAnalysisKind = InterproceduralAnalysisKind.None,
             bool pessimisticAnalysis = true,
-            InterproceduralAnalysisPredicate? interproceduralAnalysisPredicate = null)
+            InterproceduralAnalysisPredicate? interproceduralAnalysisPredicate = null,
+            ImmutableArray<INamedTypeSymbol> additionalSupportedValueTypes = default,
```
mavasani, line 43: @dotpaul Most of the DFA changes are just plumbing two new optional parameters to value content analysis so it can now track string values for custom named type operations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.OperationVisitor.cs
buyaa-n: > TryDecode can return multiple platform infos now via dataflow

Just curious why is that? Even there were multiple local variables only one can be used with the method, just not sure what exact case make it need to use/return multiple

mavasani: There is a test I added for that scenario. Even though there is a single argument, it can point to multiple potential values at runtime as assignments can be conditional

```csharp
        var windowsPlatform = OSPlatform.Create(""Windows"");
        var linuxPlatform = OSPlatform.Create(""Linux"");
        var platform = isWindows ? windowsPlatform : linuxPlatform;
        if(RuntimeInformation.IsOSPlatform(platform)) // Value here is "Windows" or "Linux"
        {
            M2();
        }
        else
        {
            {|#0:M2()|};
        }
```

buyaa-n: Ah i see, i saw that test but thought it would be clear if the value will be Windows or Linux, forgot we are doing static analysis 😛. Thanks for explanation!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatabilityAnalyzer.cs
```diff
@@ -219,7 +240,24 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
 
                 return;
 
-                OperationVisitor CreateOperationVisitor(GlobalFlowStateAnalysisContext context) => new OperationVisitor(guardMethods, osPlatformType, context);
+                OperationVisitor CreateOperationVisitor(GlobalFlowStateAnalysisContext context) => new OperationVisitor(guardMethods, osPlatformTypeArray[0], context);
```
buyaa-n, line 243:
```suggestion
                OperationVisitor CreateOperationVisitor(GlobalFlowStateAnalysisContext context) => new OperationVisitor(guardMethods, osPlatformType, context);
```

buyaa-n (approved):
Platform analyzer changes looks good, thanks!

mavasani:
Actually all the tests passed, only codecov upload is failing, so I am going to merge this to unblock @pranavkm.

@dotpaul I can address any DFA related feedback with a follow-up PR.

PR #4130, opened 2020-09-04
Prefer ExactSpelling=true for known extern methods

Mrnikbobjeff:
Implementation of [#35695](https://github.com/dotnet/runtime/issues/35695) the warning level was not explicitely stated as far as I can tell. I also extended the proposal to look at the EntryPoint parameter, as we would probably not want to flag 
`[DllImport("user32.dll", EntryPoint = "CallMsgFilterA", ExactSpelling = true)]
public XXX CallMsgFilter(...)`


Mrnikbobjeff:
Another question left open is if we want to include name manged functions. excluding them should be easy, and in my analysis no name mangled function exhibited behaviour which we want to catch with this analyzer @jkotas

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
```diff
@@ -0,0 +1,101 @@
...
+        [Fact]
+        public async Task CA1839CSharpTest()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System.Runtime.InteropServices;
+
+public class C
+{
+    [DllImport(""user32.dll"")]
+    static extern void CallMsgFilter(); // should have exactSpelling and W rule
```
jkotas, line 32: Note that this going to call `CallMsgFilterA` method because the method is not marked as `CharSet = CharSet.Unicode`. Changing it to exactSpelling and W rule will change the behavior - whether this change change is ok or not depends on the rest of the arguments to the method.

The analyzer has to take `CharSet` into account as well.

Mrnikbobjeff: So the analyzer would only propse the fix in this case when the CharSet is set to unicode as well?

jkotas: > Would this not call CallMsgFilter? That method is also present.

`CallMsgFilter` is weird. It has all 3 variants: `CallMsgFilter`, `CallMsgFilterA`, `CallMsgFilterW`. `CallMsgFilter` is an alias for `CallMsgFilterA`. So this PInvoke as written ends up calling `CallMsgFilterA`.

It may be better to demonstrate the desired behavior on more regular methods that have just A and W variants.

Here are the searches that this is about: 
https://github.com/dotnet/runtime/blob/d1c7946bb0892005a58ec647e3a6d8f3514cdc6b/src/coreclr/src/vm/method.cpp#L5383
https://github.com/dotnet/runtime/blob/d1c7946bb0892005a58ec647e3a6d8f3514cdc6b/src/coreclr/src/vm/method.cpp#L5391

We care about situations when the first call to FindEntryPoint fails (wasting cycles) and the second call succeeds. The analyzer should be about fixing these situations. 

Suggesting fixes for the other situations would be a style suggestion. Consistent style may be still a fine idea, but it would not be help with performance.

Mrnikbobjeff: So if my ordinal would not end on A or W I would invoke A by default even if the ordinals don't match and these are the diagnostics we are interested in?

jkotas: > TestA and TestW. our method name is Test and no charset parameter is is present or it is CharSet.Ansi.

Yes.

> we have TestA and TestW. our method name is Test and Charset.Unicode parameter present.

This does not have a peformance problem today. The warning maybe nice-to-have for style, not for performance.

There is one additional case to warn for. It is the case in the original issue description:

we have TestW method. our method name is TestW and Charset.Unicode parameter present.

Mrnikbobjeff: I believe I am covering all cases now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,146 @@
...
+        private static string GetEntryPointName(IMethodSymbol methodSymbol, AttributeData dllImportAttribute)
+        {
+            var hasEntryPointParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("EntryPoint", StringComparison.Ordinal));
+            return hasEntryPointParameter.Key is null ? methodSymbol.Name : hasEntryPointParameter.Value.Value.ToString();
+        }
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var model = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            INamedTypeSymbol? dllImportType = model.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute);
```
Evangelink, line 45: [optional] You could also use the `TryGet...` method instead of doing the null check.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,146 @@
...
+            if (dllImportType is null)
+                return;
+            SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            MethodDeclarationSyntax methodDeclaration = (MethodDeclarationSyntax)root.FindNode(context.Span);
+            IMethodSymbol methodSymbol = (IMethodSymbol)model.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
+            var dllImportAttribute = methodSymbol
+                .GetAttributes()
+                .FirstOrDefault(x => x.AttributeClass.Name.Equals("DllImportAttribute", StringComparison.Ordinal));
```
Evangelink, line 53: See comment on the analyzer but we usually check for the symbol rather than simply the name.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,146 @@
...
+            var dllName = dllImportAttribute.ConstructorArguments.First().Value.ToString();
```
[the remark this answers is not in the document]
Mrnikbobjeff: Even if it is guaranteed by the analyzer that it will only report if there is an argument present?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        internal static readonly Lazy<Dictionary<string, HashSet<string>>> KnownApis = new Lazy<Dictionary<string, HashSet<string>>>(() => CreateKnownApis());
+
+        private static Dictionary<string, HashSet<string>> CreateKnownApis()
+        {
+            var serializer = new XmlSerializer(typeof(KnownApi[]));
+            var str = MicrosoftNetCoreAnalyzersResources.ResourceManager.GetObject("KnownApiList", CultureInfo.InvariantCulture);
+            using var stringReader = new StringReader(str.ToString());
+            using var reader = XmlReader.Create(stringReader);
+            var knownApis = (serializer.Deserialize(reader) as KnownApi[])!;
```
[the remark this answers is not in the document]
Mrnikbobjeff: Fixed, it is a resource in our resx file. It should always be present

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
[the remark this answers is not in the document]
Evangelink: I might be missing something but how is it guaranteed?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,130 @@
...
+                {
+                    if (isCharSetUnicode && !isExactSpelling && methodName.EndsWith("W", StringComparison.Ordinal))
+                    {
+                        var diagnostic = Diagnostic.Create(WideRule, methodSymbol.Locations[0], methodName);
+                        context.ReportDiagnostic(diagnostic);
+                        return;
+                    }
+                    if (!isCharSetUnicode && !isExactSpelling && methodName.EndsWith("A", StringComparison.Ordinal))
+                    {
+                        var diagnostic = Diagnostic.Create(DefaultRule, methodSymbol.Locations[0], methodName);
```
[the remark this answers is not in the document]
Mrnikbobjeff: Thanks for that. The extensions really are helpful. It would be wounderful if some of this helpfull stuff would also be in the visual studio template! 👍

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1479,4 +1479,20 @@
   <data name="PlatformCompatibilityCheckSupportedOsMessage" xml:space="preserve">
     <value>'{0}' is supported on '{1}'</value>
   </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisDescription" xml:space="preserve">
+    <value>|{0}| is a known invocation on Windows without an equivalent -W suffix. Use ExactSpelling=true.</value>
```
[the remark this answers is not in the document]
Mrnikbobjeff: I implemented a couple of analyzers with the default Visual Studio Analyzer template, I believe the error message there has this format. No reason really

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,130 @@
...
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices
+{
+    /// <summary>
+    /// CA1839: Prefer ExactSpelling=true for known Apis analyzer
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
```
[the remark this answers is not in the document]
Mrnikbobjeff: To be perfectly honest, I never touched vbnet, I would not even know whether it works there. I will cobble up a test to verify.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,130 @@
...
+                .FirstOrDefault(x => x.AttributeClass.Name.Equals("DllImportAttribute", StringComparison.Ordinal));
+            if (dllImportAttribute is null)
+                return;
+
+            var hasEntryPointParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("EntryPoint", StringComparison.Ordinal));
+            var hasExactSpellingParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("ExactSpelling", StringComparison.Ordinal));
+            var hasCharSetParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("CharSet", StringComparison.Ordinal));
+
+            var methodName = hasEntryPointParameter.Key is null ? methodSymbol.Name : hasEntryPointParameter.Value.Value.ToString();
+            var isExactSpelling = !(hasExactSpellingParameter.Key is null) && (bool)hasExactSpellingParameter.Value.Value;
```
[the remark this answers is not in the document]
Mrnikbobjeff: using true.Equals should produce the same result without casts

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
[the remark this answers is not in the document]
Evangelink: @mavasani do you prefer to go for the usual `'{0}'` or to keep this syntax?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,146 @@
...
+                                                         async ct => { await AddExactSpelling(context.Document, methodDeclaration, editor, ct).ConfigureAwait(false); return editor.GetChangedDocument(); },
+                                                         equivalenceKey: title),
+                                        context.Diagnostics);
+                        return;
+                    }
+                }
+                if (methods.Contains(actualName + "A"))
+                {
+                    context.RegisterCodeFix(new MyCodeAction(title,
+                                                         async ct => { await AddASuffix(context.Document, methodDeclaration, actualName, editor, ct).ConfigureAwait(false); return editor.GetChangedDocument(); },
```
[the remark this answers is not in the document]
Mrnikbobjeff: I had  that originally, but I had problems replacing nodes after I called GetChangedDocument. I refactored and added a helper method to reduce duplication

Mrnikbobjeff:
@Evangelink thank you for the review, I believe I adressed all points which were not directly adressed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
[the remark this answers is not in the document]
Mrnikbobjeff: We only offer a fix if the string literal passed as parameter is present in the known api list. If there is no parameter the string can not match. `var dllName = dllImportAttribute.ConstructorArguments.FirstOrDefault().Value?.ToString() ?? string.Empty;

            if (!KnownApis.Value.TryGetValue(dllName, out var methods))
            {
                return;
            }
`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            if (model.GetDeclaredSymbol(methodDeclaration, context.CancellationToken) is not IMethodSymbol methodSymbol)
+            {
+                return;
+            }
+
+            var dllImportAttribute = methodSymbol
+                .GetAttributes()
+                .FirstOrDefault(x => x.AttributeClass.Equals(dllImportType));
+            var dllName = dllImportAttribute.ConstructorArguments.First().Value.ToString();
```
Evangelink, line 81: `dllImportAttribute` could be `null`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,154 @@
...
+            var dllImportAttribute = methodSymbol
+                .GetAttributes()
+                .FirstOrDefault(x => x.AttributeClass.Equals(dllImportType));
+
+            if (dllImportAttribute is null)
+            {
+                return;
+            }
+
+            var hasEntryPointParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("EntryPoint", StringComparison.Ordinal));
```
Evangelink, line 114: This is unused.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,154 @@
...
+            var isExactSpelling = hasExactSpellingParameter.Key is not null && bool.TryParse(hasExactSpellingParameter.Value.Value.ToString(), out var isExact) && isExact;
+            var isCharSetUnicode = hasCharSetParameter.Key is not null && Enum.TryParse<CharSet>(hasCharSetParameter.Value.Value.ToString(), out var actualCharSet) && actualCharSet == CharSet.Unicode;
+            var dllName = dllImportAttribute.ConstructorArguments.FirstOrDefault().Value?.ToString() ?? string.Empty;
+
+            if (!KnownApis.Value.TryGetValue(dllName, out var methods))
+            {
+                return;
+            }
+
+            if (methods.Contains(methodName)) //Whatever exactspelling is is wrong, wide method should be preferred
```
Evangelink, line 128: Not sure to get the first part of your comment.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                if (isCharSetUnicode && !isExactSpelling && methodName.EndsWith("W", StringComparison.Ordinal))
+                {
+                    var diagnostic = Diagnostic.Create(WideRule, methodSymbol.Locations[0], methodName);
```
Evangelink, line 132:
```suggestion
                    var diagnostic = methodSymbol.CreateDiagnostic(WideRule, methodName);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1482,4 +1482,20 @@
...
     <value>Replace with 'string.Contains'</value>
   </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisDescription" xml:space="preserve">
+    <value>|{0}| is a known invocation on Windows without an equivalent -W suffix. Use ExactSpelling=true.</value>
+  </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisMessage" xml:space="preserve">
+    <value>This method is part of the known API. Use 'ExactSpelling' to avoid name lookups on first invocation.</value>
+  </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisTitle" xml:space="preserve">
+    <value>Add 'ExactSpelling=true to DllImport declaration</value>
```
Evangelink, line 1492:
```suggestion
    <value>Add 'ExactSpelling=true' to 'DllImport' declaration</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1482,4 +1482,20 @@
...
+  </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisWithSuffixMessage" xml:space="preserve">
+    <value>|{0}| is a known invocation on Windows withan equivalent -W suffix. Use ExactSpelling=true and use the method without string encoding differences.</value>
```
Evangelink, line 1495:
```suggestion
    <value>|{0}| is a known invocation on Windows withan equivalent -W suffix. Use 'ExactSpelling=true' and use the method without string encoding differences.</value>
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
```diff
@@ -0,0 +1,87 @@
...
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.InteropServices.PreferIsExactSpellingIsTrueForKnownApisAnalyzer,
+    Microsoft.NetCore.Analyzers.InteropServices.PreferIsExactSpellingIsTrueForKnownApisFixer>;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices.UnitTests
+{
+    public sealed class PreferIsExactSpellingIsTrueForKnownApisTest
+    {
+        private DiagnosticResult CA1839_DefaultRule(int location, params string[] arguments)
```
Evangelink, line 14: [optional] `location` -> `markupId`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
```diff
@@ -0,0 +1,87 @@
...
+    static extern void abcdefg12(); // should have nothing, dll unknown
+    [DllImport(""user32.dll"", ExactSpelling = true)]
+    static extern void BroadcastSystemMessageA(); // should have nothing
+    [DllImport(""user32.dll"", ExactSpelling = true)]
+    static extern void BroadcastSystemMessage(); // should have nothing, perhaps we want to call method without a or w suffix
+    static void nonExtern() {} // should have nothing, not extern
+    static extern void onlyExtern(); // should have nothing, attribute missing
+}
+");
+
```
Evangelink, line 84:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
Evangelink, line 14: You are only expecting one argument so it'd be better to have the test helper expecting only one argument

```suggestion
        private DiagnosticResult CA1839_DefaultRule(int location, string knownApiName)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        private static void AnalyzeExternMethod(SymbolAnalysisContext context)
+        {
+            var methodSymbol = (IMethodSymbol)context.Symbol;
+            if (!methodSymbol.IsExtern)
+            {
+                return;
+            }
+            if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute, out var dllImportType))
```
Evangelink, line 101:
```suggestion

            if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute, out var dllImportType))
```

Evangelink: Alternatively you can also merge this with the previous condition.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
Mrnikbobjeff: Why would that help? As far as I can tell there are only overloads accepting a object[] as a parameter for setting of arguments, so we would simply delay the array creation slightly

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,163 @@
...
+            IReadOnlyList<SyntaxNode> arguments = generator.GetAttributeArguments(dllImportSyntax);
+
+            // [DllImport] attribute -> add or replace ExactSpelling parameter
+            SyntaxNode argumentValue = generator.LiteralExpression(actualExternalName + "A");
+            SyntaxNode newEntryPointText = generator.AttributeArgument(EntryPointText, argumentValue);
+
+            SyntaxNode? entryPointNode = FindNamedArgument(arguments, EntryPointText);
+            AddOrUpdateParameterOnAttribute(editor, dllImportSyntax!, entryPointNode, newEntryPointText);
+        }
+        private class MyCodeAction : DocumentChangeAction
```
Evangelink, line 150:
```suggestion

        private class MyCodeAction : DocumentChangeAction
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApis.cs
```diff
@@ -0,0 +1,154 @@
...
+            if (!methodSymbol.IsExtern)
+            {
+                return;
+            }
+
+            if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute, out var dllImportType))
+            {
+                return;
+            }
+            var dllImportAttribute = methodSymbol
```
Evangelink, line 106:
```suggestion

            var dllImportAttribute = methodSymbol
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1482,4 +1482,20 @@
...
     <value>Replace with 'string.Contains'</value>
   </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisDescription" xml:space="preserve">
+    <value>|{0}| is a known invocation on Windows without an equivalent -W suffix. Use 'ExactSpelling=true'.</value>
+  </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisMessage" xml:space="preserve">
+    <value>This method is part of the known API. Use 'ExactSpelling' to avoid name lookups on first invocation.</value>
+  </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisTitle" xml:space="preserve">
+    <value>Add 'ExactSpelling=true to 'DllImport' declaration</value>
```
Evangelink, line 1492: Closing quote is still missing:

```suggestion
    <value>Add 'ExactSpelling=true' to 'DllImport' declaration</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1482,4 +1482,20 @@
...
+  </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisWithSuffixMessage" xml:space="preserve">
+    <value>|{0}| is a known invocation on Windows withan equivalent -W suffix. Use ExactSpelling=true and use the method without string encoding differences.</value>
```
Evangelink, line 1495:
```suggestion
    <value>|{0}| is a known invocation on Windows withan equivalent -W suffix. Use 'ExactSpelling=true' and use the method without string encoding differences.</value>
```

Evangelink (commented):
A couple more cosmetic changes and then it's ok on my side. 

Tagging @mavasani for extra review.

Mrnikbobjeff:
Thank you @Evangelink for the detailed review. For future PR's I have a way better understanding of desired formatting and can adapt accordingly. The remaining issues should be fixed now

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        internal static readonly Lazy<Dictionary<string, HashSet<string>>> KnownApis = new Lazy<Dictionary<string, HashSet<string>>>(() => CreateKnownApis());
+
+        private static Dictionary<string, HashSet<string>> CreateKnownApis()
+        {
+            var serializer = new XmlSerializer(typeof(KnownApi[]));
+            var str = MicrosoftNetCoreAnalyzersResources.ResourceManager.GetObject("KnownApiList", CultureInfo.InvariantCulture);
+            using var stringReader = new StringReader(str.ToString());
+            using var reader = XmlReader.Create(stringReader);
+            var knownApis = (KnownApi[])serializer.Deserialize(reader);
```
mavasani, line 55: Why are we doing all these XML reader shenanigans? Can we please hard code the known APIs in source, just like all other analyzers do?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    /// <summary>
+    /// CA1839: Prefer ExactSpelling=true for known Apis analyzer
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class PreferIsExactSpellingIsTrueForKnownApisAnalyzer : DiagnosticAnalyzer
+    {
+        [Serializable]
+#pragma warning disable CA1034 // Nested types should not be visible
+        public class KnownApi //Only made public as XmlSerializer requires it
```
mavasani, line 36: Please delete and hard code the dictionary in source.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        private static void AnalyzeExternMethod(SymbolAnalysisContext context)
+        {
+            var methodSymbol = (IMethodSymbol)context.Symbol;
+            if (!methodSymbol.IsExtern)
+            {
+                return;
+            }
+
+            if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute, out var dllImportType))
```
mavasani, line 102: This needs to move into a CompilationStartAction

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
mavasani: I don’t believe this is resolved. Current analyzer implementation does not perform flow analysis, so cannot narrow down to cases that @jkotas mentioned where we only want to fire when first call fails and second succeeds. Current implementation is flow agnostic and would be a style suggestion, which does not seem to be the intent of this performance rule?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA1839 | Performance | Info | PreferIsExactSpellingIsTrueForKnownApisAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca1839)
```
mavasani, line 5: We need a new branch for post .NET5 analyzers. Master is targeting .NET5, so is not the right place for this PR.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,155 @@
...
+            if (dllImportAttribute is null)
+            {
+                return;
+            }
+
+            var hasExactSpellingParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("ExactSpelling", StringComparison.Ordinal));
+            var hasCharSetParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("CharSet", StringComparison.Ordinal));
+
+            var methodName = methodSymbol.GetActualExternName(dllImportAttribute);
+            var isExactSpelling = hasExactSpellingParameter.Key is not null && bool.TryParse(hasExactSpellingParameter.Value.Value.ToString(), out var isExact) && isExact;
```
mavasani, line 120: Seems like the second `.Value` access can throw: http://sourceroslyn.io/#Microsoft.CodeAnalysis/Symbols/TypedConstant.cs,99. You need a Kind check. Additionally, also seems the returned value can be `null`, so you need a `?.ToString()`

mavasani (changes requested):
Marking as request changes till we figure out the branching for post .NET5 analyzers. Tagging @jmarolf and @jeffhandley

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: Well there are currently 160k entries. I initially used the new generator feature to generate a compile time initialised dictionary with the constants, but the resulting compile time was an order of magnitude larger than compiling without the generated dictionary

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: See above

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
mavasani: > Well there are currently 160k entries. 

We would absolutely need performance measurements with this analyzer enabled then. It seems the initial load be pretty expensive?

Mrnikbobjeff: Yes the initial load is the heaviest part, it is below half a second though and once its done it is very cheap to query

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: How would I then access the type in the symbolaction?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
Mrnikbobjeff: We have all exported symbols in our `Dictionary<string,HashSet<string>>` . Suppose we have a method named `AbortSystemShutdown`, for which both `AbortSystemShutdownW` and `AbortSystemShutdownA` exist in our dictionary for `advapi32.dll`. As we have no attribute namedparameters present we default to `CharSet.Ansi`. As there is no known method for `AbortSystemShutdown` but for `AbortSystemShutdownA` the analyzer would warn as the first lookup fails. This is the targeted scenario is it not?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: Excluding name manged symbols this would shring to 96k method symbols

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
Evangelink: Yes that's only for the sake of clarity when you are trying to add new tests for an existing analyzer, it's sometimes hard to understand what's expected.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
jkotas: This analyzer does not need to be flow agnostic.

>  we only want to fire when first call fails and second succeeds

The first and second call in my comment referred to calls to GetProcAddress method done internally in the runtime. The comment did not refer to calls done by user code that is being analyzed.

mavasani: Ah thanks for the clarification, sorry about confusion on my side.

Mrnikbobjeff:
> Marking as request changes till we figure out the branching for post .NET5 analyzers. Tagging @jmarolf and @jeffhandley

Once this is established I have branches already implemented for all analyers marked as api-approved in the dotnet/runtime repository with the feedback from this branch already ported. I would  hold of sending a pr until this is figured out.

jmarolf:
@mavasani which release is this intended for? .NET 5  or .NET 6 Preview 1?

mavasani:
@jmarolf I believe all new .NET SDK API analyzers are now for post .NET5. @jeffhandley to confirm.

jeffhandley:
> I believe all new .NET SDK API analyzers are now for post .NET5.

That is correct. New analyzers should target 6.0 at this point.

jmarolf:
alright then this getting merged into master is the correct thing. @mavasani can you unblock this merging?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
mavasani: This comment is no longer valid. Master now targets .NET6

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+            var dllImportAttribute = methodSymbol
+                .GetAttributes()
+                .FirstOrDefault(x => x.AttributeClass.Equals(dllImportType));
+
+            if (dllImportAttribute is null)
+            {
+                return;
+            }
+
+            var hasExactSpellingParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("ExactSpelling", StringComparison.Ordinal));
```
mavasani, line 120: Seems like all the uses of values returned from `FirstOrDefault` invocations in this method need to be protected with a null check.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+            var hasCharSetParameter = dllImportAttribute.NamedArguments.FirstOrDefault(x => x.Key.Equals("CharSet", StringComparison.Ordinal));
+
+            var methodName = methodSymbol.GetActualExternName(dllImportAttribute);
+            var isExactSpelling = hasExactSpellingParameter.Key is not null && hasExactSpellingParameter.Value.Kind != TypedConstantKind.Array && bool.TryParse(hasExactSpellingParameter.Value.Value?.ToString(), out var isExact) && isExact;
```
mavasani, line 124:
```suggestion
            var isExactSpelling = hasExactSpellingParameter?.Key is not null && hasExactSpellingParameter.Value.Kind != TypedConstantKind.Array && bool.TryParse(hasExactSpellingParameter.Value.Value?.ToString(), out var isExact) && isExact;
```

mavasani: Nit: Please split the all the `&&` parts in the expression on separate lines for improved readability.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+            var isCharSetUnicode = hasCharSetParameter.Key is not null && hasCharSetParameter.Value.Kind != TypedConstantKind.Array && Enum.TryParse<CharSet>(hasCharSetParameter.Value.Value?.ToString(), out var actualCharSet) && actualCharSet == CharSet.Unicode;
```
mavasani, line 125:
```suggestion
            var isCharSetUnicode = hasCharSetParameter?.Key is not null && hasCharSetParameter.Value.Kind != TypedConstantKind.Array && Enum.TryParse<CharSet>(hasCharSetParameter.Value.Value?.ToString(), out var actualCharSet) && actualCharSet == CharSet.Unicode;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+            var dllName = dllImportAttribute.ConstructorArguments.FirstOrDefault().Value?.ToString() ?? string.Empty;
```
mavasani, line 126:
```suggestion
            var dllName = dllImportAttribute.ConstructorArguments.FirstOrDefault()?.Value?.ToString() ?? string.Empty;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+                {
+                    var diagnostic = methodSymbol.CreateDiagnostic(DefaultRule, methodName);
+                    context.ReportDiagnostic(diagnostic);
+                    return;
+                }
+            }
+
+            if (methods.Contains(methodName + "A") && !isCharSetUnicode)
+            {
+                if (isExactSpelling) //Known method has parameter set to true
```
mavasani, line 152:
```suggestion
                // Bail out if known method has parameter set to true
                if (isExactSpelling)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+            var isExactSpelling = hasExactSpellingParameter.Key is not null && hasExactSpellingParameter.Value.Kind != TypedConstantKind.Array && bool.TryParse(hasExactSpellingParameter.Value.Value?.ToString(), out var isExact) && isExact;
+            var isCharSetUnicode = hasCharSetParameter.Key is not null && hasCharSetParameter.Value.Kind != TypedConstantKind.Array && Enum.TryParse<CharSet>(hasCharSetParameter.Value.Value?.ToString(), out var actualCharSet) && actualCharSet == CharSet.Unicode;
+            var dllName = dllImportAttribute.ConstructorArguments.FirstOrDefault().Value?.ToString() ?? string.Empty;
+
+            if (!KnownApis.Value.TryGetValue(dllName, out var methods))
+            {
+                return;
+            }
+
+            if (methods.Contains(methodName))
```
mavasani, line 133: @jeffhandley I think someone from the runtime team should review the core analyzer semantics in terms of the analyzer implementation matching the recommended API usage.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: Not needed, hasCharSetParameter is a struct and can not be null, only default in which case the .Key is not null check is sufficient

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: Not needed, this is a struct which can never be null

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
Mrnikbobjeff: That would be wonderful, I am self taught and have no clue whether I did something bad when writing these analyzers, I just went with the docs samples and learned from there.

Mrnikbobjeff:
@mavasani I integrated your formatting and spelling changes. The null checks you suggested are all redundant though, as we would check non-nullable value types for null. We have to check on their reference type properties, but these checks are already there

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+{
+    /// <summary>
+    /// CA1839: Prefer ExactSpelling=true for known Apis fixer
+    /// </summary>
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = PreferIsExactSpellingIsTrueForKnownApisAnalyzer.RuleId), Shared]
+    public sealed class CSharpPreferIsExactSpellingIsTrueForKnownApisFixer : CodeFixProvider
+    {
+        private const string ExactSpellingText = "ExactSpelling";
+        private const string EntryPointText = "EntryPoint";
+        internal const string RuleId = "CA1839";
```
Evangelink, line 29: Un-needed.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(PreferIsExactSpellingIsTrueForKnownApisAnalyzer.RuleId);
+
+        private static SyntaxNode? FindNamedArgument(IReadOnlyList<SyntaxNode> arguments, string argumentName)
```
Evangelink, line 33: Could be `IEnumerable<SyntaxNode>`

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            return arguments.OfType<AttributeArgumentSyntax>().FirstOrDefault(arg => arg.NameEquals != null && arg.NameEquals.Name.Identifier.Text == argumentName);
```
Evangelink, line 35: You could use arrowed expression method.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+                if (actualName.EndsWith("A", StringComparison.OrdinalIgnoreCase))
+                {
+                    context.RegisterCodeFix(new PreferIsExactSpellingChangeAction(title,
+                                                     async ct => await ApplixFixAndGetChangedDocument(editor, AddExactSpelling(context.Document, methodDeclaration, editor, ct)).ConfigureAwait(false),
+                                                     equivalenceKey: title),
+                                    context.Diagnostics);
+                    return;
+                }
+            }
+            if (methods.Contains(actualName + "A"))
```
Evangelink, line 112:
```suggestion

            if (methods.Contains(actualName + "A"))
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                if (actualName.EndsWith("W", StringComparison.OrdinalIgnoreCase))
+                {
+                    context.RegisterCodeFix(new PreferIsExactSpellingChangeAction(title,
+                                                     async ct => await ApplixFixAndGetChangedDocument(editor, AddExactSpelling(context.Document, methodDeclaration, editor, ct)).ConfigureAwait(false),
+                                                     equivalenceKey: title),
+                                    context.Diagnostics);
+                    return;
+                }
+                if (actualName.EndsWith("A", StringComparison.OrdinalIgnoreCase))
```
Evangelink, line 103:
```suggestion

                if (actualName.EndsWith("A", StringComparison.OrdinalIgnoreCase))
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+            var editor = await DocumentEditor.CreateAsync(context.Document, context.CancellationToken).ConfigureAwait(false);
+            if (!PreferIsExactSpellingIsTrueForKnownApisAnalyzer.KnownApis.Value.TryGetValue(dllName, out var methods))
+            {
+                return;
+            }
+
+            var actualName = methodSymbol.GetActualExternName(dllImportAttribute);
+            if (methods.Contains(actualName))
+            {
+                if (actualName.EndsWith("W", StringComparison.OrdinalIgnoreCase))
```
Evangelink, line 95: Might be worth adding a comment with a small example over each `EndsWith` parts.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+                {
+                    context.RegisterCodeFix(new PreferIsExactSpellingChangeAction(title,
+                                                     async ct => await ApplixFixAndGetChangedDocument(editor, AddExactSpelling(context.Document, methodDeclaration, editor, ct)).ConfigureAwait(false),
+                                                     equivalenceKey: title),
+                                    context.Diagnostics);
+                    return;
+                }
+                if (actualName.EndsWith("A", StringComparison.OrdinalIgnoreCase))
+                {
+                    context.RegisterCodeFix(new PreferIsExactSpellingChangeAction(title,
```
Evangelink, line 105: I don't see any difference with the previous if block, if that's intended, you could merge the 2.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+                                                     equivalenceKey: title),
+                                    context.Diagnostics);
+            }
+        }
+
+        public static async Task AddExactSpelling(Document document, MethodDeclarationSyntax methodDeclaration, DocumentEditor editor, CancellationToken cancellationToken)
+        {
+            SyntaxGenerator generator = editor.Generator;
+            SemanticModel model = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+            var dllImportSyntax = methodDeclaration.AttributeLists.First(x => x.Attributes.Any(y => y.Name.ToString().Equals("DllImport", StringComparison.Ordinal)));
```
Evangelink, line 125: I don't know about that, we prefer to check the symbols  rather than just string names. If you add a test where you use the long form `DllImportAttribute` it should fail.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+            IReadOnlyList<SyntaxNode> arguments = generator.GetAttributeArguments(dllImportSyntax);
+
+            // [DllImport] attribute -> add or replace ExactSpelling parameter
+            SyntaxNode argumentValue = generator.TrueLiteralExpression();
+            SyntaxNode newExactSpellingArgument = generator.AttributeArgument(ExactSpellingText, argumentValue);
+            SyntaxNode? exactSpellingArgument = FindNamedArgument(arguments, ExactSpellingText);
+            AddOrUpdateParameterOnAttribute(editor, dllImportSyntax!, exactSpellingArgument, newExactSpellingArgument);
```
Evangelink, line 132: I think the `!` is un-needed.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        public static async Task AddASuffix(Document document, MethodDeclarationSyntax methodDeclaration, string actualExternalName, DocumentEditor editor, CancellationToken cancellationToken)
+        {
+            await AddExactSpelling(document, methodDeclaration, editor, cancellationToken).ConfigureAwait(false);
+            SyntaxGenerator generator = editor.Generator;
+            SemanticModel model = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+            var dllImportSyntax = methodDeclaration.AttributeLists.First(x => x.Attributes.Any(y => y.Name.ToString().Equals("DllImport", StringComparison.Ordinal)));
+            IReadOnlyList<SyntaxNode> arguments = generator.GetAttributeArguments(dllImportSyntax);
```
Evangelink, line 141: Same remark.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            // [DllImport] attribute -> add or replace EntryPoint parameter
+            SyntaxNode argumentValue = generator.LiteralExpression(actualExternalName + "A");
+            SyntaxNode newEntryPointText = generator.AttributeArgument(EntryPointText, argumentValue);
+
+            SyntaxNode? entryPointNode = FindNamedArgument(arguments, EntryPointText);
+            AddOrUpdateParameterOnAttribute(editor, dllImportSyntax!, entryPointNode, newEntryPointText);
```
Evangelink, line 148: Same remark for the `!`.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        public override FixAllProvider GetFixAllProvider()
+        {
+            return WellKnownFixAllProviders.BatchFixer;
+        }
```
Evangelink, line 154:
```suggestion
        => WellKnownFixAllProviders.BatchFixer;
```

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
Evangelink, line 5: Link is no longer correct.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,165 @@
...
+        internal static readonly Lazy<Dictionary<string, HashSet<string>>> KnownApis = new Lazy<Dictionary<string, HashSet<string>>>(() => CreateKnownApis());
+
+        private static Dictionary<string, HashSet<string>> CreateKnownApis()
+        {
+            var serializer = new XmlSerializer(typeof(KnownApi[]));
+            var str = MicrosoftNetCoreAnalyzersResources.ResourceManager.GetObject("KnownApiList", CultureInfo.InvariantCulture);
+            using var stringReader = new StringReader(str.ToString());
+            using var reader = XmlReader.Create(stringReader);
+            var knownApis = (KnownApi[])serializer.Deserialize(reader);
+            return knownApis.ToDictionary(x => x.Dll!, x => new HashSet<string>(x.Methods!));
```
Evangelink, line 56: You are assuming that the value is never null so it might be better to reflect that on the DTO.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(DefaultRule, WideRule);
+
+        public override void Initialize(AnalysisContext analysisContext)
+        {
+            analysisContext.EnableConcurrentExecution();
+            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            analysisContext.RegisterCompilationStartAction(ctx =>
+            {
+
```
Evangelink, line 93:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,165 @@
...
+                if (!ctx.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute, out var dllImportType))
+                {
+                    return;
+                }
+
+                ctx.RegisterSymbolAction(x => AnalyzeExternMethod(x, dllImportType), SymbolKind.Method);
```
Evangelink, line 99:
```suggestion
                if (ctx.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesDllImportAttribute, out var dllImportType))
                {
                    ctx.RegisterSymbolAction(x => AnalyzeExternMethod(x, dllImportType), SymbolKind.Method);
                }                
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisTest.cs
```diff
@@ -0,0 +1,86 @@
...
+               .WithLocation(markupId)
+               .WithArguments(knownApiName);
+
+        private DiagnosticResult CA1839_WideRule(int markupId, string knownApiName)
+           => VerifyCS.Diagnostic(PreferIsExactSpellingIsTrueForKnownApisAnalyzer.WideRule)
+               .WithLocation(markupId)
+               .WithArguments(knownApiName);
+
+        [Fact]
+        public async Task CA1839CSharpTest()
```
Evangelink, line 25: Could you please split this test into multiple? Like no diagnostic test, ExactSpelling missing, A suffix...

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1482,4 +1482,20 @@
   <data name="PreferStringContainsOverIndexOfCodeFixTitle" xml:space="preserve">
     <value>Replace with 'string.Contains'</value>
   </data>
+  <data name="PreferIsExactSpellingIsTrueForKnownApisDescription" xml:space="preserve">
+    <value>|{0}| is a known invocation on Windows without an equivalent -W suffix. Use 'ExactSpelling=true'.</value>
```
Evangelink, line 1486: I am still a little puzzled by the `|{0}|` syntax, I would go for `'{0}'`.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpPreferIsExactSpellingIsTrueForKnownApis.Fixer.cs
Mrnikbobjeff: I do not think arguments can have the same name?

Evangelink: Not sure to get what you are asking.

Mrnikbobjeff: Ah you are referring to the parameter type?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
Mrnikbobjeff: Would running the codefix provider again offer a solution? Or will it generate an invalid link again?

mavasani (changes requested):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+
+                if (!isCharSetUnicode && !isExactSpelling && methodName.EndsWith("A", StringComparison.Ordinal))
+                {
+                    var diagnostic = methodSymbol.CreateDiagnostic(DefaultRule, methodName);
+                    context.ReportDiagnostic(diagnostic);
+                    return;
+                }
+            }
+
+            if (methods.Contains(methodName + "A") && !isCharSetUnicode)
```
buyaa-n, line 155: I think this should be within the search condition above (row 138 `if (methods.Contains(methodName))`) and seems it is missing scenarios with `W` suffix @jkotas [suggested ](https://github.com/dotnet/runtime/issues/35695#issuecomment-686663918):
> When we see "Xxxx" in the entrypoint, the ExactSpelling autofixer needs to decide whether to suggest:
> 
> "XxxxW" + ExactSpelling=true. This should be suggested when we see "XxxxW" in the database.
> "Xxxx" + ExactSpelling=true. This should be suggested when we see "Xxxx" and no "XxxxW" in the database
> Leave the API alone. This should be the case when neither "Xxxx" nor "XxxxW" is in the database

0. If `methodName` exist in DB (already covered with  `if (methods.Contains(methodName))` condition):
1. If `methodName + "W"` exist  we suggest adding `"W"` suffix and setting `ExactSpelling=true`
2. If `methodName + "W"` does not exist we suggest only setting `ExactSpelling=true`

Probably we want to do same for "A" suffix but from from @elinor-fung's [comment ](https://github.com/dotnet/runtime/issues/35695#issuecomment-623803330) seems suggesting `ExactSpelling=true` when `CharSet=CharSet.Ansi` might not important or might not much helpful, if we want to cover it the rule should be same

3. If `methodName + "A"` exist  we suggest adding `"A"` suffix and setting `ExactSpelling=true`
4. If `methodName + "A"` does not exist we suggest only setting `ExactSpelling=true`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+            var dllName = dllImportAttribute.ConstructorArguments.FirstOrDefault().Value?.ToString() ?? string.Empty;
+
+            if (!KnownApis.Value.TryGetValue(dllName, out var methods))
+            {
+                return;
+            }
+
+            if (methods.Contains(methodName))
+            {
+                if (isCharSetUnicode && !isExactSpelling && methodName.EndsWith("W", StringComparison.Ordinal))
```
buyaa-n, line 140: `!isExactSpelling` better to be moved before DB search ( before `if (methods.Contains(methodName))`), here and all other occurrences below 
```suggestion
                if (isCharSetUnicode && methodName.EndsWith("W", StringComparison.Ordinal))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PreferIsExactSpellingIsTrueForKnownApisAnalyzer.cs
buyaa-n: It seems missing some scenarios which i commented [here ](https://github.com/dotnet/roslyn-analyzers/pull/4130#discussion_r537929600), but i would defer to @elinor-fung for final expected scenarios

mavasani:
PR needs to be retargeted to `release\6.0.xx` branch. Do we still plan to target this PR for .NET6 release?

buyaa-n:
Closing this PR as there were no activity for a while, @Mrnikbobjeff feel free to reopen the PR after the feedbacks addressed and when its ready for review

PR #4136, opened 2020-09-07
CA2249: Rename code fix title

### LINKED ISSUE #4134: CA2249 codefix menu item text improvement
#### Analyzer package

#### Package Version

v3.3.0

#### Diagnostic ID

[CA2249](https://docs.microsoft.com/en-ca/visualstudio/code-quality/ca2249)

#### Repro steps

image

The text "Consider ..." doesn't feel right for this menu. These are directives for action, not contemplation.

Instead it should read as "Use string.Contains instead of string.IndexOf"

Evangelink:
Fix #4134


mavasani (approved):
LGTM. We should also remove “Consider” term from the title used by the analyzer, but that can be a separate change. Thanks!

PR #4138, opened 2020-09-07
CA2014: Don't warn if a local function containing stackalloc is inside a loop

### LINKED ISSUE #4135: `CA2014: Do not use stackalloc in loops` should not be report for local function inside loop
### Analyzer

**Diagnostic ID**: [CA2014](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2014?view=vs-2019): `CA2014: Do not use stackalloc in loops`

### Analyzer source

**Version**: [SDK 5.0.100-preview.8](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug

`CA2014: Do not use stackalloc in loops` should not be report for local function inside loop

### Steps To Reproduce

```cs
        public static void XXX()
        {
            while (true)
            {
                XX();

                static void XX()
                {
                    Span<int> tmp = stackalloc int[10];
                    Console.WriteLine(tmp[0]);
                }
            }
        }
```

### Expected behavior
No CA2014, since it the same as the below code.
```cs
        public static void XXX()
        {
            while (true)
            {
                XX();
            }

                static void XX()
                {
                    Span<int> tmp = stackalloc int[10];
                    Console.WriteLine(tmp[0]);
                }
        }
```

### Actual behavior
CA2014

Youssef1313:
Fixes #4135

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
@@ -39,6 +39,12 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 // We found a stackalloc.  Walk up from it to see if it's in a loop at any level.
                 for (SyntaxNode node = ctx.Node; node != null; node = node.Parent)
                 {
+                    // Don't warn of the stackalloc is in anonymous function.
+                    if (ctx.SemanticModel.GetOperation(node, ctx.CancellationToken)?.Kind == OperationKind.AnonymousFunction)
+                    {
+                        return;
+                    }
+
                     switch (node.Kind())
                     {
                         // Don't warn if a local function containing stackalloc is inside a loop.
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
@@ -41,6 +41,10 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 {
                     switch (node.Kind())
                     {
+                        // Don't warn if a local function containing stackalloc is inside a loop.
+                        case SyntaxKind.LocalFunctionStatement:
+                            return;
+
                         // Look for loops.  We don't bother with ad-hoc loops via gotos as we're
                         // too likely to incur false positives.
                         case SyntaxKind.ForStatement:
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
@@ -39,16 +39,13 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 // We found a stackalloc.  Walk up from it to see if it's in a loop at any level.
                 for (SyntaxNode node = ctx.Node; node != null; node = node.Parent)
                 {
-                    // Don't warn of the stackalloc is in anonymous function.
-                    if (ctx.SemanticModel.GetOperation(node, ctx.CancellationToken)?.Kind == OperationKind.AnonymousFunction)
-                    {
-                        return;
-                    }
-
                     switch (node.Kind())
                     {
-                        // Don't warn if a local function containing stackalloc is inside a loop.
+                        // Don't warn if a local function or lambda containing stackalloc is inside a loop.
                         case SyntaxKind.LocalFunctionStatement:
+                        case SyntaxKind.ParenthesizedLambdaExpression:
+                        case SyntaxKind.SimpleLambdaExpression:
+                        case SyntaxKind.AnonymousMethodExpression:
                             return;
 
                         // Look for loops.  We don't bother with ad-hoc loops via gotos as we're
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpSecurityCodeFixVerifier<
     Microsoft.NetCore.Analyzers.Runtime.DoNotUseStackallocInLoopsAnalyzer,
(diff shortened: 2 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
```diff
@@ -41,6 +41,9 @@ public sealed override void Initialize(AnalysisContext context)
                 {
                     switch (node.Kind())
                     {
+                        // Don't warn if a local function containing stackalloc is inside a loop.
+                        case SyntaxKind.LocalFunctionStatement:
+                            return;
```
Evangelink, line 46:
```suggestion
                            return;
                            
```

Evangelink (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
```diff
@@ -63,6 +63,26 @@ unsafe class TestClass {
                 }");
         }
 
+        [Fact]
+        public async Task NoDiagnostics_StackallocInLoopButInsideALocalFunction()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
```
Evangelink, line 69: You will have to use the full syntax to specify C# 8 version in order to use the local function feature for your test.

Youssef1313:
This currently fail because the default language version for tests is 7.3, which doesn't support local functions.

Do you want me to update the language version globally in:

https://github.com/dotnet/roslyn-analyzers/blob/0cbdd68e14722e911d50eee64ee95823b3ffa09c/src/Test.Utilities/CSharpCodeRefactoringVerifier%601%2BTest.cs#L38

or only in the new test?

Evangelink:
@Youssef1313 I think it's best to only fix the test. Changing the main version would deserve another PR as we will need to cleanup the tests manually specifying the version 8.

Evangelink (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
```diff
@@ -41,6 +41,10 @@ public sealed override void Initialize(AnalysisContext context)
                 {
                     switch (node.Kind())
                     {
+                        // Don't warn if a local function containing stackalloc is inside a loop.
+                        case SyntaxKind.LocalFunctionStatement:
```
mavasani, line 45: Can we also bail out on OperationKind.AnonymousFunction? Please add a test for it as well.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
```diff
@@ -63,6 +64,30 @@ unsafe class TestClass {
                 }");
         }
 
+        [Fact]
+        public async Task NoDiagnostics_StackallocInLoopButInsideALocalFunction()
```
mavasani, line 68: Please add a test for lambda case.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
Evangelink: I was thinking about that but without dataflow we cannot guarantee the func won't be used within a loop. I guess it's probably better to miss a few cases rather than wrongly reporting.

stephentoub: It actually doesn't matter if the func is used in a loop. The stack memory will be released upon the func's return.

Youssef1313: @mavasani I've added a check for OperationKind.AnonymousFunction.
Should I also add one for FlowAnonymousFunction? I honestly don't know the difference ;-(

Should I also move the SyntaxKind.LocalFunctionStatement check and use OperationKind.LocalFunction instead? I think it doesn't matter, but just thought to group the checks together instead of making one dependent on SyntaxKind and the other dependent on OperationKind.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
```diff
@@ -63,6 +64,52 @@ unsafe class TestClass {
...
+        {
+            await new VerifyCS.Test
+            {
+                LanguageVersion = LanguageVersion.CSharp8,
+                TestCode = @"
+using System;
+class TestClass {
+    private static void StackallocInLoopButInsideALambda() {
+        while (true) {
+            Action a = () => {
```
Evangelink, line 102: If you don't mind also doing a check with a `Func` that'd be nice.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
```diff
@@ -39,8 +39,18 @@ public sealed override void Initialize(AnalysisContext context)
                 // We found a stackalloc.  Walk up from it to see if it's in a loop at any level.
                 for (SyntaxNode node = ctx.Node; node != null; node = node.Parent)
                 {
+                    // Don't warn of the stackalloc is in anonymous function.
+                    if (ctx.SemanticModel.GetOperation(node, ctx.CancellationToken)?.Kind == OperationKind.AnonymousFunction)
```
Evangelink, line 43: Sorry I have read too fast earlier! I think it's better not to go for the `SemanticModel.GetOperation` and so to have 2 syntax kind exceptions on the switch.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
Evangelink: Cheers @Youssef1313

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoops.cs
mavasani: My bad, I didn’t see the current code was operating on SyntaxKinds rather then operation kinds. Instead of fetching operation, it’s better to just bail out on different lambda related syntax kinds.

Youssef1313: These are `ParenthesizedLambdaExpression`, `SimpleLambdaExpression`, and `AnonymousMethodExpression` only? or I'm missing any others?

mavasani: Thanks, that seems to match compiler implementation: https://github.com/dotnet/roslyn/blob/9a658fab0803051491fa888bbed5a6668b8874e3/src/Compilers/CSharp/Portable/Syntax/SyntaxNodeExtensions.cs#L19-L31

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotUseStackallocInLoopsTests.cs
```diff
@@ -63,6 +64,75 @@ unsafe class TestClass {
...
```
mavasani, line 102: Can you also add a unit test for simple lambda expression syntax, i.e. say `Action<int> a = _ => Console.Write((stackalloc int[10]).Length);`

image

mavasani (approved):
LGTM. I can merge once you add the suggested unit test.

mavasani (approved):

PR #4139, opened 2020-09-07
CA2231: Update message for clarity

### LINKED ISSUE #4132: CA2231: Cryptic
From @ericwj in MicrosoftDocs/visualstudio-docs#5804:

> I've read the warning message for CA2231 three times and I still have no idea what it is talking about.
> 
> ```
> warning CA2231: Overload operator equals on overriding value type Equals
> ```
> 
> Yet the cause is crystal clear: [A value type overrides System.Object.Equals but does not implement the equality operator](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2231?view=vs-2019).
> 
> I am not event driven and don't follow patterns like a compiler, perhaps if the message is a directive or suggestion certainly first of all don't use **on** but **when** and generally I think there should be 1-2-3 words extra in the message to make it read like it is normal English.

Youssef1313:
Fixes #4132

Do I need a parallel doc PR?

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
@@ -413,7 +413,7 @@
     <value>In most programming languages there is no default implementation of the equality operator (==) for value types. If your programming language supports operator overloads, you should consider implementing the equality operator. Its behavior should be identical to that of Equals.</value>
   </data>
   <data name="OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage" xml:space="preserve">
-    <value>Overload operator equals on overriding value type Equals</value>
+    <value>Implement the equality operators and make their behavior identical to that of Equals method.</value>
   </data>
   <data name="PassSystemUriObjectsInsteadOfStringsTitle" xml:space="preserve">
     <value>Pass system uri objects instead of strings</value>
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
@@ -413,7 +413,7 @@
     <value>In most programming languages there is no default implementation of the equality operator (==) for value types. If your programming language supports operator overloads, you should consider implementing the equality operator. Its behavior should be identical to that of Equals.</value>
   </data>
   <data name="OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage" xml:space="preserve">
-    <value>Implement the equality operators and make their behavior identical to that of Equals method.</value>
+    <value>Implement the equality operators and make their behavior identical to that of 'Equals' method</value>
   </data>
   <data name="PassSystemUriObjectsInsteadOfStringsTitle" xml:space="preserve">
     <value>Pass system uri objects instead of strings</value>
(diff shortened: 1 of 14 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -407,7 +407,7 @@
     <value>Change '{0}' to be read-only by removing the property setter</value>
   </data>
   <data name="OverloadOperatorEqualsOnOverridingValueTypeEqualsTitle" xml:space="preserve">
-    <value>Overload operator equals on overriding value type Equals</value>
+    <value>A value type overrides System.Object.Equals but does not implement the equality operator</value>
```
Evangelink, line 410: Use single quotes around types.

```suggestion
    <value>A value type overrides 'System.Object.Equals' but does not implement the equality operator</value>
```

We tend to keep short sentences for title but I am not sure how to shrink your message.

Evangelink (commented):
Let's wait for @mavasani 's input.

mavasani:
@Youssef1313 did you intend to fix OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage? The title is not what the user was complaining about.

Youssef1313:
@mavasani Although the issue mentions that the message is confusing, but it also mentions that "Overload operator equals on overriding value type Equals" is confusing, which is the title.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Youssef1313: @Evangelink I took it just a copy/paste.

Things that can make it a (little) bit shorter:

- "the equality operator" -> "equality operators" (a little bit shorter, and more accurate to be plural I think (to represent == and !=)
- "does not" -> "doesn't" (but I'm not sure if you follow a specific convention for this and want to be consistent)
- "System.Object.Equals" -> "object.Equals" (Note lower case object to make it still accurate, but shorter at the same time)

I'll wait for @mavasani to weigh in for this.

mavasani:
@Youssef1313 the user has complained about the diagnostic message, not the title. Let’s update the message string and revert the title change.

Youssef1313:
@mavasani Great, then I probably misunderstood his complaint.

The current message is:

> In most programming languages there is no default implementation of the equality operator (==) for value types. If your programming language supports operator overloads, you should consider implementing the equality operator. Its behavior should be identical to that of Equals.

What about the following:

> Implement the equality operators and make their behavior identical to that of Equals method.

I intentionally used plural (rather than singular in the original message) as user should implement both `==` and `!=`.

Does it look good?

mavasani (changes requested):
Please update `OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage` and revert `OverloadOperatorEqualsOnOverridingValueTypeEqualsTitle`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -413,7 +413,7 @@
     <value>In most programming languages there is no default implementation of the equality operator (==) for value types. If your programming language supports operator overloads, you should consider implementing the equality operator. Its behavior should be identical to that of Equals.</value>
   </data>
   <data name="OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage" xml:space="preserve">
-    <value>Overload operator equals on overriding value type Equals</value>
+    <value>Implement the equality operators and make their behavior identical to that of Equals method.</value>
```
mavasani, line 416:
```suggestion
    <value>Implement the equality operators and make their behavior identical to that of 'Equals' method.</value>
```

mavasani (approved):

Issue #4091, opened 2020-08-28
Platform compat analyzer: should be disabled when compiling for .NET Framework


terrajobst:
We can't mark .NET Framework projects as Windows-specific (because the attribute doesn’t ship with .NET Framework). Which means if you set `<AnalysisLevel>` to `5`, all Windows-specific APIs are now flagged. This is usually not a big thing, because most APIs are in-box and in-box APIs aren’t marked. But if someone consumes a NuGet package with internal attributes, they could get false positives.

The customer can work this around by either disabling the rule or by applying the attribute to their assembly, but this would also require them to define it themselves.

We should disable the analyzer if the TFM is < `net5.0`.

jeffhandley:
We would still allow the analyzer to be forced to enabled for TFM < `net5.0` though, respecting any attributes they apply to their own APIs.

jeffhandley:
Per @mavasani:

> Note that the analyzer can read the underlying MSBuild TFM through analyzer options. So, if we wanted different default behavior based on it, it would be possible. For example, we can add a new boolean editorconfig option for this analyzer, say enabled_on_net_framework, which can be defaulted to false, and user would need an additional step 4. to set this option to true in their editorconfig to enable this analyzer on .NET Framework.

buyaa-n:
Questions around TFM edge cases:
- What if there is no TFM specified? Could I assume we should disable the analzyer to avoid false positives?
- What if multiple TFMs provided (separated by `;`). Could i assume analyzer should be enabled if at least one of them is net5.0 or above?

CC @terrajobst @jeffhandley

mavasani:
> What if multiple TFMs provided (separated by ;). Could i assume analyzer should be enabled if at least one of them is net5.0 or above?

Each compilation should only have a single TFM, even if project file has multiple TFMs. I am not suggesting we should run the analyzer if subset of TFMs are supported and subset are not supported, Immo/Jeff should make a call on that.

mavasani:
Actually, I am not sure if analyzer will have access to `TargetFrameworks` property, as each compilation is done using individual `TargetFramework` from the list of frameworks in that property. You may have to try it out...

buyaa-n:
> Actually, I am not sure if analyzer will have access to `TargetFrameworks` property, as each compilation is done using individual `TargetFramework` from the list of frameworks in that property. 

That makes sense, most likely that is true

> You may have to try it out...

Guess only option is manual testing, will give it try

[...]

PR #4153, opened 2020-09-10
By default platform compat analyzer should be disabled for TFM < 5.0

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4091

### DIFF
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -744,12 +744,12 @@ Example: `dotnet_code_quality.CA1711.allowed_suffixes = Flag|Flags`
 
 ### Enable platform compatibility analyzer for TFMs <= net5.0
 
-Option Name: `enable_platform_analyzer`
+Option Name: `enable_platform_analyzer_on_pre_net5_target`
 
-Configurable Rules: [CA14016](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1416)
+Configurable Rules: [CA14016](https://docs.microsoft.com/visualstudio/code-quality/ca1416)
 
 Option Values: Boolean values
 
 Default Value: `false`
 
-Example: `dotnet_code_quality.enable_platform_analyzer = true` or `dotnet_code_quality.CA1416.enable_platform_analyzer = true`
+Example: `dotnet_code_quality.enable_platform_analyzer_on_pre_net5_target = true` or `dotnet_code_quality.CA1416.enable_platform_analyzer_on_pre_net5_target = true`
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -741,3 +741,15 @@ Configurable Rules: [CA1711](https://docs.microsoft.com/visualstudio/code-qualit
 Option Values: List (separated by '|') of allowed suffixes
 
 Example: `dotnet_code_quality.CA1711.allowed_suffixes = Flag|Flags`
+
+### Enable platform compatibility analyzer for TFMs <= net5.0
+
+Option Name: `enable_platform_analyzer`
+
+Configurable Rules: [CA14016](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1416)
+
+Option Values: Boolean values
+
+Default Value: `false`
+
+Example: `dotnet_code_quality.enable_platform_analyzer = true` or `dotnet_code_quality.CA1416.enable_platform_analyzer = true`
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -49,7 +49,6 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         private const string IsOSPlatform = nameof(IsOSPlatform);
         private const string IsPrefix = "Is";
         private const string OptionalSuffix = "VersionAtLeast";
-        private const char SeparatorSemicolon = ';';
         private const string Net = "net";
 
         internal static DiagnosticDescriptor SupportedOsVersionRule = DiagnosticDescriptorHelper.Create(RuleId,
@@ -159,6 +158,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 {
                     return methods.Add(runtimeIsOSPlatformMethod);
                 }
+
                 return methods;
             }
 
@@ -172,30 +172,24 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         private static bool PlatformAnalysisAllowed(AnalyzerOptions options, Compilation compilation, CancellationToken token)
         {
             var tfmString = options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.TargetFramework, compilation, token);
-            if (tfmString != null)
-            {
-                foreach (var tfm in tfmString.Split(SeparatorSemicolon))
-                {
-                    if (TryParseTfm(tfm, out var platform, out var version))
-                    {
-                        if (platform.Equals(Net, StringComparison.OrdinalIgnoreCase) &&
-                            version != null &&
-                            int.TryParse(version[0].ToString(), out var value)
-                            && value >= 5)
-                        {
-                            return true;
-                        }
-                        else
-                        {
-                            return LowerTargetsEnabled(options, compilation, token);
-                        }
-                    }
-                }
 
-                return false;
+            if (tfmString != null &&
+               TryParseTfm(tfmString, out var platform, out var version))
+            {
+                if (platform.Equals(Net, StringComparison.OrdinalIgnoreCase) &&
+                    version != null &&
+                    int.TryParse(version[0].ToString(), out var value) &&
+                    value >= 5)
+                {
+                    return true;
+                }
+                else
+                {
+                    return LowerTargetsEnabled(options, compilation, token);
+                }
             }
 
-            return true;
+            return false;
         }
 
         private static bool LowerTargetsEnabled(AnalyzerOptions options, Compilation compilation, CancellationToken cancellationToken) =>
@@ -215,10 +209,12 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         version = tfm[i..];
                         return true;
                     }
+
                     platform = null;
                     return false;
                 }
             }
+
             platform = tfm;
             return true;
         }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -173,8 +173,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         {
             var tfmString = options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.TargetFramework, compilation, token);
 
-            if (tfmString != null &&
-                tfmString.Length >= 4 &&
+            if (tfmString?.Length >= 4 &&
                 tfmString.StartsWith(Net, StringComparison.OrdinalIgnoreCase) &&
                 int.TryParse(tfmString[3].ToString(), out var major) &&
                 major >= 5)
@@ -189,7 +188,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
 
         private static bool LowerTargetsEnabled(AnalyzerOptions options, Compilation compilation, CancellationToken cancellationToken) =>
             compilation.SyntaxTrees.FirstOrDefault() is { } tree &&
-            options.GetBoolOptionValue(EditorConfigOptionNames.EnablePlatformAnalyzer, SupportedOsRule, tree, compilation, false, cancellationToken);
+            options.GetBoolOptionValue(EditorConfigOptionNames.EnablePlatformAnalyzerOnPreNet5Target, SupportedOsRule, tree, compilation, false, cancellationToken);
 
         private void AnalyzeOperationBlock(
             OperationBlockStartAnalysisContext context,
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -174,50 +174,22 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             var tfmString = options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.TargetFramework, compilation, token);
 
             if (tfmString != null &&
-               TryParseTfm(tfmString, out var platform, out var version))
+                tfmString.Length >= 4 &&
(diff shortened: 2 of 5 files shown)


src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -28,6 +28,11 @@ internal static partial class EditorConfigOptionNames
         /// </summary>
         public const string ExcludeAsyncVoidMethods = "exclude_async_void_methods";
 
+        /// <summary>
+        /// Boolean option to enable platform commpatibility analyzer for lower version than 5.0.
```
mavasani, line 32: Please add a section at the end of https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#supported-editorconfig-options for the new option.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -162,6 +169,56 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
...
+        {
+            version = null;
+            for (int i = 0; i < tfm.Length; i++)
+            {
+                if (char.IsDigit(tfm[i]))
+                {
+                    if (i > 0)
+                    {
+                        platform = tfm.Substring(0, i);
+                        version = tfm[i..];
```
mavasani, line 209: Will this handle all flavored platforms mentioned here: https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#design?

buyaa-n: Yes, except `Tizen, Unity` i guess they are not expected to be supported though. Luckily we don't need to care about the different platforms, only `net5` section is matter from `net5.0-XYZ` so we are only checking if everything `before the first number` is equal to `net` and the `first number` is >= 5

mavasani: I thought the range expression will give entire substring starting from index i till the end.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -41,6 +41,89 @@ public void MethodWithArrayArgument()
...
 
+        public static IEnumerable<object[]> Create_DifferentTfms()
+        {
+            yield return new object[] { "build_property.TargetFramework = net472", false };
+            yield return new object[] { "build_property.TargetFramework = netcoreapp1.0", false };
+            yield return new object[] { "build_property.TargetFramework = dotnet", false };
+            yield return new object[] { "build_property.TargetFramework = uap10.0", false };
+            yield return new object[] { "build_property.TargetFramework = netstandard2.1", false };
+            yield return new object[] { "build_property.TargetFramework = net5", true };
+            yield return new object[] { "build_property.TargetFramework = net99", true };
```
mavasani, line 52: Can you add tests for TFM value “net5.0” and “net5.0-xxx”?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
mavasani: Can we simplify the check to:
1. .Length >= 4 AND
2. .StartsWith(“net”, OrdinalIgnoreCase) AND
3 int.TryParse(tfm[3], out int major) AND
4. major >= 5

buyaa-n: Yes, we don't really need to parse entire TFM

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
buyaa-n: Sure

docs/Analyzer Configuration.md
```diff
@@ -741,3 +741,15 @@ Configurable Rules: [CA1711](https://docs.microsoft.com/visualstudio/code-qualit
 Option Values: List (separated by '|') of allowed suffixes
 
 Example: `dotnet_code_quality.CA1711.allowed_suffixes = Flag|Flags`
+
+### Enable platform compatibility analyzer for TFMs <= net5.0
+
+Option Name: `enable_platform_analyzer`
+
+Configurable Rules: [CA14016](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1416)
```
Youssef1313, line 749:
```suggestion
Configurable Rules: [CA1416](https://docs.microsoft.com/visualstudio/code-quality/ca1416)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -162,6 +169,28 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
                     (method.Parameters.Length == 0 || method.Name.EndsWith(OptionalSuffix, StringComparison.Ordinal));
         }
 
+        private static bool PlatformAnalysisAllowed(AnalyzerOptions options, Compilation compilation, CancellationToken token)
+        {
+            var tfmString = options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.TargetFramework, compilation, token);
+
+            if (tfmString != null &&
+                tfmString.Length >= 4 &&
```
mavasani, line 177:
```suggestion
            if (tfmString?.Length >= 4 &&
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -162,6 +169,28 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
...
+                tfmString.StartsWith(Net, StringComparison.OrdinalIgnoreCase) &&
+                int.TryParse(tfmString[3].ToString(), out var major) &&
```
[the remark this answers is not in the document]
mavasani: I'll let @buyaa-n make a call on whether we want to handle that case. If we wanted to, you can possibly write this as:
```suggestion
                tfmString.Length >= 5 &&  ? int.TryParse(tfmString[3..4], out var major) : int.TryParse(tfmString[3].ToString(), out major)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -41,6 +41,91 @@ public void MethodWithArrayArgument()
...
+        public static IEnumerable<object[]> Create_DifferentTfms()
+        {
+            yield return new object[] { "build_property.TargetFramework = net472", false };
+            yield return new object[] { "build_property.TargetFramework = netcoreapp1.0", false };
+            yield return new object[] { "build_property.TargetFramework = dotnet", false };
+            yield return new object[] { "build_property.TargetFramework = uap10.0", false };
+            yield return new object[] { "build_property.TargetFramework = netstandard2.1", false };
+            yield return new object[] { "build_property.TargetFramework = net5", true };
+            yield return new object[] { "build_property.TargetFramework = net5.0-windows", true };
+            yield return new object[] { "build_property.TargetFramework = net5.0-ios14.0", true };
```
mavasani, line 53: One additional scenario:
```suggestion
            yield return new object[] { "build_property.TargetFramework = net5", true };
            yield return new object[] { "build_property.TargetFramework = net5.0", true };
            yield return new object[] { "build_property.TargetFramework = net5.0-windows", true };
            yield return new object[] { "build_property.TargetFramework = net5.0-ios14.0", true };
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -41,6 +41,91 @@ public void MethodWithArrayArgument()
...
+            yield return new object[] { "build_property.TargetFramework = net472\ndotnet_code_quality.enable_platform_analyzer=false", false };
+            yield return new object[] { "build_property.TargetFramework = netcoreapp1.0\ndotnet_code_quality.enable_platform_analyzer=true", true };
+            yield return new object[] { "build_property.TargetFramework = netcoreapp1.0\ndotnet_code_quality.CA1416.enable_platform_analyzer=false", false };
+            yield return new object[] { "build_property.TargetFramework = dotnet\ndotnet_code_quality.enable_platform_analyzer=true", true };
+            yield return new object[] { "build_property.TargetFramework = uap10.0\ndotnet_code_quality.enable_platform_analyzer=false", false };
+            yield return new object[] { "build_property.TargetFramework = netstandard2.1\ndotnet_code_quality.enable_platform_analyzer=true", true };
+            yield return new object[] { "build_property.TargetFramework = netstandard2.1\ndotnet_code_quality.enable_platform_analyzer=false", false };
+            yield return new object[] { "build_property.TargetFramework = net5\ndotnet_code_quality.enable_platform_analyzer=false", true };
```
mavasani, line 96:
```suggestion
            yield return new object[] { "build_property.TargetFramework = net5\ndotnet_code_quality.enable_platform_analyzer=false", true };
            yield return new object[] { "build_property.TargetFramework = net5.0\ndotnet_code_quality.enable_platform_analyzer=false", true };
            yield return new object[] { "build_property.TargetFramework = net5.0-windows\ndotnet_code_quality.enable_platform_analyzer=false", true };
            yield return new object[] { "build_property.TargetFramework = net5.0-ios14.0\ndotnet_code_quality.enable_platform_analyzer=false", true };
```

docs/Analyzer Configuration.md
```diff
@@ -741,3 +741,15 @@ Configurable Rules: [CA1711](https://docs.microsoft.com/visualstudio/code-qualit
 Option Values: List (separated by '|') of allowed suffixes
 
 Example: `dotnet_code_quality.CA1711.allowed_suffixes = Flag|Flags`
+
+### Enable platform compatibility analyzer for TFMs <= net5.0
+
+Option Name: `enable_platform_analyzer`
```
mavasani, line 747: Given the option is only applicable to pre-.NET5 TFMs, maybe we can make it explicit in the option name so it is clear that the option is not applicable to net5 or later TFMs? If you make this change, please update all uses of `enable_platform_analyzer` in this PR.
```suggestion
Option Name: `enable_platform_analyzer_on_pre_net5`
```

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -28,6 +28,11 @@ internal static partial class EditorConfigOptionNames
         /// </summary>
         public const string ExcludeAsyncVoidMethods = "exclude_async_void_methods";
 
+        /// <summary>
+        /// Boolean option to enable platform commpatibility analyzer for TFMs with lower version than net5.0 (https://docs.microsoft.com/visualstudio/code-quality/ca1416).
+        /// </summary>
+        public const string EnablePlatformAnalyzer = "enable_platform_analyzer";
```
mavasani, line 34: Should likely also rename this field if change the option name.

mavasani (approved):
Minor suggestions, but LGTM

docs/Analyzer Configuration.md
buyaa-n: Good catch, @mavasani JFYI the path copied from the above rules, all are having `en-us`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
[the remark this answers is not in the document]
buyaa-n: Good point, i better to account at least 2 digits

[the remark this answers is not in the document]
buyaa-n: as for valid TFM there is a char expected after `net5` i will change the condition on row 177 into `tfmString.Length >= 5`

docs/Analyzer Configuration.md
mavasani: Thanks, I'll create a separate PR to bulk remove `/en-us` from this page.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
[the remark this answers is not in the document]
mavasani: > there is a char expected after net5

I thought just `net5` is a valid TFM value, so you need to allow for length = 4?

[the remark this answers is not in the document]
buyaa-n: Yeah seems it is valid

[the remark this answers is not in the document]
buyaa-n: The problem is there is .Net framework versions that are having 2 or more digits but accounted as lower than 5.0. like `net11`. See https://docs.microsoft.com/en-us/dotnet/standard/frameworks. I think i will keep it as is for now: 1st number >= 5

image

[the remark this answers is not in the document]
Youssef1313: If net11 refers to .NET Framework 1.1, then I wonder what will happen when we get to the "unified" .NET 11? (still a lot of years though) 😄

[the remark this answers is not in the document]
buyaa-n: hopefully, that time `.NET Framework` versions will be completely deprecated/removed

docs/Analyzer Configuration.md
buyaa-n: Yeah i have considered naming something like that and thought it is a too long name for devs to memorize, but after looking naming for other options seems we try to give a more descriptive name than keeping it short, will update

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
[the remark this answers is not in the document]
mavasani: Sounds good @buyaa-n

PR #4109, opened 2020-09-01
Some security analyzer tests for #4107

dotpaul:

PR #4158, opened 2020-09-14
CA5377 shouldn't warn when unable to get control flow graph

dotpaul:
As per https://github.com/dotnet/roslyn-analyzers/pull/4109#discussion_r481397796


mavasani (approved):

dotpaul:
Thank you, @mavasani !

PR #4159, opened 2020-09-15
Update RS1025 to allow empty Initialize method

### LINKED ISSUE #4101: RS1025 (Configure generated code analysis) false positive for empty Initialize method
### Analyzer

**Diagnostic ID**: [RS1025](https://docs.microsoft.com/visualstudio/code-quality/rs1025): **Configure generated code analysis**

### Analyzer source

**Version**: v3.3.0 (Latest)

### Describe the bug

RS1025 should not be reported for empty `Initialize` methods, e.g.:

Evangelink:
Fix #4101


mavasani (changes requested):
As per https://github.com/dotnet/roslyn-analyzers/issues/4101#issuecomment-692820153. I believe this is not a false positive, such stub analyzers add unnecessary performance overhead with no overall user value add. Analyzer should be removed from the analyzer package.

PR #4160, opened 2020-09-15
Update RS1026 to allow empty Initialize method

### LINKED ISSUE #4102: RS1026 (Enable concurrent execution) false positive for empty Initialize method
### Analyzer

**Diagnostic ID**: [RS1026](https://docs.microsoft.com/visualstudio/code-quality/rs1026): **Enable concurrent execution**

### Analyzer source

**Version**: v3.3.0 (Latest)

### Describe the bug

RS1026 should not be reported for empty `Initialize` methods, e.g.:

🔗 See also #4101

Evangelink:
Fix #4102


mavasani (changes requested):
As per https://github.com/dotnet/roslyn-analyzers/issues/4101#issuecomment-692820153. I believe this is not a false positive, such stub analyzers add unnecessary performance overhead with no overall user value add. Analyzer should be removed from the analyzer package.

PR #4163, opened 2020-09-15
Clarifying titles / messages

### LINKED ISSUE #3936: CA3075 does not explain why the given member is unsafe
[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers) 3.3.0-beta2.final, [CA3075](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca3075)

No explanation is given.

Also, there are two grammar issues. The sentence doesn't have an end mark, and it uses a comma to join two independent clauses without following the comma with a conjunction like the word 'and.'

image

It took some digging to find some proper documentation in https://github.com/dotnet/roslyn-analyzers/issues/562:

> This API internally enables DTD processing on the XML reader instance used and uses UrlResolver for resolving external XML entities. The outcome is information disclosure. Content from file system or network shares for the machine processing the XML can be exposed to attacker. In addition, an attacker can use this as a DoS vector.

dotpaul:
Fixes #3936

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
@@ -1012,22 +1012,22 @@
     <value>Processing XML from untrusted data may load dangerous external references, which should be restricted by using an XmlReader with a secure resolver or with DTD processing disabled.</value>
   </data>
   <data name="UseXmlReaderForDataSetReadXml" xml:space="preserve">
-    <value>Use XmlReader For DataSet Read Xml</value>
+    <value>Use XmlReader for DataSet.ReadXml()</value>
   </data>
   <data name="UseXmlReaderForDeserialize" xml:space="preserve">
-    <value>Use XmlReader For Deserialize</value>
+    <value>Use XmlReader for XmlSerializer.Deserialize()</value>
   </data>
   <data name="UseXmlReaderForSchemaRead" xml:space="preserve">
-    <value>Use XmlReader For Schema Read</value>
+    <value>Use XmlReader for XmlSchema.Read()</value>
   </data>
   <data name="UseXmlReaderForValidatingReader" xml:space="preserve">
-    <value>Use XmlReader For Validating Reader</value>
+    <value>Use XmlReader for XmlValidatingReader constructor</value>
   </data>
   <data name="UseXmlReaderForXPathDocument" xml:space="preserve">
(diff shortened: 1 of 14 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1012,22 +1012,22 @@
...
-    <value>Use XmlReader For Validating Reader</value>
+    <value>Use XmlReader for XmlValidatingReader constructor</value>
   </data>
   <data name="UseXmlReaderForXPathDocument" xml:space="preserve">
-    <value>Use XmlReader For XPathDocument</value>
+    <value>Use XmlReader for XPathDocument constructor</value>
   </data>
   <data name="UseXmlReaderMessage" xml:space="preserve">
-    <value>This overload of the {0}.{1} method is potentially unsafe, use an overload that takes a XmlReader instance instead</value>
+    <value>This overload of the {0}.{1} method is potentially unsafe. It may enable Document Type Definition (DTD) which can be vulnerable to denial of service attacks. or might use an XmlResolver which can be vulnerable to information disclosure. Use an overload that takes a XmlReader instance instead, with DTD processing disabled and no XmlResolver.</value>
```
mavasani, line 1030:
```suggestion
    <value>This overload of the {0}.{1} method is potentially unsafe. It may enable Document Type Definition (DTD) which can be vulnerable to denial of service attacks, or might use an XmlResolver which can be vulnerable to information disclosure. Use an overload that takes a XmlReader instance instead, with DTD processing disabled and no XmlResolver.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1012,22 +1012,22 @@
     <value>Processing XML from untrusted data may load dangerous external references, which should be restricted by using an XmlReader with a secure resolver or with DTD processing disabled.</value>
   </data>
   <data name="UseXmlReaderForDataSetReadXml" xml:space="preserve">
-    <value>Use XmlReader For DataSet Read Xml</value>
+    <value>Use XmlReader for DataSet.ReadXml()</value>
```
mavasani, line 1015: Our strings are not consistent, but ideally all API references should use single quotes around them.

mavasani (approved):

dotpaul:
Thanks @mavasani !

PR #4185, opened 2020-09-21
Store the global flow values per operation as a post-step

### LINKED ISSUE #4182: PlatformCompatibilityAnalyzer seems to prevent compilation from completing
### Analyzer

**Diagnostic ID**: I think this is the PlatformCompatibilityAnalyzer but I'm not sure as the build never complete

### Analyzer source

**Version**: 5.0.100-rc.1.20452.10

### Describe the bug

The build never completes. It uses lots of CPU and the memory keeps increasing (> 2GB) when I add the `[SupportedOSPlatform]` attribute in the following code.

image

### Steps To Reproduce

Compile the following code using `dotnet build`:

````csharp
#nullable disable
using System;
using System.Diagnostics;

namespace Repro
{
    public static partial class ProcessExtensions
    {
        public static int? GetParentProcessId(this Process process)
        {
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
            {
                foreach (var entry in GetProcesses())
                {
                }
            }
            else
            {
                string line = "";
                while (true)
                {
                    if (line.StartsWith("", StringComparison.Ordinal))
                    {
                        if (true)
                        {
                        }
                    }
                }
            }

            return null;
        }

        public static IEnumerable<int> GetProcesses() => throw null;
    }
}
````

````xml
  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>
</Project>
````

### Expected behavior

The compilation should complete

### Actual behavior

The compilation doesn't complete.

- If you remove the `[SupportedOSPlatform("windows")]`, the code compiles.
- If you remove a statement from the method, it compiles.

mavasani:
Fixes #4182
Verified that the added unit test causes an infinite loop in flow analysis prior to this fix.

Analyzers executing global flow analysis need the global flow state values at each operation in the flow graph. Currently, we store these global values into the operation state map while performing analysis, but this seems to lead to incorrect analysis results. As these global analysis values per operation are not required by the flow analysis itself, we now store these values in separate global value map during analysis. These values will be stored onto the final result after analysis, so the analyzers will see only global flow state values in the result.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
@@ -52,31 +52,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                     return GlobalFlowStateAnalysisValueSet.Unknown;
                 }
 
-                return GetValueOrDefault(value);
-            }
-
-            public override GlobalFlowStateAnalysisValueSet VisitPropertyReference(IPropertyReferenceOperation operation, object? argument)
-            {
-                return GetValueOrDefault(base.VisitPropertyReference(operation, argument));
-            }
-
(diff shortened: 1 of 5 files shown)


buyaa-n (approved):
Looks good, thank you for the quick fix!

PR #4196, opened 2020-09-21
Fix DataflowOperationVisitor.VisitUnaryOperator

### LINKED ISSUE #4190: PlatformCompatibilityAnalyzer: Assertions should work the same way as nullable
Consider:

```C#
string? foo = SomeFunction();
Debug.Assert(foo != null);
// After this point, foo is considered non-null
foo.ToString(); // Does not trigger a nullable warning
```

```C#
Debug.Assert(OperatingSystem.IsWindows());
new SomeUnsupportedAPI(); // Results in a warning.
```

mavasani:
Default the Visit method to return the value for the operand. Fixes #4190


buyaa-n (approved):
Looks great, thank you! 
We might want it in RC2. CC @jeffhandley

mavasani:
Thanks @buyaa-n - retargeted.

jeffhandley (approved):
LGTM.  And if we're able to include this in RC2, that would be helpful.

PR #4199, opened 2020-09-22
[2.9.x] Move to latest CodeStyle package and fix violations

mavasani:
NOTE: Developers will require to be on VS2019 16.8 Preview3 or later after this change

### DIFF
--- a/src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
@@ -39,14 +39,11 @@ namespace Microsoft.CodeAnalysis
                     return false;
 
                 case ReportDiagnostic.Warn:
-                    switch (other)
+                    return other switch
                     {
-                        case ReportDiagnostic.Error:
-                            return true;
-
-                        default:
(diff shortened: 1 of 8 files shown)


src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
```diff
@@ -39,14 +39,11 @@ public static bool IsLessSevereThan(this ReportDiagnostic current, ReportDiagnos
                     return false;
 
                 case ReportDiagnostic.Warn:
-                    switch (other)
+                    return other switch
```
Evangelink, line 42: Isn't it better to use a ternary operator here?

Evangelink (approved):

src/Utilities/Compiler/Extensions/ReportDiagnosticExtensions.cs
mavasani: Possibly, though I just applied the suggestions that were presented by the IDE rules. I can clean this up in a follow-up change for the whole repo.

mavasani:
Thank you @Evangelink!

PR #4201, opened 2020-09-23
V3 publishing 

epananth:
V3 publishing 

with default channel : https://dnceng.visualstudio.com/internal/_build/results?buildId=826687&view=results
without default channel : https://dnceng.visualstudio.com/internal/_build/results?buildId=826493&view=results


jmarolf:
Looking at https://github.com/dotnet/arcade/issues/4283 this is a welcome change. less churn in taking arcade updates to publish to different branches. I will verify that new signed builds are getting inserted into rc2 after this is merged.

jmarolf (approved):

PR #4204, opened 2020-09-23
Refactor code to rely on CreateDiagnostic helper

Evangelink:


mavasani:
@Evangelink there are test failures

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DataSetDataTableInWebSerializableObjectGraphTests.cs
```diff
@@ -31,7 +30,7 @@ public string MyWebMethod(DataTable dataTable)
     }
 }
 ",
-                GetCSharpResultAt(10, 31, "DataTable", "DataTable"));
+                GetCSharpResultAt(10, 41, "DataTable", "DataTable"));
```
Evangelink, line 33: With the old technique, the first column of the location was the data-type, now we "correctly" report on the parameter.

mavasani: Tagging @dotpaul as this is a security rule.

src/Utilities/Compiler/Extensions/DiagnosticExtensions.cs
```diff
@@ -144,6 +154,29 @@ internal static class DiagnosticExtensions
                      messageArgs: args);
         }
 
+        public static Diagnostic CreateDiagnostic(
+            this SyntaxReference syntaxReference,
```
mavasani, line 158: I don't really like all these SyntaxReference extension methods, seem to be very specific to a set of internal analyzers in Roslyn.Diagnostic.Analyzers. Maybe it is better to move all of them to just that project inside a separate helper type?

Evangelink: I am fine with this.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DataSetDataTableInWebSerializableObjectGraphTests.cs
dotpaul: In this case, it's the type that's the issue, so I think I prefer the older technique's location.

Also, could you make all the security rule changes in the 2.9.x branch please?

Evangelink: Sure for the 2.9.x and regarding the location I wasn't sure as it was originally reporting on the full parameter (type + var name), I'll change it.

mavasani (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DataSetDataTableInWebSerializableObjectGraphTests.cs
dotpaul: Thanks @Evangelink !

PR #4207, opened 2020-09-23
Simplify wildcard matching algorithm

### LINKED ISSUE #3403: Consider revisiting or improving the wildcard matching algorithm
1. Consider revisiting this code for code refactoring/cleanup for easier future maintenance.

2. Consider tuning the cache size and hit/misses here to improve performance.

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/pull/3233_

Evangelink:
Fix #3403


src/Utilities.UnitTests/Options/SymbolNamesWithValueOptionTests.cs
```diff
@@ -271,46 +271,41 @@ public class MyClass {}
...
-        [InlineData("MyOuterClass*", "MyInnerClass")]
-        [InlineData("My*", "MyCompany")]
-        [InlineData("My*", "MyOuterClass")]
-        [InlineData("My*", "MyInnerClass")]
-        [InlineData("My*", "MyField")]
-        [InlineData("My*", "MyProperty")]
-        [InlineData("My*", "MyEvent")]
-        [InlineData("My*", "MyMethod")]
-        [InlineData("My*", "MyMethod2")]
+        // Fully qualified name without prefix
```
Evangelink, line 296: I have kept these for compatibility with the old algorithm but we might want to drop them as they imply a more heavy computation.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -271,240 +270,73 @@ private bool TryGetFirstWildcardMatch(ISymbol symbol, out KeyValuePair<string, T
...
-                    return !firstMatch.Equals(NoWildcardMatch);
-                }
-
-                // Is there any wildcard entry for this symbol kind?
-                if (!wildcardNamesBySymbolKind.ContainsKey(kind))
-                {
-                    firstMatch = NoWildcardMatch;
-                    return false;
-                }
+                var declarationIdWithoutPrefix = DocumentationCommentId.CreateDeclarationId(symbol)[2..];
```
Evangelink, line 332: I couldn't find any better way. We get called for each symbol walking up on the inheritance chain but we need the fully qualified part every time.

mavasani (approved):

PR #4211, opened 2020-09-24
Update CA1826 to add user option to exclude OrDefault methods

### LINKED ISSUE #1817: "Do not use Enumerable methods" encourages degrading readability
Analyzer package: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)
Package Version: 2.6.2-beta2
Diagnostic ID: CA1826

"Do not use Enumerable methods on indexable collections. Instead use the collection directly" makes sense for .First() and .Last(), but not for .FirstOrDefault() or .LastOrDefault(). There's no readable replacement for those methods; unless FirstOrDefault is showing up in a performance profiling tool, readability is more important.

Even .Last() is a stretch since the replacement is `list[list.Count - 1]`. That's a big drop in readability.

Evangelink:
Fix #1817

### DIFF
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -815,7 +815,7 @@ Default Value: `false`
 
 Example: `dotnet_code_quality.CA1051.exclude_structs = true`
 
-### Exclude OrDefault methods
+### Exclude FirstOrDefault and LastOrDefault methods
 
 Option Name: `exclude_ordefault_methods`
 
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -773,3 +773,15 @@ Option Values: Boolean values
 Default Value: `false`
 
 Example: `dotnet_code_quality.CA1051.exclude_structs = true`
+
+### Exclude OrDefault methods
+
+Option Name: `exclude_ordefault_methods`
+
+Configurable Rules: [CA1826](https://docs.microsoft.com/visualstudio/code-quality/CA1826)
+
+Option Values: Boolean values
+
+Default Value: `false`
+
+Example: `dotnet_code_quality.CA1826.exclude_ordefault_methods = true`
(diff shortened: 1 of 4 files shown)


docs/Analyzer Configuration.md
```diff
@@ -773,3 +773,15 @@ Option Values: Boolean values
 Default Value: `false`
 
 Example: `dotnet_code_quality.CA1051.exclude_structs = true`
+
+### Exclude OrDefault methods
```
mavasani, line 777: Can we add some additional description here for what exactly is being excluded? At least we should give an example for clarity.

src/Utilities/Compiler/Options/EditorConfigOptionNames.cs
```diff
@@ -198,5 +198,10 @@ internal static partial class EditorConfigOptionNames
         /// Boolean option to configure whether to exclude structs when considering public fields.
         /// </summary>
         public const string ExcludeStructs = "exclude_structs";
+
+        /// <summary>
+        /// Boolean option to configure whether to exclude 'OrDefault' methods.
```
mavasani, line 203: Same here, additional context would be helpful in this comment.

docs/Analyzer Configuration.md
mavasani: Maybe the title itself can be `Exclude FirstOrDefault and LastOrDefault methods` to match analyzer implementation?

mavasani (approved):
Thanks!

orchidalloy:
How do I apply this?

UmairB:
Yeah, would be great to have an example editorconfig

Evangelink:
@orchidalloy @UmairB You can add [the following line](https://github.com/dotnet/roslyn-analyzers/pull/4211/files#diff-13786715c035d15d5764ff035bd6369f4a5b617aa0016695821abf05d7ed802fR828): `dotnet_code_quality.CA1826.exclude_ordefault_methods = true`

PR #4212, opened 2020-09-24
Handle expression-bodied method in code-fixes for RS1025 and RS1026

### LINKED ISSUE #2698: RS1025 and RS1026 Rules Do Not Handle Expression-Bodied Members Correctly
**Version Used**: 
VS 2019

**Steps to Reproduce**:
1. Create an analyzer, and override `Initialize()`
2. Add a call to `RegisterSyntaxNodeAction`() with `Initialize()` as an expression-bodied member
3. Note that RS1025 and RS1026 will show up in the Error List
4. Try to apply the code fixes.

**Expected Behavior**:
The code is added by changing the method body for `Initialize()` to a block body and then adding the right call to `context`.

**Actual Behavior**:
The analyzer changes the body to a block body, but it overwrites the call to `RegisterSyntaxNodeAction()`. This is incorrect. Note that if the method body was originally a block body, the code fix works correctly.

Evangelink:
Fix #2698

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/ConfigureGeneratedCodeAnalysisFix.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/ConfigureGeneratedCodeAnalysisFix.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Composition;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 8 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/ConfigureGeneratedCodeAnalysisFix.cs
```diff
@@ -65,7 +63,7 @@ private static async Task<Document> ConfigureGeneratedCodeAnalysisAsync(Document
                 return document;
             }
 
-            var statements = generator.GetStatements(methodDeclaration);
+            var statements = GetStatements(methodDeclaration);
```
Evangelink, line 66: https://github.com/dotnet/roslyn/issues/48009

sharwell (approved):

PR #4213, opened 2020-09-24
Add null checks for access to ApplicationSyntaxReference

### LINKED ISSUE #1083: NullReference exception running
#### Analyzer package

Microsoft.Composition.Analyzers

#### Analyzer

DoNotMixAttributesFromDifferentVersionsOfMEFAnalyzer

#### Repro steps

I couldn't get a simplified repro here and mine is a bit convoluted (see below for the debug info though)

1. Open solution from https://github.com/dotnet/roslyn-project-system
2. Add a reference to Microsoft.VisualStudio.ProjectSystem.VS.Implementation
3. Add a class that consumes anything from that assembly

#### Expected behavior

Build successful with whatever message the analyzer has.

#### Actual behavior

Build error:

Error	AD0001	Analyzer threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.	Microsoft.VisualStudio.ProjectSystem.Managed.VS	E:\dotnet\roslyn-project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\CSC	1	N/A

Debugging through it the null reference comes from here:

private static void ReportDiagnostic(SymbolAnalysisContext symbolContext, INamedTypeSymbol exportedType, AttributeData problematicAttribute)
        {
            // Attribute '{0}' comes from a different version of MEF than the export attribute on '{1}'
            var diagnostic = Diagnostic.Create(Rule, problematicAttribute.ApplicationSyntaxReference.GetSyntax().GetLocation(), problematicAttribute.AttributeClass.Name, exportedType.Name);
            symbolContext.ReportDiagnostic(diagnostic);
        }

problematicAttribute.ApplicationSyntaxReference is null, presumably because the issue is flagged from the referenced assembly and not the local sources. Looks like that needs to be null checked and defaulted to Location.None if there will be cases where that property can be null.

Evangelink:
Fix #1083


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerAttributeAnalyzer.cs
```diff
@@ -115,7 +115,7 @@ protected override void AnalyzeDiagnosticAnalyzer(SymbolAnalysisContext symbolCo
                             }
                         }
 
-                        attributeSyntax = attribute.ApplicationSyntaxReference.GetSyntax(symbolContext.CancellationToken);
+                        attributeSyntax = attribute.ApplicationSyntaxReference?.GetSyntax(symbolContext.CancellationToken);
```
mavasani, line 118: This seems redundant given the null check on line 96 above.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumWithFlagsAttribute.Fixer.cs
```diff
@@ -75,9 +75,11 @@ private static SyntaxNode RemoveFlagsAttribute(SyntaxGenerator generator, Semant
             }
 
             AttributeData flagsAttribute = enumType.GetAttributes().First(a => Equals(a.AttributeClass, flagsAttributeType));
-            SyntaxNode attributeNode = flagsAttribute.ApplicationSyntaxReference.GetSyntax(cancellationToken);
+            SyntaxNode? attributeNode = flagsAttribute.ApplicationSyntaxReference?.GetSyntax(cancellationToken);
```
mavasani, line 78: Can this ever be null given that FlagsAttribute has `Inherited = false`: https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=netcore-3.1?

Evangelink: Good remark, I went defensive without asking me much question...

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumWithFlagsAttribute.Fixer.cs
```diff
@@ -75,9 +75,11 @@ private static SyntaxNode RemoveFlagsAttribute(SyntaxGenerator generator, Semant
             }
 
             AttributeData flagsAttribute = enumType.GetAttributes().First(a => Equals(a.AttributeClass, flagsAttributeType));
-            SyntaxNode attributeNode = flagsAttribute.ApplicationSyntaxReference.GetSyntax(cancellationToken);
+            SyntaxNode? attributeNode = flagsAttribute.ApplicationSyntaxReference.GetSyntax(cancellationToken);
```
mavasani, line 78: Do we have a repro case where this is null? I am not sure if this can ever be null.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ProvideObsoleteAttributeMessage.cs
```diff
@@ -75,8 +75,11 @@ private static void AnalyzeSymbol(SymbolAnalysisContext context, INamedTypeSymbo
                     if (attribute.ConstructorArguments.IsEmpty ||
                         string.IsNullOrEmpty(attribute.ConstructorArguments.First().Value as string))
                     {
-                        SyntaxNode node = attribute.ApplicationSyntaxReference.GetSyntax(context.CancellationToken);
-                        context.ReportDiagnostic(node.CreateDiagnostic(Rule, context.Symbol.Name));
+                        SyntaxNode? node = attribute.ApplicationSyntaxReference?.GetSyntax(context.CancellationToken);
+                        if (node != null)
```
mavasani, line 79: This doesn't seem correct. We should not bail out on reporting diagnostics based on lack of syntax. Couple of questions:

1. Do we know if null is possible? As in, does `ISymbol.GetAttributes` return attributes inherited from base types, which can be in metadata? I thought it only returns attributes applied directly to the symbol, and given the analyzer callbacks only happen for source symbols, all attributes should be in source. If so, null value is not possible.
2. If null is possible, can we find an alternate syntax to report diagnostic (say symbol localtion), instead of bailing out?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PInvokeDiagnosticAnalyzer.cs
```diff
@@ -209,9 +209,8 @@ public void AnalyzeSymbol(SymbolAnalysisContext context)
             }
 
             private static Location GetAttributeLocation(AttributeData attributeData)
-            {
-                return attributeData.ApplicationSyntaxReference.SyntaxTree.GetLocation(attributeData.ApplicationSyntaxReference.Span);
-            }
+                => attributeData.ApplicationSyntaxReference?.SyntaxTree.GetLocation(attributeData.ApplicationSyntaxReference.Span)
```
mavasani, line 212: Same here, would like to know when null is possible.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Resources/MarkAssembliesWithNeutralResourcesLanguage.cs
```diff
@@ -73,7 +73,10 @@ public override void Initialize(AnalysisContext analysisContext)
                     if (data != null)
                     {
                         // we have the attribute but its doing it wrong.
-                        ce.ReportDiagnostic(data.ApplicationSyntaxReference.GetSyntax(ce.CancellationToken).CreateDiagnostic(Rule));
+                        if (data.ApplicationSyntaxReference != null)
```
mavasani, line 76: Based on the implementation of `TryCheckNeutralResourcesLanguageAttribute`, I think this attribute must be in source, hence this cannot be null. Also bailing out on null syntax seems incorrect.

mavasani (changes requested):
Let's try to identify the cases where `ApplicationSyntaxReference` is possibly coming from a metadata attribute and null checks for those, instead of doing so for all places it is used.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumWithFlagsAttribute.Fixer.cs
Evangelink: If I have correctly understood your comment about base type, then for an enum it seems impossible so we don't need to check this. Is it worth adding a `RoslynDebug.Assert`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ProvideObsoleteAttributeMessage.cs
Evangelink: There is not much doc available for `GetAttributes` so that's pretty hard to say but I would say that if this is true then I don't get the issue described in the ticket.

mavasani: @Evangelink The issue specifically identified an NRE in DoNotMixAttributesFromDifferentVersionsOfMEFAnalyzer, which seems to walk the base type chain to get all applicable attributes - so for that analyzer the issue is certainly possible. We should probably narrow down this PR to fix just the cases that use GetApplicableAttributes? Also the fix should not be to bail out, but instead to report on an alternate, reasonable syntax.

src/Roslyn.Diagnostics.Analyzers/Core/DoNotMixAttributesFromDifferentVersionsOfMEF.cs
```diff
@@ -116,10 +116,17 @@ private static void AnalyzeSymbol(SymbolAnalysisContext symbolContext, IEnumerab
 
         private static void ReportDiagnostic(SymbolAnalysisContext symbolContext, INamedTypeSymbol exportedType, AttributeData problematicAttribute)
         {
-            // Attribute '{0}' comes from a different version of MEF than the export attribute on '{1}'
-            var diagnostic = problematicAttribute.ApplicationSyntaxReference.CreateDiagnostic(
-                Rule, symbolContext.CancellationToken, problematicAttribute.AttributeClass.Name, exportedType.Name);
-            symbolContext.ReportDiagnostic(diagnostic);
+            if (problematicAttribute.ApplicationSyntaxReference == null)
+            {
+                symbolContext.ReportDiagnostic(symbolContext.Symbol.CreateDiagnostic(Rule, problematicAttribute.AttributeClass.Name, exportedType.Name));
```
Evangelink, line 121: IMO, this call should not fail.

mavasani (approved):
Thanks @Evangelink!

PR #4215, opened 2020-09-24
Update RS1008 to stop reporting on nested struct

### LINKED ISSUE #845: make RS1008 to ignore nested struct as well
right now, if nested class in diagnostic analyzer has roslyn symbols as field, we don't issue RS1008 but it looks like we missed struct case.

Evangelink:
Fix #845

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.SyntaxNodeWithinAnalyzerTypeCompilationAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.SyntaxNodeWithinAnalyzerTypeCompilationAnalyzer.cs
@@ -41,7 +41,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
                 foreach (TClassDeclarationSyntax classDecl in classDecls)
                 {
                     IEnumerable<TSyntaxNodeOfInterest> syntaxNodes =
-                        classDecl.DescendantNodes(n => n is not TClassDeclarationSyntax && n is not TStructDeclarationSyntax ||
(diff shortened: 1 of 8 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerCorrectnessAnalyzer.SyntaxNodeWithinAnalyzerTypeCompilationAnalyzer.cs
```diff
@@ -39,7 +40,10 @@ protected override void AnalyzeDiagnosticAnalyzer(SymbolAnalysisContext symbolCo
                 IEnumerable<TClassDeclarationSyntax> classDecls = GetClassDeclarationNodes(namedType, symbolContext.CancellationToken);
                 foreach (TClassDeclarationSyntax classDecl in classDecls)
                 {
-                    IEnumerable<TSyntaxNodeOfInterest> syntaxNodes = classDecl.DescendantNodes(n => !(n is TClassDeclarationSyntax) || ReferenceEquals(n, classDecl)).OfType<TSyntaxNodeOfInterest>();
+                    IEnumerable<TSyntaxNodeOfInterest> syntaxNodes =
+                        classDecl.DescendantNodes(n => n is not TClassDeclarationSyntax && n is not TStructDeclarationSyntax ||
```
sharwell, line 44:
```suggestion
                        classDecl.DescendantNodes(n => n is not (TClassDeclarationSyntax or TStructDeclarationSyntax) ||
```

Evangelink: Oh that's a cool one, thanks!

mavasani (approved):
Thanks!

PR #4218, opened 2020-09-24
Switch platform compat analyzer to do an optimistic flow analysis

### LINKED ISSUE #4209: PlatformCompatibilityAnalyzer returns spurious errors in AspNetCore sources
* Clone the AspNetCore repo
* Run `restore.cmd`
* cd src/SignalR
* startvs.cmd
* Open `WebSocketsTransport.cs`
* Remove the suppression: https://github.com/dotnet/aspnetcore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/Internal/WebSocketsTransport.cs#L145
* Notice that the analyzer complains about unsupported OS platform.

mavasani:
Fixes #4209
Our flow analyses take a boolean parameter `pessimisticAnalysis` that determines the state change to be done when we encounter invocation to an unknown lambda target. Currently, the analysis was pessimistically reseting the state to unknown, but now we optimistically assume such a call does not change current flow analysis state.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -235,6 +235,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                     var analysisResult = GlobalFlowStateAnalysis.TryGetOrComputeResult(
                         cfg, context.OwningSymbol, CreateOperationVisitor, wellKnownTypeProvider,
                         context.Options, SupportedOsRule, performValueContentAnalysis,
+                        pessimisticAnalysis: false,
                         context.CancellationToken, out var valueContentAnalysisResult,
                         additionalSupportedValueTypes: osPlatformTypeArray,
                         getValueContentValueForAdditionalSupportedValueTypeOperation: GetValueContentValue);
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -235,6 +235,7 @@ private static bool PlatformAnalysisAllowed(AnalyzerOptions options, Compilation
                     var analysisResult = GlobalFlowStateAnalysis.TryGetOrComputeResult(
                         cfg, context.OwningSymbol, CreateOperationVisitor, wellKnownTypeProvider,
                         context.Options, SupportedOsRule, performValueContentAnalysis,
+                        pessimisticAnalysis: false,
```
mavasani, line 238: This is the core change needed to fix the bug. @buyaa-n would you be able to help validating if this fixes the aspnet core repo issue that @pranavkm was hitting by building locally with this change?

buyaa-n: Sure, i can validate it

buyaa-n: Yes, it was fixed with this update, thanks @mavasani

buyaa-n (approved):
Looks great, thanks!

PR #4222, opened 2020-09-25
Use tree.DiagnosticOptions workaround only when needed

Evangelink:


src/Utilities/Compiler/Extensions/DiagnosticExtensions.cs
```diff
@@ -184,15 +188,22 @@ internal static class DiagnosticExtensions
...
                 }
+#endif
 
                 ReportDiagnostic? overriddenSeverity = null;
                 foreach (var tree in compilation.SyntaxTrees)
                 {
-                    var options = (ImmutableDictionary<string, ReportDiagnostic>)s_syntaxTreeDiagnosticOptionsProperty.GetValue(tree)!;
+                    var options =
+#if CODEANALYSIS_V3_OR_BETTER
+                        tree.DiagnosticOptions;
```
mavasani, line 203: SyntaxTree.DiagnosticOptions has actually been deprecated in future Roslyn versions and always returns empty options. I need to fix up the logic here.

Evangelink: https://github.com/dotnet/roslyn-analyzers/issues/4224

PR #4223, opened 2020-09-25
Update ISymbol.IsReadOnly() to handle methods and types

Evangelink:


src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -747,18 +747,13 @@ public static bool IsConst([NotNullWhen(returnValue: true)] this ISymbol? symbol
...
+            => symbol switch
             {
                 IFieldSymbol field => field.IsReadOnly,
-
                 IPropertySymbol property => property.IsReadOnly,
-
-                // TODO: IMethodSymbol and ITypeSymbol also have IsReadOnly in Microsoft.CodeAnalysis 3.x
-                //       Add these cases once we move to the required Microsoft.CodeAnalysis.nupkg.
-
+                IMethodSymbol method => method.IsReadOnly,
```
mavasani, line 754: You may need to wrap this around preprocessor symbol checks for code analysis V3 or later

Evangelink: Oh didn't see this was failing. That makes sense, some analyzers are still based on old version of roslyn.

mavasani (approved):

PR #4226, opened 2020-09-25
Update CA2241 to handle cases from IDE0043

### LINKED ISSUE #1254: CA2241 Does not fire on skipped placeholder sequence ({0} {1} {3}).
#### Analyzer package

System.Runtime.Analyzers

#### Analyzer

ProvideCorrectArgumentsToFormattingMethods

#### Repro steps

`string.Format(CultureInfo.InvariantCulture, "Abce '{0}' efg '{1}' hij'{3}'.", p0, p1, p2));`

#### Expected behavior

Warning

#### Actual behavior

No Warning

Evangelink:
Fix #1254


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -41,45 +42,61 @@ public override void Initialize(AnalysisContext analysisContext)
...
+                    if (expectedStringFormatArgumentCount > 0)
                     {
-                        // __arglist is not supported here
-                        if (invocation.TargetMethod.IsVararg)
+                        var missingIndexes = Enumerable.Range(0, stringFormatIndexes.Max())
+                            .Except(stringFormatIndexes);
+                        if (missingIndexes.Any())
                         {
-                            // can't deal with this for now.
+                            context.ReportDiagnostic(invocation.CreateDiagnostic(Rule));
```
Evangelink, line 86: We may want to report with a different diagnostic (message + desc).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -41,45 +42,61 @@ public override void Initialize(AnalysisContext analysisContext)
...
 
-                    StringFormatInfo.Info? info = formatInfo.TryGet(invocation.TargetMethod, operationContext);
+                    StringFormatInfo.Info? info = formatInfo.TryGet(invocation.TargetMethod, context);
                     if (info == null || invocation.Arguments.Length <= info.FormatStringIndex)
                     {
                         // not a target method
                         return;
                     }
 
+                    if (info.ExpectedStringFormatArgumentCount >= 0 &&
```
Evangelink, line 60: Moved to exit before doing the parsing of the string as we don't need the value.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -63,48 +137,106 @@ public class C
...
-        Console.Write("""", 1);
-        Console.Write(""{0}"", 1, 2);
-        Console.Write(""{0} {1}"", 1, 2, 3);
-        Console.Write(""{0} {1} {2}"", 1, 2, 3, 4);
-        Console.Write(""{0} {1} {2} {3}"", 1, 2, 3, 4, 5);
+        {|#0:" + invocation + @"(""{0} {1}"", 1)|};
+        {|#1:" + invocation + @"(""{0} {1} {2}"", 1)|};
+        {|#2:" + invocation + @"(""{0} {1} {2}"", 1, 2)|};
+
+        // These cases are not handled as we cannot find a parameter named format
```
Evangelink, line 144: I was tempted to update the code to handle these cases but I was a bit worried about the perf impact of analyzing the extra overloads. Let me know what you think.

mavasani:
@jmarolf You have the most context on this analyzer. Can you please review?

Evangelink:
@mavasani @jmarolf I think I need to rewrite the analyzer to use the regex as suggested in the ticket but I am fine with having a first review to start with, it shouldn't change much the main portion of the code.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
jmarolf: I think, since we are saying we want to break with legacy FxCop behavior that introducing a new, more specific warning message would be nice.

---
In reply to: [495128627](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r495128627) [](ancestors = 495128627)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -114,13 +133,13 @@ public override void Initialize(AnalysisContext analysisContext)
...
                     }
                 }, OperationKind.Invocation);
             });
         }
 
-        private static int GetFormattingArguments(string format)
+        private static HashSet<int>? GetStringFormatItemIndexes(string format)
         {
             // code is from mscorlib
             // https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/Text/StringBuilder.cs#L1312
```
jmarolf, line 145: Can we update this link to point to the current implementation [here](https://github.com/dotnet/runtime/blob/f1e131a4bef5bd5373a3dab65523851b54a94306/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs#L1538) now that we've migrated repos?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -27,34 +27,108 @@ public class C
...
-
-        IFormatProvider p = null;
-        var f = String.Format(p, """", 1);
-        var g = String.Format(p, ""{0}"", 1, 2);
-        var h = String.Format(p, ""{0} {1}"", 1, 2, 3);
-        var i = String.Format(p, ""{0} {1} {2}"", 1, 2, 3, 4);
+        {|#0:" + invocation + @"("""", 1)|};
+        {|#1:" + invocation + @"(""{0}"", 1, 2)|};
+        {|#2:" + invocation + @"(""{0} {1}"", 1, 2, 3)|};
+        {|#3:" + invocation + @"(""{0} {1} {2}"", 1, 2, 3, 4)|};
```
jmarolf, line 33: can we add a case for `"{0} {1} {3}", 1, 2, 3`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -15,10 +15,10 @@ namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
 {
     public class ProvideCorrectArgumentsToFormattingMethodsTests
```
jmarolf, line 16: Can we add a test for `__arglist`?

jmarolf (changes requested):
looks good! I would appreciate it if we could add a few more tests while we are here and I think having a separate, more descriptive diagnostic message would be nice

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
Evangelink: Do you mean an empty arglist? We have some tests with `__arglist(5)`

jmarolf: yep

---
In reply to: [499685668](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r499685668) [](ancestors = 499685668)

Evangelink: I am not sure how to get it to work, I was expecting the following to be good but that's not the case:

```csharp
public void M(__arglist)
{
    System.Console.WriteLine("", __arglist);
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
Evangelink: I would still keep the same rule ID except if you think that's valuable to let users disable some part of the rule (I don't see the need as I don't expect much FPs).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -22,64 +24,105 @@ public class ProvideCorrectArgumentsToFormattingMethodsAnalyzer : DiagnosticAnal
...
+            DiagnosticCategory.Usage,
+            RuleLevel.BuildWarningCandidate,
+            description: s_localizableDescription,
+            isPortedFxCopRule: true,
+            isDataflowRule: false);
+
+        /// <summary>
+        /// This regex is used to remove escaped brackets from the format string before looking for valid {} pairs.
+        /// </summary>
+        private static readonly Regex s_removeEscapedBracketsRegex = new("{{");
```
Evangelink, line 54: Not sure if it's better to use the compiled regex here. It's not present in IDE0043 so I am assuming this was a conscious call.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -114,204 +157,53 @@ public override void Initialize(AnalysisContext analysisContext)
...
-            // return count of this format - {index[,alignment][:formatString]}
-            var pos = 0;
-            int len = format.Length;
-            var uniqueNumbers = new HashSet<int>();
+            var matches = s_extractPlaceholdersRegex.Matches(formatStringWithEscapedBracketsChangedToSpaces);
+            var indexes = new HashSet<int>();
 
-            // main loop
-            while (true)
+            foreach (Match match in matches)
```
Evangelink, line 176: Any preference between the explicit foreach and a `Select` call? Note that I will need to do a cast first.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -22,64 +24,105 @@ public class ProvideCorrectArgumentsToFormattingMethodsAnalyzer : DiagnosticAnal
...
+                            // provided to the invocation.
                             return;
                         }
+                    }
 
+                    // __arglist is not supported here (see https://github.com/dotnet/roslyn/issues/7346)
+                    // but we want to skip it only in C# since VB doesn't support __arglist
+                    if (info.ExpectedStringFormatArgumentCount >= 0 &&
+                        invocation.TargetMethod.IsVararg &&
+                        invocation.Language == LanguageNames.CSharp)
```
Evangelink, line 115: This filter was suggested but not present in the original implementation. Let me know if you prefer to revert this.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -22,64 +24,105 @@ public class ProvideCorrectArgumentsToFormattingMethodsAnalyzer : DiagnosticAnal
...
+                    if (expectedStringFormatArgumentCount > 0)
                     {
-                        // __arglist is not supported here
-                        if (invocation.TargetMethod.IsVararg)
+                        var missingIndexes = Enumerable.Range(0, stringFormatIndexes.Max())
+                            .Except(stringFormatIndexes);
+                        if (missingIndexes.Any())
                         {
-                            // can't deal with this for now.
+                            context.ReportDiagnostic(invocation.CreateDiagnostic(MissingFormatItemRule));
```
Evangelink, line 102: Do we want to report only on the string parameter?

Evangelink:
There is an extra user option available in [IDE0043](https://github.com/dotnet/roslyn/blob/master/src/Analyzers/Core/Analyzers/ValidateFormatString/AbstractValidateFormatStringDiagnosticAnalyzer.cs#L98-L100), do we want to integrate it?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: I would go ahead and handle them.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -342,8 +342,8 @@
   <data name="ProvideCorrectArgumentsToFormattingMethodsDescription" xml:space="preserve">
     <value>The format argument that is passed to System.String.Format does not contain a format item that corresponds to each object argument, or vice versa.</value>
   </data>
-  <data name="ProvideCorrectArgumentsToFormattingMethodsMessage" xml:space="preserve">
-    <value>Provide correct arguments to formatting methods</value>
+  <data name="ProvideCorrectArgumentsToFormattingMethodsMessageArgs" xml:space="preserve">
+    <value>Incorrect number of arguments passed to this formatting method call</value>
```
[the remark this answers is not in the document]
jmarolf: This sounds like a great idea

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1482,4 +1482,7 @@
   <data name="PreferStringContainsOverIndexOfCodeFixTitle" xml:space="preserve">
     <value>Replace with 'string.Contains'</value>
   </data>
+  <data name="ProvideCorrectArgumentsToFormattingMethodsMessageFormatItem" xml:space="preserve">
+    <value>Some string format items are missing from this formatting method call</value>
```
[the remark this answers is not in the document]
jmarolf: Considering we will "squiggle" the entire thing I think some text explaining how to fix what is wrong would help. Saying

`"Some string format items are missing from this formatting method call. Nothing was passed for index '0'"`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -22,64 +24,105 @@ public class ProvideCorrectArgumentsToFormattingMethodsAnalyzer : DiagnosticAnal
...
+            s_localizableTitle,
+            s_localizableMessageArgs,
+            DiagnosticCategory.Usage,
+            RuleLevel.BuildWarningCandidate,
+            description: s_localizableDescription,
+            isPortedFxCopRule: true,
+            isDataflowRule: false);
+
+        internal static DiagnosticDescriptor MissingFormatItemRule = DiagnosticDescriptorHelper.Create(
+            RuleId,
```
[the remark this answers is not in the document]
jmarolf: I think that users may want to configure different severities for these rules which would necessitate a different ID.

- having too few parameters, run time error, should be a warning
- having too many parameters, not a runtime error, should probably be info by default

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
jmarolf: compiling the regex has binary size implications in visual studio which we are very paranoid about. Having the regex be precompiled here is fine.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
jmarolf: I am fine with this change

jmarolf:
> There is an extra user option available in IDE0043, do we want to integrate it?

If you have the time, I would love that. I think each of these options should be a different diagnostic Id that the user can configure independently.

jmarolf:
@mavasani I would like providing too few parameters (which will cause a runtime error) to be enabled as a warning for analysis level 6. Do you have any objections?

mavasani:
> @mavasani I would like providing too few parameters (which will cause a runtime error) to be enabled as a warning for analysis level 6. Do you have any objections?

Sounds reasonable. However, lets follow a process for promoting any CA rule to a warning:

- Create a separate design issue for the proposal
- Discuss it via a design meeting with NetAnalyzers v-team OR get confirmation via email
- Open a PR to promote it as a warning
- File a doc issue on dotnet/docs to document breaking change in net6.0

jmarolf:
@mavasani sounds like a plan. Looking over my notes from the last meeting I believe this has already been approved, but I agree let's not put foist those additional requirements on this PR. I'll create a follow up PR after this is merged.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
jmarolf: I think foreach auto-casting is fine

mavasani:
>  I'll create a follow up PR after this is merged.

As part of that we should also bump the analyzer to warning in dotnet/roslyn and dotnet/runtime and confirm no false positives and acceptable performance. I think that should be our miniumn quality gate before bumping up a rule to a warning.

Evangelink:
I am working on the various open comment but I have a side question regarding the descriptors we want to create, basically we have the following use-cases:

1. Not enough args compared for well-known string format methods (overload with expected arguments)
  a. `Console.WriteLine("{0} {1}", 42);` -> causes runtime failure = Descriptor1 (e.g. `NotEnoughArgsRule`)
  b. `Console.WriteLine("{1}", 42);` -> causes runtime failure = Descriptor1 (e.g. `NotEnoughArgsRule`) but maybe also Descriptor2 (e.g. `NotEnoughArgsMissingFormatItemRule`). In this case we might want Descriptor2 to have the same ID as 1 but with a different message. I wouldn't report 2 diag on this line because we don't know what's the correct fix (i.e. it could be change `1` to `0` or add one more parameter).

2. Not enough args compared for well-known string format methods (overload without expected arguments). For example, `Console.WriteLine("{0} {1}")` or `Console.WriteLine("{1}")`, there is no runtime error so we probably want to use Descriptor3 (e.g. `FormatItemInNonStringFormatMethodRule`) and maybe Descriptor4 (e.g. `FormatItemInNonStringFormatMethodMissingFormatItemRule`) for the case with the missing `{0}`.

3. Correct number of args but missing format item (e.g. `Console.WriteLine("{1}", 1, 2)`) -> Descriptor5 (e.g. `EnoughArgsMissingFormatItemRule`).

4. Too many args for well-known string format methods. For example, `Console.WriteLine("{0}", 1, 2)` -> Descriptor6 (e.g. `TooManyArgsRule`) and for `Console.WriteLine("{1}", 1, 2, 3)` Descriptor7 (e.g. `TooManyArgsMissingFormatItemRule`).

In addition to the previous cases, there is a [user option to provide additional string format methods](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/Analyzer%20Configuration.md#additional-string-formatting-methods), shall we reuse the previous descriptors or create a new set of them? Especially for case 1, where we are not sure if these additional methods will cause the runtime error so we might not want them to be raised as `Warning` by default.

Regarding the user-option available in [roslyn - IDE0043](https://github.com/dotnet/roslyn/blob/4af4f1a670cf0b1219cc340087f94091d1c21e1f/src/Analyzers/CSharp/Tests/ValidateFormatString/ValidateFormatStringTests.cs#L562-L591) looking at the tests it seems that this is used to validate what I have called missing format item here. Do we want to keep this as a separate option?

jmarolf:
> I am working on the various open comment but I have a side question regarding the descriptors we want to create, basically we have the following use-cases:

my take on this is: what is the minimal set of things the user would want to configure? I cannot think of a reason I would want to have different severities for 1a vs. 1b. 

> In addition to the previous cases, there is a user option to provide additional string format methods, shall we reuse the previous descriptors or create a new set of them?

I don't think this should necessarily block this, but I do believe we will want this functionality eventually.

> Regarding the user-option available in roslyn - IDE0043 looking at the tests it seems that this is used to validate what I have called missing format item here. Do we want to keep this as a separate option?

I think this should be a separate diagnostic.

Evangelink:
> my take on this is: what is the minimal set of things the user would want to configure? I cannot think of a reason I would want to have different severities for 1a vs. 1b.

I agree but we still want to have 2 descriptors (with same ID) to have different (specialized) messages and description, don't we? 

>> In addition to the previous cases, there is a user option to provide additional string format methods, shall we reuse the previous descriptors or create a new set of them?

> I don't think this should necessarily block this, but I do believe we will want this functionality eventually.

Sorry but I am not sure to get what you wanted to say. I was wondering if we do want to have separate IDs for user-configured methods as we can't know for sure if they are going to cause runtime failure.

mavasani:

jmarolf:
> I agree but we still want to have 2 descriptors (with same ID) to have different (specialized) messages and description, don't we?

ah, yes. I am comfortable having as many messages as possible, frankly. I think these are all the cases we are aware of correct?

1. Not enough args `CA2241` (will be a warning that is on by default in .NET 6)
    1. `Console.WriteLine("{0} {1}", 42);`: **"No format argument specified for index '{1}'."**
        1. I would ideally like us to do special handling so the location we report is the `{1}` in the string to make it clear to the user what we are referring to in the diagnostic message.
    1. `Console.WriteLine("{1} {2}", 42, other);`: **"Format argument '42' does not correspond to any format index. Format index '{2}' is also unused, is there an off by one error?"**
        1. This is the most complex case. We might omit it and instead emit both 1i and 2i here if we think placing the error locations make it clear what is happening.
    1. `Console.WriteLine("{1}", 42);`: **"Format index '{1}' does not correspond to any format argument, did you mean '{0}'?"**
        1. special case but I think this would be helpful for new programmers that are not aware this is 0-indexed
1. Too many args `CAXXX` (will be a set at info severity)
    1. `Console.WriteLine("{1}", 42, other);`: "Too many arguments supplied, argument 'other' is not used."
		1. I would want the error location to be "other" here. We could experiment with having the severity by unnecessary here.
    1. `Console.WriteLine("{0} {1}")`: "No format arguments given for string, though format items are specified. Did you mean to escape the braces?"

Evangelink:
I am wondering what can be done in term of diagnostic ID, I mean could we depreciate an old ID? I'd like to avoid having a too big gap between the 2 analyzers ID. The next available ID is CA2250. Except if we want to have 2 separate categories too.

I am also wondering if we want to simply bail-out for C# varargs or if we still want to report for missing format indexes? Related to this subject, I would need help for the syntax of the empty `__arglist`, I cannot make it compile.

I will push a WIP soon, here are the remaining steps:
1. Change location of the diagnostics
2. Update all title, message and descriptions
3. Add support for detection of format items in overload without the format string argument
4. Ensure I have ported all the editorconfig from CA2241 and IDE0043

mavasani:
@Evangelink Let's not deprecate CA2241 ID, let's just repurpose it for one of the new buckets. Microsoft.CodeAnalysis.FxCopAnalyzers is soon going to be deprecated in favor of Microsoft.CodeAnalysis.NetAnalyzers, and we do  not need to preserve any backcompat behavior from FxCop in the latter, so I wouldn't worry about breaking changes in CA2241.

Evangelink:
It just feel a little weird as a user to have two really related rules so separated while I don't see much people who wouldn't both running.

mavasani:
> It just feels a little weird as a user to have two really related rules so separated while I don't see much people who wouldn't both running.

We already have multiple such cases, so it should be fine. Rule IDs should essentially be considered opaque.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -371,7 +850,18 @@ class Test
...
-        var a = MyFormat("""", 1);
+        // Too many args
+        var s1 = MyFormat(""{0}"", 1, 2);
+        // Too many args and missing format index
+        var s2 = MyFormat(""{1}"", 1, 2, 3);
+
+        // Not enough args
+        var s3 = MyFormat(""{0} {1}"", 1);
+        // Not enough args and missing format index
+        var s4 = MyFormat(""{0} {2}"", 1);
```
Evangelink, line 861: I am tempted to have a separate set of descriptors for these cases (they will use CA2241 as the too many  cases) because we are not sure these will lead to runtime failure as this is a custom implementation.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -446,7 +960,18 @@ class Test
...
```
Evangelink, line 971: Same comment regarding the separate descriptors.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -27,34 +28,51 @@ public class C
...
-        var c = String.Format(""{0} {1}"", 1, 2, 3);
-        var d = String.Format(""{0} {1} {2}"", 1, 2, 3, 4);
-        var e = string.Format(""{0} {0}"", 1, 2);
-
-        IFormatProvider p = null;
-        var f = String.Format(p, """", 1);
-        var g = String.Format(p, ""{0}"", 1, 2);
-        var h = String.Format(p, ""{0} {1}"", 1, 2, 3);
-        var i = String.Format(p, ""{0} {1} {2}"", 1, 2, 3, 4);
+        {|#0:" + invocation + @"("""", 1)|};
```
Evangelink, line 31: @jmarolf so on these cases I need to report on all extra args, is it right?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -63,82 +81,181 @@ public class C
 {
     void Method()
     {
-        Console.Write("""", 1);
-        Console.Write(""{0}"", 1, 2);
-        Console.Write(""{0} {1}"", 1, 2, 3);
-        Console.Write(""{0} {1} {2}"", 1, 2, 3, 4);
-        Console.Write(""{0} {1} {2} {3}"", 1, 2, 3, 4, 5);
+        {|#0:" + invocation + @"(""{1}"", 1, 2, 3)|};
```
Evangelink, line 84: @jmarolf here I need to report on all extra args + somewhere for the missing format indexes (any suggestion?)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -147,134 +264,361 @@ public class C
 {
     void Method()
     {
-        var s = String.Format(""{0} {1} {2} {3}"", new object[] {1, 2});
-        Console.Write(""{0} {1} {2} {3}"", new object[] {1, 2, 3, 4, 5});
-        Console.WriteLine(""{0} {1} {2} {3}"", new object[] {1, 2, 3, 4, 5});
+        {|#0:" + invocation + @"(""{0} {1}"", 1)|};
```
Evangelink, line 267: Here I highlight all format indexes without associated args, right?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -147,134 +264,361 @@ public class C
...
+            await VerifyCS.VerifyAnalyzerAsync(@"
 using System;
 
 public class C
 {
     void Method()
     {
-        Console.Write(""{0} {1} {2} {3} {4}"", 1, 2, 3, 4, __arglist(5));
-        Console.WriteLine(""{0} {1} {2} {3} {4}"", 1, 2, 3, 4, __arglist(5));
+        {|#0:" + invocation + @"(""{1}"", 1)|};
```
Evangelink, line 312: How to highlight the missing indexes?

mavasani:
@jmarolf Would be you able to review the latest commit and answer @Evangelink's questions?

jmarolf:
@mavasani I'll review this after lunch, thanks for the ping

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
```diff
@@ -456,9 +981,15 @@ void M1(string param)
 
             if (editorConfigText.Length > 0)
             {
-                csharpTest.ExpectedDiagnostics.Add(
-                    // Test0.cs(8,17): warning CA2241: Provide correct arguments to formatting methods
-                    GetCSharpResultAt(8, 17));
+                // TODO: Make sure to report only the right diagnostic not both
```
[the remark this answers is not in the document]
jmarolf: @sharwell as he is the expert on test framework helpers. I am not aware of one.

---
In reply to: [508698915](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508698915) [](ancestors = 508698915)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: I think thats what we want. Once we get the preview out for .NET 6 well see if feedback directs us to change this.

---
In reply to: [508700724](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508700724) [](ancestors = 508700724)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: That would be my preference

---
In reply to: [508701842](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508701842) [](ancestors = 508701842)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: In _this_ case I think we want to have a diagnostic on `, 1` and say that is will not be used.

---
In reply to: [508702128](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508702128) [](ancestors = 508702128)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: for this case I think the most useful location would be `1,` and `3` indicating that these are unused.

---
In reply to: [508701363](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508701363) [](ancestors = 508701363)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: I would be fine with not issuing these diagnostics at all to start. Yes, historically we've done this but and I can see the utility but I would not turn these on for folks by default. This is always going to be something that you need to opt into. I like the idea of having a separate ID for custom methods so folks can still get these warnings on their custom loggers.

---
In reply to: [508699816](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508699816) [](ancestors = 508699816)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethodsTests.cs
jmarolf: yep, lets add a new descriptor for this case.

---
In reply to: [508700027](https://github.com/dotnet/roslyn-analyzers/pull/4226#discussion_r508700027) [](ancestors = 508700027)

jmarolf:
@Evangelink done with review pass, please ping me if I missed something.

Evangelink:
@jmarolf Thank you for the various feedback! I will try to work on the pending changes by the end of the week (I might need a bit of time to get back to what's done and what's left to do, it's no longer really fresh in my mind).

mavasani (dismissed):

jmarolf:
@Evangelink gentle ping, do you have time to retarget this?

Evangelink:
@jmarolf I will try to finish this PR by the end of next week. Looking back at it, there is still a lot of work to do.

mavasani:
@Evangelink Would you have time to proceed further on this PR? Thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideCorrectArgumentsToFormattingMethods.cs
```diff
@@ -18,73 +20,160 @@ namespace Microsoft.NetCore.Analyzers.Runtime
...
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+        // e.g. Console.WriteLine("{0} {1}", 42);
+        internal static DiagnosticDescriptor NotEnoughArgsRule = DiagnosticDescriptorHelper.Create(
+            NotEnoughArgsRuleId,
+            s_localizableTitle,
+            s_localizableMessageArgs,
+            DiagnosticCategory.Usage,
+            RuleLevel.BuildWarningCandidate,
+            description: s_localizableDescription,
+            isPortedFxCopRule: true,
```
Youssef1313, line 39: Were all rules supported by FxCop?

Youssef1313:
Ping @Evangelink

PR #4228, opened 2020-09-26
CA1068: Allow caller attributes to come after cancellation token

### LINKED ISSUE #4227: CA1068 Should ignore reserved attributes from System.Runtime.CompilerServices
### Analyzer
CA1068 - CA1068:CancellationToken parameters must come last

### Describe the improvement

Compiler populated parameters as defined [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/caller-information) (i.e.  `System.Runtime.CompilerServices.CallerFilePathAttribute`) are not explicitly passed, so putting the cancellation token after them would necessitate using the parameter name to call the method with a cancellation token.

Ignore optional parameters with compiler populated attributes when considering which parameter is last, or possibly verify that they are always last because it would be messy to put the cancellation token between them.

Youssef1313:
Fixes #4227

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -63,6 +63,20 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                         last--;
                     }
 
+                    // Ignore parameters that have any of these attributes.
+                    // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
+                    var builder = ImmutableHashSet.CreateBuilder<INamedTypeSymbol>();
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerFilePathAttribute).FullName));
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName));
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
+                    var callerInformationAttributes = builder.ToImmutable();
+
+                    while (last >= 0
+                        && HasCallerInformationAttribute(methodSymbol.Parameters[last], callerInformationAttributes))
+                    {
+                        last--;
+                    }
+
                     // Skip optional parameters, UNLESS one of them is a CancellationToken
                     // AND it's not the last one.
                     if (last >= 0 && methodSymbol.Parameters[last].IsOptional
@@ -97,20 +111,6 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     if (last >= 0
                         && iprogressType != null
                         && methodSymbol.Parameters[last].Type.OriginalDefinition.Equals(iprogressType))
-                    {
-                        last--;
-                    }
-
-                    // Ignore parameters that hav any of these attributes.
-                    // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
-                    var builder = ImmutableHashSet.CreateBuilder<INamedTypeSymbol>();
-                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerFilePathAttribute).FullName));
-                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName));
-                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
-                    var callerInformationAttributes = builder.ToImmutable();
-
-                    while (last >= 0
-                        && HasCallerInformationAttribute(methodSymbol.Parameters[last], callerInformationAttributes))
                     {
                         last--;
                     }
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -138,10 +138,8 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
         }
 
         private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableHashSet<INamedTypeSymbol> callerAttributes)
-        {
-            return parameter.GetAttributes().Any(
+            => parameter.GetAttributes().Any(
                 attribute => callerAttributes.Any(
                     callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));;
-        }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -103,9 +103,9 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     // Ignore parameters that hav any of these attributes.
                     // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
                     var callerInformationAttributes = ImmutableArray.Create<INamedTypeSymbol>(
-                        compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerFilePathAttribute).FullName),
-                        compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerLineNumberAttribute).FullName),
-                        compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerMemberNameAttribute).FullName));
+                        compilationContext.Compilation.GetTypeByMetadataName(typeof(CallerFilePathAttribute).FullName),
+                        compilationContext.Compilation.GetTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName),
+                        compilationContext.Compilation.GetTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
 
                     while (last >= 0
                         && HasCallerInformationAttribute(methodSymbol.Parameters[last], callerInformationAttributes))
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -102,10 +102,11 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
 
                     // Ignore parameters that hav any of these attributes.
                     // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
-                    var callerInformationAttributes = ImmutableArray.Create<INamedTypeSymbol>(
-                        compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerFilePathAttribute).FullName),
-                        compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName),
-                        compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
+                    var builder = ImmutableHashSet.CreateBuilder<INamedTypeSymbol>();
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerFilePathAttribute).FullName));
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName));
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
+                    var callerInformationAttributes = builder.ToImmutable();
 
                     while (last >= 0
                         && HasCallerInformationAttribute(methodSymbol.Parameters[last], callerInformationAttributes))
@@ -135,7 +136,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             });
         }
 
-        private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableArray<INamedTypeSymbol> callerAttributes)
+        private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableHashSet<INamedTypeSymbol> callerAttributes)
         {
             if (!parameter.IsOptional)
             {
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -151,7 +151,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     return true;
                 }
             }
-            retrun false;
+            return false;
         }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
@@ -108,7 +108,6 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                         compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerMemberNameAttribute).FullName));
 
                     while (last >= 0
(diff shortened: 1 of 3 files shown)


Evangelink:
@Youssef1313 Have you been able to reproduce the issue? I couldn't...

Youssef1313:
@Evangelink  I haven't tried, this PR was just a quick edit from GitHub UI 😄 
But it's strange if the problem doesn't reproduce, I can't see anywhere in the existing code that it treats these attributes in any special way.

Youssef1313:
@Evangelink I read through the code again and I think I got it. The cancellation token should be optional to reproduce.
I'll open a draft PR to check test results from CI.

Evangelink:
@Youssef1313  I don't think so, some tests are raising an issue without having the parameter with a default value:

```csharp
[Fact]
        public async Task DiagnosticOnExtensionMethodWhenCancellationTokenIsNotFirstParameter()
        {
            var test = @"
using System.Threading;
static class C1
{
    public static void M1(this object p1, CancellationToken p2, object p3)
    {
    }
}";

            var expected = VerifyCS.Diagnostic().WithLocation(5, 24).WithArguments("C1.M1(object, System.Threading.CancellationToken, object)");
            await VerifyCS.VerifyAnalyzerAsync(test, expected);
        }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -99,6 +100,20 @@ public override void Initialize(AnalysisContext context)
...
 
+                    // Ignore parameters that hav any of these attributes.
+                    // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
+                    var callerInformationAttributes = ImmutableArray.Create<INamedTypeSymbol>(
+                        compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerFilePathAttribute).FullName),
+                        compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerLineNumberAttribute).FullName),
+                        compilationContext.Compilation.GetTypeByMetadaName(typeof(CallerMemberNameAttribute).FullName));
+
+                    while (last >= 0
+                        && methodSymbol.Parameters[last].IsOptional
```
Youssef1313, line 111: This is just to avoid the loop in `HasCallerInformationAttribute` if the parameter isn't optional.

Youssef1313:
@Evangelink If optional parameter exists, the analyzer behavior is different based on whether the cancellation token is optional or not.

- If cancellation token is optional: report diagnostic.
- If cancellation token isn't optional: don't report because it cannot come last in this case.

Youssef1313:
@mavasani This is preferred to be squashed when merging.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -99,6 +100,19 @@ public override void Initialize(AnalysisContext context)
                         last--;
                     }
 
+                    // Ignore parameters that hav any of these attributes.
+                    // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
+                    var callerInformationAttributes = ImmutableArray.Create<INamedTypeSymbol>(
+                        compilationContext.Compilation.GetTypeByMetadataName(typeof(CallerFilePathAttribute).FullName),
+                        compilationContext.Compilation.GetTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName),
+                        compilationContext.Compilation.GetTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
```
Evangelink, line 108: You are using the wrong API.

```suggestion
                        compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerFilePathAttribute).FullName),
                        compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName),
                        compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -99,6 +100,19 @@ public override void Initialize(AnalysisContext context)
                         last--;
                     }
 
+                    // Ignore parameters that hav any of these attributes.
+                    // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
+                    var callerInformationAttributes = ImmutableArray.Create<INamedTypeSymbol>(
```
Evangelink, line 105: This doesn't exclude the null cases, I think it'd be better to use a builder and rely on the `AddIfNotNull` method.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -120,5 +134,24 @@ public override void Initialize(AnalysisContext context)
...
+
+            var attributes = parameter.GetAttributes();
+            foreach (var attribute in attributes)
+            {
+                if (callerAttributes.Any(callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)))
+                {
+                    return true;
+                }
+            }
+            return false;
```
Evangelink, line 154: You could rewrite as a linq expression.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -334,5 +334,47 @@ End Function
             VerifyVB.Diagnostic().WithLocation(7, 21)
                 .WithArguments("Public Function SomeAsync(cancellationToken As System.Threading.CancellationToken, progress1 As System.IProgress(Of Integer), progress2 As System.IProgress(Of Integer)) As System.Threading.Tasks.Task"));
         }
+
+        [Fact]
```
Evangelink, line 338:
```suggestion
        [Fact, WorkItem(4227, "https://github.com/dotnet/roslyn-analyzers/issues/4227")]
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -334,5 +334,47 @@ End Function
...
+        [CallerFilePath] string sourceFilePath = """",
+        [CallerLineNumber] int sourceLineNumber = 0)
+    {
+        throw new NotImplementedException();
+    }
+}");
+        }
+
+        [Fact]
+        public async Task CA1068_CallerAttributesWithOptionalCancellationToken()
```
Evangelink, line 360:
```suggestion
        [Fact, WorkItem(4227, "https://github.com/dotnet/roslyn-analyzers/issues/4227")]
```

Evangelink, line 360: You could merge the 2 methods if you want.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
Youssef1313: @Evangelink Yeah looks like Compilation.GetTypeByMetadataName is even banned here.
Do you know why? I used it before in a personal source generator and it's used heavily in dotnet/roslyn without issues.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
Evangelink: Could you also add a case where the `CancellationToken` parameter is last and one where it is in between some CallerAttribute args.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
Youssef1313: @Evangelink that will need to use `Intersect` right? or I'm missing another better way?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -334,5 +334,89 @@ End Function
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+public class C
+{
+    public Task SomeAsync([CallerMemberName] string memberName = """",
+        [CallerFilePath] string sourceFilePath = """",
+        CancellationToken cancellationToken = default,
+        [CallerLineNumber] int sourceLineNumber = 0)
```
Youssef1313, line 415: @Evangelink This case you came up with is really interesting.

Should the analyzer flag it or not?
This input obviously incorrect, but the analyzer suggests to move cancellationToken to the end, which is also incorrect.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -334,5 +334,89 @@ End Function
...
+
+public class C
+{
+    public Task SomeAsync(CancellationToken cancellationToken = default,
+        [CallerMemberName] string memberName = """",
+        [CallerFilePath] string sourceFilePath = """",
+        [CallerLineNumber] int sourceLineNumber = 0)
+    {
+        throw new NotImplementedException();
+    }
```
Youssef1313, line 376: This test is failing ;-(

I'll debug it soon and fix the formatting issues.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
Evangelink: I haven't tested locally but I think the following should work `parameter.GetAttributes().Any(attribute => callerAttributes.Any(callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
Evangelink: Yes I know why :) Basically the resolution done within the `GetTypeByMetadataName` is expensive so we are building a cache to ensure the expensive part is done only once per compilation.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -120,5 +136,24 @@ public override void Initialize(AnalysisContext context)
...
+        }
```
Youssef1313, line 157:
```suggestion
        private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableHashSet<INamedTypeSymbol> callerAttributes)
        {
            return parameter.GetAttributes().Any(
                attribute => callerAttributes.Any(
                    callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));;
        }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -120,5 +136,12 @@ public override void Initialize(AnalysisContext context)
                 SymbolKind.Method);
             });
         }
+
+        private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableHashSet<INamedTypeSymbol> callerAttributes)
+        {
+            return parameter.GetAttributes().Any(
+                attribute => callerAttributes.Any(
+                    callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));;
+        }
```
Evangelink, line 145: [Optional] You could rewrite as a expression-bodied method.

```suggestion
        private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableHashSet<INamedTypeSymbol> callerAttributes)
            => parameter.GetAttributes().Any(
                attribute => callerAttributes.Any(
                    callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));;
```

Evangelink (approved):
LGTM

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -334,5 +334,89 @@ End Function
...
+}");
+        }
+
+        [Fact, WorkItem(4227, "https://github.com/dotnet/roslyn-analyzers/issues/4227")]
+        public async Task CA1068_CallerAttributesWithOptionalCancellationToken2()
```
Evangelink, line 381:
```suggestion
        public async Task CA1068_CallerAttributesWithOptionalCancellationTokenAsLastParameter()
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLastTests.cs
```diff
@@ -334,5 +334,89 @@ End Function
...
+        [CallerLineNumber] int sourceLineNumber = 0,
+        CancellationToken cancellationToken = default)
+    {
+        throw new NotImplementedException();
+    }
+}");
+        }
+
+        [Fact, WorkItem(4227, "https://github.com/dotnet/roslyn-analyzers/issues/4227")]
+        public async Task CA1068_CallerAttributesWithOptionalCancellationToken3()
```
Evangelink, line 402:
```suggestion
        public async Task CA1068_CallerAttributesWithOptionalCancellationTokenAsMiddleParameter()
```

Evangelink (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -120,5 +136,10 @@ public override void Initialize(AnalysisContext context)
                 SymbolKind.Method);
             });
         }
+
+        private static bool HasCallerInformationAttribute(IParameterSymbol parameter, ImmutableHashSet<INamedTypeSymbol> callerAttributes)
+            => parameter.GetAttributes().Any(
+                attribute => callerAttributes.Any(
+                    callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));;
```
Youssef1313, line 143:
```suggestion
                    callerAttribute => SymbolEqualityComparer.Default.Equals(callerAttribute, attribute.AttributeClass)));
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
```diff
@@ -61,6 +63,20 @@ public override void Initialize(AnalysisContext context)
                         last--;
                     }
 
+                    // Ignore parameters that have any of these attributes.
+                    // C# reserved attributes: https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/caller-information
+                    var builder = ImmutableHashSet.CreateBuilder<INamedTypeSymbol>();
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerFilePathAttribute).FullName));
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerLineNumberAttribute).FullName));
+                    builder.AddIfNotNull(compilationContext.Compilation.GetOrCreateTypeByMetadataName(typeof(CallerMemberNameAttribute).FullName));
```
mavasani, line 71: Fetching well know types must be done in the compilation start action. Please move this outside the symbol action callback.

mavasani: Additionally, please define [WellKnownTypeNames](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/WellKnownTypeNames.cs) for each of these attribute types instead of `typeof(CallerFilePathAttribute).FullName`

mavasani (commented):
Change looks good, but needs moving well known type computation logic into CompilationStartAction.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CancellationTokenParametersMustComeLast.cs
Evangelink: Oh I totally missed those!

mavasani (approved):

PR #4232, opened 2020-09-27
RethrowToPreserveStackDetailsAnalyzer: bail early on records

Youssef1313:
While walking up the tree, we can't find catch clauses if we arrived to a record declaration (just the same as class declaration).

The change here shouldn't change any behavior (i.e. it's un-testable), and only to avoid walking up the tree unnecessarily.

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
@@ -62,6 +62,7 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
                     case SyntaxKind.AnonymousMethodExpression:
                     case SyntaxKind.ClassDeclaration:
                     case SyntaxKind.StructDeclaration:
+                    case SyntaxKind.RecordDeclaration:
                         return;
                 }
             }
--- a/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
@@ -62,7 +62,9 @@ namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
                     case SyntaxKind.AnonymousMethodExpression:
                     case SyntaxKind.ClassDeclaration:
                     case SyntaxKind.StructDeclaration:
+#if CODEANALYSIS_V3_OR_BETTER
                     case SyntaxKind.RecordDeclaration:
+#endif
                         return;
                 }
             }
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
@@ -495,6 +495,23 @@ End Class
 ");
         }
 
+        [Fact]
+        public async Task CA2200_NoDiagnosticsForFieldInClassOrRecord()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+
+public class C
+{
+    private string s = default(string) ?? throw new Exception();
+}
+
+public record R
+{
+    private string s = default(string) ?? throw new Exception();
+}");
+        }
+
         private static DiagnosticResult GetCA2200BasicResultAt(int line, int column)
             => VerifyVB.Diagnostic()
                 .WithLocation(line, column);


src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
```diff
@@ -62,6 +62,7 @@ private static void AnalyzeNode(SyntaxNodeAnalysisContext context)
                     case SyntaxKind.AnonymousMethodExpression:
                     case SyntaxKind.ClassDeclaration:
                     case SyntaxKind.StructDeclaration:
+                    case SyntaxKind.RecordDeclaration:
```
Evangelink, line 65: Same remark as for the other PR, you will have to use the condition compilation (https://github.com/dotnet/roslyn-analyzers/pull/4223/files#diff-d8afcf96221c79be0c7c1d6e2defd475R754=

Youssef1313:
```suggestion
#if CODEANALYSIS_V3_OR_BETTER
                    case SyntaxKind.RecordDeclaration:
#endif
```

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
```diff
@@ -62,6 +62,9 @@ private static void AnalyzeNode(SyntaxNodeAnalysisContext context)
                     case SyntaxKind.AnonymousMethodExpression:
                     case SyntaxKind.ClassDeclaration:
                     case SyntaxKind.StructDeclaration:
+#if CODEANALYSIS_V3_OR_BETTER
+                    case SyntaxKind.RecordDeclaration:
```
Evangelink, line 66: I did a bit of digging and I think we can add a test for the `struct`, `class` and `record` cases doing something like:

```csharp
using System;

public class C
{
    private string? s = default(string) ?? throw new Exception();
}
```

Evangelink:
@mavasani Any idea why there is no explicit stop for the method, property, event and also local func? If that's ok for you we could create a separate ticket and handle it.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
Youssef1313: This will return very early:

https://github.com/dotnet/roslyn-analyzers/blob/7641932817a0871edc7ba15a22f357a8581ef57c/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs#L24-L29

It shouldn't get to this code path at all

Youssef1313:
> 
> 
> @mavasani Any idea why there is no explicit stop for the method, property, event and also local func? If that's ok for you we could create a separate ticket and handle it.

To even avoid all this, the analyzer can register for a `catch` instead of `throw`, then find all descendants that are `throw` statements.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
sharwell: What about this?

```csharp
public class C
{
  private string? S
  {
    get
    {
      throw new Exception();
    }
  }
}
```

sharwell: 📝  This is part of the reason why I love code coverage tooling as part of CI builds. If you have code coverage tooling and make an assumption like this for testing, you can see that test cases were eliminated if someone comes along in the future and adds the special cases for accessors or properties.

Youssef1313: @sharwell That's correct. I was also mistaken of the sample provided by @Evangelink, the expression isn't null. My comment (https://github.com/dotnet/roslyn-analyzers/pull/4232#discussion_r495897805) looks a bit funny.

The suggested tests should improve test coverage (but still there is no behavioral change - it just stops walking up the tree when it's known we can't meet a catch clause).

But I think my suggestion in https://github.com/dotnet/roslyn-analyzers/pull/4232#issuecomment-699975946 should make things a little bit faster in most cases.

The current approach walks up the tree for throw expressions. (the registered action is for SyntaxKind.ThrowStatement).

Registering for CatchStatement instead should be a bit faster for most cases, it also doesn't require this list of switch to return early.

What do you think?

Evangelink: @Youssef1313 Having some tests showing nothing is reported is a good test IMO.

Regarding the modification of the analyzer to register on the catch and then walking down, as far as I remember this is a "bad" idea because when you have the callback you don't have any guarantee that what is below the callback has been created. Doing the walk will force the creation but it doesn't go in the normal route. I might be a little wrong on the exact explanation but I remember that going up was ok but down was to avoid.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -495,6 +495,23 @@ End Class
...
+public class C
+{
+    private string s = default(string) ?? throw new Exception();
+}
+
+public record R
+{
+    private string s = default(string) ?? throw new Exception();
+}");
+        }
```
Evangelink, line 513:
```suggestion
        [Theory]
        [InlineData("class")]
        [InlineData("struct")]
        [InlineData("record")]
        public async Task CA2200_NoDiagnosticsForFieldThrowExpression(string typeKind)
        {
            await VerifyCS.VerifyAnalyzerAsync(@"
using System;

public " + typeKind + @" C
{
    private string s = default(string) ?? throw new Exception();
}");
        }
```

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
```diff
@@ -62,6 +62,9 @@ private static void AnalyzeNode(SyntaxNodeAnalysisContext context)
                     case SyntaxKind.AnonymousMethodExpression:
```
Evangelink, line 62: You could do so in a separate PR but it'd be nice to also have a test (if they are not already present) for action, func and local func.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
Evangelink: @sharwell I was actually wondering if there was a reason not to have any stop on member declaration because I don't see the need to walk-up anymore. If you agree, I will create a ticket for it.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
Youssef1313: @Evangelink I think it won't work with structs that way. It will result in a compile error.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpRethrowToPreserveStackDetails.cs
mavasani: This analyzer can be rewritten as a very simple IOperation analyzer, which will drastically simplify the implementation, make it language agnostic and improve the overall performance. We will also not need these very specific syntax checks. We just walk up the operation tree until we reach a catch operation, or a lambda/local function operation.

Evangelink: @mavasani I was waiting for this feedback, I have a PR ready for it :)

Youssef1313:
I'm closing in favor of https://github.com/dotnet/roslyn-analyzers/issues/4240

PR #4233, opened 2020-09-27
Add RecordDeclaration to IsNamedTypeDeclarationBlock

Youssef1313:
I'm not sure if this changes behavior or not.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
@@ -14,7 +14,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Analyzers.MetaAnalyzers
             switch (syntax.Kind())
             {
                 case SyntaxKind.ClassDeclaration:
- #if CODEANALYSIS_V3_OR_BETTER
+#if CODEANALYSIS_V3_OR_BETTER
                 case SyntaxKind.RecordDeclaration:
 #endif
                 case SyntaxKind.StructDeclaration:
--- a/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
@@ -14,6 +14,7 @@ namespace Microsoft.CodeAnalysis.CSharp.Analyzers.MetaAnalyzers
             switch (syntax.Kind())
             {
                 case SyntaxKind.ClassDeclaration:
+                case SyntaxKind.RecordDeclaration:
                 case SyntaxKind.StructDeclaration:
                 case SyntaxKind.EnumDeclaration:
                 case SyntaxKind.InterfaceDeclaration:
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
```diff
@@ -14,6 +14,7 @@ protected override bool IsNamedTypeDeclarationBlock(SyntaxNode syntax)
             switch (syntax.Kind())
             {
                 case SyntaxKind.ClassDeclaration:
+                case SyntaxKind.RecordDeclaration:
```
Evangelink, line 17: Let's add a test for this.

Evangelink:
@Youssef1313 Build is failing because the syntax is not found on all compilations. You might need to have this case handled only for Roslyn 3+ (see https://github.com/dotnet/roslyn-analyzers/pull/4223/files#diff-d8afcf96221c79be0c7c1d6e2defd475R754)

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
Youssef1313, line 17:
```suggestion
 #if CODEANALYSIS_V3_OR_BETTER
                case SyntaxKind.RecordDeclaration:
#endif
```

Youssef1313: @Evangelink I'm not sure how the test should look like to have a different behavior before and after.

Evangelink: I haven't tested but you might be able to copy a test raising an issue and change class to be record and show that this does raise an issue.

Youssef1313: @Evangelink tests are based on classes that inherit DiagnosticAnalyzer. Records on the other hand cannot inherit classes.
I'll debug this to make sure if records can come across this code path (e.g. record nested inside a class that inherit diagnostic analyzer)

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
```diff
@@ -14,6 +14,9 @@ protected override bool IsNamedTypeDeclarationBlock(SyntaxNode syntax)
             switch (syntax.Kind())
             {
                 case SyntaxKind.ClassDeclaration:
+ #if CODEANALYSIS_V3_OR_BETTER
```
Youssef1313, line 17:
```suggestion
#if CODEANALYSIS_V3_OR_BETTER
```

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpDiagnosticAnalyzerAPIUsageAnalyzer.cs
Evangelink: Oh I was left on an old proposition for the record where the inheritance was allowed but it seems it is no longer the case (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/records). So basically what's the purpose of your change of there is no way to reach this point?

Youssef1313: @Evangelink I didn't mean this is unreachable with my previous comment. (I also didn't mean this is reachable).

Until now I'm not sure whether this is reachable or not.
And if it's reachable, will this have a behavior change, or just returning early for some cases.

The method is called in two different places in the analyzer, which should have two different purposes. But the current implementation has struct, enum, interface. If all of these are reachable, then records shouldn't different.

I'll debug soon to get feel of how the code flows and get an understanding of it.

mavasani:
@Youssef1313 are you able to add a regression test for this?

Youssef1313:
@mavasani Nope, I *think* this helps to bail-out early for irrelevant nodes.

PR #4238, opened 2020-09-29
Improve test coverage for RethrowToPreserveStackDetailsAnalyzer

### LINKED ISSUE #4236: Improve test coverage for RethrowToPreserveStackDetailsAnalyzer
You could do so in a separate PR but it'd be nice to also have a test (if they are not already present) for action, func and local func.

_Originally posted by @Evangelink in https://github.com/dotnet/roslyn-analyzers/pull/4232#r496092997_

Evangelink:
Fix #4236

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Testing;
 using Test.Utilities;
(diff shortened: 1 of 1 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -495,6 +496,68 @@ End Class
...
+    void CatchAndRethrowExplicitly()
+    {
+        try
+        {
+        }
+        catch (Exception e)
+        {
+            Action DoThrowAction = () =>
+            {
+                throw e;
```
Evangelink, line 544: Maybe we do want to report here as we do report for the inner throw in https://github.com/dotnet/roslyn-analyzers/blob/master/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs#L324-L351

mavasani: I don't think we want to report for throw crossing method boundaries, this lambda can be invoked after say `e` has been re-assigned a different exception. That would be a false positive.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -495,6 +496,68 @@ End Class
...
+        {
+        }
+        catch (Exception e)
+        {
+
+            DoThrow();
+
+            void DoThrow()
+            {
+                [|throw e;|]
```
mavasani, line 519: This indicates there is a bug. Likely following will show a violation, when it should not:
```csharp
    void CatchAndRethrowExplicitly()
    {
        try
        {
        }
        catch (Exception e)
        {
            e = new MyException();
            DoThrow();
            void DoThrow()
            {
                throw e; // This should not fire.
            }
        }
    }
```

mavasani (approved):

PR #4241, opened 2020-09-29
Update CA1812 to handle types declared through DebuggerTypeProxy

### LINKED ISSUE #1708: CA1812 Shouldn't Fire for DebuggerTypeProxy

Evangelink:
Fix #1708


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+
+                bool CanBeDesignerAttributeContext(INamedTypeSymbol type)
+                    => designerAttributeSymbol != null && (type.TypeKind == TypeKind.Interface || type.TypeKind == TypeKind.Class);
+
+                INamedTypeSymbol? FindTypeIfDesignerAttribute(AttributeData attribute, Compilation compilation)
+                {
+                    RoslynDebug.Assert(designerAttributeSymbol != null);
+
+                    if ((attribute.ConstructorArguments.Length != 1 && attribute.ConstructorArguments.Length != 2) ||
+                        !attribute.AttributeClass.Equals(designerAttributeSymbol))
```
Youssef1313, line 207: Should this use SymbolEqualityComparer? I'm not sure, just asking.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+                }
+
+                bool CanBeDesignerAttributeContext(INamedTypeSymbol type)
+                    => designerAttributeSymbol != null && (type.TypeKind == TypeKind.Interface || type.TypeKind == TypeKind.Class);
+
+                INamedTypeSymbol? FindTypeIfDesignerAttribute(AttributeData attribute, Compilation compilation)
+                {
+                    RoslynDebug.Assert(designerAttributeSymbol != null);
+
+                    if ((attribute.ConstructorArguments.Length != 1 && attribute.ConstructorArguments.Length != 2) ||
```
Youssef1313, line 206: You may want to use the new C# 9.0 syntax `is not (1 or 2)`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+                        !attribute.AttributeClass.Equals(designerAttributeSymbol))
+                    {
+                        return null;
+                    }
+
+                    switch (attribute.ConstructorArguments[0].Value)
```
Youssef1313, line 212: Can this cause out of range exception?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+                    {
+                        case string designerTypeName when designerTypeName != null:
```
Youssef1313, line 214: The when is redundant here.

```suggestion
                        case string designerTypeName:
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+                            {
+                                if (IsTypeInCurrentAssembly(designerTypeName, compilation, out var namedType))
+                                {
+                                    return namedType;
+                                }
+                                break;
+                            }
+
+                        case INamedTypeSymbol namedType when namedType != null:
```
Youssef1313, line 223:
```suggestion
                        case INamedTypeSymbol namedType:
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+                    RoslynDebug.Assert(debuggerTypeProxyAttributeSymbol != null);
+
+                    if (!attribute.AttributeClass.Equals(debuggerTypeProxyAttributeSymbol))
+                    {
+                        return null;
+                    }
+
+                    switch (attribute.ConstructorArguments[0].Value)
+                    {
+                        case string typeName when typeName != null:
```
Youssef1313, line 244:
```suggestion
                        case string typeName:
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
```diff
@@ -197,6 +172,90 @@ public sealed override void Initialize(AnalysisContext analysisContext)
...
+                            {
+                                if (IsTypeInCurrentAssembly(typeName, compilation, out var namedType))
+                                {
+                                    return namedType;
+                                }
+                                break;
+                            }
+
+                        case INamedTypeSymbol namedType when namedType != null:
```
Youssef1313, line 253:
```suggestion
                        case INamedTypeSymbol namedType:
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
Evangelink: We have previously check there is 1 or 2 args so IMO it should not.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
Evangelink: TBH I am not sure, I did see the warning in VS but as far as I can tell it's not available in all compilation paths.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
Youssef1313: Yes, but it's `||`ed with `attribute.AttributeClass.Equals(designerAttributeSymbol)`.

If, for some reason, the runtime in a future version introduced a parameterless constructor, would this cause a problem?
Probably the other check needs to be `&&` instead? Or I might be missing something.

Evangelink: If there is a number of args different from 1 and 2 -> bail-out. If it is not the right attribute -> bail-out. So for me there is no problem when reaching this point we are guarantee to have 1 or 2 params (so 0 exists).

If a parameterless constructor is introduced, then it won't match the 1st condition (i.e. 1 or 1 params) so that's safe too.

Youssef1313: @Evangelink Oh yes my mistake.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClassesTests.cs
```diff
@@ -1323,24 +1323,31 @@ internal class CSomeClass {}
...
         }
 
-        [Fact, WorkItem(2957, "https://github.com/dotnet/roslyn-analyzers/issues/2957")]
-        public async Task CA1812_DesignerAttributeTypeName_NoDiagnostic()
+        [Theory]
+        [WorkItem(2957, "https://github.com/dotnet/roslyn-analyzers/issues/2957")]
+        [WorkItem(1708, "https://github.com/dotnet/roslyn-analyzers/issues/1708")]
+        [InlineData("System.ComponentModel.DesignerAttribute")]
+        [InlineData("System.Diagnostics.DebuggerTypeProxyAttribute")]
+        public async Task CA1812_DesignerAttributeTypeName_NoDiagnostic(string attributeFullName)
```
mavasani, line 1331: Rename the test as this is no longer specific to designer attribute? Same for other modified tests..

mavasani (approved):

mavasani:
Merging as the suggestion is pretty minor, may not really be required to be addressed. If you wish to address it, feel free to create a separate PR.

PR #4247, opened 2020-09-30
Rewrite RethrowToPreserveStackDetailsAnalyzer as operation analyzer

### LINKED ISSUE #4240: Re-write RethrowToPreserveStackDetailsAnalyzer
This analyzer can be rewritten as a very simple IOperation analyzer, which will drastically simplify the implementation, make it language agnostic and improve the overall performance. We will also not need these very specific syntax checks. We just walk up the operation tree until we reach a catch operation, or a lambda/local function operation.

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/pull/4232#discussion_r496779275_

Evangelink:
Fix #4240 and address @mavasani comments on https://github.com/dotnet/roslyn-analyzers/pull/4238/files


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -25,9 +28,46 @@ public abstract class RethrowToPreserveStackDetailsAnalyzer : DiagnosticAnalyzer
...
+                while (ancestor != null &&
+                    ancestor.Kind != OperationKind.AnonymousFunction &&
+                    ancestor.Kind != OperationKind.LocalFunction)
+                {
+                    if (ancestor.Kind == OperationKind.CatchClause &&
+                        ancestor is ICatchClauseOperation catchClause &&
+                        catchClause.Locals.Contains(localReference.Local) &&
+                        !IsReassignedInCatch(catchClause, localReference))
+                    {
+                        context.ReportDiagnostic(throwOperation.CreateDiagnostic(Rule));
```
mavasani, line 55: I think the original code had a false positive bug, and your change will also have the same bug:

```csharp
using System;

public class Test
{
    public void M1()
    {
        try
        {
        }
        catch (Exception ex1)
        {
            try
            {
            }
            catch (Exception ex2)
            {
                throw ex1;  // Changing this to a 'rethrow' will change semantics.
            }
        }
    }
}
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -5,10 +5,13 @@
 using Microsoft.CodeAnalysis;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.CodeAnalysis.CSharp;
```
mavasani, line 9: Why do you need `using Microsoft.CodeAnalysis.CSharp;`?

mavasani: Hmm interesting. Seems like it is not necessary but unnecessary using is not triggered for it.

Evangelink: Oh awesome! I did reported an issue for this on roslyn but couldn't manage to create a reproducer I could share (it was on some internal company code).

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
Evangelink: Want me to fix it as part of this PR or through another one?

mavasani: I'll leave it to your preference.

mavasani (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
mavasani:
```suggestion
```

mavasani:
@Evangelink Let me know if you'd prefer to fix the nested catch issue with a separate PR or this one.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -5,10 +5,13 @@
...
 
 namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
 {
-    public abstract class RethrowToPreserveStackDetailsAnalyzer : DiagnosticAnalyzer
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class RethrowToPreserveStackDetailsAnalyzer : DiagnosticAnalyzer
     {
         internal const string RuleId = "CA2200";
         private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.RethrowToPreserveStackDetailsTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
```
Youssef1313, line 16:
```suggestion
        private static readonly LocalizableString s_localizableTitle = new(nameof(MicrosoftCodeQualityAnalyzersResources.RethrowToPreserveStackDetailsTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources));
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -25,9 +28,46 @@ public abstract class RethrowToPreserveStackDetailsAnalyzer : DiagnosticAnalyzer
...
+
+                if (throwOperation.Exception is not ILocalReferenceOperation localReference)
+                {
+                    return;
+                }
+
+                IOperation ancestor = throwOperation;
+                while (ancestor != null &&
+                    ancestor.Kind != OperationKind.AnonymousFunction &&
+                    ancestor.Kind != OperationKind.LocalFunction)
```
Youssef1313, line 47: OperationKind.MethodBody ?

Youssef1313: I'm not sure what OperationKind.FlowAnonymousFunction. But assuming it's some special kind of AnonymousFunction, you might want to consider it.

mavasani: I don't think you need to handle MethodBody, that will anyway be the root operation.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -25,9 +27,49 @@ public abstract class RethrowToPreserveStackDetailsAnalyzer : DiagnosticAnalyzer
...
+                {
+                    if (ancestor.Kind == OperationKind.CatchClause &&
+                        ancestor is ICatchClauseOperation catchClause)
```
Youssef1313, line 50: Is the first check necessary?
I think ***every*** ICatchClauseOperation should have OperationKind.CatchClause and vice-versa.

```suggestion
                    if (ancestor is ICatchClauseOperation catchClause)
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
mavasani: Lets avoid such generic style changes in this PR. If desired, this should be done across the repo in a separate PR.

mavasani (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
Evangelink: @Youssef1313 I am doing the first check for a fast bail-out, I don't know if that's something we care about. Some analyzers do some don't.

@mavasani any opinion on this?

mavasani: Seems fine either way to me.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
Evangelink: > I don't think you need to handle MethodBody, that will anyway be the root operation.

Yes, exactly! 

> I'm not sure what OperationKind.FlowAnonymousFunction. But assuming it's some special kind of AnonymousFunction, you might want to consider it.

Looking at the [doc](https://docs.microsoft.com/dotnet/api/microsoft.codeanalysis.flowanalysis.iflowanonymousfunctionoperation) it seems fine to ignore.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
Youssef1313: If it's known that the second check is slower, then keep both. I wasn't aware that the first check is faster than the second.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -6,10 +6,10 @@
 using Test.Utilities;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
-    Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines.CSharpRethrowToPreserveStackDetailsAnalyzer,
+    Microsoft.CodeQuality.Analyzers.QualityGuidelines.RethrowToPreserveStackDetailsAnalyzer,
     Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
 using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
-    Microsoft.CodeQuality.VisualBasic.Analyzers.QualityGuidelines.BasicRethrowToPreserveStackDetailsAnalyzer,
+    Microsoft.CodeQuality.Analyzers.QualityGuidelines.RethrowToPreserveStackDetailsAnalyzer,
     Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
```
Youssef1313, line 13: I see there is a file `RethrowToPreserveStackDetailsTests.Fixer`. Not sure why the test is using `EmptyCodeFixProvider`?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -511,12 +502,39 @@ void CatchAndRethrowExplicitly()
...
+                throw e;
+            }
+        }
+    }
+}
+");
+        }
+
+        [Fact]
+        public async Task CA2200_NoDiagnosticsForThrowCaughtExceptionInLocalMethodAfterReassignment()
```
Youssef1313, line 518: Considering having a test for top-level statements (not necessarily in this PR).

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
Evangelink: :)

jmarolf (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
CyrusNajmabadi: the copmiler may be seeing a type in that and considering it during lookup while doing lambda overload resolution.  If you comment out your `context =>` lambda does this go away?

CyrusNajmabadi: if so, this is one of those cases where the namespace is marked by the compiler because it did examien it, but eventually ended up with a resolution taht did not use anything from it.  if so, this is a known limitation/extremely-difficult-problem in this space.

PR #4249, opened 2020-09-30
Update CA1062 to handle NotNullAttribute

### LINKED ISSUE #4248: CA1062 should check .NET [NotNull] attribute
### Analyzer

**Diagnostic ID**: [CA1062](https://docs.microsoft.com/visualstudio/code-quality/CA1062)

### Describe the improvement

The rule CA1062 currently checks for attributes named `ValidatedNotNullAttribute` on guard/validate methods to determine if a parameter has been checked for `null` - like so:

```c#
public static class Guard
{
    public static void IsNotNull<T>([ValidatedNotNull] T value, string paramName)
    {
        if (value is null)
        {
        }
    }
}

...

public void MyMethod(string someValue)
{
    Guard.IsNotNull(someValue, nameof(someValue);

    if (someValue.Length > 0) // member access is ok
    ...
}
```

The class `ValidatedNotNullAttribute` is not part of .NET itself. Instead it has to be implemented by the user themselves.

Since C# 8 (.NET Standard 2.1, .NET Core 3.0), .NET comes with the [NotNullAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute) that does exactly the same thing:

> Specifies that an input argument was not `null` when the call returns.

So, CA1062 should check for this attribute as well.

At the same places where CA1062 currently checks for the `ValidatedNotNullAttribute`, it should also check for `System.Diagnostics.CodeAnalysis.NotNullAttribute`.

### Additional context

See also: https://github.com/dotnet/csharplang/discussions/3954

Evangelink:
Fix #4248

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
@@ -568,6 +568,40 @@ End Class
         }
 
         [Fact]
+        public async Task NotNullAttribute_PossibleNullRefUsage_NoDiagnostic()
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.NetCore.NetCoreApp31,
+                TestCode = @"
+using System.Diagnostics.CodeAnalysis;
+
+public class Test
+{
+    public void M1([NotNull]string str)
+    {
+        var x = str.ToString();
+    }
+}
+",
+            }.RunAsync();
+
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ValidateArgumentsOfPublicMethodsTests.cs
```diff
@@ -567,6 +567,40 @@ End Class
             GetBasicResultAt(8, 34, "Sub Test.M1(str As String, str2 As String)", "str2"));
         }
 
+        [Fact]
```
mavasani, line 570: WorkItem attribute for all tests?

mavasani (approved):

PR #4252, opened 2020-09-30
Update RS1010 to not report on overload with nestedActions

### LINKED ISSUE #3475: RS1010 (use equivalence key) false positive for nested code actions
#### Analyzer package

Roslyn analyzers

#### Package Version

2.9.8

#### Diagnostic ID

RS1010

#### Repro steps

Use the `CodeAction.Create` overload that takes a list of nested code actions.

#### Expected behavior

No diagnostic. There is no way to set the equivalence key for this overload.

#### Actual behavior

RS1010 is reported, but not fixable.

Evangelink:
Fix #3475

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
@@ -280,7 +280,8 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
                     IParameterSymbol param = invocation.TargetMethod.Parameters.FirstOrDefault(p => p.Name == EquivalenceKeyParameterName);
                     if (param == null)
                     {
-                        return true;
+                        // User is calling an overload without the equivalenceKey parameter
+                        return false;
                     }
 
                     foreach (var argument in invocation.Arguments)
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeAnalysis.Analyzers/UnitTests/FixAnalyzers/FixerWithFixAllAnalyzerTests.cs
```diff
@@ -375,6 +377,63 @@ private string GetKey()
             await TestCSharpCoreAsync(source, missingGetFixAllProviderOverrideDiagnostic, withCustomCodeActions: true);
         }
 
+        [Fact, WorkItem(3475, "https://github.com/dotnet/roslyn-analyzers/issues/3475")]
+        public async Task CSharp_CodeActionCreateNestedActions_NoDiagnostics()
+        {
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.Default.AddPackages(ImmutableArray.Create(new PackageIdentity("Microsoft.CodeAnalysis", "3.3.0"))),
```
Evangelink, line 385: Overload is not available before this version.

src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
```diff
@@ -280,7 +280,8 @@ static bool IsViolatingCodeActionCreateInvocation(IInvocationOperation invocatio
                     IParameterSymbol param = invocation.TargetMethod.Parameters.FirstOrDefault(p => p.Name == EquivalenceKeyParameterName);
                     if (param == null)
```
Evangelink, line 281: We could be checking that the overload contains the `nestedActions` param but I think that the current check is good enough.

mavasani: So the assumption here is CodeAction.Create will only have overloads without `equivalenceKey` parameter when not applicable? Seems reasonable.

mavasani (approved):

PR #4255, opened 2020-09-30
Upload coverage to codecov.io is flaky, so do not fail build on uploa…

mavasani:
…d failure

We are seeing lot of failures in debug queue due to codecov upload failures.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,7 +41,7 @@ jobs:
       condition: always()
     - script: dotnet msbuild -binaryLogger:artifacts\log\$(_configuration)\uploadCodeCov.binlog;ProjectImports=Embed -restore eng/CodeCoverage.proj
       displayName: Upload coverage to codecov.io
-      condition: always() # Upload coverage to codecov.io is flaky, so do not fail build on upload failure
+      condition: and(always(), eq(variables['_codeCoverage'], True)) # Upload coverage to codecov.io is flaky, so do not fail build on upload failure
     - task: PublishBuildArtifacts@1
       displayName: Publish Logs
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,7 +41,7 @@ jobs:
       condition: always()
     - script: dotnet msbuild -binaryLogger:artifacts\log\$(_configuration)\uploadCodeCov.binlog;ProjectImports=Embed -restore eng/CodeCoverage.proj
       displayName: Upload coverage to codecov.io
-      condition: and(succeeded(), eq(variables['_codeCoverage'], True))
+      condition: always() # Upload coverage to codecov.io is flaky, so do not fail build on upload failure
     - task: PublishBuildArtifacts@1
       displayName: Publish Logs
       inputs:


jmarolf (approved):

azure-pipelines.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       condition: always()
     - script: dotnet msbuild -binaryLogger:artifacts\log\$(_configuration)\uploadCodeCov.binlog;ProjectImports=Embed -restore eng/CodeCoverage.proj
       displayName: Upload coverage to codecov.io
-      condition: and(succeeded(), eq(variables['_codeCoverage'], True))
+      condition: always() # Upload coverage to codecov.io is flaky, so do not fail build on upload failure
```
mavasani, line 44:
```suggestion
      condition: and(always(), eq(variables['_codeCoverage'], True)) # Upload coverage to codecov.io is flaky, so do not fail build on upload failure
```

sharwell:
Are we using the latest version of the uploader? Curious what the error(s) are.

azure-pipelines.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       condition: always()
     - script: dotnet msbuild -binaryLogger:artifacts\log\$(_configuration)\uploadCodeCov.binlog;ProjectImports=Embed -restore eng/CodeCoverage.proj
       displayName: Upload coverage to codecov.io
-      condition: and(succeeded(), eq(variables['_codeCoverage'], True))
+      condition: and(always(), eq(variables['_codeCoverage'], True)) # Upload coverage to codecov.io is flaky, so do not fail build on upload failure
```
Evangelink, line 44: Is it possible to introduce some retry pattern?

mavasani: I don't see a reason for it - if someone really wants to look at codecov data and upload failed, they can manually re-run the debug queue on the PR. I don't believe queue should fail if upload of such additional information fails. We already take this approach if publishing test results fail, so this is identical.

mavasani:
> Are we using the latest version of the uploader? Curious what the error(s) are.

I am not sure, but regardless, I don't believe this should be a required gate. It should be optional, just like publishing test results.

sharwell (changes requested):
Can we try updating the codecov uploader tool first?

azure-pipelines.yml
Evangelink: I agree with the fact it should not fail the build, I just think that the coverage metric is useful so having a retry seemed like a good option.

sharwell:
> We are seeing lot of failures in debug queue due to codecov upload failures.

I see a few failures (6% of total for the month), but they tend to be bucketed together occurring a few weeks apart. It seems more like a case where we'd work with the partner to improve the upload reliability.

jmarolf:
While I agree with the sentiment @sharwell, we should update the uploader and we should engage to see if we can make things more resilient. I do not think we should block this PR. Unless we are prepared to say that code coverage data is important enough to fail the build, logicallyy it should be optional.

> In the past, this led to extremely long delays between when a configuration error led to continuous upload failures, and when they were found/fixed.

Still don't see this as being useful. We should investigate warnings in the builds, but I think blocking all codeflow for this is overkill.

mavasani:
> In the past, this led to extremely long delays between when a configuration error led to continuous upload failures, and when they were found/fixed. Required uploads tend to provide better long-term build stability.

Lets try to get whatever enforcement/requirement model we want for codecov to flow from arcade. If all arcade based repos mandate this, I am fine to do so for this repo. Otherwise, we are just losing productivity here.

sharwell:
@mavasani can we make it a separate build job (e.g. what dotnet/roslyn#47249 is doing? That would allow two things:

1. We could make it a GitHub check that is not required
2. We could retry _just the upload_ without forcing a new build

mavasani:
@sharwell Those all seem good long term improvements. This PR is aimed at unblocking unnecessary build failures. Feel free to file a follow-up issue to track these improvements.

mavasani:
Filed https://github.com/dotnet/roslyn-analyzers/issues/4261 for long term improvements for codecov. I am merging this PR to unblock..

PR #4256, opened 2020-09-30
Fix unspeakable main name

### LINKED ISSUE #4253: Review IsTopLevelStatementsEntryPointMethod implementation
This was implemented in https://github.com/dotnet/roslyn-analyzers/pull/4066/files#diff-b16e801c575072ba829f30b96c9d2362R692-R694

The current implementation:

```csharp
        public static bool IsTopLevelStatementsEntryPointMethod([NotNullWhen(true)] this IMethodSymbol? methodSymbol)
            => methodSymbol?.ContainingType.IsTopLevelStatementsEntryPointType() == true &&
               methodSymbol.IsStatic &&
               methodSymbol.Name switch
               {
                   "$Main" => true,
                   "<$Main>$" => true,
                   _ => false
               };
```

The generated entry point for top-level is `"<Main>$"` which isn't covered by the method.

## Actions needed

- [x] Investigate how the tests passed when the method name is `"<Main>$"` which is not covered.
- [ ] Use the public API from Roslyn:

    https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Compilers/Core/Portable/Symbols/WellKnownMemberNames.cs#L348-L356

Youssef1313:
Related to #4253 (keeping the issue open to track using the compiler API)

### DIFF
--- a/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
@@ -690,7 +690,7 @@ namespace Analyzer.Utilities.Extensions
                methodSymbol.Name switch
                {
                    "$Main" => true,
-                   "<$Main>$" => true,
+                   "<Main>$" => true,
                    _ => false
                };
     }


eng/Versions.props
```diff
@@ -23,7 +23,7 @@
     <MicrosoftNetCompilersToolsetVersion>$(MicrosoftNETCoreCompilersPackageVersion)</MicrosoftNetCompilersToolsetVersion>
     <CodeStyleAnalyersVersion>$(MicrosoftNETCoreCompilersPackageVersion)</CodeStyleAnalyersVersion>
     <!-- Roslyn -->
-    <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisVersion>3.8.0-3.final</MicrosoftCodeAnalysisVersion>
```
mavasani, line 26: Please don't bump the repo version so aggressively. This prevents all analyzer packages to run on older bits.

mavasani (changes requested):
As commented. Just upgrading to this constant API does not warrant bumping up the code analysis version for the entire repo.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -686,12 +686,6 @@ public static bool IsAutoPropertyAccessor(this IMethodSymbol methodSymbol)
         /// </summary>
         public static bool IsTopLevelStatementsEntryPointMethod([NotNullWhen(true)] this IMethodSymbol? methodSymbol)
             => methodSymbol?.ContainingType.IsTopLevelStatementsEntryPointType() == true &&
-               methodSymbol.IsStatic &&
-               methodSymbol.Name switch
-               {
-                   "$Main" => true,
-                   "<$Main>$" => true,
```
mavasani, line 693: We can just change this to `"<Main>$"`
https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Compilers/Core/Portable/Symbols/WellKnownMemberNames.cs#L351

src/Utilities/Compiler/Extensions/INamedTypeSymbolExtensions.cs
```diff
@@ -275,11 +275,6 @@ public static bool IsXUnitTestAttribute(this INamedTypeSymbol attributeClass, Co
         /// Check if the given <paramref name="typeSymbol"/> is an implicitly generated type for top level statements.
         /// </summary>
         public static bool IsTopLevelStatementsEntryPointType([NotNullWhen(true)] this INamedTypeSymbol? typeSymbol)
-            => typeSymbol?.IsStatic == true && typeSymbol.Name switch
-            {
-                "$Program" => true,
-                "<Program>$" => true,
```
mavasani, line 281: This seems to be just fine.

mavasani (commented):
Seems like this needs a single char deletion fix :-)

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Youssef1313: @mavasani When the code in tests is compiled, it's compiled against an older compiler version that produces `$Main` instead of `<Main>$`. I wanted to remove the old name from the extension method, so that required a newer version of the compiler. But as this is causing problems, I'll revert.

Youssef1313: Do you want to keep the issue open to track using the compiler API when it's good time to update the version?

mavasani: If you need new bits only for test you can bump up the below: https://github.com/dotnet/roslyn-analyzers/blob/193a3b623223f6c4a015d9faafb874fd739dff61/eng/Versions.props#L28

We should probably rename that property name to suggest that is for tests.

mavasani: > Do you want to keep the issue open to track using the compiler API when it's good time to update the version?

That sounds reasonable. Thanks.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -686,12 +686,6 @@ public static bool IsAutoPropertyAccessor(this IMethodSymbol methodSymbol)
...
-                   _ => false
-               };
+               methodSymbol.IsStatic && methodSymbol.Name == "<Main>$";
```
mavasani, line 689: Please retain the old logic as `"$Main"` was used in older compiler bits which is supported by the analyzer package. I think you should only fix the typo here.

Youssef1313: I thought that's not important as the compiler version has this isn't a stable release. I'll revert the last 2 commits.

mavasani: > as the compiler version has this isn't a stable release

Ah, if you are sure it wasn't a stable release, then it would be fine to remove those cases. By the way, there are test failures.

Youssef1313: > Ah, if you are sure it wasn't a stable release, then it would be fine to remove those cases

@mavasani I think it was shipped only in a preview version of Visual Studio. But might be better to confirm with the compiler team. Not sure who may be able help.

src/Utilities/Compiler/Extensions/INamedTypeSymbolExtensions.cs
```diff
@@ -275,11 +275,6 @@ public static bool IsXUnitTestAttribute(this INamedTypeSymbol attributeClass, Co
...
-                _ => false
-            };
+            => typeSymbol?.IsStatic == true && typeSymbol.Name = "<Program>$";
```
Youssef1313, line 278:
```suggestion
            => typeSymbol?.IsStatic == true && typeSymbol.Name == "<Program>$";
```

Evangelink:
@Youssef1313 I am having some trouble understanding the link between the name of the PR and its content. What am I missing?

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Youssef1313: @mavasani The current test failures seems unrelated to top-level statements. Can you have a look?

For example:

https://github.com/dotnet/roslyn-analyzers/blob/a23fde8d03fcec0c8e5beb6b1f3882048ab8435b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ImplementIDisposableCorrectlyTests.cs#L745-L798

This test doesn't produce any diagnostics, while 2 are expected. The number of failing tests is **huge**.

Youssef1313:
> 
> 
> @Youssef1313 I am having some trouble understanding the link between the name of the PR and its content. What am I missing?

Did you mean the "unspeakable" part of the title? That's what the compiler team calls these generated members names.

Evangelink:
> 
> 
> > @Youssef1313 I am having some trouble understanding the link between the name of the PR and its content. What am I missing?
> 
> Did you mean the "unspeakable" part of the title? That's what the compiler team calls these generated members names.

Oh I see, thanks :)

Youssef1313:
Given these test failures here (whose reason is known), I won't update the version to use the compiler API and just fix the typo.

mavasani (approved):

PR #4259, opened 2020-09-30
Create markdownlint workflow

### LINKED ISSUE #4258: Setup markdown linter for CI
> This is a one-off change. Regarding the various documentation files this is a manual modification but we don't have any automatic linter that would prevent new errors to be introduced.

@mavasani @Evangelink Do you want a linter workflow for PRs?
dotnet/docs is using one, it should be copy/paste to set the workflow for this repo.

https://github.com/dotnet/docs/blob/master/.github/workflows/markdownlint.yml
https://github.com/dotnet/docs/blob/master/.github/workflows/markdownlint-problem-matcher.json

We can temporarily exclude `AnalyzerReleases.*.md`.

_Originally posted by @Youssef1313 in https://github.com/dotnet/roslyn-analyzers/pull/4141#issuecomment-701539631_

Youssef1313:
Fixes #4258


Youssef1313:
@mavasani Are workflows disabled in this repository?

mavasani:
@jmarolf should know

.github/workflows/markdownlint-problem-matcher.json
```diff
@@ -0,0 +1,17 @@
+{
```
Evangelink, line 1: Can you point me to some doc for this json configuration?

Youssef1313: @Evangelink For the file you're currently commenting on, I don't know unfortunately. The information I have on it that it makes GitHub prints out the failures in the files changed tab. which is easier than checking the logs.

Youssef1313: @Evangelink  Found some doc:

https://github.com/actions/toolkit/blob/7ae226462403d6bceb659e29211bc8e812843503/docs/problem-matchers.md

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -2676,7 +2676,7 @@ Applications available over HTTPS must use secure cookies.
 |CodeFix|False|
 ---
 
-## [CA5383](https://docs.microsoft.com/visualstudio/code-quality/ca5383): Ensure Use Secure Cookies In ASP.Net Core
+## [CA5383](https://docs.microsoft.com/visualstudio/code-quality/ca5383): Ensure Use Secure Cookies In ASP.NET Core
```
Youssef1313, line 2679: These changes were automated when I had proper_names from dotnet/docs config. I think parallel changes to resx is needed.

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
```diff
@@ -198,7 +198,7 @@ The 'helpLinkUri' value is used to show information when this diagnostic in the
 
 ## RS1016: Code fix providers should provide FixAll support
 
-A CodeFixProvider should provide FixAll support to enable users to fix multiple instances of the underlying diagnostic with a single code fix. See documenation at https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md for further details.
+A CodeFixProvider should provide FixAll support to enable users to fix multiple instances of the underlying diagnostic with a single code fix. See documenation at <https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md> for further details.
```
Evangelink, line 201: Is it really the right fix? Isn't it better to do something like:

```suggestion
A CodeFixProvider should provide FixAll support to enable users to fix multiple instances of the underlying diagnostic with a single code fix. See [documenation](https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md) for further details.
```

src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
```diff
@@ -10,109 +12,114 @@ Release tracking analyzer enables third party analyzer packages to define analyz
...
 
 ## Example
+
 Consider the following example:
 
 1. `AnalyzerReleases.Shipped.md`:
 
-```
-## Release 1.0
+   ```md
```
Evangelink, line 32: Not sure about these leading spaces.

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
Youssef1313: Hyperlinking "documentation" is certainly better. The other fix was automated using `markdownlint -f` command.

src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
Youssef1313: @Evangelink The leading spaces makes the content to appear correctly under the list item (i.e. there is a **visual** change here). The absence of the leading spaces also violates some rule, basically because the code snippet becomes outside the list item, which means the whole list is ended. Hence, the other list item that starts with `2.` is considered as a new list (instead of becoming a new list *item* under the same list). The linter triggers an error when a new list starts with number other than `1`.

Youssef1313: The HTML before the leading spaces looks like this:

image

Note there are 2 `<ol>` tags, which the second one starts with 2.

Youssef1313:
markdownlint now passes.

.github/workflows/markdownlint-problem-matcher.json
Evangelink: Thanks!

docs/Analyzer reference page template.md
```diff
@@ -28,4 +28,4 @@
 
 ## Related rules
 
-[RULEID: Friendly related rule name](https://stable-uris-r-us.com/MyRuleId_MyFriendlyRuleName.md)
\ No newline at end of file
+[RULEID: Friendly related rule name](https://stable-uris-r-us.com/MyRuleId_MyFriendlyRuleName.md)
```
Evangelink, line 31: Is it a line ending change?

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
```diff
@@ -198,7 +198,7 @@ The 'helpLinkUri' value is used to show information when this diagnostic in the
 
 ## RS1016: Code fix providers should provide FixAll support
 
-A CodeFixProvider should provide FixAll support to enable users to fix multiple instances of the underlying diagnostic with a single code fix. See documenation at https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md for further details.
+A CodeFixProvider should provide FixAll support to enable users to fix multiple instances of the underlying diagnostic with a single code fix. See [documenation](https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md) for further details.
```
Evangelink, line 201: Giving some thoughts about it and keeping in mind that this file is auto-generated I think that it might be easier to go for your original solution (i.e. `<` + url + `>`).

Youssef1313: @Evangelink If this is auto-generated, we shouldn't edit it manually. The original code that generates it should be modified.

mavasani: Yes, this is auto-generated. You should instead modify the generator as suggested by @Evangelink  https://github.com/dotnet/roslyn-analyzers/blob/dc649547a3d3a80d4e8caffc7bdd31cf3e70cb92/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs#L271

To re-execute the generator, invoke `msbuild /t:pack` on the entire solution or the selected nuget project within https://github.com/dotnet/roslyn-analyzers/tree/master/nuget

src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
mavasani: Please revert the changes here. This md file is an additional file used by the analyzer, and I am not sure the analyzer that parses it handles these spaces. Also the code fix that produces this md will not add these spaces, so user would try hand editing, which is undesirable.

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md
```diff
@@ -2688,7 +2688,7 @@ By default, the Trusted Root Certification Authorities certificate store is conf
 |CodeFix|False|
 ---
 
-## [CA5382](https://docs.microsoft.com/visualstudio/code-quality/ca5382): Use Secure Cookies In ASP.Net Core
+## [CA5382](https://docs.microsoft.com/visualstudio/code-quality/ca5382): Use Secure Cookies In ASP.NET Core
```
mavasani, line 2691: This is auto-generated file, please revert changes. These are actually resource strings provided by the analyzer.

src/Microsoft.CodeAnalysis.FxCopAnalyzers/RulesMissingDocumentation.md
```diff
@@ -2,5 +2,5 @@
 
 Rule ID | Missing Help Link | Title |
 --------|-------------------|-------|
-CA1416 | https://docs.microsoft.com/visualstudio/code-quality/ca1416 | Validate platform compatibility |
+CA1416 | <https://docs.microsoft.com/visualstudio/code-quality/ca1416> | Validate platform compatibility |
```
mavasani, line 5: This is auto-generated file by tool, please revert and change the tool if required: https://github.com/dotnet/roslyn-analyzers/blob/dc649547a3d3a80d4e8caffc7bdd31cf3e70cb92/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs#L467

src/Microsoft.CodeAnalysis.VersionCheckAnalyzer/Microsoft.CodeAnalysis.VersionCheckAnalyzer.md
```diff
@@ -2,7 +2,7 @@
 
 ## CA9999: Analyzer version mismatch
 
-Analyzers in this package require a certain minimum version of Microsoft.CodeAnalysis to execute correctly. Refer to https://docs.microsoft.com/visualstudio/code-quality/install-fxcop-analyzers#fxcopanalyzers-package-versions to install the correct analyzer version.
+Analyzers in this package require a certain minimum version of Microsoft.CodeAnalysis to execute correctly. Refer to <https://docs.microsoft.com/visualstudio/code-quality/install-fxcop-analyzers#fxcopanalyzers-package-versions> to install the correct analyzer version.
```
mavasani, line 5: Revert

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
Evangelink: I know this is auto-generated. My point was to actually say that as it is auto-generated we could easily do a regex replacement to wrap the url in `<>` but it would be hard to do the change you have done.

.markdownlint.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "proper-names": {
+        "code_blocks": false,
+        "names": [
+            "ASP.NET",
+            "PowerShell",
+            "C#"
+        ]
+    }
+}
```
sharwell, line 17: 📝 The indentation in this file does not adhere to **.editorconfig** for this repository

docs/Analyzer reference page template.md
sharwell: It adds the final newline

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
sharwell: 💭 I think I would expect this form for external links:

```
[Title (site)](link)
```

Where in this case `site` would either say `github.com` or, since it's a different project on the same site, `dotnet/roslyn`.

src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
sharwell: @mavasani This is not going to render with spaces, so the result for users is unchanged.

src/Microsoft.CodeAnalysis.FxCopAnalyzers/Microsoft.CodeAnalysis.FxCopAnalyzers.md
sharwell: @Youssef1313 If you need to make a change here because something is wrong, you can do so by updating the original source that produces this file.

src/PerformanceSensitiveAnalyzers/PerformanceSensitiveAnalyzers.md
```diff
@@ -1 +1 @@
-This is forked from [01c6dd2bf7dc509de289da60b78df8440ce4c11d@Microsoft/RoslynClrHeapAllocationAnalyzer](https://github.com/Microsoft/RoslynClrHeapAllocationAnalyzer/commit/01c6dd2bf7dc509de289da60b78df8440ce4c11d)
\ No newline at end of file
+This is forked from [01c6dd2bf7dc509de289da60b78df8440ce4c11d@Microsoft/RoslynClrHeapAllocationAnalyzer](https://github.com/Microsoft/RoslynClrHeapAllocationAnalyzer/commit/01c6dd2bf7dc509de289da60b78df8440ce4c11d)
```
sharwell, line 1:
```suggestion
This is forked from Microsoft/RoslynClrHeapAllocationAnalyzer@01c6dd2bf7dc509de289da60b78df8440ce4c11d
```

sharwell:
@Youssef1313 I marked this as a draft. For generated files, you have two options:

1. Exclude the files from analysis, revert the changes made to them, and file bug(s) to fix the generation process
2. Fix the generation process here and generation the files using the updated tool

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
Youssef1313: > 
> 
> 💭 I think I would expect this form for external links:
> 
> ```
> [Title (site)](link)
> ```
> 
> Where in this case `site` would either say `github.com` or, since it's a different project on the same site, `dotnet/roslyn`.

This message part is auto generated from resx. It can show outside markdown context. So the message itself can't be updated to use markdown syntax. The code that generates the markdown could possibly find links and convert them, but how it should decide the display text for the link?

mavasani:

Youssef1313:
@mavasani Since GitHub Actions are disabled, are you okay if I switched the markdownlint to azure-pipeline job?

mavasani:
> Since GitHub Actions are disabled, are you okay if I switched the markdownlint to azure-pipeline job?

Sounds good

mavasani (approved):

mavasani:
@Youssef1313 Can you please resolve the merge conflicts?

mavasani:
@Youssef1313 There are failures in the linter test

mavasani:
Ping @Youssef1313

Youssef1313:
I'm not sure how to fix the generation process to detect the current violation and fix it.
Currently the generated markdown contains `Something<T>`, and the `<T>` isn't displayed because parsers treat that as an HTML tag. Will take a look on how to properly escape the `<`.

Evangelink:
I think that's probably good enough, if at some point we face an issue we will have a case to refine the match pattern. As it's a good practice to prefix generic type with `T` we might want to look for `<T....>`, WDYT?

Evangelink:
Or we could also open a FP ticket on the markdown linter so that their algorithm is improved to not report here :)

Evangelink:
@Youssef1313 Could you manage to fix the remaining issues so that we can merge this?

mavasani:
> @Youssef1313 Could you manage to fix the remaining issues so that we can merge this?

Agreed, this PR will very quickly go out-of-date with more changes in the repo.

Youssef1313:
Will work on it soon.

azure-pipelines.yml
```diff
@@ -110,4 +110,13 @@ jobs:
...
\ No newline at end of file
+      condition: always()
+
+- job: Markdownlint
+  pool:
+      vmImage: ubuntu-latest
+  steps:
+    - script: sudo npm install -g markdownlint-cli
+      displayName: Install markdownlint-cli
+    - script: markdownlint '**/*.md' -i '**/AnalyzerReleases.*.md' -i 'docs/Analyzer reference page template.md' -i 'eng/common/**/*.md'
```
Youssef1313, line 121: The `-i` switch used here is "ignore". The currently ignored files have some violations. We probably want to fix them and enable lint for these files in the future.

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -573,7 +573,8 @@ void createAnalyzerRulesMissingDocumentationFile()
                         helpLinkUri = $"<{helpLinkUri}>";
                     }
 
-                    var line = $"{ruleId} | {helpLinkUri} | {descriptor.Title.ToString(CultureInfo.InvariantCulture)} |";
+                    var escapedTitle = descriptor.Title.ToString(CultureInfo.InvariantCulture).Replace("<T>", "\\<T>");
```
Evangelink, line 576: Do we want to wait for an issue to appear or try to be a little more generic here? We could potentially do a regex replace of something like `<T[a-zA-Z]*?>`.

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -361,6 +361,10 @@ void createAnalyzerDocumentationFile()
                     description = Regex.Replace(description, "(\r?\n)", "$1$1");
                     // Escape generic arguments to ensure they are not considered as HTML elements
                     description = Regex.Replace(description, "(<.+?>)", "\\$1");
+                    // Add angle brackets around links to prevent violating MD034:
+                    // https://github.com/DavidAnson/markdownlint/blob/82cf68023f7dbd2948a65c53fc30482432195de4/doc/Rules.md#md034---bare-url-used
+                    // Regex taken from: https://github.com/DavidAnson/markdownlint/blob/59eaa869fc749e381fe9d53d04812dfc759595c6/helpers/helpers.js#L24
+                    description = Regex.Replace(description, @"(?:http|ftp)s?:\/\/[^\s\]""']*(?:\/|[^\s\]""'\W])", "<$0>");
```
Evangelink, line 367: You might want to add `?` after the `*` in the regex for better results.

```suggestion
                    description = Regex.Replace(description, @"(?:http|ftp)s?:\/\/[^\s\]""']*?(?:\/|[^\s\]""'\W])", "<$0>");
```

Youssef1313: @Evangelink I took the regex as-is from markdownlint repository.

Do you have an example where the current regex won't work but the modified will? In that case it would be a possible bug in markdownlint then.

Evangelink: It won't be a bug, just a performance hotpoint. This is related to some backtracking, I don't remember where is the doc.

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
Youssef1313: @Evangelink I wanted to keep it simple, if issues appeared later, either your regex or the regex from markdownlint repository should work.

sharwell: Why not escape all `<` characters?

```suggestion
                    var escapedTitle = descriptor.Title.ToString(CultureInfo.InvariantCulture).Replace("<", "\\<");
```

Youssef1313: @sharwell Sounds reasonable in the context for rule title.

Evangelink:
Looks like you have been missing the `msbuild /t:pack`

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
Youssef1313: @Evangelink I reverted your change here, it caused the closing angle bracket to be in the wrong position.

image

Evangelink: Oh it makes sense based on the rest of the regex, my bad. To use this non-greedy version we would need to tweak the rest. I guess that's fine as-is.

azure-pipelines.yml
Evangelink: Do we have a ticket tracking this remaining task?

Evangelink (approved):

mavasani:
Thank you @Youssef1313!

PR #4262, opened 2020-10-01
Loc Checkin 9/30/2020

cristianosuzuki77:
Latest translations.

PR #4267, opened 2020-10-01
Use WithArguments instead of WithMessage in unit test

buyaa-n:
For fixing test failures for localization PR https://github.com/dotnet/roslyn-analyzers/pull/4262#issuecomment-702287401

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
@@ -86,8 +86,8 @@ class Test
     }
 }" + MockAttributesCsSource + MockOperatingSystemApiSource;
 
-            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(15, 9)
-                .WithMessage("'Test.Api()' is unsupported on 'windows'"));
+            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.UnsupportedOsRule)
+                .WithLocation(15, 9).WithArguments("Test.Api()", "windows"));
         }
 
         [Fact]
@@ -210,10 +210,10 @@ namespace PlatformCompatDemo.Bugs.GuardsAroundSupported
 }" + MockAttributesCsSource + MockOperatingSystemApiSource;
 
             await VerifyAnalyzerAsyncCs(source,
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(15, 17).WithMessage("'Target.SupportedOnWindowsAndBrowser()' is supported on 'windows'"),
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(16, 17).WithMessage("'Target.SupportedOnWindows10AndBrowser()' is supported on 'windows' 10.0 and later"),
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(24, 17).WithMessage("'Target.SupportedOnWindows10AndBrowser()' is supported on 'browser'"),
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(48, 17).WithMessage("'Target.SupportedOnWindows10AndBrowser()' is supported on 'browser'"));
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(15, 17).WithArguments("Target.SupportedOnWindowsAndBrowser()", "browser"),
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(16, 17).WithArguments("Target.SupportedOnWindows10AndBrowser()", "browser"),
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(24, 17).WithArguments("Target.SupportedOnWindows10AndBrowser()", "browser"),
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(48, 17).WithArguments("Target.SupportedOnWindows10AndBrowser()", "browser"));
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -210,10 +210,10 @@ class Target
...
 
             await VerifyAnalyzerAsyncCs(source,
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(15, 17).WithMessage("'Target.SupportedOnWindowsAndBrowser()' is supported on 'windows'"),
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(16, 17).WithMessage("'Target.SupportedOnWindows10AndBrowser()' is supported on 'windows' 10.0 and later"),
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(24, 17).WithMessage("'Target.SupportedOnWindows10AndBrowser()' is supported on 'browser'"),
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(48, 17).WithMessage("'Target.SupportedOnWindows10AndBrowser()' is supported on 'browser'"));
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(15, 17).WithArguments("Target.SupportedOnWindowsAndBrowser()", "browser"),
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(16, 17).WithArguments("Target.SupportedOnWindows10AndBrowser()", "browser"),
```
Youssef1313, line 214: I would be surprised if this test passed. Shouldn't "10.0 and later" be passed as a third argument? and the rule should remain `SupportedOsVersionRule`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -86,8 +86,8 @@ void Api()
     }
 }" + MockAttributesCsSource + MockOperatingSystemApiSource;
 
-            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(15, 9)
-                .WithMessage("'Test.Api()' is unsupported on 'windows'"));
+            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.UnsupportedOsRule)
+                .WithLocation(15, 9).WithArguments("Test.Api()", "windows"));
```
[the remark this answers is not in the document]
mavasani: Because all rules have same ID and just differed by message, and the prior implementation was explicitly overwriting the message.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -1055,7 +1055,7 @@ public static void WindowsSpecificApi()
 
             await VerifyAnalyzerAsyncCs(source,
                 VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.UnsupportedOsRule).WithLocation(16, 13)
-                .WithMessage("'Some.WindowsSpecificApi()' is supported on 'windows' 10.0 and later"));
+                .WithArguments("Some.WindowsSpecificApi()", "windows", "10.0"));
```
Youssef1313, line 1058: Should be `SupportedOsVersionRule`?
As mentioned in previous comment, the test - for some reason that I don't know - will pass on its current state.

mavasani (approved):
LGTM

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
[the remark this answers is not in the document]
buyaa-n: Seems when there is a message check it is not checking the rule and arguments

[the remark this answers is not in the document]
Youssef1313: @mavasani In that case I think it's more descriptive (and less misleading) to pass the ID only to this method (if it has an overload that accepts string id).

There is no need to address that in this PR though to unblock the loc PR.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: Well there is 2 diagnostics one for browser and the other for Windows 10.0 and later, as we can add only one markup for diagnostic the other is here, so there is nothing really change for the test

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: Good catch/point, maybe it passed because has diagnostic for it warns for `UnsupportedOsRule` too, but yes `SupportedOsVersionRule` is the correct one

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
Youssef1313: @buyaa-n Does that mean when there are multiple diagnostics in the same span, the test need to only verify *one* of them?

buyaa-n: > @buyaa-n Does that mean when there are multiple diagnostics in the same span, the test need to only verify one of them?

Not test need to verify both of them, here one is verified by the markup syntax `[|ExpectedDiagnostic|]` in the source, i don't know if there is a way to mark there is 2 diagnostics, so the other one is added here as a parameter.

Youssef1313: @buyaa-n Got it. But if I understand that correctly, there is no way to test the **arguments** for both diagnostics. One of them is tested against its existence, and the other is tested against both existence and correctness of arguments.

There could be a room of improvement here in the tests infrastructure.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -86,8 +86,8 @@ void Api()
     }
 }" + MockAttributesCsSource + MockOperatingSystemApiSource;
 
-            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsVersionRule).WithLocation(15, 9)
-                .WithMessage("'Test.Api()' is unsupported on 'windows'"));
```
Evangelink, line 90: @mavasani Would it be worth banning the `WithMessage` API? It seems that we favor using `WithArguments` to avoid translation issues.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: > there is no way to test the arguments for both diagnostics. One of them is tested against its existence, and the other is tested against both existence and correctness of arguments.

You can do that for both, but for markup syntax you cannot verify the arguments inline, but you can number the markups and add corresponding arguments later for each markup. for this case for me it was enough to verify the message/location/arguments for only one of the diagnostics

sharwell: You can verify arguments using markup by using the markup syntax `{|#0:code|}` and then constructing the expected diagnostic with `.WithLocation(0).WithArguments(...)`. Multiple explicit diagnostics can reference the same location by identifier. You can also have two diagnostics at the same location by nesting: `[|[|code|]|]`, or `{|#0:{|#1:code|} more code|}`.

sharwell: As an aside, never use the two-argument form of `WithLocation`. Either use `WithLocation(int)` or use `WithSpan` to specify full spans.

Youssef1313: > As an aside, never use the two-argument form of `WithLocation`. Either use `WithLocation(int)` or use `WithSpan` to specify full spans.

I'll file an issue to improve this later.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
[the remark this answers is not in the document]
mavasani: A better way for such cases is to tag each location in source with an index, and then use Descriptor + Arguments to validate message. See below example: https://github.com/dotnet/roslyn-analyzers/blob/2e221c3b000bbd7d5e618115b3e43ad831fca32b/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs#L67

https://github.com/dotnet/roslyn-analyzers/blob/2e221c3b000bbd7d5e618115b3e43ad831fca32b/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzerTests.cs#L4469-L4471

mavasani:
@buyaa-n I am merging this to unblock loc PR. Feel free to cleanup the tests with a follow-up PR, if required.

Issue #4235, opened 2020-09-28
Improve code fixer heuristic employed by CA1835


stephentoub:
### Analyzer

**Diagnostic ID**: [CA1835](https://docs.microsoft.com/visualstudio/code-quality/ca1835)

### Describe the improvement

The analyzer currently suggests changing a call like:
```C#
await stream.ReadAsync(buffer, 0, buffer.Length);
```
to instead be:
```C#
await stream.ReadAsync(buffer.AsMemory(0, buffer.Length));
```

While correct, for the common case of a 0 offset and a length equal to the buffer's length, it is more complex than needed (with a bit more overhead, and could instead be just:
```C#
await stream.ReadAsync(buffer);
```

### Describe suggestions on how to achieve the rule

Apply a simpler code fix if the second argument is 0 and the third argument is just a Length property accessor on the first argument.

cc: @carlossanlop, @jeffhandley

carlossanlop:
Thanks @stephentoub, this is a good improvement to consider.

It would apply for both `ReadAsync` and `WriteAsync`, since in both cases, it would make sense to use `0` as the `offset` parameter, and `buffer.Length` as the `count` parameter.

Question - How smart do we want this to be? For example, if the user saves the `buffer.Length` into a variable, and passes this variable as the `count` parameter, should the analyzer check if the variable's original value is `buffer.Length`?:
  ```cs
  int length = buffer.Length;
  stream.WriteAsync(buffer, 0, length);
  ```
My gut says that maybe we don't want to invest in adding this level of complexity, but let me know if we should.

I'll assign this to myself.

stephentoub:
> if the user saves the buffer.Length into a variable, and passes this variable as the count parameter, should the analyzer check if the variable's original value is buffer.Length?:

I don't think that's necessary or worth additional complexity (at least not until we get evidence that it would meaningfully help). The common case, at least for ReadAsync, is to just do `buffer, 0, buffer.Length`.

PR #4269, opened 2020-10-01
Fix to improve code fixer heuristic employed by CA1835

carlossanlop:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4235

For invocations of ReadAsync/WriteAsync that consider the full buffer (from `0` to `buffer.Length`):
```cs
await stream.ReadAsync(buffer, 0, buffer.Length);
```

Instead of suggesting this:
```cs
await stream.ReadAsync(buffer.AsMemory(0, buffer.Length));
```

We now want to suggest this:
```cs
await stream.ReadAsync(buffer);
```

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -58,14 +58,14 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
                 // First argument should be an identifier name node
                 bufferValueNode is IdentifierNameSyntax firstArgumentIdentifierName &&
                 // Second argument should be a literal expression node with a constant value of zero
-                model.GetConstantValue(offsetValueNode) is Optional<object> optionalValue && optionalValue.HasValue && optionalValue.Value is int value && value == 0 &&
+                model.GetConstantValue(offsetValueNode) is Optional<object> optionalValue && optionalValue.HasValue && optionalValue.Value is 0 &&
                 // Third argument should be a member access node...
                 countValueNode is MemberAccessExpressionSyntax thirdArgumentMemberAccessExpression &&
                 thirdArgumentMemberAccessExpression.Expression is IdentifierNameSyntax thirdArgumentIdentifierName &&
                 // whose identifier is that of the first argument...
                 firstArgumentIdentifierName.Identifier.ValueText == thirdArgumentIdentifierName.Identifier.ValueText &&
                 // and the member name is `Length`
-                thirdArgumentMemberAccessExpression.Name.Identifier.ValueText == "Length";
+                thirdArgumentMemberAccessExpression.Name.Identifier.ValueText == WellKnownMemberNames.LengthPropertyName;
         }
     }
 }
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -52,15 +52,13 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
             return false;
         }
 
-        protected override bool IsPassingZeroAndBufferLength(SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode)
+        protected override bool IsPassingZeroAndBufferLength(SemanticModel model, SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode)
         {
             return
                 // First argument should be an identifier name node
                 bufferValueNode is IdentifierNameSyntax firstArgumentIdentifierName &&
-                // Second argument should be a literal expression node...
-                offsetValueNode is LiteralExpressionSyntax secondArgumentLiteralExpression &&
-                // and its value should be zero
-                secondArgumentLiteralExpression.Token.ValueText == "0" &&
+                // Second argument should be a literal expression node with a constant value of zero
+                model.GetConstantValue(offsetValueNode) is Optional<object> optionalValue && optionalValue.HasValue && optionalValue.Value is int value && value == 0 &&
                 // Third argument should be a member access node...
                 countValueNode is MemberAccessExpressionSyntax thirdArgumentMemberAccessExpression &&
                 thirdArgumentMemberAccessExpression.Expression is IdentifierNameSyntax thirdArgumentIdentifierName &&
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -51,5 +51,23 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
             }
             return false;
         }
+
+        protected override bool IsPassingZeroAndBufferLength(SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode)
+        {
+            return
+                // First argument should be an identifier name node
+                bufferValueNode is IdentifierNameSyntax firstArgumentIdentifierName &&
+                // Second argument should be a literal expression node...
+                offsetValueNode is LiteralExpressionSyntax secondArgumentLiteralExpression &&
+                // and its value should be zero
+                secondArgumentLiteralExpression.Token.ValueText == "0" &&
+                // Third argument should be a member access node...
+                countValueNode is MemberAccessExpressionSyntax thirdArgumentMemberAccessExpression &&
+                thirdArgumentMemberAccessExpression.Expression is IdentifierNameSyntax thirdArgumentIdentifierName &&
+                // whose identifier is that of the first argument...
+                firstArgumentIdentifierName.Identifier.ValueText == thirdArgumentIdentifierName.Identifier.ValueText &&
+                // and the member name is `Length`
+                thirdArgumentMemberAccessExpression.Name.Identifier.ValueText == "Length";
+        }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -39,7 +39,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
         // Verifies if the user passed `0` as the 1st argument (`offset`) and `buffer.Length` as the 2nd argument (`count`),
         // where `buffer` is the name of the variable passed as the 0th argument.
-        protected abstract bool IsPassingZeroAndBufferLength(SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode);
+        protected abstract bool IsPassingZeroAndBufferLength(SemanticModel model, SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode);
 
         public sealed override ImmutableArray<string> FixableDiagnosticIds =>
             ImmutableArray.Create(PreferStreamAsyncMemoryOverloads.RuleId);
@@ -94,7 +94,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             string title = MicrosoftNetCoreAnalyzersResources.PreferStreamAsyncMemoryOverloadsTitle;
 
-            Task<Document> createChangedDocument(CancellationToken _) => FixInvocation(doc, root, invocation, invocation.TargetMethod.Name,
+            Task<Document> createChangedDocument(CancellationToken _) => FixInvocation(model, doc, root,
+                                                         invocation, invocation.TargetMethod.Name,
                                                          bufferOperation.Value.Syntax, isBufferNamed,
                                                          offsetOperation.Value.Syntax, isOffsetNamed,
                                                          countOperation.Value.Syntax, isCountNamed,
@@ -108,7 +109,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 context.Diagnostics);
         }
 
-        private Task<Document> FixInvocation(Document doc, SyntaxNode root, IInvocationOperation invocation, string methodName,
+        private Task<Document> FixInvocation(SemanticModel model, Document doc, SyntaxNode root,
+            IInvocationOperation invocation, string methodName,
             SyntaxNode bufferValueNode, bool isBufferNamed,
             SyntaxNode offsetValueNode, bool isOffsetNamed,
             SyntaxNode countValueNode, bool isCountNamed,
@@ -122,7 +124,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             // Depending on the arguments being passed to Read/WriteAsync, it's the substitution we will make
             SyntaxNode replacedInvocationNode;
 
-            if (IsPassingZeroAndBufferLength(bufferValueNode, offsetValueNode, countValueNode))
+            if (IsPassingZeroAndBufferLength(model, bufferValueNode, offsetValueNode, countValueNode))
             {
                 // Remove 0 and buffer.length
                 replacedInvocationNode =
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
@@ -37,6 +37,10 @@ namespace Microsoft.NetCore.Analyzers.Runtime
         // Verifies if a namespace has already been added to the usings/imports list.
         protected abstract bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> importList);
 
+        // Verifies if the user passed `0` as the 1st argument (`offset`) and `buffer.Length` as the 2nd argument (`count`),
+        // where `buffer` is the name of the variable passed as the 0th argument.
+        protected abstract bool IsPassingZeroAndBufferLength(SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode);
+
         public sealed override ImmutableArray<string> FixableDiagnosticIds =>
             ImmutableArray.Create(PreferStreamAsyncMemoryOverloads.RuleId);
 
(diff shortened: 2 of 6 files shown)


src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -51,5 +51,23 @@ protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> impo
...
+
+        protected override bool IsPassingZeroAndBufferLength(SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode)
+        {
+            return
+                // First argument should be an identifier name node
+                bufferValueNode is IdentifierNameSyntax firstArgumentIdentifierName &&
+                // Second argument should be a literal expression node...
+                offsetValueNode is LiteralExpressionSyntax secondArgumentLiteralExpression &&
+                // and its value should be zero
+                secondArgumentLiteralExpression.Token.ValueText == "0" &&
```
carlossanlop, line 63: @mavasani is this the proper way to check if the value of the token is exactly `0`? Same question for line 70 to check for `Length`.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -51,5 +51,23 @@ protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> impo
...
+                // Third argument should be a member access node...
+                countValueNode is MemberAccessExpressionSyntax thirdArgumentMemberAccessExpression &&
+                thirdArgumentMemberAccessExpression.Expression is IdentifierNameSyntax thirdArgumentIdentifierName &&
+                // whose identifier is that of the first argument...
+                firstArgumentIdentifierName.Identifier.ValueText == thirdArgumentIdentifierName.Identifier.ValueText &&
```
carlossanlop, line 68: Similarly, is this the best way to verify that both the `buffer` argument and the `buffer.Length` argument are the same variable?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: A better approach is to check for constant value for the syntax using this API: http://sourceroslyn.io/#Microsoft.CodeAnalysis/Compilation/SemanticModel.cs,743

mavasani: You can also remove the check `offsetValueNode is LiteralExpressionSyntax secondArgumentLiteralExpression`, instead just check constant value of syntax node for second argument.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
mavasani: I think this is the best possible heuristic..

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -112,26 +116,39 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
-                namedStartNode.WithTriviaFrom(offsetValueNode),
-                namedLengthNode.WithTriviaFrom(countValueNode));
+            // Depending on the arguments being passed to Read/WriteAsync, it's the substitution we will make
+            SyntaxNode replacedInvocationNode;
 
-            // Generate the new buffer argument, ensuring we include the argument name if the user originally indicated one
-            SyntaxNode namedBufferWithAsMemoryInvocationNode =
-                    (isBufferNamed ? generator.Argument(name: "buffer", RefKind.None, asMemoryInvocationNode) : asMemoryInvocationNode)
-                .WithTriviaFrom(bufferValueNode);
+            if(IsPassingZeroAndBufferLength(bufferValueNode, offsetValueNode, countValueNode))
```
Evangelink, line 125:
```suggestion
            if (IsPassingZeroAndBufferLength(bufferValueNode, offsetValueNode, countValueNode))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -112,26 +116,39 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            {
+                // Remove 0 and buffer.length
+                replacedInvocationNode =
+                        (isBufferNamed ? generator.Argument(name: "buffer", RefKind.None, bufferValueNode) : bufferValueNode)
```
Evangelink, line 129:
```suggestion
                    (isBufferNamed ? generator.Argument(name: "buffer", RefKind.None, bufferValueNode) : bufferValueNode)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -112,26 +116,39 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                // Generate an invocation of the AsMemory() method from the byte array object, using the correct named arguments
+                SyntaxNode asMemoryExpressionNode = generator.MemberAccessExpression(bufferValueNode.WithoutTrivia(), memberName: "AsMemory");
+                SyntaxNode asMemoryInvocationNode = generator.InvocationExpression(
+                    asMemoryExpressionNode,
+                    namedStartNode.WithTriviaFrom(offsetValueNode),
+                    namedLengthNode.WithTriviaFrom(countValueNode));
+
+                // Generate the new buffer argument, ensuring we include the buffer argument name if the user originally indicated one
+                replacedInvocationNode =
+                        (isBufferNamed ? generator.Argument(name: "buffer", RefKind.None, asMemoryInvocationNode) : asMemoryInvocationNode)
```
Evangelink, line 149:
```suggestion
                    (isBufferNamed ? generator.Argument(name: "buffer", RefKind.None, asMemoryInvocationNode) : asMemoryInvocationNode)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
Evangelink, line 129: I have noticed you are using a lot the following pattern: `boolVar ? generator.Argument(name: someName, RefKind.None, node) : node`. You could probably extract it as a function taking the 3 parameters.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb
```diff
@@ -59,5 +59,38 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
 
         End Function
 
+        Protected Overrides Function IsPassingZeroAndBufferLength(bufferValueNode As SyntaxNode, offsetValueNode As SyntaxNode, countValueNode As SyntaxNode) As Boolean
+
+            ' First argument should be an identifier name node
+            Dim firstArgumentIdentifierName As IdentifierNameSyntax = TryCast(bufferValueNode, IdentifierNameSyntax)
```
Evangelink, line 65: [optional] You could get rid of the various `As XXX` as you are doing a direct assignment.

```suggestion
            Dim firstArgumentIdentifierName = TryCast(bufferValueNode, IdentifierNameSyntax)
```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb
```diff
@@ -59,5 +59,38 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
...
+            Dim thirdArgumentMemberAccessExpression As MemberAccessExpressionSyntax = TryCast(countValueNode, MemberAccessExpressionSyntax)
+            If thirdArgumentMemberAccessExpression Is Nothing Then
+                Return False
+            End If
+
+            ' whose identifier is an identifier name node, and its value is the same as the value of first argument, and the member name is `Length`
+            Dim thirdArgumentIdentifierName As IdentifierNameSyntax = TryCast(thirdArgumentMemberAccessExpression.Expression, IdentifierNameSyntax)
+            If thirdArgumentIdentifierName Is Nothing Or
+            thirdArgumentIdentifierName.Identifier.ValueText IsNot firstArgumentIdentifierName.Identifier.ValueText Or
+            thirdArgumentMemberAccessExpression.Name.Identifier.ValueText IsNot "Length" Then
```
Evangelink, line 87:
```suggestion
                thirdArgumentIdentifierName.Identifier.ValueText IsNot firstArgumentIdentifierName.Identifier.ValueText Or
                thirdArgumentMemberAccessExpression.Name.Identifier.ValueText IsNot "Length" Then
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: I'm using the conditional operator _precisely_ because the checks are very simple and don't deserve their own function.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb
carlossanlop: Nice. I'll do that change in this whole file.
This actually sounds like a candidate for a Roslyn analyzer 😄. Visual Studio is currently not giving me that suggestion.
If no one objects, I'll create the issue with this suggestion later today.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb
```diff
@@ -59,5 +59,38 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
...
+        Protected Overrides Function IsPassingZeroAndBufferLength(bufferValueNode As SyntaxNode, offsetValueNode As SyntaxNode, countValueNode As SyntaxNode) As Boolean
+
+            ' First argument should be an identifier name node
+            Dim firstArgumentIdentifierName = TryCast(bufferValueNode, IdentifierNameSyntax)
+            If firstArgumentIdentifierName Is Nothing Then
+                Return False
+            End If
+
+            ' Second argument should be a literal expression node and its value should be zero
+            Dim secondArgumentLiteralExpression = TryCast(offsetValueNode, LiteralExpressionSyntax)
```
mavasani, line 71: Same suggestion here, you can just check for ConstantValue from semantic model.

mavasani (approved):
Overall LGTM. You can improve the heuristic by using SemanticModel.GetConstantValue API.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: So it seems I would change my code to something like this:
```suggestion
                model.GetConstantValue(secondArgumentLiteralExpression) is Optional<object> optionalObject &&
                is Optional<object> optionalValue && optionalValue.HasValue &&
                optionalValue.Value is string value && value.Length == 0 &&
```

But I'm wondering why would that be an improvement? Wouldn't all those additional comparisons and casts have a performance impact?

mavasani: You will not have `optionalValue.Value is string value && value.Length == 0` but instead `optionalValue.Value is int value && value == 0`. The code allows for cases were instead of `0` you used some constant field assigned to `0` and also feels much cleaner then explicitly checking for syntax token `0`. Its fine even to retain the current code if you feel this case is not too important.

carlossanlop: > _optionalValue.Value is string value && value.Length == 0_

Sorry, I pasted the wrong text. I was looking for usage examples of your suggestion and found one with those parameters. I then adapted it to my code, but I didn't realize that I ended up pasting here the code from the other example. I edited my auto-suggestion.

But the general question still applies: the fact that I'm doing more castings and comparisons that directly checking ValueText.

Maybe adding the extra casts is not that impactful, and I think it's important to call the right code. I don't know if comparing strings could cause trouble if the text has an unexpected property, like an unusual encoding.

carlossanlop:
FYI I opened a proposal to suggest simplifying `TryCast` calls in VB: https://github.com/dotnet/roslyn-analyzers/issues/4276

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -51,5 +51,21 @@ protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> impo
...
             return false;
         }
+
+        protected override bool IsPassingZeroAndBufferLength(SemanticModel model, SyntaxNode bufferValueNode, SyntaxNode offsetValueNode, SyntaxNode countValueNode)
+        {
+            return
+                // First argument should be an identifier name node
+                bufferValueNode is IdentifierNameSyntax firstArgumentIdentifierName &&
+                // Second argument should be a literal expression node with a constant value of zero
+                model.GetConstantValue(offsetValueNode) is Optional<object> optionalValue && optionalValue.HasValue && optionalValue.Value is int value && value == 0 &&
```
Youssef1313, line 61: Not tested locally, but I think should work

```suggestion
                model.GetConstantValue(offsetValueNode) is Optional<object> optionalValue && optionalValue.HasValue && optionalValue.Value is 0 &&
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
```diff
@@ -51,5 +51,21 @@ protected override bool IsSystemNamespaceImported(IReadOnlyList<SyntaxNode> impo
...
+                // Third argument should be a member access node...
+                countValueNode is MemberAccessExpressionSyntax thirdArgumentMemberAccessExpression &&
+                thirdArgumentMemberAccessExpression.Expression is IdentifierNameSyntax thirdArgumentIdentifierName &&
+                // whose identifier is that of the first argument...
+                firstArgumentIdentifierName.Identifier.ValueText == thirdArgumentIdentifierName.Identifier.ValueText &&
+                // and the member name is `Length`
+                thirdArgumentMemberAccessExpression.Name.Identifier.ValueText == "Length";
```
Youssef1313, line 68: If WellKnownMemberNames is available to use from here, we can use it:

https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Compilers/Core/Portable/Symbols/WellKnownMemberNames.cs#L337

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb
```diff
@@ -59,5 +59,45 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
...
+            Dim thirdArgumentMemberAccessExpression = TryCast(countValueNode, MemberAccessExpressionSyntax)
+            If thirdArgumentMemberAccessExpression Is Nothing Then
+                Return False
+            End If
+
+            ' whose identifier is an identifier name node, and its value is the same as the value of first argument, and the member name is `Length`
+            Dim thirdArgumentIdentifierName = TryCast(thirdArgumentMemberAccessExpression.Expression, IdentifierNameSyntax)
+            If thirdArgumentIdentifierName IsNot Nothing And
+                thirdArgumentIdentifierName.Identifier.Text = firstArgumentIdentifierName.Identifier.Text And
+                thirdArgumentMemberAccessExpression.Name.Identifier.Text = "Length" Then
```
Youssef1313, line 94: WellKnownMemberNames.LengthPropertyName if it's available to use

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferStreamAsyncMemoryOverloads.Fixer.cs
carlossanlop: Thanks, yes it works. Constant pattern check with `is`: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#constant-pattern

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb
```diff
@@ -59,5 +59,45 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
...
+            Dim firstArgumentIdentifierName = TryCast(bufferValueNode, IdentifierNameSyntax)
+            If firstArgumentIdentifierName Is Nothing Then
+                Return False
+            End If
+
+            ' Second argument should be a literal expression node with a constant value...
+            Dim optionalValue As [Optional](Of Object) = model.GetConstantValue(offsetValueNode)
+
+            ' And must be an integer...
+            If Not optionalValue.HasValue Or TypeOf optionalValue.Value IsNot Integer Then
```
carlossanlop, line 74: @Youssef1313 I'll also test your suggestion here (`optionalValue.Value IsNot 0`), it should probably work on VB too.

stephentoub:
Thanks for improving this.

PR #4285, opened 2020-10-03
Fix bug when exception declared in filter is thrown

### LINKED ISSUE #4280: CA2200: Rethrow to preserve stack details - should not warn when throwing different exception
### Analyzer

**Diagnostic ID**: [CA2200](https://docs.microsoft.com/visualstudio/code-quality/ca2200): `Rethrow to preserve stack details`

### Analyzer source

**Version**: [SDK 5.0.100-preview.8](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug
The caught exception `ex` is only thrown when it is mapped into a different exception by the `MapSendException(...)` method. The analyzer should not warn in this case, further I would assume CA2200 should not warn when the catch block is throwing a different exception (`mappedEx`) than the catched exception (`ex`)

```csharp
    try
    {
          ... 
    }
    catch (Exception ex) when (MapSendException(ex, cancellationToken, out Exception mappedEx))
    {
        throw mappedEx;
    }
```

The exact warning found [here](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L751), related to https://github.com/dotnet/runtime/pull/42957#discussion_r498532030

### Expected behavior
Should not warn in above case
### Actual behavior
Warning

Youssef1313:
Fixes #4280

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
@@ -49,7 +49,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                     if (ancestor.Kind == OperationKind.CatchClause &&
                         ancestor is ICatchClauseOperation catchClause)
                     {
-                        if ((catchClause.ExceptionDeclarationOrExpression is not IVariableDeclaratorOperation variableDeclaratorOperation || variableDeclaratorOperation.Symbol == localReference.Local) &&
+                        if ((catchClause.ExceptionDeclarationOrExpression is not IVariableDeclaratorOperation variableDeclaratorOperation || SymbolEqualityComparer.Default.Equals(variableDeclaratorOperation.Symbol, localReference.Local)) &&
                             catchClause.Locals.Contains(localReference.Local) &&
                             !IsReassignedInCatch(catchClause, localReference))
                         {
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
@@ -49,7 +49,8 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                     if (ancestor.Kind == OperationKind.CatchClause &&
                         ancestor is ICatchClauseOperation catchClause)
                     {
-                        if (catchClause.Locals.Contains(localReference.Local) &&
+                        if ((catchClause.ExceptionDeclarationOrExpression is not IVariableDeclaratorOperation variableDeclaratorOperation || variableDeclaratorOperation.Symbol == localReference.Local) &&
+                            catchClause.Locals.Contains(localReference.Local) &&
                             !IsReassignedInCatch(catchClause, localReference))
                         {
                             context.ReportDiagnostic(throwOperation.CreateDiagnostic(Rule));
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -49,7 +49,8 @@ public override void Initialize(AnalysisContext analysisContext)
                     if (ancestor.Kind == OperationKind.CatchClause &&
                         ancestor is ICatchClauseOperation catchClause)
                     {
-                        if (catchClause.Locals.Contains(localReference.Local) &&
+                        if ((catchClause.ExceptionDeclarationOrExpression is not IVariableDeclaratorOperation variableDeclaratorOperation || variableDeclaratorOperation.Symbol == localReference.Local) &&
```
Evangelink, line 52: You should not be comparing the symbols with `==`.

Evangelink (approved):
Small suggestion for the symbol comparison but otherwise LGTM.

I totally didn't think about this scenario (never used it).

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
Youssef1313, line 52:
```suggestion
                        if ((catchClause.ExceptionDeclarationOrExpression is not IVariableDeclaratorOperation variableDeclaratorOperation || SymbolEqualityComparer.Default.Equals(variableDeclaratorOperation.Symbol, localReference.Local)) &&
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetails.cs
```diff
@@ -49,7 +49,8 @@ public override void Initialize(AnalysisContext analysisContext)
                     if (ancestor.Kind == OperationKind.CatchClause &&
                         ancestor is ICatchClauseOperation catchClause)
                     {
-                        if (catchClause.Locals.Contains(localReference.Local) &&
+                        if ((catchClause.ExceptionDeclarationOrExpression is not IVariableDeclaratorOperation variableDeclaratorOperation || SymbolEqualityComparer.Default.Equals(variableDeclaratorOperation.Symbol, localReference.Local)) &&
```
mavasani, line 52: Why are we skipping the check if this not a variable declaration?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -92,6 +92,30 @@ End Class
...
+using System;
+
+public abstract class C
+{
+    public void M()
+    {
+        try
+        {
+        }
+        catch (Exception ex) when (Map(ex, out Exception mappedEx))
```
mavasani, line 108: Can you add a similar test where we have ‘catch(Exception) when M(out Exception ex)’ followed by ‘throw ex’? That should not fire.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/RethrowToPreserveStackDetailsTests.cs
```diff
@@ -92,6 +92,30 @@ End Class
...
+using System;
+
+public abstract class C
+{
+    public void M()
+    {
+        try
+        {
+        }
+        catch (Exception) when (Map(out Exception ex))
```
mavasani, line 108: @Youssef1313 please add both tests - one in the repro and one that you added here based on PR feedback. Sorry if my comment was unclear.

Youssef1313: @mavasani My bad. I thought I copied the existing test and modified the copy.

mavasani (approved):

PR #4304, opened 2020-10-09
(XUnit) SkippableFact and FXCop code analysis rule CA1822: Mark members as static

Atrejoe:
### Analyzer

**Diagnostic ID**: [CA1822](https://docs.microsoft.com/visualstudio/code-quality/CA1822)

### Describe the improvement

As (X)unit tests should be non-static, allowing test setup and teardown in class constructors and destructors, I think this check should not apply.
FxCop already recognizes XUnit tests by looking at the `FactAttribute`.

I guess that that that can be assigned to FactAttribute instead of strictly being equals to FactAttribute, it would be fine, as SkippableFactAttribute (from [AArnott/Xunit.SkippableFact ](https://github.com/AArnott/Xunit.SkippableFact) / [NuGet Gallery](https://www.nuget.org/packages/Xunit.SkippableFact)) inherits from FactAttribute.

I think this already attempted at https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/Extensions/INamedTypeSymbolExtensions.cs#L266, but maybe I don't understand completely.

See also: https://github.com/AArnott/Xunit.SkippableFact/issues/23

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -298,6 +298,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
 
             // XUnit attributes
             Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitFactAttribute));
+            Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitSkippableFactAttribute));
 
             // NUnit Attributes
             Add(wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.NUnitFrameworkSetUpAttribute));
--- a/src/Utilities/Compiler/WellKnownTypeNames.cs
+++ b/src/Utilities/Compiler/WellKnownTypeNames.cs
@@ -267,6 +267,7 @@ namespace Analyzer.Utilities
         public const string MicrosoftVisualStudioTestToolsUnitTestingStringAssert = "Microsoft.VisualStudio.TestTools.UnitTesting.StringAssert";
         public const string XunitAssert = "Xunit.Assert";
         public const string XunitFactAttribute = "Xunit.FactAttribute";
+        public const string XunitSkippableFactAttribute = "Xunit.SkippableFactAttribute";
         public const string XunitTheoryAttribute = "Xunit.TheoryAttribute";
         public const string NUnitFrameworkAssert = "NUnit.Framework.Assert";
         public const string NUnitFrameworkOneTimeSetUpAttribute = "NUnit.Framework.OneTimeSetUpAttribute";


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=4304) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=4304) before we can accept your contribution.<br/><br/>:x: Atrejoe [sign now](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=4304)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/roslyn-analyzers?pullRequest=4304) it.</sub>

Youssef1313:
I see a check for inheritance in:

https://github.com/dotnet/roslyn-analyzers/blob/32e083168cc884d3453abeb3d5d4d457c9d5e496/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs#L199

Not sure why it's not working properly, will need to debug. But I don't feel the analyzer should keep adding any inherited attributes.

Evangelink:
Hi @Atrejoe,

Looking at the code, I can confirm that there is no need for any modification at all to have the correct behavior. Besides, you are not targeting the correct branch. I will file a separate PR to add a unit test showing that it's working properly.

Evangelink (changes requested):
See #4307

Atrejoe:
Hi @Evangelink,

Thanks for creating the test. This looks like it demonstrates correct behavior.
But then again, this would also tell me that the validation error I received could not have been there in the first place, right?

Maybe having a look at the attribute in the library referred to may spark an Idea:
https://github.com/AArnott/Xunit.SkippableFact/blob/master/src/Xunit.SkippableFact/SkippableFactAttribute.cs

But if you are saying, "I've got this, this is not a problem (anymore)", that's fine too.

I targeted this branch as I was trying to supply a fix for .Net Core 3.1. The main branch appeared to be for 5.0. But again, I may not understand correctly.

sharwell (changes requested):
This should already be handled since `SkippableFactAttribute` extends `FactAttribute`. @Atrejoe can you provide more specific details regarding the situation where you are seeing a diagnostic, since it should not be showing up?

Atrejoe:
@sharwell and @Evangelink it appears that on a new Xunit test application, confirms that there is not problem.

This may have been because I was migrating a legacy library using legacy static code analysis, besides roslyn-analyzers in `Microsoft.CodeAnalysis.FxCopAnalyzers` 

Demonstration:
```csharp
namespace XUnitTestProjectDemo
{
    /// <summary>
    /// This is a sample unittest
    /// </summary>
    public class UnitTestSample
    {
        /// <summary>
        /// Demonstrates regular XUnit tests are fine
        /// </summary>
        [Fact]
        public void Test()
        {

        }

        /// <summary>
        /// Demonstrates Skippable tests (derived from FactAttribute) are fine too
        /// </summary>
        [SkippableFact]
        public void TestSkippable()
        {

        }
    }

    /// <summary>
    /// 
    /// </summary>
    public class Demo
    {
        /// <summary>
        /// Demonstrates CS1822 is checked.
        /// </summary>
        public void Demonstrate() { 
        }
    }
}
```

image

Atrejoe:
No, it turns out I was using an old version of FxCopAnalyzers (2.6.3 to be exact) :man_facepalming:

After upgrading this message disappeared.
My apologies for wasting your time!

sharwell:
@Atrejoe No problem, thanks for the update!  👍

Evangelink:
@Atrejoe Thanks for the feedback!

> But then again, this would also tell me that the validation error I received could not have been there in the first place, right?

> Maybe having a look at the attribute in the library referred to may spark an Idea:
https://github.com/AArnott/Xunit.SkippableFact/blob/master/src/Xunit.SkippableFact/SkippableFactAttribute.cs

> But if you are saying, "I've got this, this is not a problem (anymore)", that's fine too.

Totally sorry about that, I shall have given more explanation. I did meant that the issue was fixed at some point in the codebase and so using latest version no longer reproduce the issue.

PR #4312, opened 2020-10-12
Remove more false positives from CA3075

### LINKED ISSUE #3937: CA3075 warns that XmlReader.Create(string) is unsafe even though it delegates to a call that CA3075 considers safe
[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers) 3.3.0-beta2.final, [CA3075](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca3075)

These calls both delegate to `XmlReader.Create("x", settings: null, inputContext: null)` and are exactly equivalent:

```cs
// ⚠ CA3075 An insecure overload of XmlReader.Create which does not accept an XmlReaderSettings argument.
//                 ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓
using var reader = XmlReader.Create("x");
```

```cs
// No warning
using var reader = XmlReader.Create("x", settings: null);
```

dotpaul:
Fixes #3937

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetFramework.Analyzers/DoNotUseInsecureDtdProcessingXmlReaderCreateWrongOverloadTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetFramework.Analyzers/DoNotUseInsecureDtdProcessingXmlReaderCreateWrongOverloadTests.cs
@@ -21,7 +21,7 @@ namespace Microsoft.NetFramework.Analyzers.UnitTests
         ////    => VerifyVB.Diagnostic(DoNotUseInsecureDtdProcessingAnalyzer.RuleXmlReaderCreateWrongOverload).WithLocation(line, column);
 
         [Fact]
-        public async Task UseXmlReaderCreateWrongOverloadShouldGenerateDiagnostic()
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetFramework.Analyzers/DoNotUseInsecureDtdProcessingXmlReaderCreateWrongOverloadTests.cs
```diff
@@ -166,13 +158,12 @@ Class TestClass
...
     End Sub
-End Class",
-                GetCA3075XmlReaderCreateWrongOverloadBasicResultAt(8, 26)
+End Class"
             );
         }
 
         [Fact]
-        public async Task UseXmlReaderCreateInsecureOverloadInCatchShouldGenerateDiagnostic()
+        public async Task UseXmlReaderCreateInsecureOverloadInCatchShouldGenerateNotDiagnostic()
```
mavasani, line 166:
```suggestion
        public async Task UseXmlReaderCreateInsecureOverloadInCatchShouldNotGenerateDiagnostic()
```

mavasani (approved):

PR #4320, opened 2020-10-13
Fix localization bug

Youssef1313:
1. The manual change made was using `LocalizableResourceString` instead of `string`.
2. After this change, violations of RS1031 and RS1032 appeared. This was fixed using the codefix associated with it.
3. Then, I ran `msbuild RoslynAnalyzers.sln /t:pack` to update the XLF and generated files to match the changes in step 2.


Evangelink (approved):
LGTM

Youssef1313:
@Evangelink @mavasani Is it intended that RS1031 and RS1032 doesn't report violations when `string` is used?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_ReleaseTracking.cs
```diff
@@ -114,8 +114,8 @@ public sealed partial class DiagnosticDescriptorCreationAnalyzer
 
         internal static readonly DiagnosticDescriptor InvalidHeaderInAnalyzerReleasesFileRule = new DiagnosticDescriptor(
             id: DiagnosticIds.InvalidEntryInAnalyzerReleasesFileRuleId,
-            title: CodeAnalysisDiagnosticsResources.InvalidEntryInAnalyzerReleasesFileRuleTitle,
-            messageFormat: CodeAnalysisDiagnosticsResources.InvalidHeaderInAnalyzerReleasesFileRuleMessage,
+            title: new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.InvalidEntryInAnalyzerReleasesFileRuleTitle), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources)),
+            messageFormat: new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.InvalidHeaderInAnalyzerReleasesFileRuleMessage), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources)),
```
Youssef1313, line 118: Is this a problem from the original code?
I can't see a reason why the title is `InvalidEntryInAnalyzerReleasesFileRuleTitle` and the message is `InvalidHeaderInAnalyzerReleasesFileRuleMessage`. But that's out-of-scope for this PR.

Evangelink:
Not expected, we do try to check various cases through this method: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs#L665-L702

I would need to debug to see why it does not work.

PR #4322, opened 2020-10-13
CA2243: Update the rule to exclude ObsoleteAttribute.UrlFormat

### LINKED ISSUE #3635: CA2243 false positive when using ObsoleteAttribute.UrlFormat
#### Analyzer package and version

(Latest checked in to dotnet/runtime.)

#### Diagnostic ID

CA2243: Attribute string literals should parse correctly
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2243

#### Problem description

In .NET 5.0, we're extending the type `ObsoleteAttribute` by adding some new properties. See https://github.com/dotnet/runtime/issues/33089 for more info.

```cs
namespace System
{
    public partial class ObsoleteAttribute : Attribute
    {
        // New:
        public string? DiagnosticId { get; set; }
        public string? UrlFormat { get; set; }
    }
}
```

The _UrlFormat_ property is typed as _string_, and even though it has "Url" in the property name it's not intended to represent a well-formed URL. Rather, it's intended to represent a format string such as `"https://example.com/{0}/"`, where the compiler will essentially perform a `string.Format(UrlFormat, DiagnosticId)` when generating the friendly URL to display to the developer.

However, CA2243 sees the literal `'{'` and `'}'` characters in the string and deduces that the URL itself is not well-formed. This leads to false positive CA2243 hits for whoever ends up using the newly introduced extensibility features.

My recommendation is that CA2243 special-case this property and ignore it.

/cc @terrajobst

Evangelink:
Fix #3635

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AttributeStringLiteralsShouldParseCorrectly.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AttributeStringLiteralsShouldParseCorrectly.cs
@@ -243,16 +243,16 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
     internal class ValueValidator
     {
-        private readonly string[]? _ignoredNames;
+        private readonly string[] _ignoredNames;
 
         public ImmutableArray<string> AcceptedTokens { get; }
         public string TypeName { get; }
         public Func<string, bool> IsValidValue { get; }
 
         public bool IsIgnoredName(string name)
-            => _ignoredNames != null && _ignoredNames.Contains(name, StringComparer.OrdinalIgnoreCase);
+            => _ignoredNames.Contains(name, StringComparer.OrdinalIgnoreCase);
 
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AttributeStringLiteralsShouldParseCorrectly.cs
```diff
@@ -243,20 +243,18 @@ private static void Analyze(Action<Diagnostic> reportDiagnostic, AttributeData a
 
     internal class ValueValidator
     {
-        private readonly string? _ignoredName;
+        private readonly string[]? _ignoredNames;
```
Youssef1313, line 246: The compiler doesn't pass a null array for params.

You can just use `string[]`.

mavasani: +1 You can also get rid of null check in `IsIgnoredName`.

mavasani (approved):
LGTM. Should be good to merge once you address @Youssef1313's suggestion.

PR #4325, opened 2020-10-13
Initial implementation of add missing interpolation token analyzer

### LINKED ISSUE #2767: New rule: Interpolated strings that are missing the $ special character
It'd be great if the correctness analyzers included a rule that analyzed strings to detect cases where an interpolated string was intended but the `$` character was erroneously left off.  You can see some examples of this in corefx here:
https://github.com/dotnet/corefx/pull/40322

There are probably a wide variety of heuristics that could be used, but I'm imagining something like looking at strings in a context where interpolation could be used, and if it contains matching and non-nested `{`,`}` pairs, then if it could be compiles successfully with `$` (or maybe if everything inside the `{`,`}` is a valid expression and can be bound), the rule triggers.

Youssef1313:
~~The analyzer is currently not very intelligent.~~
~~It can't detect cases like: `"{AMethodCall()}"` or `"{object.Property}"`.~~
~~Not sure what's the correct API to use that can validate cases like this?~~

(The above is outdated now)

Fixes #2767

TODOs:

- [ ] Tests suggested by @Evangelink 
- [ ] Review potential false positives stated by @Evangelink in https://github.com/dotnet/roslyn-analyzers/pull/4325#discussion_r678214029.


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,107 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AddMissingInterpolationTokenAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2017";
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AddMissingInterpolationTokenTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AddMissingInterpolationTokenMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+            DiagnosticCategory.Reliability,
```
Youssef1313, line 22: Not sure if this is the correct category.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1454,4 +1454,10 @@
   <data name="MarkAttributesWithAttributeUsageCodeFix" xml:space="preserve">
     <value>Apply 'AttributeUsageAttribute'</value>
   </data>
-</root>
+  <data name="AddMissingInterpolationTokenMessage" xml:space="preserve">
+    <value>The string may have a missing interpolation token</value>
```
Evangelink, line 1458: What about something like `This string literal uses composite formatting but is not an interpolated string. Did you miss the '$' prefix?`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1454,4 +1454,10 @@
...
+  </data>
+  <data name="AddMissingInterpolationTokenTitle" xml:space="preserve">
+    <value>The string may have a missing interpolation token</value>
```
Evangelink, line 1461: What about `Possible missing interpolation token`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
Evangelink: I don't know, I would have put it under `Usage` but I have opened a ticket for `CA2009` which is under `Reliability` at the moment.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,107 @@
...
+
+                var parts = GetInterpolationParts(stringText);
+                if (parts.All(s => IsValidInterpolationPart(s, literalOperation.SemanticModel, literalOperation.Syntax.SpanStart)))
+                {
+                    context.ReportDiagnostic(literalOperation.CreateDiagnostic(Rule));
+                }
+            }, OperationKind.Literal);
+        }
+
+        private static bool IsValidInterpolationPart(string s, SemanticModel semanticModel, int position)
```
Evangelink, line 53: Looking at https://docs.microsoft.com/dotnet/csharp/language-reference/tokens/interpolated#structure-of-an-interpolated-string I think that your validation algorithm is wrong.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                // Numerical literals are valid interpolation from language perspective.
+                // But we don't want the analyzer to flag for them. So,
+                // they're invalid from analyzer perspective.
+                return false;
+            }
+            return !semanticModel.LookupSymbols(position, name: s).IsDefaultOrEmpty;
+        }
+
+        private static IEnumerable<string> GetInterpolationParts(string s)
```
Evangelink, line 65: I think you want to reuse the parsing regex from the [CA2243 PR](https://github.com/dotnet/roslyn-analyzers/pull/4226/files#diff-822f292292d15a89b293bf2c681da69718067c9bfb2c12968980921dee24b1f9R166)

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.CodeAnalysis.Testing;
+using Test.Utilities;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.QualityGuidelines.AddMissingInterpolationTokenAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.CodeQuality.Analyzers.UnitTests.QualityGuidelines
+{
+    public class AddMissingInterpolationTokenTests
```
Evangelink, line 13: Please add the following tests:

- string literal assigned to a variable.
- string that looks like an interpolated string inside a `string.Format` call. Do we want to report? More generally if the string is passed to a `format` parameter do we still want to report?
- cases where you use the `alignment` and `formatString` parts of the tokens (see [doc](https://docs.microsoft.com/dotnet/csharp/language-reference/tokens/interpolated#structure-of-an-interpolated-string))

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA2017 | Reliability | Info | AddMissingInterpolationTokenAnalyzer, [Documentation](https://docs.microsoft.com/visualstudio/code-quality/ca2017)
```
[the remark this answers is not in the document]
Youssef1313: @Evangelink The fix will reflect only after publishing the next version and updating the package version used in the repository.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
Youssef1313: @Evangelink Will the regex you used detect nesting?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Youssef1313: > string that looks like an interpolated string inside a string.Format call. Do we want to report? More generally if the string is passed to a format parameter do we still want to report?

Did you mean something like `string.Format("{0} - {1}", foo, bar)`? If so, that's not something the analyzer should report.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
Evangelink: Not sure honestly, it will need some testing. Tagging @jmarolf in case he has a better idea for this handling.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Evangelink: More like do we want to report on `string.Format("{x}")` and also `string.Format("{0}")`. I would tend to say no but in this case the code needs to be tweaked to avoid reporting.

Youssef1313: Currently, the code reports for `string.Format("{x}")` **if** there is a variable called `x` in scope.
But it doesn't report for `string.Format("{0}")`

mavasani:

mavasani (changes requested):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
Youssef1313: @Evangelink The implementation currently handles the very simple cases only. I think I'd need some better way to get this information using some compiler APIs. Do you have ideas?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Youssef1313: @Evangelink This will trigger RS1032 because it's multi-sentence and doesn't end with a period. I believe this is a false-positive. Should we update RS1032 to handle both question marks and periods exactly the same? or just suppress the diagnostic?

mavasani:
Sorry @Youssef1313 for the delay in review. Can you please retarget the PR to `release\6.0.1xx` branch?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1,5 +1,10 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
 
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA2250 | Usage | Info | AddMissingInterpolationTokenAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2250)
```
mavasani, line 6: ID has already been used up now.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,107 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AddMissingInterpolationTokenAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2250";
```
Youssef1313, line 17:
```suggestion
        internal const string RuleId = "CA2251";
```

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif
```diff
@@ -3450,6 +3450,26 @@
             ]
           }
         },
+        "CA2251": {
+          "id": "CA2251",
+          "shortDescription": "Possible missing interpolation token",
+          "fullDescription": "This string literal uses composite formatting but is not an interpolated string. Did you miss the '$' prefix?",
+          "defaultLevel": "note",
+          "helpUri": "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2250",
```
Youssef1313, line 3458:
```suggestion
          "helpUri": "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2251",
```

src/NetAnalyzers/RulesMissingDocumentation.md
```diff
@@ -11,3 +11,4 @@ CA1842 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-r
 CA1843 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1843> | Do not use 'WaitAll' with a single task |
 CA1844 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1844> | Provide memory-based overrides of async methods when subclassing 'Stream' |
 CA1845 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1845> | Use span-based 'string.Concat' |
+CA2250 | https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2250 | Possible missing interpolation token |
```
Youssef1313, line 14:
```suggestion
CA2251 | https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2251 | Possible missing interpolation token |
```

src/NetAnalyzers/RulesMissingDocumentation.md
```diff
@@ -11,3 +11,4 @@ CA1842 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-r
 CA1843 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1843> | Do not use 'WaitAll' with a single task |
 CA1844 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1844> | Provide memory-based overrides of async methods when subclassing 'Stream' |
 CA1845 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1845> | Use span-based 'string.Concat' |
+CA2251 | https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2251 | Possible missing interpolation token |
```
Youssef1313, line 14:
```suggestion
CA2251 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2251> | Possible missing interpolation token |
```

Youssef1313:
> Sorry @Youssef1313 for the delay in review. Can you please retarget the PR to `release\6.0.1xx` branch?

No problem! I rebased onto `release/6.0.1xx`.

Youssef1313 (commented):
I rewrote the analyzer in a much better way. This is ready for another review modulo VB implementation.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,53 @@
...
+            var annotation = new SyntaxAnnotation();
+            if (SyntaxFactory.ParseExpression("$" + operation.Syntax.ToString(), options: operation.Syntax.SyntaxTree.Options).WithAdditionalAnnotations(annotation)
+                is not InterpolatedStringExpressionSyntax dummyNode)
+            {
+                return false;
+            }
+
+            var root = operation.Syntax.SyntaxTree.GetRoot();
+            root = root.ReplaceNode(operation.Syntax, dummyNode);
+            dummyNode = (InterpolatedStringExpressionSyntax)root.GetAnnotatedNodes(annotation).Single();
```
Evangelink, line 27: I am not sure if that's a good solution to throw in case of multiple matches.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (!operation.SemanticModel.TryGetSpeculativeSemanticModel(operation.Syntax.SpanStart, dummyNode.FirstAncestorOrSelf<StatementSyntax>(), out var model))
+            {
+                return false;
+            }
+
+            var interpolations = dummyNode.Contents.OfType<InterpolationSyntax>();
+            bool hasNonConstantInterpolation = false;
+            foreach (var interpolation in interpolations)
+            {
+                if (interpolation.Expression is LiteralExpressionSyntax)
```
Evangelink, line 37: I don't remember what @sharwell said but I recall some discussion about `Kind`/`IsKind`.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    continue;
+                }
+
+                hasNonConstantInterpolation = true;
+                var info = model.GetSymbolInfo(interpolation.Expression);
```
Evangelink, line 43: [cosmetic] I would get rid of the `info` var and call `.Symbol` directly on `GetSymbolInfo(..)` in the `if`. Or at least transform `info` into `symbol` and get the symbol.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
+{
+    public abstract class AbstractAddMissingInterpolationTokenAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2253";
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AddMissingInterpolationTokenTitle), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+#pragma warning disable RS1032 // Define diagnostic message correctly - the analyzer wants a period after the existing question mark.
```
Evangelink, line 18: That's funny, I recall having handle various punctuation in the initial implementation. I would recommend to create a ticket to fix the implementation and link it in the code, WDYT?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,51 @@
...
+            new LocalizableResourceString(nameof(MicrosoftCodeQualityAnalyzersResources.AddMissingInterpolationTokenMessage), MicrosoftCodeQualityAnalyzersResources.ResourceManager, typeof(MicrosoftCodeQualityAnalyzersResources)),
+#pragma warning restore RS1032 // Define diagnostic message correctly
+            DiagnosticCategory.Usage,
+            RuleLevel.IdeSuggestion,
+            description: null,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
```
Evangelink, line 27: Could be `sealed` as we don't want children to change this.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        private protected abstract bool ShouldReport(ILiteralOperation operation);
+
+        public override void Initialize(AnalysisContext context)
```
Evangelink, line 31: Could be `sealed` as we don't want children to change this.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterOperationAction(context =>
+            {
+                var literalOperation = (ILiteralOperation)context.Operation;
+                if (!literalOperation.ConstantValue.HasValue ||
```
Evangelink, line 38: [cosmetic] I would personally either:
- revert these conditions and merge them with the next `if` (`ShouldReport`)
- revert the `ShouldReport` and merge it with this `if`.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpAddMissingInterpolationToken : AbstractAddMissingInterpolationTokenAnalyzer
+    {
+        private protected override bool ShouldReport(ILiteralOperation operation)
+        {
+            var annotation = new SyntaxAnnotation();
+            if (SyntaxFactory.ParseExpression("$" + operation.Syntax.ToString(), options: operation.Syntax.SyntaxTree.Options).WithAdditionalAnnotations(annotation)
```
Evangelink, line 19: @Youssef1313 You said

> It can't detect cases like: "{AMethodCall()}" or "{object.Property}".

Is it because of the implementation of `SyntaxFactory.ParseExpression`? Has this changed between this initial PR and now?

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
Youssef1313: @Evangelink `.IsKind` is faster than type checking and faster than `Kind`. However:

- This syntax node isn't associated with one single kind. It could be string literal, numeric literal, true/false. So I didn't bother.
- The difference in most cases is unlikely to be noticeable.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
Youssef1313: > Is it because of the implementation of SyntaxFactory.ParseExpression? Has this changed between this initial PR and now?


Indeed. The initial implementation is what was buggy and unreliable. This is outdated now.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
Youssef1313: → #5310

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
Youssef1313: @Evangelink This should never happen. We annotated one node only. So `GetAnnotatedNodes` shouldn't return multiple nodes.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,52 @@
...
+                }
+
+                hasNonConstantInterpolation = true;
+                if (model.GetSymbolInfo(interpolation.Expression).Symbol is null)
+                {
+                    return false;
+                }
+            }
+
+            return hasNonConstantInterpolation;
```
Evangelink, line 49: I missed that in the previous review but I don't see the point of this variable, it's always true at this point. I would simply return true here and get rid of the varaible.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.Fixer.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Composition;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeQuality.Analyzers.QualityGuidelines;
+
+namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public sealed class CSharpAddMissingInterpolationTokeFixer : AbstractAddMissingInterpolationTokenFixer
```
Evangelink, line 12: Typo: missing `n` in `Token`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.Fixer.cs
```diff
@@ -0,0 +1,46 @@
...
+            root = root.ReplaceNode(node, newNode);
+            return Task.FromResult(document.WithSyntaxRoot(root));
+        }
+
+        private protected abstract SyntaxNode GetReplacement(SyntaxNode node);
+
+        private class MyCodeAction : DocumentChangeAction
+        {
+            public MyCodeAction(Func<CancellationToken, Task<Document>> createChangedDocument)
+                : base("", createChangedDocument, "")
```
Evangelink, line 40: Code fix is missing a title and the equivalency key.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.Fixer.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+            }
+        }
+
```
Evangelink, line 44:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            var fixedCode = @"
+using System;
+
+class Program
+{
+    public static void Main()
+    {
+        int x = 5;
+        Console.WriteLine($""{x}, {0}"");
```
Evangelink, line 98: Not part of this PR/analyzer but I am wondering if we should have some analyzer to simplify some interpolated parts (like this `{0}`).

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            var code = @"
+using System;
+
+class Program
+{
+    public static void Main()
+    {
+        int x = 5;
+        Console.WriteLine([|""{x}, {0}""|]);
```
Evangelink, line 86: Could we add a test for a case like: `Console.WriteLine(""{x}, {0}"", 1, 2)`? I would expect not to have the diagnostic but I could understand raising anyways because we are in a mixing scenario.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            var code = @"
+using System;
+
+class Program
+{
+    public static void Main()
+    {
+        int x = 5;
+        Console.WriteLine([|""{M(x)}""|]);
```
Evangelink, line 115: Could we add a test with a property and/or method chaining call? Something like `{x.Prop.Other.Get()}`.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Evangelink: > * cases where you use the `alignment` and `formatString` parts of the tokens (see [doc](https://docs.microsoft.com/dotnet/csharp/language-reference/tokens/interpolated#structure-of-an-interpolated-string))

Could we add tests for these?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines.CSharpAddMissingInterpolationToken,
+    Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines.CSharpAddMissingInterpolationTokeFixer>;
+
+namespace Microsoft.CodeQuality.Analyzers.UnitTests.QualityGuidelines
+{
+    public class AddMissingInterpolationTokenTests
```
Evangelink, line 11: Could we add tests for special cases (see [doc](https://docs.microsoft.com/dotnet/csharp/language-reference/tokens/interpolated#special-characters)):
- escaped curly brace
- ternary operator
- interpolated verbatim string

Evangelink (commented):
One more round of review. Thanks and sorry @Youssef1313 :)

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
Youssef1313: @Evangelink It should be false in cases like `"Hello world"`, or `"Hello, {0}"`. (ie, no interpolations or all interpolation are literals)

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Youssef1313: Indeed the current implementation will raise a diagnostic. The suggestion by @sharwell at https://github.com/dotnet/roslyn-analyzers/issues/2767#issuecomment-656355022 was saying:

> My initial suggestion is exclude cases where **all** of the interpolation expressions are non-negative integer literals.

I'd say we

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.cs
Evangelink: Ahh sorry! With the comments in between I thought that the closing curly brace was for the if not the loop... Feel free to resolve.

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpAddMissingInterpolationToken.Fixer.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeQuality.Analyzers.QualityGuidelines;
+
+namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public sealed class CSharpAddMissingInterpolationTokenFixer : AbstractAddMissingInterpolationTokenFixer
+    {
+        private protected override SyntaxNode GetReplacement(SyntaxNode node)
+            => SyntaxFactory.ParseExpression("$" + node.ToString());
```
Youssef1313, line 15: I've always wished if the `additionalProperties` dictionary on diagnostics was generic. I'd have passed the syntax node as already calculated in the analyzer.

alenros:
I'm so excited for this, I'm either forgetting to add the $ or deciding to make the string interpolation in the middle of writing the string.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Youssef1313: @Evangelink Do you mean an analyzer for replacing `{0}` in interpolated string with just `0`? For this specific case, the behavior can differ per the current culture I believe.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Youssef1313: @Evangelink Do we still need the suggested tests given that I'm no longer manually parsing the string, but falling back to speculative semantic model?

Youssef1313:
@mavasani Can you take a look please? Thanks!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,77 @@
...
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            CreateLocalizableResourceString(nameof(AddMissingInterpolationTokenTitle)),
+            CreateLocalizableResourceString(nameof(AddMissingInterpolationTokenMessage)),
+            DiagnosticCategory.Usage,
+            RuleLevel.IdeSuggestion,
+            description: null,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
```
Evangelink, line 27:
```suggestion
        public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Determines whether the given interpolated string node has the following properties:
+        /// <list type="number">
+        /// <item><description>All parts are bindable.</description></item>
+        /// <item><description>There is at least one non-literal expression.</description></item>
+        /// </list>
+        /// </summary>
+        /// <param name="node"></param>
+        /// <param name="model"></param>
+        /// <returns></returns>
```
Evangelink, line 70:
```suggestion
        /// <summary>
        /// Determines whether the given interpolated string node has the following properties:
        /// <list type="number">
        /// <item><description>All parts are bindable.</description></item>
        /// <item><description>There is at least one non-literal expression.</description></item>
        /// </list>
        /// </summary>
```

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Evangelink: I would still add them to be sure some know edge cases are covered but I understand if you don't. Maybe that's worth to add at least the test for ternary operator as it forces you to include parenthesis.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationTokenTests.cs
Evangelink: Yeah I meant that and you are probably right about the culture :)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
```diff
@@ -0,0 +1,77 @@
...
+    using static MicrosoftCodeQualityAnalyzersResources;
+    public abstract class AbstractAddMissingInterpolationTokenAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2259";
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            CreateLocalizableResourceString(nameof(AddMissingInterpolationTokenTitle)),
+            CreateLocalizableResourceString(nameof(AddMissingInterpolationTokenMessage)),
+            DiagnosticCategory.Usage,
+            RuleLevel.IdeSuggestion,
```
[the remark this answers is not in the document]
Youssef1313: @Evangelink I'm not sure. @stephentoub What do you think? (btw, I'm not sure the rule requires review in dotnet/runtime first)

Youssef1313:
@Evangelink I'll add the tests once we're sure the rule is approved by the dotnet/runtime team.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AddMissingInterpolationToken.cs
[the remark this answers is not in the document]
stephentoub: > Is it the right level?

Yes, I think suggestion makes sense as the default level, assuming we're confident this will have a very low false positive rate.  We don't want to add lots of bugs if someone does a fix-all.

> I'm not sure the rule requires review in dotnet/runtime first

It doesn't.  This is about C# syntax / misuse rather than about APIs.

stephentoub:
@mavasani, @Youssef1313, are you still working on this?

Youssef1313:
> @mavasani, @Youssef1313, are you still working on this?

it needs a review and adding few tests IIRC.

Issue #3608, opened 2020-05-07
New Analyzer: Warn when calling Enumerable.Cast<T>/OfType<T> with incompatible types


fowl2:
The following will always throw an `InvalidCastException` at runtime:

`new int[] { 1 }.Cast<string>().ToList();`
`new float[] { 1f }.Cast<double>().ToList();`
`new int[] { 1 }.Cast<long>().ToList();`

Perhaps worse, this will return an enumerable that fully enumerates the source and then returns *no items silently*:

`enumerableOfApple.OfType<Orange>()`

Both of these situations are likely to be a mistake.

I'd never played with Roslyn before but I thought it'd be fun so I [hacked together something](https://github.com/fowl2/roslyn-analyzers/blob/EnumerableCasts/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs). The rules seem quite complex so I have no confidence I got it right - or even if it's possible to spot 100% of cases - but it seems to be able to pick up the most obvious problems and I think that's still valuable. 

Also, naming things and writing descriptive messages is really hard 😅.

An interesting thing I learnt was that enums can be cast to/from their underlying type and to other enums with the same underlying type this way, but numeric types can't be. I guess the numeric type conversions must be a language thing and so the compiled `Cast<T>()` can't know them.

Anyway I hope this is useful, even just for the idea.

Thanks

PS. It's amazing how good the getting started docs are here! (The comments in the code are a little sparse though ;))

Evangelink:
@fowl2 Do you want to give it a go and we will give some feedback or do you want me to take care of this one?

fowl2:
I'll give it a go! Created draft PR #4328 to gather feedback.

buyaa-n:
The API Proposal https://github.com/dotnet/runtime/issues/33770 ready for review

carlossanlop:
The API proposal has been approved, @fowl2

PR #4328, opened 2020-10-14
New Analyzer: Warn when calling Enumerable.Cast<T>/OfType<T> with incompatible types 

fowl2:
Warn if use of `Enumerable.Cast<T>` would always fail at runtime or `OfType<T>()` would provably result in an empty sequence, because we can see that the input type in the sequence will never be the specified type.

RuleId: CA2019
Category: Reliability

related: https://github.com/dotnet/roslyn-analyzers/issues/3608, https://github.com/dotnet/runtime/issues/33770

- [x] Tests pass
- [x] Targeted to ~release/7.0.1xx~ main
- [x] Run against 
   - [x] dotnet/roslyn-analyzers 
   - [x] dotnet/runtime
   - [ ] dotnet/roslyn 
- [ ] Docs

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
@@ -1008,6 +1008,12 @@
   <data name="DoNotAddArchiveItemPathToTheTargetFileSystemPathMessage" xml:space="preserve">
     <value>When creating path for '{0} in method {1}' from relative archive item path to extract file and the source is an untrusted zip archive, make sure to sanitize relative archive item path '{2} in method {3}'</value>
   </data>
+  <data name="DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesTitle" xml:space="preserve">
+    <value>Do not call Enumerable.Cast&lt;T&gt; or Enumerable.OfType&lt;T&gt; with incompatible types</value>
+  </data>
+  <data name="DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesMessageOfType" xml:space="preserve">
+    <value>This call will always result in an empty sequence because type '{0}' is incompatible with type '{1}'</value>
+  </data>
   <data name="DoNotCreateTasksWithoutPassingATaskSchedulerTitle" xml:space="preserve">
     <value>Do not create tasks without passing a TaskScheduler</value>
   </data>
@@ -1721,6 +1727,15 @@
   </data>
   <data name="ModuleInitializerAttributeShouldNotBeUsedInLibrariesTitle" xml:space="preserve">
     <value>The 'ModuleInitializer' attribute should not be used in libraries</value>
+  </data>
+  <data name="DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesDescription" xml:space="preserve">
+    <value>Enumerable.Cast&lt;T&gt; and Enumerable.OfType&lt;T&gt; require compatible types to function expectedly.  
+The generic cast (IL 'unbox.any') used by the sequenced returned by Enumerable.Cast&lt;T&gt; will throw InvalidCastException at runtime on elements of the types specified.  
+The generic type check (C# 'is' operator/IL 'isinst') used by Enumerable.OfType&lt;T&gt; will never succeed with elements of types specified, resulting in an empty sequence.  
+Widening and user defined conversions are not supported with generic types.</value>
+  </data>
+  <data name="DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesMessageCast" xml:space="preserve">
+    <value>Type '{0}' is incompatible with type '{1}' and cast attempts will throw InvalidCastException at runtime</value>
   </data>
   <data name="UseAsyncMethodInAsyncContextMessage" xml:space="preserve">
     <value>'{0}' synchronously blocks. Await '{1}' instead.</value>
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
@@ -1730,7 +1730,7 @@
   </data>
   <data name="DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesDescription" xml:space="preserve">
     <value>Enumerable.Cast&lt;T&gt; and Enumerable.OfType&lt;T&gt; require compatible types to function expectedly.  
-The generic cast (IL 'unbox.any') used by the sequenced returned by Enumerable.Cast&lt;T&gt; will throw InvalidCastException at runtime on elements of the types specified.  
+The generic cast (IL 'unbox.any') used by the sequence returned by Enumerable.Cast&lt;T&gt; will throw InvalidCastException at runtime on elements of the types specified.  
 The generic type check (C# 'is' operator/IL 'isinst') used by Enumerable.OfType&lt;T&gt; will never succeed with elements of types specified, resulting in an empty sequence.  
 Widening and user defined conversions are not supported with generic types.</value>
   </data>
--- a/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
+++ b/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
@@ -1860,7 +1860,7 @@ Some built-in operators added in .NET 7 behave differently when overflowing than
 |CodeFix|False|
 ---
 
-## [CA2021](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2019): Do not call Enumerable.Cast\<T> or Enumerable.OfType\<T> with incompatible types
+## [CA2021](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2021): Do not call Enumerable.Cast\<T> or Enumerable.OfType\<T> with incompatible types
 
 Enumerable.Cast\<T> and Enumerable.OfType\<T> require compatible types to function expectedly.  
 
--- a/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
+++ b/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
@@ -1860,6 +1860,24 @@ Some built-in operators added in .NET 7 behave differently when overflowing than
 |CodeFix|False|
 ---
 
+## [CA2021](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2019): Do not call Enumerable.Cast\<T> or Enumerable.OfType\<T> with incompatible types
+
+Enumerable.Cast\<T> and Enumerable.OfType\<T> require compatible types to function expectedly.  
+
+The generic cast (IL 'unbox.any') used by the sequenced returned by Enumerable.Cast\<T> will throw InvalidCastException at runtime on elements of the types specified.  
+
+The generic type check (C# 'is' operator/IL 'isinst') used by Enumerable.OfType\<T> will never succeed with elements of types specified, resulting in an empty sequence.  
+
+Widening and user defined conversions are not supported with generic types.
+
+|Item|Value|
+|-|-|
+|Category|Reliability|
+|Enabled|True|
+|Severity|Warning|
+|CodeFix|False|
+---
+
 ## [CA2100](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2100): Review SQL queries for security vulnerabilities
 
 SQL queries that directly use user input can be vulnerable to SQL injection attacks. Review this SQL query for potential vulnerabilities, and consider using a parameterized SQL query.
--- a/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
+++ b/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
@@ -1864,7 +1864,7 @@ Some built-in operators added in .NET 7 behave differently when overflowing than
 
 Enumerable.Cast\<T> and Enumerable.OfType\<T> require compatible types to function expectedly.  
 
-The generic cast (IL 'unbox.any') used by the sequenced returned by Enumerable.Cast\<T> will throw InvalidCastException at runtime on elements of the types specified.  
+The generic cast (IL 'unbox.any') used by the sequence returned by Enumerable.Cast\<T> will throw InvalidCastException at runtime on elements of the types specified.  
 
 The generic type check (C# 'is' operator/IL 'isinst') used by Enumerable.OfType\<T> will never succeed with elements of types specified, resulting in an empty sequence.  
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
@@ -135,7 +135,6 @@ class C
         _ = {|#41:(new Apple[0]).Cast<Shoe>()|}; // error
         _ = {|#42:(new Shoe[0]).Cast<Fruit>()|}; // error
         _ = {|#43:(new Shoe[0]).Cast<Apple>()|}; // error
-        _ = {|#44:(new Shoe[0]).Cast<Shoe>()|}; // error
 
         // interface to class
         _ = (new ICar[0]).Cast<Plant>(); // subclass of Plant could implement ICar
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
@@ -4,7 +4,6 @@ using System;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.Testing;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
     Microsoft.NetCore.Analyzers.Runtime.DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzer,
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
@@ -1,0 +1,723 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Testing;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzer,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
+{
+    public class DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests
+    {
+        private readonly DiagnosticDescriptor castRule = DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzer.CastRule;
+        private readonly DiagnosticDescriptor ofTypeRule = DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzer.OfTypeRule;
+
+        [Fact]
+        public async Task OnlyWellKnownIEnumerable()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+namespace System.Linq
+{ 
+    interface IEnumerable {}
+    // this 'IEnumerable<T>' isn't 'well known', so the analyzer won't fire
+    interface IEnumerable<T> : IEnumerable {} 
+
+    static class Enumerable
+    {
+        public static IEnumerable<T> OfType<T>(this IEnumerable ienum) => null;
+    }
+
+    class C
+    {
+        void M()
+        {
+            _ = Enumerable.OfType<string>(default(IEnumerable<int>));
+            _ = default(IEnumerable<int>).OfType<string>();
+        }
+    }
+}
+");
+        }
+
+        [Fact()]
+        public async Task UnrelatedMethodsDontTrigger()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System.Linq
+{
+    static class Enumerable
+    {
+        // missing 'this'
+        public static IEnumerable<T> Cast<T>(IEnumerable ienum) => null;
+        // wrong parameter type
+        public static IEnumerable<T> Cast<T>(this IEnumerable<int> ienum) => null;
+        // too many parameters
+        public static IEnumerable<T> Cast<T>(this IEnumerable ienum, object distraction) => null;
+        // too many type parameters
+        public static IEnumerable<T> Cast<T, T2>(this IEnumerable ienum) => null;
+    }
+
+    class C
+    {
+        void M()
+        {
+            IEnumerable<int> e = default;
+            _ = Enumerable.Cast<string>(e);
+            _ = e.Cast<string>();
+            _ = e.Cast<string>(null);
+            _ = e.Cast<string, object>();
+        }
+    }
+}
+");
+        }
+
+        [Fact]
+        public async Task NonGenericCasesCSharp()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System.Linq;
+class Fruit {}
+class Apple : Fruit {}
+class Shoe {}
+
+interface ICar {}
+
+interface IPlant {}
+interface ITree : IPlant {}
+interface IGrass : IPlant {}
+
+class Plant : IPlant {}
+class Tree : ITree {}
+sealed class Grass : IGrass {}
+
+class C
+{
+    public void M()
+    {
+        _ = (new int[0]).OfType<object>();
+        _ = {|#10:(new int[0]).OfType<string>()|};
+        _ = (new object[0]).OfType<string>();
+
+        // we don't look through extra casts
+        _ = (new System.Uri[0] as object[]).OfType<string>();
+        _ = ((object[])new System.Uri[0]).OfType<string>();
+
+        // expression syntax
+        _ = from string s in new object[0]
+            select s;
+        _ = {|#11:from int i in new string[0]|}
+            select i;
+        _ = {|#12:Enumerable.Cast<string>(new int[0])|};
+
+        // interfaces
+        _ = (new ICar[0]).Cast<ICar>();
+        _ = (new ICar[0]).Cast<IPlant>();
+        _ = (new IPlant[0]).Cast<IPlant>();
+        _ = (new IPlant[0]).Cast<IPlant>();
+
+        // classes
+        _ = (new Fruit[0]).Cast<Fruit>(); // identity
+        _ = (new Fruit[0]).Cast<Apple>(); // upcast
+        _ = {|#40:(new Fruit[0]).Cast<Shoe>()|}; // error
+        _ = (new Apple[0]).Cast<Fruit>(); // downcast
+        _ = (new Apple[0]).Cast<Apple>(); // identity
+        _ = {|#41:(new Apple[0]).Cast<Shoe>()|}; // error
+        _ = {|#42:(new Shoe[0]).Cast<Fruit>()|}; // error
+        _ = {|#43:(new Shoe[0]).Cast<Apple>()|}; // error
+        _ = {|#44:(new Shoe[0]).Cast<Shoe>()|}; // error
+
+        // interface to class
+        _ = (new ICar[0]).Cast<Plant>(); // subclass of Plant could implement ICar
+        _ = (new ICar[0]).Cast<Tree>(); // subclass of Tree could implement ICar
+        _ = {|#50:(new ICar[0]).Cast<Grass>()|}; // subclass of Grass could not implement ICar, as Grass is sealed
+        _ = (new ICar[0]).Cast<Shoe>(); // subclass of Shoe could implement ICar
+
+        _ = (new IPlant[0]).Cast<Plant>();
+        _ = (new IPlant[0]).Cast<Tree>();
+        _ = (new IPlant[0]).Cast<Grass>();
+        _ = (new IPlant[0]).Cast<Shoe>();
+
+        _ = (new ITree[0]).Cast<Plant>();
+        _ = (new ITree[0]).Cast<Tree>();
+        _ = {|#51:(new ITree[0]).Cast<Grass>()|}; // subclass of Grass could not implement ICar, as Grass is sealed
+        _ = (new ITree[0]).Cast<Shoe>();
+
+        _ = (new IGrass[0]).Cast<Plant>();
+        _ = (new IGrass[0]).Cast<Tree>();
+        _ = (new IGrass[0]).Cast<Grass>();
+        _ = (new IGrass[0]).Cast<Shoe>();
+
+        // class to interface
+        _ = (new Plant[0]).Cast<ICar>();
+        _ = (new Tree[0]).Cast<ICar>();
+        _ = {|#52:(new Grass[0]).Cast<ICar>()|}; // grass doesn't implement ICar, and is sealed so a subclass couldn't either
+        _ = (new Shoe[0]).Cast<ICar>();
+
+        _ = (new Plant[0]).Cast<IPlant>();
+        _ = (new Tree[0]).Cast<IPlant>();
+        _ = (new Grass[0]).Cast<IPlant>();
+        _ = (new Shoe[0]).Cast<IPlant>();
+
+        _ = (new Plant[0]).Cast<ITree>();
+        _ = (new Tree[0]).Cast<ITree>();
+        _ = {|#53:(new Grass[0]).Cast<ITree>()|};
+        _ = (new Shoe[0]).Cast<ITree>();
+
+        _ = (new Plant[0]).Cast<IGrass>();
+        _ = (new Tree[0]).Cast<IGrass>();
+        _ = (new Grass[0]).Cast<IGrass>();
+        _ = (new Shoe[0]).Cast<IGrass>();
+    }
+}
+",
+
+VerifyCS.Diagnostic(ofTypeRule).WithLocation(10).WithArguments("int", "string"),
+VerifyCS.Diagnostic(castRule).WithLocation(11).WithArguments("string", "int"),
+VerifyCS.Diagnostic(castRule).WithLocation(12).WithArguments("int", "string"),
+
+VerifyCS.Diagnostic(castRule).WithLocation(40).WithArguments("Fruit", "Shoe"),
+VerifyCS.Diagnostic(castRule).WithLocation(41).WithArguments("Apple", "Shoe"),
+VerifyCS.Diagnostic(castRule).WithLocation(42).WithArguments("Shoe", "Fruit"),
+VerifyCS.Diagnostic(castRule).WithLocation(43).WithArguments("Shoe", "Apple"),
+
+VerifyCS.Diagnostic(castRule).WithLocation(50).WithArguments("ICar", "Grass"),
+VerifyCS.Diagnostic(castRule).WithLocation(51).WithArguments("ITree", "Grass"),
+VerifyCS.Diagnostic(castRule).WithLocation(52).WithArguments("Grass", "ICar"),
+VerifyCS.Diagnostic(castRule).WithLocation(53).WithArguments("Grass", "ITree")
+);
+        }
+
+        [Fact]
+        public async Task ArrayCSharp()
+        {
+            Assert.Throws<InvalidCastException>(()
+                => (new int[][] { new int[] { 1 } }).Cast<object[]>().ToArray());
+            Assert.Throws<InvalidCastException>(()
+                => (new[] { 1 }).Cast<string>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new object[][] { Array.Empty<object>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new ValueType[][] { Array.Empty<ValueType>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new object[][] { Array.Empty<ValueType>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new ValueType[][] { Array.Empty<ValueType>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new int[][] { Array.Empty<int>() }).Cast<ValueType[]>().ToArray());
+
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+using System.Linq;
+
+class C
+{
+    void M()
+    {
+        Enumerable.OfType<object[]>(new int[0][]);
+        Enumerable.OfType<object[]>(new string[0][]);
+        Enumerable.OfType<object[]>(new object[0][]);
+
+        // these should all be errors, but not all implemented
+        Enumerable.OfType<int[]>(new object[0][]);
+        Enumerable.OfType<string[]>(new object[0][]);
+
+        {|#12:Enumerable.OfType<string[]>(new int[0][])|};
+        {|#13:Enumerable.OfType<string>(new int[0][])|};
+
+        // multidimensional arrays don't implement IEnumberable<T>
+
+        // IEnumerable<string[,]> != string[,]
+        {|#20:Enumerable.OfType<string[,]>(new string[0,0])|}; 
+
+        // these will throw
+        {|#21:Enumerable.OfType<string[]>(new int[0,0])|};
+        {|#22:Enumerable.OfType<string[,]>(new int[0,0])|};
+
+        // arrays of multidimensional arrays are checked, including rank
+        {|#30:Enumerable.OfType<string[]>(new string[0][,])|};
+        {|#31:Enumerable.OfType<string[,]>(new string[0][])|};
+        {|#32:Enumerable.OfType<string[]>(new int[0][,])|};
+        {|#33:Enumerable.OfType<string[,]>(new int[0][])|};
+
+        // all arrays can be cast to and from System.Array
+        Enumerable.OfType<Array>(Enumerable.Empty<Array>());
+
+        Enumerable.OfType<Array>(Enumerable.Empty<object[]>());
+        Enumerable.OfType<Array>(Enumerable.Empty<object[]>());
+        Enumerable.OfType<object[]>(Enumerable.Empty<Array>());
+        Enumerable.OfType<string[]>(Enumerable.Empty<Array>());
+
+        Enumerable.OfType<Array>(Enumerable.Empty<ValueType[]>());
+        Enumerable.OfType<ValueType[]>(Enumerable.Empty<Array>());
+
+        // but not non-arrays
+        {|#41:Enumerable.OfType<Array>(Array.Empty<int>())|};
+        Enumerable.OfType<Array>(Array.Empty<object>());
+        {|#42:Enumerable.OfType<Array>(Array.Empty<string>())|};
+        {|#43:Enumerable.OfType<Array>(Array.Empty<ValueType>())|};
+        {|#44:Enumerable.OfType<Array>(Array.Empty<Enum>())|};
+
+        {|#51:Enumerable.OfType<int>(Array.Empty<Array>())|};
+        Enumerable.OfType<object>(Array.Empty<Array>());
+        {|#52:Enumerable.OfType<string>(Array.Empty<Array>())|};
+        {|#53:Enumerable.OfType<ValueType>(Array.Empty<Array>())|};
+        {|#54:Enumerable.OfType<Enum>(Array.Empty<Array>())|};
+
+        // this might work
+        Enumerable.OfType<int[]>(Enumerable.Empty<ValueType[]>());
+        // this is not allowed, but not implemented yet
+        Enumerable.OfType<ValueType[]>(Enumerable.Empty<int[]>());
+    }
+}",
+    //   VerifyCS.Diagnostic(ofTypeRule).WithLocation(10).WithArguments("int[]", "object[]"),
+    //   VerifyCS.Diagnostic(ofTypeRule).WithLocation(11).WithArguments("string[]", "object[]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(12).WithArguments("int[]", "string[]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(13).WithArguments("int[]", "string"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(20).WithArguments("string", "string[*,*]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(21).WithArguments("int", "string[]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(22).WithArguments("int", "string[*,*]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(30).WithArguments("string[*,*]", "string[]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(31).WithArguments("string[]", "string[*,*]"),
+    VerifyCS.Diagnostic(ofTypeRule).WithLocation(32).WithArguments("int[*,*]", "string[]"),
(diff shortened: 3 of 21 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=4328) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=4328) before we can accept your contribution.<br/><br/>:x: fowl2 [sign now](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=4328)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/roslyn-analyzers?pullRequest=4328) it.</sub>

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    return;
+                }
+
+                compilationStartContext.RegisterOperationAction(operationContext =>
+                {
+                    var invocation = (IInvocationOperation)operationContext.Operation;
+
+                    var targetMethod = invocation.TargetMethod.ReducedFrom ?? invocation.TargetMethod;
+                    if (targetMethod == null || !methodMetadataNames.TryGetValue(targetMethod.Name, out var rule))
```
mavasani, line 76: TargetMethod cannot be null
```suggestion
                    if (!methodMetadataNames.TryGetValue(targetMethod.Name, out var rule))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                                                                             s_localizableCastDescription,
+                                                                             isPortedFxCopRule: false,
+                                                                             isDataflowRule: false);
+
+        internal static DiagnosticDescriptor OfTypeRule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                             s_localizableOfTypeTitle,
+                                                                             s_localizableOfTypeMessage,
+                                                                             DiagnosticCategory.Reliability,
+                                                                             RuleLevel.IdeSuggestion,
+                                                                             description: null,
```
mavasani, line 43: We generally recommend specifying a description with detailed explanation of the issue and recommended fix.

fowl2:
That should be all the warnings resolved except for the rule id. Not sure how those are are assigned.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
   <data name="DoNotAddArchiveItemPathToTheTargetFileSystemPathMessage" xml:space="preserve">
     <value>When creating path for '{0} in method {1}' from relative archive item path to extract file and the source is an untrusted zip archive, make sure to sanitize relative archive item path '{2} in method {3}'</value>
   </data>
+  <data name="DoNotCallEnumerableCastThatWillFailDescription" xml:space="preserve">
+    <value>If the source sequence contains any elements, the sequence returned by Cast will throw InvalidCastException at runtime when enumerated. Change the type(s) to those that are compatible or remove the call.</value>
+  </data>
+  <data name="DoNotCallEnumerableCastThatWillFailMessage" xml:space="preserve">
+    <value>Enumerable.Cast&lt;T&gt; will '{0}' can not be castetype '{1}' will fail</value>
```
Evangelink, line 1096: What about `Type '{0}' cannot be converted to '{1}' and will lead to a runtime failure`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
   <data name="DoNotAddArchiveItemPathToTheTargetFileSystemPathMessage" xml:space="preserve">
     <value>When creating path for '{0} in method {1}' from relative archive item path to extract file and the source is an untrusted zip archive, make sure to sanitize relative archive item path '{2} in method {3}'</value>
   </data>
+  <data name="DoNotCallEnumerableCastThatWillFailDescription" xml:space="preserve">
+    <value>If the source sequence contains any elements, the sequence returned by Cast will throw InvalidCastException at runtime when enumerated. Change the type(s) to those that are compatible or remove the call.</value>
```
Evangelink, line 1093:
```suggestion
    <value>If the source sequence contains any element, the call to 'Cast' will throw an  'InvalidCastException' at runtime when enumerated. Make sure the types are compatible or remove this call.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
...
+  </data>
+  <data name="DoNotCallEnumerableCastThatWillFailMessage" xml:space="preserve">
+    <value>Enumerable.Cast&lt;T&gt; will '{0}' can not be castetype '{1}' will fail</value>
+  </data>
+  <data name="DoNotCallEnumerableCastThatWillFailTitle" xml:space="preserve">
+    <value>Do not call Enumerable.Cast&lt;T&gt; with incompatible types</value>
+  </data>
+  <data name="DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesMessage" xml:space="preserve">
+    <value>Source sequence elements are of type '{1}' which is not compatible with provided result element type '{0}'</value>
```
Evangelink, line 1102: `Type '{0}' cannot be converted to '{1}' and will results in empty sequence`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
...
+  </data>
+  <data name="DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesTitle" xml:space="preserve">
+    <value>The enumerable returned by this call to OfType&lt;T&gt; will never return any values</value>
+  </data>
+  <data name="DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesMessageDescription" xml:space="preserve">
+    <value>The type of the elements in the source sequence is not compatible with the type passed to OfType&lt;T&gt;, therefore no elements will ever be returned. Change the type of the elements in the source sequence, the type passed to OfType&lt;T&gt;, or remove the call.</value>
```
Evangelink, line 1108:
```suggestion
    <value>This call to 'OfType' will always results in an empty sequence. Make sure the types are compatible or remove this call.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// CAxxxx: Do not call Enumerable.Cast with incompatible types.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotCallEnumerableCastThatWillFailAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA9999";
```
Evangelink, line 21: Let's use the next available in `Usage` (i.e. `CA2250`), see https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/usage-warnings

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+        private static readonly LocalizableString s_localizableOfTypeMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableOfTypeDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesMessageDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor CastRule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                             s_localizableCastTitle,
+                                                                             s_localizableCastMessage,
+                                                                             DiagnosticCategory.Reliability,
+                                                                             RuleLevel.BuildWarningCandidate,
```
Evangelink, line 34: Use `IdeSuggestion`. A rule should never be promoted to warning without having been verified and validated by MS.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                                                                             s_localizableCastDescription,
+                                                                             isPortedFxCopRule: false,
+                                                                             isDataflowRule: false);
+
+        internal static DiagnosticDescriptor OfTypeRule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                             s_localizableOfTypeTitle,
+                                                                             s_localizableOfTypeMessage,
+                                                                             DiagnosticCategory.Reliability,
+                                                                             RuleLevel.BuildWarningCandidate,
```
Evangelink, line 43: Same remark.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+            => methodMetadataNames.Values.ToImmutableArray();
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(compilationStartContext =>
+            {
+                var enumerableType = compilationStartContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable);
```
Evangelink, line 64: You could use the `TryGet` method as you bail-out when any is not present.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
+            => methodMetadataNames.Values.ToImmutableArray();
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(compilationStartContext =>
```
Evangelink, line 62: Use `context` instead of `compilationStartContext`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                var enumerableType = compilationStartContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable);
+                var genericIEnumerableType = compilationStartContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericIEnumerable1);
+
+                if (enumerableType == null || genericIEnumerableType == null)
+                {
+                    return;
+                }
+
+                compilationStartContext.RegisterOperationAction(operationContext =>
```
Evangelink, line 72: Use `context` instead of `operationContext`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    var invocation = (IInvocationOperation)operationContext.Operation;
+
+                    var targetMethod = invocation.TargetMethod.ReducedFrom ?? invocation.TargetMethod;
```
Evangelink, line 76:
```suggestion
                    var targetMethod = invocation.TargetMethod.OriginalDefinition;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                    if (!methodMetadataNames.TryGetValue(targetMethod.Name, out var rule))
```
Evangelink, line 77: We usually prefer a symbol comparison rather than a name comparison.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                        return;
+                    }
+
+                    var numberOfArguments = invocation.IsExtensionMethodAndHasNoInstance() ? 1 : 0;
+                    if (!invocation.Arguments.HasExactly(numberOfArguments))
+                    {
+                        return;
+                    }
+
+                    var arg = invocation.IsExtensionMethodAndHasNoInstance()
```
Evangelink, line 93: I need to look for it but I am sure that we have something similar and that there is a little trick for vbnet.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
```
Evangelink, line 4:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
```diff
@@ -0,0 +1,130 @@
...
+using Microsoft.CodeAnalysis.Testing;
+
```
Evangelink, line 6:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
```diff
@@ -0,0 +1,130 @@
...
+using Xunit;
+
```
Evangelink, line 8:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
```
Evangelink, line 6:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
```diff
@@ -0,0 +1,130 @@
...
+    public void M()
+    {
+        var a = (new int[0]).OfType<object>();
+        var b = (new int[0]).OfType<string>();
+        var c = (new object[0]).OfType<string>();
+    }
+}
+",
+    // /0/Test0.cs(9,17): info CA9999: If the source sequence contains any elements, the sequence returned by Cast will throw InvalidCastException at runtime when enumerated.
+    VerifyCS.Diagnostic(ofTypeRule).WithSpan(9, 17, 9, 46).WithArguments("int", "string")
```
Evangelink, line 39: Could you please use inline markup notation rather than `WithSpan` it is easier to read and comprehend (e.g. `var a = (new int[0]).{|#0:OfType<object>()|}`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
...
     <value>When creating path for '{0} in method {1}' from relative archive item path to extract file and the source is an untrusted zip archive, make sure to sanitize relative archive item path '{2} in method {3}'</value>
   </data>
+  <data name="DoNotCallEnumerableCastThatWillFailDescription" xml:space="preserve">
+    <value>If the source sequence contains any elements, the sequence returned by Cast will throw InvalidCastException at runtime when enumerated. Change the type(s) to those that are compatible or remove the call.</value>
+  </data>
+  <data name="DoNotCallEnumerableCastThatWillFailMessage" xml:space="preserve">
+    <value>Enumerable.Cast&lt;T&gt; will '{0}' can not be castetype '{1}' will fail</value>
+  </data>
+  <data name="DoNotCallEnumerableCastThatWillFailTitle" xml:space="preserve">
+    <value>Do not call Enumerable.Cast&lt;T&gt; with incompatible types</value>
```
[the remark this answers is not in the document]
fowl2: I don't think rule titles support format items, I can't see any others that are using them.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
fowl2: Is 'converted' the correct terminology?  How about `Type '{0}' is incompatible with type '{1}' and will cause an exception at runtime`?

Evangelink: From the doc page they say:

> If an element cannot be converted to type `TResult`, this method throws a `InvalidCastException`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
Evangelink: Except if I am missing something this isn't fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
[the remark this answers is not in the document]
fowl2: ok I'll leave it as

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
fowl2: this does seem pretty messy. I remember looking for quite a while but couldn't find anything.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
fowl2: Sorry, I'm ticking them off as I go. Pushed now fix now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1,17 +1,17 @@
...
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+
```
Youssef1313, line 14: Was this file edited manually?
It's strange to see these extra spaces.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
   <data name="DoNotAddArchiveItemPathToTheTargetFileSystemPathMessage" xml:space="preserve">
     <value>When creating path for '{0} in method {1}' from relative archive item path to extract file and the source is an untrusted zip archive, make sure to sanitize relative archive item path '{2} in method {3}'</value>
   </data>
+  <data name="DoNotCallEnumerableCastThatWillFailDescription" xml:space="preserve">
+    <value>If the source sequence contains any element, the call to 'Cast' will throw an  'InvalidCastException' at runtime when enumerated. Make sure the types are compatible or remove this call.</value>
```
Youssef1313, line 1093: remove extra space.

```suggestion
    <value>If the source sequence contains any element, the call to 'Cast' will throw an 'InvalidCastException' at runtime when enumerated. Make sure the types are compatible or remove this call.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
fowl2: found it! `GetInstance()`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
fowl2: This is the format output by the test runner - is there any way to get it in this format or will this have to be done manually?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
fowl2: How the the approach I've just pushed?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+                        {
+                            return false;
+                        }
+
+                        if (castTo.DerivesFrom(castFrom))
+                        {
+                            return false;
+                        }
+
+                        return true;
```
Evangelink, line 144:
```suggestion
                        return !castFrom.DerivesFrom(castTo) &&
                        	!castTo.DerivesFrom(castFrom));
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
```diff
@@ -0,0 +1,127 @@
...
+    );
+        }
+
+        [Fact]
+        public async Task DiagnosticCasesVB()
```
Evangelink, line 41: It'd be nice to have an equivalent in term of cases used for C# and vbnet.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1089,6 +1089,24 @@
...
+    <value>Type '{0}' is incompatible with type '{1}' and will cause an exception at runtime</value>
+  </data>
+  <data name="DoNotCallEnumerableCastThatWillFailTitle" xml:space="preserve">
+    <value>Do not call 'Enumerable.Cast&lt;T&gt;' with incompatible types</value>
+  </data>
+  <data name="DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesMessage" xml:space="preserve">
+    <value>Type '{0}' is incompatible with type '{1}' and will result in an empty sequence</value>
+  </data>
+  <data name="DoNotCallEnumerablOfTypeThatWillNeverReturnAnyValuesTitle" xml:space="preserve">
+    <value>The enumerable returned by this call to OfType&lt;T&gt; will never return any values</value>
```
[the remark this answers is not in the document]
fowl2: the other strings don't seem to quote these

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+                    var instanceArg = invocation.GetInstance(); // "this" argument of an extension method
+
+                    // because the type of the parameter is actually the non-generic IEnumerable,
+                    // we have to reach back through conversion operator(s) to get the "real" type of the argument
+                    if (instanceArg is not IConversionOperation conversionOperation)
+                    {
+                        return;
+                    }
+
+                    var argIEnumerableType = conversionOperation.Operand.Type.AllInterfaces.SingleOrDefault(t => t.OriginalDefinition.SpecialType == SpecialType.System_Collections_Generic_IEnumerable_T);
```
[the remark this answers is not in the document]
fowl2: I have no idea what more than one `System.Collections.Generic.IEnumerable<T>` would means. I think a failure would be justified?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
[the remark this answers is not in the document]
fowl2: ahh I'll do it anyway

buyaa-n (changes requested):
@fowl2 The analyzer proposal has not yet been approved, please wait until the proposal goes through API review

carlossanlop:
The proposal has been approved so we can unblock the PR, @buyaa-n

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
fowl2: Will do.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqEnumerable, out var enumerableType))
+                {
+                    return;
+                }
+
+                var methodRuleDictionary = methodMetadataNames
+                    .SelectMany(m => enumerableType.GetMembers(m.MethodName)
+                                                   .OfType<IMethodSymbol>()
+                                                   .Where(method => method.TypeParameters.HasExactly(1)
+                                                                 && method.Parameters.HasExactly(1))
```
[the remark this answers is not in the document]
fowl2: @mavasani any thoughts on the best way of locating a particular member?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+                    {
+                        return;
+                    }
+
+                    var castFrom = argIEnumerableType.TypeArguments.Single();
+                    var castTo = invocation.TargetMethod.TypeArguments.Single();
+
+                    if (CastWillAlwaysFail(castFrom, castTo))
+                    {
+                        context.ReportDiagnostic(invocation.CreateDiagnostic(rule, castFrom.ToDisplayString(), castTo.ToDisplayString()));
```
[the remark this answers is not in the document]
fowl2: @mavasani is `ToDisplayString` ok in a diagnostic? Some existing analyzers do use it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
fowl2: I'll (hopefully) get a chance to work on improving the tests this weekend. @Evangelink if you could provide feedback on the other comments, that's be great.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System.Linq;
+
+class C
+{
+    public void M()
+    {
+        var a = (new int[0]).OfType<object>();
+        var b = (new int[0]).OfType<string>();
+        var c = (new object[0]).OfType<string>();
```
buyaa-n, line 31: Could we have tests for `.Cast<T>()`?

buyaa-n (changes requested):
Overall looks good @fowl2 thank you for your contribution, added few comment, and seems we could add some more unit tests.
Not sure if you read our [Definition of done](https://github.com/dotnet/roslyn-analyzers/blob/master/docs/NetCore_GettingStarted.md#definition-of-done) list, it would be great if you could try testing the new analyzer with runtime repo, please let us know if you would like to give it a try

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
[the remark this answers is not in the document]
buyaa-n: It should be OK, we might prefer to set the display format though

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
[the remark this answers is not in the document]
mavasani: Yes, we never use `SingleOrDefault` in analyzers looking for specific symbol structure as that can cause analyzer crash for invalid code. I agree on using `FirstOrDefault`. Additionally, `Operand.Type` can be null as well, so needs to be `Operand.Type?.AllInterfaces`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,160 @@
...
+                        return;
+                    }
+
+                    var argIEnumerableType = conversionOperation.Operand.Type.AllInterfaces.SingleOrDefault(t => t.OriginalDefinition.SpecialType == SpecialType.System_Collections_Generic_IEnumerable_T);
+                    if (argIEnumerableType is null)
+                    {
+                        return;
+                    }
+
+                    var castFrom = argIEnumerableType.TypeArguments[0];
```
mavasani, line 106: Can the type argument count be 0 for invalid code? I'd rather be defensive and check `argIEnumerableType.TypeArguments.IsEmpty` in the above bail out check.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,160 @@
...
+                    var castTo = invocation.TargetMethod.TypeArguments[0];
```
mavasani, line 107: Same here - we should check for `invocation.TargetMethod.TypeArguments.IsEmpty`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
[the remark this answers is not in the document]
mavasani: `ToDisplayString` can produce really large strings in general, but I think it would be fine for type arguments.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,160 @@
...
+
+                    if (CastWillAlwaysFail(castFrom, castTo))
+                    {
+                        context.ReportDiagnostic(invocation.CreateDiagnostic(rule, castFrom.ToDisplayString(), castTo.ToDisplayString()));
+                    }
+                }, OperationKind.Invocation);
+            });
+
+            static bool CastWillAlwaysFail(ITypeSymbol castFrom, ITypeSymbol castTo)
```
mavasani, line 116: Hmm, I am wondering if we can use any compiler API here instead of trying to mimic conversion logic. @333fred Is there any compiler API that given two type symbols returns the `Conversion` info for it?

fowl2: One little/important detail here is that there are some subtle differences between the language  and IL casts. Eg. C# will emit calls to explicit and numeric conversion operators but the runtime will not. 

Note to self: ensure there is a test for this.

mavasani:

mavasani (dismissed):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
MaStr11: Would also be good to have a test for `.Cast<T>` in [query expression syntax](https://docs.microsoft.com/en-US/dotnet/csharp/programming-guide/concepts/linq/converting-data-types)

```CS
_ = from string s in new object[0]
    select s;
```

buyaa-n:
@fowl2 could you address the comments? Do you need a help with anything?

Also please retarget to https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx

fowl2:
@buyaa-n sorry I haven't had time to look at this - looks like I'll have some spare time this week to get back to you

danmoseley:
@fowl2 could you give us an update? It would be great to have this analyzer. Do you expect to have more time?

buyaa-n:
ping @fowl2 again, are you planning to work on it

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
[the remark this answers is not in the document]
fowl2: Am I understanding that would cause this not to trigger without the extension method syntax?

ie. is there a reason why there should be a difference between
 `(new string[] { "not an int" }).Cast<int>();` and
 `Enumerable.Cast<int>(new string[] { "not an int" });`?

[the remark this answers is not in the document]
buyaa-n: > Am I understanding that would cause this not to trigger without the extension method syntax?

Yes, and it is better to only restrict to the existing implementation of `Enumerable.Cast<T>/OfType<T>` which takes `this` parameter. There is no other overload for Enumerable.Cast<T>,  and if an overload is added in the future we don't know that overload would have the same behavior and needs to warn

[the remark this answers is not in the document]
buyaa-n: > ie. is there a reason why there should be a difference between

I don't see any difference between them for the analyzer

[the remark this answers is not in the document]
fowl2: ahh, I was confusing method definition and call site. I used `method.IsExtensionMethod` instead.

fowl2:
@buyaa-n it looks like the tooling doesn't really like having two different messages for the same rule id, should I split this into `CA2017` and `CA2018`?

buyaa-n:
> @buyaa-n it looks like the tooling doesn't really like having two different messages for the same rule id, should I split this into CA2017 and CA2018?

Why? Different messages should be OK, it seems some more files not generated yet, please run `msbuild /t:pack`

I see the message : `One or more auto-generated documentation files were either edited manually, or not updated. Please revert changes made to the following files (if manually edited) and run "msbuild /t:pack" for each solution at the root of the repo to automatically update them` in build failure

fowl2:
> Why? Different messages should be OK, it seems some more files not generated yet, please run `msbuild /t:pack`

After doing this `src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.{md,sarif}` only contains the first description.
 
> By the way you don't need to retarget to `6.0.0-xyz` branches anymore as 6.0.0 is cut for new analyzers, this will be merged into main and will be added to 7.0

I've merged from `main`.

----

Spent a bit of time trying to work out how to test adding `System.Linq.Enumerable` overloads but it seems more trouble than it's worth. Anyway, still working on tests, I'll resolve the conversations when ready.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
fowl2: I added a bunch of c# tests, not exactly equivalent, but hopefully enough.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzerTests.cs
fowl2: added some for both

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
fowl2: I couldn't end up finding any API for this.

buyaa-n:
> After doing this src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.{md,sarif} only contains the first description.

Yes, it is generated with the title/description of each rule, a rule could have multiple messages but should have only one title and one description. Could you integrate the title and description into one? Something like `Do not call Enumerable.Cast or Enumerable.OfType with incompatible types.`

> I've merged from main.

I was wondering why the https://github.com/dotnet/roslyn-analyzers/blob/08382fdbdfe11a6af166d8ebe6110699551f03ab/src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md has only one analyzer while we have added many in 6.0. Got confirmed that `main` is only for nuget releases, no more new analyzers added there, you need to retarget the `release/7.0.1xx` branch, see https://github.com/dotnet/roslyn-analyzers/blob/main/docs/NetCore_GettingStarted.md#branch-definitions for more info

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastThatWillFailAnalyzer.cs
```diff
@@ -0,0 +1,319 @@
...
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// CA2017: Do not call Enumerable.Cast or Enumerable.OfType with incompatible types.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class DoNotCallEnumerableCastThatWillFailAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string CastRuleId = "CA2017";
+        internal const string OfTypeRuleId = "CA2017";
```
buyaa-n, line 20: The same id would not work, if you don't want to merge the title and description we can consider using a different id, that might need to go through API review cc @terrajobst

fowl2:
having trouble due to https://github.com/dotnet/docs/issues/24894

thousands of 
> CSC : error AD0001: Analyzer 'Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.PropertiesShouldNotBeWriteOnlyAnalyzer' threw an exception of type 'System.InvalidOperationException' with message 'Passing '.editorconfig' files as addition
> al files is no longer needed. It will be implicitly discovered (if the file is in the project's directory or any ancestor directory), or it should be converted into a 'globalconfig'. See 'https://docs.microsoft.com/dotnet/fundamentals/code
> -analysis/configuration-files'.'. [\roslyn-analyzers\src\Microsoft.CodeAnalysis.AnalyzerUtilities\Microsoft.CodeAnalysis.AnalyzerUtilities.csproj]

Youssef1313:
@fowl2 Can you try deleting this:

https://github.com/dotnet/roslyn-analyzers/blob/1ffb3b40946918f1900d566f2ad3ddd3aaa11bdd/Directory.Build.props#L55-L57


@jeffhandley @buyaa-n Anyone knows how this can happen locally but not in CI? The error seems to make sense to me and it should have been reported by CI. This seems quite odd to me 👀

fowl2:
removing that line seems to have done something

I don't think I've copied the right file, as I'm now getting 3x
> error CA1069: The enum member 'IdeSuggestion' has the same constant value '3' as member 'BuildWarningCandidate'

But nothing for this rule - which should at least be picking up something from its own unit test. 

Which package+version am I supposed to be overwriting? There's about 20 different ones [in here](https://github.com/dotnet/roslyn-analyzers/blob/release/7.0.1xx/eng/Versions.props).

fowl2:
@Youssef1313 do you have any hints as to how to run test against `dotnet/roslyn-analyzers`?

Youssef1313:
> @Youssef1313 do you have any hints as to how to run test against `dotnet/roslyn-analyzers`?

Can you try the instructions [here](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/NetCore_GettingStarted.md#testing-against-the-runtime-and-roslyn-analyzers-repo)?

fowl2:
Yes that's what I was trying. See my previous comment :)

Youssef1313:
Probably @buyaa-n or @mavasani can help better here.

fowl2:
done:
* rebased on main: still builds ✅🎉
* got it running against `roslyn-analyzers`: no problems found
* double checked by adding a problem: found 🎉

to be:
* check against `runtime` (ran into some python dep issue) and `roslyn` (unrelated laptop death)

fowl2:
I've rebased to main, changed to new diagnostic ID (again) and updated to the new `DiagnosticDescriptor`/`LocalizableString` pattern.

Ran out of time/cpu again trying to run it against `runtime` and `rosyln`, could I ask someone else have a go to help get this over the line? :)

buyaa-n (commented):
> Ran out of time/cpu again trying to run it against runtime and rosyln, could I ask someone else have a go to help get this over the line? :)

Tested with runtime looks good and no diagnostic found, thank you.

Overall looks great to me, left a few questions

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,722 @@
...
+                => (new ValueType[][] { Array.Empty<ValueType>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new object[][] { Array.Empty<ValueType>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new ValueType[][] { Array.Empty<ValueType>() }).Cast<int[]>().ToArray());
+
+            Assert.Throws<InvalidCastException>(()
+                => (new int[][] { Array.Empty<int>() }).Cast<ValueType[]>().ToArray());
```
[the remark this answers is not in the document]
fowl2: the idea here was two-fold:
1. it was useful to double check the logic while writing the analyzer
2. this could potentially detect if the behavior changes in future runtime/bcl updates. This is pretty unlikely though, so I wouldn't object to removing them.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,722 @@
...
+        // classes
+        _ = (new Fruit[0]).Cast<Fruit>(); // identity
+        _ = (new Fruit[0]).Cast<Apple>(); // upcast
+        _ = {|#40:(new Fruit[0]).Cast<Shoe>()|}; // error
+        _ = (new Apple[0]).Cast<Fruit>(); // downcast
+        _ = (new Apple[0]).Cast<Apple>(); // identity
+        _ = {|#41:(new Apple[0]).Cast<Shoe>()|}; // error
+        _ = {|#42:(new Shoe[0]).Cast<Fruit>()|}; // error
+        _ = {|#43:(new Shoe[0]).Cast<Apple>()|}; // error
+        _ = {|#44:(new Shoe[0]).Cast<Shoe>()|}; // error
```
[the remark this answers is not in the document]
fowl2: The comment is wrong - it doesn't warn here and there's no `VerifyCS.Diagnostic(castRule).WithLocation(44)` below.

I've removed the line.

buyaa-n (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DoNotCallEnumerableCastOrOfTypeWithIncompatibleTypesAnalyzerTests.cs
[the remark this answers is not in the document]
buyaa-n: I am OK with keeping, defer to @mavasani

buyaa-n:
Thank you @fowl2 LGTM!

@mavasani please look at the open questions and you also have out-dated change request

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -1860,6 +1860,24 @@ Some built-in operators added in .NET 7 behave differently when overflowing than
...
 
+## [CA2021](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2021): Do not call Enumerable.Cast\<T> or Enumerable.OfType\<T> with incompatible types
+
+Enumerable.Cast\<T> and Enumerable.OfType\<T> require compatible types to function expectedly.  
+
+The generic cast (IL 'unbox.any') used by the sequenced returned by Enumerable.Cast\<T> will throw InvalidCastException at runtime on elements of the types specified.  
+
+The generic type check (C# 'is' operator/IL 'isinst') used by Enumerable.OfType\<T> will never succeed with elements of types specified, resulting in an empty sequence.  
+
+Widening and user defined conversions are not supported with generic types.
```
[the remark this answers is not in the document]
buyaa-n: Confirmed offline that it is OK

PR #4333, opened 2020-10-14
Re-enable RS1024

Youssef1313:
The linked issue was fixed. This is **not** ready to merge until I check whether the current used version in the repository has this fix or not.

The current version doesn't have the fix, and it's very tedious to manually fix those.
I was hoping that the FixAll will fix what it can, and leave what it can't, but it doesn't fix any.

I filed https://github.com/dotnet/roslyn/issues/48607


Youssef1313:
@mavasani There are **many** violations of this rule. It was disabled because a bug in "codefix" which doesn't affect the build at all. I believe we should get it back soon regardless whether the current version has the fix.

Youssef1313:
@mavasani The current codefix has a bug https://github.com/dotnet/roslyn-analyzers/issues/4337. Also, SymbolEqualityComparer isn't *always* available to use. That makes things more harder to handle. Before I work more on this, I want to make sure that you want to enable the rule.

mavasani:
> Before I work more on this, I want to make sure that you want to enable the rule.

@Youssef1313 We certainly want to enable this rule in this repo.

mavasani:
> Also, SymbolEqualityComparer isn't always available to use

@Youssef1313 We can start by enabling the rule only for https://github.com/dotnet/roslyn-analyzers/tree/master/src/NetAnalyzers if that makes it easier for you.

sharwell:
We want to enable this rule, but the Fix All needs to work first. The only reason it's disabled is it's broken.

Evangelink:
Might be worth having a new-look when #4477 is merged.

mavasani:
@Youssef1313 https://github.com/dotnet/roslyn-analyzers/pull/4477 has been merged. Want to take another crack at this after upgrading our analyzer package reference?

Youssef1313:
@mavasani The repository is using 3.3.2 which is the latest on nuget.org. The dotnet6 feed doesn't contain this package, and the dotnet5 feed contains an older version. So it looks like the fix from #4477 isn't yet released?

PR #4340, opened 2020-10-15
Fix bugs in CallSiteImplicitAllocationAnalyzer

Youssef1313:
The bug stated in https://github.com/microsoft/RoslynClrHeapAllocationAnalyzer/issues/84 existed in the copy of the analyzer in this repository too.

I couldn't fix it without refactoring. Relying on syntax here is very error-prone. So I switched to getting `IOperation`s. However, the analyzer is still a syntax one its base class is so.

### DIFF
--- a/src/PerformanceSensitiveAnalyzers/CSharp/Analyzers/CallSiteImplicitAllocationAnalyzer.cs
+++ b/src/PerformanceSensitiveAnalyzers/CSharp/Analyzers/CallSiteImplicitAllocationAnalyzer.cs
@@ -68,20 +68,16 @@ namespace Microsoft.CodeAnalysis.CSharp.PerformanceSensitiveAnalyzers
             //     static void Fun2(int i = 0, params object[] args) {}
             foreach (var argument in invocationOperation.Arguments)
             {
-                if (argument.ArgumentKind != ArgumentKind.ParamArray)
+                if (argument.ArgumentKind == ArgumentKind.ParamArray)
                 {
-                    continue;
+                    // Up to net45 the System.Array.Empty<T> singleton didn't existed so an empty params array was still causing some memory allocation.
+                    if (argument.IsImplicit &&
+                        (!compilationHasSystemArrayEmpty || (argument.Value as IArrayCreationOperation)?.Initializer.ElementValues.IsEmpty == true))
+                    {
+                        reportDiagnostic(node.CreateDiagnostic(ParamsParameterRule));
+                    }
+                    break;
                 }
-
-                bool isEmpty = (argument.Value as IArrayCreationOperation)?.Initializer.ElementValues.IsEmpty == true;
-
-                // Up to net45 the System.Array.Empty<T> singleton didn't existed so an empty params array was still causing some memory allocation.
-                if (argument.IsImplicit && (!isEmpty || !compilationHasSystemArrayEmpty))
-                {
-                    reportDiagnostic(node.CreateDiagnostic(ParamsParameterRule));
-                }
-
-                break;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
```diff
@@ -161,6 +161,34 @@ public void Testing()
             await VerifyCS.VerifyAnalyzerAsync(sampleProgram);
         }
 
+        [Fact]
+        public async Task ParamsIsPrecededByOptionalParameters()
```
Evangelink, line 165: Can you do something equivalent for vbnet?

Youssef1313: @Evangelink This is a C# only analyzer.
I think we should file an issue to support VB for it. It seems reasonable to be applicable to both languages.

src/PerformanceSensitiveAnalyzers/CSharp/Analyzers/CallSiteImplicitAllocationAnalyzer.cs
```diff
@@ -44,61 +42,46 @@ internal sealed class CallSiteImplicitAllocationAnalyzer : AbstractAllocationAna
...
-            if (arguments.Count != methodInfo.Parameters.Length)
-            {
-                reportDiagnostic(invocationExpression.CreateDiagnostic(ParamsParameterRule, EmptyMessageArgs));
-            }
-            else
-            {
-                var lastIndex = arguments.Count - 1;
-                var lastArgumentTypeInfo = semanticModel.GetTypeInfo(arguments[lastIndex].Expression, cancellationToken);
-                if (lastArgumentTypeInfo.Type != null && !lastArgumentTypeInfo.Type.Equals(methodInfo.Parameters[lastIndex].Type))
+                bool isEmpty = (argument.Value as IArrayCreationOperation)?.Initializer.ElementValues.IsEmpty == true;
```
genlu, line 76: We can break out of the loop here if `argument.IsImplicit` is false

Youssef1313: @genlu I already break at the end of the loop.

genlu: By breaking here, we can avoid calculating `isEmpty` and (IMO) it makes the code easier to follow. Alternatively, we can rewrite the body of the `foreach` loop to make the flow clearer

```cs
foreach (var argument in invocationOperation.Arguments)
{
    if (argument.ArgumentKind == ArgumentKind.ParamArray && argument.IsImplicit)
    {
        if (!compilationHasSystemArrayEmpty || (argument.Value as IArrayCreationOperation)?.Initializer.ElementValues.IsEmpty == true)
            reportDiagnostic(node.CreateDiagnostic(ParamsParameterRule));

        break;
    }
}
```

genlu:
Looks like there's a couple test failures. @Youssef1313 Will you be able to take a look at those? Thanks

src/PerformanceSensitiveAnalyzers/UnitTests/CallSiteImplicitAllocationAnalyzerTests.cs
genlu: It should be straight forward to convert this to a language agnostic analyzer, since it's already based on IOperation. Could you please file an issue to track it? Thanks!

genlu (approved):

PR #4356, opened 2020-10-22
Update comment for clarity

Youssef1313:
The code wasn't previously working as expected. `GenerateAnalyzerRulesMissingDocumentationFile` was already true in CI builds. Now with the CI validation introduced recently, we even don't need it to be false for CI at all.

### DIFF
--- a/eng/GenerateAnalyzerNuspec.targets
+++ b/eng/GenerateAnalyzerNuspec.targets
@@ -41,8 +41,8 @@
 
     <PropertyGroup Condition="'$(GenerateAnalyzerRulesMissingDocumentationFile)' == ''">
       <!-- Generate rules missing documentation file by default for local builds and normal CI builds (to validate the existing files). -->
-      <!-- Skip it for internal signed builds that are kicked off after each PR merge builds for improved CI throughput. -->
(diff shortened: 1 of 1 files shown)


mavasani (changes requested):
This code is still valid for internal signed builds that are kicked off after each PR merge - it improves throughput + these signed build machines may or may not have internet access.

Youssef1313:
@mavasani Thanks for clarification. I updated the comment to be more clear about that to avoid any future confusion.

eng/GenerateAnalyzerNuspec.targets
```diff
@@ -40,7 +40,8 @@
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(GenerateAnalyzerRulesMissingDocumentationFile)' == ''">
-      <!-- Generate rules missing documentation file by default for local builds. Skip it for CI/signed builds for improved CI throughput. -->
+      <!-- Generate rules missing documentation file by default for local builds and normal CI builds (to validate the existing files). -->
+      <!-- Skip it for internal signed builds that are kicked off after each PR merge builds for improved CI throughput. -->
```
mavasani, line 44: Move the  comment to next line?

mavasani (approved):
Thanks!

PR #4360, opened 2020-10-23
[Port] Add CompilationEnd custom tag to diagnostics reported in compilation end actions

mavasani:
Port of https://github.com/dotnet/roslyn-analyzers/pull/4355 to release/5.0.1xx branch

PR #4363, opened 2020-10-23
Remove 'isReportedAtCompilationEnd: true' for compilation end analyzers ported to symbol start/end analyzers

mavasani:
Follow-up to #4360


sharwell (approved):

mavasani:
Thanks @sharwell!

PR #4370, opened 2020-10-26
Order APIs case-insensitive with case-sensitive fallback

sharwell:


mavasani (approved):

mavasani:
Merging as tests passed, only the CodeCov upload failed.

Issue #4021, opened 2020-08-15
[Platform Compatibility Analyzer] Indicate why the current code violates the rule


jeffhandley:
### Analyzer

**Diagnostic ID**: [CA1416](https://docs.microsoft.com/visualstudio/code-quality/ca1716) Platform Compatibility Analyzer

### Describe the improvement

When a diagnostic is raised, the message includes information about the target method, such as:

> 'API' requires 'windows' 10.1.2.3 or later
> 'API' is not supported or has been removed from 'browser'

But the diagnostic doesn't indicate what platform support was determined for the call site. We could guide users to resolve issues more easily by including in the message something indicating what support the current code has.

> 'API' requires 'windows' 10.1.2.3 or later; this code is reachable with 'windows' 10.0
> 'API' requires 'windows' 10.1.2.3 or later; this code is reachable without 'windows'
> 'API' is not supported or has been removed from 'browser'; this code is reachable with 'browser'

terrajobst:
Maybe. I'm concerned that they will not understand what this means.

buyaa-n:
The final messaging updated by @jeffhandley's [suggestion ](https://github.com/dotnet/roslyn-analyzers/pull/3917#issuecomment-674370508) 
Existing Message | Suggested Message
-- | --
'API' requires 'windows' | 'API' is supported in 'windows'
'API' requires 'windows' 10.0 or later | 'API' is supported in 'windows' 10.0 and later
'API' has been deprecated since 'windows' 0.0 | 'API' is deprecated in 'windows'
'API' has been deprecated since 'windows' 10.0 | 'API' is deprecated in 'windows' 10.0 and later
'API' is not supported or has been removed from 'windows' | 'API' is unsupported in 'windows'
'API' is not supported or has been removed since 'windows' 10.0 | 'API' is unsupported in 'windows' 10.0 and later

Now we are receiving feedbacks that `'API' is supported in 'windows'` is not that `obvious what the problem is` @stephentoub  suggested to change it to  `'API' is only supported in 'windows'`. Further when there is multiple supported OSs devs gonna see messages for each:
```csharp
  [SupportedOSPlatform("Windows")]
  [SupportedOSPlatform("Linux")]
  public void SupportedOnWindowsAndLinuxOnly() { }

  public void Caller()
  {
      SupportedOnWindowsAndLinuxOnly(); // warns: 'SupportedOnWindowsAndLinuxOnly' is supported on 'Windows'  
[...]

PR #4373, opened 2020-10-27
Improve messaging of Platform compat analyzer

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4021

Proposed messaging changes include:

List | Api attributes | Call site attribute | Existing Message | Ptoposed Message
-- | -- | -- | -- | --
Allow list | [SupportedOSPlatform("windows")] | No attribute | 'API' is supported on 'windows' | 'API' is only supported on: 'windows', but this call site is reachable on all platforms.
Allow list | [SupportedOSPlatform("windows")] | [UnsupportedOSPlatform("windows")] | 'API' is supported on 'windows' | 'API' is only supported on: 'windows', but this call site is is unreachable on 'windows'.
Allow list | [SupportedOSPlatform("windows")]  [SupportedOSPlatform("ios14.0")] | [SupportedOSPlatform("android")] [SupportedOSPlatform("linux")] | 2 warnings for each support  | 'API' is only supported on: 'windows', 'ios' 14.0 and later, but this call site is reachable on: 'linux', 'android'.
Allow list | [SupportedOSPlatform("windows")] [UnsupportedOSPlatform("windows10.0")] | no attribute | 'API' is unsupported on 'windows' 10.0 and later | 'API' is unsupported on: 'windows' from version 10.0 and later, but this call site is reachable on all platforms.
Allow list | [SupportedOSPlatform("windows10.0")] [UnsupportedOSPlatform("windows12.0")] | [SupportedOSPlatform("windows")] | 2 warnings for support and unsupported | 'API' is only supported on: 'windows' from version 10.0 to 12.0, but this call site is reachable on 'windows'.
Deny list | [UnsupportedOSPlatform("windows")] [SupportedOSPlatform("windows10.0")] | No Attribute | 'API' is supported on 'windows' 10.0 and later | 'API' is supported on: 'windows' from version 10.0 and later, but this call site is reachable on all platforms.
Deny list | [UnsupportedOSPlatform("windows")] | [SupportedOSPlatform("windows")] | 'API' is unsupported on 'windows' | 'API' is unsupported on: 'windows', but this call site is reachable on 'windows'.
Deny list | [UnsupportedOSPlatform("windows")] [SupportedOSPlatform("windows10.0")] [UnsupportedOSPlatform("windows12.0")] |  No attributes | 2 warnings for unsupported and supported | 'API' is supported on: 'windows' from version 10.0 to 12.0, but this call site is reachable on all platforms.

cc @jeffhandley @terrajobst @stephentoub

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -6,7 +6,9 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
@@ -34,12 +36,11 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         internal const string RuleId = "CA1416";
         private static readonly ImmutableArray<string> s_osPlatformAttributes = ImmutableArray.Create(SupportedOSPlatformAttribute, UnsupportedOSPlatformAttribute);
 
-        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableSupportedOsMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckSupportedOsMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableSupportedOsVersionMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckSupportedOsVersionMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableUnsupportedOsMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckUnsupportedOsMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableUnsupportedOsVersionMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckUnsupportedOsVersionMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableOnlySupportedMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityOnlySupportedMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableSupportedMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilitySupportedMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableUnsupportedMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityUnsupportedMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
 
         // We are adding the new attributes into older versions of .Net 5.0, so there could be multiple referenced assemblies each with their own
         // version of internal attribute type which will cause ambiguity, to avoid that we are comparing the attributes by their name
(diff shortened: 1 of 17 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1473,13 +1470,34 @@
...
+    <value>'{0}' from version {1} and later, </value>
+  </data>
+  <data name="PlatformFromVersionToVersionSuffix" xml:space="preserve">
+    <value>'{0}' from version {1} to {2}, </value>
+  </data>
+  <data name="PlatformVersionAndLaterSuffix" xml:space="preserve">
+    <value>'{0}' {1} and later, </value>
+  </data>
+  <data name="PlatformVersionAndBeforeSuffix" xml:space="preserve">
+    <value>'{0}' {1} and before, </value>
```
buyaa-n, line 1501: Was need to construct messages from several parts, might not good for localization, for using predefined strings for all possible combinations will need to add many more rules ...

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -86,8 +86,8 @@ void Api()
     }
 }" + MockAttributesCsSource + MockOperatingSystemApiSource;
 
-            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.UnsupportedOsRule)
-                .WithLocation(15, 9).WithArguments("Test.Api()", "windows"));
+            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule)
+                .WithLocation(15, 9).WithArguments("Test.Api()", "'windows' from version 10.0.19041 and later, but this call site is reachable on all platforms."));
```
buyaa-n, line 90: This arguments check will be removed if we go with this solution as its not good for localization

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Yes, I have been told multiple times by LOC team that combining different resource strings to form the final string is _never_ a good solution. I don't know of a better alternative though.

jeffhandley: The guidance I've collected over the years for this scenario is:

- Never concatenate strings together; always have a single message since sentence structures are not consistent across languages
- Consider using bullets instead of lists in a sentence
- Consider producing multiple resource strings to represent different scenarios and programmatically choose the correct resource string; such as:
    - When there is a single item in the list
    - When there are two items in the list
    - When there are three or more items in the list

I don't think it's going to surface here, but a related tip worth sharing is that we can never assume quantities are as simple as having a singular and plural form.

- Some languages (e.g. Russian) have different words not just for singular and plural, but for 0, 1, 2, 3, 4, and 5 of something
- Therefore you cannot assume messages like "{0} platform" or "{1} platforms" will suffice
- Instead, favor using a label and data combo such as "Number of platforms: {0}"
- But if natural language is preferred, then utilize specific resource strings to capture the most common quantities (One, Two, Three, Four, Five) and a label/data combo message as the catch-all

buyaa-n: Ok, then a separate descriptor for each message is preferred, it would be great to get dom feedback for the messages produced. I have added them as a comment to the test cases for review

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -1378,16 +1361,16 @@ class Caller
...
         public static void TestWithBrowserUnsupported()
         {
-            [|Target.SupportedOnWindows()|];
-            [|Target.SupportedOnBrowser()|];
-            [|Target.SupportedOnWindowsAndBrowser()|]; // two diagnostics expected
-            [|Target.SupportedOnWindowsAndUnsupportedOnBrowser()|];
+            [|Target.SupportedOnWindows()|];  // 'Target.SupportedOnWindows()' is only supported on: 'windows', but this call site is reachable on all platforms.
+            [|Target.SupportedOnBrowser()|];  // 'Target.SupportedOnBrowser()' is only supported on: 'browser', but this call site is unreachable on 'browser'.
+            [|Target.SupportedOnWindowsAndBrowser()|]; // 'Target.SupportedOnWindowsAndBrowser()' is only supported on: 'browser', 'windows', but this call site is unreachable on 'browser'.
+            [|Target.SupportedOnWindowsAndUnsupportedOnBrowser()|]; // 'Target.SupportedOnWindowsAndUnsupportedOnBrowser()' is only supported on: 'windows', but this call site is reachable on all platforms.
```
buyaa-n, line 1367: Produced messages added as a comment for better review

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
jeffhandley: I'm not sure if you've already made changes here based on the feedback above, but to expand on my notes above:

- Do not use one resource string for what the API is annotated with and another resource string for what the call site supports; instead, you need unique resource strings for each combination such as `ApiSupportedOnPlatformsCallSupportedOnAllPlatforms`
- There should be zero string concatenation to build a message (no StringBuilders)
- You can however use nested resources where you get one parameterized message and use that as a value inside another parameterized message -- such as having a message that represents a list of platforms, getting one value that is the list of supported platforms and another value that is the list of reachable platforms, and using those as placeholders inside the full message
- Assuming you use `String.Join` to create the string that represents the list of platforms, the item separator string also needs to be localizable.  In English, it would be `", "`, but there will be some languages that use a different separator string.

I also recommend structuring the overall message into 2 sentences in English (but still as one message):

`'{API}' is supported on: {supportedPlatformsList}. This call site is reachable on: {reachablePlatformsList}.`

This approach will avoid trouble with commas being used both in the platforms lists and within the sentence structure itself; and different languages would then be able to deal with this in targeted ways.

jeffhandley (commented):
Overall, I'd like to see the platform/version data getting passed around as data and to then only have 1 place responsible for converting the data into localizable strings--right now, the construction of the messages is spread around into multiple places (as a StringBuilder).  I also left more comments about how to construct the messages in a more localizable way.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: > Assuming you use String.Join to create the string that represents the list of platforms, the item separator string also needs to be localizable. In English, it would be ", ", but there will be some languages that use a different separator string.

Interesting, thanks 

> '{API}' is supported on: {supportedPlatformsList}. This call site is reachable on: {reachablePlatformsList}.

The {supportedPlatformsList} and {reachablePlatformsList} still would need to use `'{platform}' {version} and before`, in case any of platforms versioned

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -822,7 +946,7 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
                         {
                             // Call site has no attributes for this platform, check if MsBuild list has it,
                             // then if call site has deny list, it should support its later support
-                            if (msBuildPlatforms.Contains(platformName) &&
+                            if (msBuildPlatforms.IndexOf(platformName, 0, StringComparer.OrdinalIgnoreCase) != -1 &&
```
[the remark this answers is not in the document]
buyaa-n: These are platform names from MSBuild plus custom platforms added using `<SupportedPlatform Include="browser"/>`, technically it could match substring, but i don't recall any platform could be sub string of others

buyaa-n:
I need to base it from 5.0.x.x so closing this PR, feedbacks will be applied on the new branch with new PR

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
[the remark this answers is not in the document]
jeffhandley: It would be safer to guard against the possibility of false-positive substring matches; it's possible that years from now, such a scenario will occur, and this code would be hard to track down.

PR #4374, opened 2020-10-27
Support foreach over parameters in DoNotCopyValue

sharwell:
Builds on #4368

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
@@ -547,38 +547,45 @@ namespace Roslyn.Diagnostics.Analyzers
                     CheckLocalSymbolInUnsupportedContext(operation, local);
                 }
 
-                var instance = operation.Collection;
+                // 'foreach' operations have an identity conversion for the collection property, and then invoke the
+                // GetEnumerator method.
+                var instance = operation.Collection as IConversionOperation;
                 var instance2 = (operation.Collection as IConversionOperation)?.Operand;
 
-                switch (Acquire(operation.Collection))
+                if (instance2 is null)
                 {
-                    case RefKind.Ref:
-                        // No special requirements
-                        break;
+                    // Didn't match the known pattern
+                    instance = null;
+                }
+                else if (instance?.Conversion is not { IsIdentity: true, MethodSymbol: null })
+                {
+                    // Not a supported conversion
+                    instance = null;
+                    instance2 = null;
+                }
+                else
+                {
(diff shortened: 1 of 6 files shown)


src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
```diff
@@ -9,6 +9,8 @@
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
```
mavasani, line 13: Not recommended in the shared layer. Please move to the C# layer.

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
```diff
@@ -547,14 +549,36 @@ public override void VisitForEachLoop(IForEachLoopOperation operation)
...
+
+                switch (Acquire(operation.Collection))
                 {
-                    instance = null;
-                    instance2 = null;
+                    case RefKind.Ref:
+                        // No special requirements
+                        break;
+
+                    case RefKind.RefReadOnly when operation.Syntax is CommonForEachStatementSyntax syntax && operation.SemanticModel.GetForEachStatementInfo(syntax).GetEnumeratorMethod is { IsReadOnly: true }:
```
mavasani, line 559: This is getting too difficult to manage - IMO this analyzer is not sustainable unless rewritten as a flow-based analyzer. I'll let you decide how to take this forward.

mavasani (approved):
At this point, I am not sure my reviews are valuable for this analyzer :-) - the analyzer feels way too complex and syntax/language construct dependent to me.

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
sharwell: > IMO this analyzer is not sustainable unless rewritten as a flow-based analyzer.

Flow based analysis does not help here. Every language construct needs to be allow-listed as it is encountered.

src/Roslyn.Diagnostics.Analyzers/Core/DoNotCopyValue.cs
sharwell: Is there a language independent way to get the `GetEnumerator` method used by the `IForEachLoopOperation`?

sharwell: ➡️ Moved language-specific functionality to derived types

PR #4377, opened 2020-10-28
Implement TemporaryArrayAsRefAnalyzer

sharwell:
This analyzer enforces correct use of otherwise-unsafe method `TemporaryArrayExtensions.AsRef`.


src/Roslyn.Diagnostics.Analyzers/Core/TemporaryArrayAsRefAnalyzer.cs
```diff
@@ -0,0 +1,74 @@
...
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterCompilationStartAction(context =>
+            {
+                var temporaryArrayExtensions = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisSharedCollectionsTemporaryArrayExtensions);
+                if (temporaryArrayExtensions is null)
+                    return;
+
+                var temporaryArrayAsRef = (IMethodSymbol)temporaryArrayExtensions.GetMembers("AsRef").SingleOrDefault();
```
CyrusNajmabadi, line 44: this throws if there are two members with that name.  is that ok?

src/Roslyn.Diagnostics.Analyzers/Core/xlf/RoslynDiagnosticsAnalyzersResources.de.xlf
```diff
@@ -272,6 +272,21 @@
         <target state="translated">Zusätzliche Leerzeilen entfernen</target>
         <note />
       </trans-unit>
+      <trans-unit id="TemporaryArrayAsRefDescription">
+        <source>Instance of TemporaryArray&lt;T&gt;.AsRef() must be a using variable.</source>
```
CyrusNajmabadi, line 276:
```suggestion
        <source>Instance of TemporaryArray&lt;T&gt;.AsRef() must be a 'using' variable.</source>
```

CyrusNajmabadi (approved):

src/Roslyn.Diagnostics.Analyzers/Core/TemporaryArrayAsRefAnalyzer.cs
sharwell: Yes, it would produce a build diagnostic

src/Roslyn.Diagnostics.Analyzers/Core/xlf/RoslynDiagnosticsAnalyzersResources.de.xlf
sharwell: ➡️ Updated

src/Roslyn.Diagnostics.Analyzers/Core/TemporaryArrayAsRefAnalyzer.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        private static void AnalyzeInvocation(OperationAnalysisContext context, IMethodSymbol temporaryArrayAsRef)
+        {
+            var invocation = (IInvocationOperation)context.Operation;
+            var targetMethod = invocation.TargetMethod.ReducedFrom ?? invocation.TargetMethod;
+            if (!Equals(targetMethod.OriginalDefinition, temporaryArrayAsRef))
+                return;
+
+            var instance = invocation.Instance ?? invocation.Arguments.FirstOrDefault()?.Value;
```
mavasani, line 59: Use the extension method `IInvocationOperation.GetReceiverType`? If that doesn't work, seems to be a good candidate to add a parallel extension method that matches your scenario.

sharwell: I'm not sure that helps here (don't need the type).

PR #4395, opened 2020-10-31
AvoidUninstantiatedInternalClasses: Don't report on static classes

### LINKED ISSUE #4389: CA1812 incorrect message about static class not being instantiated
This is a regression - the problem didn't happen in v3.3.0.

### Analyzer

**Diagnostic ID**: [CA1812](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1812): `Avoid uninstantiated internal classes`

### Analyzer source

**NuGet Package**: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

**Version**: v3.3.1 (Latest)

### Steps To Reproduce

Create an internal static class without any public or internal methods.

```csharp
    internal static class Foo
    {
    }
```

### Actual behavior

Warning reported:

> CA1812: Foo is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it static (Shared in Visual Basic)

### Additional context

This message doesn't make any sense - the class is already marked as `static`.

Youssef1313:
Fixes #4389

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/AvoidUninstantiatedInternalClasses.cs
@@ -346,6 +346,11 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
                 return true;
             }
 
+            if (type.IsStaticHolderType())
+            {
+                return true;
+            }
+
             return false;
         }
         public static bool IsMefExported(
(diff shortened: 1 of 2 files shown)


Youssef1313:
@Evangelink Let me know if this is the right fix and I'll update the tests.

Note: I think a non-static class containing only static methods will now report diagnostic while it didn't use to in the past. Is this behavior desired? To me, this is reasonable.

mavasani (approved):
LGTM

PR #4390, opened 2020-10-29
Warn for platform specific type used as generic type parameter

buyaa-n:
Warn when platform-specific type used as generic type parameter, useful when a type created using reflections

Fixes https://github.com/dotnet/roslyn-analyzers/issues/4362

PR #4403, opened 2020-11-02
Flag when platform specific type used as generic type parameter

### LINKED ISSUE #4362: Platform compatibility analyzer not warning for platform specific type parameter
Platform compatibility analyzer not warning for platform-specific type parameter, seems we could warn for any type parameter usage of platform-specific API. Related to [PR review comment](https://github.com/dotnet/runtime/pull/43363#discussion_r510934435) :

> To cover both could we warn for any type of parameter usages? If that could cause any false positives?

My understanding of the intention here:

1. The whole class/type is marked as `[UnsupportedOSPlatform("browser")]`, meaning that "this class is not going to work on that platform".
2. The method call `SomeGenericMethod<AboveType>()` should warn about using this type. It is not supported on this platform.

I guess I don't see how a false positive could occur here. You are using a type that is not supported on the platform. It seems like it should warn.

_Originally posted by @eerhardt in https://github.com/dotnet/runtime/pull/43363#discussion_r511064337_

buyaa-n:
Warn when platform-specific type used as generic type parameter, useful when a type created using reflections

Fixes #4362

Cherry-picked from https://github.com/dotnet/roslyn-analyzers/pull/4390 to target `release\5.0.2xx` branch


buyaa-n:
@jeffhandley confirmed that we want it for 5.0.2xx and we have a separate issue for tracking inheritance related scenarios, I think this PR good to go now @mavasani please take another look

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -583,6 +600,36 @@ static void RemoveOtherSupportsOnDifferentPlatforms(SmallDictionary<string, Plat
...
+                [NotNullWhen(true)] out ISymbol? symbol, ImmutableArray<ITypeSymbol> typeArguments)
+            {
+                foreach (var typeArgument in typeArguments)
+                {
+                    if (platformSpecificMembers.TryGetValue(typeArgument, out var foundAttributes) &&
+                        foundAttributes != null &&
+                        foundAttributes.Any())
+                    {
+                        symbol = typeArgument;
+                        return true;
```
mavasani, line 613: Hmm I am not clear on this - are we just considering the first type argument symbol that has any platform attributes? What if there are multiple type arguments with platform attributes, but the one with violation type argument is not the first one, say `if (!IsWindowsOnly()) { GenericMethod<BrowserOnlyType, WindowsOnlyType>(); }`. I believe this should still flag the method invocation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -549,15 +549,32 @@ static void RemoveOtherSupportsOnDifferentPlatforms(SmallDictionary<string, Plat
...
                 if (accessor != null)
                 {
                     symbol = accessor;
                 }
             }
+            else if (symbol is IMethodSymbol method)
+            {
+                if (method.IsGenericMethod && TryGetSymbol(platformSpecificMembers, out var foundSymbol, method.TypeArguments))
+                {
+                    symbol = foundSymbol;
```
mavasani, line 564: For this case, I feel we should analyze both the original method symbol and also the found type symbol (or in fact all type symbols that have platform attributes). The current implementation means that we will not validate the platform attributes on the method itself if we find a type argument with platform attributes.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: > are we just considering the first type argument symbol that has any platform attributes? What if there are multiple type arguments with platform attributes, but the one with violation type argument is not the first one

Yes, that is right, i will update to get the correct symbol with your example, thanks

> I believe this should still flag the method invocation.

This is flagging the method invocation, because the diagnostic attributes tied to the operations only, and some operation's symbol is not one which has platform-specific attributes. In some cases, one of the child symbols of the current symbol might be the one we want to report the diagnostic. But yes here it might give the wrong symbol when there were many platform-specific type parameters.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -693,10 +740,50 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
             }
             else
             {
-                CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, symbol);
+                if (!CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, symbol) && symbol is IMethodSymbol method)
```
buyaa-n, line 743: Here we are checking the operation (method) first, if it has attributed, else we will check the method's type parameters

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
mavasani: > This is flagging the method invocation, because the diagnostic attributes tied to the operations only

I meant will this flag `if (IsBrowserOnly()) { WindowsOnlyMethod<BrowserOnlyType>(); }`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
mavasani: Ah got you, thanks. Can we add unit test that I mentioned in https://github.com/dotnet/roslyn-analyzers/pull/4403#discussion_r517676488?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: This section is for ReportDiagnostics(...), i.e. the analysis already done and reporting diagnostic, here just trying load the correct symbol for the message, the analysis section done here https://github.com/dotnet/roslyn-analyzers/pull/4403/files#r517676304

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: Sure i will add a test for that

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: > I meant will this flag if (IsBrowserOnly()) { WindowsOnlyMethod<BrowserOnlyType>(); }?

Yes it might flag the wrong symbol, `BrowserOnlyType` instead `WindowsOnlyMethod`, for your other example: `if (IsBrowserOnly()) { GenericMethod<BrowserOnlyType, WindowsOnlyType>(); }` it might also have incorrect message `'BrowserOnlyType' is supported on 'windows'` instead `'WindowsOnlyType' is supported on 'windows'`, good catch, i will add test for both case

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -706,20 +702,56 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+                    {
+                        if (TryCopyAttributesNotSuppressedByMsBuild(operationAttributes, msBuildPlatforms, out var copiedAttributes))
+                        {
+                            platformSpecificOperations.TryAdd(new KeyValuePair<IOperation, ISymbol>(operation, symbol), copiedAttributes);
+                        }
+                    }
+                }
+            }
+
+            static void CheckImmedaiteAttributes(IOperation operation, OperationAnalysisContext context, PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>, SmallDictionary<string, PlatformAttributes>> platformSpecificOperations,
```
mavasani, line 718:
```suggestion
            static void CheckImmediateAttributes(IOperation operation, OperationAnalysisContext context, PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>, SmallDictionary<string, PlatformAttributes>> platformSpecificOperations,
```

mavasani: It seems like the new added method is the exact clone of `CheckOperationAttributes`. Instead of duplicate implementations, can we just past in a new flag to `CheckOperationAttributes`, say `bool onlyImmediateSymbol`, which is then passed down to `TryGetOrCreatePlatformAttributes` to perform the appropriate logic?

buyaa-n: I have actually moved the duplicate section into a separate method

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: When an operation symbol has associated symbols also need to be analyzed (like property or event accessors, generic type parameters) with the previous solution when a parent has diagnostic the child symbols were having repeated diagnostic and it was difficult to get the real symbol (parent or child, which child etc) having the diagnostic from operations map, refactored it to diagnose only immediate attributes for child symbols and keep operation and symbol both in the map

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -658,81 +625,154 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
                 }
             }
-        }
 
-        private static bool TryCopyAttributesNotSuppressedByMsBuild(SmallDictionary<string, PlatformAttributes> operationAttributes,
-            ImmutableArray<string> msBuildPlatforms, out SmallDictionary<string, PlatformAttributes> copiedAttributes)
-        {
-            copiedAttributes = new SmallDictionary<string, PlatformAttributes>(StringComparer.OrdinalIgnoreCase);
-            foreach (var (platformName, attributes) in operationAttributes)
+            static bool TryGetOrCreateImmediatePlatformAttributes(ISymbol symbol, ConcurrentDictionary<ISymbol, SmallDictionary<string, PlatformAttributes>?> platformSpecificMembers,
```
mavasani, line 749: Why do we need a separate method `TryGetOrCreateImmediatePlatformAttributes`? Can we instead just add a new bool, say `immediateOnly` to `TryGetOrCreatePlatformAttributes` such that [this part](https://github.com/dotnet/roslyn-analyzers/blob/03a1c1666000aad58d660db44806710f7e56326c/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L968-L986) of that method is executed only when this bool is false?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -658,81 +625,154 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+                                {
+                                    CheckImmediateAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, tArgument);
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            static void CheckImmediateAttributes(IOperation operation, OperationAnalysisContext context, PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>,
```
mavasani, line 680: Feels like you don't need this new method and can revert back to the original `CheckOperationAttributes` once you add a new boolean flag `immediateOnly` that is passed down to this method and also to `TryGetOrCreatePlatformAttributes`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -658,81 +625,154 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+                SmallDictionary<string, PlatformAttributes>> platformSpecificOperations, ConcurrentDictionary<ISymbol, SmallDictionary<string, PlatformAttributes>?> platformSpecificMembers,
+                ImmutableArray<string> msBuildPlatforms, ISymbol symbol)
+            {
+                if (TryGetOrCreateImmediatePlatformAttributes(symbol, platformSpecificMembers, out var operationAttributes))
+                {
+                    CheckWithCallSiteAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, symbol, operationAttributes);
+                }
+            }
+
+            static void AnalyzePropertyAccessors(IOperation operation, OperationAnalysisContext context, PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>,
```
mavasani, line 690: Feels like you can also revert these new local functions to reduce the noise in this PR - instead just change the original code to pass in `immediateOnly: true` to `CheckOperationAttributes` for accessors.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: We can do that, for me it was more descriptive, but yes it is causing more noise in the PR, i will revert some of the changes

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -658,23 +625,25 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
         {
             var symbol = GetOperationSymbol(operation);
 
-            if (symbol == null)
+            if (symbol == null || symbol is ITypeSymbol type && type.SpecialType != SpecialType.None)
             {
                 return;
             }
 
+            CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, symbol, true);
```
mavasani, line 638: Nit: consider using named argument for literals
```suggestion
            CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, symbol, checkParents: true);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -658,23 +625,25 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+
             if (symbol is IPropertySymbol property)
             {
                 foreach (var accessor in GetPropertyAccessors(property, operation))
                 {
                     if (accessor != null)
                     {
-                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor);
+                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor, false);
```
mavasani, line 646:
```suggestion
                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor, checkParents: false);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -684,36 +653,71 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
 
                 if (accessor != null)
                 {
-                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor);
+                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor, false);
```
mavasani, line 656:
```suggestion
                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor, checkParents: false);
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -131,6 +131,85 @@ class Target
...
+using System;
+using System.Runtime.Versioning;
+
+[SupportedOSPlatform(""windows"")]
+class WindowsOnlyType {}
+
+class GenericClass<T>
+{
+    public static void M<V>() { }
+    public static void M2() {}
```
mavasani, line 197: Can you add a non-generic field (say `Field`) and property (say `Property`) to this type and then verify that following are flagged: `[|GenericClass<<WindowsOnlyType>.Field()|];` and `[|GenericClass<<WindowsOnlyType>.Property()|];` inside `Test.M1()` in this test?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -684,36 +653,71 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
                 }
-                else
+            }
+            else if (symbol is IMethodSymbol method)
+            {
+                AnalyzeTypeParameters(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, method);
+            }
+
+            static void CheckTypeArguments(ImmutableArray<ITypeSymbol> namedTypes, IOperation operation, OperationAnalysisContext context,
```
mavasani, line 664: I think the parameter should be renamed as type arguments are not guaranteed to be named type, then can be any type symbol:
```suggestion
            static void CheckTypeArguments(ImmutableArray<ITypeSymbol> typeArguments, IOperation operation, OperationAnalysisContext context,
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -684,36 +653,71 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+                PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>, SmallDictionary<string, PlatformAttributes>> platformSpecificOperations,
+                ConcurrentDictionary<ISymbol, SmallDictionary<string, PlatformAttributes>?> platformSpecificMembers, ImmutableArray<string> msBuildPlatforms)
+            {
+                foreach (var typeArgument in namedTypes)
                 {
-                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, iEvent);
+                    if (typeArgument.SpecialType == SpecialType.None)
+                    {
+                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, typeArgument, false);
```
mavasani, line 672:
```suggestion
                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, typeArgument, checkParents: false);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -684,36 +653,71 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+
+                        if (typeArgument is INamedTypeSymbol nType && nType.IsGenericType)
+                        {
+                            foreach (var tArgument in nType.TypeArguments)
+                            {
+                                if (tArgument.SpecialType == SpecialType.None)
+                                {
+                                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, tArgument, false);
```
mavasani, line 680:
```suggestion
                                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, tArgument, checkParents: false);
```

mavasani (commented):
Overall looking good - want to make sure that we handle field/property/event access off a generic type with disallowed type argument. Looks good to sign off once that is addressed. Thanks!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
buyaa-n: Sure

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -59,6 +59,79 @@ void Api()
...
+        {
+            [|WindowsOnlyMethod<BrowserOnlyType>()|];  // should flag for WindowsOnlyMethod method and BrowserOnlyType parameter
+            [|GenericMethod<WindowsOnlyType, BrowserOnlyType>()|]; // should flag for WindowsOnlyType and BrowserOnlyType parameters
+        }
+    }
+}
+" + MockAttributesCsSource + MockOperatingSystemApiSource;
+            await VerifyAnalyzerAsyncCs(csSource, s_msBuildPlatforms,
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(24, 13).WithArguments("Test.WindowsOnlyMethod<BrowserOnlyType>()", "windows"),
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.SupportedOsRule).WithLocation(25, 13).WithArguments("WindowsOnlyType", "windows"));
```
jeffhandley, line 132: Should there be an expected diagnostic here for `BrowserOnlyType` per the comment above?

```csharp
// should flag for WindowsOnlyType and BrowserOnlyType parameter
```

buyaa-n: Yes, that is matched with the markup syntax, in the else section of the condition 2 diagnostics expected for each invocation, those extra diagnostics are matched here

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -684,36 +653,61 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+                foreach (var typeArgument in typeArguments)
                 {
-                    if (TryGetOrCreatePlatformAttributes(context.ContainingSymbol, platformSpecificMembers, out var callSiteAttributes))
+                    if (typeArgument.SpecialType == SpecialType.None)
+                    {
+                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, typeArgument, checkParents: false);
+
+                        if (typeArgument is INamedTypeSymbol nType && nType.IsGenericType)
+                        {
+                            foreach (var tArgument in nType.TypeArguments)
```
mavasani, line 681: Instead of this for loop, should we just make a recursive call to `CheckTypeArguments` here? This will ensure things like `GenericType<SomeType<WindowsOnlyType>>.NonGenericMethod();` get flagged where the violating type is a nested type argument. I think to avoid infinite recursion, we probably need to maintain a worklist/set of types currently being processed by `CheckTypeArguments` so we don't repeatedly call it for the same type when we have broken code, say `GenericType<SomeType<GenericType>>.NonGenericMethod();`?

buyaa-n: Sure, this PR change is causing lots of new warnings in runtime, i am digging into that since yesterday, will apply your suggestion after that

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -684,36 +653,66 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
...
+                SmallDictionary<string, PlatformAttributes>> platformSpecificOperations, ConcurrentDictionary<ISymbol, SmallDictionary<string, PlatformAttributes>?> platformSpecificMembers,
+                ImmutableArray<string> msBuildPlatforms, ISymbol symbol, bool checkParents)
+            {
+                if (TryGetOrCreatePlatformAttributes(symbol, checkParents, platformSpecificMembers, out var operationAttributes))
                 {
-                    if (TryGetOrCreatePlatformAttributes(context.ContainingSymbol, platformSpecificMembers, out var callSiteAttributes))
+                    var containingSymbol = context.ContainingSymbol;
+                    if (containingSymbol is IMethodSymbol method && method.IsAccessorMethod())
+                    {
+                        containingSymbol = method.AssociatedSymbol;
```
buyaa-n, line 701: As we are not adding parent attributes to the accessor methods so that it will not report a duplicate warning when the containing symbol had any attribute (for example property has no accessor level attribute but the entire property has attribute). But when the property getter or setter any statements inside it became the containing symbol for those statements. When accessors used as call site, as their attributes are not including the parent attributes call site attributes are incomplete and causing warnings. So here when accessor is the containing symbol change it to the AssociatedSymbol which is the real Property or Event which has complete attributes for call site

mavasani (approved):

buyaa-n:
Retargeted to master, as there is no any other change merging

PR #4417, opened 2020-11-08
Update CA1001 to support excluded_symbol_names option

### LINKED ISSUE #4385: CA1004: allow to disable the rule for each field
### Analyzer

**Diagnostic ID**: [CA1716](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1716)

### Describe the improvement
The rule CA1004 reports classes that contain fields are that are IDisposable and are not properly disposed.
In case of suppression of the rule, it gets suppressed for the full class.
It would be nice if:
- it could keep the current behavior, when applied to a whole class. I.e. suppressing at class level will ignore any field in the class itself;
- but also introduce the possibility to suppress it at field level, so I can ignore one field, but still keep the rule valid for the class, so if later another developer will introduce a new IDisposable field, it will be correctly picked up by the rule.

The improvement would be the possibility to suppress the rule at field level.

Evangelink:
Fix #4385

### DIFF
--- a/docs/Analyzer Configuration.md
+++ b/docs/Analyzer Configuration.md
@@ -261,6 +261,7 @@ Example: `dotnet_code_quality.try_determine_additional_string_formatting_methods
 Option Name: `excluded_symbol_names`
 
 Configurable Rules:
+[CA1001](https://docs.microsoft.com/visualstudio/code-quality/ca1001),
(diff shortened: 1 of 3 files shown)


mavasani (approved):
Thank you @Evangelink!

PR #4438, opened 2020-11-19
Improve messaging for Platform Compat Analyzer

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4021
This PR only has messaging updates, no any functional change or bugfix
The messages update into the following structure:
List | Messaging structure
-- | -- 
Allow list | '{0}' is only supported on: {1}. This call site is reachable on all platforms. 
Allow list | '{0}' is only supported on: {1}. This call site is unreachable on: {2}.
Allow list | '{0}' is only supported on: {1}. This call site is reachable on: {2}. 
Deny list | '{0}' is unsupported on: {1}, This call site is reachable on all platforms.
Deny list | '{0}' is unsupported on: {1}, This call site is reachable on: {2}.
Deny list | '{0}' is supported on: {1}, This call site is reachable on all platforms.
Deny list | '{0}' is supported on: {1}, This call site is reachable on: {2}.

Examples: 

- `'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser'. This call site is reachable on all platforms.`
- `'SupportedOnWindowsUnsupportedOnWindows2004()' is only supported on: 'windows' 10.0.2004 and before. This call site is reachable on: 'Windows'.`
- `SupportedOnWindows()' is only supported on: 'windows'. This call site is reachable on: 'windows', 'browser'.`
- `'SupportedOnWindowsAndBrowser()' is only supported on: 'browser', 'windows'. This call site is unreachable on: 'browser'.`
- `'UnsupportedOnWindows()' is unsupported on: 'windows'. This call site is reachable on all platforms.`
- `'UnsupportedOnBrowser()' is unsupported on: 'browser'. This call site is reachable on: 'windows', 'browser'.`
- `'UnsupportedOnWindowsSupportedOn1903UnsupportedOn2004()' is supported on: 'windows' from version 10.0.1903 to 10.0.2004. This call site is reachable on all platforms.`
- `'UnsupportedOnWindowsSupportedOn1903()' is supported on: 'windows' 10.0.1903 and later. This call site is reachable on: 'windows' 10.0.2000 and before`

cc @jeffhandley @stephentoub  @terrajobst

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Linq;
 using System.Threading;
 using Analyzer.Utilities;
@@ -34,12 +35,15 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         internal const string RuleId = "CA1416";
         private static readonly ImmutableArray<string> s_osPlatformAttributes = ImmutableArray.Create(SupportedOSPlatformAttribute, UnsupportedOSPlatformAttribute);
 
-        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableSupportedOsMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckSupportedOsMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableSupportedOsVersionMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckSupportedOsVersionMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableUnsupportedOsMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckUnsupportedOsMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableUnsupportedOsVersionMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckUnsupportedOsVersionMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
-        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityCheckDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableOnlySupportedCsAllPlatform = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityOnlySupportedCsAllPlatformMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableOnlySupporteCsReachable = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityOnlySupportedCsReachableMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableOnlySupporteCsUnreachable = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityOnlySupportedCsUnreachableMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizablSupporteCsAllPlatform = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilitySupportedCsAllPlatformMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizablSupporteCsReachable = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilitySupportedCsReachableMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableUnsupportedCsAllPlatform = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityUnsupportedCsAllPlatformMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableUnsupportedCsReachable = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityUnsupportedCsReachableMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
 
         // We are adding the new attributes into older versions of .Net 5.0, so there could be multiple referenced assemblies each with their own 
         // version of internal attribute type which will cause ambiguity, to avoid that we are comparing the attributes by their name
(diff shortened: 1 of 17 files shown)


mavasani:
> `'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser'. This call site is reachable on all platforms.`

It feels like the message states two statements, but it is not clear if the problem is with the first part or second part and what should be done to resolve the issue. Some other message suggestions:

- `'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser', but this call site is reachable on all platforms`
- This call site is reachable on all platforms, but `'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser'

terrajobst:
Can we change the output to emit the platforms on separate lines, rather than a comma list in the in message?

Here is what we do when the target framework OS version doesn't exist:

image

When the user expands the node:

image

On the command line it looks as follows:

> error NETSDK1140: 10.0.19.0 is not a valid TargetPlatformVersion for Windows. Valid versions include:
> error NETSDK1140: 10.0.19041.0
> error NETSDK1140: 10.0.18362.0
> error NETSDK1140: 10.0.17763.0
> error NETSDK1140: 8.0
> error NETSDK1140: 7.0

Thoughts?

buyaa-n:
> * `'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser', but this call site is reachable on all platforms`

@mavasani FYI this was the message we had in my previous PR and changed it by the review comment https://github.com/dotnet/roslyn-analyzers/pull/4373#discussion_r518489206

> * This call site is reachable on all platforms, but 'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser'

This message looks good to me, I can update it if others agree with changing like this

buyaa-n:
> Can we change the output to emit the platforms on separate lines, rather than a comma list in the in message?

@terrajobst  I am wondering how that message would look like in VS and if VS support that, looks it should look like something like this: 

'API()' is only supported on: 'windows'
'browser'
'Linux'
This call site is reachable on 'windows'
'ios'
'Linux'

mavasani:
@terrajobst @buyaa-n I don't know if the C# and VB compiler can handle analyzer messages with newlines - https://github.com/dotnet/roslyn/issues/42616 indicates this will lead to a compiler crash, but you may have to verify.

mavasani:
> This message looks good to me, I can update it if others agree with changing like this

Thanks @buyaa-n, it was just a personal opinion - if your proposed messages seem fine to others, I don't have strong push back.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -413,45 +400,45 @@ public static void TestWithGuardMethods()
...
 
             if (OperatingSystemHelper.IsWindows())
             {
-                [|Target.UnsupportedInWindows()|]; // row 31 expected diagnostic - windows unsupported
+                [|Target.UnsupportedInWindows()|]; // 'Target.UnsupportedInWindows()' is unsupported on: 'windows'. This call site is reachable on: 'Windows'.
                 [|Target.UnsupportedInWindows10()|]; // expected diagnostic - windows 10 unsupported
                 Target.UnsupportedOnBrowser();
                 [|Target.UnsupportedOnWindowsAndBrowser()|]; // expected diagnostic - windows unsupported
-                [|Target.UnsupportedOnWindows10AndBrowser()|]; // expected diagnostic - windows 10 unsupported
+                [|Target.UnsupportedOnWindows10AndBrowser()|]; // 'Target.UnsupportedOnWindows10AndBrowser()' is unsupported on: 'windows' 10.0 and later. This call site is reachable on: 'Windows'.
```
jeffhandley, line 423: > 'Target.UnsupportedOnWindows10AndBrowser()' is unsupported on: 'windows' 10.0 and later. This call site is reachable on: 'Windows'.

In this example, it might not be clear that the call site is reachable on all versions of Windows, since that isn't called out explicitly.  And the casing of 'windows' vs. 'Windows' could lead a user to think there's a case sensitivity problem causing the error instead of the all versions aspects.

What would you think about the following?

1. Always lower-casing the platforms
2. Using a message that includes "(all versions)" when that makes a difference

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -558,38 +621,237 @@ static void RemoveOtherSupportsOnDifferentPlatforms(SmallDictionary<string, Plat
...
+                 SmallDictionary<string, PlatformAttributes> attributes, SmallDictionary<string, PlatformAttributes>? callsiteAttributes)
+            {
+                var supportedRule = GetSupportedPlatforms(attributes, out var platformNames);
+                var platforms = string.Join(MicrosoftNetCoreAnalyzersResources.CommaSeparator, platformNames);
+                var callSitePlatforms = string.Join(MicrosoftNetCoreAnalyzersResources.CommaSeparator, GetCallsitePlatforms(attributes, callsiteAttributes, out var callsite));
+                var rule = supportedRule ? SwitchSupportedRule(callsite) : SwitchRule(callsite, true);
+
+                context.ReportDiagnostic(operation.CreateDiagnostic(rule, operationName, platforms, callSitePlatforms));
+
+                static DiagnosticDescriptor SwitchSupportedRule(Callsite callste)
```
jeffhandley, line 657: Nit: Is `callste` a typo (intending `callsite`)?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -558,38 +621,237 @@ static void RemoveOtherSupportsOnDifferentPlatforms(SmallDictionary<string, Plat
...
+                            }
+                            supportedRule = false;
+                        }
+                    }
+                    return supportedRule;
                 }
+            }
 
-                if (attribute.UnsupportedFirst != null)
+            static DiagnosticDescriptor SwitchRule(Callsite callste, bool unsupported)
```
jeffhandley, line 716: Nit: Is `callste` a typo (intending `callsite`)?

jeffhandley:
I prefer the approach of 2 sentences, as to me it makes it easier to see the separation between the comma-separated list of supported platforms and the comma-separated list of reachable platforms when the message is complex. I thought it could make translation a little easier as well.

jeffhandley (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: > 1. Always lower-casing the platforms

Platforms are printed how they provided in the attribute, i think not all platform names suitable for lower casing (should be the same as how we use in TFM), further, we could address platform name casing with the other analyzer which would warn for platform name/versions correctness and attributes consistency

> 2. Using a message that includes "(all versions)" when that makes a difference

I will give it try

jeffhandley: > not all platform names suitable for lower casing

Makes sense.  I presume when we detect the code is within a guard, we're getting the platform name from the `Is<Platform>` method, just stripping off the `Is`, which means we'll always be proper cased ("Windows").  The first character is likely where we'll see the difference.

Drawing attention to the version mismatch is going to be the key then.  We should consider including in the documentation a note that platform names are case-insensitive, in case anyone gets that impression and goes to the docs to check if that's the case.

buyaa-n:
> I prefer the approach of 2 sentences, as to me it makes it easier to see the separation between the comma-separated list of supported platforms and the comma-separated list of reachable platforms when the message is complex. I thought it could make translation a little easier as well.

Completely agree, how about adding `But` in front of the 2nd sentence? Might make the 2 sentences more connected `'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser'. But this call site is reachable on all platforms.`

buyaa-n:
Tagging @gewarren @JasonCard for language and localization suggestions

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: > I presume when we detect the code is within a guard, we're getting the platform name from the Is<Platform> method, just stripping off the Is, which means we'll always be proper cased ("Windows"). The first character is likely where we'll see the difference.

Oh yeah in this case it is produced from the guard methods, that is why it is capital

> Drawing attention to the version mismatch is going to be the key then. We should consider including in the documentation a note that platform names are case-insensitive, in case anyone gets that impression and goes to the docs to check if that's the case.

Good point

mavasani:
I don't think starting a sentence with "But" is gramatically correct. I am fine with your current implementation if the doc team feels the message is clear and grammatically correct.

gewarren:
I prefer this variation, because the issue is with the call site, not the method being called, so it seems you want to put that first:

`This call site is reachable on all platforms, but <method> is only supported on: 'windows' 10.0 and later, 'browser'.`

However, what about separating the platforms with "or" instead of commas?

`This call site is reachable on all platforms, but <method> is only supported on 'windows' 10.0 and later or 'browser'.`

JasonCard:
I agree with the approach to separate into 2 sentences, which should be easier to translate. Especially with placeholders in the sentences. It may also make sense to put the call site info first, but I'm not sure if adding "But" would always make sense in this case if call site is "unreachable":
```
This call site is unreachable on: 'browser', 'Linux'. But 'SupportedOnWindowsAndBrowser()' is only supported on: 'browser', 'windows'.
```  
cc adding @cristianosuzuki77 if he had any additional input.

buyaa-n:
Thank you all for your input, sounds like we want to move the call site in front, keep the 2 sentences, do not need `But`:
List | Messaging structure
-- | -- 
Allow list | This call site is reachable on all platforms. '{Api}' is only supported on: {platforms}.
Allow list | This call site is unreachable on: {platforms}. '{Api}' is only supported on: {platforms}.
Allow list | This call site is reachable on: {platforms}. '{Api}' is only supported on: {platforms}. 
Deny list | This call site is reachable on all platforms. '{Api}' is unsupported on: {platforms}.
Deny list | This call site is reachable on: {platforms}. '{Api}' is unsupported on: {platforms}.
Deny list | This call site is reachable on all platforms. '{Api}' is supported on: {platforms}. 
Deny list | This call site is reachable on: {platforms}. '{Api}' is supported on: {platforms}.

Examples: 

- `This call site is reachable on all platforms.'SupportedOnWindows10AndBrowser()' is only supported on: 'windows' 10.0 and later, 'browser'.`
- ` This call site is reachable on: 'Windows' all versions.'SupportedOnWindowsUnsupportedOnWindows2004()' is only supported on: 'windows' 10.0.2004 and before.`
- `This call site is reachable on: 'windows', 'browser'. 'SupportedOnWindows()' is only supported on: 'windows'.`
- `This call site is unreachable on: 'browser'. 'SupportedOnWindowsAndBrowser()' is only supported on: 'browser', 'windows'.`
- `This call site is reachable on all platforms. 'UnsupportedOnWindows()' is unsupported on: 'windows'.`
- `This call site is reachable on: 'windows', 'browser'. 'UnsupportedOnBrowser()' is unsupported on: 'browser'.`
- `This call site is reachable on all platforms. 'UnsupportedOnWindowsSupportedOn1903UnsupportedOn2004()' is supported on: 'windows' from version 10.0.1903 to 10.0.2004.`
- `This call site is reachable on: 'windows' 10.0.2000 and before. 'UnsupportedOnWindowsSupportedOn1903()' is supported on: 'windows' 10.0.1903 and later.`

jeffhandley (approved):

cristianosuzuki77:
Please add the examples as a resx comment. It will help understanding the possible values of the placeholders.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -558,38 +621,261 @@ static void RemoveOtherSupportsOnDifferentPlatforms(SmallDictionary<string, Plat
...
+                    ReportUnsupportedDiagnostic(operation, context, operationName, attributes, csAttributes);
                 }
+                break;
+            }
+
+            static void ReportSupportedDiagnostic(IOperation operation, OperationBlockAnalysisContext context, string operationName,
+                 SmallDictionary<string, PlatformAttributes> attributes, SmallDictionary<string, PlatformAttributes>? callsiteAttributes)
+            {
+                var supportedRule = GetSupportedPlatforms(attributes, callsiteAttributes, out var platformNames);
+                var platforms = string.Join(MicrosoftNetCoreAnalyzersResources.CommaSeparator, platformNames);
```
mavasani, line 651: Why do we need this to be localized? I think it would be fine to just use the hard coded string `, ` here and below when joining platform names.

buyaa-n: It was added by a [comment](https://github.com/dotnet/roslyn-analyzers/pull/4373#discussion_r518489206): `Assuming you use String.Join to create the string that represents the list of platforms, the item separator string also needs to be localizable. In English, it would be ", ", but there will be some languages that use a different separator string.`

buyaa-n:
> Please add the examples as a resx comment. It will help understanding the possible values of the placeholders.

Examples added @cristianosuzuki77 @gewarren @JasonCard please take a look the localization

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
mavasani: Thanks. I think we might have few other comma separated `string.Join` operations in the repo that use the hard coded string `, ` to generate the diagnostic argument.

jeffhandley: Let's confirm with @JasonCard. Do you want the `, ` separator to be localizable? Some localization teams I've worked with in the past have and some have not. If you don't think we need it to be a resource string, we can hard-code it.

JasonCard: Comma can be different for Asian languages, so I believe it should be localizable, but I'm not familiar with this resource string and how it will be used, so @cristianosuzuki77 (loc owner for SW/product side) can double confirm.

cristianosuzuki77:
thanks for adding the examples. The comment on comma is correct, the word separator is different for example Japananese and Simplified Chinese (eg English: System reserved names, including 'CON', 'AUX', PRN', 'COM1' or 'LPT2' 
Japanese: システム予約名 ('CON'、'AUX'、'PRN'、'COM1'、'LPT2' など) 
Simplified Chinese: 系統保留名稱，包括 'CON'、'AUX'、'PRN'、'COM1' 或 'LPT2' 

Please add a real example as a resx comment, so the intent of asking the comma to be translated is clear to the translators as well.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: > Please add a real example as a resx comment, so the intent of asking the comma to be translated is clear to the translators as well.

Example added as a resx comment, how it is looks:
https://github.com/dotnet/roslyn-analyzers/blob/104dc0fc9785e7ea405607fbc6ff0970a1226b19/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx#L1489-L1491
FYI we are omitted `'or', 'and'` conjunctions for simplicity

buyaa-n:
> Please add a real example as a resx comment, so the intent of asking the comma to be translated is clear to the translators as well.

Example added as a resx comment, how it is looks:
https://github.com/dotnet/roslyn-analyzers/blob/104dc0fc9785e7ea405607fbc6ff0970a1226b19/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx#L1489-L1491
FYI we are omitted `'or', 'and'` conjunctions for simplicity

cristianosuzuki77 (approved):
These examples are good enough for translation.

buyaa-n:
Thanks you all for review/suggestions, waiting for @jmarolf's sign off for merging into `release/5.0.2xx` branch

jmarolf (approved):

PR #4440, opened 2020-11-19
Fix fast-path cache

sharwell:
For default **.editorconfig** scenarios, `SymbolNamesWithValueOption<TValue>.Empty` is used. This has an empty set of wildcard match names, but still caches symbols. Since the `Empty` value is static, this means the process keeps a cached copy of _all_ symbols for which `Contains` is called until the process runs out of memory.

This issue is also the reason why other `ConditionalWeakTable<TKey, TValue>` instances were never releasing values, since the `Compilation` keys were retaining strong roots through this path.

Fixes [AB#1242125](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1242125)

### DIFF
--- a/src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
+++ b/src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
@@ -22,9 +22,8 @@ namespace Analyzer.Utilities
         public static readonly SymbolNamesWithValueOption<TValue> Empty = new SymbolNamesWithValueOption<TValue>();
         internal static KeyValuePair<string, TValue> NoWildcardMatch => default;
 
-#pragma warning disable CA1051 // Do not declare visible instance fields
-        internal /* for testing purposes */ readonly ImmutableDictionary<string, TValue> _names;
-        internal /* for testing purposes */ readonly ImmutableDictionary<ISymbol, TValue> _symbols;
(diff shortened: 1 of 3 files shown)


src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -277,7 +275,6 @@ private bool TryGetFirstWildcardMatch(ISymbol symbol, out KeyValuePair<string, T
             if (_wildcardNamesBySymbolKind.IsEmpty)
             {
                 firstMatch = NoWildcardMatch;
-                _wildcardMatchResult.AddOrUpdate(symbol, firstMatch, (s, match) => NoWildcardMatch);
```
sharwell, line 280: 📝 This is the bug fix. The rest is code cleanup.

mavasani (approved):
LGTM, but I would certainly want @Evangelink's review to confirm it does not regress perf on the path where the user has actually specified this option.

sharwell:
> does not regress perf

`_wildcardNamesBySymbolKind` is immutable, so if it is empty for one call through this path it will be empty for all subsequent calls through this instance. This is an earlier return than the first point where this cache is ever used.

mavasani:
@sharwell I can release new 3.3.2 analyzer NuGet packages once this PR is merged in.

PR #4449, opened 2020-11-20
Update the compiler to the 3.8.0 release

sharwell:


eng/Versions.props
```diff
@@ -23,11 +23,12 @@
     <ReportGeneratorVersion>4.3.6</ReportGeneratorVersion>
     <!-- Use the correct compiler version -->
     <UsingToolMicrosoftNetCompilers>true</UsingToolMicrosoftNetCompilers>
-    <CodeStyleAnalyersVersion>3.8.0-3.20428.4</CodeStyleAnalyersVersion>
+    <MicrosoftNetCompilersToolsetVersion>3.8.0</MicrosoftNetCompilersToolsetVersion>
```
Evangelink, line 26: This seems to be a new property but there is no file updated to be using it.

mavasani: This was explicitly removed by @mmitche as this causes a dependency chain issue when inserting analyzers into the SDK repo. We have been asked explicitly to not reference `MicrosoftNetCompilersToolsetVersion`, but instead use in-built compiler toolset from arcade version. Also adding @jmarolf

mavasani (changes requested):
We have been asked to not reference `MicrosoftNetCompilersToolsetVersion`

eng/Versions.props
sharwell: The Microsoft.Net.Compilers.Toolset dependency does not propagate transitively.

mavasani: @sharwell Let me add you to the email thread.

mavasani (approved):

Evangelink:
@mavasani So is it ok to use `MicrosoftNetCompilersToolsetVersion`?

mavasani:
@Evangelink Based on offline conversations, it was agreed it is fine. When we are nearing an official release as part of .NET SDK, we would have to remove it.

PR #4463, opened 2020-11-26
Update CA2227 to not report on init only property

### LINKED ISSUE #4461: CA2227 false positive on record type
### Analyzer

**Diagnostic ID**: [CA2227](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2227): `Collection properties should be read only`

### Analyzer source

**Version**: [SDK 5.0.100](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug

The analyzer incorrectly reports a CA2227 for a list property of a record.

### Steps To Reproduce

Write this code in a project where CA2227 is enabled

```csharp
public record MyRecord(IList<Item> Items);
```

### Expected behavior

No warning

### Actual behavior

CA2227 warning on `Items`:

> Change 'Items' to be read-only by removing the property setter

But here the setter is an `init` accessor (compiler-generated), which means the property can't be reassigned after creation.

## Additional context

The problem also occurs for properties with init-only setters on non-record types.
The analyzer should look for a `modreq` of type `IsExternalInit` on the setter's return parameter, as specified in the [init-only setter proposal](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init)

Evangelink:
Fix #4461

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
@@ -438,6 +438,7 @@ class A
             await new VerifyCS.Test
             {
                 LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
                 TestCode = @"
 using System.Collections.Generic;
 
@@ -451,6 +452,7 @@ public record MyRecord(IList<int> Items);",
             await new VerifyCS.Test
             {
                 LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
                 TestCode = @"
 using System.Collections.Generic;
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
@@ -431,5 +431,39 @@ class A
 }
 ");
         }
+
+        [Fact, WorkItem(4461, "https://github.com/dotnet/roslyn-analyzers/issues/4461")]
+        public async Task CA2227_CSharp_InitPropertyRecord()
+        {
+            await new VerifyCS.Test
+            {
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                TestCode = @"
+using System.Collections.Generic;
+
+public record MyRecord(IList<int> Items);",
+            }.RunAsync();
+        }
+
+        [Fact, WorkItem(4461, "https://github.com/dotnet/roslyn-analyzers/issues/4461")]
+        public async Task CA2227_CSharp_InitProperty()
+        {
+            await new VerifyCS.Test
+            {
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                TestCode = @"
+using System.Collections.Generic;
+
+class C
+{
+    public IList<int> L { get; init; }
+}
+
+struct S
+{
+    public IList<int> L { get; init; }
+}",
+            }.RunAsync();
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
```diff
@@ -431,5 +431,39 @@ class A
 }
 ");
         }
+
+        [Fact, WorkItem(4461, "https://github.com/dotnet/roslyn-analyzers/issues/4461")]
+        public async Task CA2227_CSharp_InitPropertyRecord()
```
Evangelink, line 436: Test are failing because some type is not imported, I suspect there might be a need to update the test sdk but I am not entirely sure.

@sharwell may I ask your help?

Evangelink: FYI the error is `error CS0518: Predefined type 'System.Runtime.CompilerServices.IsExternalInit' is not defined or imported`

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
```diff
@@ -431,5 +431,39 @@ class A
...
+        {
+            await new VerifyCS.Test
+            {
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                TestCode = @"
+using System.Collections.Generic;
+
+public record MyRecord(IList<int> Items);",
```
Youssef1313, line 444: @Evangelink For the issue you're facing in https://github.com/dotnet/roslyn-analyzers/pull/4463/files#r531133849

I think you can fix it like this:

```suggestion
public record MyRecord(IList<int> Items);

namespace System.Runtime.CompilerServices
{
    class IsExternalInit {}
}",
```

Youssef1313: The tests of the IDE-side of roslyn currently follow that approach:

See https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Analyzers/CSharp/Tests/AddAccessibilityModifiers/AddAccessibilityModifiersTests.cs#L163

Evangelink: It should work but my understanding was that we try not to rely on fake class and instead use the real code.

Youssef1313: @Evangelink The real class is already empty.

https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/IsExternalInit.cs

Evangelink: I meant we replaced the various fake class in the tests of roslyn-analyzer to rely on having a reference to the real dll (direct dll or nuget package). I think here we should also do the same.

Youssef1313: @Evangelink I'm not sure what is the thing that determines the runtime version under which the tests are run.
Could updating MicrosoftCodeAnalysisTestingVersion help? Not sure.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/CollectionPropertiesShouldBeReadOnlyTests.cs
```diff
@@ -431,5 +431,39 @@ class A
 }
 ");
         }
+
+        [Fact, WorkItem(4461, "https://github.com/dotnet/roslyn-analyzers/issues/4461")]
+        public async Task CA2227_CSharp_InitPropertyRecord()
+        {
+            await new VerifyCS.Test
+            {
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
```
Youssef1313, line 440: This *might* work, @Evangelink 

```suggestion
                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
```

Evangelink: Of course it will work! Totally forgot about that, thanks!

mavasani (approved):

PR #4477, opened 2020-12-01
Fix RS1024: Do not report on overloads not comparing symbols

### LINKED ISSUE #4413: False positive for RS1024 (Compare symbols correctly) when not comparing symbols
It seems that in some cases, when a method has an overload that takes an `IEqualityComparer<T>`, and one of the generic arguments to that method is `ISymbol` or a type that derives from `ISymbol`, the analyzer assumes that the symbols are being compared. That is not necessarily true.

### Steps To Reproduce

Consider this method:

```cs
ILookup<string, IFieldSymbol> GetNames(IFieldSymbol[] fields) => fields.ToLookup(f => f.Name);
```

### Expected behavior

No symbols are being compared, so no diagnostic is reported.

Note that the method has an overload taking a `IEqualityComparer<string>`, it seems this may confuse the analyzer.

### Actual behavior

The following diagnostic is reported:

    warning RS1024: Compare symbols correctly

Evangelink:
Fix #4413


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -182,22 +184,28 @@ private static void HandleBinaryOperator(in OperationAnalysisContext context, IN
...
                 if (method is null)
                 {
                     return false;
                 }
                 else if (!method.TypeArguments.IsEmpty)
                 {
-                    return IsSymbolType(method.TypeArguments[0], symbolType);
+                    var destinationTypeIndex = method.TypeParameters
+                        .Select((type, index) => type.Name == "TKey" ? index : -1)
+                        .FirstOrDefault(x => x >= 0);
```
Evangelink, line 197: /!\ This is a bit hacky. /!\ 

I am not sure whether we can safely rely on a generic type parameter name.

As far as I can tell, the non-hacky solution forces us to resolves all overloads, and map them with the index we should look at.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -168,9 +169,10 @@ private static void HandleBinaryOperator(in OperationAnalysisContext context, IN
                 default:
                     if (equalityComparerMethods.TryGetValue(method.Name, out var possibleMethodTypes))
                     {
+                        var xx = method.GetOverloads();
```
mavasani, line 172:
```suggestion
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
mavasani: I'll let @ryzngard review/approve this change.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -170,7 +171,7 @@ private static void HandleBinaryOperator(in OperationAnalysisContext context, IN
                     {
                         if (symbolEqualityComparerType != null &&
```
KrisVandermotten, line 172: Suggestion: collapse the two `if` statements, and move the `symbolEqualityComparerType != null` check before the `TryGetValue` call.

Evangelink: I'd rather do the change in a separate PR as it's not really related to this fix.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
Evangelink: Ping @ryzngard

ryzngard: Looking, sorry for delay

ryzngard: Agree this does seem hacky, but maybe no more hacky than depending on ordering. At the end of the day we're at least working on a known set of functions that we don't expect to change drastically, so if it works for the known set maybe it's good enough.

I was looking at the test you wrote without the change, and a surprising piece for me: the type arguments are not in the order I would expect. 

Method display is showing `ILookup<string, IFieldSymbol>` but type arguments are `{ IFieldSymbol, string }`. Going to try and find documentation to see if that's expected.

ryzngard: Oh nm, this makes sense. Just got slightly confused with how the information was display. `ToLookup<IFieldSymbol, string>` is the method here so the first type argument is correct. Maybe for things like `ToLookup`, `ToDictionary`, and `GroupBy` we should be looking at the return type and it's typearguments instead of the method signature arguments

Evangelink: I can but this will add some complexity to the current solution. Let me know what's your preference.

mavasani: Tagging @ryzngard for last question

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
+using System.Diagnostics;
```
Youssef1313, line 10: nit: Move before `using Analyzer.Utilities`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
Evangelink: Ping @ryzngard

Evangelink: Ping @ryzngard

ryzngard: This looks fine for now

ryzngard (approved):

PR #4489, opened 2020-12-03
Use package name suffix for property names used inside `AddGlobalAnalyzerConfigForPackage` target

mavasani:
This prevents clashes between property values when executed in context of different analyzer packages. Found this issue while dogfooding the NetAnalyzers package, when also referencing another analyzer package produced from this repo.


mavasani:
Merging as this is a trivial change needed to unblock package release.

PR #4503, opened 2020-12-04
PostRelease activities after release of new analyzer packages

mavasani:
Also migrated dogfood analyzers from FxCopAnalyzers to NetAnalayzers, with AnalysisMode = AllEnabledByDefault, which enabled additional CA analyzers in the repo. I have fixed/disabled the additional analyzers as appropriate.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -252,6 +252,25 @@ dotnet_code_quality.CA1720.api_surface = public
 # CA1715: Identifiers should have correct prefix
 dotnet_code_quality.CA1715.exclude_single_letter_type_parameters = true
 
+# CA1002: Do not expose generic lists
+dotnet_diagnostic.CA1002.severity = suggestion
+
+# CA1024: Use properties where appropriate
+dotnet_diagnostic.CA1024.severity = suggestion
+
+# CA1033: Interface methods should be callable by child types
+dotnet_diagnostic.CA1033.severity = suggestion
+
+# CA1711: Identifiers should not have incorrect suffix
+dotnet_diagnostic.CA1711.severity = suggestion
+
+# CA1309: Use ordinal string comparison
+dotnet_diagnostic.CA1309.severity = suggestion
+
+# CA1725: Parameter names should match base declaration
+# TODO: Enable this rule as warning and fix violations.
+dotnet_diagnostic.CA1725.severity = suggestion
+
 ### Configuration for PublicAPI analyzers executed on this repo ###
 [*.{cs,vb}]
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -261,6 +261,9 @@ dotnet_diagnostic.CA1024.severity = suggestion
 # CA1033: Interface methods should be callable by child types
 dotnet_diagnostic.CA1033.severity = suggestion
 
+# CA1307: Specify StringComparison for clarity
+dotnet_diagnostic.CA1307.severity = suggestion
+
 # CA1711: Identifiers should not have incorrect suffix
 dotnet_diagnostic.CA1711.severity = suggestion
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -23,6 +23,9 @@
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
+    <!-- CA1014: Mark assemblies with CLSCompliant -->
+    <NoWarn>$(NoWarn),CA1014</NoWarn>
+    
     <!--
       Make sure any documentation comments which are included in code get checked for syntax during the build, but do
       not report warnings for missing comments.
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -23,9 +23,6 @@
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
-    <!-- CA1014: Mark assemblies with CLSCompliant -->
-    <NoWarn>$(NoWarn),CA1014</NoWarn>
-    
     <!--
       Make sure any documentation comments which are included in code get checked for syntax during the build, but do
       not report warnings for missing comments.
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -8,9 +8,6 @@
 
     <DefineConstants Condition="'$(BUILDING_VSIX)' == 'true'">$(DefineConstants),BUILDING_VSIX</DefineConstants>
     <DefineConstants Condition="'$(LEGACY_CODE_METRICS_MODE)' == 'true'">$(DefineConstants),LEGACY_CODE_METRICS_MODE</DefineConstants>
-
-    <AnalysisLevel>latest</AnalysisLevel>
-    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
   </PropertyGroup>
 
   <!-- Test runner configuration -->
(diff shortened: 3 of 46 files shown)


.editorconfig
```diff
@@ -252,6 +252,25 @@ dotnet_code_quality.CA1720.api_surface = public
...
+# CA1002: Do not expose generic lists
+dotnet_diagnostic.CA1002.severity = suggestion
+
+# CA1024: Use properties where appropriate
+dotnet_diagnostic.CA1024.severity = suggestion
+
+# CA1033: Interface methods should be callable by child types
+dotnet_diagnostic.CA1033.severity = suggestion
+
+# CA1711: Identifiers should not have incorrect suffix
```
mavasani, line 264:
```suggestion
# CA1307: Specify StringComparison for clarity
dotnet_diagnostic.CA1307.severity = suggestion

# CA1711: Identifiers should not have incorrect suffix
```

Directory.Build.props
```diff
@@ -23,6 +23,9 @@
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
+    <!-- CA1014: Mark assemblies with CLSCompliant -->
```
Evangelink, line 26: Out of curiosity why this change? I might miss something but either we want to be compliant so let's fix  this or we don't care and so let's disable the rule.

src/Directory.Build.props
```diff
@@ -8,6 +8,9 @@
 
     <DefineConstants Condition="'$(BUILDING_VSIX)' == 'true'">$(DefineConstants),BUILDING_VSIX</DefineConstants>
     <DefineConstants Condition="'$(LEGACY_CODE_METRICS_MODE)' == 'true'">$(DefineConstants),LEGACY_CODE_METRICS_MODE</DefineConstants>
+
```
Evangelink, line 11: You probably either want to remove this empty line or move the following entries to a different `PropertyGroup`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/RegisterActionAnalyzer.cs
```diff
@@ -369,7 +369,9 @@ private void AnalyzeInvocation(SyntaxNodeAnalysisContext context)
                                         {
                                             symbol = semanticModel.GetSymbolInfo(argument, context.CancellationToken).Symbol;
                                             if (symbol != null &&
+#pragma warning disable CA1508 // Avoid dead conditional code - TODO: File an issue.
```
Evangelink, line 372: TODO should be fixed/handled.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DataSetDataTableInWebSerializableObjectGraphTests.cs
```diff
@@ -90,7 +90,9 @@ public class MyClass
 
         private static async Task VerifyServiceModelCSharpAsync(string source, params DiagnosticResult[] expected)
         {
+#pragma warning disable CA5386 // Avoid hardcoding SecurityProtocolType value
```
Evangelink, line 93: Might be worth updating the analyzer to not report on test method, what do you think?

src/Tools/GenerateAnalyzerNuspec/Program.cs
```diff
@@ -58,7 +58,9 @@
         result.AppendLine($"    <{name}>{value}</{name}>");
     }
 
+#pragma warning disable CA1508 // Avoid dead conditional code - TODO: File a bug
```
Evangelink, line 61: TODO

src/Utilities/Compiler/Extensions/SymbolVisibility.cs
```diff
@@ -4,7 +4,9 @@
 
 namespace Analyzer.Utilities.Extensions
 {
+#pragma warning disable CA1027 // Mark enums with FlagsAttribute
```
Evangelink, line 7: Interesting, do we need the values? It would prevent this warning.

src/Utilities/Compiler/Lightup/LightupHelpers.cs
```diff
@@ -7,6 +7,8 @@
 using System.Reflection;
 using Microsoft.CodeAnalysis;
 
+#pragma warning disable CA2201 // Do not raise reserved exception types - 'throw new NullReferenceException()' is used in this file
```
Evangelink, line 10: Would be nice to be able to exclude all rules on a given folder/namespace (Lightup folder).

.editorconfig
```diff
@@ -252,6 +252,28 @@ dotnet_code_quality.CA1720.api_surface = public
...
+dotnet_diagnostic.CA1033.severity = suggestion
+
+# CA1307: Specify StringComparison for clarity
+dotnet_diagnostic.CA1307.severity = suggestion
+
+# CA1711: Identifiers should not have incorrect suffix
+dotnet_diagnostic.CA1711.severity = suggestion
+
+# CA1309: Use ordinal string comparison
+dotnet_diagnostic.CA1309.severity = suggestion
```
Evangelink, line 271: Why not as warning?

Directory.Build.props
mavasani: We don't care about this rule, the NoWarn is turning it off, but I can also instead move it to https://github.com/dotnet/roslyn-analyzers/blob/master/eng/Analyzers_ShippingRules.ruleset. Let me do that change.

.editorconfig
mavasani: There are many violations, so probably best to address it in a follow-up PR.

src/Directory.Build.props
mavasani: Sure, let me do the latter and add a comment for clarity.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/RegisterActionAnalyzer.cs
mavasani, line 372:
```suggestion
#pragma warning disable CA1508 // Avoid dead conditional code - https://github.com/dotnet/roslyn-analyzers/issues/4519
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Security/DataSetDataTableInWebSerializableObjectGraphTests.cs
mavasani: I think it is unlikely that we see this violation in test code except the test project in this repo. Not sure if it is worth the effort.

src/Tools/GenerateAnalyzerNuspec/Program.cs
mavasani:
```suggestion
#pragma warning disable CA1508 // Avoid dead conditional code - https://github.com/dotnet/roslyn-analyzers/issues/4520
```

src/Utilities/Compiler/Extensions/SymbolVisibility.cs
mavasani: Possibly not, but I want to avoid any real semantic changes in this PR.

src/Utilities/Compiler/Lightup/LightupHelpers.cs
mavasani: Good idea, let me add an editorconfig to the folder.

Evangelink: I forgot about that xD I was already thinking about some exclusion pattern. I like the simple solution :)

mavasani:
Thanks for the review @Evangelink - I think I addressed all your feedback. Can you please take another look?

Evangelink (approved):

mavasani:
Thanks @Evangelink!

PR #4505, opened 2020-12-05
Fix wording of one of the CA2231 messages

jnm2:
- An article is needed since the word 'method' is there
- Flows better and matches other messages when it's just a name and not quoted as code

I used VS Code's replace-in-all-files across the repo.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
@@ -413,7 +413,7 @@
     <value>In most programming languages there is no default implementation of the equality operator (==) for value types. If your programming language supports operator overloads, you should consider implementing the equality operator. Its behavior should be identical to that of Equals.</value>
   </data>
   <data name="OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage" xml:space="preserve">
-    <value>Implement the equality operators and make their behavior identical to that of 'Equals' method</value>
+    <value>Implement the equality operators and make their behavior identical to that of the 'Equals' method</value>
   </data>
   <data name="PassSystemUriObjectsInsteadOfStringsTitle" xml:space="preserve">
     <value>Pass system uri objects instead of strings</value>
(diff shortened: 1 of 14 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/xlf/MicrosoftCodeQualityAnalyzersResources.cs.xlf
```diff
@@ -718,7 +718,7 @@
         <note />
       </trans-unit>
       <trans-unit id="OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage">
-        <source>Implement the equality operators and make their behavior identical to that of 'Equals' method</source>
+        <source>Implement the equality operators and make their behavior identical to that of the Equals method</source>
```
Youssef1313, line 721: I think xlf files shouldn't be manually modified to allow the localized messages to be reviewed again after this change.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -413,7 +413,7 @@
     <value>In most programming languages there is no default implementation of the equality operator (==) for value types. If your programming language supports operator overloads, you should consider implementing the equality operator. Its behavior should be identical to that of Equals.</value>
   </data>
   <data name="OverloadOperatorEqualsOnOverridingValueTypeEqualsMessage" xml:space="preserve">
-    <value>Implement the equality operators and make their behavior identical to that of 'Equals' method</value>
+    <value>Implement the equality operators and make their behavior identical to that of the Equals method</value>
```
Evangelink, line 416: I don't have a strong opinion on the way forward but as far as I know the idea is to always put the single quotes when the word refers to a type, method... so I would stick to `the 'Equals'`.

Cc @mavasani

jnm2: I agree with quotes for things that are code but not for proper names. 'Equals' could be either depending on how it is used in the message. In "the Equals method," it's being used as a proper name and not as a quote of C# syntax as such.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/xlf/MicrosoftCodeQualityAnalyzersResources.cs.xlf
jnm2: Thanks! Done.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: I am not a native English speaker so I won't make any decision on that but I think I would have _wrongly_ written "identical to that of the `Equals` method." xD

Anyways @mavasani will review and give his opinion here.

mavasani: > This is my reasoning: if you were writing about it here, you would say either "identical to that of the Equals method" or "identical to that of Equals." Not "identical to that of the Equals method."

Seems reasonable to me. @Evangelink is correct that we try to put actual symbol names in quotes, but this change seems fine.

mavasani (approved):

PR #4522, opened 2020-12-07
Fix CA1307 and CA1310 to not report on IQueryable context.

### LINKED ISSUE #3492: CA1307 reported for LINQ Expression, suggested fix fails at runtime
#### Analyzer package

#### Package Version

v2.9.8 (latest release)

#### Diagnostic ID

Example: [CA1307](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1307)

#### Repro steps

```csharp
new DbContext().MyEntities.Where(x => x.Name.StartsWith("Hello"))
```

#### Expected behavior

No diagnostic is reported; actual comparison is done inside SQL so the "thread" is not relevant (I think).

#### Actual behavior

Issue CA1307 is reported, suggesting "StartsWith(string, StringComparison)":

> The behavior of 'string.StartsWith(string)' could vary based on the current user's locale settings. Replace this call in '...' with a call to 'string.StartsWith(string, System.StringComparison)'.

However that suggestion fails at runtime:

> System.NotSupportedException: 'LINQ to Entities does not recognize the method 'Boolean StartsWith(System.String, System.StringComparison)' method, and this method cannot be translated into a store expression.'

Evangelink:
Fix #3492


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
```diff
@@ -82,6 +84,18 @@ public override void Initialize(AnalysisContext analysisContext)
                         return;
                     }
 
+                    if (iqueryableType != null)
+                    {
+                        var ancestorInvocation = invocationExpression.GetAncestor<IInvocationOperation>(OperationKind.Invocation);
```
Evangelink, line 89: We could consider moving this block check at the time where we report if there is any performance concern with getting up the operation tree.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
```diff
@@ -82,6 +84,18 @@ public override void Initialize(AnalysisContext analysisContext)
...
+
+                        // If we are in a IQueryable context, it's possible that the underlying call doesn't have the comparison
+                        // option so we want to bail-out to be safe.
+                        if (ancestorInvocation != null && iqueryableType.Equals(ancestorInvocation.Type?.OriginalDefinition))
```
Evangelink, line 93: My first thought was to not stop on the first invocation ancestor but it doesn't seem to be needed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
mavasani: This seems reasonable, but I am wondering if we can do a more targeted fix. For example, I am not sure if this will be behave correctly when crossing lambda/local function boundary.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
```diff
@@ -81,6 +86,19 @@ public override void Initialize(AnalysisContext analysisContext)
                         return;
                     }
 
+                    if (linqExpressionType != null)
+                    {
+                        var enclosingLambdaOrLocalFunc = invocationExpression.GetAncestor(s_LambdaOrLocalFunctionKinds);
```
Evangelink, line 91: @mavasani This new implementation should not cross lambdas boundaries.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
```diff
@@ -81,6 +86,19 @@ public override void Initialize(AnalysisContext analysisContext)
...
+
+                        // Check if we are in a Expression<Func<T...>> context, in which case it is possible
```
Evangelink, line 93: Rather than `IQueryable`, I am now checking for `Expression<TDelegate>` (as suggested by @damieng).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
mavasani: Thanks, this looks much simpler. Do you think it is worth moving the entire added code block here into an extension method on IOperation, say `IsInExpressionTree` or some such, given we might likely need more analyzers to bail out in expression tree context?

mavasani (approved):
LGTM.

Let me know if you prefer moving to a helper extension method, otherwise I can merge.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs
Evangelink: Let's wait for the next use to create the helper.

Issue #4342, opened 2020-10-19
CA1801 False positive on event handlers set in XAML for UWP apps


JuanMejiaVelez:
### Analyzer
Microsoft.CodeAnalysis.FxCopAnalyzers

**Diagnostic ID**: [CA2013](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2013): `Do not use ReferenceEquals with value types`

### Analyzer source
**NuGet Package**: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

**Version**: v3.3.0 (Latest)

### Describe the bug
The analyzer is flagging event handlers that are set in XAML instead of code-behind.

### Steps To Reproduce
When creating XAML pages on UWP, and adding event handlers in XAML code, the event handler is flagged in the C# code as parameters not being used or sometimes as the event should be marked static.
When the event is subscribe in code-behind this does not repro.

For WinUI. NavigationView it reproes 100%

### Expected behavior
Event handler is not flagged as offending CA1801
### Actual behavior
Event handler is flagged.

mavasani:
Tagging @Evangelink - I thought this was handled in https://github.com/dotnet/roslyn-analyzers/pull/3155

mavasani:
Also related to https://github.com/dotnet/roslyn/issues/38192 - underlying issue is with XAML tooling not correctly auto-updating the code behind file, and they have closed it as won't fix.

Evangelink:
@JuanMejiaVelez Would you mind testing if you have the same issue with `IDE0060`? `CA1801` is deprecated in favor of the other rule and I don't have UWP on my machine so I can't really check if the false-positive also exists in the other rule.

Evangelink:
Ping @JuanMejiaVelez

Evangelink:
Ping @JuanMejiaVelez

Issue #3752, opened 2020-06-15
CA1801: False negatives due to argument validation


bordecal:
The current implementation of CA1801 does not exclude common argument validation approaches from detected parameter "usage", resulting in a potentially high false negative rate for a code base that makes heavy use of argument validation.

The rule would be quite a bit more useful if it included an option for ignoring validation "usage", including both inline validation and validation via helper methods (_à la_ https://docs.microsoft.com/en-us/dotnet/api/microsoft.requires).  Ideally, this would allow for a more than just `ValidatedNotNullAttribute` for identifying a validation helper method since not all argument validation involves null checks (e.g.: https://docs.microsoft.com/en-us/dotnet/api/microsoft.requires.range).

#### Analyzer package

[Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

[CA1801](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1801) (ReviewUnusedParameters)

#### Repro steps

When a parameter's only usage inside a method is for argument validation (whether inline or via a validation method), a CA1801 violation is not flagged.  For example, a CA1801 violation will not be flagged for the `x` parameter in either `DoSomething1` or `DoSomething2` below.  (The same result occurs even if `nameof()` is not used to retrieve the parameter name.)

```csharp
public void DoSomething1(object x)
{
    if (x == null)
    {
        throw new ArgumentNullException(nameof(x));
    }
}

public void DoSomething2(object x)
{
    Argument.NotNull(x, nameof(x));
}
```

where the following are present for `DoSomething2`:

```csharp
[AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
public sealed class ValidatedNotNullAttribute : Attribute
{
}

public static class Argument
{
    public static T NotNull<T>([ValidatedNotNull] this T value, string parameterName)
    {
        return value ?? throw new ArgumentNullException(parameterName);
    }
}
```

#### Expected behavior
Validation should not be sufficient to consider a parameter "used" for the purposes of CA1801.

#### Actual behavior
Use of a parameter for either inline validation or validation via a helper method hides a potential CA1801 problem.

Evangelink:
[...]

PR #4535, opened 2020-12-08
Remove rule CA1801 - replaced by IDE0060

### LINKED ISSUE #4498: Decide which analyzer to keep: CA1801 vs IDE0060
### Analyzer

**Diagnostic ID**: [CA1801](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1801)

### Describe the improvement

Currently 2 rules compete and they have both different implementation and "scope" (findings and issues), rather than "wasting" time maintaining both I would suggest that a clear cut is made and only one is kept. When the choice is made, we should ensure that the winning rule is correctly matching both scopes.

Tagging @mavasani

Evangelink:
Fix #4498


Youssef1313:
There are open issues regarding some false positives.

https://github.com/dotnet/roslyn-analyzers/issues/3752
https://github.com/dotnet/roslyn-analyzers/issues/4342

We need to check whether these false positives exist in IDE0060 or not. Then close them or move to roslyn depending on whether IDE0060 has the same issues or not.

Evangelink:
@mavasani It's written not to do it manually but there was no way to apply the code-fix for the removed rule so I did it manually. If that's not expected we should create a ticket.

Youssef1313:
> @mavasani It's written not to do it manually but there was no way to apply the code-fix for the removed rule so I did it manually. If that's not expected we should create a ticket.

I think the codefix should be able to do that. If it doesn't, then it's a codefix issue.

Evangelink:
Well the issue is only available in the error-list and from there we cannot run a code-fix so I think that's a more global topic (to run a code-fix from the error window).

Youssef1313:
@Evangelink Oh correct. The diagnostic in this case doesn't have a location ;-(

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/Maintainability/CSharpReviewUnusedParameters.Fixer.cs
```diff
@@ -1,28 +0,0 @@
...
-using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeQuality.Analyzers.Maintainability;
-
-namespace Microsoft.CodeQuality.CSharp.Analyzers.Maintainability
-{
-    /// <summary>
-    /// CA1801: Review unused parameters
-    /// </summary>
-    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
-    public sealed class CSharpReviewUnusedParametersFixer : ReviewUnusedParametersFixer
```
mavasani, line 15: IDE0060 does not have a code fix. CodeFix for CA1801 had some shortcomings and heuristics. I think we should consider opening a Roslyn issue to author a code fix for IDE0060, which uses this code fix implementation as a reference.

mavasani (approved):
Thank you @Evangelink. Please make sure to open a doc PR to mention deprecation once this is merged.

mavasani:
> Well the issue is only available in the error-list and from there we cannot run a code-fix so I think that's a more global topic (to run a code-fix from the error window).

CA1801 should have a code fix from editor, the diagnostic does have a source file location. Is that not what you are observing?

mavasani:
> It's written not to do it manually but there was no way to apply the code-fix for the removed rule so I did it manually

Ah, you mean the diagnostic to update the analyzer releases file? Sorry, I misunderstood. In that case, yes, that is by design as that diagnostic cannot be mapped to any source location. Manual update is fine for rule removal.

mavasani (approved):

Evangelink:
@mavasani deprecation doc is already done and available live (see https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1801)

PR #4543, opened 2020-12-09
Remove pragma for SymbolNamesWithValueOption

### LINKED ISSUE #4350: Remove the pragma suppression added to 'SymbolNamesWithValueOption'
Remove the pragma suppression added to `SymbolNamesWithValueOption` to unblock https://github.com/dotnet/roslyn-analyzers/pull/4347

Evangelink:
Fix #4350

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
@@ -177,9 +177,9 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             {
                 // User specific mapping has higher priority than hardcoded one
                 if (userMap.TryGetValue(type.OriginalDefinition, out var suffix) &&
-                    !string.IsNullOrWhiteSpace(suffix))
+                    !RoslynString.IsNullOrWhiteSpace(suffix))
                 {
-                    suffixInfo = SuffixInfo.Create(suffix!, canSuffixBeCollection: false);
+                    suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
                     return true;
                 }
 
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
@@ -179,7 +179,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                 if (userMap.TryGetValue(type.OriginalDefinition, out var suffix) &&
                     !string.IsNullOrWhiteSpace(suffix))
                 {
-                    suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
+                    suffixInfo = SuffixInfo.Create(suffix!, canSuffixBeCollection: false);
                     return true;
                 }
 
--- a/src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
+++ b/src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
@@ -224,7 +224,7 @@ namespace Analyzer.Utilities
         /// <summary>
         /// Gets the value associated with the specified symbol in the option specification.
         /// </summary>
-        public bool TryGetValue(ISymbol symbol, [MaybeNullWhen(false)] out TValue? value)
+        public bool TryGetValue(ISymbol symbol, out TValue? value)
         {
             if (_symbols.TryGetValue(symbol, out value) || _names.TryGetValue(symbol.Name, out value))
             {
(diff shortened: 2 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
```diff
@@ -179,7 +179,7 @@ private static void AnalyzeCompilationStart(CompilationStartAnalysisContext cont
                 if (userMap.TryGetValue(type.OriginalDefinition, out var suffix) &&
                     !string.IsNullOrWhiteSpace(suffix))
                 {
-                    suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
+                    suffixInfo = SuffixInfo.Create(suffix!, canSuffixBeCollection: false);
```
Evangelink, line 182: The `string.IsNullOrWhiteSpace` is not taken into account.

sharwell: ❕  Use `RoslynString.IsNullOrWhiteSpace` instead of a suppression.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -224,7 +224,7 @@ public bool Contains(ISymbol symbol)
         /// <summary>
         /// Gets the value associated with the specified symbol in the option specification.
         /// </summary>
-        public bool TryGetValue(ISymbol symbol, [MaybeNullWhen(false)] out TValue value)
+        public bool TryGetValue(ISymbol symbol, [MaybeNullWhen(false)] out TValue? value)
```
sharwell, line 227: ❕ The attribute here no longer makes sense.

```suggestion
        public bool TryGetValue(ISymbol symbol, out TValue? value)
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
Evangelink: Thank you @sharwell

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -373,16 +373,14 @@ internal TestAccessor(SymbolNamesWithValueOption<TValue> symbolNamesWithValueOpt
         /// </example>
         public sealed class NameParts
         {
-            public NameParts(string symbolName, TValue associatedValue = default)
+            public NameParts(string symbolName, TValue? associatedValue = default)
```
sharwell, line 376: From a code quality standpoint, it would be vastly preferable to remove the default value here.

```suggestion
            public NameParts(string symbolName, TValue associatedValue)
```

Most of the rest of the changes in this pull request are not necessary with this change.

Evangelink: I will give it another try but I started with that and couldn't do the fix.

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
```diff
@@ -19,11 +19,11 @@ internal sealed class SymbolNamesWithValueOption<TValue> : IEquatable<SymbolName
         internal const SymbolKind AllKinds = SymbolKind.ErrorType;
         internal const char WildcardChar = '*';
 
-        public static readonly SymbolNamesWithValueOption<TValue> Empty = new SymbolNamesWithValueOption<TValue>();
-        internal static KeyValuePair<string, TValue> NoWildcardMatch => default;
+        public static readonly SymbolNamesWithValueOption<TValue> Empty = new();
+        internal static KeyValuePair<string, TValue?> NoWildcardMatch => default;
```
Youssef1313, line 23: This had no nullable warning before. Is it one of the known pitfalls to nullable feature?

[Minimal repro](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4GJPoA2hEwhcABHJqeQLABQjAAgEwCMLADBc+wCwBuRowDScAJ4A1CIXRwAChACWCADzLMMJL3ZcAfBQDWANwoBeCgBM4AMwhEYwhnwCcAClMA6GXLheAFQB7AGUYBE0Ac3cAShjBIA===)

src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs
Evangelink: I confirm that I am blocked, basically now [line 92](https://github.com/dotnet/roslyn-analyzers/pull/4543/files#diff-d26839098724560378df49fbfe7309653539ef2089e47babbee13a07e63ec988R92) we need a default value, so I need to pass it to the method signature and going up the chain I end up to places using `ISymbol` so I can't pass a value here. Note that in this case we pass the lambda so we should not use the default value.

sharwell: I'm going to do one more pass locally to experiment with this and see if I need to file a compiler feature request.

Evangelink: Sure

sharwell:
Alternate implementation provided in #4553

PR #4547, opened 2020-12-10
Add support for reading AnalysisLevelSuffix

mavasani:
Follow-up to https://github.com/dotnet/sdk/pull/14833 towards addressing part of https://github.com/dotnet/roslyn/issues/49250#issuecomment-736879786

Apart from reading `AnalysisLevelSuffix`, we also support reading the `AnalysisMode` property for back compat and also allow additional property values `Minimum` and `Recommended`.

### DIFF
--- a/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
+++ b/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
@@ -713,7 +713,9 @@ Rule ID | Missing Help Link | Title |
                         {
                             CreateGlobalconfig(
                                 analyzerGlobalconfigsDir,
-                                $"AnalysisLevel_{analysisLevelVersionString}_{analysisMode}.editorconfig",
+#pragma warning disable CA1308 // Normalize strings to uppercase
+                                $"AnalysisLevel_{analysisLevelVersionString}_{analysisMode!.ToString()!.ToLowerInvariant()}.editorconfig",
+#pragma warning restore CA1308 // Normalize strings to uppercase
                                 $"Rules from '{version}' release with '{analysisMode}' analysis mode",
                                 $"Rules with enabled-by-default state from '{version}' release with '{analysisMode}' analysis mode. Rules that are first released in a version later than '{version}' are disabled.",
                                 (AnalysisMode)analysisMode!,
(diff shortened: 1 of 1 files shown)


src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -713,7 +713,9 @@ bool createGlobalConfigFiles()
                         {
                             CreateGlobalconfig(
                                 analyzerGlobalconfigsDir,
-                                $"AnalysisLevel_{analysisLevelVersionString}_{analysisMode}.editorconfig",
+#pragma warning disable CA1308 // Normalize strings to uppercase
+                                $"AnalysisLevel_{analysisLevelVersionString}_{analysisMode!.ToString()!.ToLowerInvariant()}.editorconfig",
```
Evangelink, line 717: Isn't it better to have assertions or null check?

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -1435,8 +1464,10 @@ private enum RulesetKind
         private enum AnalysisMode
         {
             Default,
-            AllDisabledByDefault,
-            AllEnabledByDefault,
+            None,
+            Minimum,
+            Recommended,
+            All
```
Youssef1313, line 1470: Isn't this a breaking change?

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
mavasani: I believe the C# compiler is generating nullable warnings because both Enum.GetValues and ToString are annotated as something that can return null, which is obviously not possible here.

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
mavasani: No, the updates to the generated targets above ensure back compat and that old values are correctly mapped: `AllEnabledByDefault` will be mapped to `All` and `AllDisabledByDefault` will be mapped to `None`.

mavasani: I have also manually verified this by trying out the locally built package.

Evangelink (approved):

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
Youssef1313: @mavasani The comment above the enum is now misleading I think.

It assumes that the names in the enum are used directly in AnalysisLevel MSBuild property.

PR #4551, opened 2020-12-10
Do not report RS1025 on empty Initialize method

### LINKED ISSUE #4101: RS1025 (Configure generated code analysis) false positive for empty Initialize method

Evangelink:
Fix #4101

@sharwell It was actually hard to implement because of technical difference between C# and VB.NET.

### DIFF
--- a/src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
+++ b/src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
@@ -2,6 +2,8 @@
 
 #if HAS_IOPERATION
 
+using System;
+using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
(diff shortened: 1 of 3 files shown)


mavasani (changes requested):
I don't believe this is a problem with the analyzer as mentioned in https://github.com/dotnet/roslyn-analyzers/issues/4101#issuecomment-692820153. I will post a comment on the issue and try to close it out as by-design

src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
```diff
@@ -10,13 +12,41 @@ namespace Analyzer.Utilities.Extensions
 {
     internal static class OperationBlockAnalysisContextExtension
     {
-#pragma warning disable RS1012 // Start action has no registered actions.
         public static bool IsMethodNotImplementedOrSupported(this OperationBlockStartAnalysisContext context)
```
mavasani, line 15: This is already a complicated method and the underlying issue absolutely does not justify adding so much complexity to it.

Evangelink: Shall I open a PR with the changes on the helper or do you think it's not worth it and we can redo them later if needs be?

mavasani: @Evangelink What changes are needed to helper? Is it just the removal of `RS1012` suppression?

Evangelink: My question do we want to keep the helper to detect empty method/block? If so I will do a separate PR with it, if not then I don't do it (we don't have the need now).

mavasani: I would prefer to avoid any code changes to helper unless it is required by some caller.

PR #4553, opened 2020-12-10
Fix nullability for SymbolNamesWithValueOption

### LINKED ISSUE #4350: Remove the pragma suppression added to 'SymbolNamesWithValueOption'

sharwell:
* Support global options in Roslyn 3.7+
* Fix nullability for `SymbolNamesWithValueOption`

Fixes #4350

### DIFF
--- a/src/Utilities.UnitTests/Options/SymbolNamesWithValueOptionTests.cs
+++ b/src/Utilities.UnitTests/Options/SymbolNamesWithValueOptionTests.cs
@@ -19,7 +19,7 @@ namespace Analyzer.Utilities.UnitTests.Options
         public void WhenNoSymbolNames_ReturnsEmpty()
         {
             // Arrange & act
-            var options = SymbolNamesWithValueOption<Unit>.Create(ImmutableArray<string>.Empty, GetCompilation(), null, null);
+            var options = SymbolNamesWithValueOption<Unit>.Create(ImmutableArray<string>.Empty, GetCompilation(), null, static name => new SymbolNamesWithValueOption<Unit>.NameParts(name, Unit.Default));
 
             // Assert
             Assert.Equal(SymbolNamesWithValueOption<Unit>.Empty, options);
@@ -37,7 +37,7 @@ namespace Analyzer.Utilities.UnitTests.Options
                 {
                     callCount++;
                 }
-                return new SymbolNamesWithValueOption<Unit>.NameParts(symbolName);
+                return new SymbolNamesWithValueOption<Unit>.NameParts(symbolName, Unit.Default);
             }
 
             // Act
@@ -56,7 +56,7 @@ namespace Analyzer.Utilities.UnitTests.Options
             var symbolNames = ImmutableArray.Create("MyNamespace", "MyClass", "MyMethod()", "MyProperty", "MyEvent", "MyField");
 
             // Act
-            var options = SymbolNamesWithValueOption<Unit>.Create(symbolNames, compilation, null, null);
+            var options = SymbolNamesWithValueOption<Unit>.Create(symbolNames, compilation, null, static name => new SymbolNamesWithValueOption<Unit>.NameParts(name, Unit.Default));
 
             // Assert
             Assert.Equal(symbolNames.Length, options.GetTestAccessor().Names.Count);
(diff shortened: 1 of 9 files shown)


src/Utilities/Compiler/Analyzer.Utilities.projitems
```diff
@@ -11,6 +11,9 @@
   <PropertyGroup Condition="'$(MicrosoftCodeAnalysisVersion)' == '' OR '$(MicrosoftCodeAnalysisVersion)' &gt;= '3.3'">
     <DefineConstants>$(DefineConstants),CODEANALYSIS_V3_OR_BETTER</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(MicrosoftCodeAnalysisVersion)' == '' OR '$(MicrosoftCodeAnalysisVersion)' &gt;= '3.7'">
+    <DefineConstants>$(DefineConstants),CODEANALYSIS_V3_7_OR_BETTER</DefineConstants>
```
mavasani, line 15: I am wondering that now that we don't support FxCopAnalyzers, would it be worth just repurposing `CODEANALYSIS_V3_OR_BETTER` define above to mean `3.7` and also bump up the underlying `Microsoft.CodeAnalysis` version used by the repo to `3.7`? What do you think? https://github.com/dotnet/roslyn-analyzers/blob/690be1702a6ae8f915da3bf6cc1b935e3ff264dd/eng/Versions.props#L28?

sharwell: No objection, but perhaps better as a follow-up?

mavasani: Sounds good.

mavasani: I am still confused though, when is this condition true? None of our existing packages in this repo set `MicrosoftCodeAnalysisVersion` to 3.7 or greater.

sharwell: Roslyn.Diagnostics.Analyzers sets it.

mavasani: Ah, thanks :-)

mavasani (approved):

src/Utilities.UnitTests/Options/SymbolNamesWithValueOptionTests.cs
```diff
@@ -19,7 +19,7 @@ public sealed class SymbolNamesWithValueOptionTests
         public void WhenNoSymbolNames_ReturnsEmpty()
         {
             // Arrange & act
-            var options = SymbolNamesWithValueOption<Unit>.Create(ImmutableArray<string>.Empty, GetCompilation(), null, null);
+            var options = SymbolNamesWithValueOption<Unit>.Create(ImmutableArray<string>.Empty, GetCompilation(), null, static name => new SymbolNamesWithValueOption<Unit>.NameParts(name, Unit.Default));
```
Evangelink, line 22: That looks like a possible good improvement for the make static rule, WDYT?

sharwell: I like it, but maybe should be a separate diagnostic ID so it can be turned off?

Evangelink: I am actually curious if it would be hard technically to allow fine-tuning of rule descriptor through the `.editorconfig` so that we can keep rules with a same scope under the same ID but allow specific descriptors to be tweaked.

For example, `dotnet_code_quality.CAXXXX.severity = warning` would keep the same behavior (it controls all descriptor for this ID) but `dotnet_code_quality.CAXXXX-desc1.severity = warning` only change the severity of one of the descriptors.

sharwell: Or something like https://github.com/dotnet/roslyn/issues/27712?

mavasani: I believe @sharwell has a compiler design proposal for allowing sub-diagnostics for similar configuration - different severity buckets for same ID. We should wait for that feature before attempting anything similar ourselves from an analyzer package.

Evangelink: Of course you thought about it before me xD And I even upvoted but forgot about it.

PR #4554, opened 2020-12-10
Couple of fixes for CA1508

### LINKED ISSUE #4509: CA1508: False-positive for generic
### Analyzer

**Diagnostic ID**: [CA1508](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1508): `Avoid dead conditional code`

### Analyzer source

**Version**: v5.0.1 (Latest)

### Describe the bug

CA1508 `'this.m_myValue' is never 'null'` is reported for the code of `ToString()` in the following class:

```c#
public class MyClass<T>
{
    private T m_myValue;

    public MyClass(T myValue)
    {
        this.m_myValue = myValue;
    }

    /// <inheritdoc />
    public override string ToString()
    {
        return this.m_myValue?.ToString() ?? "";
    }
}
```

It appears that the code analysis engine treats the generic parameter `T` as "not nullable" - possibly because I have `<Nullable>enable</Nullable>` (nullable reference types) defined in my project file. However, as far as I understand it, unconstrained generic types can still be null - even if nullable reference types are enabled.

### Steps To Reproduce

Paste the class above in a project that:

* targets .NET 5
* has nullable reference types enabled
* has CA1508 enabled

### Expected behavior

CA1508 should not be reported in this case.

### Actual behavior

CA1508 is reported in this case.

### LINKED ISSUE #4548: CA1508 False positive for Activator.CreateInstance
### Analyzer

**Diagnostic ID**: CA1508

### Analyzer source

**Version**: [SDK 5.0.101](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug

`Activator.CreateInstance` can return `null` when used with a `Nullable<T>` type.

````
var value = Activator.CreateInstance(typeof(int?));
if (value is null) // CA1508 'value is null' is always 'false'
{
    Console.WriteLine("null");
}
````

In the previous code `value is null` is true. Whereas CA1508 thinks it is always false.

### PR DESCRIPTION
**Pull request by mavasani**
Recommend to review by commit-by-commit

Fixes #4509
Fixes #4548

### DIFF
--- a/src/Utilities/Compiler/Lightup/ITypeSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Lightup/ITypeSymbolExtensions.cs
@@ -16,9 +16,6 @@ namespace Analyzer.Utilities.Lightup
         public static NullableAnnotation NullableAnnotation(this ITypeSymbol typeSymbol)
             => s_nullableAnnotation(typeSymbol);
 
-        public static bool IsAnnotatedAsNullable(this ITypeSymbol typeSymbol)
-            => s_nullableAnnotation(typeSymbol) == Lightup.NullableAnnotation.Annotated;
-
         public static ITypeSymbol WithNullableAnnotation(this ITypeSymbol typeSymbol, NullableAnnotation nullableAnnotation)
             => s_withNullableAnnotation(typeSymbol, nullableAnnotation);
     }
--- a/src/Utilities/Compiler/Lightup/ITypeSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Lightup/ITypeSymbolExtensions.cs
@@ -16,6 +16,9 @@ namespace Analyzer.Utilities.Lightup
         public static NullableAnnotation NullableAnnotation(this ITypeSymbol typeSymbol)
             => s_nullableAnnotation(typeSymbol);
 
+        public static bool IsAnnotatedAsNullable(this ITypeSymbol typeSymbol)
+            => s_nullableAnnotation(typeSymbol) == Lightup.NullableAnnotation.Annotated;
+
         public static ITypeSymbol WithNullableAnnotation(this ITypeSymbol typeSymbol, NullableAnnotation nullableAnnotation)
             => s_withNullableAnnotation(typeSymbol, nullableAnnotation);
     }
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
@@ -14,6 +14,8 @@ using Microsoft.CodeAnalysis.Operations;
 
 namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.PointsToAnalysis
 {
+    using NullableAnnotation = Analyzer.Utilities.Lightup.NullableAnnotation;
+
     public partial class PointsToAnalysis : ForwardDataFlowAnalysis<PointsToAnalysisData, PointsToAnalysisContext, PointsToAnalysisResult, PointsToBlockAnalysisResult, PointsToAbstractValue>
     {
         /// <summary>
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs [hunk 1]
```diff
@@ -923,22 +925,38 @@ private PointsToAbstractValue VisitInvocationCommon(IOperation operation, IOpera
...
                      method.ContainingType.SpecialType != SpecialType.None ||
                      method.ReturnType is INamedTypeSymbol namedType &&
                      method.ContainingType.DerivesFromOrImplementsAnyConstructionOf(namedType.OriginalDefinition));
+                if (!isFactoryMethod)
+                {
+                    return false;
+                }
+
+                // Activator.CreateInstance can return 'null'
+                if (method.Name.Equals("CreateInstance", StringComparison.Ordinal) &&
```
Youssef1313, line 952: Is this necessary? The method is correctly annotated.

https://docs.microsoft.com/en-us/dotnet/api/system.activator.createinstance?view=net-5.0#System_Activator_CreateInstance_System_Type_

mavasani: I added it because the VB test was failing.

Youssef1313: @mavasani In that case, I think we should at least check for `CreateInstanceFrom` as well.

Or, if there is a way to find the correct annotation of C# code (in that case, the runtime) that's being consumed in VB.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
```diff
@@ -923,22 +925,38 @@ private PointsToAbstractValue VisitInvocationCommon(IOperation operation, IOpera
...
             /// Returns true if this special static factory method whose name starts with "Create", such that
+            /// method's return type is not nullable, i.e. 'type?', and
             /// method's containing type is static OR a special type OR derives from or is same as the type of the field/property/method return.
             /// For example: class SomeType { static SomeType CreateXXX(...); }
             /// </summary>
-            private static bool IsSpecialFactoryMethod(IMethodSymbol method)
+            private static bool IsSpecialFactoryMethodReturningNonNullValue(IMethodSymbol method, WellKnownTypeProvider wellKnownTypeProvider)
             {
-                return method.IsStatic &&
+                var isFactoryMethod = method.IsStatic &&
```
Evangelink, line 939: Could we split this condition? It's pretty complex to read with all the mix of AND and OR conditions.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
Evangelink: It's probably good to add a comment here because when looking at the doc we see that the API returns `object?` so it seems weird to have a special case.

Evangelink (commented):
I keep saying it but I am always amazed how "simple" the fixes are. When are you doing a public dataflow structure training ;) I'd be happy to join!

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
mavasani: Thanks, done.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/PointsToAnalysis/PointsToAnalysis.PointsToDataFlowOperationVisitor.cs
mavasani: Sure, done.

src/Utilities/Compiler/Lightup/ITypeSymbolExtensions.cs
```diff
@@ -16,6 +16,9 @@ internal static class ITypeSymbolExtensions
         public static NullableAnnotation NullableAnnotation(this ITypeSymbol typeSymbol)
             => s_nullableAnnotation(typeSymbol);
 
+        public static bool IsAnnotatedAsNullable(this ITypeSymbol typeSymbol)
```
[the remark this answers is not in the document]
mavasani: Thanks, I have removed this helper.

Evangelink (approved):

mavasani:
Thanks @Evangelink!

PR #4564, opened 2020-12-11
Prevent use of language-specific symbols in language-agnostic layers

### LINKED ISSUE #4323: Ban language-specific symbols in language-agnostic layers

Evangelink:
Fix #4323


src/Utilities/Workspaces/Workspaces.Utilities.projitems
```diff
@@ -13,8 +13,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)SolutionChangeAction.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SyntaxGeneratorExtensions.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SyntaxNodeExtensions.cs" />
-  </ItemGroup>
+  </ItemGroup>  
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersion)" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="$(MicrosoftCodeAnalysisVersion)" />
```
Evangelink, line 18: I couldn't find how to have a conditional include here that would say include the C# workspace nuget if the project name (or assembly name) contains `CSharp`, include VB for vb and fallback to the common.

@sharwell any idea?

src/Microsoft.CodeAnalysis.Analyzers/CSharp/Microsoft.CodeAnalysis.CSharp.Analyzers.csproj
```diff
@@ -9,4 +9,7 @@
   <ItemGroup>
     <InternalsVisibleTo Include="Microsoft.CodeAnalysis.Analyzers.UnitTests" />
   </ItemGroup>
+  <ItemGroup>
```
Evangelink, line 12: We used to automatically include C# and VB.NET workspace nugets, now I manually add the specific on the specific layer

src/NetAnalyzers/CSharp/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
```
Evangelink, line 2: It's considered as a new rule, I don't know if that's a problem.

src/NetAnalyzers/CSharp/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
...
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA2014 | Reliability | Warning | CSharpDoNotUseStackallocInLoopsAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2014)
```
mavasani, line 5: This is an already shipped rule in .NET 5 release. It should be manually moved to shipped file.

mavasani: Also I would have expected build to fail as `CA2014` is in the core shipped analyzer releases file, but the rule is no longer in that project. You would have to delete the existing `CA2014` entry in that file.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpDoNotUseStackallocInLoops.cs
```diff
@@ -7,12 +7,13 @@
...
 using Microsoft.CodeAnalysis.Operations;
+using Microsoft.NetCore.Analyzers;
 
-namespace Microsoft.NetCore.Analyzers.Runtime
+namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
 {
     /// <summary>CA2014: Do Not Use Stackalloc In Loops.</summary>
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    public sealed class DoNotUseStackallocInLoopsAnalyzer : DiagnosticAnalyzer
+    public sealed class CSharpDoNotUseStackallocInLoopsAnalyzer : DiagnosticAnalyzer
```
Evangelink, line 16: I don't know if there is any preference between having a language specific (syntax) analyzer being fully in the language layer or still having the base class in the common layer. I went with 1st option here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/Helpers/InsecureObjectGraphResult.cs
```diff
@@ -87,7 +86,7 @@ public string GetDisplayString()
             else if (this.InsecureAttributeTypedConstant != null)
             {
                 TypedConstant t = (TypedConstant)this.InsecureAttributeTypedConstant!;
-                return t.ToCSharpString();
+                return t.ToString(); // TODO: What to do here
```
Evangelink, line 89: @dotpaul May I have some help to refactor this bit?

src/Utilities/Workspaces/Workspaces.Utilities.projitems
sharwell: IMO the **projitems** should never have `ProjectReference`, but other people feel differently about this.

sharwell: I think the most common expectation would be **Workspaces.Utilities** references language-agnostic only, and **Workspaces.Utilities.CSharp** references the C# assemblies (i.e. split the shared projects the same way normal assemblies are split).

Evangelink: I also prefer not to have references in shared projects and to be honest with new project system I have stopped using shared project, I simply add manual reference to the files/folder I want to share. But I also understand how it makes it easier to share this common dependency.

Regarding `Workspaces.Utilities.CSharp` we currently don't have it and don't have C# specific code. I was thinking about a `Choose` just to help with avoiding the reference on each project.

If you all agree then I suggest to keep what I have done in this PR, otherwise I am open to suggestion :)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/Helpers/InsecureObjectGraphResult.cs
dotpaul: I'll take a look later this week. Last time I looked there wasn't a nice language agnostic solution, which is why it's using ToCSharpString(). Maybe we can get to the syntax node for the attribute's constructor argument somehow.

Evangelink: Yep the problem is just that it is accessing C# specific from the common layer.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Security/CSharpDataSetDataTableInIFormatterSerializableObjectGraphAnalyzer.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Security
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpDataSetDataTableInIFormatterSerializableObjectGraphAnalyzer
+        : DataSetDataTableInIFormatterSerializableObjectGraphAnalyzer
+    {
+        protected override string ToString(TypedConstant typedConstant)
```
Evangelink, line 14: I went for the abstract method but I could also pass the function to the ctor, let me know your preference.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/Helpers/InsecureObjectGraphResult.cs
```diff
@@ -78,7 +77,7 @@ public Location GetLocation()
         /// Gets the display string of <see cref="InsecureSymbol"/> or <see cref="InsecureAttribute"/>.
         /// </summary>
         /// <returns>Display string of <see cref="InsecureSymbol"/> or <see cref="InsecureAttribute"/>.</returns>
-        public string GetDisplayString()
+        public string GetDisplayString(Func<TypedConstant, string> typedConstantToString)
```
Evangelink, line 80: @dotpaul I went for a bit "hacky" solution but trying to have this class abstract was causing too many changes and it was hard to by-pass the static `GetOrCreate` method.

Evangelink:
Hum I guess that this needs to be changed to target `2.9.x`, isn't it? cc @mavasani

Evangelink:
Not sure what is going-on here, I can't repro locally.

src/NetAnalyzers/CSharp/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1 @@
-; Please do not edit this file manually, it should only be updated through code fix application.
+; Please do not edit this file manually, it should only be updated through code fix application.### New Rules
```
mavasani, line 1:
```suggestion
; Please do not edit this file manually, it should only be updated through code fix application.
```

src/NetAnalyzers/CSharp/AnalyzerReleases.Shipped.md
```diff
@@ -0,0 +1,7 @@
+## Release 5.0
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA2014 | Reliability | Warning | CSharpDoNotUseStackallocInLoopsAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2014)
```
mavasani, line 7: If you define a base type `DoNotUseStackallocInLoopsAnalyzer` and have the descriptor in this base type instead of the C# specific type, then you will not need any changes to any of the Analyzer releases file. This will also ensure that all analyzer releases and released analyzers are in a single file in the core project. Would you be able to make this change?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpDoNotUseStackallocInLoops.cs
mavasani: Let's switch to option 2 to help ensure all AnalyzerReleases are in the core layer.

src/Utilities/Workspaces/Workspaces.Utilities.projitems
mavasani: I have no preference here - I am fine with whatever is easier for you @Evangelink :-)

mavasani (commented):
Overall LGTM. I'd like to avoid any changes to AnalyzerReleases as part of the PR by avoiding moving descriptors to language specific layer. I can approve and merge once that is done.

Thanks for the work @Evangelink!

mavasani:
Tagging @jmarolf again - seems like random tests in `DiagnosticDescriptorCreationAnalyzerTests` are failing on Ubuntu Debug queue on various PRs, but not in any others. We probably should mark this test type with WindlowsOnly fact and file a tracking issue for it.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpDoNotUseStackallocInLoops.cs
Evangelink: That's ok, I will do some changes to get to this direction.

mavasani (approved):
Thanks @Evangelink!

PR #4571, opened 2020-12-14
RS1024: fix false positive on object creation

### LINKED ISSUE #4568: Constructing a `HashSet<ISymbol>` with `SymbolEqualityComparer` should not generate any warnings
### Analyzer

**Diagnostic ID**: RS1024: `Compare symbols correctly`

### Analyzer source

**NuGet Package**: [Microsoft.CodeAnalysis.Analyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Analyzers)

**Version**: 3.3.2 (Latest)

### Describe the bug

When constructing a `HashSet<ISymbol>` or a `Dictionary<ISymbol, _>` with the constructor override accepting an `IEqualityComparer`, passing in either `SymbolEqualityComparer.Default` or `SymbolEqualityComparer.IncludeNullability` still results is a warning.

### Steps To Reproduce

1. Create an new source generator project
2. Within the `Execute` metod of the generator, write the line:
`HashSet<ISymbol> childTypes = new(SymbolEqualityComparer.Default);`
3. A warning will get reported on the constructor call

See [here](https://github.com/Leemyy/SourceGenExperiments/blob/b73a6d6f25fc95bc0f4caa405db452940c95f094/ExperimentsGenerator/InheritanceGenerator.cs#L49) for an example of a (non-minimal) repro.

### Expected behavior

No warning should be reported.

### Actual behavior

A warning about improper Symbol comparison is reported.

## Additional context

- The warning gets reported both for `HashSet<T>` and `Dictionary<K, V>`
- Using a qualified constructor insead of a target-typed constuctor makes no difference
- The warning gets reported whenever the generic parameter is/derives from `ISymbol`

Evangelink:
Fix #4568

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
@@ -215,7 +215,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
                 collectionTypes.Contains(createdType.OriginalDefinition) &&
                 !createdType.TypeArguments.IsEmpty &&
                 IsSymbolType(createdType.TypeArguments[0], symbolType) &&
-                !objectCreation.Arguments.Any(arg => IsSymbolType(arg.Type, symbolEqualityComparerType)))
+                !objectCreation.Arguments.Any(arg => IsSymbolType(arg.Value, symbolEqualityComparerType)))
             {
                 context.ReportDiagnostic(objectCreation.CreateDiagnostic(CollectionRule));
             }
--- a/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
@@ -866,7 +866,9 @@ public class C
             }.RunAsync();
         }
 
-        [Fact, WorkItem(4470, "https://github.com/dotnet/roslyn-analyzers/issues/4470")]
+        [Fact]
+        [WorkItem(4470, "https://github.com/dotnet/roslyn-analyzers/issues/4470")]
+        [WorkItem(4568, "https://github.com/dotnet/roslyn-analyzers/issues/4568")]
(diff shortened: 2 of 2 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -215,7 +215,7 @@ static bool HasFirstTypeArgumentSymbolType(IMethodSymbol? method, INamedTypeSymb
                 collectionTypes.Contains(createdType.OriginalDefinition) &&
                 !createdType.TypeArguments.IsEmpty &&
                 IsSymbolType(createdType.TypeArguments[0], symbolType) &&
-                !objectCreation.Arguments.Any(arg => IsSymbolType(arg.Type, symbolEqualityComparerType)))
+                !objectCreation.Arguments.Any(arg => IsSymbolType(arg.Value, symbolEqualityComparerType)))
```
mavasani, line 218: @ryzngard can you please review?

Evangelink: It's the second similar fix I am doing which makes me question a change on Roslyn.

ryzngard: Looking at this one right after #4477

ryzngard: > It's the second similar fix I am doing which makes me question a change on Roslyn.

Just to make sure I'm interpreting correctly. You hit these issues while working in Roslyn and are fixing here?

ryzngard (approved):
It's unfortunate how easy this mistake was to miss :/. Thanks for fixing!

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
Evangelink: Sorry no, I meant that this is the second issue raised on roslyn-analyzers related to getting the type on a conversion. I am wondering if the code used to work and it stopped when we have moved to a newer version of Roslyn. My point is to say, maybe there is a regression on Roslyn side regarding these operations (Type wasn't null and now is).

PR #4575, opened 2020-12-15
Remove second descriptor because of FP rate

### LINKED ISSUE #4432: CA2248 has incorrect and undocumented diagnostic
### Analyzer

**Diagnostic ID**: [CA2248](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2248): `Provide correct enum argument to Enum.HasFlag`

### Analyzer source

**NuGet Package**: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

**Version**: v3.3.1 (Latest)

### Describe the bug

The CA2248 diagnostic incorrectly fires on uses of `HasFlag` that appears on enum types that are missing a `[Flags]` attribute, claiming that "this call will always return 'false'" when in fact it *may* return true just like any other enum, as attested to by this C# Interactive Window output:

```
> enum A { B = 1
. }
> A a = A.B;
> a.HasFlag(A.B)
true
> 
```

### Steps To Reproduce

Given this code:

```cs
            if (((__VSFINDOPTIONS)grfOptions).HasFlag(__VSFINDOPTIONS.FR_Find) ||
                ((__VSFINDOPTIONS)grfOptions).HasFlag(__VSFINDOPTIONS.FR_Replace))
```

Where the enum is defined as:

```cs
public enum __VSFINDOPTIONS
{
FR_Find = 0x100000,
FR_Replace = 0x400000
}
```

### Expected behavior

No warning. In all cases, whether the enum is attributed with a `FlagsAttribute` or not, or even if its members are defined with flags-friendly values, HasFlag will indeed return `true` if the given value is set as if it _were_ a flags enum.

### Actual behavior

The analyzer incorrectly claims:
 
> This call will always returns 'false' because the enum type '__VSFINDOPTIONS' is not marked with 'FlagsAttribute'

It does _not_ return false, so the message is dangerously misleading as someone might remove the code that is guarded by the condition that allegedly will never evaluate to true.

Evangelink:
Fix #4432

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
-    that is mostly human readable. The generation and parsing of the
-    various data types are done through the TypeConverter classes
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+    
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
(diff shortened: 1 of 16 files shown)


mavasani (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlagTests.cs
```diff
@@ -92,40 +92,6 @@ End Class
 ");
         }
 
-        [Fact]
-        public async Task CA2248_EnumTypesAreSameButNotFlag_NoDiagnostic()
```
sharwell, line 96: 💡 Consider keeping this test but updating it to show the new outcome

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
```diff
@@ -54,26 +43,18 @@ public override void Initialize(AnalysisContext context)
...
-                        invocation.Arguments[0].Value is not IConversionOperation conversion ||
-                        invocation.Instance.Type == null ||
-                        invocation.Instance.Type.TypeKind == TypeKind.TypeParameter ||
-                        conversion.Operand.Type?.TypeKind == TypeKind.TypeParameter)
-                    {
-                        return;
-                    }
-
-                    if (!invocation.Instance.Type.Equals(conversion.Operand.Type))
+                    if (invocation.TargetMethod.ContainingType.SpecialType == SpecialType.System_Enum &&
```
sharwell, line 46: 💭 I think I would have preferred to not change the flow control layout here, but I suppose what's done is done

sharwell (approved):
LGTM. If another test already covers the case I commented on, feel free to mention the test name instead of adding a new one.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/ProvideCorrectArgumentToEnumHasFlag.cs
Evangelink: I can revert, it just seemed weird to have all the conditions for the return and then one if for the actual condition.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1,17 +1,17 @@
...
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+
```
Youssef1313, line 14: What made these spaces to appear?

Evangelink: As I haven't done any manual change I guess that's VS.

Youssef1313: @Evangelink Not the first time I see this issue without manual changes.

@mavasani Any idea where to report this possible flaky bug?

AArnott: I suspect VS's resgen tool is consistently adding them, and editors configured to trim trailing whitespace remove it when the file is manually edited. 
I tend to "let VS win" in whitespace conflicts when that is the easiest thing to maintain.

Filing a suggestion bug against resgen to stop emitting the trailing whitespace in the first place would be Goodness.

AArnott (approved):

Issue #4545, opened 2020-12-09
CA1416 should Ignore diagnostics on throw statements


buyaa-n:
### Describe the improvement

When supported attributes added by SDK for the targeted platform we are seeing warnings in the statements just throwing PNSE, the warnings for the resource string used for the exception message which we couldn't separate for specific build.
   
Related to https://github.com/dotnet/runtime/issues/44231.

### Describe suggestions on how to achieve the rule

We could ignore diagnostics in throw statements as they are not platform specific operation at all

### Additional context
For example in `net5.0-Browser` targeted build `SupportedOSPlatform("Browser")` attribute will be added to the project's AssemblyInfo.cs by MSBuld, which makes everything included in the assembly are `Browser` only including resource strings
```csharp
[System.Runtime.Versioning.UnsupportedOSPlatformAttribute("browser")]
public abstract partial class Aes : System.Security.Cryptography.SymmetricAlgorithm
{
    protected Aes() { throw new System.PlatformNotSupportedException(System.SR.SystemSecurityCryptographyAlgorithms_PlatformNotSupported);  }
    public static new System.Security.Cryptography.Aes Create() { throw new System.PlatformNotSupportedException(System.SR.SystemSecurityCryptographyAlgorithms_PlatformNotSupported);  }
    public static new System.Security.Cryptography.Aes? Create(string algorithmName) { throw new System.PlatformNotSupportedException(System.SR.SystemSecurityCryptographyAlgorithms_PlatformNotSupported);  }
}
```
So for the above code we will see warnings:
```
\runtime\artifacts\obj\System.Security.Cryptography.Algorithms\net6.0-Browser-Debug\System.Security.Cryptography.Algorithms.notsupported.cs(13,118):
 error CA1416: This call site is unreachable on: 'Browser'. 'SR.SystemSecurityCryptographyAlgorithms_PlatformNotSupported' is only supported on: 'Browser'.
\runtime\artifacts\obj\System.Security.Cryptography.Algorithms\net6.0-Browser-Debug\System.Security.Cryptography.Algorithms.notsupported.cs(12,74):
 error CA1416: This call site is unreachable on: 'Browser'. 'SR.SystemSecurityCryptographyAlgorithms_PlatformNotSupported' is only supported on: 'Browser'. 
\runtime\artifacts\obj\System.Security.Cryptography.Algorithms\net6.0-Browser-Debug\System.Security.Cryptography.Algorithms.notsupported.cs(14,139):
[...]

PR #4576, opened 2020-12-15
CA1416 Do not warn for OS dependent messages for creating NotSupportedExceptions

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4545

We should ignore diagnostics in throw statements as they are not platform specific operation and causing false positives, i made sure it ignores exception messages for a block not only just throwing, but having some other statements too, i think this will be more useful for conditional throwing like:
```csharp
[SupportedOSPlatform("browser")]
public static class SR
{
    public static string Message {get; set;}
    public static void M1() { }
}

public class Test
{
    void ConditionallyThrowWithStringConstructor()
    {
        SR.M1(); // warns 
        if (!OperatingSystem.IsBrowser())
        {
            throw new PlatformNotSupportedException(SR.Message); // not warns
        }
        SR.M1() // not warns
    }
}
```
We might want to merge it into 5.0.XX cc @jeffhandley


mavasani (approved):

mavasani:
> We might want to merge it into 5.0.XX cc @jeffhandley

Sounds reasonable to me.

jeffhandley:
> > We might want to merge it into 5.0.XX cc @jeffhandley
> 
> Sounds reasonable to me.

Yep - me too

buyaa-n:
Created a [new PR](https://github.com/dotnet/roslyn-analyzers/pull/4577) as i was need to branch from 5.0.2xx, closing this one

PR #4577, opened 2020-12-15
CA1416 Do not warn for OS dependent messages for creating NotSupportedExceptions

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4545. Related to https://github.com/dotnet/roslyn-analyzers/pull/4576 raising PR for 5.0.2

@jmarolf this PR is not introducing any new warning, instead reduces false positives, so we decided to merge it to 5.0.2 
CC @jeffhandley @mavasani 

We should ignore diagnostics in throw statements as they are not platform specific operation and causing false positives, i made sure it ignores exception messages for a block not only just throwing, but having some other statements too, i think this will be more useful for conditional throwing like:
```csharp
[SupportedOSPlatform("browser")]
public static class SR
{
    public static string Message {get; set;}
    public static void M1() { }
}

public class Test
{
    void ConditionallyThrowWithStringConstructor()
    {
        SR.M1(); // warns 
        if (!OperatingSystem.IsBrowser())
        {
            throw new PlatformNotSupportedException(SR.Message); // not warns
        }
        SR.M1() // not warns
    }
}
```


mavasani (approved):

jeffhandley (approved):

buyaa-n:
Seems @jmarolf is out, i am gonna merge it as it is need for fixing some of runtime warnings

PR #4591, opened 2020-12-17
Remove workaround in AnalysisEntityFactory for a possible IOperation bug

### LINKED ISSUE #1602: Remove workaround in AnalysisEntityFactory for a possible IOperation bug
See unit test `MemberReferenceInQueryFromClause_Disposed_NoDiagnostic` (added with https://github.com/dotnet/roslyn-analyzers/pull/1600/commits/84627cea7805f4999f071d37fe21b73a69141ccf)

```
using System;
using System.Linq;

class A : IDisposable
{
    public void Dispose()
    {
    }
}

class B: IDisposable
{
    public C C { get; }
    public void Dispose()
    {
    }
}

class C
{
    public ImmutableArray<A> ArrayOfA { get; }
}

class Test
{
    void M1(ImmutableArray<B> arrayOfB)
    {
        var x = from b in arrayOfB
            from a in b.C.ArrayOfA
            select a;
        var y = new A();
        y.Dispose();
    }
}
```

IOperation for `b` in `from a in b.C.ArrayOfA` has `OperationKind.None` with no children and null type. This seems incorrect. A workaround has been added to `AnalysisEntityFactory`.

This issue tracks:
1. Filing a corresponding bug in Roslyn, after confirming it is indeed a bug.
2. Remove the workaround in `AnalysisEntityFactory` after we move to Roslyn bits with the fix.

Evangelink:
Fix #1602

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/AnalysisEntityFactory.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/AnalysisEntityFactory.cs
@@ -451,12 +451,6 @@ namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow
                 return false;
             }
 
-            // Workaround for https://github.com/dotnet/roslyn-analyzers/issues/1602
-            if (instance != null && instance.Type == null)
-            {
-                return false;
-            }
-
             PointsToAbstractValue? instanceLocation = null;
             AnalysisEntity? parent = null;
             if (instance?.Type != null)
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs
@@ -8098,10 +8098,17 @@ Class Test
 End Class");
         }
 
-        [Fact, WorkItem(1602, "https://github.com/dotnet/roslyn-analyzers/issues/1602")]
-        public async Task MemberReferenceInQueryFromClause_Disposed_NoDiagnostic()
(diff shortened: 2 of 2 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/AnalysisEntityFactory.cs
```diff
@@ -451,12 +451,6 @@ private AnalysisEntity GetOrCreateForFlowCapture(CaptureId captureId, ITypeSymbo
                 return false;
             }
 
-            // Workaround for https://github.com/dotnet/roslyn-analyzers/issues/1602
```
mavasani, line 454: @Evangelink Can you try to extend the failing unit test mentioned in this issue to run with all available PointsToAnalysisKind option values? Basically the below test should be converted: https://github.com/dotnet/roslyn-analyzers/blob/405247d4063733ef5465361001ad98974d9ba82b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs#L8102
so it has `InlineData` as: https://github.com/dotnet/roslyn-analyzers/blob/405247d4063733ef5465361001ad98974d9ba82b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScopeTests.cs#L8703-L8706

mavasani: I think the workaround is still needed for `PointsToAnalysisKind.Complete` non-default option value.

Evangelink: It worked for all the cases.

mavasani (approved):
Thanks @Evangelink!

PR #4594, opened 2020-12-17
Remove stale TODO comment

### LINKED ISSUE #2009: IMethodSymbolExtensions.GetTopmostOperationBlock() gets one more operation than expected
#### Branch

master

#### File Name

\src\Utilities\Extensions\IOperationExtensions.cs, \src\Utilities\Extensions\IMethodSymbolExtensions.cs

#### Problematic Code

ImmutableArray<IOperation> GetOperations(this ImmutableArray<IOperation> blockOperations) in IOperationExtensions.cs, IBlockOperation GetTopmostOperationBlock(this IMethodSymbol method, Compilation compilation, CancellationToken cancellationToken = default(CancellationToken)) in IMethodSymbolExtensions.cs

#### Test Code
```
[Fact]
        public void TestDelegateCreationDiagnostic()
        {
            VerifyBasic(@"
Imports System.Net
Imports System.Net.Security
Imports System.Security.Cryptography.X509Certificates

Namespace TestNamespace
    Class TestClass
        Sub TestMethod()
            System.Net.ServicePointManager.ServerCertificateValidationCallback = New System.Net.Security.RemoteCertificateValidationCallback(AddressOf AcceptAllCertifications)
        End Sub

        Function AcceptAllCertifications(ByVal sender As Object, ByVal certification As System.Security.Cryptography.X509Certificates.X509Certificate, ByVal chain As System.Security.Cryptography.X509Certificates.X509Chain, ByVal sslPolicyErrors As System.Net.Security.SslPolicyErrors) As Boolean
            Return True
        End Function
    End Class
End Namespace",
            GetBasicResultAt(9, 82, DoNotDisableCertificateValidation.Rule, "MethodReference"));
        }
```

#### Expected behavior

When use GetTopmostOperationBlock() to deal with the test code, it should get some children operations and the last 2 children operation should be implicit and whose kinds should be "Label" and "Return", then pass the result to GetOperations() to remove the last 2 operations.

#### Actual behavior

GetTopmostOperationBlock gets 5 children operations, and whose kinds are "Return", "Literal", "Label", "Return" and "LocalReference". It seems there's one more implicit operation "LocalReference" than expected.

Evangelink:
Fix #2009

Testing on `master`, it seems we now have 4 children: the variable, the return and the 2 implicit operations `Label` and `Return` mentioned in the `Expected behavior` section.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotDisableCertificateValidation.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/DoNotDisableCertificateValidation.cs
@@ -112,8 +112,7 @@ namespace Microsoft.NetCore.Analyzers.Security
                                             return;
                                         }
 
-                                        // TODO(LINCHE): This is an issue tracked by #2009. We filter extraneous based on IsImplicit.
-                                        var targetOperations = ImmutableArray.ToImmutableArray(blockOperation.Descendants()).WithoutFullyImplicitOperations();
+                                        var targetOperations = blockOperation.Descendants().ToImmutableArray().WithoutFullyImplicitOperations();
                                         alwaysReturnTrue = AlwaysReturnTrue(targetOperations);
                                         break;
                                 }


mavasani:
@Evangelink This needs to target 2.9.x branch. Tagging @dotpaul 

Also tagging @jmarolf as it seems Ubuntu Debug is showing intermittent test failures.

Evangelink:
Why does it need to be so complex. Closng this PR and I will open a new one targeting the right branc.

mavasani:
> Why does it need to be so complex.

Because large number of internal teams in Microsoft are still on VS2017 and they can only consume 2.9.x packages, so we need to continue maintaining that branch.

Evangelink:
Sorry wasn't talking about the need to support the branch 2.9.x, more about the github ui that was applying all commits when I change the target branch rather than only trying to cherry-pick commits.

PR #4615, opened 2020-12-20
Consolidate tests

Youssef1313:
@sharwell This is based on your suggestion. I'll try to do that to more tests. (**Review commit-by-commit**)

Is there a better way to organize the commits to ease review? The diff on GitHub is a bit hard to follow, especially if I'm going to do that for more and more tests.


Youssef1313:
@sharwell What's the difference between `VerifyCS.VerifyCodeFixAsync(code, code);` and `VerifyCS.VerifyCodeFixAsync(code);` (if any)

sharwell:
> `VerifyCS.VerifyCodeFixAsync(code)`

😕 I didn't think a single-argument form existed for this method.

> `VerifyCS.VerifyCodeFixAsync(code, code)`

This verifies two things:

1. The diagnostics match the values from markup (often no diagnostic)
2. The `CodeFixProvider` used in the `VerifyCS` alias does not offer any fixes for the diagnostics

The second verification is a slightly stronger statement than just saying no changes are applies, since it's possible for some code fixes to offer a fix which then takes no action.

Youssef1313:
> 😕 I didn't think a single-argument form existed for this method.

@sharwell Ah sorry, I meant `VerifyCS.VerifyAnalyzerAsync(code)` vs `VerifyCS.VerifyCodeFixAsync(code, code)`. But it's clear now thanks to your explanation.

sharwell:
> So I think I should always (or in most cases) prefer VerifyCodeFixAsync(code, code) rather than VerifyAnalyzerAsync(code) due to the point in your previous comment. Is that right?

Yes, that's correct.

Evangelink:
@Youssef1313 Could you fix the conflict?

Evangelink:
@sharwell Could you review please?

sharwell (approved):
I reordered the methods so the final diff was minimized

Evangelink:
@Youssef1313 could you please run a pack command?

Youssef1313:
It shouldn't be needed. I'm going to close and re-open for a new build.

mavasani (approved):

PR #4621, opened 2020-12-21
Update CA1820 to not report on expression trees

### LINKED ISSUE #1508: CA1820 incorrectly fires in LINQ expressions
#### Analyzer package

Microsoft Code Analysis, version 2.6.0.6241303

#### Analyzer

CA1820

#### Repro steps

Consider the following program:

    using System.Linq;

    class C
    {
        void M(IQueryable<string> strings)
        {
            var q = from s in strings
                    where s == ""
                    select s;
        }
    }

#### Expected behavior

CA1820 is not raised. Indeed, the comparison to `""` happens in a LINQ expression.

#### Actual behavior

CA1820 is raised:

    Test for empty strings using 'string.Length' property or 'string.IsNullOrEmpty' method instead of an Equality check.	
    Comparing strings by using the String.Length property or the String.IsNullOrEmpty method is significantly faster than using Equals.

Given that the comparison happens in a LINQ expression, there is no way to tell if using string.Length will be faster or not. In fact, it is very much possible that the LINQ provider will not be able to translate a comparison to String.Length.

Note that [U2U1100](https://github.com/u2uconsult/codeanalyzers/blob/master/docs/U2U1100.md) from the [U2U Consult Performance Analyzers for C# 7](https://marketplace.visualstudio.com/items?itemName=vs-publisher-363830.U2UConsultPerformanceCodeAnalyzersforC7) does not have this problem.

Evangelink:
Fix #1508

### DIFF
--- a/src/Utilities/Compiler/Extensions/IOperationExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IOperationExtensions.cs
@@ -508,10 +508,25 @@ namespace Analyzer.Utilities.Extensions
 
         private static readonly ImmutableArray<OperationKind> s_LambdaAndLocalFunctionKinds =
             ImmutableArray.Create(OperationKind.AnonymousFunction, OperationKind.LocalFunction);
+
         public static bool IsWithinLambdaOrLocalFunction(this IOperation operation, [NotNullWhen(true)] out IOperation? containingLambdaOrLocalFunctionOperation)
         {
             containingLambdaOrLocalFunctionOperation = operation.GetAncestor(s_LambdaAndLocalFunctionKinds);
             return containingLambdaOrLocalFunctionOperation != null;
+        }
+
+        public static bool IsWithinExpressionTree(this IOperation operation,
+            [NotNullWhen(true)] INamedTypeSymbol? linqExpressionTreeType)
+        {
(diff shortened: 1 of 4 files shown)


src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -508,12 +508,27 @@ void ProcessLocalOrParameter(ISymbol symbol)
...
             ImmutableArray.Create(OperationKind.AnonymousFunction, OperationKind.LocalFunction);
+
         public static bool IsWithinLambdaOrLocalFunction(this IOperation operation, [NotNullWhen(true)] out IOperation? containingLambdaOrLocalFunctionOperation)
         {
             containingLambdaOrLocalFunctionOperation = operation.GetAncestor(s_LambdaAndLocalFunctionKinds);
             return containingLambdaOrLocalFunctionOperation != null;
         }
 
+        public static bool IsWithinExpressionTree(this IOperation operation,
+            [NotNullWhen(true)] INamedTypeSymbol? linqExpressionTreeType)
```
Evangelink, line 519: I have seen two different approaches in the codebase and I am not sure which one is the preferred one:
1. Take the named type
2. Take the compilation and retrieve the named type from within the helper

mavasani (approved):

Issue #4268, opened 2020-10-01
Never use the two-argument form of `WithLocation`.


Youssef1313:
As an aside, never use the two-argument form of `WithLocation`. Either use `WithLocation(int)` or use `WithSpan` to specify full spans.

_Originally posted by @sharwell in https://github.com/dotnet/roslyn-analyzers/pull/4267#discussion_r498492605_

Evangelink:
Not sure whether there is any interest in keeping this ticket open. We usually try to suggest the new API during reviews.

Shall we ban the use of the other overload and add suppressions for all existing methods?

cc @mavasani

sharwell:
> Shall we ban the use of the other overload and add suppressions for all existing methods?

That could work as a starting point.

Seems like we could keep this issue open for someone to eventually fix the existing cases and remove the suppressions.

Evangelink:
One case which isn't easy with the inline markup is when we test user-option and for some of them we have the error and for others we don't. @sharwell Do you think it could be worth to update the test sdk to allow some kind of `.EnableWhen(Func<bool>)` which would allow to always define the markup syntax but control whether or not the issue is reported?

sharwell:
The `{|#0:...|}` markup works for that case. If there is no corresponding explicit diagnostic with `WithLocation(0)`, then location `0` in markup is simply ignored.

Evangelink:
I have had test failures saying it cannot find any diagnostic for the markup ID XXX

sharwell:
> I have had test failures saying it cannot find any diagnostic for the markup ID XXX

That would occur for the syntax `{|ID:...|}` and `[|...|]`, but would not occur for `{|#0:...|}`.

Evangelink:
I am pretty sure I saw this behavior but I can't repro it so sorry for the comments :s

PR #4631, opened 2020-12-22
Ban non-markupKey WithLocation overloads

Evangelink:
Relates to #4268. Maybe we want to close the associated ticket and create a new one to get rid of the suppressions.

Shall we also ban the use of `WithSpan` methods?

I have experienced a problem with the fix-all solution: I did a full compilation of the solution, clicked the suppress all in solution and after 10min it was still "loading", I have clicked "Cancel" and tried again it was the same result. Going for project by project solution only takes some seconds to load-up. I don't know if there is a problem with roslyn/VS, I think my machine is pretty decent so I would rule that out.


sharwell:
> Shall we also ban the use of `WithSpan` methods?

No. The goal of this change is ensuring diagnostic locations are validated for both the start and end locations. The `WithLocation(int)` overload is easier to write and maintain, but the `WithSpan` forms provide the same level of assurance.

Evangelink:
I wasn't sure if we wanted to ensure "only" precise location of issues or also more readable issue locations (the markups are easier to read and understand than the `WithSpan`.

sharwell (approved):

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlersTests.cs
```diff
@@ -25,8 +25,12 @@ public class Program
     public void Handler1(object sender, EventArgs args) {}
     protected void Handler2(object sender, EventArgs args) {}
 }",
+#pragma warning disable RS0030 // Do not used banned APIs
                 VerifyCS.Diagnostic().WithLocation(6, 17).WithArguments("Handler1"),
+#pragma warning restore RS0030 // Do not used banned APIs
+#pragma warning disable RS0030 // Do not used banned APIs
```
Youssef1313, line 31: nit: redundant.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlersTests.cs
```diff
@@ -38,8 +42,12 @@ End Sub
     Protected Sub Handler2(ByVal sender As Object, ByVal args As EventArgs)
     End Sub
 End Class",
+#pragma warning disable RS0030 // Do not used banned APIs
                 VerifyVB.Diagnostic().WithLocation(5, 16).WithArguments("Handler1"),
+#pragma warning restore RS0030 // Do not used banned APIs
+#pragma warning disable RS0030 // Do not used banned APIs
```
Youssef1313, line 48: nit: redundant

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -2151,10 +2159,14 @@ internal static class StaticClass
     }
 }";
 
+#pragma warning disable RS0030 // Do not used banned APIs
             await VerifyAnalyzerAsyncCs(source, VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.UnsupportedCsReachable).
                 WithLocation(11, 9).WithArguments("C.StaticClass.LinuxMethod()", "'linux'", "'Linux'"),
+#pragma warning restore RS0030 // Do not used banned APIs
+#pragma warning disable RS0030 // Do not used banned APIs
```
Youssef1313, line 2166: nit: redundant

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/ReviewVisibleEventHandlersTests.cs
sharwell: I saw these, but figured they are fine since the cleanup pass will fix one file at a time.

PR #4635, opened 2020-12-26
Exclude GetPinnableReference from CA1024 analysis

### LINKED ISSUE #4634: CA1024: False positive for special method GetPinnableReference
### Analyzer

**Diagnostic ID**: [CA1024](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1024): Use properties where appropriate

### Analyzer source

**Version**: 5.0.1 (Latest)

### Describe the bug

Rule CA1024 doesn't treat `GetPinnableReference` (used for allowing the type to be `fixed`) as a special method name, like it does with `GetHashCode` or `GetEnumerator` and instead triggers a warning on such methods.

### Steps To Reproduce

1. Add some code looking something like this:
```cs
public unsafe class Foo {
    private int _bar;
    public ref int GetPinnableReference() { return ref _bar; }
    public void Baz() { fixed (int* bar = this) { } }
}
```
2. Enable the CA1024 diagnostic
3. Analyze the code
3. Note the warning regarding `GetPinnableReference` being better off as a property.

Youssef1313:
Fixes #4634

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
@@ -77,7 +77,8 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     if (methodSymbol.IsGenericMethod ||
                         methodSymbol.IsVirtual ||
                         methodSymbol.IsOverride ||
-                        methodSymbol.Name is GetHashCodeName or GetEnumeratorName or GetPinnableReferenceName ||
+                        methodSymbol.Name is GetHashCodeName or GetEnumeratorName ||
+                        methodSymbol is { Name: GetPinnableReferenceName, ReturnsByRef: true } ||
                         methodSymbol.ContainingType.GetMembers(methodSymbol.Name).Length > 1 ||
                         taskTypes.Contains(methodSymbol.ReturnType.OriginalDefinition) ||
                         methodSymbol.IsImplementationOfAnyImplicitInterfaceMember())
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
@@ -78,7 +78,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                         methodSymbol.IsVirtual ||
                         methodSymbol.IsOverride ||
                         methodSymbol.Name is GetHashCodeName or GetEnumeratorName ||
-                        methodSymbol is { Name: GetPinnableReferenceName, ReturnsByRef: true } ||
+                        (methodSymbol.Name == GetPinnableReferenceName && (methodSymbol.ReturnsByRef || methodSymbol.ReturnsByRefReadonly)) ||
                         methodSymbol.ContainingType.GetMembers(methodSymbol.Name).Length > 1 ||
                         taskTypes.Contains(methodSymbol.ReturnType.OriginalDefinition) ||
                         methodSymbol.IsImplementationOfAnyImplicitInterfaceMember())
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
```diff
@@ -117,6 +117,11 @@ public IEnumerator GetEnumerator()
         return null;
     }
 
+    public int GetPinnableReference() // Ideally, the method return type must be `ref`.
```
Youssef1313, line 120: Let me know if you want this to report a diagnostic, and only ref-returns doesn't.

sharwell: Does pattern-based `fixed` allow `GetPinnableReference` to return a pointer type?

I believe the exclusion should only apply to methods that match the pattern by filtering to `ref` and `ref readonly` returns.

Evangelink: According to the doc it's even more restrictive than this: https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/fixed-statement

sharwell: The `ref` check seems easy and catches most cases. I don't think we need to check the unmanaged type restriction.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
```diff
@@ -117,6 +117,11 @@ public IEnumerator GetEnumerator()
         return null;
     }
 
+    public ref string GetPinnableReference() // If the method isn't ref-returning, there will be a diagnostic.
```
Evangelink, line 120: Could you add a test for `readonly ref` (or the other way around, I can't recall the expected order). Also, could you add a test for a managed type.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
```diff
@@ -76,8 +77,8 @@ public override void Initialize(AnalysisContext analysisContext)
                     if (methodSymbol.IsGenericMethod ||
                         methodSymbol.IsVirtual ||
                         methodSymbol.IsOverride ||
-                        methodSymbol.Name == GetHashCodeName ||
-                        methodSymbol.Name == GetEnumeratorName ||
+                        methodSymbol.Name is GetHashCodeName or GetEnumeratorName ||
+                        (methodSymbol.Name == GetPinnableReferenceName && (methodSymbol.ReturnsByRef || methodSymbol.ReturnsByRefReadonly)) ||
```
Youssef1313, line 81: Is there a better way to write this condition using the new pattern matching enhancement?

Evangelink: It's non C# 9 but you could simply extract the code in a function `IsPinnableReferenceMethod`, also this could maybe be part of the helpers, I suspect some other rules will have to use it.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriateTests.cs
Evangelink: I think you missed the managed type case.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UsePropertiesWhereAppropriate.cs
mavasani: Do we need to special case ref and ref readonly checks only to methods named `GetPinnableReference` or should we do it regardless of the method name? @sharwell for thoughts.

mavasani: Tag @sharwell again for the last question. Otherwise, LGTM.

mavasani: Going to keep the current restrictive check - if needed, we can adjust it in future.

mavasani (approved):

PR #4641, opened 2020-12-31
Use and enforce expression body for properties and indexers

Evangelink:


src/Tools/ReleaseNotesUtil/RuleInfo.cs
```diff
@@ -53,25 +53,14 @@ public RuleInfo()
...
-                    (!string.IsNullOrWhiteSpace(this.Description)
-                        ? this.Description
-                        : this.MessageFormat)
-                    ?? string.Empty;
-            }
-        }
+            => (!string.IsNullOrWhiteSpace(this.Description)
+                    ? this.Description
+                    : this.MessageFormat)
+                ?? string.Empty;
```
sharwell, line 64: ⬆️ This change is the reason I do not like this rule.

sharwell (commented):
Not going to block this if someone else wants it, but not a rule I would turn on for my own repositories in its current form. 😄

src/Tools/ReleaseNotesUtil/RuleInfo.cs
Evangelink: Is there any ticket on Roslyn to enable some tweak of the rule? We could think of some option that would not suggest expression-body if the content is a ternary for example or if it is a "complex" expression (to be defined).

Evangelink:
@sharwell I am fine if we close this one out. I am just trying to apply some common styles/pattern across the repro, it would be easier if we could have some chat from time to time on the directions we(you) want to take.

src/Utilities/Compiler/CodeMetrics/ComputationalComplexityMetrics.cs
```diff
@@ -403,43 +403,24 @@ public ComputationalComplexityMetrics Union(ComputationalComplexityMetrics other
...
-        }
+            => _symbolUsageCounts + _constantUsageCounts;
 
-        public long Vocabulary               //n
-        {
-            // n = n1 + n2
-            get { return DistinctOperators + DistinctOperands; }
-        }
+        public long Vocabulary               // n = n1 + n2
+            => DistinctOperators + DistinctOperands;
```
Youssef1313, line 418: There were 2 comments, one of them was removed.
Luckily, I feel the one that was removed is really extra and unneeded, but the codefix should preserve trivia correctly.

Evangelink: I did the removal manually because I wasn't sure how to merge the 2 comments. I have also created an issue on roslyn because the code-fix breaks trivia.s

Evangelink:
@mavasani Let me know if you feel like we should use expression-bodies or drop this PR.

mavasani:
> Let me know if you feel like we should use expression-bodies or drop this PR.

These analyzers are supported on build in the CodeStyle NuGet and the .NET 5.0 SDK, so I think your PR is good dogfooding exercise for enforcing them on build. Let's go ahead with the PR.

.editorconfig
```diff
@@ -94,9 +94,9 @@ csharp_style_expression_bodied_local_functions = false:none
 csharp_style_expression_bodied_lambdas = true:none
 
 # Prefer property-like constructs to have an expression-body
-csharp_style_expression_bodied_properties = true:none
-csharp_style_expression_bodied_indexers = true:none
-csharp_style_expression_bodied_accessors = true:none
+csharp_style_expression_bodied_properties = true:warning
+csharp_style_expression_bodied_indexers = true:warning
+csharp_style_expression_bodied_accessors = true:warning
```
mavasani, line 99: All style rules are defaulted to warnings, so this is redundant.
```suggestion
```

https://github.com/dotnet/roslyn-analyzers/blob/5102a0d89ef178b0ca741634ee006597bf5df2ed/.editorconfig#L188-L189

.editorconfig
```diff
@@ -94,9 +94,9 @@ csharp_style_expression_bodied_local_functions = false:none
 csharp_style_expression_bodied_lambdas = true:none
 
 # Prefer property-like constructs to have an expression-body
```
mavasani, line 96:
```suggestion
```

src/Tools/ReleaseNotesUtil/RuleInfo.cs
mavasani: I don't understand how the original code is more readable then the new code though. I think we need to do either of 2 things here:

1. Rewrite the code in block body form breaking down into multiple statements.
2. Use the expression body form suggested in this PR.

I am fine with either approach. I think the form in the original code is probably the least preferred form.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DictionaryAnalysisData.cs
```diff
@@ -56,15 +56,10 @@ public ICollection<TKey> Keys
...
-        {
-            get
-            {
-                // "Values" might be accessed during dispose.
-                //Debug.Assert(!IsDisposed);
-                return _coreAnalysisData.Values;
-            }
-        }
+        public ICollection<TValue> Values =>
+            // "Values" might be accessed during dispose.
```
mavasani, line 60: I think we may want to consider loosening the analyzer here and not report an issue when there are comments on a separate line in the body. I don't think users would prefer expression body with comments in between the `=>` and the expression. @Evangelink can you file a Roslyn issue for this?

mavasani (approved):
Overall LGTM. Lets try this out for a bit, and if the analyzer gets too annoying we can downgrade it to an IDE suggestion in future.

@Evangelink I'll wait for you to respond to comments before merge.

src/Utilities/Compiler/CodeMetrics/ComputationalComplexityMetrics.cs
Evangelink: See https://github.com/dotnet/roslyn/issues/50181

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DictionaryAnalysisData.cs
Evangelink: Ticket created https://github.com/dotnet/roslyn/issues/50229

src/Tools/ReleaseNotesUtil/RuleInfo.cs
Evangelink: Let me know your preferences.

mavasani (approved):

Evangelink:
> @Evangelink I'll wait for you to respond to comments before merge.

I think I answered them.

src/Tools/ReleaseNotesUtil/RuleInfo.cs
CyrusNajmabadi: We have a value of `when_on_single_line` for this feature.

PR #4644, opened 2020-12-31
Fix various nullability warnings

Evangelink:


sharwell (changes requested):
Marking request changes _only_ so I have more time to review... will follow up shortly

sharwell:
@Evangelink note that the current recommendation for enumerator implementations is to write nullability annotations under the condition that the enumerator is used correctly.

From https://github.com/dotnet/runtime/blob/aebd598664e7d931eee19eb1779c0ace5d1c3600/docs/coding-guidelines/api-guidelines/nullability.md:

> **DO NOT** be concerned with the validity of an enumerator's Current annotations when used before MoveNext has been called or after MoveNext has returned false. As with Dispose, such use is considered invalid, and we don't want to harm the correct consumption of Current by catering to incorrect consumption.

src/Utilities/Compiler/SmallDictionary.cs
```diff
@@ -541,7 +541,10 @@ public Enumerator(SmallDictionary<K, V> dict)
                     }
                 }
 
-                public K Current => _current.Key;
+                public K Current
+                    => _current is null
+                        ? throw new InvalidOperationException()
+                        : _current.Key;
```
sharwell, line 547: ❗ The previous code is the expected form. The result of the `Current` property is unspecified when does not follow a `MoveNext()` call that returned true.

This applies to all three cases from the PR.

src/Utilities/Compiler/SmallDictionary.cs
```diff
@@ -571,7 +574,7 @@ private void PushIfNotNull(AvlNode? child)
                 {
                     if (child != null)
                     {
-                        _stack.Push(child);
+                        _stack?.Push(child);
```
sharwell, line 577: 💡 Rather than use `?.`, you can move this to be a local function of the `MoveNext()` method, and the NRT analysis will see that it is only called when `_stack` is non-null.

This applies to all three cases from the PR.

sharwell (changes requested):

Evangelink:
Thanks for the guide, I wasn't aware of it. I will take care of the changes you suggested.

src/Utilities/Compiler/SmallDictionary.cs
Evangelink: If I revert this part I end up with a `CS8602` except if I do revert the nullability of the field in which case it's back to square 1 with `CS8618`.

src/Utilities/Compiler/SmallDictionary.cs
Evangelink: I have done the local function but with `struct` I can't access fields from local func so I had to pass it along.

src/Utilities/Compiler/SmallDictionary.cs
Evangelink: @sharwell What's your preference to fix the error on the property? Shall I just drop this PR?

mavasani:
@sharwell Can you please review latest version?
@Evangelink Can you please fix failures?

src/Utilities/Compiler/SmallDictionary.cs
sharwell: Either of the following is acceptable:

1. Make `_current` non-nullable, which is an accurate representation of its state after `MoveNext()` returns true
2. Return `_current!.Key`

sharwell:
The unresolved failures were due to an open question for me, which I have now answered. Sorry for the delay!

Evangelink:
@mavasani @sharwell I believe the PR is ready

Evangelink:
Ping @sharwell @mavasani

Evangelink:
@mavasani @sharwell are you still interesting in moving this PR forward?

Evangelink:
Ping again @mavasani @sharwell

Evangelink:
Ping @mavasani

mavasani (approved):

Evangelink:
Thank you @mavasani

PR #4647, opened 2020-12-31
Add analyzer/fixer for informal IDE brace style with statements.

CyrusNajmabadi:
Adds a new analyzer/fixer for the informal IDE style that we shouldn't bunch of statements too closely.  Specifically, if we have a multiline block, and that block is followed by a statement, then there should be a blank line between the block and the statement.  In other words, we prefer:

```c#
while (...)
{
   // ...
}

if (...)
{
}

return;
```

over

```c#
while (...)
{
   // ...
}
if (...)
{
}
return;
```


CyrusNajmabadi:
Could someone remind me how to run this against Roslyn itself?  I want to see if i missed any cases.  Thanks!

sharwell:
Here's the test cases for this rule:

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/LayoutRules/SA1513CSharp7UnitTests.cs

CyrusNajmabadi:
> Here's the test cases for this rule:

Added the cases.  There are two differences between SA1513 and my code currently.  Specifically:

1.  My code (currently) doesn't touch curlies outside of block-statements.  I feel like code to adjust how *members* are spaced should be different, and out of scope of this analyzer/fixer.
2. My code does not touch a block-statement followed by a `case`-clause.  This is up in the air for me if we would insist this be fixed or not.  

Note that in both of these cases we simply do not care what the code is.  So that leaves us in the state we are in today.  IMO, we can decide to be more stringent about the coding here later if we want to.  

Thoughts @sharwell ?

CyrusNajmabadi:
@sharwell https://github.com/dotnet/roslyn-analyzers/pull/4647#issuecomment-753149235 ^ plz :)

CyrusNajmabadi:
Can see results of this here: https://github.com/dotnet/roslyn/pull/50201

sharwell:
@CyrusNajmabadi thanks for detailing the differences. No objections.

CyrusNajmabadi:
@sharwell  i have no idea why this particular test is failing :)

CyrusNajmabadi:
Specifically:  

```
Microsoft.CodeQuality.Analyzers.QualityGuidelines.UnitTests.DoNotInitializeUnnecessarilyTests.Diagnostics_InitializerRemoved

Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessageException : Context: Fix all in solution\r\nExpected '2' iterations but found '1' iterations.\r\nAssert.Equal() Failure\r\nExpected: 2\r\nActual:   1
```

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsCodeFixProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+            var document = context.Document;
+            var diagnostic = context.Diagnostics.First();
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    RoslynDiagnosticsAnalyzersResources.Add_blank_line_after_block,
+                    c => UpdateDocumentAsync(document, diagnostic, c),
+                    RoslynDiagnosticsAnalyzersResources.Add_blank_line_after_block),
+                context.Diagnostics);
+            return Task.CompletedTask;
+
```
CyrusNajmabadi, line 40:
```suggestion
```

Evangelink:
I am really surprised that https://github.com/dotnet/roslyn/issues/50195 got closed and this analyzer pops-up as a Roslyn specific analyzer.

src/Roslyn.Diagnostics.Analyzers/CSharp/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
```
Youssef1313, line 3: @Evangelink Didn't you fix this before (the empty line before the table)? or I'm mis-remembering?

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsCodeFixProvider.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+#nullable disable warnings
```
Youssef1313, line 3: Why not enable nullable warnings here?

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// {
+    /// }
+    /// 
+    /// NextStatement();
+    /// </code>
+    /// 
+    /// Specifically, all blocks followed by another statement must have a blank line between them.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer : DiagnosticAnalyzer
```
Youssef1313, line 34: In roslyn-analyzers, it was preferred that all analyzers (even language-specific) be implemented as abstract classes in the Core layer to exist in the proper tracking analyzers file.

Tagging @mavasani to confirm.

CyrusNajmabadi:
> I am really surprised that dotnet/roslyn#50195 got closed and this analyzer pops-up as a Roslyn specific analyzer.

Why?  As I said in that issue:

> Closing out. Such an analyzer can already be written today. There is no need for it to be in roslyn.

This pr follows that opinion. I don't think this should be part of roslyn. It can just be something written externally.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
Youssef1313, line 34: I think it shouldn't implement DiagnosticAnalyzer directly, but AbstractBuiltInCodeStyleDiagnosticAnalyzer instead.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        private void AnalyzeBlock(SyntaxNodeAnalysisContext context)
+        {
+            var block = (BlockSyntax)context.Node;
+
+            // Don't examine broken blocks.
+            var closeBrace = block.CloseBraceToken;
+            if (closeBrace.IsMissing)
+                return;
```
Youssef1313, line 69: Do you prefer braces for single-line if statements?

CyrusNajmabadi: I do not. I prefer the not be included in that case

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
CyrusNajmabadi: That type is not in this repo :-)

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsCodeFixProvider.cs
CyrusNajmabadi: Copied from another file. I tried enabling this and it was pretty annoying.

src/Roslyn.Diagnostics.Analyzers/UnitTests/BlankLines/BlankLinesBetweenStatementsTests_CSharp.cs
```diff
@@ -0,0 +1,1006 @@
...
+{
+    int X { get; }
+    int Y { get; }
+}";
+
+            await new Verify.Test()
+            {
+                TestCode = code,
+                FixedCode = code,
+            }.RunAsync();
```
Youssef1313, line 31: await Verify.VerifyCodeFixAsync(code, code); is shorter.

Evangelink:
I don't want to pollute your PR too much so I am just going to ask one last question. Aren't all rules described here: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#c-formatting-rules part of the `roslyn` repo? If they are what makes the new line after brace different from all the other new line rules?

src/Roslyn.Diagnostics.Analyzers/CSharp/AnalyzerReleases.Unshipped.md
Evangelink: I don't recall doing the fix :)

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsCodeFixProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+        public override ImmutableArray<string> FixableDiagnosticIds
+            => ImmutableArray.Create(RoslynDiagnosticIds.BlankLinesBetweenStatementsRuleId);
+
+        public override FixAllProvider GetFixAllProvider()
+            => WellKnownFixAllProviders.BatchFixer;
+
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var document = context.Document;
+            var diagnostic = context.Diagnostics.First();
```
Evangelink, line 32: We have some code-fixes doing a loop over the `Diagnostics` and some other doing `.First()` what is the right way?

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
Evangelink: Yep that was the decision, see https://github.com/dotnet/roslyn-analyzers/pull/4564#discussion_r545292242

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsCodeFixProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers.BlankLines
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class CSharpBlankLinesBetweenStatementsCodeFixProvider : CodeFixProvider
```
Evangelink, line 19: @mavasani I don't know if the decision to have a base-class in the core layer even for language specific analyzers also applies to code-fix.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSyntaxNodeAction(AnalyzeBlock, SyntaxKind.Block);
+        }
+
+        private void AnalyzeBlock(SyntaxNodeAnalysisContext context)
```
Evangelink, line 62:
```suggestion
        private static void AnalyzeBlock(SyntaxNodeAnalysisContext context)
```

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+            var closeBrace = block.CloseBraceToken;
+            if (closeBrace.IsMissing)
+                return;
+
+            // If the close brace itself doesn't have a newline, then ignore this.  This is a case of series of
+            // statements on the same line.
+            if (!closeBrace.TrailingTrivia.Any())
+                return;
+
+            if (closeBrace.TrailingTrivia.Last().Kind() != SyntaxKind.EndOfLineTrivia)
```
Evangelink, line 76: [question] Isn't it better to use `IsKind` rather than `Kind()` and comparison?

CyrusNajmabadi:
> If they are what makes the new line after brace different from all the other new line rules?

I believe the set in roslyn is the core set we want to support (esp. for legacy reasons). I don't want to add new ad-hoc rules like this. :-)

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+                if (trivia.Kind() == SyntaxKind.WhitespaceTrivia)
+                    continue;
+
+                // got something that wasn't whitespace.  Bail out as we don't want to place any restrictions on this code.
+                return;
+            }
+
+            context.ReportDiagnostic(Diagnostic.Create(
+                Rule,
+                closeBrace.GetLocation()));
```
Evangelink, line 105:
```suggestion
            context.ReportDiagnostic(closeBrace.CreateDiagnostic(Rule));
```

src/Roslyn.Diagnostics.Analyzers/UnitTests/BlankLines/BlankLinesBetweenStatementsTests_CSharp.cs
```diff
@@ -0,0 +1,1006 @@
...
+        {
+            var code =
+@"
+class C
+{
+    int X { get; }
+    int Y { get; }
+}";
+
+            await new Verify.Test()
```
Evangelink, line 27: I would not test the `FixedCode` when we don't expect a diagnostic to be raised. If you go with this change, it applies to some other tests.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
CyrusNajmabadi: i'm fine with that.  i was plannign on doing a similar analyzer/fixer for VB as well.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
CyrusNajmabadi: Done.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
CyrusNajmabadi: I don't see either as being better or more appropriate.  it's not realistically going to matter.  and, if it ever does, i'm fine wiht an analyzer/fixer ensuring consistent use of hte right method.

Youssef1313:
You'll need to run `msbuild /t:pack` at repo root to update auto-generated files.

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsDiagnosticAnalyzer.cs
CyrusNajmabadi: Nice.  thanks :)

src/Roslyn.Diagnostics.Analyzers/UnitTests/BlankLines/BlankLinesBetweenStatementsTests_CSharp.cs
CyrusNajmabadi: I disagree... i'm trying to doc expected behavior here about what is/isn't controlled by this analyzer.  this is very normal IMO for full testing of these types.

CyrusNajmabadi:
> You'll need to run msbuild /t:pack at repo root to update auto-generated files.

Thanks!

src/Roslyn.Diagnostics.Analyzers/CSharp/BlankLines/CSharpBlankLinesBetweenStatementsCodeFixProvider.cs
Youssef1313: https://github.com/dotnet/roslyn/pull/44824#discussion_r439754184

Evangelink: It might be worth doing an anlyzer for this then.

CyrusNajmabadi: I think this logic is correct. If there were multiple diagnostics for the same token, we'd only want to process one of them anyways. If we looked at all diagnostics, we'd have to dedupe. So this has the same effect.

mavasani:
> I am really surprised that dotnet/roslyn#50195 got closed and this analyzer pops-up as a Roslyn specific analyzer.

Personally, I completely agree. There is nothing Roslyn.sln specific about all these formatting analyzers that we are adding them to Roslyn.Diagnostics.Analyzers, our private package. These are extremely useful rules, which should either ship as part of [IDE0055: Fix formatting](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#rule-id-ide0055-fix-formatting) OR a new IDExxxx diagnostic for experimental formatting rules. Either way - the current approach is causing us to block shipping these useful formatting rules to VS customers, which is very unfortunate.

mavasani:
> This pr follows that opinion. I don't think this should be part of roslyn. It can just be something written externally.
> This was also something Manish asked if I would do when I did my last set of analyzers :-)

Cyrus, this was done as part of a temporary experiment to enable dogfooding these rules on Roslyn.sln. I had presented my pushback on this approach even back then, but agreed contingent on the plan that these analyzers will be deleted in 16.8 and appropriately moved into Roslyn repo as part of existing or new formatting diagnostic ID. This unfortunately did not happen.

I have started an email thread around this and proposed we start with deleting these temporary formatting rules from Roslyn.Diagnostics.Analyzers, self-inflict some pain on Roslyn.sln devs and help fast track shipping these as part of VS/CodeStyle NuGet for external customers.

mavasani (changes requested):
Marking this as blocked until we finalize the long term ship strategy for these new, useful formatting rules to external customers. I have no pushback on the analyzer/fixer as such.

CyrusNajmabadi:
Sure. I can move back to roslyn.

CyrusNajmabadi:
This merged into roslyn itself.

PR #4650, opened 2021-01-01
MarkMembersAsStatic: Exclude indexers

### LINKED ISSUE #4651: CA1822 - Indexers cannot be marked as static
Using .NET 5.0 in VS2019 16.8.3:

```csharp
class MyDict
{
    // CA1822	Member 'this[]' does not access instance data and can be marked as static
    public int this[int x] => 42;
}
```

More version info:
```
C# Tools   3.8.0-5.20604.10+9ed4b774d20940880de8df1ca8b07508aa01c8cd
```

Youssef1313:
Fixes #4651

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -111,15 +111,18 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                         return;
                     }
 
-                    bool isInstanceReferenced = false;
+                    bool isInstanceReferenced = methodSymbol.AssociatedSymbol.IsIndexer();
 
-                    blockStartContext.RegisterOperationAction(operationContext =>
+                    if (!isInstanceReferenced)
                     {
-                        if (((IInstanceReferenceOperation)operationContext.Operation).ReferenceKind == InstanceReferenceKind.ContainingTypeInstance)
+                        blockStartContext.RegisterOperationAction(operationContext =>
                         {
-                            isInstanceReferenced = true;
-                        }
-                    }, OperationKind.InstanceReference);
+                            if (((IInstanceReferenceOperation)operationContext.Operation).ReferenceKind == InstanceReferenceKind.ContainingTypeInstance)
+                            {
+                                isInstanceReferenced = true;
+                            }
+                        }, OperationKind.InstanceReference);
+                    }
 
                     blockStartContext.RegisterOperationBlockEndAction(blockEndContext =>
                     {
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -111,15 +111,18 @@ void OnOperationBlockStart(OperationBlockStartAnalysisContext blockStartContext)
                         return;
                     }
 
-                    bool isInstanceReferenced = false;
+                    bool isInstanceReferenced = methodSymbol.AssociatedSymbol.IsIndexer();
 
-                    blockStartContext.RegisterOperationAction(operationContext =>
+                    if (!isInstanceReferenced)
```
Youssef1313, line 116: This might have performance benefits as we now not required to do a RegisterOperationAction for indexers.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -335,6 +335,8 @@ public int Prop1
     public void SomeNotImplementedMethod() => throw new System.NotImplementedException();
 
     public void SomeNotSupportedMethod() => throw new System.NotSupportedException();
+
+    public int this[int x] => 42;
```
Evangelink, line 339: Could you add a similar test for VB.NET?

Evangelink (approved):

mavasani:
I am still trying to understand what realistic user code requires to write an indexer that does not access instance data. I feel we should Won't fix the issue until we have a valid use case.

mavasani:
I am wondering if the better fix here is to report CA1822 with a different message for indexers: `Avoid indexers that do no access instance data`. If required, users can still add a source suppression with a documenting comment why they need such an odd indexer.

Evangelink:
If we go with a different message we probably should do the same thing for properties.

mavasani:
> If we go with a different message we probably should do the same thing for properties.

Possibly, but that will not be in scope of this PR. Note that properties that do not access instance data is a very realistic user scenario, but writing an indexer does not access instance data seems convoluted (unless someone can provide a valid use case).

Evangelink:
> Possibly, but that will not be in scope of this PR.

Agree

> Note that properties that do not access instance data is a very realistic user scenario, but writing an indexer does not access instance data seems convoluted (unless someone can provide a valid use case).

Yes but it is still very unlikely that you can't have them `static`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
mavasani: I think the code will be simpler if reverted to original form and just add the below if statement above this:

```csharp
if (methodSymbol.AssociatedSymbol.IsIndexer())
{
    return;
}
```

mavasani: Actually, can we just enhance the checks in `ShouldAnalyze` to check `!methodSymbol.AssociatedSymbol.IsIndexer()`?

mavasani (changes requested):
Seems fine to bail out on indexers given the context provided by customer on the issue, but lets simplify the fix to just move the new check into `ShouldAnalyze`

mavasani (approved):

PR #4667, opened 2021-01-05
Post release activities for NetAnalyzers release 5.0.3

mavasani:
https://github.com/dotnet/roslyn-analyzers/releases/tag/5.0.3


Evangelink (approved):
I prefer the old release notes with the explicit list of changes better :)

mavasani:
> I prefer the old release notes with the explicit list of changes better

Release notes utility tool written by @dotpaul lists any added/removed rules, plus rules whose severity or enabled by default has changed. The tool does not identify explicit list of all the bug fixes, so in case of patch releases with only bug fixes, there is no automated way to generate such release notes. If you are able to enhance https://github.com/dotnet/roslyn-analyzers/tree/master/src/Tools/ReleaseNotesUtil to do so, that would be great :)

PR #4706, opened 2021-01-17
CA1825: Bail-out for expression trees

### LINKED ISSUE #4665: CA1825 - Avoid unnecessary zero-length array allocations deemed dangerous in expression trees
```csharp
Expression<Func<int[]>> f = () => new int[0];
```

suggests changing to using `Array.Empty<int>()` which LINQ query providers may not expect. In fact, the C# compiler avoids optimization of `new int[0]` to `Array.Empty<int>()` for empty `params` arrays inside expression trees, see https://github.com/dotnet/roslyn/issues/1103. Therefore it's questionable than any analyzer suggests such a rewrite, even for manual empty "new array initialization" because whatever library that consumes the expression tree at runtime may not be hardended to expect a `MethodCallExpression` where a `NewArrayExpression` occurred before.

Youssef1313:
Fixes #4665

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidZeroLengthArrayAllocationsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidZeroLengthArrayAllocationsTests.cs
@@ -708,5 +708,37 @@ class C
 #pragma warning restore RS0030 // Do not used banned APIs
                 fixedSource);
         }
+
+        [Fact]
+        [WorkItem(4665, "https://github.com/dotnet/roslyn-analyzers/issues/4665")]
+        public async Task NoDiagnosticInExpressionTree_CSharp()
+        {
+            const string source = @"
+using System;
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidZeroLengthArrayAllocationsTests.cs
```diff
@@ -708,5 +708,36 @@ class C
...
+{
+    Expression<Func<int[]>> f = () => new int[0];
+}
+";
+            await VerifyCS.VerifyCodeFixAsync(source, source);
+        }
+
+        [Fact]
+        [WorkItem(4665, "https://github.com/dotnet/roslyn-analyzers/issues/4665")]
+        public async Task NoDiagnosticInExpressionTree_VisualBasic()
```
Youssef1313, line 730: 🔴 This test has passed before fixing the issue. It shouldn't have passed. Need to figure out why this happened.

mavasani (approved):
Thank you @Youssef1313!

PR #4709, opened 2021-01-18
Add test: InstanceMemberUsedInXml_NoDiagnostic

Youssef1313:
This false positive was caught in dotnet/roslyn.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -115,7 +115,10 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                     // Workaround for https://github.com/dotnet/roslyn/issues/27564
                     blockStartContext.RegisterOperationAction(operationContext =>
                     {
-                        isInstanceReferenced = true;
+                        if (!operationContext.IsOperationNoneRoot())
+                        {
+                            isInstanceReferenced = true;
+                        }
                     }, OperationKind.None);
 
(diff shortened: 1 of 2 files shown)


Youssef1313:
@mavasani This looks like a Roslyn issue. The operation tree doesn't contain an `IInstanceReferenceOperation` for the case under test. I believe it should have it. I opened https://github.com/dotnet/roslyn/issues/50582

mavasani:
> Do you want me to try a workaround in the analyzer? or keep things as-is until the compiler issue is resolved?

Yes, please add a workaround in the analyzer. https://github.com/dotnet/roslyn/issues/27564 is very tricky to fix, with non-trivial feature work. We have already employed similar workarounds in similar analyzers, such as https://github.com/dotnet/roslyn/issues/33142

Youssef1313:
I can bail-out for OperationKind.None, which may indeed fix the false positive, but have the risk of having false negatives. Should I attempt to get all `IdentifierName` syntaxes inside `XmlEmbeddedExpression`, then use the semantic model to know what this identifier is referring to?

mavasani:
`OperationKind.None` should be very rare, I think it is fine to skip analysis when we have an operation block with any node that we have no knowledge about.

mavasani:
@Youssef1313 Do you plan to add the workaround to the analyzer in this PR or add a skipped attribute to the test?

Youssef1313:
@mavasani An existing test has failed due to bailing out for OperationKind.None

mavasani:
> An existing test has failed due to bailing out for OperationKind.None. But that test already has compile error.

That seems to be due to attributes having `OperationKind.None`. Can you expand the check to also verify that calling this extension method doesn't return true? https://github.com/dotnet/roslyn-analyzers/blob/a0522afd52b25ad211298e308fd5bbc7734865e0/src/Utilities/Compiler/Extensions/IOperationExtensions.cs#L189-L197

mavasani (approved):

PR #4720, opened 2021-01-20
Revert workaround for old SyntaxGenerator bug

### LINKED ISSUE #4717: Remove workaround and make fixer language-agnostic
https://github.com/dotnet/roslyn-analyzers/blob/b148fa89c974a72c28184931767835709f34d1fa/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/Maintainability/CSharpUseNameofInPlaceOfString.Fixer.cs#L18-L28

The SyntaxGenerator bug was fixed in https://github.com/dotnet/roslyn/pull/42214

Youssef1313:
Fixes #4717


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/UseNameofInPlaceOfString.Fixer.cs
```diff
@@ -47,15 +46,15 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+        private static async Task<Document> ReplaceWithNameOf(Document document, SyntaxNode nodeToReplace,
             string stringText, CancellationToken cancellationToken)
         {
             var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
             var generator = editor.Generator;
 
             var trailingTrivia = nodeToReplace.GetTrailingTrivia();
             var leadingTrivia = nodeToReplace.GetLeadingTrivia();
-            SyntaxNode nameOfExpression = GetNameOfExpression(generator, stringText)
+            var nameOfExpression = generator.NameOfExpression(generator.IdentifierName(stringText))
```
Youssef1313, line 57: The usage of explicit type generates an IDE0007 warning locally. I'm completely confused how this doesn't fail CI?

Youssef1313: Actually, there are **tons** of warnings locally. Including RS1024 which is explicitly disabled in editorconfig

image

image

Evangelink (approved):

sharwell:
Currently the analyzers support Roslyn 3.3.1 and newer. The `SyntaxGenerator` bug was only fixed for 3.6.0.

Seems like this change needs to be blocked on a future update (unknown time) when this version is updated to 3.6.0 or newer:

https://github.com/dotnet/roslyn-analyzers/blob/33582d0369adb7657731b611f4811e981dcb2f7f/eng/Versions.props#L31

mavasani:
> Seems like this change needs to be blocked on a future update (unknown time) when this version is updated to 3.6.0 or newer:

@sharwell There is no API dependency in the fix here that requires us to update to newer Microsoft.CodeAnalysis package. I feel we should do the following:

1. Take this change as-is
2. If user installs this package and uses the code fix on a VS version before VS2019 16.6, they will hit the original bug. We would treat this as a VS bug which has been fixed and request customer to move to VS2019 16.6 or later. IMO, that is a reasonable story.

mavasani (approved):
LGTM, but will wait for @sharwell's approval.

sharwell:
@mavasani if a user hits this and reports it, the issue is going to get routed to you anyway 😄

Youssef1313:
@sharwell Users can never hit this bug, even with the buggy behavior of SyntaxGenerator. The IDE will re-parse the text produced anyway. This workaround was mainly done for the tests here to pass, as the tests don't reparse the syntax tree.

mavasani:
@Youssef1313 - Ah yes, thanks for the clarification. I recall the issue now, and the fact that it was test-only workaround. This gives me lot more confidence to go ahead with the change.

sharwell:
> This workaround was mainly done for the tests here to pass, as the tests don't reparse the syntax tree.

The tests always reparse the syntax tree:

https://github.com/dotnet/roslyn-sdk/blob/6f9cf96ec9e94453ac750c50c3b84c554037edcb/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest%601.cs#L182

Sounds like the change that made this PR viable was switching to the new test library a while back.

Youssef1313:
> The tests always reparse the syntax tree:

@sharwell Thanks for clarifying. It should the old tests which didn't reparse.

PR #4723, opened 2021-01-21
Fix outdated rule id in comments

Youssef1313:
https://github.com/dotnet/roslyn-analyzers/blob/4bbb229fda9e3e2bdf9389350ade042237075cec/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs#L16

Not of much value, but did it anyway as a small cleanup.


genlu (approved):
:shipit:

PR #4725, opened 2021-01-21
Override Stream ReadAsync/WriteAsync Analyzer

NewellClark:
This fixes [issue #33789](https://github.com/dotnet/runtime/issues/33789). 

I ran the analyzer against dotnet/runtime and found several violations:
- [DeflateStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs), [ChunkedMemoryStream](https://github.com/dotnet/runtime/blob/master/src/libraries/Common/src/System/IO/ChunkedMemoryStream.cs), [RequestStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs), and [NetworkStreamWrapper](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Requests/src/System/Net/NetworkStreamWrapper.cs) were all trivial, and have been dealt with in [this PR](https://github.com/dotnet/runtime/pull/47125). 
- [CryptoStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs) was non-trivial, and is being worked on in [this PR](https://github.com/dotnet/runtime/pull/47207).
- [DeflateManagedStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateManaged/DeflateManagedStream.cs) is internal, and calls into the [IFileFormatReader](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateManaged/FileFormats.cs) interface, which is also internal. I was unable to find any classes that actually implement this interface throughout dotnet/runtime. 
We could do what we're doing for `CryptoStream` and check if the `Memory<byte>` is actually an array, and if not simply do the copy, but since this interface is internal, I figured we might be able to explore other options.
- [WebSocketHttpListenerDuplexStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.HttpListener/src/System/Net/Windows/WebSockets/WebSocketHttpListenerDuplexStream.cs) is also internal, and from what I could find, `ReadAsync`/`WriteAsync` are only ever called from [WebSocketBase](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.HttpListener/src/System/Net/Windows/WebSockets/WebSocketBase.cs), which only calls the array-based override. Unfortunately, `WebSocketHttpListenerDuplexStream` also calls back into `WebSocket`, and it passes the array it was given into multiple different methods on `WebSocket`, and some methods on [WebSocketBuffer](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.HttpListener/src/System/Net/Windows/WebSockets/WebSocketBuffer.cs). Providing memory-based overrides would require almost a complete rewrite of `WebSocketBuffer` as well as changes to code that consumes it. Since nothing currently calls the memory-based overrides on `WebSocketHttpListenerDuplexStream`, I think we should strongly consider suppressing the warning.


mavasani:
@NewellClark Seems like your PR has many more changes then intended.

mavasani:
The target branch seems to be incorrect `dotnet:merges/master-to-release/6.0.1xx-preview1`. Did you intend to target `master` or `release/6.0.1xx-preview1`?

NewellClark:
> Did you intend to target `master` or `release/6.0.1xx-preview1`? 
Oops, so I did. I will fix it.

PR #4687, opened 2021-01-12
(1/4) Issue 35343 - Replace Dictionary<,>.Keys.Contains with Dictionary<,>.ContainsKey

NewellClark:
Fixes [issue #35343](https://github.com/dotnet/runtime/issues/35343). 

I've implemented this to report diagnostics whenever an invocation of the form  `dictionary.Keys.Contains(TKey)` or `dictionary.Values.Contains(TValue)` is encountered, provided that:
1) `dictionary`'s compile-time type implements `IDictionary<TKey, TValue>`
2) `Contains()` returns `bool` and accepts one argument (not counting the first argument if it is an extension method), and
3) The compile-time type of `dictionary` has an available matching `bool ContainsKey(TKey)` or `bool ContainsValue(TValue)` method, respectively. 

I ran the analyzer against runtime, roslyn-analyzers, and roslyn, and found no violations.

PR #4726, opened 2021-01-21
(2/4) Override Stream ReadAsync/WriteAsync Analyzer

NewellClark:
Fixes [issue #33789](https://github.com/dotnet/runtime/issues/33789). 

I ran the analyzer against dotnet/runtime and found several violations:
- [DeflateStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs), [ChunkedMemoryStream](https://github.com/dotnet/runtime/blob/master/src/libraries/Common/src/System/IO/ChunkedMemoryStream.cs), [RequestStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs), and [NetworkStreamWrapper](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Requests/src/System/Net/NetworkStreamWrapper.cs) were all trivial, and have been dealt with in [this PR](https://github.com/dotnet/runtime/pull/47125). 
- [CryptoStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs) was non-trivial, and is being worked on in [this PR](https://github.com/dotnet/runtime/pull/47207).
- [DeflateManagedStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateManaged/DeflateManagedStream.cs) is internal, and calls into the [IFileFormatReader](https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateManaged/FileFormats.cs) interface, which is a dead interface and is being removed in [this PR](https://github.com/dotnet/runtime/pull/47408). Once that gets merged, I'll memorify `DeflateManagedStream`, which will have no extra copies. Edit: I've memorified `DeflateManagedStream` ([PR](https://github.com/dotnet/runtime/pull/48085)).
- [WebSocketHttpListenerDuplexStream](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.HttpListener/src/System/Net/Windows/WebSockets/WebSocketHttpListenerDuplexStream.cs) is also internal, and from what I could find, `ReadAsync`/`WriteAsync` are only ever called from [WebSocketBase](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.HttpListener/src/System/Net/Windows/WebSockets/WebSocketBase.cs), which only calls the array-based override. Unfortunately, `WebSocketHttpListenerDuplexStream` also calls back into `WebSocket`, and it passes the array it was given into multiple different methods on `WebSocket`, and some methods on [WebSocketBuffer](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.HttpListener/src/System/Net/Windows/WebSockets/WebSocketBuffer.cs). Providing memory-based overrides would require almost a complete rewrite of `WebSocketBuffer` as well as changes to code that consumes it. Since nothing currently calls the memory-based overrides on `WebSocketHttpListenerDuplexStream`, I think we should strongly consider suppressing the warning. 

Note that I'm using `CA1840` as the diagnostic ID because I used `CA1839` in [another analyzer](https://github.com/dotnet/roslyn-analyzers/pull/4687) that has yet to be merged in. Not sure how that type of situation is typically handled.


NewellClark:
@mavasani Sorry about that. I've retargeted this PR to the correct branch.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,208 @@
...
+            DiagnosticCategory.Performance,
+            RuleLevel.IdeSuggestion,
+            s_localizableDescription,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private const string ReadAsyncName = "ReadAsync";
+        private const string WriteAsyncName = "WriteAsync";
```
Youssef1313, line 42: nit:

```suggestion
        private const string ReadAsyncName = nameof(System.IO.Stream.ReadAsync);
        private const string WriteAsyncName = nameof(System.IO.Stream.WriteAsync);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,208 @@
...
+            public ITypeSymbol ReadOnlyMemoryOfByteType { get; }
+            public IMethodSymbol ReadAsyncArrayMethod { get; }
+            public IMethodSymbol ReadAsyncMemoryMethod { get; }
+            public IMethodSymbol WriteAsyncArrayMethod { get; }
+            public IMethodSymbol WriteAsyncMemoryMethod { get; }
+        }
+
+        private static LocalizableString CreateResource(string resourceName)
+        {
+            return new LocalizableResourceString(resourceName, Resx.ResourceManager, typeof(Resx));
```
Youssef1313, line 205: Should that be `MicrosoftNetCoreAnalyzersResources.ResourceManager` and `typeof(MicrosoftNetCoreAnalyzersResources)`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,208 @@
...
+            return derivedType.GetMembers(overriddenMethod.Name)
+                .SingleOrDefault(x =>
+                {
+                    return x.IsOverride && overriddenMethod.Equals(x.GetOverriddenMember(), SymbolEqualityComparer.Default);
+                }) as IMethodSymbol;
+        }
+
+        //  We will not be doing any comparisons on this type.
+#pragma warning disable CA1815 // Override equals and operator equals on value types
+        private readonly struct RequiredSymbols
```
Youssef1313, line 177: Not sure if we could consider a `record` here?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        //  If the specified type overrides the specified method on its immediate base class, returns the method symbol representing the 
+        //  overriding method. Returns null if the specified type does not override the specified method.
+        private static IMethodSymbol? GetSymbolForOverridingMethod(ITypeSymbol derivedType, IMethodSymbol overriddenMethod)
+        {
+            RoslynDebug.Assert(derivedType.BaseType.Equals(overriddenMethod.ContainingType, SymbolEqualityComparer.Default));
+
+            return derivedType.GetMembers(overriddenMethod.Name)
+                .SingleOrDefault(x =>
```
Youssef1313, line 169: Can it happen there are multiple symbols with these criteria, hence the call to `SingleOrDefault` throws an exception?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,208 @@
...
+                writeAsyncArrayMethod, writeAsyncMemoryMethod);
+            return true;
+        }
+
+        private static IMethodSymbol? GetOverload(ITypeSymbol containingType, string methodName, params ITypeSymbol[] argumentTypes)
+        {
+            return containingType.GetMembers(methodName)
+                .SingleOrDefault(symbol => symbol is IMethodSymbol m && IsMatch(m, argumentTypes)) as IMethodSymbol;
+
+            static bool IsMatch(IMethodSymbol method, ITypeSymbol[] argumentTypes)
```
Youssef1313, line 147: @Evangelink Was there a helper method that does this?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: I placed a `using Resx = Microsoft.NetCore.Analyzers.MicrosoftNetCoreAnalyzersResources;` at the top of the file to reduce line-length in places.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: I figured we could save on the allocation by using a struct, since it's not copied (it's literally just passed out of `TryGetRequiredSymbols` via an **out** parameter), and I've heard allocations are a pain point in the compiler.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
Youssef1313: @NewellClark Ah thanks for clarifying. I didn't notice that line.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: It should never happen. We check if

1. The method has a specific name,
2. It is an overriding method (which means its original definition must be a virtual method, which rules out any method-hiding)
3. The types of its arguments match exactly.

I don't see any way in C# or VB for a class to have two methods that match all of those criteria. I used `SingleOrDefault` deliberately to document the impossibility of the above scenario.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: The helper method requires instances of `ParameterInfo`, which themselves require `INamedTypeSymbol` instances. It was significantly shorter to write it out my way. I did originally write it out using the helper method, but it was significantly longer and more verbose.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: Oops, that was sloppy of me. I will fix that here and in other places where I should have used `nameof`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
Youssef1313: @NewellClark Yes, this most likely shouldn't happen for valid code. But my concern was about "invalid" code. The analyzers are run against code that has compile error.

Consider checking what would happen if the same method is just repeated (which will have a compile-error). Will GetMembers return repeated methods for this error scenario?

NewellClark: @Youssef1313 Good catch. Just tested it and it does indeed crash. Working on a fix right now.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
```diff
@@ -0,0 +1,1476 @@
...
+            string code = $@"
+{CSUsings}
+namespace Testopolis
+{{
+    public class {{|#0:FooStream|}} : Stream
+    {{
+        {CSAbstractMembers}
+        {CSReadAsyncArray}
+    }}
+}}";
```
Evangelink, line 36: [cosmetic] I am fine if you keep it this way but I find the double curly braces hard to read.

```suggestion
            string code = @"
" + CSUsings + @"
namespace Testopolis
{
    public class {|#0:FooStream|} : Stream
    {
        " + CSAbstractMembers + @"
        " + CSReadAsyncArray + @"
    }
}";
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,14 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="ProvideStreamMemoryBasedAsyncOverridesDescription" xml:space="preserve">
+    <value>The default implementation of '{2}' delegates to '{1}', which is innefficient as it requires an extra copy. It also results in extra allocations as it returns Task instead of ValueTask. Consider moving your implementation to '{2}' and have '{1}' delegate to '{2}'.</value>
```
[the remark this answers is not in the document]
NewellClark: Oops. So I did. `{0}` is used in message, but not description. I can change description so that it includes a `{0}`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,219 @@
...
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+
+            if (!TryGetRequiredSymbols(compilation, out RequiredSymbols symbols))
+                return;
+
+            context.RegisterSymbolAction(AnalyzeNamedType, SymbolKind.NamedType);
+
+            void AnalyzeNamedType(SymbolAnalysisContext context)
```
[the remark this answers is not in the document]
NewellClark: It uses the `symbols` struct produced by the call to `TryGetRequiredSymbols(...)` on line 55.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,219 @@
...
+
+            if (memoryOfByteType is null || readOnlyMemoryOfByteType is null || streamType is null || cancellationTokenType is null)
+            {
+                requiredSymbols = default;
+                return false;
+            }
+
+            // We know that there are no compiler errors on streamType, so we use 'SingleOrDefault'.
+
+            var readAsyncArrayMethod = GetOverloads(streamType, ReadAsyncName, byteArrayType, int32Type, int32Type, cancellationTokenType).SingleOrDefault();
```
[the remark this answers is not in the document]
NewellClark: Maybe use `FirstOrDefault()` but assert if the length of the array is more than one?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: Going forward, I will do it your way.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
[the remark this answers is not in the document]
NewellClark: I figured since we're looking up methods on framework types (`Stream`) that are in metadata, and therefore could never contain compiler errors. Now that I think about it though, if this analyzer is running against `System.Private.CoreLib`, these framework types could actually contain compiler errors, and we definitely don't want the analyzer to crash. I suppose we could detect whether `streamType` is in source code or metadata, but it's probably not worth the hassle.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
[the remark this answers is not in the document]
Evangelink: My bad!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                RoslynDebug.Assert(arrayBasedOverride.OverriddenMethod is not null);
+
+                //  We want to underline the name of the violating type in the class declaration. If the violating type
+                //  is a partial class, we underline all partial declarations.
+
+                return Diagnostic.Create(
+                    Rule,
+                    violatingType.Locations[0],
+                    violatingType.Locations.Skip(1),
```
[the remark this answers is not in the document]
NewellClark: I've fixed this in the latest commit.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
Evangelink: Let's wait for @sharwell or @mavasani opinion here.

NewellClark:
This should be linked to issue #33789.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,220 @@
...
+                    violatingType.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat),
+                    arrayBasedOverride.OverriddenMethod.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat),
+                    memoryBasedMethod.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat));
+            }
+        }
+
+        private static bool TryGetRequiredSymbols(Compilation compilation, out RequiredSymbols requiredSymbols)
+        {
+            var int32Type = compilation.GetSpecialType(SpecialType.System_Int32);
+            var byteType = compilation.GetSpecialType(SpecialType.System_Byte);
```
Youssef1313, line 112: @mavasani Can these be null under any case?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
```diff
@@ -0,0 +1,1476 @@
...
+            var test = new VerifyVB.Test
+            {
+                TestCode = code,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                ExpectedDiagnostics =
+                {
+                    VerifyVB.Diagnostic(Rule)
+                        .WithLocation(0)
+                        .WithLocation(1)
+                        .WithArguments("River", displayArrayMethod, displayMemoryMethod)
```
Youssef1313, line 253: This is very confusing. Is this supposed to be two diagnostics?
I believe that only one diagnostic is passed to the test.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
```diff
@@ -0,0 +1,1476 @@
...
+            var test = new VerifyCS.Test
+            {
+                TestState = { Sources = { fooSource, barSource, bazSource } },
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                ExpectedDiagnostics =
+                {
+                    VerifyCS.Diagnostic(Rule)
+                        .WithLocation(0)
+                        .WithLocation(1)
+                        .WithLocation(2)
```
Youssef1313, line 299: Same comment here.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
```diff
@@ -0,0 +1,1476 @@
...
+                        .WithArguments("River", CSDisplayWriteAsyncArray, CSDisplayWriteAsyncMemory)
+                }
+            };
+            return test.RunAsync();
+        }
+
+        [Fact]
+        public Task BothArrayOverrides_MultiplePartialsInSameFile_REportsAllLocations_VB()
```
Youssef1313, line 399: nit: typo

```suggestion
        public Task BothArrayOverrides_MultiplePartialsInSameFile_ReportsAllLocations_VB()
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
```diff
@@ -0,0 +1,1476 @@
...
+        End Function";
+        private const string VBHideExplicitWriteAsyncArray = @"Public Overloads Function WriteAsync(buffer As Byte, offset As Integer, count As Integer, cancellationToken As CancellationToken) As Task
+            Throw New NotImplementedException()
+        End Function";
+        private const string VBHideExplicitWriteAsyncMemory = @"Public Overloads Function WriteAsync(buffer As ReadOnlyMemory(Of Byte), Optional cancellationToken As CancellationToken = Nothing) As ValueTask
+            Throw New NotImplementedException()
+        End Function";
+
+        private static DiagnosticDescriptor Rule => ProvideStreamMemoryBasedAsyncOverrides.Rule;
+        private static string RuleId => ProvideStreamMemoryBasedAsyncOverrides.RuleId;
```
Youssef1313, line 1463: Could thie be a constant?

```suggestion
        private const string RuleId = ProvideStreamMemoryBasedAsyncOverrides.RuleId;
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
```diff
@@ -0,0 +1,1476 @@
...
+            Throw New NotImplementedException()
+        End Function";
+        private const string VBHideExplicitWriteAsyncArray = @"Public Overloads Function WriteAsync(buffer As Byte, offset As Integer, count As Integer, cancellationToken As CancellationToken) As Task
+            Throw New NotImplementedException()
+        End Function";
+        private const string VBHideExplicitWriteAsyncMemory = @"Public Overloads Function WriteAsync(buffer As ReadOnlyMemory(Of Byte), Optional cancellationToken As CancellationToken = Nothing) As ValueTask
+            Throw New NotImplementedException()
+        End Function";
+
+        private static DiagnosticDescriptor Rule => ProvideStreamMemoryBasedAsyncOverrides.Rule;
```
Youssef1313, line 1462: Could this be a readonly field?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: The location that is underlined when this diagnostic is reported is the name of the class in the class declaration. If the class is a partial class, all partial declarations need to be underlined. That is what this test is checking: that when you have multiple partial declarations in the same file, all partial declarations will be underlined for a single violation.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: Same as my previous comment, but this time we are checking whether all locations are reported when the partials are in separate files.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: I suppose it could. I figured since it's just a shorthand for accessing the field on the analyzer class, it makes sense just to make it a property that simply returns the field.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
Youssef1313: @NewellClark If I understand correctly, you're passing only one diagnostic. When you chain `WithLocation` calls, only the latest one takes effect.

I believe if you removed the chain and kept the latest WithLocation call only, the test will still pass. Could you confirm that locally on your machine?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: Same here. It's just forwarding to the analyzer class. Is there an advantage that a readonly field has over a readonly property that just forwards to a different readonly field?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
Youssef1313: @NewellClark for the purpose of test. It probably doesn't matter a lot. But it's better practice to use private constant.

See [CA1802](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1802?view=vs-2019).

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: Just tested it. The test fails if we remove any of the `WithLocation` calls, and the error message reports the diagnostics at the removed locations as the unexpected diagnostics.

Youssef1313: @NewellClark Good to know! Thanks.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: CA1802 talks about using a static readonly field instead of a constant. A non-auto readonly property has no associated storage, and thus does not need a static constructor to initialize it at runtime.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryAsyncOverridesTests.cs
NewellClark: @Youssef1313 by the way, would it be possible to have this pull request linked to [issue 33789](https://github.com/dotnet/runtime/issues/33789)?

Youssef1313: @NewellClark The issue won't be auto-closed when the PR is merged. But a Microsoft employee will close it when he merges your PR.

Youssef1313 (commented):
Quoting part from the runtime issue:

> The rule should flag types derived from Stream that override `BeginRead`/`EndRead` or `BeginWrite`/`EndWrite` but that don't override `ReadAsync` or `WriteAsync`.

You're probably missing this case?

NewellClark:
> Quoting part from the runtime issue:
> 
> > The rule should flag types derived from Stream that override `BeginRead`/`EndRead` or `BeginWrite`/`EndWrite` but that don't override `ReadAsync` or `WriteAsync`.
> 
> You're probably missing this case?

When they did the API review in [this video](https://www.youtube.com/watch?v=Dcu0Cz69onc&t=0h20m44s), it was decided to ignore the `BeginRead`/`EndRead` cases and only flag for the task-based methods that accept arrays. The logic was that nobody has used `BeginRead`/`EndRead` since like 2001.

Youssef1313:
> > Quoting part from the runtime issue:
> > > The rule should flag types derived from Stream that override `BeginRead`/`EndRead` or `BeginWrite`/`EndWrite` but that don't override `ReadAsync` or `WriteAsync`.
> > 
> > 
> > You're probably missing this case?
> 
> When they did the API review in [this video](https://www.youtube.com/watch?v=Dcu0Cz69onc&t=0h20m44s), it was decided to ignore the `BeginRead`/`EndRead` cases and only flag for the task-based methods that accept arrays. The logic was that nobody has used `BeginRead`/`EndRead` since like 2001.

Yeah I see. I had my comment updated.

carlossanlop:
@mavasani is `dotnet:release/6.0.1xx-preview1` the correct branch? Is there a new rule stating we should not merge directly to `main`?

mavasani:
I believe this should go into preview4 branch now. @jmarolf to confirm.

jmarolf:
@mavasani correct! preview 3 has released this should be retargeted to release/6.0.1xx-preview4

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,14 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="ProvideStreamMemoryBasedAsyncOverridesDescription" xml:space="preserve">
+    <value>The default implementation of '{2}' delegates to '{1}', which is innefficient as it requires an extra copy. It also results in extra allocations as it returns Task instead of ValueTask. Consider moving {0}'s implementation to '{2}' and have '{1}' delegate to '{2}'.</value>
```
```suggestion
    <value>The default implementation of '{2}' delegates to '{1}', which is innefficient as it requires an extra copy. It also results in extra allocations as it returns 'Task' instead of 'ValueTask'. Consider moving {0}'s implementation to '{2}' and have '{1}' delegate to '{2}'.</value>
```

mavasani: Actually, you cannot have format arguments `{0}`, `{1}`, etc. in description string, it is a static description per-descriptor, not something instantiated per-diagnostic. You should instead follow what @carlossanlop did in a similar analyzer and use more generic description string that is applicable for all violations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,215 @@
...
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+
+            if (!TryGetRequiredSymbols(compilation, out RequiredSymbols symbols))
+                return;
+
+            context.RegisterSymbolAction(AnalyzeNamedType, SymbolKind.NamedType);
+
+            void AnalyzeNamedType(SymbolAnalysisContext context)
```
mavasani, line 60:
```suggestion
            context.RegisterSymbolAction(AnalyzeNamedType, SymbolKind.NamedType);
            return;

            // Local functions
            void AnalyzeNamedType(SymbolAnalysisContext context)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
mavasani: Yes, they can be null for broken compilation references. We should have a null check for these types as well.

src/NetAnalyzers/RulesMissingDocumentation.md
```diff
@@ -2,3 +2,4 @@
 
 Rule ID | Missing Help Link | Title |
 --------|-------------------|-------|
+CA1840 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1840> | Provide memory-based overrides of async methods when subclassing 'Stream' |
```
mavasani, line 5: ID has already been used, please choose a new one.

mavasani (commented):
Overall looks great. Please retarget to preview4 branch.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: Fixed

src/NetAnalyzers/RulesMissingDocumentation.md
NewellClark: Changed ID to ca1842

mavasani:
@NewellClark Sorry for not getting back to this PR review sooner, but unfortunately even preview4 is now locked. Can you please retarget to https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx? Promise this one is the last branch retarget request. Again, very sorry for this additional work.

Can you also address the remaining review comments? I should be able to merge once both of these are taken care of.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: @mavasani Fixed. Sorry I missed this earlier.

NewellClark:
> @NewellClark Sorry for not getting back to this PR review sooner, but unfortunately even preview4 is now locked. Can you please retarget to https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx? Promise this one is the last branch retarget request. Again, very sorry for this additional work.
> 
> Can you also address the remaining review comments? I should be able to merge once both of these are taken care of.

@mavasani Fixed. Should I retarget for all my other analyzer PRs that are targeted to preview4?

mavasani:
> Should I retarget for all my other analyzer PRs that are targeted to preview4?

Yes please. preview4 branch is now locked.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
```diff
@@ -0,0 +1,225 @@
...
+            static Diagnostic CreateDiagnostic(INamedTypeSymbol violatingType, IMethodSymbol arrayBasedOverride, IMethodSymbol memoryBasedMethod)
+            {
+                RoslynDebug.Assert(arrayBasedOverride.OverriddenMethod is not null);
+
+                //  We want to underline the name of the violating type in the class declaration. If the violating type
+                //  is a partial class, we underline all partial declarations.
+
+                return violatingType.CreateDiagnostic(
+                    Rule,
+                    violatingType.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat),
```
mavasani, line 106: We normally use just the type/method `.Name` in diagnostic messages. `MinimallyQualifiedFormat` actually produces a qualified name, which lead to very long names.

mavasani (approved):
LGTM.

I can merge once you change the diagnostic messages to use simple names instead of ToDisplayString.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ProvideStreamMemoryBasedAsyncOverrides.cs
NewellClark: @mavasani Fixed. I've changed the message so it makes sense with simple names (the message was originally worded with the assumption that the arguments of each overload would be displayed).

mavasani (approved):
Thanks!

mavasani:
@NewellClark Can you please help us in creating documentation PRs for all the new .NET6 analyzers that you have contributed? Please see the item 4. in https://github.com/dotnet/roslyn-analyzers/blob/main/GuidelinesForNewRules.md.

Also tagging @gewarren as we now have 8 new undocumented CA rules which have been implemented in the `release\6.0.1xx` branch for .NET6: https://github.com/dotnet/roslyn-analyzers/blob/release/6.0.1xx/src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md.

NewellClark:
> @NewellClark Can you please help us in creating documentation PRs for all the new .NET6 analyzers that you have contributed? Please see the item 4. in https://github.com/dotnet/roslyn-analyzers/blob/main/GuidelinesForNewRules.md.
> 
> Also tagging @gewarren as we now have 8 new undocumented CA rules which have been implemented in the `release\6.0.1xx` branch for .NET6: https://github.com/dotnet/roslyn-analyzers/blob/release/6.0.1xx/src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md.

Absolutely. I'm on it.

PR #4731, opened 2021-01-22
draft - enable ca1508

Youssef1313:
It's not enabled currently due to AllEnabledByDefault bug for which the fix isn't shipped yet. This PR isn't meant to be merged, but to check if there are any false positive to CA1508 in case any of them can be fixed before the next release.


Youssef1313:
Build passed. Closing

PR #4740, opened 2021-01-27
CA1416 handle call-site empty sets causing warnings when TFM attributes applied

buyaa-n:
Related to https://github.com/dotnet/runtime/issues/44231

[By design](https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md#platform-context), SDK adds SuuportedOSPlatform attribute with the targeted OS platform from TFMs, which works well for normal condition but for multitargeted env like `runtime` repo it is causing warnings mentioned in  https://github.com/dotnet/runtime/issues/44231 and requiring clear separation of platform-specific APIs which forcing developers to duplicate their code to conform to the strict supported-platform name rules. After evaluating of possible solution and several discussions we have decided to resolve it at the analyzer level. The TFM target attributes are causing inconsistent scenarios in the call site and the real attributes within that assembly are being ignored or malfunctioned which causing the warnings, this shouldn't happen in normal development. By the main rule of "Child API attributes cannot widen parent or containing type attributes" when the inconsistency happens in the call site and call site evaluated to an empty set of support not gonna warn within that empty set context.

cc @jeffhandley @terrajobst  

Examples:
```cs
[SupportedOSPlatform("windows7")]
class Test1
{
    [UnsupportedOSPlatform("windows9")]
    void StillSupportedOnWindows7and8()
    {
        SupportedOnWindows10(); // Warning; calling from Windows 7-8 is still supported but it will fail
    }
    [UnsupportedOSPlatform("windows7")]
    void NotSupportedAnywhere()
    {
        SupportedOnWindows10(); // No warning; no valid call sites
    }
    [SupportedOSPlatform("windows10")]
    void SupportedOnWindows10()
    {
    }
}
class Test2
{
    class Test2MacOs
    {
        void WindowsOnly()
        {
            // No valid call sites; no warnings no matter what
        }
        [SupportedOSPlatform("linux")]
        void LinuxOnly()
        {
            // No valid call sites; no warnings no matter what
        }
    }
}
```

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using Analyzer.Utilities;
 
 namespace Microsoft.NetCore.Analyzers.InteropServices
 {
@@ -22,15 +23,28 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         ///  - UnsupportedFirst - keeps the lowest version of [UnsupportedOSPlatform] attribute found
         ///  - UnsupportedSecond - keeps the second lowest version of [UnsupportedOSPlatform] attribute found
         /// </summary>
-        private class PlatformAttributes
+        private class Versions
         {
             public Version? SupportedFirst { get; set; }
             public Version? SupportedSecond { get; set; }
             public Version? UnsupportedFirst { get; set; }
             public Version? UnsupportedSecond { get; set; }
-            public Callsite Callsite { get; set; }
-            public bool HasAttribute() => SupportedFirst != null || UnsupportedFirst != null ||
+            public bool IsSet() => SupportedFirst != null || UnsupportedFirst != null ||
                         SupportedSecond != null || UnsupportedSecond != null;
+        }
+
+        private class PlatformAttributes
+        {
+            public PlatformAttributes() { }
+
+            public PlatformAttributes(Callsite callsite, SmallDictionary<string, Versions> platforms)
+            {
+                Callsite = callsite;
+                Platforms = platforms;
+            }
+
+            public SmallDictionary<string, Versions>? Platforms { get; set; }
+            public Callsite Callsite { get; set; }
         }
     }
 }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
@@ -33,7 +33,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         SupportedSecond != null || UnsupportedSecond != null;
         }
 
-        private class PlatformAttributes
+        private sealed class PlatformAttributes
         {
             public PlatformAttributes() { }
 
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
@@ -28,6 +28,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             public Version? SupportedSecond { get; set; }
             public Version? UnsupportedFirst { get; set; }
             public Version? UnsupportedSecond { get; set; }
+            public Callsite Callsite { get; set; }
             public bool HasAttribute() => SupportedFirst != null || UnsupportedFirst != null ||
                         SupportedSecond != null || UnsupportedSecond != null;
         }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -233,8 +233,8 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             }
 
             var osPlatformTypeArray = osPlatformType != null ? ImmutableArray.Create(osPlatformType) : ImmutableArray<INamedTypeSymbol>.Empty;
-            var platformSpecificOperations = PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>, (SmallDictionary<string, Versions> attributes,
-                SmallDictionary<string, Versions>? csAttributes)>.GetInstance();
+            var platformSpecificOperations = PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>,
+                (SmallDictionary<string, Versions> attributes, SmallDictionary<string, Versions>? csAttributes)>.GetInstance();
 
             context.RegisterOperationAction(context =>
             {
@@ -251,7 +251,8 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             {
                 try
                 {
-                    if (platformSpecificOperations.IsEmpty || guardMethods.IsEmpty ||
+                    if (platformSpecificOperations.IsEmpty ||
+                        guardMethods.IsEmpty ||
                         !(context.OperationBlocks.GetControlFlowGraph() is { } cfg))
                     {
                         return;
@@ -416,7 +417,8 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                     attribute.UnsupportedFirst = null;
                                 }
 
-                                if (attribute.UnsupportedSecond != null && attribute.UnsupportedSecond <= info.Version)
+                                if (attribute.UnsupportedSecond != null &&
+                                    attribute.UnsupportedSecond <= info.Version)
                                 {
                                     attribute.UnsupportedSecond = null;
                                 }
@@ -1064,8 +1066,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                             platformSpecificOperations.TryAdd(new KeyValuePair<IOperation, ISymbol>(operation, symbol), (notSuppressedAttributes, callSiteAttributes.Platforms));
                         }
                     }
-                    else if (callSiteAttributes.Callsite != Callsite.Empty &&
-                        TryCopyAttributesNotSuppressedByMsBuild(operationAttributes.Platforms!, msBuildPlatforms, out var copiedAttributes))
+                    else if (TryCopyAttributesNotSuppressedByMsBuild(operationAttributes.Platforms!, msBuildPlatforms, out var copiedAttributes))
                     {
                         platformSpecificOperations.TryAdd(new KeyValuePair<IOperation, ISymbol>(operation, symbol), (copiedAttributes, null));
                     }
@@ -1174,8 +1175,9 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                 diagnosticAttribute.SupportedFirst = (Version)attributeToCheck.Clone();
                             }
 
-                            if (attribute.UnsupportedFirst != null && (!mandatorySupportFound.Value ||
-                                !(SuppressedByCallSiteSupported(attribute, callSiteAttribute.SupportedFirst) ||
+                            if (attribute.UnsupportedFirst != null &&
+                                (!mandatorySupportFound.Value ||
+                                  !(SuppressedByCallSiteSupported(attribute, callSiteAttribute.SupportedFirst) ||
                                   SuppressedByCallSiteUnsupported(callSiteAttribute, attribute.UnsupportedFirst))))
                             {
                                 diagnosticAttribute.UnsupportedFirst = (Version)attribute.UnsupportedFirst.Clone();
@@ -1257,7 +1259,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         }
                     }
                     else if (msBuildPlatforms.Contains(platformName, StringComparer.OrdinalIgnoreCase) &&
-                        !callSiteAttributes.Values.Any(v => v.SupportedFirst != null))
+                             !callSiteAttributes.Values.Any(v => v.SupportedFirst != null))
                     {
                         // if MsBuild list contain the platform and call site has no any other supported attribute it means global, so need to warn
                         diagnosticAttribute.UnsupportedFirst = (Version)attribute.UnsupportedFirst.Clone();
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -1505,21 +1505,17 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                             }
                         }
                     }
-                    // For allow list if child narrowing supported platfroms by having less platforms support than parent,
-                    // not existing parent platforms should be removed
+
                     if (notFoundPlatforms.Count > 0)
                     {
+                        // For allow list if child narrowing supported platforms by having less platforms support than parent,
+                        // not existing parent platforms should be removed
                         if (supportFound)
                         {
-                            childAttributes = new SmallDictionary<string, Versions>(StringComparer.OrdinalIgnoreCase);
-                            foreach (var (platform, attributes) in pAttributes)
+                            foreach (var platform in notFoundPlatforms)
                             {
-                                if (!notFoundPlatforms.Contains(platform))
-                                {
-                                    childAttributes.Add(platform, attributes);
-                                }
+                                parentAttributes.Platforms!.Remove(platform);
                             }
-                            parentAttributes.Platforms = childAttributes;
                         }
                         else
                         {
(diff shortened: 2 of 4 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
```diff
@@ -22,14 +23,28 @@ public sealed partial class PlatformCompatibilityAnalyzer
...
                         SupportedSecond != null || UnsupportedSecond != null;
         }
+
+        private class PlatformAttributes
+        {
+            public PlatformAttributes() { }
+
+            public PlatformAttributes(Callsite callsite, SmallDictionary<string, Versions> platforms)
+            {
+                Callsite = callsite;
```
buyaa-n, line 42: Because of attributes kept in the cache i needed to add Call-site info for subsequent calls made within that call site

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -293,12 +289,7 @@ private static bool PlatformAnalysisAllowed(AnalyzerOptions options, Compilation
                 }
                 finally
                 {
-                    // Workaround for https://github.com/dotnet/roslyn/issues/46859
-                    // Do not free in presence of cancellation.
-                    if (!context.CancellationToken.IsCancellationRequested)
-                    {
-                        platformSpecificOperations.Free(context.CancellationToken);
-                    }
```
buyaa-n, line 301: `PooledConcurrentDictionary.Free()` is already checking this, don't need to double-check here https://github.com/dotnet/roslyn-analyzers/blob/5102a0d89ef178b0ca741634ee006597bf5df2ed/src/Utilities/Compiler/PooledObjects/PooledConcurrentDictionary.cs#L36-L41

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1509,43 +1497,66 @@ static void MergePlatformAttributes(ImmutableArray<AttributeData> immediateAttri
...
+                            }
+                        }
+                    }
+                    // For allow list if child narrowing supported platfroms by having less platforms support than parent,
+                    // not existing parent platforms should be removed
+                    if (notFoundPlatforms.Count > 0)
+                    {
+                        if (supportFound)
+                        {
+                            childAttributes = new SmallDictionary<string, Versions>(StringComparer.OrdinalIgnoreCase);
```
buyaa-n, line 1512: Because `SmallDictionary` doesn't have the `Remove` method added this temporary workaround, planning to add remove method to the `SmallDictionary`, then this will be updated

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1509,43 +1497,66 @@ static void MergePlatformAttributes(ImmutableArray<AttributeData> immediateAttri
...
+                            foreach (var (platform, attributes) in pAttributes)
+                            {
+                                if (!notFoundPlatforms.Contains(platform))
+                                {
+                                    childAttributes.Add(platform, attributes);
+                                }
+                            }
+                            parentAttributes.Platforms = childAttributes;
```
buyaa-n, line 1520: Because SmallDictionary doesn't have the Remove method added this temporary workaround, planning to add remove method to the SmallDictionary, then this will be updated

mavasani:
@buyaa-n just trying to understand the expected semantics for the below example you provided:

```csharp
class Test2
{
    class Test2MacOs
    {
        void WindowsOnly()
        {
            // No valid call sites; no warnings no matter what
        }
        [SupportedOSPlatform("linux")]
        void LinuxOnly()
        {
            // No valid call sites; no warnings no matter what
        }
    }
}
```

Applying `macos` SupportedOSPlatform attribute on `Test2MacOs` but not `windows` is equivalent to `[UnsupportedOSPlatform("windows")]` being applied to `Test2MacOs`. Due to this, `WindowsOnly` method with `[SupportedOSPlatform("windows")]` becomes a method with an empty set of supported attributes, hence has no valid call sites. Is this a correct analysis?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
```diff
@@ -22,14 +23,28 @@ public sealed partial class PlatformCompatibilityAnalyzer
...
             public Version? SupportedFirst { get; set; }
             public Version? SupportedSecond { get; set; }
             public Version? UnsupportedFirst { get; set; }
             public Version? UnsupportedSecond { get; set; }
-            public bool HasAttribute() => SupportedFirst != null || UnsupportedFirst != null ||
+            public bool IsSet() => SupportedFirst != null || UnsupportedFirst != null ||
                         SupportedSecond != null || UnsupportedSecond != null;
         }
+
+        private class PlatformAttributes
```
mavasani, line 36:
```suggestion
        private sealed class PlatformAttributes
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -251,12 +251,8 @@ private static bool PlatformAnalysisAllowed(AnalyzerOptions options, Compilation
...
                 try
                 {
-                    if (platformSpecificOperations.IsEmpty)
-                    {
-                        return;
-                    }
-
-                    if (guardMethods.IsEmpty || !(context.OperationBlocks.GetControlFlowGraph() is { } cfg))
+                    if (platformSpecificOperations.IsEmpty || guardMethods.IsEmpty ||
+                        !(context.OperationBlocks.GetControlFlowGraph() is { } cfg))
```
mavasani, line 255: Nit: either separate out each part of the condition into a separate line OR keep the entire condition on the same line.
```suggestion
                    if (platformSpecificOperations.IsEmpty ||
                        guardMethods.IsEmpty ||
                        !(context.OperationBlocks.GetControlFlowGraph() is { } cfg))
```

buyaa-n:
> Applying `[SupportedOSPlatform("macos")]` attribute on `Test2MacOs` but not `[SupportedOSPlatform("windows")]` is equivalent to `[UnsupportedOSPlatform("windows")]` being applied to `Test2MacOs`. Due to this, `WindowsOnly` method with `[SupportedOSPlatform("windows")]` becomes a method with an empty set of supported attributes, hence has no valid call sites. Is this a correct analysis?

Yes that is correct, it is equivalent to `[UnsupportedOSPlatform("windows")]` being applied

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
mavasani: Also tagging @CyrusNajmabadi - seems like a good candidate for your whitespace/newline formatting analyzers

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
mavasani: Yes, thanks.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1074,17 +1058,16 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
 
                     if (TryGetOrCreatePlatformAttributes(containingSymbol, true, platformSpecificMembers, out var callSiteAttributes))
                     {
-                        if (IsNotSuppressedByCallSite(operationAttributes, callSiteAttributes, msBuildPlatforms, out var notSuppressedAttributes))
+                        if (callSiteAttributes.Callsite != Callsite.Empty &&
```
mavasani, line 1061: This check is being repeated in both if and else if parts. Consider moving the check to outer if, i.e.

```csharp
if (TryGetOrCreatePlatformAttributes(containingSymbol, true, platformSpecificMembers, out var callSiteAttributes) &&
    callSiteAttributes.Callsite != Callsite.Empty)
```

buyaa-n: Good point, those are not exactly equivalent, else part should only work when there were no platform attributes found in call-site, but the `callSiteAttributes.Callsite != Callsite.Empty` in else part is redundant, leftover from my previous tries

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -3198,6 +3198,232 @@ internal static class UnsupportedWindows6_10_Ios_1_5_11
...
+        NoAttribute();
+    }
+    [SupportedOSPlatform(""iOS"")]
+    public void M3 ()
+    {
+        [|WindowsOnly()|];
+        IosOnly();
+        NoAttribute();
+    }
+}";
```
jeffhandley, line 3374: These tests are very illustrative of the behavior. :+1:

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
```diff
@@ -22,14 +23,28 @@ public sealed partial class PlatformCompatibilityAnalyzer
...
+        {
+            public PlatformAttributes() { }
+
+            public PlatformAttributes(Callsite callsite, SmallDictionary<string, Versions> platforms)
+            {
+                Callsite = callsite;
+                Platforms = platforms;
+            }
+
+            public SmallDictionary<string, Versions>? Platforms { get; set; }
```
mavasani, line 46: Is this property ever set to null via the property setter? Both the assignment sites I see below seem to guarantee the value assigned will be non-null. If that is true, I'd prefer this field be made non-nullable so all the read sites for the property don't have to add a `!`

jeffhandley:
@buyaa-n Do I interpret this change correctly in that there's no assembly-level special-casing of the empty set behavior and that this is the desired "correct" fix to achieve empty set detection?

With these changes, how many of the `dotnet/runtime` browser-build warnings does this clean up, and how many are left?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Data.cs
buyaa-n: By default, the empty constructor used for creating the attributes, and if there were no attributes found for the symbol this property will never be set. https://github.com/dotnet/roslyn-analyzers/blob/8aee460a17c3f05eca78305a00725e9480316fa9/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L1392

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -171,32 +171,32 @@ public class Test
 {
     void ThrowWithStringArgument()
     {
-        [|SR.Message|] = ""Warns not reachable on Browser"";
+        SR.Message = ""nOT Warns not reachable on Browser"";
```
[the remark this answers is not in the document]
buyaa-n: Yes, i was forgot to update messages, and later updated to "Not warn" for all, but your version is more meaningful

buyaa-n:
> @buyaa-n Do I interpret this change correctly in that there's no assembly-level special-casing of the empty set behavior and that this is the desired "correct" fix to achieve empty set detection?

Yes it does not include my previous solution with `assembly-level special-casing`, it is based on our last discussion `the desired "correct" fix to achieve empty set detection`

> With these changes, how many of the `dotnet/runtime` browser-build warnings does this clean up, and how many are left?

All are cleaned up, no one left

jeffhandley:
> Yes it does not include my previous solution with `assembly-level special-casing`, it is based on our last discussion `the desired "correct" fix to achieve empty set detection`

:100: /cc @terrajobst -- the correct fix is in here, so no need for a follow-up item

> All are cleaned up, no one left

Great!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1509,43 +1499,62 @@ static void MergePlatformAttributes(ImmutableArray<AttributeData> immediateAttri
                                     }
                                 }
                             }
-                            // other platform attributes are ignored as the list couldn't be extended
+                            else // not existing parent platforms might need to be removed
```
mavasani, line 1502: We generally prefer comments to be in a separate line inside the block. @CyrusNajmabadi another good analyzer rule for whitespace/newline related formatting style suggestions.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -3198,6 +3198,232 @@ internal static class UnsupportedWindows6_10_Ios_1_5_11
...
 
+        [Fact]
+        public async Task EmptyCallsiteReferencesLocalFunctionNotWarns()
+        {
+            var source = @"
+using System.Runtime.Versioning;
+[SupportedOSPlatform(""Browser"")]
+public class Test
+{
+    [UnsupportedOSPlatform(""browser"")]
```
mavasani, line 3209: Just curious - does the reverse annotation work? i.e. if I mark the containing assembly or type as unsupported for a platform A and just a single method/property inside it as supported, will we get warnings inside that method/property if it has guarded runtime checks within it?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -3198,6 +3198,232 @@ internal static class UnsupportedWindows6_10_Ios_1_5_11
...
+    }
+
+    [SupportedOSPlatform(""WINDOWS"")]
+    void WindowsOnly() // Nothing should warn
+    {
+        ApiUnsupportedOnBrowser();
+        Console.Beep(10, 20);
+    }
+
+    [SupportedOSPlatform(""Linux"")] // Not valid because Linux is not in parent list of platforms
```
mavasani, line 3417: Why is this the case? If I had a type such that only one of the method supported a platform, why would I be required to add the supported attribute on the whole type as opposed to just the method?

src/Utilities/Compiler/SmallDictionary.cs
```diff
@@ -68,6 +68,114 @@ private int GetHashCode(K k)
             return Comparer.GetHashCode(k);
         }
 
+        public bool IsEmpty => _root == null;
+
+        public void Remove(K key)
```
mavasani, line 73: Thank you for adding this functionality @buyaa-n! I am going to skip reviewing this helper in detail as it seems to be the standard AVL node removal algorithm. Hopefully in future, such a SmallDictionary data structure can be added to BCL and we will not require to maintain this type in both Roslyn repo and this repo.

mavasani (approved):
LGTM. Have some minor questions, but none of those seem blocking.

src/Utilities/Compiler/SmallDictionary.cs
buyaa-n: Yes, it is by the standard algorithm, manually tested with several scenarios plus new tests added with this PR also testing it. 

> Hopefully in future, such a SmallDictionary data structure can be added to BCL and we will not require to maintain this type in both Roslyn repo and this repo

Yes, I remember cc @terrajobst was pointing that out before

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
buyaa-n: > Just curious - does the reverse annotation work? i.e. if I mark the containing assembly or type as unsupported for a platform A and just a single method/property inside it as supported

Only parent attribute would work, as a child cannot extend parent support supported attribute on M2 will be ignored
```cs
[UnsupportedOSPlatform(""Browser"")] // Deny list => unsupported only on browser but supported on all others 
public class Test
{
    [SupportedOSPlatform(""browser"")] // child cannot extend parent support, annotation willl be ignored
    public void M2 ()  // the above annotation doesn't produce empty set as parent only unsupported on browser but supported on all other 
    {
         // All platforms except browser can access here, so we would warn for platform-specific APIs usage 
    }
}
```
> will we get warnings inside that method/property if it has guarded runtime checks within it?

Good question, this combination is not producing the empty set, so we should warn

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
buyaa-n: Right, as you mentioned it is not a meaningful or realistic scenario, but it is happening in multitargeted builds for runtime repo. We have assemblies for multiple platforms where some APIs annotated as platform-specific, the assembly doesn't have any platform attributes but it does build for each platform specified: 
```xml
 <TargetFrameworks>net5.0-windows;net5.0-Unix;net5.0-tvOS;net5.0-Browser</TargetFrameworks>
```
For example, when it builds for windows `net5.0-windows` target MSBuild adds assembly-level [SupportedOSPlatform("windows")] attributes then the individual APIs having different platforms support would produce this scenario and we see warnings for those targeted builds, this is the main reason we are doing this change

PR #4741, opened 2021-01-27
CA1802: Don't report diagnostic for preview feature constant interpolated strings

### LINKED ISSUE #4732: CA1802 false positive on static readonly interpolated strings in Visual Studio 16.9 Preview 3 targeting net472
🔗 Also filed as [AB#1287517](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1287517)

### Analyzer

**Diagnostic ID**: [CA1802](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1802): `Use Literals Where Appropriate`

### Analyzer source

**Version**: 5.0.3 (Latest)

### Describe the bug

CA1802 is false positively triggering on static readonly strings in Visual Studio 16.9 Preview 3 targeting net472

Similar reported issue: #1175

### Steps To Reproduce

Create a project with the following files in Visual Studio 16.9 Preview 3:

ClassLibrary1.csproj
```
    <PropertyGroup>
        <TargetFramework>net472</TargetFramework>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" />
    </ItemGroup>
</Project>
```

.editorconfig
```
[*.cs]
dotnet_diagnostic.CA1802.severity = error
```

Class1.cs
```
namespace ClassLibrary1
{
    public class Class1
    {
        private const string foo = "foo";
        private static readonly string fooBar = $"{foo}bar";
    }
}
```

In the above class, CA1802 triggers on line `private static readonly string fooBar = $"{foo}bar";`

However, when changing the line to the suggested `private const string fooBar = $"{foo}bar";`, we get a new error:
```
Error CS8652 : The feature 'constant interpolated strings' is currently in Preview and *unsupported*. To use Preview features, use the 'preview' language version.
```

Enabling `preview` language version resolves the `CS8652` as expected, but I would like to use the default language version.

### Expected behavior

CA1802 should not trigger here

### Actual behavior

CA 1802 triggers here

## Additional context

Things tested that do not trigger CA1802 (as expected):
- CA1802 does not appear to be trigger using Visual Studio 16.9 Preview 2 or 16.8.4 (stable)
- CA1802 does not appear to be trigger when targeting `netstandard2.0`, `netcoreapp2.0`, or `net5.0` in the .csproj file

Youssef1313:
Fixes #4732


eng/Versions.props
```diff
@@ -30,7 +30,7 @@
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
     <MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>3.7.0</MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>
-    <MicrosoftCodeAnalysisVersionForTests>3.8.0</MicrosoftCodeAnalysisVersionForTests>
+    <MicrosoftCodeAnalysisVersionForTests>3.9.0-3.final</MicrosoftCodeAnalysisVersionForTests>
```
Youssef1313, line 33: Version 3.8.0 doesn't produce operation tree where interpolated string has constant value. I had to update to be able to test.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/UseLiteralsWhereAppropriate.cs
```diff
@@ -95,5 +101,12 @@ public override void Initialize(AnalysisContext context)
                 OperationKind.FieldInitializer);
             });
         }
+
+        private static bool IsConstantInterpolatedStringSupported()
+        {
+            // TODO: When constant interpolated string is supported in a stable language version (most likely C# 10), this method should be updated.
+            // The feature is currently available for preview only.
+            return false;
+        }
```
Youssef1313, line 110: I didn't return true for preview version based on https://github.com/dotnet/roslyn-analyzers/issues/4732#issuecomment-767891878

When this is implemented later, it will need to be abstract with language-specific implementation.

A possible another approach would be to introduce a new service that gives information about features.

I'd love to be able to do the following (or similar):

```csharp
var service = compilation.GetService<ILanguageFeatureSupportService>();
if (service.IsConstantInterpolationSupported(compilation)
{
    // do something.
}
```

Youssef1313:
@mavasani Test failure is for another analyzer CA1834. This happened because the operation tree changed after updating the compiler used in tests. Can you give it a look? Both behaviors can be reasoned about, so I don't know what will be your preference.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/UseLiteralsWhereAppropriate.cs
```diff
@@ -95,5 +101,12 @@ public override void Initialize(AnalysisContext context)
                 OperationKind.FieldInitializer);
             });
         }
+
+        private static bool IsConstantInterpolatedStringSupported()
+        {
+            // TODO: When constant interpolated string is supported in a stable language version (most likely C# 10), this method should be updated.
```
mavasani, line 107: Can you instead check the LanguageVersion from the Parse options and get rid of the TODO? You would need to add analyzer sub-types to C# and VB layers to access the parse options.

mavasani (changes requested):
Lets address the TODO by checking the parse options.

mavasani (approved):
Thanks!

mavasani:
@Youssef1313 tests are failing

Youssef1313:
@mavasani Failing test is for another analyzer. Could you check what should be the correct behavior for it?

mavasani:
> Failing test is for another analyzer. Could you check what should be the correct behavior for it?

@Youssef1313 You need to update the failing test to expect an analyzer diagnostic given that with the new MS.CA package for testing, interpolated string that is made up of constant constituent parts is also considered a constant now.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringTests.cs
```diff
@@ -493,11 +493,12 @@ private void TestMethod(int value)
...
             StringBuilder sb = new StringBuilder();
             const string ch = ""a"";
-            sb.Append($""{ch}"");
+            sb.Append([|$""{ch}""|]);
         } 
     } 
 }";
-            await VerifyCS.VerifyAnalyzerAsync(interpolatedString_cs);
+
+            await VerifyCS.VerifyCodeFixAsync(interpolatedString_cs, interpolatedString_cs);
```
Youssef1313, line 501: @mavasani No fix is offered for this edge case. Is this behavior okay or should I file an issue?

mavasani: I think it is fine.

mavasani (approved):

PR #4764, opened 2021-02-02
(3/4) Use Span-Based String Concat Analyzer and Fixer

NewellClark:
Fixes [issue #33777](https://github.com/dotnet/runtime/issues/33777). 
Issues in dotnet/runtime have been addressed.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseSpanBasedStringConcatDescription" xml:space="preserve">
+    <value>'Substring' is O(n) in the length of the substring, whereas 'AsSpan' is O(1). When using the result of 'Substring' in a string concatenation, it is more efficient to instead call 'AsSpan' and use 'string.Concat', instead of 'Substring' and a concatenation operator. </value>
+  </data>
+  <data name="UseSpanBasedStringConcatMessage" xml:space="preserve">
+    <value>Use span-based string.Concat and 'AsSpan' instead of 'Substring'</value>
```
mavasani, line 1517:
```suggestion
    <value>Use span-based 'string.Concat' and 'AsSpan' instead of 'Substring'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,13 @@
...
+  </data>
+  <data name="UseSpanBasedStringConcatTitle" xml:space="preserve">
+    <value>Use span-based string.Concat</value>
```
mavasani, line 1520:
```suggestion
    <value>Use span-based 'string.Concat'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,253 @@
...
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using Resx = Microsoft.NetCore.Analyzers.MicrosoftNetCoreAnalyzersResources;
+using Analyzer.Utilities.PooledObjects;
+using System.Collections.Generic;
+using System.Linq;
+using System.Diagnostics.CodeAnalysis;
```
Youssef1313, line 14: nit: Group all `System.*` usings to the top.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
```diff
@@ -0,0 +1,829 @@
...
+        public Task NestedViolations_AreReportedAndFixed_VB(string testStatements, string fixedStatements, int[] locations)
+        {
+            var test = new VerifyVB.Test
+            {
+                TestCode = VBUsings + VBWithBody(testStatements),
+                FixedCode = VBUsings + VBWithBody(fixedStatements),
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                NumberOfIncrementalIterations = -10,
+                NumberOfFixAllIterations = -10,
+                NumberOfFixAllInDocumentIterations = -10
```
sharwell, line 326: 📝 If the default number of iterations isn't working, you should use _positive_ numbers for the replacement. Otherwise it's impossible to tell from reading the test exactly what the unexpected behavior will feel like to a user.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,253 @@
...
+                }
+
+                void ReportDiagnosticsOnRootConcatOperationsWithSubstringCalls(OperationBlockAnalysisContext context)
+                {
+                    //  We report diagnostics for all top-most concat operations that contain substring invocations
+                    //  when there is an applicable span-based overload of string.Concat
+                    foreach (var operation in topMostConcatOperations)
+                    {
+                        var chain = FlattenBinaryOperation(operation);
+                        if (chain.Any(IsAnySubstringInvocation) && symbols.TryGetRoscharConcatMethodWithArity(chain.Length, out var _))
```
Youssef1313, line 92: nit:
```suggestion
                        if (chain.Any(IsAnySubstringInvocation) && symbols.TryGetRoscharConcatMethodWithArity(chain.Length, out _))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
```diff
@@ -0,0 +1,175 @@
...
+                    .WithTrailingTrivia(trailingTrivia);
+
+                var newRoot = generator.ReplaceNode(root, concatExpressionSyntax, concatMethodInvocationSyntax);
+
+                //  Make sure 'System' namespace is imported.
+                if (!IsSystemNamespaceImported(generator.GetNamespaceImports(newRoot)))
+                {
+                    var systemNamespaceImport = generator.NamespaceImportDeclaration(nameof(System));
+                    newRoot = generator.AddNamespaceImports(newRoot, systemNamespaceImport);
+                }
```
Youssef1313, line 108: I think it's automatically imported in VB. The following (without `Imports System`) compiles on my machine:

```vb
Module Program
    Sub Main(args As String())
        Console.WriteLine("Hello World!")
    End Sub
End Module
```

If I added the extra `Imports System`, Visual Studio fades it out as unnecessary:

image

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,253 @@
...
+            }
+
+            public static bool TryGetSymbols(Compilation compilation, out RequiredSymbols symbols)
+            {
+                var stringType = compilation.GetSpecialType(SpecialType.System_String);
+                var roscharType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemReadOnlySpan1)
+                    ?.Construct(compilation.GetSpecialType(SpecialType.System_Char));
+                var memoryExtensionsType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemMemoryExtensions);
+
+                if (roscharType is null || memoryExtensionsType is null)
```
Youssef1313, line 176: I'd be more pessimistic and check against `stringType` for null. It's unlikely to be null, but I think it may be.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
```diff
@@ -0,0 +1,829 @@
...
+        public Task SingleViolationInOneBlock_ReportedAndFixed_CS(string testStatements, string fixedStatements)
+        {
+            var test = new VerifyCS.Test
+            {
+                TestCode = CSUsings + CSWithBody(testStatements),
+                FixedCode = CSUsings + CSWithBody(fixedStatements),
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                ExpectedDiagnostics = { VerifyCS.Diagnostic(Rule).WithLocation(0) }
+            };
+            return test.RunAsync();
```
Youssef1313, line 87: Consider making the tests `async`, then `await test.RunAsync()`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
```diff
@@ -0,0 +1,829 @@
...
+object thing = bar;
+var s = {0};";
+            var culture = CultureInfo.InvariantCulture;
+            string testStatements = string.Format(culture, format, $@"{{|#0:{testExpression}|}}");
+            string fixedStatements = string.Format(culture, format, fixedExpression);
+
+            var test = new VerifyCS.Test
+            {
+                TestState = { Sources = { CSUsings + CSWithBody(testStatements), helperTypes } },
+                FixedState = { Sources = { CSUsings + CSWithBody(fixedStatements), helperTypes } },
```
Youssef1313, line 581: Was it intentional to use TestState and FixedState instead of TestCode and FixedCode in this specific test?

Youssef1313 (commented):
Consider having tests against null-conditional access operator (`?.`).

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
NewellClark: Yes, because I defined some helper types in a separate source file.

Youssef1313: @NewellClark Good. I didn't notice that.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
NewellClark: `public async Task TestMethod() => await test.RunAsync();` is semantically identical to 
`public Task TestMethod() => return test.RunAsync();`, but potentially slower. If we needed to await multiple async calls, it would make sense to mark the method as `async`, but since we have just that one `async` call, it makes sense to simply return the `Task` produced by `test.RunAsync()`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
NewellClark: Good to know. IIRC I think I set it to -10 because the order in which the violations were being fixed was inconsistent, which made it sometimes take different numbers of iterations. Is there any way to tell the test framework which order to apply the fixes in?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
NewellClark: It's weird; when I build a project in Visual Studio, the `Imports System` statement isn't necessary. But when I don't import the namespace in the analyzer test, I get compiler error BC0456 "'AsSpan' is not a member of 'String'". Is the auto-importing of `System` a compiler switch in VB? I'll look into this some more.

NewellClark: I figured it out. In VB, you can define "implicitly imported" namespaces for your project. Those namespaces will be considered to be imported for all files in the project. `System` is imported by default in visual studio, but apparently it's not imported by default in the test projects. I'm going to fix the analyzer so it detects implicitly-imported namespaces, and write tests that ensure `Imports System` is only added when it is not imported, either explicitly or implicitly.

NewellClark:
> Consider having tests against null-conditional access operator (`?.`).

Do you mean for `Substring` invocations, as in `foo?.Substring(1) + bar`?

Youssef1313:
> > Consider having tests against null-conditional access operator (`?.`).
> 
> @Youssef1313 Do you mean for `Substring` invocations, as in `foo?.Substring(1) + bar`?

Yes.

NewellClark:
> > > Consider having tests against null-conditional access operator (`?.`).
> > 
> > 
> > @Youssef1313 Do you mean for `Substring` invocations, as in `foo?.Substring(1) + bar`?
> 
> Yes.

What should the fixer do for `foo?.Substring(1)`? Because the span-based equivalent is something like 
`foo is null ? default : foo.AsSpan(1)`, which seems kind of verbose for the fixer to be inserting. Should we ignore cases where `Substring` is accessed conditionally?

Youssef1313:
> > > > Consider having tests against null-conditional access operator (`?.`).
> > > 
> > > 
> > > @Youssef1313 Do you mean for `Substring` invocations, as in `foo?.Substring(1) + bar`?
> > 
> > 
> > Yes.
> 
> What should the fixer do for `foo?.Substring(1)`? Because the span-based equivalent is something like
> `foo is null ? default : foo.AsSpan(1)`, which seems kind of verbose for the fixer to be inserting. Should we ignore cases where `Substring` is accessed conditionally?

I'm not sure whether the fixer should do something or not. But I think at least the analyzer should flag these.

@mavasani @bartonjs What do you think?

bartonjs:
> I'm not sure whether the fixer should do something or not. But I think at least the analyzer should flag these.

That's a reasonable starting position.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: fixed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcatTests.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Fixed. Didn't know you could do that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Fixed.

NewellClark:
I've made the requested changes. 

The fixer will now **only** add `Imports System` in VB if it is not specified as an implicit global import. `System` is imported by default in projects created in Visual Studio, but it can be toggled off, so I've added tests ensuring the fixer always does the right thing regardless of user configs. 

Analyzer detects `foo?.Substring(1) + whatever` but fixer won't fix it, as previously discussed.

NewellClark:
How do we want to handle character arguments? I was thinking `stackalloc char[] { charVar }`, but that could potentially run into [CA2014](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2014) if the concatenation is inside a loop.

stephentoub:
> How do we want to handle character arguments?

Can you provide an example?

NewellClark:
> > How do we want to handle character arguments?
> 
> Can you provide an example?

Sure. `foo.Substring(1) + charVar`. At present, that gets turned into `string.Concat(foo.AsSpan(1), charVar.ToString())`. It would be faster to convert it to `string.Concat(foo.AsSpan(1), stackalloc char[] { charVar })`, but that could potentially run into [CA2014](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2014), if the expression happened to be in a loop.

stephentoub:
> foo.Substring(1) + charVar

If `charVar` is a literal, we should make it a string literal.

If `charVar` is a variable, I don't think the rule should fire.

stephentoub:
> that gets turned into string.Concat(foo.AsSpan(1), charVar.ToString())

We definitely don't want that :wink:

And as you say, it shouldn't stackalloc.

If this were code in dotnet/runtime, we would likely use `MemoryMarshal.CreateReadOnlySpan(ref charVar, 1)`, but I don't want a fixer to generate that code, as it's too advanced / unsafe for general consumption.

I don't really see the variable char case as being different from a variable of any other type.

NewellClark:
> I don't really see the variable char case as being different from a variable of any other type.

Do you want the rule to only fire when all operands are strings, with the exception of character literals? Currently, it converts non-string operands by calling `ToString` on them, which I figured made sense for most objects, since that's what concatenation does anyway.

stephentoub:
> Currently, it converts non-string operands by calling ToString on them, which I figured made sense for most objects, since that's what concatenation does anyway.

I want to be careful about causing developers to write code that makes it perform better as a moment-in-time improvement but then actually hurts things in the long run.  For example, today, yes, string.Concat(object, object, object) will calls ToString on each argument.  But we might change that.  If you look at string.Format, for example, it doesn't just always call ToString on each argument: if an argument implements ISpanFormattable, it will try to use its TryFormat to write directly into the underlying buffer.  If this analyzer got developers to change their code to explicitly ToString, then we'd lose the benefits of such a change in the future, while also having them write more complicated code.

> So you want the rule to only fire when all operands are strings, with the exception of character literals?

I think so, yes.  Character literals are already special-cased by the compiler in this regard:
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgZgAIS6BhOgbyrs4fsNIAY6AWQAUvAdgB2ABwCuGAJR0AvAD46k2RjoBqOgHIwegNxUAvkA=
so it seems reasonable to do so here as well.

NewellClark (commented):
What should the fixer do when we have both conditional and non-conditional substring calls in the same concatenation?
`foo.Substring(1) + bar?.Substring(1)`
Should the fixer provide the partial fix:
`string.Concat(foo.AsSpan(1), bar?.Substring(1)`, 
or should we simply report without fixing if _any_ substring call is conditional?

NewellClark:
I've applied changes requested by @Evangelink on [my other PR](https://github.com/dotnet/roslyn-analyzers/pull/4806).

mavasani (changes requested):
This PR needs to be rebased to preview4 branch

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,16 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseSpanBasedStringConcatDescription" xml:space="preserve">
+    <value>'Substring' is O(n) in the length of the substring, whereas 'AsSpan' is O(1). When using the result of 'Substring' in a string concatenation, it is more efficient to instead call 'AsSpan' and use 'string.Concat', instead of 'Substring' and a concatenation operator.</value>
```
```suggestion
    <value>It is more efficient to use 'AsSpan' and 'string.Concat', instead of 'Substring' and a concatenation operator.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,16 @@
...
+  </data>
+  <data name="UseSpanBasedStringConcatMessage" xml:space="preserve">
+    <value>Use span-based 'string.Concat' and 'AsSpan' instead of 'Substring'</value>
+  </data>
+  <data name="UseSpanBasedStringConcatTitle" xml:space="preserve">
+    <value>Use span-based 'string.Concat'</value>
+  </data>
+  <data name="UseSpanBasedStringConcatCodeFixTitle" xml:space="preserve">
+    <value>Replace concatenation with 'string.Concat' and replace 'Substring' with 'AsSpan'</value>
```
mavasani, line 1523: Code fix title should be succinct
```suggestion
    <value>Use 'AsSpan' with 'string.Concat'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+            s_localizableMessage,
+            DiagnosticCategory.Performance,
+            RuleLevel.IdeSuggestion,
+            s_localizableDescription,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        private const string SubstringName = nameof(string.Substring);
+        private const string AsSpanName = nameof(MemoryExtensions.AsSpan);
+        private const string ConcatName = nameof(string.Concat);
```
mavasani, line 38: Do we really need these fields? Can we just inline the nameof expressions?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+
+        // Use readonly struct instead of record type to save on allocations, since it's not passed by-value.
+        // We aren't comparing these.
+#pragma warning disable CA1815 // Override equals and operator equals on value types
+        internal readonly struct RequiredSymbols
+#pragma warning restore CA1815 // Override equals and operator equals on value types
+        {
+            private RequiredSymbols(
+                INamedTypeSymbol stringType, INamedTypeSymbol roscharType,
+                IMethodSymbol substring1, IMethodSymbol substring2,
```
mavasani, line 208: Kindly use descriptive symbol names
```suggestion
                IMethodSymbol substringStart, IMethodSymbol substringStartLength,
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+                IMethodSymbol asSpan1, IMethodSymbol asSpan2)
```
mavasani, line 209:
```suggestion
                IMethodSymbol asSpanStart, IMethodSymbol asSpanStartLength)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+                ReadOnlySpanOfCharType = roscharType;
+                SubstringStart = substring1;
+                SubstringStartLength = substring2;
+                AsSpanStart = asSpan1;
+                AsSpanStartLength = asSpan2;
+
+                RoslynDebug.Assert(
+                    StringType is not null && ReadOnlySpanOfCharType is not null &&
+                    SubstringStart is not null && SubstringStartLength is not null &&
+                    AsSpanStart is not null && AsSpanStartLength is not null);
```
mavasani, line 221: I don't believe this is required - the repo is nullable enabled and defining parameters without `?` ensures that a potential value passed in at callsite is flagged by the compiler with nullable warning.
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+            context.RegisterCompilationStartAction(OnCompilationStart);
+        }
+
+        private void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            if (!RequiredSymbols.TryGetSymbols(context.Compilation, out RequiredSymbols symbols))
+                return;
+
+            context.RegisterOperationBlockStartAction(OnOperationBlockStart);
+
```
mavasani, line 62: Please add explicit return statement for control flow clarity.
```suggestion
            context.RegisterOperationBlockStartAction(OnOperationBlockStart);
            return;

            // Local functions
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+            void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
+            {
+                //  Maintain set of all top-most concat operations so we don't report sub-expressions of an
+                //  already-reported violation. 
+                //  We also don't report any diagnostic if the concat operation has too many operands for the span-based
+                //  Concat overloads to handle.
+                var topMostConcatOperations = PooledConcurrentSet<IBinaryOperation>.GetInstance();
```
mavasani, line 69: Please register an OperationBlockEnd action to dispose this pooled set

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+
+                context.RegisterOperationAction(PopulateTopMostConcatOperations, OperationKind.Binary);
+                context.RegisterOperationBlockEndAction(ReportDiagnosticsOnRootConcatOperationsWithSubstringCalls);
+
+                void PopulateTopMostConcatOperations(OperationAnalysisContext context)
+                {
+                    //  If the current operation is a string-concatenation operation, walk up to the top-most concat
+                    //  operation and add it to the set.
+                    var binary = (IBinaryOperation)context.Operation;
+                    if (!TryGetTopMostConcatOperation(binary, out var topMostConcatOperation))
```
mavasani, line 79: This approach seems fine, but seems you can do without `topMostConcatOperations` and reporting in the end action. Instead you can just have the following in `IBinaryOperation` callback:

```csharp
  if (IsTopMostConcatOperation(binary) && ShouldBeReported(binary))
  {
      context.ReportDiagnostic(binary.CreateDiagnostic(Rule));
  }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+            bool IsAnyDirectOrConditionalSubstringInvocation(IOperation operation)
+            {
+                if (operation is IConditionalAccessOperation conditionallAccessOperation)
+                    operation = conditionallAccessOperation.WhenNotNull;
+
+                return operation is IInvocationOperation invocation && symbols.IsAnySubstringMethod(invocation.TargetMethod);
+            }
+        }
+
+        internal static ImmutableArray<IOperation> FlattenBinaryOperation(IBinaryOperation root)
```
mavasani, line 140: Might be much simpler to just write a specialized [OperationWalker](http://sourceroslyn.io/#Microsoft.CodeAnalysis/Operations/OperationWalker.cs,11) here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            if (operand is not IConversionOperation conversion)
+                return operand;
+            if (!conversion.IsImplicit || conversion.Conversion.IsUserDefined)
+                return conversion;
+
+            switch (conversion.Language)
+            {
+                case LanguageNames.CSharp when conversion.Type.SpecialType is SpecialType.System_Object:
+                case LanguageNames.VisualBasic when conversion.Type.SpecialType is SpecialType.System_Object or SpecialType.System_String:
```
mavasani, line 193: Please move all language specific logic to language specific layers instead of checking language names - you already have similar abstract methods and language specific operation walking code, so please move this out as well.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
```diff
@@ -0,0 +1,165 @@
...
+            var cancellationToken = context.CancellationToken;
+            var model = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+            var compilation = model.Compilation;
+            SyntaxNode root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+
+            if (!RequiredSymbols.TryGetSymbols(compilation, out var symbols))
+                return;
+            if (root.FindNode(context.Span, getInnermostNodeForTie: true) is not SyntaxNode concatExpressionSyntax)
+                return;
+            //  OperatorKind will be BinaryOperatorKind.Concatenate, even when '+' is used instead of '&' in Visual Basic.
```
mavasani, line 51: Nit: newline before comment
```suggestion

            //  OperatorKind will be BinaryOperatorKind.Concatenate, even when '+' is used instead of '&' in Visual Basic.
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
```diff
@@ -0,0 +1,165 @@
...
+            if (model.GetOperation(concatExpressionSyntax, cancellationToken) is not IBinaryOperation concatOperation || concatOperation.OperatorKind is not (BinaryOperatorKind.Add or BinaryOperatorKind.Concatenate))
+                return;
+
+            var operands = UseSpanBasedStringConcat.FlattenBinaryOperation(concatOperation);
+            //  Bail out if we don't have a long enough span-based string.Concat overload.
```
mavasani, line 56:
```suggestion

            //  Bail out if we don't have a long enough span-based string.Concat overload.
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
```diff
@@ -0,0 +1,165 @@
...
+            var model = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+            var compilation = model.Compilation;
+            SyntaxNode root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+
+            if (!RequiredSymbols.TryGetSymbols(compilation, out var symbols))
+                return;
+            if (root.FindNode(context.Span, getInnermostNodeForTie: true) is not SyntaxNode concatExpressionSyntax)
+                return;
+            //  OperatorKind will be BinaryOperatorKind.Concatenate, even when '+' is used instead of '&' in Visual Basic.
+            if (model.GetOperation(concatExpressionSyntax, cancellationToken) is not IBinaryOperation concatOperation || concatOperation.OperatorKind is not (BinaryOperatorKind.Add or BinaryOperatorKind.Concatenate))
```
mavasani, line 52: Please break down the expression into separate lines for readability

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
```diff
@@ -0,0 +1,165 @@
...
+            if (!operands.Any(IsAnyNonConditionalSubstringInvocation))
+                return;
+
+            var codeAction = CodeAction.Create(
+                Resx.UseSpanBasedStringConcatCodeFixTitle,
+                FixConcatOperationChain,
+                Resx.UseSpanBasedStringConcatCodeFixTitle);
+            context.RegisterCodeFix(codeAction, diagnostic);
+
+            bool IsAnyNonConditionalSubstringInvocation(IOperation operation)
```
mavasani, line 70:
```suggestion
            return;

            // Local functions
            bool IsAnyNonConditionalSubstringInvocation(IOperation operation)
```

mavasani (commented):
Overall LGTM. Thanks for extensive tests.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: The problem with this approach is that we may end up reporting the same violation multiple times. I remember encountering this issue while writing the analyzer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Okay, I'll reimplement this using an `OperationWalker`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
NewellClark: Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Inlined.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: You're right. I've done that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.cs
NewellClark: Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
NewellClark: Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseSpanBasedStringConcat.Fixer.cs
```diff
@@ -0,0 +1,174 @@
...
+                    .WithTrailingTrivia(trailingTrivia);
+
+                SyntaxNode newRoot = generator.ReplaceNode(root, concatExpressionSyntax, concatMethodInvocationSyntax);
+
+                //  Import 'System' namespace if it's absent.
+                if (!IsSystemNamespaceImported(context.Document.Project, generator.GetNamespaceImports(newRoot)))
+                {
+                    SyntaxNode systemNamespaceImport = generator.NamespaceImportDeclaration(nameof(System));
+                    newRoot = generator.AddNamespaceImports(newRoot, systemNamespaceImport);
+                }
```
Youssef1313, line 118: Would you be able to check if calling `WithAddImportsAnnotation` on `concatMethodInvocationSyntax` eliminate the need of `IsSystemNamespaceImported`?

NewellClark: `WithAddImportsAnnotation` only works for syntax nodes that represent types (just tested it).
I did change the VB fixer to use a case-insensitive comparison (good catch).

mavasani:
Same request and apologies as https://github.com/dotnet/roslyn-analyzers/pull/4726#issuecomment-842094914. This needs to be targeted to `release\6.0.1xx`. Thanks!

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -9,6 +9,11 @@ CA1841 | Performance | Info | PreferDictionaryContainsMethods, [Documentation](h
 CA1842 | Performance | Info | DoNotUseWhenAllOrWaitAllWithSingleArgument, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1842)
 CA1843 | Performance | Info | DoNotUseWhenAllOrWaitAllWithSingleArgument, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1843)
 
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
```
mavasani, line 14:
```suggestion
```

mavasani (approved):

PR #4770, opened 2021-02-03
Resources pass for code quality analyzers (add single ticks around placeholders)

Youssef1313:

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
@@ -118,58 +118,58 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="AvoidAsyncVoidTitle" xml:space="preserve">
-    <value>Avoid Async Void</value>
+    <value>Avoid 'async void'.</value>
   </data>
   <data name="AvoidAsyncVoidDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="AvoidAsyncVoidMessage" xml:space="preserve">
-    <value>Avoid Async Void</value>
+    <value>Avoid 'async void'.</value>
   </data>
   <data name="AsyncMethodNamesShouldEndInAsyncTitle" xml:space="preserve">
-    <value>Async Method Names Should End in Async</value>
+    <value>Async method names should end in 'Async'</value>
   </data>
   <data name="AsyncMethodNamesShouldEndInAsyncDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="AsyncMethodNamesShouldEndInAsyncMessage" xml:space="preserve">
-    <value>Async Method Names Should End in Async</value>
+    <value>Async method names should end in 'Async'</value>
   </data>
   <data name="DoNotPassAsyncLambdasAsVoidReturningDelegateTypesTitle" xml:space="preserve">
-    <value>Don't Pass Async Lambdas as Void Returning Delegate Types</value>
+    <value>Don't pass async lambdas as void returning delegate types</value>
   </data>
   <data name="DoNotPassAsyncLambdasAsVoidReturningDelegateTypesDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="DoNotPassAsyncLambdasAsVoidReturningDelegateTypesMessage" xml:space="preserve">
-    <value>Don't Pass Async Lambdas as Void Returning Delegate Types</value>
+    <value>Don't pass async lambdas as void returning delegate types</value>
   </data>
   <data name="DoNotStoreAsyncLambdasAsVoidReturningDelegateTypesTitle" xml:space="preserve">
-    <value>Don't Store Async Lambdas as Void Returning Delegate Types</value>
+    <value>Don't store async lambdas as void returning delegate types</value>
   </data>
   <data name="DoNotStoreAsyncLambdasAsVoidReturningDelegateTypesDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="DoNotStoreAsyncLambdasAsVoidReturningDelegateTypesMessage" xml:space="preserve">
-    <value>Don't Store Async Lambdas as Void Returning Delegate Types</value>
+    <value>Don't store async lambdas as void returning delegate types</value>
   </data>
   <data name="PropagateCancellationTokensWhenPossibleTitle" xml:space="preserve">
-    <value>Propagate CancellationTokens When Possible</value>
+    <value>Propagate 'CancellationToken's when possible</value>
   </data>
   <data name="PropagateCancellationTokensWhenPossibleDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="PropagateCancellationTokensWhenPossibleMessage" xml:space="preserve">
-    <value>Propagate CancellationTokens When Possible</value>
+    <value>Propagate 'CancellationToken's when possible</value>
   </data>
   <data name="DoNotMixBlockingAndAsyncTitle" xml:space="preserve">
-    <value>Don't Mix Blocking and Async</value>
+    <value>Don't mix blocking and async</value>
   </data>
   <data name="DoNotMixBlockingAndAsyncDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="DoNotMixBlockingAndAsyncMessage" xml:space="preserve">
-    <value>Don't Mix Blocking and Async</value>
+    <value>Don't mix blocking and async</value>
   </data>
   <data name="TypesThatOwnDisposableFieldsShouldBeDisposableTitle" xml:space="preserve">
     <value>Types that own disposable fields should be disposable</value>
@@ -208,13 +208,13 @@
     <value>The default value of an uninitialized enumeration, just as other value types, is zero. A nonflags-attributed enumeration should define a member by using the value of zero so that the default value is a valid value of the enumeration. If an enumeration that has the FlagsAttribute attribute applied defines a zero-valued member, its name should be ""None"" to indicate that no values have been set in the enumeration.</value>
   </data>
   <data name="EnumsShouldHaveZeroValueMessageFlagsRename" xml:space="preserve">
-    <value>In enum {0}, change the name of {1} to 'None'</value>
+    <value>In enum '{0}', change the name of '{1}' to 'None'</value>
   </data>
   <data name="EnumsShouldHaveZeroValueMessageFlagsMultipleZeros" xml:space="preserve">
-    <value>Remove all members that have the value zero from {0} except for one member that is named 'None'</value>
+    <value>Remove all members that have the value zero from '{0}' except for one member that is named 'None'</value>
   </data>
   <data name="EnumsShouldHaveZeroValueMessageNotFlagsNoZeroValue" xml:space="preserve">
-    <value>Add a member to {0} that has a value of zero with a suggested name of 'None'</value>
+    <value>Add a member to '{0}' that has a value of zero with a suggested name of 'None'</value>
   </data>
   <data name="AbstractTypesShouldNotHaveConstructorsTitle" xml:space="preserve">
     <value>Abstract types should not have public constructors</value>
@@ -229,10 +229,10 @@
     <value>Mark assemblies with CLSCompliant</value>
   </data>
   <data name="MarkAssembliesWithClsCompliantDescription" xml:space="preserve">
-    <value>The Common Language Specification (CLS) defines naming restrictions, data types, and rules to which assemblies must conform if they will be used across programming languages. Good design dictates that all assemblies explicitly indicate CLS compliance by using CLSCompliantAttribute . If this attribute is not present on an assembly, the assembly is not compliant.</value>
+    <value>The Common Language Specification (CLS) defines naming restrictions, data types, and rules to which assemblies must conform if they will be used across programming languages. Good design dictates that all assemblies explicitly indicate CLS compliance by using 'CLSCompliantAttribute' . If this attribute is not present on an assembly, the assembly is not compliant.</value>
   </data>
   <data name="MarkAssembliesWithClsCompliantMessage" xml:space="preserve">
-    <value>Mark assemblies with CLSCompliant</value>
+    <value>Mark assemblies with 'CLSCompliant'</value>
   </data>
   <data name="MarkAssembliesWithAssemblyVersionTitle" xml:space="preserve">
     <value>Mark assemblies with assembly version</value>
@@ -244,28 +244,28 @@
     <value>Mark assemblies with assembly version</value>
   </data>
   <data name="MarkAssembliesWithComVisibleTitle" xml:space="preserve">
-    <value>Mark assemblies with ComVisible</value>
+    <value>Mark assemblies with 'ComVisible'</value>
   </data>
   <data name="MarkAssembliesWithComVisibleDescription" xml:space="preserve">
     <value>ComVisibleAttribute determines how COM clients access managed code. Good design dictates that assemblies explicitly indicate COM visibility. COM visibility can be set for the whole assembly and then overridden for individual types and type members. If this attribute is not present, the contents of the assembly are visible to COM clients.</value>
   </data>
   <data name="MarkAssembliesWithComVisibleMessageNoAttribute" xml:space="preserve">
-    <value>Because {0} exposes externally visible types, mark it with ComVisible(false) at the assembly level and then mark all types within the assembly that should be exposed to COM clients with ComVisible(true).</value>
+    <value>Because '{0}' exposes externally visible types, mark it with 'ComVisible(false)' at the assembly level and then mark all types within the assembly that should be exposed to COM clients with 'ComVisible(true)'.</value>
   </data>
   <data name="MarkAssembliesWithComVisibleMessageAttributeTrue" xml:space="preserve">
     <value>Consider changing the ComVisible attribute on {0} to false, and opting in at the type level.</value>
   </data>
   <data name="MarkAttributesWithAttributeUsageTitle" xml:space="preserve">
-    <value>Mark attributes with AttributeUsageAttribute</value>
+    <value>Mark attributes with 'AttributeUsageAttribute'</value>
   </data>
   <data name="MarkAttributesWithAttributeUsageDescription" xml:space="preserve">
-    <value>When you define a custom attribute, mark it by using AttributeUsageAttribute to indicate where in the source code the custom attribute can be applied. The meaning and intended usage of an attribute will determine its valid locations in code.</value>
+    <value>When you define a custom attribute, mark it by using 'AttributeUsageAttribute' to indicate where in the source code the custom attribute can be applied. The meaning and intended usage of an attribute will determine its valid locations in code.</value>
   </data>
   <data name="MarkAttributesWithAttributeUsageMessageDefault" xml:space="preserve">
-    <value>Specify AttributeUsage on {0}</value>
+    <value>Specify AttributeUsage on '{0}'</value>
   </data>
   <data name="MarkAttributesWithAttributeUsageMessageInherited" xml:space="preserve">
-    <value>Even though attribute {0} inherits AttributeUsage from its base type, you should consider explicitly specifying AttributeUsage on the type to improve code readability and documentation.</value>
+    <value>Even though attribute '{0}' inherits 'AttributeUsage' from its base type, you should consider explicitly specifying 'AttributeUsage' on the type to improve code readability and documentation.</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsTitle" xml:space="preserve">
     <value>Define accessors for attribute arguments</value>
@@ -274,31 +274,31 @@
     <value>Attributes can define mandatory arguments that must be specified when you apply the attribute to a target. These are also known as positional arguments because they are supplied to attribute constructors as positional parameters. For every mandatory argument, the attribute should also provide a corresponding read-only property so that the value of the argument can be retrieved at execution time. Attributes can also define optional arguments, which are also known as named arguments. These arguments are supplied to attribute constructors by name and should have a corresponding read/write property.</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageDefault" xml:space="preserve">
-    <value>Add a public read-only property accessor for positional argument {0} of Attribute {1}</value>
+    <value>Add a public read-only property accessor for positional argument '{0}' of Attribute '{1}'</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageRemoveSetter" xml:space="preserve">
-    <value>Remove the property setter from {0} or reduce its accessibility because it corresponds to positional argument {1}</value>
+    <value>Remove the property setter from '{0}' or reduce its accessibility because it corresponds to positional argument '{1}'</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageIncreaseVisibility" xml:space="preserve">
-    <value>If {0} is the property accessor for positional argument {1}, make it public</value>
+    <value>If '{0}' is the property accessor for positional argument '{1}', make it public</value>
   </data>
   <data name="UsePropertiesWhereAppropriateTitle" xml:space="preserve">
     <value>Use properties where appropriate</value>
   </data>
   <data name="UsePropertiesWhereAppropriateDescription" xml:space="preserve">
-    <value>A public or protected method has a name that starts with ""Get"", takes no parameters, and returns a value that is not an array. The method might be a good candidate to become a property.</value>
+    <value>A public or protected method has a name that starts with 'Get', takes no parameters, and returns a value that is not an array. The method might be a good candidate to become a property.</value>
   </data>
   <data name="UsePropertiesWhereAppropriateMessage" xml:space="preserve">
     <value>Use properties where appropriate</value>
   </data>
   <data name="MarkEnumsWithFlagsTitle" xml:space="preserve">
-    <value>Mark enums with FlagsAttribute</value>
+    <value>Mark enums with 'FlagsAttribute'</value>
   </data>
   <data name="MarkEnumsWithFlagsDescription" xml:space="preserve">
-    <value>An enumeration is a value type that defines a set of related named constants. Apply FlagsAttribute to an enumeration when its named constants can be meaningfully combined.</value>
+    <value>An enumeration is a value type that defines a set of related named constants. Apply 'FlagsAttribute' to an enumeration when its named constants can be meaningfully combined.</value>
   </data>
   <data name="MarkEnumsWithFlagsMessage" xml:space="preserve">
-    <value>Mark enums with FlagsAttribute</value>
+    <value>Mark enums with 'FlagsAttribute'</value>
   </data>
   <data name="InterfaceMethodsShouldBeCallableByChildTypesTitle" xml:space="preserve">
     <value>Interface methods should be callable by child types</value>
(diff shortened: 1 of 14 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/xlf/MicrosoftCodeQualityAnalyzersResources.cs.xlf
```diff
@@ -93,8 +93,8 @@
         <note />
       </trans-unit>
       <trans-unit id="AvoidOutParametersTitle">
-        <source>Avoid out parameters</source>
-        <target state="translated">Vyhněte se výstupním parametrům</target>
+        <source>Avoid 'out' parameters</source>
+        <target state="needs-review-translation">Vyhněte se výstupním parametrům</target>
```
Youssef1313, line 97: My change here will most likely prevent the translation of out. I think that's correct, but pointing to this in case that's not what you want.

Evangelink:
I am fine with this change but in the past, I got asked (suggested :)) to avoid these manual fixes and instead create an analyzer.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -118,58 +118,58 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="AvoidAsyncVoidTitle" xml:space="preserve">
-    <value>Avoid Async Void</value>
+    <value>Avoid 'async void'.</value>
```
Evangelink, line 121: There shouldn't be a period for title.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -118,58 +118,58 @@
...
   </data>
   <data name="AvoidAsyncVoidDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="AvoidAsyncVoidMessage" xml:space="preserve">
-    <value>Avoid Async Void</value>
+    <value>Avoid 'async void'.</value>
```
Evangelink, line 127: No trailing period for single-sentence message.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: The analyzer should have caught that, is it broken?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -118,58 +118,58 @@
...
   <data name="DoNotStoreAsyncLambdasAsVoidReturningDelegateTypesDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="DoNotStoreAsyncLambdasAsVoidReturningDelegateTypesMessage" xml:space="preserve">
-    <value>Don't Store Async Lambdas as Void Returning Delegate Types</value>
+    <value>Don't store async lambdas as void returning delegate types</value>
   </data>
   <data name="PropagateCancellationTokensWhenPossibleTitle" xml:space="preserve">
-    <value>Propagate CancellationTokens When Possible</value>
+    <value>Propagate 'CancellationToken's when possible</value>
```
Evangelink, line 157: Do we need the plural form?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -118,58 +118,58 @@
...
   <data name="PropagateCancellationTokensWhenPossibleDescription" xml:space="preserve">
     <value>#N/A</value>
   </data>
   <data name="PropagateCancellationTokensWhenPossibleMessage" xml:space="preserve">
-    <value>Propagate CancellationTokens When Possible</value>
+    <value>Propagate 'CancellationToken's when possible</value>
   </data>
   <data name="DoNotMixBlockingAndAsyncTitle" xml:space="preserve">
-    <value>Don't Mix Blocking and Async</value>
+    <value>Don't mix blocking and async</value>
```
Evangelink, line 166: Shall it be `'async'`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -244,28 +244,28 @@
...
-    <value>Mark attributes with AttributeUsageAttribute</value>
+    <value>Mark attributes with 'AttributeUsageAttribute'</value>
   </data>
   <data name="MarkAttributesWithAttributeUsageDescription" xml:space="preserve">
-    <value>When you define a custom attribute, mark it by using AttributeUsageAttribute to indicate where in the source code the custom attribute can be applied. The meaning and intended usage of an attribute will determine its valid locations in code.</value>
+    <value>When you define a custom attribute, mark it by using 'AttributeUsageAttribute' to indicate where in the source code the custom attribute can be applied. The meaning and intended usage of an attribute will determine its valid locations in code.</value>
   </data>
   <data name="MarkAttributesWithAttributeUsageMessageDefault" xml:space="preserve">
-    <value>Specify AttributeUsage on {0}</value>
+    <value>Specify AttributeUsage on '{0}'</value>
```
Evangelink, line 265:
```suggestion
    <value>Specify 'AttributeUsage' on '{0}'</value>
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -274,31 +274,31 @@
     <value>Attributes can define mandatory arguments that must be specified when you apply the attribute to a target. These are also known as positional arguments because they are supplied to attribute constructors as positional parameters. For every mandatory argument, the attribute should also provide a corresponding read-only property so that the value of the argument can be retrieved at execution time. Attributes can also define optional arguments, which are also known as named arguments. These arguments are supplied to attribute constructors by name and should have a corresponding read/write property.</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageDefault" xml:space="preserve">
-    <value>Add a public read-only property accessor for positional argument {0} of Attribute {1}</value>
+    <value>Add a public read-only property accessor for positional argument '{0}' of Attribute '{1}'</value>
```
Evangelink, line 277: Either `attribute` (lower-case first letter) or `'Attribute'`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -274,31 +274,31 @@
...
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageRemoveSetter" xml:space="preserve">
-    <value>Remove the property setter from {0} or reduce its accessibility because it corresponds to positional argument {1}</value>
+    <value>Remove the property setter from '{0}' or reduce its accessibility because it corresponds to positional argument '{1}'</value>
   </data>
   <data name="DefineAccessorsForAttributeArgumentsMessageIncreaseVisibility" xml:space="preserve">
-    <value>If {0} is the property accessor for positional argument {1}, make it public</value>
+    <value>If '{0}' is the property accessor for positional argument '{1}', make it public</value>
```
Evangelink, line 283: What about `public`? Is the the keyword or a short-hand for public and protected?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -632,19 +632,19 @@
     <value>Change the type of property '{0}' from 'string' to 'System.Uri'</value>
   </data>
   <data name="ImplementIDisposableCorrectlyTitle" xml:space="preserve">
-    <value>Implement IDisposable Correctly</value>
+    <value>Implement 'IDisposable' Correctly</value>
```
Evangelink, line 635:
```suggestion
    <value>Implement 'IDisposable' correctly</value>
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -632,19 +632,19 @@
...
   </data>
   <data name="ImplementIDisposableCorrectlyDescription" xml:space="preserve">
-    <value>All IDisposable types should implement the Dispose pattern correctly.</value>
+    <value>All 'IDisposable' types should implement the Dispose pattern correctly.</value>
```
Evangelink, line 638:
```suggestion
    <value>All 'IDisposable' types should implement the 'Dispose' pattern correctly.</value>
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -632,19 +632,19 @@
...
+    <value>Remove 'IDisposable' from the list of interfaces implemented by '{0}' as it is already implemented by base type '{1}'</value>
   </data>
   <data name="ImplementIDisposableCorrectlyMessageFinalizeOverride" xml:space="preserve">
-    <value>Remove the finalizer from type '{0}', override Dispose(bool disposing), and put the finalization logic in the code path where 'disposing' is false. Otherwise, it might lead to duplicate Dispose invocations as the Base type '{1}' also provides a finalizer.</value>
+    <value>Remove the finalizer from type '{0}', override 'Dispose(bool disposing)', and put the finalization logic in the code path where 'disposing' is false. Otherwise, it might lead to duplicate Dispose invocations as the Base type '{1}' also provides a finalizer.</value>
   </data>
   <data name="ImplementIDisposableCorrectlyMessageDisposeOverride" xml:space="preserve">
-    <value>Remove '{0}', override Dispose(bool disposing), and put the dispose logic in the code path where 'disposing' is true</value>
+    <value>Remove '{0}', override 'Dispose(bool disposing)', and put the dispose logic in the code path where 'disposing' is true</value>
```
Evangelink, line 647: Shall it be `'true'`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -857,10 +857,10 @@
     <value>For value types, the inherited implementation of Equals uses the Reflection library and compares the contents of all fields. Reflection is computationally expensive, and comparing every field for equality might be unnecessary. If you expect users to compare or sort instances, or to use instances as hash table keys, your value type should implement Equals.</value>
   </data>
   <data name="OverrideEqualsAndOperatorEqualsOnValueTypesMessageEquals" xml:space="preserve">
-    <value>{0} should override Equals</value>
+    <value>'{0}' should override 'Equals'</value>
   </data>
   <data name="OverrideEqualsAndOperatorEqualsOnValueTypesMessageOpEquality" xml:space="preserve">
-    <value>{0} should override the equality (==) and inequality (!=) operators</value>
+    <value>'{0}' should override the equality (==) and inequality (!=) operators</value>
```
Evangelink, line 863: Might be worth quoting the operators.

Evangelink: Also this message is pretty csharp focused, is it ok?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -878,28 +878,28 @@
     <value>The strong name protects clients from unknowingly loading an assembly that has been tampered with. Assemblies without strong names should not be deployed outside very limited scenarios. If you share or distribute assemblies that are not correctly signed, the assembly can be tampered with, the common language runtime might not load the assembly, or the user might have to disable verification on his or her computer.</value>
   </data>
   <data name="AssembliesShouldHaveValidStrongNamesMessageNoStrongName" xml:space="preserve">
-    <value>Sign {0} with a strong name key.</value>
+    <value>Sign '{0}' with a strong name key.</value>
```
Evangelink, line 881: Message should not end with period.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -878,28 +878,28 @@
...
   </data>
   <data name="AssembliesShouldHaveValidStrongNamesMessageNotValid" xml:space="preserve">
-    <value>Verify that {0} has a valid strong name before deploying.</value>
+    <value>Verify that '{0}' has a valid strong name before deploying.</value>
```
Evangelink, line 884: Message should not end with period.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -908,13 +908,13 @@
     <value>Generate missing operators</value>
   </data>
   <data name="OverrideEqualsOnOverloadingOperatorEqualsCodeActionTitle" xml:space="preserve">
-    <value>Override object.Equals</value>
+    <value>Override 'object.Equals'</value>
```
Evangelink, line 911: Sometimes we use `object` and sometimes `Object`, do we want to be consistent? I'd vote for the alias `object` as it is the recommended use in code.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -968,7 +968,7 @@
     <value>Append .ConfigureAwait(true)</value>
   </data>
   <data name="ImplementIEquatableWhenOverridingObjectEqualsDescription" xml:space="preserve">
-    <value>When a type T overrides Object.Equals(object), the implementation must cast the object argument to the correct type T before performing the comparison. If the type implements IEquatable&lt;T&gt;, and therefore offers the method T.Equals(T), and if the argument is known at compile time to be of type T, then the compiler can call IEquatable&lt;T&gt;.Equals(T) instead of Object.Equals(object), and no cast is necessary, improving performance.</value>
+    <value>When a type T overrides 'Object.Equals(object)', the implementation must cast the object argument to the correct type T before performing the comparison. If the type implements 'IEquatable&lt;T&gt;', and therefore offers the method 'T.Equals(T)', and if the argument is known at compile time to be of type 'T', then the compiler can call 'IEquatable&lt;T&gt;.Equals(T)' instead of 'Object.Equals(object)', and no cast is necessary, improving performance.</value>
```
Evangelink, line 971: Shall we quote the generic type `T`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1014,34 +1014,34 @@
     <value>A member calls a potentially dangerous or problematic method.</value>
   </data>
   <data name="AvoidCallingProblematicMethodsMessageSystemGCCollect" xml:space="preserve">
-    <value>Remove the call to GC.Collect from {0}. It is usually unnecessary to force garbage collection, and doing so can severely degrade performance.</value>
+    <value>Remove the call to GC.Collect from '{0}'. It is usually unnecessary to force garbage collection, and doing so can severely degrade performance.</value>
```
Evangelink, line 1017:
```suggestion
    <value>Remove the call to 'GC.Collect' from '{0}'. It is usually unnecessary to force garbage collection, and doing so can severely degrade performance.</value>
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1014,34 +1014,34 @@
...
+    <value>Remove the call to 'System.Type.InvokeMember' with 'BindingFlags.NonPublic' from '{0}'. Taking a dependency on a private member increases the chance of a breaking change in the future.</value>
   </data>
   <data name="AvoidCallingProblematicMethodsMessageCoInitializeSecurity" xml:space="preserve">
-    <value>{0} is a P/Invoke declaration to an OLE32 API that cannot be reliably called after the runtime has been initialized. The workaround is to write an unmanaged shim that will call the routine and then activate and call into managed code. You can do this using an export from a mixed-mode C++ DLL, by registering a managed component for use by COM, or by using the runtime hosting API.</value>
+    <value>'{0}' is a P/Invoke declaration to an OLE32 API that cannot be reliably called after the runtime has been initialized. The workaround is to write an unmanaged shim that will call the routine and then activate and call into managed code. You can do this using an export from a mixed-mode C++ DLL, by registering a managed component for use by COM, or by using the runtime hosting API.</value>
```
Evangelink, line 1029: Shall we quote `OLE32`?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1014,34 +1014,34 @@
...
+    <value>'{0}' is a P/Invoke declaration to an OLE32 API that cannot be reliably called against a runtime callable wrapper (a managed object wrapping a COM object). Runtime callable wrappers dynamically fetch interface pointers so the effect of the call might be arbitrarily lost. Runtime callable wrappers for a given COM object are also shared across an application domain so the call could possibly affect other users. Replace this call with a native wrapper COM object for the interface pointer that does the appropriate CoSetProxyBlanket calls.</value>
   </data>
   <data name="AvoidCallingProblematicMethodsMessageSystemRuntimeInteropServicesSafeHandleDangerousGetHandle" xml:space="preserve">
-    <value>Remove the call to SafeHandle.DangerousGetHandle from {0}</value>
+    <value>Remove the call to 'SafeHandle.DangerousGetHandle' from '{0}'</value>
   </data>
   <data name="AvoidCallingProblematicMethodsMessageSystemReflectionAssemblyLoadFrom" xml:space="preserve">
-    <value>Remove the call to Assembly.LoadFrom from {0}</value>
+    <value>Remove the call to 'Assembly.LoadFile' from '{0}'</value>
```
Evangelink, line 1038: Message was `LoadFrom` and you changed it to `LoadFile`.

Evangelink (changes requested):
Requesting change just to wait from @mavasani or @sharwell input.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Youssef1313, line 121: A few of the manual fixes, like this one for example, wouldn't be caught by an analyzer.

Youssef1313:
@Evangelink Thanks for review!
I'll wait for @mavasani or @sharwell if they'd accept these manual changes before moving forward with more changes.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: A lot of entries in the form `Axxx Bxxx Cxxx` could be relatively easily caught. 

Regarding quoting keywords, we could find a way, although it might be less elegant.

I am more concerned about types and properties.

sharwell, line 121: 💡 This needs `{Locked="async void"}` as the message. The same would apply for other uses of keywords and literals.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
sharwell: I would prefer:

```suggestion
    <value>Propagate 'CancellationToken' when possible</value>
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
sharwell, line 166: Should we be using `Do not` here?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: Is it a roslyn/roslyn-analyzer specific syntax?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
sharwell: I don't think we should quote it.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
sharwell: 🤯

Youssef1313: @Evangelink Copy/paste error 😕 

Good catch!
I'll review all changes again, and address feedbacks from you and @sharwell.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Youssef1313: @Evangelink Good catch. The reason for that is simply this is a dead resource. It's not used currently. Probably the relevant analyzer was deleted at some point, and it was forgotten to remove these resources.

Youssef1313: There are lots of dead resources. Let me cleanup this first in a separate PR.

Youssef1313:
PR should be smaller now after deleting unused resources and fixing the conflicts. I'm going to temporary convert to a draft until I manually re-review the changes and address the requested feedbacks.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -154,13 +154,13 @@
     <value>The default value of an uninitialized enumeration, just as other value types, is zero. A nonflags-attributed enumeration should define a member by using the value of zero so that the default value is a valid value of the enumeration. If an enumeration that has the FlagsAttribute attribute applied defines a zero-valued member, its name should be ""None"" to indicate that no values have been set in the enumeration.</value>
   </data>
   <data name="EnumsShouldHaveZeroValueMessageFlagsRename" xml:space="preserve">
-    <value>In enum {0}, change the name of {1} to 'None'</value>
+    <value>In enum '{0}', change the name of '{1}' to 'None'</value>
```
sharwell, line 157: 💡 It may be good to lock `None`

```suggestion
    <value>In enum '{0}', change the name of '{1}' to 'None'</value>
    <comment>{Locked="None"}</comment>
```

This would apply to all identifiers in resources representing API names.

We can approach this in two ways:

1. Have a first pass that adds `'` around true placeholders (e.g. `'{0}'`) and a second pass that addresses identifiers (e.g. `'None'` or `'FlagsAttribute'`)
2. Have one pass that covers both cases for a single resource file

Whichever way we go, the locking for names should be done in the pass that updates API names.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -190,28 +190,28 @@
     <value>Mark assemblies with assembly version</value>
   </data>
   <data name="MarkAssembliesWithComVisibleTitle" xml:space="preserve">
-    <value>Mark assemblies with ComVisible</value>
+    <value>Mark assemblies with 'ComVisible'</value>
```
sharwell, line 193: @rainersigwald We are considering updating all analyzer titles and messages to treat API terms the same way we treat placeholders for type or member names, with single quotes around them. As a person particularly interested in English grammar, I would value your opinion on this proposal.

sharwell, line 193: 💭 We should probably be consistent in titles regarding the use of `Attribute` suffix.

rainersigwald: I think it makes sense and is consistent. :shipit:

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Youssef1313: @Evangelink This is used by the localization team to tell them don't translate this part. Probably in any project in the organization.

@sharwell Trying to apply Locked everywhere in the PR will make it VERY big, and possibly hard to review. But let me know if you still prefer to do that.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -130,37 +130,37 @@
...
   <data name="UseGenericEventHandlerInstancesForEventMessage" xml:space="preserve">
-    <value>Change the event '{0}' to replace the type '{1}' with a generic EventHandler, for example EventHandler&lt;T&gt;, where T is a valid EventArgs</value>
+    <value>Change the event '{0}' to replace the type '{1}' with a generic 'EventHandler', for example 'EventHandler&lt;T&gt;', where 'T' is a valid 'EventArgs'</value>
   </data>
   <data name="UseGenericEventHandlerInstancesForEventDescription" xml:space="preserve">
     <value>A delegate that handles a public or protected event does not have the correct signature, return type, or parameter names.</value>
   </data>
   <data name="UseGenericEventHandlerInstancesForEvent2Message" xml:space="preserve">
-    <value>Change the event '{0}' to use a generic EventHandler by defining the event type explicitly, for e.g. Event MyEvent As EventHandler(Of MyEventArgs).</value>
+    <value>Change the event '{0}' to use a generic 'EventHandler' by defining the event type explicitly, for example, 'Event MyEvent As EventHandler(Of MyEventArgs)'.</value>
```
Youssef1313, line 145: Why is this using VB-specific syntax? 😨 
The analyzer seems applicable to both languages?
https://github.com/dotnet/roslyn-analyzers/blob/main/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UseGenericEventHandlerInstances.cs#L22

mavasani: Yes, we should fix that. Additionally, we should also not be giving examples in the diagnostic message. Instead the example should go into the diagnostic description.

Youssef1313: @mavasani I'm going to open a separate issue tracking this to first confirm this is reachable in C#, and see how can we make the message read well for both languages.

FYI: the message here was ending in a period while it shouldn't. The analyzer didn't catch that initially because it thought it's a muli-sentence message (because of `e.g.` - the analyzer treated `.` as end of sentence, hence decided it's multi-sentence). I'm not sure this is fixable.

```suggestion
    <value>Change the event '{0}' to use a generic 'EventHandler' by defining the event type explicitly, for example, 'Event MyEvent As EventHandler(Of MyEventArgs)'</value>
```

Issue #4544, opened 2020-12-09
Flaky unit test DisposeMethodsShouldCallBaseClassDisposeTests.ErrorCase_NoDiagnostic


mavasani:
This test is occasionally failing in our signed/CI builds. Seems like compiler is sometimes reporting an extra CS error. We may want to consider updating/removing this test from this repo to workaround this (while also trying to identify a repro case for compiler issue).

**Failing test**: https://github.com/dotnet/roslyn-analyzers/blob/a23fde8d03fcec0c8e5beb6b1f3882048ab8435b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DisposeMethodsShouldCallBaseClassDisposeTests.cs#L1083-L1117

**Test failure log:**
```
Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessage
Exception : Context: Diagnostics of test state
Mismatch between number of diagnostics returned, expected "7" actual "8"

Diagnostics:
// /0/Test0.cs(2,29): error CS0246: El nombre del tipo o del espacio de nombres 'IComparable<>' no se encontró (¿falta una directiva using o una referencia de ensamblado?)
DiagnosticResult.CompilerError("CS0246").WithSpan(2, 29, 2, 43).WithArguments("IComparable<>"),
// /0/Test0.cs(3,16): error CS0246: El nombre del tipo o del espacio de nombres 'IComparable<>' no se encontró (¿falta una directiva using o una referencia de ensamblado?)
DiagnosticResult.CompilerError("CS0246").WithSpan(3, 16, 3, 30).WithArguments("IComparable<>"),
// /0/Test0.cs(3,16): error CS0246: El nombre del tipo o del espacio de nombres 'IComparable<>' no se encontró (¿falta una directiva using o una referencia de ensamblado?)
DiagnosticResult.CompilerError("CS0246").WithSpan(3, 16, 3, 30).WithArguments("IComparable<>"),
// /0/Test0.cs(10,22): error CS1061: "T" no contiene una definición para "CompareTo" ni un método de extensión accesible "CompareTo" que acepte un primer argumento del tipo "T" (¿falta alguna directiva using o una referencia de ensamblado?)
DiagnosticResult.CompilerError("CS1061").WithSpan(10, 22, 10, 31).WithArguments("T", "CompareTo"),
[...]

PR #4774, opened 2021-02-03
Fix line endings for public api fixers

### LINKED ISSUE #4749: PublicAPIAnalyzer CodeFix does not respect existing newline style
https://github.com/dotnet/roslyn-analyzers/blob/62f71ea5c1ad04f9c4e5579c5c2b97fab16ff967/src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs#L110

Instead of reading the line endings that already exist in the file, the public API analyzers use the default from the environment. This is a problem on Windows when autocrlf is off, or when the source is shared with a Linux partition (such as WSL or when running Windows in a VM) and the line endings are set to just `\n`. Instead, it should be looking at the existing line endings in the file and, if they are just `\n`, only using that, instead of unconditionally using the environment-default newline.

Youssef1313:
Fixes #4749

### DIFF
--- a/src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
+++ b/src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
@@ -31,7 +31,7 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
         public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
         {
             var project = context.Document.Project;
-            var publicSurfaceAreaDocument = GetUnshippedDocument(project);
+            var publicSurfaceAreaDocument = PublicApiFixHelpers.GetUnshippedDocument(project);
 
             foreach (Diagnostic diagnostic in context.Diagnostics)
             {
@@ -49,16 +49,6 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
             }
 
             return Task.CompletedTask;
-        }
-
-        internal static TextDocument? GetUnshippedDocument(Project project)
-        {
-            return project.AdditionalDocuments.FirstOrDefault(doc => doc.Name.Equals(DeclarePublicApiAnalyzer.UnshippedFileName, StringComparison.Ordinal));
-        }
-
-        internal static TextDocument? GetShippedDocument(Project project)
-        {
-            return project.AdditionalDocuments.FirstOrDefault(doc => doc.Name.Equals(DeclarePublicApiAnalyzer.ShippedFileName, StringComparison.Ordinal));
         }
 
         private static async Task<Solution> GetFixAsync(TextDocument? publicSurfaceAreaDocument, Project project, string newSymbolName, ImmutableHashSet<string> siblingSymbolNamesToRemove, CancellationToken cancellationToken)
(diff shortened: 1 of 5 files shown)


src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
```diff
@@ -136,7 +143,14 @@ private static SourceText RemoveSymbolNamesFromSourceText(SourceText sourceText,
             List<string> lines = GetLinesFromSourceText(sourceText);
             IEnumerable<string> newLines = lines.Where(line => !linesToRemove.Contains(line));
 
-            SourceText newSourceText = sourceText.Replace(new TextSpan(0, sourceText.Length), string.Join(Environment.NewLine, newLines) + GetEndOfFileText(sourceText));
+            var endOfLine = Environment.NewLine;
+            if (sourceText.Lines.Count > 1)
+            {
+                var firstLine = sourceText.Lines[0];
+                endOfLine = sourceText.ToString(new TextSpan(firstLine.End, firstLine.EndIncludingLineBreak - firstLine.End));
+            }
```
sharwell, line 151: ❔ Can we make this an extension method? It seems to be used in several places.

sharwell (changes requested):
* Concept looks good
* Can be improved by the use of an extension method
* Tests are needed

src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs
```diff
@@ -51,16 +51,6 @@ public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
...
 
-        internal static TextDocument? GetUnshippedDocument(Project project)
-        {
-            return project.AdditionalDocuments.FirstOrDefault(doc => doc.Name.Equals(DeclarePublicApiAnalyzer.UnshippedFileName, StringComparison.Ordinal));
-        }
-
-        internal static TextDocument? GetShippedDocument(Project project)
-        {
-            return project.AdditionalDocuments.FirstOrDefault(doc => doc.Name.Equals(DeclarePublicApiAnalyzer.ShippedFileName, StringComparison.Ordinal));
-        }
```
Youssef1313, line 62: These were used by different code fixes. So I moved them to the new helper class.

Youssef1313:
Closing and re-opening to re-run CI due to failing flaky test: https://github.com/dotnet/roslyn-analyzers/issues/4544

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1481,6 +1481,46 @@ public class C
             await VerifyCSharpAdditionalFileFixAsync(source, shippedText, unshippedText, fixedUnshippedText);
         }
 
+        [Theory]
+        [WorkItem(4749, "https://github.com/dotnet/roslyn-analyzers/issues/4749")]
+        [InlineData("\r\n")] // Windows line ending.
+        [InlineData("\n")] // Linux line ending.
+        [InlineData("\r")] // macOS line ending.
```
sharwell, line 1488: This was only the Mac line ending prior to OSX. There is no need to support it.

```suggestion
```

sharwell: Mildly interesting side note: the first release of OSX predated .NET Framework 1.0 by almost a year, so C# code should really never need to handle these.

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -2167,7 +2206,7 @@ public class {|RS0016:{|RS0016:C2|}|} { }
             await VerifyCSharpAdditionalFileFixAsync(source, shippedText, unshippedText, fixedUnshippedText);
         }
 
-        [WindowsOnlyTheory]
+        [Theory]
```
Youssef1313, line 2209: I expect this test to be fixed on Linux now. Waiting for CI Linux build to confirm.

Evangelink:
@sharwell Could you review again please?

sharwell (approved):

PR #4787, opened 2021-02-05
Implement 'Apply trait to class' refactoring

sharwell:
/cc @tmat @CyrusNajmabadi

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpApplyTraitToClass.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpApplyTraitToClass.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Composition;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeRefactorings;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    [ExportCodeRefactoringProvider(LanguageNames.CSharp, Name = nameof(CSharpApplyTraitToClass))]
+    [Shared]
+    public sealed class CSharpApplyTraitToClass : AbstractApplyTraitToClass<AttributeSyntax>
+    {
+        private protected override IRefactoringHelpers RefactoringHelpers => CSharpRefactoringHelpers.Instance;
+
+        protected override SyntaxNode? GetTypeDeclarationForNode(SyntaxNode reportedNode)
+            => reportedNode.FirstAncestorOrSelf<TypeDeclarationSyntax>();
+    }
+}
--- a/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
+++ b/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
@@ -390,4 +390,7 @@
   <data name="DoNotCopyValueNoReturnValueFromReferenceMessage" xml:space="preserve">
     <value>Cannot return a value from a reference to non-copyable type '{0}'</value>
   </data>
+  <data name="ApplyTraitToContainingType" xml:space="preserve">
+    <value>Apply trait to containing type</value>
+  </data>
 </root>
--- a/src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
@@ -1,0 +1,175 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Test.Utilities;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeRefactoringVerifier<
+    Roslyn.Diagnostics.CSharp.Analyzers.CSharpApplyTraitToClass>;
+using VerifyVB = Test.Utilities.VisualBasicCodeRefactoringVerifier<
+    Roslyn.Diagnostics.VisualBasic.Analyzers.VisualBasicApplyTraitToClass>;
+
+namespace Roslyn.Diagnostics.Analyzers.UnitTests
+{
+    public class ApplyTraitToClassTests
+    {
+        [Theory]
+        [InlineData("A", "")]
+        [InlineData("", "A")]
+        [InlineData("A", "A")]
+        public async Task MoveTraitToType_MovesSecond_CSharp(string name, string value)
+        {
+            var source = $@"
+using Xunit;
+
+class C
+{{
+    [$$Trait(""{name}"", ""{value}"")]
+    public void Method() {{ }}
+
+    [Fact, Trait(""{name}"", ""{value}"")]
+    public void Method2() {{ }}
+}}
+";
+            var fixedSource = $@"
+using Xunit;
+
+    [Trait(""{name}"", ""{value}"")]
+class C
+{{
+    public void Method() {{ }}
+
+    [Fact]
+    public void Method2() {{ }}
+}}
+";
+
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithXUnit,
+                TestCode = source,
+                FixedCode = fixedSource,
+            }.RunAsync();
+        }
+
+        [Theory]
+        [InlineData("A", "")]
+        [InlineData("", "A")]
+        [InlineData("A", "A")]
+        public async Task MoveTraitToType_MovesSecond_VisualBasic(string name, string value)
+        {
+            var source = $@"
+Imports Xunit
+
+Class C
+    <$$Trait(""{name}"", ""{value}"")>
+    Public Sub Method()
+    End Sub
+
+    <Fact, Trait(""{name}"", ""{value}"")>
+    Public Sub Method2()
+    End Sub
+End Class
+";
+            var fixedSource = $@"
+Imports Xunit
+
+    <Trait(""{name}"", ""{value}"")>
+Class C
+    Public Sub Method()
+    End Sub
+
+    <Fact>
+    Public Sub Method2()
+    End Sub
+End Class
+";
+
+            await new VerifyVB.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithXUnit,
+                TestCode = source,
+                FixedCode = fixedSource,
+            }.RunAsync();
+        }
+
+        [Theory]
+        [InlineData("A", "")]
+        [InlineData("", "A")]
+        [InlineData("A", "A")]
+        public async Task MoveTraitToType_MovesOnlyFirst_CSharp(string name, string value)
+        {
+            var source = $@"
+using Xunit;
+
+class C
+{{
+    [$$Trait("""", """")]
+    public void Method() {{ }}
+
+    [Fact, Trait(""{name}"", ""{value}"")]
+    public void Method2() {{ }}
+}}
+";
+            var fixedSource = $@"
+using Xunit;
+
+    [Trait("""", """")]
+class C
+{{
+    public void Method() {{ }}
+
+    [Fact, Trait(""{name}"", ""{value}"")]
+    public void Method2() {{ }}
+}}
+";
+
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithXUnit,
+                TestCode = source,
+                FixedCode = fixedSource,
+            }.RunAsync();
+        }
+
+        [Theory]
+        [InlineData("A", "")]
+        [InlineData("", "A")]
+        [InlineData("A", "A")]
+        public async Task MoveTraitToType_MovesOnlyFirst_VisualBasic(string name, string value)
+        {
+            var source = $@"
+Imports Xunit
+
+Class C
+    <$$Trait("""", """")>
+    Public Sub Method()
+    End Sub
+
+    <Fact, Trait(""{name}"", ""{value}"")>
+    Public Sub Method2()
+    End Sub
+End Class
+";
+            var fixedSource = $@"
+Imports Xunit
+
+    <Trait("""", """")>
+Class C
+    Public Sub Method()
+    End Sub
+
+    <Fact, Trait(""{name}"", ""{value}"")>
+    Public Sub Method2()
+    End Sub
+End Class
+";
+
+            await new VerifyVB.Test
+            {
+                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithXUnit,
+                TestCode = source,
+                FixedCode = fixedSource,
+            }.RunAsync();
+        }
+    }
+}
--- a/src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
@@ -33,7 +33,7 @@ class C
             var fixedSource = $@"
 using Xunit;
 
-    [Trait(""{name}"", ""{value}"")]
+[Trait(""{name}"", ""{value}"")]
 class C
 {{
     public void Method() {{ }}
@@ -73,7 +73,7 @@ End Class
             var fixedSource = $@"
 Imports Xunit
 
-    <Trait(""{name}"", ""{value}"")>
+<Trait(""{name}"", ""{value}"")>
 Class C
     Public Sub Method()
     End Sub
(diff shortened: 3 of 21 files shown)


src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            var syntaxTree = await document.GetSyntaxTreeAsync(cancellationToken).ConfigureAwait(false);
+            var syntaxRoot = await syntaxTree!.GetRootAsync(cancellationToken).ConfigureAwait(false);
+            var reportedNode = syntaxRoot.FindNode(sourceSpan, getInnermostNodeForTie: true);
+            var typeDeclaration = GetTypeDeclarationForNode(reportedNode);
+            if (typeDeclaration is null)
+                return document;
+
+            var syntaxGenerator = SyntaxGenerator.GetGenerator(document);
+            if (syntaxGenerator.TryGetContainingDeclaration(reportedNode, DeclarationKind.Attribute) is not { } attribute)
```
mavasani, line 67: This code seems to differ from Register callback where we use TryGetRelevantNodeAsync. Is it possible to refactor and share code which tries to go from context to containing method?

mavasani (approved):

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+                syntaxGenerator.GetMembers(typeDeclaration),
+                (originalNode, replacementNode) =>
+                {
+                    foreach (var attribute in syntaxGenerator.GetAttributes(originalNode))
+                    {
+                        var attributeType = semanticModel.GetTypeInfo(attribute, cancellationToken);
+                        if (attributeType.Type is null)
+                            attributeType = semanticModel.GetTypeInfo(attribute.ChildNodes().First(), cancellationToken);
+
+                        if (attributeType.Type is not { Name: "TraitAttribute", ContainingNamespace: { Name: "Xunit", ContainingNamespace: { IsGlobalNamespace: true } } })
```
CyrusNajmabadi, line 91: consider extracting helper for this check.

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
```diff
@@ -390,4 +390,7 @@
   <data name="DoNotCopyValueNoReturnValueFromReferenceMessage" xml:space="preserve">
     <value>Cannot return a value from a reference to non-copyable type '{0}'</value>
   </data>
+  <data name="ApplyTraitToContainingType" xml:space="preserve">
+    <value>Apply trait to containing type</value>
```
CyrusNajmabadi, line 394: Consider `'Move trait etc.'` to make it clearer that it's going to move these and remove from all methods?

CyrusNajmabadi (approved):

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
sharwell: I used 'Apply' for one main reason: this refactoring _does not_ verify that all test methods use the trait consistently before offering to apply it. In other words, it is possible for this refactoring to add traits to tests which did not previously have the trait.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: @mavasani do we have a helper for this already in the repository?

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: `TryGetRelevantNodeAsync` resolves a position or span to a syntax node with a specific type. When the refactoring implementation is called, we are no longer operating on the caret position but instead are operating on the span already resolved by `TryGetRelevantNodeAsync`.

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpApplyTraitToClass.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeRefactorings;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Roslyn.Diagnostics.Analyzers;
+
+namespace Roslyn.Diagnostics.CSharp.Analyzers
+{
+    [ExportCodeRefactoringProvider(LanguageNames.CSharp, Name = nameof(CSharpApplyTraitToClass))]
+    [Shared]
+    public sealed class CSharpApplyTraitToClass : AbstractApplyTraitToClass<AttributeSyntax>
```
Youssef1313, line 14: nit: Consider renaming to end in "Refactoring"

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.CodeAnalysis.Text;
+
+namespace Roslyn.Diagnostics.Analyzers
+{
+    public abstract class AbstractApplyTraitToClass<TAttributeSyntax> : CodeRefactoringProvider
+        where TAttributeSyntax : SyntaxNode
+    {
+        protected AbstractApplyTraitToClass()
+        {
+        }
```
Youssef1313, line 20: Can be deleted I think. The ctor of abstract classes is protected by default.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+            context.RegisterRefactoring(
+                CodeAction.Create(
+                    RoslynDiagnosticsAnalyzersResources.ApplyTraitToContainingType,
+                    cancellationToken => ApplyTraitToClassAsync(context.Document, location.SourceSpan, cancellationToken),
+                    nameof(AbstractApplyTraitToClass<TAttributeSyntax>)));
+        }
+
+        private async Task<Document> ApplyTraitToClassAsync(Document document, TextSpan sourceSpan, CancellationToken cancellationToken)
+        {
+            var semanticModel = (await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false))!;
```
Youssef1313, line 57: Semantic model was already retrieved in the caller-side. Consider passing it instead.

src/Roslyn.Diagnostics.Analyzers/VisualBasic/VisualBasicApplyTraitToClass.vb
```diff
@@ -0,0 +1,29 @@
...
+Imports Roslyn.Diagnostics.Analyzers
+
+Namespace Roslyn.Diagnostics.VisualBasic.Analyzers
+    <ExportCodeRefactoringProvider(LanguageNames.VisualBasic, Name:=NameOf(VisualBasicApplyTraitToClass))>
+    <[Shared]>
+    Public NotInheritable Class VisualBasicApplyTraitToClass
+        Inherits AbstractApplyTraitToClass(Of AttributeSyntax)
+
+        Public Sub New()
+        End Sub
```
Youssef1313, line 17: Is this needed?

src/Roslyn.Diagnostics.Analyzers/VisualBasic/VisualBasicApplyTraitToClass.vb
```diff
@@ -0,0 +1,29 @@
...
+
+        Private Protected Overrides ReadOnly Property RefactoringHelpers As IRefactoringHelpers
+            Get
+                Return VisualBasicRefactoringHelpers.Instance
+            End Get
+        End Property
```
Youssef1313, line 23:
```suggestion
        Private Protected Overrides ReadOnly Property RefactoringHelpers As IRefactoringHelpers = VisualBasicRefactoringHelpers.Instance
```

src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+    [Fact, Trait(""{name}"", ""{value}"")]
+    public void Method2() {{ }}
+}}
+";
+            var fixedSource = $@"
+using Xunit;
+
+    [Trait(""{name}"", ""{value}"")]
+class C
```
Youssef1313, line 37: Can the refactoring put the attribute with correct indentation?

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            var attribute = await context.TryGetRelevantNodeAsync<TAttributeSyntax>(RefactoringHelpers).ConfigureAwait(false);
+            if (attribute is null)
+            {
+                // No attribute in context
+                return;
+            }
+
+            var syntaxGenerator = SyntaxGenerator.GetGenerator(context.Document);
+            if (syntaxGenerator.TryGetContainingDeclaration(attribute, DeclarationKind.Method) is null)
```
Youssef1313, line 36: A little bit lost why a "SyntaxGenerator" is needed to "know info" about a node?

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+        private async Task<Document> ApplyTraitToClassAsync(Document document, TextSpan sourceSpan, CancellationToken cancellationToken)
+        {
+            var semanticModel = (await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false))!;
+
+            var syntaxTree = await document.GetSyntaxTreeAsync(cancellationToken).ConfigureAwait(false);
+            var syntaxRoot = await syntaxTree!.GetRootAsync(cancellationToken).ConfigureAwait(false);
+            var reportedNode = syntaxRoot.FindNode(sourceSpan, getInnermostNodeForTie: true);
+            var typeDeclaration = GetTypeDeclarationForNode(reportedNode);
+            if (typeDeclaration is null)
+                return document;
```
Youssef1313, line 64: Is this possible to happen? If happened, would it result in a refactoring action being offered that basically does "nothing"?

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            if (syntaxGenerator.TryGetContainingDeclaration(attribute, DeclarationKind.Method) is not { } method)
+            {
+                // The attribute is not applied to a method
+                return document;
+            }
+
+            var expectedAttributeData = semanticModel.GetDeclaredSymbol(method, cancellationToken).GetAttributes()
+                .Single(attributeData => attributeData.ApplicationSyntaxReference is not null && attribute.Span.Contains(attributeData.ApplicationSyntaxReference.Span));
```
Youssef1313, line 79: Can typing fast lead to span being changed, hence `Single(...)` can fail? I might be overthinking though :smile:

src/Roslyn.Diagnostics.Analyzers/VisualBasic/VisualBasicApplyTraitToClass.vb
sharwell: I am not a fan of this form in Visual Basic. The code I wrote is the shortest code that the language supports for the behavior I intended.

src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
sharwell: I'm not sure why it does this. 😆

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: I'm not sure. It's a convenient public API for language-agnostic data though.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: This case is not expected.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: Everything should be consistent when we get here. The `Single` check is not expected to fail, but would provide a stack trace and context in the unexpected case we failed to account for some form of attribute.

src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
Youssef1313: If SyntaxGenerator is supposed to add the attribute in proper formatting. Then it's most likely a SyntaxGenerator bug. Otherwise, I think we can annotate the node to be formatted.

sharwell: I'll add a formatting annotation and see if it fixes it.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: @mavasani any problem with avoiding the capture and just calling `GetSemanticModelAsync` again?

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: ⏱ Fixing this by using `GetOrCreateTypeByMetadataName`

src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx
sharwell: ➡️ Not planning to change this for now, but see previous comment for details

src/Roslyn.Diagnostics.Analyzers/CSharp/CSharpApplyTraitToClass.cs
sharwell: ➡️ Not planning to change this for now. For better or worse, the current name matches the majority pattern seen in this project.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: ⏱ Will remove

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: ⏱ Will capture state

src/Roslyn.Diagnostics.Analyzers/VisualBasic/VisualBasicApplyTraitToClass.vb
sharwell: ⏱ Will remove

src/Roslyn.Diagnostics.Analyzers/VisualBasic/VisualBasicApplyTraitToClass.vb
sharwell: ➡️ Leaving this one

src/Roslyn.Diagnostics.Analyzers/UnitTests/ApplyTraitToClassTests.cs
sharwell: ⏱ Formatter annotation fixed it

src/Roslyn.Diagnostics.Analyzers/Core/AbstractApplyTraitToClass.cs
sharwell: ➡️ Leaving this one for now since it's working well

src/Roslyn.Diagnostics.Analyzers/VisualBasic/VisualBasicApplyTraitToClass.vb
CyrusNajmabadi: I'm with Youssef on this.  The latter form is much nicer and clearer.  Making the singleton type one field larger is well worth it for the savings in verbosity here.  The former form just means so much more boilerplate scaffolding one has to deal with when reviewing the code.  It doesn't earn its weight so to speak.

sharwell: I'm happy to change this to the shorter form if/when https://github.com/dotnet/vblang/issues/61 is implemented (or the code in this project is ported to C#).

PR #4788, opened 2021-02-05
Rename the test accessor instance variable

sharwell:


src/Roslyn.Diagnostics.Analyzers/UnitTests/CreateTestAccessorTests.cs
```diff
@@ -32,11 +32,11 @@ internal TestAccessor GetTestAccessor()
 
     internal readonly struct TestAccessor
     {
-        private readonly TestClass _testClass;
```
sharwell, line 35: 📝 This form was getting extremely long in practice.

mavasani (approved):

PR #4789, opened 2021-02-06
update versions

jmarolf:


mavasani (dismissed):
Not sure if master should target 6.x. We want to release NuGet packages from master onto nuget.org prior to .NET 6.0 release

mavasani (approved):

PR #4790, opened 2021-02-06
Indicate that info-level messages can remain

jeffhandley:
Per [this discussion](https://github.com/dotnet/runtime/pull/47868#issuecomment-773635368), it could be helpful to call out that we don't expect all info-level messages be resolved as part of the Definition of Done for libraries analyzers.


jeffhandley:
@newellclark If you could review this as well, I'd appreciate it.  Does this make the definition of done more clear regarding info-level messages?

NewellClark (approved):

mavasani (approved):

buyaa-n (approved):

PR #4794, opened 2021-02-06
Delete unused resources

Youssef1313:
@Evangelink @sharwell This will probably make the other PR shorter and easier to review.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
@@ -117,60 +117,6 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="AvoidAsyncVoidTitle" xml:space="preserve">
-    <value>Avoid Async Void</value>
-  </data>
-  <data name="AvoidAsyncVoidDescription" xml:space="preserve">
-    <value>#N/A</value>
-  </data>
-  <data name="AvoidAsyncVoidMessage" xml:space="preserve">
-    <value>Avoid Async Void</value>
-  </data>
-  <data name="AsyncMethodNamesShouldEndInAsyncTitle" xml:space="preserve">
-    <value>Async Method Names Should End in Async</value>
-  </data>
-  <data name="AsyncMethodNamesShouldEndInAsyncDescription" xml:space="preserve">
-    <value>#N/A</value>
-  </data>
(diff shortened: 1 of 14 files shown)


drewnoakes:
Did you find these manually? I wonder if there's an opportunity to detect unused resources with an analyzer.

I suspect unused resources are quite common. I made a similar pass in the project system a while ago and reduced our binary size by a noticeable amount.

Evangelink:
It seems rather easy to add this analyzer. We have already implemented something to read resources so we could simply look if the resource is used.

@mavasani shall we still merge this as-is or close the PR and wait for the analyzer?

Youssef1313:
> Did you find these manually? I wonder if there's an opportunity to detect unused resources with an analyzer.
> 
> I suspect unused resources are quite common. I made a similar pass in the project system a while ago and reduced our binary size by a noticeable amount.

Found these by looking at the generated file and finding resources where CodeLens says: "0 references".

mavasani:
> It seems rather easy to add this analyzer. We have already implemented something to read resources so we could simply look if the resource is used.

Agreed. It would be possible to write an analyzer for detecting unused resources, but we cannot add a code fix here as it needs full project analysis and also the lightbulb needs to be in resx file, we do not support code fixes in non-source files yet. +1 to add this analyzer though.

mavasani:
> @mavasani shall we still merge this as-is or close the PR and wait for the analyzer?

@Evangelink Seems fine to go either route. If you or @Youssef1313 are keen on implementing the analyzer, then keeping these unused resources in our repo to test out the new analyzer also seems fine. I'll let you both make a call on this.

Youssef1313:
> > @mavasani shall we still merge this as-is or close the PR and wait for the analyzer?
> 
> @Evangelink Seems fine to go either route. If you or @Youssef1313 are keen on implementing the analyzer, then keeping these unused resources in our repo to test out the new analyzer also seems fine. I'll let you both make a call on this.

I was only deleting unused resources when I find that "Title" and "Message" and "Description" are all unused.

But there are still unused resources, for example `MarkAssembliesWithComVisibleMessageNoAttribute` that belongs to <https://github.com/dotnet/roslyn-analyzers/blob/f167914d442d0ae0055b28a1c92e5b9d446fdbdb/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/MarkAssembliesWithComVisible.cs>, which may not be needed to be deleted, but instead it should be used there.

So there is still room to test the analyzer when written.

Also note that there are other resource files that I didn't touch which can have unused resources as well.

mavasani:
@Youssef1313 Sounds good to me. Can you please resolve merge conflicts?

mavasani (approved):

Evangelink (approved):

drewnoakes:
I filed https://github.com/dotnet/roslyn-analyzers/issues/4865 to track the idea of having an analyzer detect unused resources.

Evangelink:
Ping @mavasani

mavasani (approved):

Issue #4721, opened 2021-01-21
Performance sensitive analyzers package


Youssef1313:
I'm having some questions (and one suggestion) regarding this package:

- (suggestion) The package currently supports C# only and is based on syntax analysis. It could be improved to support VB using operation analysis, or add VB syntax analysis as well.
- Currently all rules are in `AnalyzerReleases.Unshipped`, is there any work to release it?
- It appears that this package was a direct port from <https://github.com/microsoft/RoslynClrHeapAllocationAnalyzer> even with the same diagnostic ids. So, when releasing, should it use different diagnostic ids? I'm not sure what will happen if a user has a reference to the one in the other repo and this one?

mavasani:
Tagging @genlu @sharwell 

IMO, this is still an experimental package, with no concrete plans at the moment towards a release.

Youssef1313:
@mavasani Is it worth attempting to re-write these analyzers as operation analyzers for now? or let this task when some plans regarding release are made?

mavasani:
@Youssef1313 We have https://github.com/dotnet/roslyn-analyzers/pull/3016 and https://github.com/dotnet/roslyn-analyzers/pull/2228 to already kick start that effort. Would need someone to help review/drive those PRs to start with.

genlu:
>(suggestion) The package currently supports C# only and is based on syntax analysis. It could be improved to support VB using operation analysis, or add VB syntax analysis as well.

Yes, this is tracked by an umbrella issue https://github.com/dotnet/roslyn-analyzers/issues/2120

> So, when releasing, should it use different diagnostic ids?

Tracked by https://github.com/dotnet/roslyn-analyzers/issues/2259. I'm inclining to have new IDs (but not with prefix "CA", probably "PS"). @sharwell thoughts?

>  is there any work to release it?

One thing that blocks the publication of this package is editorconfig support. When I ported it over from `RoslynClrHeapAllocationAnalyzer`, I made the analysis conditioned on the presence of an attribute, since which is our main use case (in Roslyn repo). To make it behave the same as the original (therefore useful for most users), we need an option to say "analyzing annotated symbols only or everything". Once we have this in place, I think we will be able release it.

Youssef1313:
> we need an option to say "analyzing annotated symbols only or everything". Once we have this in place, I think we will be able release it.

I'll try to look into that.

genlu:
@Youssef1313 That'd be awesome! Thanks!

[...]

PR #4796, opened 2021-02-06
WIP on public API analyzer symbol exclusion

roji:
This is a draft for adding `excluded_symbol_names` support to PublicApiAnalyzers, to see that I'm on the right track.

One thing I ran up against... Ideally the user defines `dotnet_public_api_analyzer.excluded_symbol_names` in .editorconfig, and it applies to all the analyzers. However, the current infrastructure for global/specific options - which `excluded_symbol_names` depends on - supports either a totally global option (for all analyzers, not just PublicApiAnalyzers), or for specific rules; there's no middle ground for defining stuff on a prefix like `dotnet_public_api_analyzer`.

I can work on adding that 3rd layer of options if it seems to make sense - let me know.

/cc @mavasani @drewnoakes

### DIFF
--- a/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
+++ b/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
@@ -121,7 +121,7 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
             internal void OnSymbolAction(SymbolAnalysisContext symbolContext)
             {
                 var obsoleteAttribute = symbolContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemObsoleteAttribute);
-                OnSymbolActionCore(symbolContext.Symbol, symbolContext.ReportDiagnostic, obsoleteAttribute);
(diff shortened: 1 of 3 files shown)


roji:
> However, the current infrastructure for global/specific options - which excluded_symbol_names depends on - supports either a totally global option (for all analyzers, not just PublicApiAnalyzers), or for specific rules; there's no middle ground for defining stuff on a prefix like dotnet_public_api_analyzer.

One more thought - it actually seems to make more sense to replace the global layer with a "prefix" layer; since after this PR we plan to have `excluded_symbol_names` under both `dotnet_code_quality` and `dotnet_public_api_analyzer`, treating it (or anything else) as truly global doesn't seem to make sense. So an option lookup would be either by rule ID or by prefix.

Youssef1313:
> One thing I ran up against... Ideally the user defines dotnet_public_api_analyzer.excluded_symbol_names in .editorconfig, and it applies to all the rules (RS0016, RS0022, RS0026...). However, the current infrastructure for global/specific options - which excluded_symbol_names depends on - supports either a totally global option (for all analyzers, not just PublicApiAnalyzers), or for specific rules; there's no middle ground for defining stuff on a prefix like dotnet_public_api_analyzer.

I wanted this while working on https://github.com/dotnet/roslyn-analyzers/issues/4721#issuecomment-767902262

I needed to add an option that applies to all rules in a specific package (the perf sensitive package). Something like `dotnet_performance_sensitive.require_attribute = true # or false.`

PR #4797, opened 2021-02-07
[New Analyzer/Fixer] Prefer static HashData method over ComputeHash

wzchua:
Resolves dotnet/runtime#40579


wzchua:
I'm confused about the target branch. targeting master first for the CI

Youssef1313:
This should target a release/6.0.1xx branch. But not sure if it should target preview1 or preview2 branch. @mavasani to confirm the correct branch.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,81 @@
...
+                        context.RegisterCodeFix(GetCodeAction(context, hashTypeName, bufferArgNode, codeFixer), diagnostics);
+                        return;
+                    }
+            }
+        }
+
+        private static CodeAction GetCodeAction(CodeFixContext context, string hashTypeName, SyntaxNode bufferArgNode, ApplyCodeFixAction codeFixer)
+        {
+            return CodeAction.Create(
+                   title: MicrosoftNetCoreAnalyzersResources.PreferHashDataOverComputeHashAnalyzerTitle,
```
Youssef1313, line 60: @mavasani @Evangelink Should the title for the codefix start with "Use" instead (i.e. different from analyzer title)?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,81 @@
...
+                       SyntaxGenerator generator = editor.Generator;
+
+                       // hashTypeName.HashData
+                       var hashData = generator.MemberAccessExpression(generator.IdentifierName(hashTypeName), PreferHashDataOverComputeHashAnalyzer.HashDataMethodName);
+                       var hashDataInvoked = generator.InvocationExpression(hashData, bufferArgNode);
+                       codeFixer(editor, hashDataInvoked);
+
+                       return editor.GetChangedDocument();
+                   },
+                   equivalenceKey: MicrosoftNetCoreAnalyzersResources.PreferHashDataOverComputeHashAnalyzerTitle);
```
Youssef1313, line 73: IIRC, @sharwell said somewhere that equivalenceKey shouldn't be localized. So you should probably use `nameof` here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,298 @@
...
+            CA1842,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Performance,
+            RuleLevel.IdeSuggestion,
+            description: s_localizableStringDescription,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(StringRule);
```
Youssef1313, line 39: nit:

```suggestion
        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(StringRule);

```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,298 @@
...
+        {
+            var compilation = context.Compilation;
+            var hashAlgoBaseType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographyHashAlgorithm);
+            var sha256Type = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographySHA256);
+            var byteType = context.Compilation.GetSpecialType(SpecialType.System_Byte);
+
+            if (hashAlgoBaseType is null || sha256Type is null || byteType is null)
+            {
+                return;
+            }
```
Youssef1313, line 57: nit:

```suggestion
            }

```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,298 @@
...
+            var byteArrayParameter = ParameterInfo.GetParameterInfo(byteType, true, 1);
+
+            // method introdued in Net5
+            var hashDataMethodType = sha256Type.GetMembers(HashDataMethodName).OfType<IMethodSymbol>().GetFirstOrDefaultMemberWithParameterInfos(byteArrayParameter);
+            if (hashDataMethodType is null)
+            {
+                return;
+            }
```
Youssef1313, line 65: nit:

```suggestion
            }

```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,298 @@
...
+                    computeHashVariableMap.Free(context.CancellationToken);
+                    createdSymbolMap.Free(context.CancellationToken);
+                    localReferenceMap.Free(context.CancellationToken);
+                }
+
+                bool IsHashCreateMethod(IMethodSymbol methodSymbol)
+                {
+                    return methodSymbol.ContainingType.Inherits(hashAlgoBaseType) &&
+                        methodSymbol.ReturnType.Inherits(hashAlgoBaseType) &&
+                        methodSymbol.Name.Equals(CreateMethodName, StringComparison.Ordinal);
```
Youssef1313, line 198: Should the comparison be case-insensitive for VB?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,298 @@
...
+        private static Diagnostic CreateDiagnostics(IInvocationOperation computeHashMethod, INamedTypeSymbol staticHashMethodType, ImmutableArray<Location> fixerLocations)
+        {
+            var dictBuilder = ImmutableDictionary.CreateBuilder<string, string?>();
+            dictBuilder.Add(TargetHashTypeName, staticHashMethodType.Name);
+
+            return computeHashMethod.CreateDiagnostic(StringRule,
+                fixerLocations,
+                dictBuilder.ToImmutable(),
+                staticHashMethodType.ToDisplayString());
+        }
```
Youssef1313, line 258: nit:

```suggestion
        }

```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,298 @@
...
+            {
+                nodeToRemove = lineDeclaration.Syntax.GetLocation();
+            }
+            else
+            {
+                nodeToRemove = declaratorOperation.Syntax.GetLocation();
+            }
+
+            return ImmutableArray.Create(bufferArgLocation, nodeToRemove);
+        }
```
Youssef1313, line 275: nit:

```suggestion
        }

```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicPreferHashDataOverComputeHash.Fixer.vb
```diff
@@ -0,0 +1,31 @@
...
+Imports Microsoft.CodeAnalysis
+Imports Microsoft.CodeAnalysis.CodeFixes
+Imports Microsoft.CodeAnalysis.Editing
+Imports Microsoft.CodeAnalysis.Formatting
+Imports Microsoft.CodeAnalysis.VisualBasic.Syntax
+Imports Microsoft.NetCore.Analyzers.Performance
+
+Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
+    <ExportCodeFixProvider(LanguageNames.VisualBasic)>
+    Public Class BasicPreferHashDataOverComputeHashFixer : Inherits PreferHashDataOverComputeHashFixer
```
Youssef1313, line 13:
```suggestion
    Public NotInheritable Class BasicPreferHashDataOverComputeHashFixer : Inherits PreferHashDataOverComputeHashFixer
```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicPreferHashDataOverComputeHash.Fixer.vb
```diff
@@ -0,0 +1,31 @@
...
+        Protected Overrides Function TryGetCodeFixer(computeHashNode As SyntaxNode, bufferArgNode As SyntaxNode, nodeToRemove As SyntaxNode, <NotNullWhen(True)> ByRef codeFixer As ApplyCodeFixAction) As Boolean
```
Youssef1313, line 14: Not sure if the NotNullWhen attribute is necessary in VB?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,636 @@
...
+                TestCode = source,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                LanguageVersion = VisualBasic.LanguageVersion.Latest,
+                FixedCode = corrected,
+            };
+
+            test.ExpectedDiagnostics.AddRange(expected);
+            return test.RunAsync();
+        }
+        private static Task TestCS(string source, params DiagnosticResult[] expected)
```
Youssef1313, line 623: I see the usage of this method isn't passing any "expected" diagnostics. Can you remove this params?

Note that if there are any expected diagnostics, you should always pass a fixed code. If the analyzer doesn't offer a codefix for the specific situation, the fixed code should be the same as the original code. This is to ensure that the fixer doesn't provide an action that does nothing.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicPreferHashDataOverComputeHash.Fixer.vb
```diff
@@ -0,0 +1,31 @@
...
+            Dim usingStatement = TryCast(nodeToRemove, UsingStatementSyntax)
+            Dim usingBlock = TryCast(usingStatement.Parent, UsingBlockSyntax)
```
Youssef1313, line 16: Can the first cast fail? This will cause a NRE on the second statement.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
wzchua: I ran a test with `create()`. The `methodSymbol.Name` is still `Create`

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicPreferHashDataOverComputeHash.Fixer.vb
wzchua: Added check

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
wzchua: I deleted the overload

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicPreferHashDataOverComputeHash.Fixer.vb
wzchua: I'm not sure either. VS generated this

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,624 @@
...
+using System;
+using System.Security.Cryptography;
+
+public class Test
+{
+    public static void TestMethod()
+    {
+        var buffer = new byte[1024];
+        {|#2:var sha256 = SHA256.Create();|}
+        byte[] digest = {|#0:sha256.ComputeHash({|#1:buffer|})|};
```
Youssef1313, line 72: I'm a little bit confused why there are 3 diagnostics here?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,304 @@
...
+            var hashAlgoBaseType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographyHashAlgorithm);
+            var sha256Type = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographySHA256);
+            var byteType = context.Compilation.GetSpecialType(SpecialType.System_Byte);
+
+            if (hashAlgoBaseType is null || sha256Type is null || byteType is null)
+            {
+                return;
+            }
+
+            var byteArrayParameter = ParameterInfo.GetParameterInfo(byteType, true, 1);
```
Youssef1313, line 60: Can you use named arguments for `true` and `1` so it's more clear what these are by looking to the code?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
wzchua: I'm capturing AdditionalLocations for the fixer. Those show up in the diagnostics

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
wzchua: Ok

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="PreferHashDataOverComputeHashAnalyzerTitle" xml:space="preserve">
+    <value>Prefer static 'HashData' method over 'ComputeHash'</value>
+  </data>
+  <data name="PreferHashDataOverComputeHashAnalyzerDescription" xml:space="preserve">
+    <value>It is simpler to use the static 'HashData' method over 'ComputeHash'.</value>
```
carlossanlop, line 1526: I think this description should also give a short explanation of _why_ it is simpler.

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -1308,6 +1308,18 @@ Marshalling of 'StringBuilder' always creates a native buffer copy, resulting in
 |CodeFix|False|
 ---
 
+## [CA1842](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1842): Prefer static 'HashData' method over 'ComputeHash'
+
+It is simpler to use the static 'HashData' method over 'ComputeHash'.
```
carlossanlop, line 1313: Same comment about expanding the description to explain _why_.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,304 @@
...
+        }
+
+        private void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+            var hashAlgoBaseType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographyHashAlgorithm);
+            var sha256Type = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographySHA256);
+            var byteType = context.Compilation.GetSpecialType(SpecialType.System_Byte);
+
+            if (hashAlgoBaseType is null || sha256Type is null || byteType is null)
```
carlossanlop, line 55: It's preferred to return as early as possible when something goes wrong. So I suggest you check if an object is null right after you call the method to obtain it, instead of obtaining all 3 first, then checking if any of the 3 was null.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,304 @@
...
+            {
+                return;
+            }
+
+            var byteArrayParameter = ParameterInfo.GetParameterInfo(byteType, isArray: true, arrayRank: 1);
+
+            // method introdued in Net5
```
carlossanlop, line 62:
```suggestion
            // method introduced in .NET 5.0
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,624 @@
...
+                vbInput,
+                vbFix,
+                VerifyCS.Diagnostic(PreferHashDataOverComputeHashAnalyzer.StringRule)
+                .WithArguments("System.Security.Cryptography.SHA256")
+                .WithLocation(0)
+                .WithLocation(1)
+                .WithLocation(2));
+
+        }
+
```
carlossanlop, line 595: I don't see any negative cases: tests where the analyzer would detect some of the items, but not all of them, and offer no fix. Can you please add those?

carlossanlop: I also don't see any test cases where there is more code inside the body of the `using`. That code needs to be preserved.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,624 @@
...
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                LanguageVersion = LanguageVersion.Preview,
+                FixedCode = corrected,
+            };
+
+            test.ExpectedDiagnostics.AddRange(expected);
+            return test.RunAsync();
+        }
+
+        private static Task TestVB(string source, string corrected, params DiagnosticResult[] expected)
```
carlossanlop, line 610: There are only two VB tests. Can you please add the same tests in both C# and VB? Ideally, in two separate files.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,624 @@
...
+        {
+            var test = new VerifyVB.Test
+            {
+                TestCode = source,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                LanguageVersion = VisualBasic.LanguageVersion.Latest,
```
carlossanlop, line 616: Should we also verify that the analyzer won't be triggered in older versions of .NET, or older versions of C#?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,624 @@
...
+";
+
+            await TestCS(
+                csInput,
+                csFix,
+                VerifyCS.Diagnostic(PreferHashDataOverComputeHashAnalyzer.StringRule)
+                .WithArguments("System.Security.Cryptography.SHA256")
+                .WithLocation(0)
+                .WithLocation(1));
+        }
```
carlossanlop, line 58: You could wrap the methods that use the same argument by putting the whole call into a helper method.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
wzchua: I created more VB Tests. There's 1 case in VB that does not work. 
When you call `SHA256.Create()` to a `HashAlgorithm` variable, the subclass `SHA256` information is lost in VB.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
wzchua: I have the test also run against NetCoreApp31

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpPreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+                case VariableDeclaratorSyntax variableDeclaratorSyntax:
+                    {
+                        codeFixer = (editor, hashDataInvoked) =>
+                        {
+                            editor.ReplaceNode(computeHashNode, hashDataInvoked);
+                            editor.RemoveNode(variableDeclaratorSyntax);
+                        };
+                        return true;
+                    }
```
Youssef1313, line 39: This code looks duplicated, can't it be a single case? e.g `case LocalDeclarationStatementSyntax or VariableDeclaratorSyntax` and use `editor.RemoveNode(nodeToRemove)`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,318 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(StringRule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(OnCompilationStart);
+        }
+
+        private void OnCompilationStart(CompilationStartAnalysisContext context)
```
Youssef1313, line 48: Can be made static?

carlossanlop:
There's a conflicting file. Can you please resolve it, @wzchua ?

wzchua:
Ok, I'll look at this some time next week

wzchua:
@carlossanlop resolved

mavasani:
@wzchua Can you please retarget the PR to `release\6.0.1xx` branch?

mavasani (changes requested):
Marking as request changes so the PR is targeting the correct branch, `release/6.0.1xx`

wzchua:
Done

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1635,4 +1635,13 @@
   <data name="DoNotUseWhenAllWithSingleTaskFix" xml:space="preserve">
     <value>Replace 'WhenAll' call with argument</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="PreferHashDataOverComputeHashAnalyzerTitle" xml:space="preserve">
+    <value>Prefer static 'HashData' method over 'ComputeHash'</value>
+  </data>
+  <data name="PreferHashDataOverComputeHashAnalyzerDescription" xml:space="preserve">
+    <value>It is simpler to use the static 'HashData' method over creating and managing a HashAlgorithm instance to call 'ComputeHash'.</value>
```
mavasani, line 1642: Given this is an analyzer with "Performance" category, I think the messaging should be slightly different.
```suggestion
    <value>It is more efficient to use the static 'HashData' method over creating and managing a HashAlgorithm instance to call 'ComputeHash'.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,128 @@
...
+        protected abstract class HashDataCodeAction : CodeAction
+        {
+            protected HashDataCodeAction(Document document, string hashTypeName, SyntaxNode bufferArgNode, SyntaxNode computeHashNode)
+            {
+                Document = document;
+                HashTypeName = hashTypeName;
+                BufferArgNode = bufferArgNode;
+                ComputeHashNode = computeHashNode;
+            }
+            public override string Title => MicrosoftNetCoreAnalyzersResources.PreferHashDataOverComputeHashAnalyzerTitle;
```
mavasani, line 76: We should not use the same title as the analyzer title, which generally tends to be longer (normally contains cause and effect). Please a different and much more succinct title that describes the code _fix_, say `Use 'HashData'` or `Replace with 'HashData'`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,318 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Performance
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class PreferHashDataOverComputeHashAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string CA1847 = nameof(CA1847);
+        internal const string TargetHashTypeName = nameof(TargetHashTypeName);
```
mavasani, line 20: Probably rename this to say something like `TargetHashTypeDiagnosticPropertyKey` to make it more explicit that is a a key for an entry added to diagnostic's property bag.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,318 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(OnCompilationStart);
+        }
+
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+
+            var hashAlgoBaseType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographyHashAlgorithm);
+            if (hashAlgoBaseType is null)
```
mavasani, line 53:
```suggestion
            if (!compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographyHashAlgorithm, out var hashAlgoBaseType))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,318 @@
...
+            {
+                return;
+            }
+
+            var sha256Type = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographySHA256);
```
mavasani, line 58: Use the `TryGetOrCreate` API for all well-know types and you can also probably combine all of them into a single if statement with `||` clauses.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,318 @@
...
+                return;
+
+                void CaptureHashLocalReferenceOperation(OperationAnalysisContext context)
+                {
+                    var localReferenceOperation = (ILocalReferenceOperation)context.Operation;
+                    if (!localReferenceOperation.Local.Type.Inherits(hashAlgoBaseType))
+                    {
+                        return;
+                    }
+                    localReferenceMap.TryAdd(localReferenceOperation, localReferenceOperation.Local);
```
mavasani, line 107: Nit: can just invert if condition and move the TryAdd call inside the if body.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,318 @@
...
+            void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
+            {
+                var computeHashVariableMap = PooledConcurrentDictionary<ILocalReferenceOperation, IInvocationOperation>.GetInstance();
+                var createdSymbolMap = PooledConcurrentDictionary<ILocalSymbol, DeclarationTuple>.GetInstance();
+                var localReferenceMap = PooledConcurrentDictionary<ILocalReferenceOperation, ILocalSymbol>.GetInstance();
+
+                context.RegisterOperationAction(CaptureHashLocalReferenceOperation, OperationKind.LocalReference);
+                context.RegisterOperationAction(CaptureCreateOrComputeHashInvocationOperation, OperationKind.Invocation);
+                context.RegisterOperationAction(CaptureHashObjectCreationOperation, OperationKind.ObjectCreation);
+                context.RegisterOperationBlockEndAction(OnOperationBlockEnd);
```
mavasani, line 97: So it seems like you have a moderately complex state based analysis per each operation block. Would it be possible to add a detailed comment with an example at the start of this method to explain the core logic. See here for a very nice example provided by @NewellClark: https://github.com/dotnet/roslyn-analyzers/blob/d464b5f144f57be88d6af32c779e5ce3949132fa/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferDictionaryContainsMethods.cs#L79-L101

wzchua: I have added the comments

wzchua:
@mavasani friendly ping

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,1188 @@
...
+        Dim aboveLine = 20
+        Dim digest As Byte() = SHA256.HashData(buffer)
+        Dim belowLine = 10
+    End Sub
+End Class
+";
+            await TestVBWithVariable(
+                vbInput,
+                vbFix,
+                "System.Security.Cryptography.SHA256");
```
bartonjs, line 1097: All of these tests used SHA256.  The analyzer/fixer should also detect MD5, SHA1, SHA384, and SHA512.

wzchua: The tests use SHA256 only for convenience. Let me see how I can expand that in a nice pattern

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,1188 @@
...
+            await TestCSWithVariable(
+                csInput,
+                csFix,
+                "System.Security.Cryptography.SHA256");
+        }
+
+        [Fact]
+        public async Task BasicCreateHelperUsingBlockCastedCaseNoFix()
+        {
+            // unable to get type info for SHA256
```
wzchua, line 597: Something has changed since this test was written. The typeinfo can now be obtained.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
wzchua: Updated

wzchua: The analyzer does specifically uses `SHA256.HashData` method as an early termination test but it is not tied to the analyzer search logic

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(PreferHashDataOverComputeHashAnalyzer.CA1848);
+
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var computeHashNode = root.FindNode(context.Span, getInnermostNodeForTie: true);
+            var diagnostics = context.Diagnostics[0];
```
Youssef1313, line 24: rename to singular 'diagnostic'

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (computeHashNode is null || bufferArgNode is null)
+            {
+                return;
+            }
+            var hashTypeName = diagnostics.Properties[PreferHashDataOverComputeHashAnalyzer.TargetHashTypeDiagnosticPropertyKey];
+
+            switch (diagnostics.AdditionalLocations.Count)
+            {
+                case 1:
+                    {
```
Youssef1313, line 36: Are these braces needed?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.Fixer.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            protected HashDataCodeAction(Document document, string hashTypeName, SyntaxNode bufferArgNode, SyntaxNode computeHashNode)
+            {
+                Document = document;
+                HashTypeName = hashTypeName;
+                BufferArgNode = bufferArgNode;
+                ComputeHashNode = computeHashNode;
+            }
+            public override string Title => MicrosoftNetCoreAnalyzersResources.PreferHashDataCodefixTitle;
+            public override string EquivalenceKey => MicrosoftNetCoreAnalyzersResources.PreferHashDataCodefixTitle;
```
Youssef1313, line 77:
```suggestion
            public override string EquivalenceKey => nameof(MicrosoftNetCoreAnalyzersResources.PreferHashDataCodefixTitle);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,339 @@
...
+                    computeHashVariableMap.Free(context.CancellationToken);
+                    createdSymbolMap.Free(context.CancellationToken);
+                    localReferenceMap.Free(context.CancellationToken);
+                }
+
+                bool IsHashCreateMethod(IMethodSymbol methodSymbol)
+                {
+                    return methodSymbol.ContainingType.Inherits(hashAlgoBaseType) &&
+                        methodSymbol.ReturnType.Inherits(hashAlgoBaseType) &&
+                        methodSymbol.Name.Equals(CreateMethodName, StringComparison.Ordinal);
```
buyaa-n, line 232: The value of `CreateMethodName` is `System.Security.Cryptography.SHA256.Create` isn't it mean it will not work for the other hash algorithms?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,339 @@
...
+                // Patterns we are looking for:
+                // Pattern #1                                               Pattern #2
+                // var obj = #HashCreation#;              or                #HashCreation#.CompuateHash(buffer);
+                // ...
+                // obj.CompuateHash(buffer); //x1
+
+                // The core search logic is split into 3 groups
+                // 1. Scan all invocations:
+                //  a. if it is a Hash Create static method, store its local symbol + declaration operation and type symbols
+                //  b. if it is HashAlgorithm.ComputeHash
```
buyaa-n, line 93: Don't we want to cover/replace
`public bool TryComputeHash(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesWritten)` with 
`public static bool TryHashData(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesWritten)`?

buyaa-n (commented):
@wzchua thank you for your contribution, left some comments/questions, overall looks good to go after the questions answered

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,339 @@
...
+                !compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSecurityCryptographySHA256, out var sha256Type))
+            {
+                return;
+            }
+
+            var byteType = context.Compilation.GetSpecialType(SpecialType.System_Byte);
+            if (byteType is null)
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
Youssef1313: @buyaa-n Indeed the check is redundant. But not for that reason. The type can indeed be missing in corner cases, but `GetSpecialType` will return an error symbol instead of null for such cases. The same isn't true for `GetTypeByMetadataName`, which can return null.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
wzchua: The value is `Create`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,1487 @@
...
+using System;
+using System.Security.Cryptography;
+
+public class Test
+{{
+    public static void TestMethod({hashType} hash)
+    {{
+        var buffer = new byte[1024];
+        int aboveLine = 20;
+        byte[] digest = hash.ComputeHash(buffer);
```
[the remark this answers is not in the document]
wzchua: Yes, I strictly followed the rules as described in the issue

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
wzchua: hmm, there's also `ComputeHash(Byte[], Int32, Int32)`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-                
-    There are any number of "resheader" rows that contain simple
```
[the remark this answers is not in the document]
wzchua: this is my editor applying the .editorconfig when I resolved the merge conflict.
Let me disable that and revert this

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
wzchua: So, this would be the expected result? 
I think it's okay
```cs
               string csInput = $@"
using System;
using System.Security.Cryptography;
public class Test
{{
    public static void TestMethod({hashType} hash)
    {{
        var buffer = new byte[1024];
        int aboveLine = 20;
        byte[] digest = {{|#0:hash.ComputeHash({{|#1:buffer|}})|}}
        int belowLine = 10;
    }}
}}
";
                string csFix = $@"
using System;
using System.Security.Cryptography;
public class Test
{{
    public static void TestMethod({hashType} hash)
    {{
        var buffer = new byte[1024];
        int aboveLine = 20;
        byte[] digest = {hashType}.HashData(buffer);
        int belowLine = 10;
    }}
}}
";
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,1487 @@
...
+public class Test
+{{
+    public static void TestMethod()
+    {{
+        var buffer = new byte[1024];
+        var hasher = {hashType}.Create();
+        int aboveLine = 20;
+        byte[] digest = hasher.ComputeHash(buffer);
+        int belowLine = 10;
+        byte[] digest2 = hasher.ComputeHash(buffer);
```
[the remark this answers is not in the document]
wzchua: Hmm, I think it's okay actually. 
For the fixer, I have an idea where if there's more than 1 ComputeHash, it will just replace ComputeHash.
If it's the last/only one, the fixer can delete the `Create` method.

However, would there be an issue with the `Fix All` function?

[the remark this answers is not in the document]
wzchua: It seems I have to add a Fix All

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
```diff
@@ -0,0 +1,1487 @@
...
+{{
+    public static void TestMethod()
+    {{
+        var buffer = new byte[1024];
+        using (var hasher = {hashType}.Create())
+        {{
+            int aboveLine = 20;
+            byte[] digest = hasher.ComputeHash(buffer);
+            int belowLine = 10;
+            byte[] digest2 = hasher.ComputeHash(buffer);
```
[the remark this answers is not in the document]
bartonjs: Yes, this is a safe replacement.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
wzchua: I've opt to disable FixAll

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
wzchua: so if there's no Create, I should not report? I'll update again tmr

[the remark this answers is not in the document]
buyaa-n: > * If the hashType is just `HashAlgorithm` then it's polymorphic and there's no fix.

The analyzer checks if the algorithm has the static HashData(...) method

> * Even if the hashType is one of the algorithms, they're not sealed, polymorphism could have been desired.
> * And even if not, there's data bleed between hash.TransformBlock and hash.ComputeHash, so the incoming object could have "intentionally corrupted" state that's supposed to mix in.

@bartonjs thanks for the clarification 

> so if there's no Create, I should not report? I'll update again tmr

@wzchua sorry about that and thanks for the update!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
buyaa-n: Right forgot that 😆

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
buyaa-n: > hmm, there's also ComputeHash(Byte[], Int32, Int32)

I don't see a direct replacement for that, probably safe to leave it.

@bartonjs is is safe to replace 
`public bool TryComputeHash(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesWritten)` with 
`public static bool TryHashData(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesWritten)`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
buyaa-n: > However, would there be an issue with the Fix All function?
> I've opt to disable FixAll

Not sure, though i am OK with disabling, ping @jmarolf @sharwell

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
bartonjs: `ComputeHash(byte[] arr, int offset, int length)` would map to `HashData(arr.AsSpan(offset, length))`

And, yeah, instance `TryComputeHash` can go to static `TryHashData`.

jeffhandley:
It looks like this is getting close to being ready to merge. Because this would be an Info-level diagnostic by default, I'm comfortable with this getting merged in for the .NET 6.0 RC2 release. Thanks for the contribution on this, @wzchua!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
wzchua: I've done this but I have to get more test written

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
wzchua: > And even if not, there's data bleed between hash.TransformBlock and hash.ComputeHash, so the incoming object could have "intentionally corrupted" state that's supposed to mix in.

does this mean if there's a call to `hash.TransformBlock` and then a call to `hash.ComputeHash`, I shouldn't report?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
GrabYourPitchforks: To simplify the analyzer, my preference would be that it fire only on this simple pattern:

1. The analyzer detects a fresh instantiation of the algorithm (via Create or whatever other mechanism is appropriate); and
2. The *only* thing done with this instance before disposal is one or more calls to ComputeHash.

If anything else at all is done with this instance - passed as an argument to a method, returned out of the method, somebody makes a copy of the local, somebody calls GetType on it, *anything* - the analyzer should not fire. Disposal is optional and is allowed as long as it's the very last thing done with the instance.

This means we don't have to worry about the TransformBlock / TransformFinalBlock cases, and we don't have to worry about other cases where somebody may have mutated the state of the internal object to some unknown value. Even though this analyzer isn't 100% comprehensive, it should be sufficient to capture the 95% usage case.

If we want to be smarter about fixing TransformBlock / TransformFinalBlock, we can save that for a future analyzer.

wzchua: Yes, at the moment, it checks if the instance is only calling `ComputeHash`.

There isn't a check on an explicit call to dispose atm.

wzchua: What's the recommended approach in a roslyn analyzer to detect an invocation is the last invocation?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
wzchua: I gave doing a custom fixallprovider a go but I'm running into differences in the formatting done. How do i replicate the formatting done in a normal codefix?
Is there an api to pass the changed document? Or is creating the SourceText the way in a fixallprovider?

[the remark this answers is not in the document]
buyaa-n: > I gave doing a custom fixallprovider a go but I'm running into differences in the formatting done. How do i replicate the formatting done in a normal codefix?
Is there an api to pass the changed document? Or is creating the SourceText the way to go in a fixallprovider?

I am not familiar with custom fixallprovider, pinging @mavasani @jmarolf @sharwell 

In case there is no such API and it has only issue in formatting i think that is OK

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
buyaa-n: > What's the recommended approach in a roslyn analyzer to detect an invocation is the last invocation?

Don't know if there is a way to detect that, not sure why/for what you need that, i think whatever you are trying to do you can do that in the `OperationBlockEndAction` as that would run after all invocations done

wzchua: There's no check on Dispose being the last method invocation in the logic at the moment.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHashTests.cs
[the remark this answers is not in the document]
wzchua: There's a case that is complicated for the fixall, I'm still looking into that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
buyaa-n: > There's no check on Dispose being the last method invocation in the logic at the moment.

I don't think you need to check if the disposal is the last invocation (if it is not the last invocation then it would throw Disposed exception anyways), i think the only thing matter is `Dispose` is allowed to be called on the instance other than `ComputeHash` which can be done in `OperationBlockEndAction`

wzchua: ok, sound good

wzchua:
Is it alright if I just squashed the commits into 1? It's getting messy imo

buyaa-n (approved):
Thank you @wzchua for your contribution, and applying the comments promptly, left few NITs, overall looks great to me 🎉 

CC @mavasani @jmarolf

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/PreferHashDataOverComputeHash.cs
```diff
@@ -0,0 +1,721 @@
...
+                //      1. if its instance is a local reference, store its local reference + invocation
+                //      2. if its instance is the creation of a hash instance, we found pattern #2. Report diagnostic.
+                //  c. if it is HashAlgorithm.Dispose, store the invocation operation
+                // 2. Find all HashAlgorithm object creation (new) and store its local symbol + declaration operation and type symbols
+                // 3. Find all HashAlgorithm local references and store them
+
+                // At OperationBlockEnd:
+                // 1. Compile a symbol to dispose-invocation-array map
+                // 1. Create a set of local reference whose symbols appears only once excluding dispose invocation
+                // 2. Iterate the invocation, only report the invocation
```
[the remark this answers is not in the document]
wzchua: rewrote the comment

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicPreferHashDataOverComputeHash.Fixer.vb
```diff
@@ -71,21 +71,45 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
...
-                        Return SyntaxFactory.InvocationExpression(hashData, argumentList)
+                        Dim arg = argumentList.Arguments(0)
+                        If arg.IsNamed Then
+                            arg = DirectCast(arg, SimpleArgumentSyntax).WithNameColonEquals(SyntaxFactory.NameColonEquals(SyntaxFactory.IdentifierName("source")))
+                        End If
+                        Dim args = SyntaxFactory.ArgumentList(SyntaxFactory.SingletonSeparatedList(Of ArgumentSyntax)(arg))
+                        Return SyntaxFactory.InvocationExpression(hashData, args)
                     Case PreferHashDataOverComputeHashAnalyzer.ComputeType.ComputeHashSection
+                        Dim list = argumentList.Arguments.ToList()
+                        Dim firstArg = list.Find(Function(a) a.IsNamed = False OrElse DirectCast(a, SimpleArgumentSyntax).NameColonEquals.Name.Identifier.Text.Equals("buffer", StringComparison.Ordinal))
```
Youssef1313, line 82: VB is case insensitive. I think this should be `OrdinalIgnoreCase`.

wzchua: Yes, just tested. This has to be `OrdinalIgnoreCase`.

jeffhandley:
Hooray! Thanks for the contribution @wzchua. 👍

PR #4798, opened 2021-02-07
Non-copyable analysis for fields and auto-properties

sharwell:
* Do not allow a non-copyable field to be defined in a copyable type
* Do not allow a non-copyable auto-property to be defined (the getter in this case always copies the backing field value)

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs
@@ -41,6 +41,12 @@ namespace Roslyn.Diagnostics.Analyzers
         private static readonly LocalizableString s_localizableNoUnboxingMessage = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.DoNotCopyValueNoUnboxingMessage), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
         private static readonly LocalizableString s_localizableNoUnboxingDescription = new LocalizableResourceString(nameof(RoslynDiagnosticsAnalyzersResources.DoNotCopyValueNoUnboxingDescription), RoslynDiagnosticsAnalyzersResources.ResourceManager, typeof(RoslynDiagnosticsAnalyzersResources));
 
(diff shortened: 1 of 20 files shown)


src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs
```diff
@@ -178,6 +238,140 @@ public void Dispose()
...
+
+            public override void DefaultVisit(ISymbol symbol)
+            {
+                base.DefaultVisit(symbol);
+            }
+
+            public override void VisitAlias(IAliasSymbol symbol)
+            {
+                base.VisitAlias(symbol);
+            }
```
CyrusNajmabadi, line 267: i don't quite get the purpose of the overrides.  is it just to be explicit that this is the behavior you want?  (if so, that's fine.  just trying to understand).

src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs
```diff
@@ -178,6 +238,140 @@ public void Dispose()
...
+
+            public override void VisitPointerType(IPointerTypeSymbol symbol)
+            {
+                base.VisitPointerType(symbol);
+            }
+
+            public override void VisitProperty(IPropertySymbol symbol)
+            {
+                // Auto-properties cannot have non-copyable types
```
CyrusNajmabadi, line 354: probably worth commenting why?

CyrusNajmabadi (approved):

src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs
sharwell: Yes, this analyzer overrides everything.

src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs
sharwell: The user-facing description of the rule includes the explanation. The getter for an auto-property always copies the backing field, which is not allowed for a non-copyable type.

sharwell: I updated the code to include more complete descriptions. The user facing messages still have descriptions as well.

PR #4806, opened 2021-02-09
(4/4) Prefer 'AsSpan' over 'Substring' analyzer

NewellClark:
Fixes [issue #33784](https://github.com/dotnet/runtime/issues/33784). 
Violations found in dotnet/runtime:
- `System.Private.DataContractSerialization` ([PR](https://github.com/dotnet/runtime/pull/48274))
- `System.Private.Xml` ([PR](https://github.com/dotnet/runtime/pull/48277))

Edit: ambiguous cases are now reported, but not fixed. 

> I'm marking this as a draft for now because I wanted to ask what we should do when only some of the `Substring` calls can be converted to `AsSpan` (currently it only reports when all `Substring` invocations in the call can be converted). 
> ```C#
> using Roschar = System.ReadOnlySpan<char>
> // No overload takes 3 roschars
> public void Consume(string a, string b, string c) { }
> public void Consume(Roschar a, string b, string c) { }
> public void Consume(string a, Roschar b, Roschar c) { }
> 
> public void NotAmbiguous(string foo)
> {
>     Consume(foo.Substring(1), foo.Substring(2), foo.Substring(3));
> 
>     // Candidates
>     Consume(foo.AsSpan(1), foo.Substring(2), foo.Substring(3));    // Option 1
>     Consume(foo.Substring(1), foo.AsSpan(2), foo.AsSpan(3));       // Option 2
> }
> ```
> Should we mirror overload resolution and have the fixer choose the "best" option here? Seems like the right thing to do. But what about when it's ambiguous?
> 
> ```C#
> using Roschar = System.ReadOnlySpan<char>;
> 
> public void Consume(string left, string right) { }
> public void Consume(Roschar left, string right) { }
> public void Consume(string left, Roschar right) { }
> 
> public void Ambiguous(string foo)
> {
>     Consume(foo.Substring(1), foo.Substring(2));
> 
>     // Candidates
>     Consume(foo.AsSpan(1), foo.Substring(2));    // Option 1
>     Consume(foo.Substring(1), foo.AsSpan(2));    // Option 2
>  }
> ```
> When it's ambiguous, should we not report at all? Or should we report, but not fix?


sharwell:
Is this a common case?

I would typically prefer one of the following:

* Report a diagnostic and provide two possible fixes
* Report a diagnostic but do not provide a fix

NewellClark:
> * Report a diagnostic but do not provide a fix

Sounds reasonable. I don't think it's super common, so I'll do this for now, unless we find evidence that it's a common scenario.

NewellClark:
@sharwell I've implemented it so that ambiguous cases are reported, but not fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="PreferAsSpanOverSubstringDescription" xml:space="preserve">
+    <value>'AsSpan' accomplishes the same thing as 'Substring' without incurring the allocation and O(n) string copy.</value>
+  </data>
+  <data name="PreferAsSpanOverSubstringMessage" xml:space="preserve">
+    <value>Prefer 'AsSpan' over 'Substring' when span-based overloads are available.</value>
```
Evangelink, line 1517: Looks like the resx analyzer is not working. Message is not supposed to end with period.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.Fixer.cs
```diff
@@ -0,0 +1,100 @@
...
+            if (!RequiredSymbols.TryGetSymbols(compilation, out RequiredSymbols symbols))
+                return;
+
+            var bestCandidates = PreferAsSpanOverSubstring.GetBestSpanBasedOverloads(symbols, reportedInvocation, context.CancellationToken);
+            //  We only apply the fix if there is an unambiguous best overload.
+            if (bestCandidates.Length != 1)
+                return;
+            IMethodSymbol spanBasedOverload = bestCandidates[0];
+
+            string title = MicrosoftNetCoreAnalyzersResources.PreferAsSpanOverSubstringTitle;
```
Evangelink, line 47: It's better to have a separate entry for the codefix title. We usually prefer them to correspond to the action, in this case you probably want the title to be `Replace 'Substring' with 'AsSpan'`.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferAsSpanOverSubstring.Fixer.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class CSharpPreferAsSpanOverSubstringFixer : PreferAsSpanOverSubstringFixer
+    {
+        private protected override void ReplaceInvocationMethodName(SyntaxEditor editor, SyntaxNode memberInvocation, string newName)
+        {
+            var cast = (InvocationExpressionSyntax)memberInvocation;
+            var memberAccessSyntax = (MemberAccessExpressionSyntax)cast.Expression;
```
Evangelink, line 19: Are we sure we can't be on something else than `MemberAccessExpresionSyntax` here? Maybe a member binding or something like this.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+{
+    /// <summary>
+    /// CA1842: Prefer 'AsSpan' over 'Substring'.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class PreferAsSpanOverSubstring : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1842";
+
+        private static readonly LocalizableString s_localizableTitle = CreateResource(nameof(Resx.PreferAsSpanOverSubstringTitle));
```
Evangelink, line 26: Could you please inline the descriptors? I don't remember exactly what was the issue but we had some hidden problem with using a helper method.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+        }
+
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            if (!RequiredSymbols.TryGetSymbols(context.Compilation, out RequiredSymbols symbols))
+                return;
+
+            context.RegisterOperationBlockStartAction(OnOperationBlockStart);
+
+            void OnOperationBlockStart(OperationBlockStartAnalysisContext context)
```
Evangelink, line 56: [minor] Method could be made `static`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+                var invocations = PooledConcurrentSet<IInvocationOperation>.GetInstance();
+
+                context.RegisterOperationAction(context =>
+                {
+                    var argument = (IArgumentOperation)context.Operation;
+                    if (!symbols.IsAnySubstringInvocation(WalkDownImplicitConversions(argument.Value)))
+                        return;
+                    if (argument.Parent is not IInvocationOperation invocation)
+                        return;
+                    invocations.Add(invocation);
```
Evangelink, line 67: Feel free to ignore this suggestion

```suggestion
                    if (symbols.IsAnySubstringInvocation(WalkDownImplicitConversions(argument.Value))
                        && argument.Parent is not IInvocationOperation invocation)
                    {
                        invocations.Add(invocation);
                    }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+                            Diagnostic diagnostic = invocation.CreateDiagnostic(Rule);
+                            context.ReportDiagnostic(diagnostic);
+                        }
+                    }
+                    invocations.Free(context.CancellationToken);
+                });
+            }
+        }
+
+        internal static IOperation WalkDownImplicitConversions(IOperation operation)
```
Evangelink, line 88: I am wondering if we should update the `WalkDownConversions` helper to have an optional predicate function `Func<IConversionOperation, bool>`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+                if (symbol is not INamedTypeSymbol type)
+                    type = symbol.ContainingType;
+
+                return type;
+            }
+        }
+
+        private static bool DerivesFromOrEqualTo(ITypeSymbol derived, ITypeSymbol candidateBase)
+        {
+            return derived.DerivesFrom(candidateBase, baseTypesOnly: true) || SymbolEqualityComparer.Default.Equals(derived, candidateBase);
```
Evangelink, line 221: Looking at the implementation of `DerivesFrom` it seems to already handle the equals case.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+                symbols = new RequiredSymbols(
+                    stringType, roscharType,
+                    memoryExtensionsType,
+                    substring1, substring2,
+                    asSpan1, asSpan2);
+                return true;
+            }
+
+            public INamedTypeSymbol StringType { get; }
+            public INamedTypeSymbol RoscharType { get; }
```
Evangelink, line 291: [minor] Could we use a more explicit name? I had to look at the creation of the `roscharType` variable to understand what this is about.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,312 @@
...
+            public INamedTypeSymbol MemoryExtensionsType { get; }
+            public IMethodSymbol Substring1 { get; }
+            public IMethodSymbol Substring2 { get; }
+            public IMethodSymbol AsSpan1 { get; }
+            public IMethodSymbol AsSpan2 { get; }
```
Evangelink, line 296: [minor] Same remark on naming. May I suggest something like:
`XXX1` => `XXXStart`
`XXX2` => `XXXStartLength`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: I think that would be a good idea. I've wanted that function on several other analyzers. While we're talking about upgrading helpers, I've also wanted a `GetFirstOrDefaultWithParamater` overload that takes `ITypeSymbol`s instead of `ParameterInfo`s.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferAsSpanOverSubstring.Fixer.cs
NewellClark: The only way this could be a `MemberBindingExpressionSyntax` is if the substring invocation is conditional, `foo?.Substring(1)`. 
This can't possibly be true because the analyzer detects `IArgumentOperation`s, and after walking down implicit conversions, the syntax of `foo?.Substring(1)` would be `ConditionalAccessExpressionSyntax`, and not `InvocationExpressionSyntax`.
I'll add some assertions and comments to enforce this invariant.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: It captures `symbols`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.Fixer.cs
NewellClark: Added.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed. I'll also fix this on my other analyzer PRs.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed. I'll also fix this on my other PRs.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
Evangelink: My bad, it's not easy to see captured variables in github.

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
```diff
@@ -12,7 +12,7 @@
 Design: CA2210, CA1000-CA1070
 Globalization: CA2101, CA1300-CA1310
 Mobility: CA1600-CA1601
-Performance: HA, CA1800-CA1838
+Performance: HA, CA1800-CA1842
```
Evangelink, line 15: Why the jump from 38 to 42?

src/Utilities/Compiler/Extensions/IEnumerableOfIMethodSymbolExtensions.cs
```diff
@@ -147,6 +147,42 @@ internal static class IEnumerableOfIMethodSymbolExtensions
...
+        {
+            if (members is null)
+                return null;
+
+            foreach (var member in members)
+            {
+                if (Predicate(member))
+                    return member;
+            }
+            return null;
```
Evangelink, line 171:
```suggestion

            return null;
```

src/Utilities/Compiler/Extensions/IEnumerableOfIMethodSymbolExtensions.cs
```diff
@@ -147,6 +147,42 @@ internal static class IEnumerableOfIMethodSymbolExtensions
...
+            {
+                if (member.Parameters.Length != expectedParameterTypesInOrder.Count)
+                    return false;
+
+                for (int index = 0; index < expectedParameterTypesInOrder.Count; index++)
+                {
+                    if (!member.Parameters[index].Type.Equals(expectedParameterTypesInOrder[index]))
+                        return false;
+                }
+                return true;
```
Evangelink, line 183:
```suggestion

                return true;
```

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -667,6 +667,23 @@ public static IOperation WalkDownConversion(this IOperation operation)
...
         }
 
+        /// <summary>
+        /// Walks down consecutive conversion operations that satisfy <paramref name="predicate"/> until an operand is reached that
+        /// either isn't a conversion or doesn't satisfy <paramref name="predicate"/>.
+        /// </summary>
+        /// <param name="operation">The starting operation.</param>
+        /// <param name="predicate">A predicate to filter conversion operations.</param>
+        /// <returns>The first operation that either isn't a conversion or doesn't satisfy <paramref name="predicate"/>.</returns>
+        public static IOperation WalkDownConversion(this IOperation operation, Func<IConversionOperation, bool> predicate)
```
Evangelink, line 677: Maybe this could be merged with the previous and allow nullable `Func`. Note that I am not convinced whether it brings much.

Evangelink:
@sharwell @mavasani Could one of you review this PR? Thank you.

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
NewellClark: Because my other analyzer PRs use CA1839-CA1841.

Evangelink: Makes sense, maybe you could add a [X/Y] prefix on your PRs so that we merge them in the right order.

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
NewellClark: I did end up needing it on [one of my other PRs](https://github.com/dotnet/roslyn-analyzers/pull/4764/files). I can see it being useful in the future, especially for dealing with arguments where you don't want to skip through explicit casts.

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
NewellClark: I've prefixed them [(A)](https://github.com/dotnet/roslyn-analyzers/pull/4687), [(B)](https://github.com/dotnet/roslyn-analyzers/pull/4726), [(C)](https://github.com/dotnet/roslyn-analyzers/pull/4764), and this one is (D). I think some of them may be in different milestones. Do you want me to change the IDs around so that earlier milestones get merged first?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferAsSpanOverSubstring.Fixer.cs
```diff
@@ -0,0 +1,41 @@
...
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class CSharpPreferAsSpanOverSubstringFixer : PreferAsSpanOverSubstringFixer
+    {
+        private protected override void ReplaceNonConditionalInvocationMethodName(SyntaxEditor editor, SyntaxNode memberInvocation, string newName)
+        {
+            var cast = (InvocationExpressionSyntax)memberInvocation;
+
+            //  We know that cast.Expression can't be a MemberBindingExpressionSyntax because the analyzer only reports 
+            //  non-conditional substring invocations, i.e. not `foo?.Substring(1)`
+            RoslynDebug.Assert(cast.Expression is MemberAccessExpressionSyntax);
```
mavasani, line 23: This assert seems redundant given the direct cast below is exactly the same - it doesn't matter if the assert fires or a cast exception is thrown.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1534,6 +1534,18 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="PreferAsSpanOverSubstringDescription" xml:space="preserve">
+    <value>'Substring' performs an O(n) string copy, while 'AsSpan' does not.</value>
```
mavasani, line 1538: This should be more descriptive. Maybe:
```suggestion
    <value>'AsSpan' is more efficient then 'Substring'. 'Substring' performs an O(n) string copy, while 'AsSpan' does not and has a constant cost.</value>
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpPreferAsSpanOverSubstring.Fixer.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,294 @@
...
+                        //  We search for an overload of the invoked member whose signature matches the signature of
+                        //  the invoked member, except with ReadOnlySpan<char> substituted in for some of the 
+                        //  arguments that are Substring invocations.
+                        if (!GetBestSpanBasedOverloads(symbols, invocation, context.CancellationToken).IsEmpty)
+                        {
+                            Diagnostic diagnostic = invocation.CreateDiagnostic(Rule);
+                            context.ReportDiagnostic(diagnostic);
+                        }
+                    }
+                    invocations.Free(context.CancellationToken);
```
mavasani, line 82: Nit: add newline between closing brace and next statement
```suggestion

                    invocations.Free(context.CancellationToken);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,294 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(OnCompilationStart);
+        }
+
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            if (!RequiredSymbols.TryGetSymbols(context.Compilation, out RequiredSymbols symbols))
+                return;
+
+            context.RegisterOperationBlockStartAction(OnOperationBlockStart);
```
mavasani, line 54:
```suggestion
            context.RegisterOperationBlockStartAction(OnOperationBlockStart);
            return;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,294 @@
...
+                    {
+                        return -1;
+                    }
+                }
+
+                return replacementCount;
+            }
+        }
+
+        private static IEnumerable<IMethodSymbol> GetAllAccessibleOverloadsIncludingSelf(IInvocationOperation invocation, CancellationToken cancellationToken)
```
mavasani, line 171: Can we use this extension method? https://github.com/dotnet/roslyn-analyzers/blob/1fa83fe751c6e245f20ab57cd863b5d17098520b/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs#L441

NewellClark: That method is poorly named. Basically, it looks at a particular invocation operation and finds all overloads of the invoked method that would also be legal at that location, taking into account the type of the receiver (for instance methods) or type (for static methods). It's pretty specific to this analyzer; I can't make a version that uses an `IMethodSymbol` instead of an `IInvocationOperation` because it depends on the receiver of the invocation. I'll rename it to something more descriptive to make this clearer. 

If we did want to add an extension method that does this, it couldn't go in `IMethodSymbolExtensions` because the extended receiver would need to be `IInvocationOperation`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.Fixer.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var document = context.Document;
+            var token = context.CancellationToken;
+            SyntaxNode root = await document.GetSyntaxRootAsync(token).ConfigureAwait(false);
+            SemanticModel model = await document.GetSemanticModelAsync(token).ConfigureAwait(false);
+            var compilation = model.Compilation;
+
+            if (root.FindNode(context.Span, getInnermostNodeForTie: true) is not SyntaxNode reportedNode || model.GetOperation(reportedNode, token) is not IInvocationOperation reportedInvocation)
```
mavasani, line 37: Nit: can you please move the second part of the condition to the next line for readability?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.Fixer.cs
```diff
@@ -0,0 +1,101 @@
...
+                return;
+            if (!RequiredSymbols.TryGetSymbols(compilation, out RequiredSymbols symbols))
+                return;
+
+            var bestCandidates = PreferAsSpanOverSubstring.GetBestSpanBasedOverloads(symbols, reportedInvocation, context.CancellationToken);
+            //  We only apply the fix if there is an unambiguous best overload.
```
mavasani, line 43: Nit: blank line before comment
```suggestion

            //  We only apply the fix if there is an unambiguous best overload.
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.Fixer.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (bestCandidates.Length != 1)
+                return;
+            IMethodSymbol spanBasedOverload = bestCandidates[0];
+
+            string title = MicrosoftNetCoreAnalyzersResources.PreferAsSpanOverSubstringCodefixTitle;
+            var codeAction = CodeAction.Create(title, CreateChangedDocument, title);
+            context.RegisterCodeFix(codeAction, context.Diagnostics);
```
mavasani, line 50:
```suggestion
            context.RegisterCodeFix(codeAction, context.Diagnostics);
            return;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,298 @@
...
+            var model = invocation.SemanticModel;
+            int location = invocation.Syntax.SpanStart;
+            var instance = invocation.Instance;
+
+            IEnumerable<IMethodSymbol> allOverloads;
+            if (method.IsStatic)
+            {
+                allOverloads = model.LookupStaticMembers(location, method.ContainingType, method.Name).OfType<IMethodSymbol>();
+            }
+            //  Ensure protected members can only be invoked on instances that are known to be instances of the accessing class.
```
mavasani, line 187: Nit
```suggestion

            //  Ensure protected members can only be invoked on instances that are known to be instances of the accessing class.
```

mavasani, line 187: Please do no add comments between and if and else blocks. Can you either move the comment at start of the if-else OR inside the else block?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,298 @@
...
+            else if (instance is not null)
+            {
+                var enclosingType = GetEnclosingType(model, location, cancellationToken);
+                allOverloads = model.LookupSymbols(location, instance.Type, method.Name).OfType<IMethodSymbol>();
+                if (instance.Type.DerivesFrom(enclosingType, baseTypesOnly: true) || instance is IInstanceReferenceOperation)
+                {
+                    allOverloads = allOverloads.Union(model.LookupBaseMembers(location, method.Name).OfType<IMethodSymbol>());
+                }
+            }
+            //  This can happen when compiling invalid code.
```
mavasani, line 197: Same here - please move the comment

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,298 @@
...
+            else
+            {
+                allOverloads = Enumerable.Empty<IMethodSymbol>();
```
mavasani, line 200:
```suggestion
                return Enumerable.Empty<IMethodSymbol>();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -0,0 +1,298 @@
...
+            }
+
+            return allOverloads.Where(x => x.IsStatic == method.IsStatic && SymbolEqualityComparer.Default.Equals(x.ReturnType, method.ReturnType));
+
+            static INamedTypeSymbol GetEnclosingType(SemanticModel model, int location, CancellationToken cancellationToken)
+            {
+                ISymbol symbol = model.GetEnclosingSymbol(location, cancellationToken);
```
mavasani, line 207: Can this be null? If so, can `symbol.ContainingType` below throw a null reference exception?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstringTests.cs
```diff
@@ -0,0 +1,1854 @@
...
+                    3, 1, 1
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(Data_NestedViolations))]
+        public Task NestedViolations_AreAllReportedAndFixed_CS(
+            string receiverClass, string testExpression, string fixedExpression, int[] locations,
+            int? incrementalIterations = null, int? fixAllInDocumentIterations = null, int? fixAllIterations = null)
```
mavasani, line 699: All the yield return values in `Data_NestedViolations` seems to provide non-null values for these 3 parameters. Can these parameters be marked non-optional?

mavasani: Also it seems `fixAllInDocumentIterations` and `fixAllIterations` is always 1. Do we need a parameter for these as all?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstringTests.cs
```diff
@@ -0,0 +1,1854 @@
...
+        {
+            get
+            {
+                string members = @"
+public void Consume(string a, string b) { }
+public void Consume(Roschar a, string b) { }
+public void Consume(string a, Roschar b) { }";
+                yield return new[]
+                {
+                    CS.WithBody(WithKey(@"Consume(foo.Substring(1), foo.Substring(2))", 0) + ';', members)
```
mavasani, line 1103: We can potentially consider showing 2 code fixes here, one for each best candidate, but given this would be a rare case, current approach of not showing code fix is also reasonable.

mavasani (approved):
Overall looks great to me and extensive unit tests give a lot of confidence, thanks!

Have some minor comments - I can merge once they are resolved.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstringTests.cs
NewellClark: @mavasani I'm happy to implement that functionality. How do I write tests that expect two code fixes for one violation? Is there another analyzer that does this that I could look at as an example?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: https://github.com/dotnet/roslyn-analyzers/blob/3f765296d9ea2fbcd88252a472e603b6c700e363/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs#L190
I don't believe so. Even if the operation is a top-level statement, `GetEnclosingSymbol` should still return the `IMethodSymbol` representing the implicit entry-point. I'll add a test to ensure this is the case.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstringTests.cs
NewellClark: I think at one point I had different values for these parameters. I'll mark them non-optional and remove `fixAllInDocumentIterations` and `fixAllIterations`, as they are always 1.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.Fixer.cs
NewellClark: Fixed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstringTests.cs
mavasani: @sharwell Might have suggestions on this - I don't believe we need this as part of this PR. Feel free to file a separate issue or open a follow-up PR for it.

mavasani:
@NewellClark Can you please resolve merge conflicts?

mavasani (approved):

PR #4807, opened 2021-02-09
Introduce a base class for globalization analyzers

### LINKED ISSUE #4629: Globalization rules should not be raised when InvariantGlobalization is set
### Analyzer

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/globalization-warnings

### Describe the improvement

Currently these rules are all firing (e.g. CA1310 Specify StringComparison for correctness) despite having InvariantGlobalization turned on in the project file

```
<PropertyGroup>
    <InvariantGlobalization>true</InvariantGlobalization>
</PropertyGroup>
```

as per https://github.com/dotnet/runtime/blob/master/docs/design/features/globalization-invariant-mode.md

This option makes the warnings moot, so they should not be raised.

Youssef1313:
Fixes #4629

Tagging @tarekgh for confirmation that [Globalization](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/globalization-warnings) rules shouldn't be produced if the option is set to true.

Hide whitespace changes when viewing diff will ease reviewing.

### DIFF
--- a/src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
+++ b/src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.NetAnalyzers
+{
+    public abstract class AbstractGlobalizationDiagnosticAnalyzer : DiagnosticAnalyzer
+    {
+        protected virtual GeneratedCodeAnalysisFlags GeneratedCodeAnalysisFlags { get; } = GeneratedCodeAnalysisFlags.None;
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags);
+            context.RegisterCompilationStartAction(context =>
+            {
+                var value = context.Options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.InvariantGlobalization, context.Compilation, context.CancellationToken);
+                if (value is not "true")
+                {
+                    InitializeWorker(context);
+                }
+            });
+        }
+
+        protected abstract void InitializeWorker(CompilationStartAnalysisContext compilationContext);
+    }
+}
--- a/src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
+++ b/src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
@@ -16,13 +16,13 @@ namespace Microsoft.CodeAnalysis.NetAnalyzers
             context.RegisterCompilationStartAction(context =>
             {
                 var value = context.Options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.InvariantGlobalization, context.Compilation, context.CancellationToken);
-                if (value is not "true")
+                if (value != "true")
                 {
                     InitializeWorker(context);
                 }
             });
         }
 
-        protected abstract void InitializeWorker(CompilationStartAnalysisContext compilationContext);
+        protected abstract void InitializeWorker(CompilationStartAnalysisContext context);
     }
 }
--- a/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
+++ b/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
@@ -104,6 +104,11 @@ namespace GenerateDocumentationAndConfigFiles
 
                 foreach (var analyzer in analyzers)
                 {
+                    if (analyzer.SupportedDiagnostics.Length == 0)
+                    {
+                        continue;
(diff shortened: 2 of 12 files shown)


src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -108,6 +108,11 @@ public static int Main(string[] args)
 
                     foreach (var rule in analyzer.SupportedDiagnostics)
                     {
+                        if (rule.Category == "Globalization")
+                        {
+                            Debug.Assert(analyzerType.BaseType?.Name == "AbstractGlobalizationDiagnosticAnalyzer");
+                        }
```
Youssef1313, line 114: This doesn't seem to be the correct/ideal place for this assertion. It makes more sense if this is turned into a unit test. Placing it here temporarily.

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -108,6 +108,11 @@ public static int Main(string[] args)
 
                     foreach (var rule in analyzer.SupportedDiagnostics)
                     {
+                        if (rule.Category == "Globalization")
+                        {
+                            Debug.Assert(analyzerType.BaseType?.Name == "AbstractGlobalizationDiagnosticAnalyzer", $"Analyzer {analyzerType.Name} was expected to inherit AbstractGlobalizationDiagnosticAnalyzer.");
```
Youssef1313, line 113: This may need to be converted to a loop (i.e. checking the inheritance hierarchy, not only the direct parent)

Youssef1313:
There seem to be CA1508 warnings that are **correct**. I can't figure out why these doesn't exist in master?

src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.NetAnalyzers
+{
+    public abstract class AbstractGlobalizationDiagnosticAnalyzer : DiagnosticAnalyzer
+    {
+        protected virtual GeneratedCodeAnalysisFlags GeneratedCodeAnalysisFlags { get; } = GeneratedCodeAnalysisFlags.None;
```
Youssef1313, line 10: Should I remove this and always use `None`?

safern:
cc: @jeffhandley @GrabYourPitchforks @mavasani 

I think it makes sense to not raise the warnings if the MSBuild property is set to true.

jeffhandley (approved):
LGTM. @buyaa-n or @carlossanlop: could one of you review this too please?

buyaa-n (commented):
Overall looks good to me, thanks @Youssef1313, could you add tests with the property set for those analyzers?

src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
buyaa-n: i think we can keep it for future use

src/NetAnalyzers/Core/AbstractGlobalizationDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public sealed override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags);
+            context.RegisterCompilationStartAction(context =>
+            {
+                var value = context.Options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.InvariantGlobalization, context.Compilation, context.CancellationToken);
+                if (value is not "true")
```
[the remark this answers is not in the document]
Youssef1313: Oh sorry for missing this PR. I don't know if it's case sensitive or not. Let me use `!=` for now. I'm not sure why I used `is not` :smile:

```suggestion
                if (value != "true")
```

buyaa-n (approved):
Thanks @Youssef131, there is one outstanding question did you tried to add tests with the property set (using editconfig file) for those analyzers? No need to be for all, but at least for one of them? Overall looks good, it would be great if we could have one test real scenario

buyaa-n:
Thanks, @Youssef1313, looks great to me. @mavasani do you have any comments/concerns?

mavasani (approved):

alexrp:
@Youssef1313 @tarekgh Sorry to dig up this old PR, but I have a concern about CA1307: It seems to me that this rule still adds value even with `InvariantGlobalization` in that it promotes clarity. It's not always clear if an API is case sensitive or not which is an aspect of string comparison that still matters with `InvariantGlobalization` (for ASCII text anyway).

For an example of an API that isn't clear about this, see ASP.NET Core's [`PathString.StartsWithSegments()`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.pathstring.startswithsegments) which performs a case-**insensitive** comparison by default. Without `InvariantGlobalization`, CA1307 flags calls to this method which I would say is helpful, and I'd like for it to do so even with `InvariantGlobalization`.

PR #4829, opened 2021-02-11
Comment out rule ids that're ported to roslyn

Youssef1313:
Related to dotnet/roslyn#50358

I made this change to avoid accidental removal/re-use of these IDs.

### DIFF
--- a/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs
+++ b/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs
@@ -54,8 +54,8 @@ namespace Roslyn.Diagnostics.Analyzers
         public const string PublicApiFileMissing = "RS0048";
         public const string TemporaryArrayAsRefRuleId = "RS0049";
 
-        public const string WrapStatementsRuleId = "RS0100";
-        public const string BlankLinesRuleId = "RS0101";
-        public const string BracePlacementRuleId = "RS0102";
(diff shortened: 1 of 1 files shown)


src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticIds.cs
```diff
@@ -54,8 +54,8 @@ internal static class RoslynDiagnosticIds
         public const string PublicApiFileMissing = "RS0048";
         public const string TemporaryArrayAsRefRuleId = "RS0049";
 
-        public const string WrapStatementsRuleId = "RS0100";
-        public const string BlankLinesRuleId = "RS0101";
-        public const string BracePlacementRuleId = "RS0102";
+        //public const string WrapStatementsRuleId = "RS0100"; // Now ported to dotnet/roslyn https://github.com/dotnet/roslyn/pull/50358
+        //public const string BlankLinesRuleId = "RS0101"; // Now ported to dotnet/roslyn https://github.com/dotnet/roslyn/pull/50358
+        //public const string BracePlacementRuleId = "RS0102"; // Now ported to dotnet/roslyn https://github.com/dotnet/roslyn/pull/50358
```
Youssef1313, line 59: I assume these IDs shouldn't be re-used.

If I'm wrong, then we can entirely delete it instead of commenting it.

mavasani: Fine with just deleting these.

sharwell: No, this was definitely correct.

sharwell (approved):

PR #4838, opened 2021-02-12
CA-1418 Validate platform string

buyaa-n:
### Validates platform string 

Provided in constructor arguments `[SupportedOSPlatform("platform")]`, `[UnsupportedOSPlatform("platform")]` attributes and` OperatingSystem.IsOSPlatform("platform")`, `OperatingSystem.IsOSPlatformVersionAtLeast("platform", 1)` methods

Fixes https://github.com/dotnet/runtime/issues/45851

- Validates if platform name is known. The know platform names list is the combined set from two places:
  1. OperatingSystem.Is(PlatformName)[VersionAtLeast](). That's the target framework's known set of platforms.
  2. The project's MSBuild item group SupportedPlatform. This is the project's specific knowledge of known platforms. This allows class library authors to add custom platform name
- Optionally validates version part if provided

Example: 

```C#
class Some
{    
    public static void GuardedCall()
    {
        if (OperatingSystem.IsOSPlatform("x")) // The platform 'x' is not a known platform name
            SomeType.Api1();

        if (OperatingSystem.IsOSPlatformVersionAtLeast("x", 1)) // The platform 'x' is not a known platform name
            SomeType.Api1();
    }
}

class SomeType
{
    [SupportedOSPlatform("x")] // The platform 'x' is not a known platform name
    public static void Api1() { }

    [UnsupportedOSPlatform("y")] // The platform 'y' is not a known platform name
    public static void Api2() { }

    // The '7' is not a valid version. Please provide an input having two to four numbers separated with a dot.
    [UnsupportedOSPlatform("windows7")] 
    public event EventHandler Windows7Event;
}
```


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        [Fact]
+        public async Task InvalidPlatformNameOrVersionWarns()
+        {
+            var csSource = @"
+using System.Runtime.Versioning;
+
+public class Test
+{
+    [[|UnsupportedOSPlatform(""window7.0"")|]] // The platform 'window' is not a known platform name
```
buyaa-n, line 113: Did not found a way for reporting the diagnostics at the attribute's argument's location. Seems I could do some workaround like creating a new location by subtracting the attribute name length or something like that, but not sure if that is the desired approach, any suggestions @mavasani?

jeffhandley (approved):
Looks good to me. I just have a question about the versions that might have been covered during the API review of this already: Should we respect the number of version segments each OSPlatform supports? Several only support 3 segments; some don't support versions at all.

buyaa-n:
> Looks good to me. I just have a question about the versions that might have been covered during the API review of this already: Should we respect the number of version segments each OSPlatform supports? Several only support 3 segments; some don't support versions at all.

I don't think that is not covered during API review, i am not sure if that information for all OS are available somewhere, if we could get that info we can add that check

jeffhandley:
If the platform was found in the `OperatingSystem.Is<Platform>` check, we could check to see if there's a `...VersionAtLeast` function as well. If not, versions aren't supported; if so, then the min/max numbers of parameters would define the supported segments.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,176 @@
...
+                context.ReportDiagnostic(argument.Syntax.CreateDiagnostic(UnknownPlatform, platformName));
+            }
+        }
+
+        private static void AnalyzeSymbol(Action<Diagnostic> reportDiagnostic, ISymbol symbol, INamedTypeSymbol supportedAttrbute,
+            INamedTypeSymbol unsupportedAttribute, ImmutableArray<string> knownPlatforms, CancellationToken token)
+        {
+            foreach (var attribute in symbol.GetAttributes())
+            {
+                if (supportedAttrbute.Equals(attribute.AttributeClass.OriginalDefinition) ||
```
Youssef1313, line 106: I think this should use SymbolEqualityComprare.
Currently RS1024 is disabled due to a codefix bug, usually it reports these.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,176 @@
...
+                    unsupportedAttribute.Equals(attribute.AttributeClass.OriginalDefinition))
```
Youssef1313, line 107: Same as above.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
```diff
@@ -0,0 +1,389 @@
...
+        }
+
+        private static VerifyVB.Test PopulateTestVb(string sourceCode, params DiagnosticResult[] expected)
+        {
+            var test = new VerifyVB.Test
+            {
+                TestCode = sourceCode,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                MarkupOptions = MarkupOptions.UseFirstDescriptor,
+                TestState = { },
```
Youssef1313, line 383: Is this needed? If so, maybe consider a comment?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
```diff
@@ -0,0 +1,389 @@
...
+
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.InteropServices.UseValidPlatformString,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices.UnitTests
+{
+    public class UseValidPlatformStringTest
+    {
+        private const string s_msBuildPlatforms = "build_property._SupportedPlatformList=CustomPlatform";
```
Youssef1313, line 18: AFAIK, some editorconfig options doesn't do anything without something like `[*.cs]`.

@mavasani Can you confirm whether this is correct for this *specific* option?

If so, then there is a test gap.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
```diff
@@ -0,0 +1,389 @@
...
+            await test.RunAsync();
+        }
+
+        private static VerifyCS.Test PopulateTestCs(string sourceCode, params DiagnosticResult[] expected)
+        {
+            var test = new VerifyCS.Test
+            {
+                TestCode = sourceCode,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                MarkupOptions = MarkupOptions.UseFirstDescriptor,
```
Youssef1313, line 356: Is the other descriptor not tested currently?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
buyaa-n: I think it was needed when setting `params DiagnosticResult[]`, but for these tests that is not used, i might remove it

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
buyaa-n: No, i guess if the analyzer has multiple descriptors it requires to choose which one to use, but for the test scenarios there is no case it could return more than one descriptor

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
buyaa-n: > AFAIK, some editorconfig options doesn't do anything without something like [*.cs].

hm, not sure, i was using it like this previously and it still works here too

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
Youssef1313: > for the test scenarios there is no case it could return more than one descriptor

I mean which tests use which descriptor?

If I understand this property correctly, it means that diagnostics specified as markup (i.e. `[|SomeCode|]`) are expected to have the first descriptor. And because all tests are using it, this means the other descriptor is never encountered in tests.

Correct me if I'm wrong please.

Youssef1313: Never mind, Just read the documentation and found:

> Use the first ***matching*** diagnostic descriptor when multiple diagnostics match the syntax.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
buyaa-n: > I think this should use SymbolEqualityComprare.

Is that a type? I cannot find that in the repo

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
buyaa-n: > Use the first matching diagnostic descriptor when multiple diagnostics match the syntax.

yes, and i meant there is no test/syntax returning multiple descriptor

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
buyaa-n: never mind, there was a typo, found it

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
jeffhandley: I think showing the diagnostic on the entire attribute is just fine.

buyaa-n:
> If the platform was found in the `OperatingSystem.Is<Platform>` check, we could check to see if there's a `...VersionAtLeast` function as well. If not, versions aren't supported; if so, then the min/max numbers of parameters would define the supported segments.

@jeffhandley added validation for the number of version segments each OSPlatform supports, as version is optional we need only a max number of parameters. You might want to review messaging, sample messages are below:

```csharp
// versions are not supported case:
[SupportedOSPlatform("Linux4.1")|}] // The platform 'Linux' is not versioned, please remove version part
public void SupportedLinux_41() { }

// a platform support Max 4 version segments
[SupportedOSPlatform("Android4.8.1.2.3")|}] // The '4.8.1.2.3' is not a valid version. Please provide an input having 2 to 4 numbers separated with a dot.
public void SupportedOSPlatformAndroid5Parts() { }

// a platform support Max 3 version segments
[UnsupportedOSPlatform("Ios14.1.2.3")] // The '14.1.2.3' is not a valid version. Please provide an input having 2 to 3 numbers separated with a dot.
public void UnsupportedOSPlatformIos4PartsInvalid() { }

**Comment by jeffhandley** [comment]
I suggest the following messages. Each also shows using lower-case for the platform names in the messages to recommend that practice.

```csharp
// versions are not supported case:
[SupportedOSPlatform("Linux4.1")|}] // Version '4.1' is not valid for the platform 'linux'. Do not use versions for this platform.
public void SupportedLinux_41() { }

// a platform support Max 4 version segments
[SupportedOSPlatform("Android4.8.1.2.3")|}] // Version '4.8.1.2.3' is not valid for the platform 'android'. Use a version with 2-4 parts for this platform.
public void SupportedOSPlatformAndroid5Parts() { }

// a platform support Max 3 version segments
[UnsupportedOSPlatform("Ios14.1.2.3")] // Version '14.1.2.3' is not valid for the platform 'ios'. Use a version with 2-3 parts for this platform.
public void UnsupportedOSPlatformIos4PartsInvalid() { }

// a platform supporting Max 2 version segments
[UnsupportedOSPlatform("foo1.2.3"] // Version '1.2.3' is not valid for the platform 'foo'. Use a version with 2 parts for this platform.
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,241 @@
...
+namespace Microsoft.NetCore.Analyzers.InteropServices
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class UseValidPlatformString : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1418";
+        private static readonly ImmutableArray<SymbolKind> s_symbols = ImmutableArray.Create(SymbolKind.NamedType, SymbolKind.Method, SymbolKind.Property, SymbolKind.Field, SymbolKind.Event);
+        private static readonly ImmutableArray<string> methodNames = ImmutableArray.Create("IsOSPlatform", "IsOSPlatformVersionAtLeast");
+        private const string IsPrefix = "Is";
+        private const string OptionalSuffix = "VersionAtLeast";
```
jeffhandley, line 23: Nit:
```suggestion
        private const string VersionSuffix = "VersionAtLeast";
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,241 @@
...
+                foreach (var platform in msBuildPlatforms)
+                {
+                    if (!knownPlatforms.ContainsKey(platform))
+                    {
+                        knownPlatforms.Add(platform, 4); // Default version count is 4
+                    }
+                }
+            }
+
+            static int ExtractPlatformAndVersionCount(IMethodSymbol method, out string platformName)
```
jeffhandley, line 116: I suggest having the return type be a `Tuple<string platformName, int versionParts>`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,19 @@
...
+  </data>
+  <data name="UseValidPlatformStringTitle" xml:space="preserve">
+    <value>Use valid platform string</value>
+  </data>
+  <data name="UseValidPlatformStringUnknownPlatform" xml:space="preserve">
+    <value>The platform '{0}' is not a known platform name</value>
+  </data>
+  <data name="UseValidPlatformStringInvalidVersion" xml:space="preserve">
+    <value>The '{0}' is not a valid version. Please provide an input having 2 to {1} numbers separated with a dot.</value>
+  </data>
```
jeffhandley, line 1533: You'll want a message for when there are exactly 2 versions so that it doesn't show a range of "2 to 2"

jeffhandley (approved):
I have a couple suggestions and I also commented with suggested messages, but otherwise looks good to me. Once my feedback is addressed, I approve without needing to re-review.

buyaa-n:
> I suggest the following messages. Each also shows using lower-case for the platform names in the messages to recommend that practice.

Thanks, I like the messages except the lower casing part, isn't it better to report back the platform names as it is provided by the user for expressing that there is nothing wrong with the platform name itself?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: For the existing `OperatingSystem.Is<Platform>VersionAtLeast(...)` APIs there is no variant having 2 version parts, i am not sure if we will have a platform with a max of 2 version parts in the future, do you think that is possible for the future platforms?

jeffhandley: I'm not sure, but I'd rather not have to worry about updating this if one should emerge.

jeffhandley:
> Thanks, I like the messages except the lower casing part, isn't it better to report back the platform names as it is provided by the user for expressing that there is nothing wrong with the platform name itself?

Fair enough; thanks for considering it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: I'm gonna use a resource string `Version '{0}' is not valid for platform '{1}'. Use a version with 2{2} parts for this platform.` where `3` or `4` part will be added with dash `-`, otherwise empty string which can express exact 2 versions, hope there is no localization issue with that cc @cristianosuzuki77

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+        }
+
+        private static void AnalyzeAttribute(Action<Diagnostic> reportDiagnostic, AttributeData attributeData, PooledDictionary<string, int> knownPlatforms, CancellationToken token)
+        {
+            var constructorArguments = attributeData.ConstructorArguments;
+
+            if (constructorArguments.Length == 1)
+            {
+                if (constructorArguments[0].Value is string value)
```
mavasani, line 165: This will throw if the argument kind is array: http://sourceroslyn.io/#Microsoft.CodeAnalysis/Symbols/TypedConstant.cs,79

I would suggest adding a `.Kind != TypedConstantKind.Array` before accessing `.Value`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    AnalyzeStringParameter(reportDiagnostic, attributeData.ApplicationSyntaxReference.GetSyntax(token), knownPlatforms, value);
```
mavasani, line 167: `ApplicationSyntaxReference` can be null for attributes inherited from base types that are in metadata. I'd suggest adding `attributeData.ApplicationSyntaxReference != null` check at the start of the method.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,20 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseValidPlatformStringDescription" xml:space="preserve">
+    <value>Platform compatibility analyzer would not work properly if invalid platform name or version used.</value>
```
```suggestion
    <value>Platform compatibility analyzer requires a valid platform name and version.</value>
```

mavasani (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -0,0 +1,241 @@
...
+
+            static (string platformName, int versionPartsCount) ExtractPlatformAndVersionCount(IMethodSymbol method)
+            {
+                var name = method.Name;
+                if (name.EndsWith(VersionSuffix, StringComparison.Ordinal))
+                {
+                    return (name[2..(name.Length - VersionSuffix.Length)], method.Parameters.Length);
+                }
+
+                return (name[2..], 0);
```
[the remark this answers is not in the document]
buyaa-n: Yes all method names will always start with `Is` prefix, it is called after validating the method name using below `NameAndParametersValid(IMethodSymbol method)` which is making sure that the methods starts with `Is` prefix

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
buyaa-n: This is called only if the attribute were `SupportedOSPlatform` or `UnsupportedOSPlatform` attribute, which only accepts a string argument, if anyone calls them with an array type there will be a compiler error, do you think that check is still needed?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
buyaa-n: Good point, i guess the API nullability was not annotated yet, updating

PR #4846, opened 2021-02-13
Disable generated code analysis for CA-1416 

buyaa-n:
As discussed with @terrajobst and @jeffhandley disabling it in 5.0. We will make it configurable by the user within 6.0 when `AnalysisContext.AnalyzerOptions.AnalyzerConfigOptionsProvider.GlobalOptions` became available in the repo

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -125,7 +125,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         public override void Initialize(AnalysisContext context)
         {
             context.EnableConcurrentExecution();
-            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
             context.RegisterCompilationStartAction(context =>
             {


jeffhandley (approved):

terrajobst (approved):
This makes sense to me. @mavasani @sharwell, any thoughts?

sharwell:
This seems like a valid warning to report in generated code. Is there a reason we need to disable it?

buyaa-n:
> This seems like a valid warning to report in generated code. Is there a reason we need to disable it?

Yes, but this might not necessary in most cases, especially for users having huge generated files this could cause slowness in the build, we want to make it off by default and user-configurable instead

Youssef1313:
I agree with @sharwell here.

- Warnings are valid and can be helpful for generated code.
- This breaks developers' experience who are already aware this is reported for generated code.

Have this shown any clear performance impact in practice?

sharwell:
@Youssef1313 We had a case where this analyzer was _suspected_ to be causing a performance problem (which led to investigating it), but measurement later ended up showing that this analyzer wasn't even registering as taking any significant time at all.

mavasani:
@sharwell This PR is unrelated to the performance issue you are citing. I have clarified with @buyaa-n offline that the PR is purely driven by a functional design decision that this diagnostic is unlikely to be useful for any customers in generated code. This aspect will be made configurable in 6.0 branch for customers who want to see these diagnostics in generated code.

mavasani (approved):

buyaa-n:
> @sharwell This PR is unrelated to the performance issue you are citing. I have clarified with @buyaa-n offline that the PR is purely driven by a functional design decision that this diagnostic is unlikely to be useful for any customers in generated code. This aspect will be made configurable in 6.0 branch for customers who want to see these diagnostics in generated code.

Right, thanks @mavasani for clarification

I did not mean the analyzer is not useful for generated code, but for me generated code is rarely gets updated/changed, so running the analyzer for each build would be redundant, better to be enabled when needed and off by default.

PR #4862, opened 2021-02-17
Faster build

sharwell:
Disable rules which cause severe build-time overhead.

### DIFF
--- a/eng/Analyzers_ShippingRules.ruleset
+++ b/eng/Analyzers_ShippingRules.ruleset
@@ -49,5 +49,6 @@
     <Rule Id="CA5390" Action="None" />
     <Rule Id="CA5403" Action="None" />
     <Rule Id="CA1508" Action="None" />
+    <Rule Id="CA2000" Action="None" />
   </Rules>
 </RuleSet>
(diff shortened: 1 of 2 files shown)


mavasani (commented):
Can you please share more details? We try to run all our analyzers for dogfooding them, and I think that is more critical than build time performance gain.

sharwell:
A rebuild takes 82 seconds prior to this change if VS is not open, and 34 seconds after this change, so it's approximately 2.4x faster. We can enable these rules again one we eliminate the _literal_ 100GiB allocation overhead.

sharwell:
@mavasani Updating Microsoft.CodeAnalysis.NetAnalyzers allows all but two of the analyzers to be enabled again without the performance penalty.

eng/Analyzers_ShippingRules.ruleset
```diff
@@ -32,4 +32,8 @@
     <!-- TODO: Remove this exclusion and fix related issues, tracked by https://github.com/dotnet/roslyn-analyzers/issues/3915 -->
     <Rule Id="RS0041" Action="None" />
   </Rules>
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.NetAnalyzers">
+    <Rule Id="CA1508" Action="None" />
```
mavasani, line 36: Can you please file work items to enable them?

mavasani (approved):

eng/Analyzers_ShippingRules.ruleset
sharwell: Filed #4914 and #4915

mavasani (approved):

PR #4863, opened 2021-02-17
Use System.HashCode for hash code calculations

sharwell:
* Port `System.HashCode` to `Analyzer.Utilities.RoslynHashCode`
* Rewrite expensive hash code calculations to use `RoslynHashCode` instead

Fixes AB#1274151


mavasani (approved):

mavasani:
@dotpaul Do you prefer this change be cherry-picked or targeted to 2.9.x branch? Also, would you be able to suggested some repos that you run the analyzers on for benchmarking?

dotpaul:
2.9.x if it's easy. I'll mail you the repo I've been using.

sharwell:
@dotpaul this is now retargeted to 2.9.x. I have a separate branch locally to aid in the eventual merge to master. Now we just need a good way to test before/after this change.

sharwell:
Two passes of `DisposeObjectsBeforeLosingScope` on Roslyn.sln before this change:

```
Found 2495 diagnostics in 117854ms (81067498528 bytes allocated)
Execution time (ms):            2246508.1989
Execution time (ms):            425694.1512

Found 2495 diagnostics in 112558ms (80943298256 bytes allocated)
Execution time (ms):            1981944.213
Execution time (ms):            561009.2706
```

Two passes after this change:

```
Found 2495 diagnostics in 108546ms (68307329032 bytes allocated)
Execution time (ms):            2280859.5691
Execution time (ms):            387885.7823

Found 2495 diagnostics in 110265ms (68199640448 bytes allocated)
Execution time (ms):            2070796.6508
Execution time (ms):            540579.0154
```

The execution times seem to be a wash, but allocations are reduced by about 13GB.

src/Utilities/Compiler/HashUtilities.cs
```diff
@@ -1,63 +1,76 @@
...
         }
 
-        internal static int Combine<T>(ImmutableHashSet<T> set) => Combine(set, 0);
-        internal static int Combine<T>(ImmutableHashSet<T> set, int currentKey)
-            => Combine(set.Select(element => element.GetHashCodeOrDefault()).Order(),
-                       set.Count,
-                       currentKey);
+        internal static void Combine<T>(ImmutableHashSet<T> set, ref RoslynHashCode hashCode)
+        {
+            foreach (var element in set)
```
dotpaul, line 54: Is enumerating a hash set (and dictionary) with the same elements always going to have a stable order? We used to sort the elements' hash codes before combining.

sharwell: `ImmutableHashSet<T>` internally sorts elements by hash code. This is not the same as other collections but works for what we need here.

stephentoub:
> but allocations are reduced by about 13GB

I've not looked at the whole change, but I'm curious how changing the hash code calculations resulted in such an allocation savings...?

sharwell:
> I've not looked at the whole change, but I'm curious how changing the hash code calculations resulted in such an allocation savings...

Previously, the hash codes of individual fields were collected into an array, and then the final hash code was calculated over the array. Now we pass the `HashCode` instance directly and apply the updates to it from each field without the intermediate array.

sharwell:
@dotpaul Did you want to see anything else before merging this?

dotpaul (approved):
LGTM :ship:

PR #4868, opened 2021-02-17
CA2259: Use implicitly assignable foreach variable type

maxkoshevoi:
Fixes: https://github.com/dotnet/roslyn/issues/54712

image

- There should be no false positives since it's reported only when `IsAssignableTo` is false.
- This analyzer reports on unobvious explicit conversion done under the hood by `foreach` loop. This conversion will likely result in runtime exception, but it's not guaranteed.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseAssignableTypeForForeachVariableDescription" xml:space="preserve">
+    <value>There is no implicit conversion between collection element and iteratoi variable, this will result in runtime exception.</value>
```
Evangelink, line 1523: Small typo. Note that you will have to regenerate the resources after applying this fix.

```suggestion
    <value>There is no implicit conversion between collection element and iterator variable, this will result in runtime exception.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+            context.RegisterCompilationStartAction(compilationContext =>
+                compilationContext.RegisterOperationAction(context =>
+                {
+                    AnalyzeLoop(context, context.Compilation);
+                },
+                OperationKind.Loop)
+            );
```
Evangelink, line 39: Cosmetic suggestion:

```suggestion
                compilationContext.RegisterOperationAction(
                	context => AnalyzeLoop(context, context.Compilation),
                	OperationKind.Loop)
            );
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            ITypeSymbol collectionElementType = loopInfo.ElementType;
+            ITypeSymbol variableType = ((IVariableDeclaratorOperation)loop.LoopControlVariable).Symbol.Type;
+
+            if (collectionElementType.IsAssignableTo(variableType, compilation))
+            {
+                return;
+            }
+
+            context.ReportDiagnostic(Diagnostic.Create(Rule, syntax.ForEachKeyword.GetLocation(), collectionElementType.Name, variableType.Name));
```
Evangelink, line 39:
```suggestion
            if (collectionElementType.IsAssignableTo(variableType, compilation))
            {
                return;
            }

            context.ReportDiagnostic(syntax.ForEachKeyword.CreateDiagnostic(Rule, collectionElementType.Name, variableType.Name));
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.NetCore.Analyzers.Runtime;
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpUseAssignableTypeForForeachVariable : UseAssignableTypeForForeachVariable
+    {
+        protected override void AnalyzeLoop(OperationAnalysisContext context, Compilation compilation)
```
Evangelink, line 16: Open question: Just wondering if it would be worth transferring some of the logic in the base class and having the subclass only proving some info (the syntax, the `GetForEachStatementInfo`).

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,42 @@
...
+            var loopInfo = loop.SemanticModel.GetForEachStatementInfo(syntax);
+            if (!loopInfo.CurrentConversion.Exists ||
+                !loopInfo.CurrentConversion.IsImplicit ||
+                !loopInfo.CurrentConversion.IsIdentity)
+            {
+                return;
+            }
+
+            ITypeSymbol collectionElementType = loopInfo.ElementType;
+            ITypeSymbol variableType = ((IVariableDeclaratorOperation)loop.LoopControlVariable).Symbol.Type;
```
Evangelink, line 32: Are we sure this cast is safe? Also isn't it possible to end-up with a null symbol which would result in a NRE on the `.Type` call?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
...
+  </data>
+  <data name="UseAssignableTypeForForeachVariableMessage" xml:space="preserve">
+    <value>Cannot convert type '{0}' to '{1}'</value>
+  </data>
+  <data name="UseAssignableTypeForForeachVariableTitle" xml:space="preserve">
+    <value>Use valid foreach variable type</value>
```
Evangelink, line 1529: What about osmething like: `No implicit conversion between these types`? Or `Cannot convert these types`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseAssignableTypeForForeachVariableDescription" xml:space="preserve">
+    <value>There is no implicit conversion between collection element and iteratoi variable, this will result in runtime exception.</value>
+  </data>
+  <data name="UseAssignableTypeForForeachVariableMessage" xml:space="preserve">
+    <value>Cannot convert type '{0}' to '{1}'</value>
```
Evangelink, line 1526: What about: `Cannot find any implicit conversion between type '{0}' and '{1}'`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAssignableTypeForForeachVariableTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAssignableTypeForForeachVariableMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAssignableTypeForForeachVariableDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Usage,
+            RuleLevel.BuildWarning,
```
Evangelink, line 22: @mavasani to confirm but I think you should not set it as build warning.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+        }
+    }
+}";
+
+            await VerifyCS.VerifyAnalyzerAsync(test);
+        }
+
+        private static DiagnosticResult GetCSharpResultAt(int line, int column)
+#pragma warning disable RS0030 // Do not used banned APIs
```
Evangelink, line 461: We try to no longer use this syntax but instead either provide a full location with `WithSpan` or by using the markup syntax this helps with having more precise control over how issues are displayed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
```diff
@@ -0,0 +1,466 @@
...
+using System.Collections.Generic;
+
+namespace ConsoleApplication1
+{
+    class Program
+    {   
+        void Main()
+        {
+            var x = new List<A>();
+            foreach (B item in x)
```
Evangelink, line 50: As you have some arguments to provide you need to use the markup syntax with an id

(note that I am not sure where you report the issue so you will have to tweak the location)
```suggestion
            {|#0:foreach (B item in x)|}
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
```diff
@@ -0,0 +1,466 @@
...
+            {
+            }
+        }
+    }
+
+    class A { }
+    class B : A { }
+}";
+
+            await VerifyCS.VerifyAnalyzerAsync(test, GetCSharpResultAt(11, 13).WithArguments("A", "B"));
```
Evangelink, line 60:
```suggestion
            await VerifyCS.VerifyAnalyzerAsync(test, VerifyCS.Diagnostic.WithLocation(0).WithArguments("A", "B"));
```

Evangelink (changes requested):
Please update all tests to use the `WithSpan` or the markup syntax (I have left one exemple).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,44 @@
...
+            description: s_localizableDescription,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```
Youssef1313, line 32: I think we shouldn't warn on generated code.

```suggestion
            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
Youssef1313, line 60: Consider using VerifyCodeFixAsync and pass both the source and fixedSource the same value.

This will ensure that:

- *If* a code fix is added and it's applicable to this case, the tests will fail and be updated with the codefix behavior.
- *If* a code fix is added but it's not applicable to this test, the test will only fail **if** the codefix provides an action that does nothing, and will pass if no action is offered.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
Evangelink: Assuming the implementation is not missing some cases, there will be a runtime exception when calling this loop so it seems reasonable to me to analyze/report.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    public abstract class UseAssignableTypeForForeachVariable : DiagnosticAnalyzer
```
Youssef1313, line 10: I don't know if this is applicable to VB, but noting in case if it's forgotten.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
Youssef1313: @Evangelink Good call!

The cast is **not** safe for a case like `foreach (var (a, b) in x)`.

A test demonstrating this should be added.

*found it in roslyn's tests*:

https://github.com/dotnet/roslyn/blob/51718d0c6a6c3adba33efebf815e68f194b82a39/src/Compilers/CSharp/Test/IOperation/IOperation/IOperationTests_IForEachLoopStatement.cs#L1101-L1102

That's why I love roslyn being open source ❤️

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
mavasani: Agree with @Youssef1313 - only critical security analyzers should warn on generated code.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
Youssef1313, line 32: This is safer to avoid crashes.

```suggestion
            ITypeSymbol? variableType = (loop.LoopControlVariable as IVariableDeclaratorOperation)?.Symbol?.Type;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
Evangelink, line 1526: Looking at the tests, it seems that we can report on cases where it's possible the runtime would work so we should probably be less "strict" here.

```csharp
using System;

public class A {}
public class B : A {}
public class Program
{
    public static void Main(string[] args)
    {
        var list = new List<A> { new B(), new B() };
        //Your code goes here
        
        foreach (B item in list) // We report here but in practice this code doesn't fail
            Console.WriteLine(item);
    }
}
```

maxkoshevoi: That's correct. If explicit conversion exists, but implicit one doesn't analyzer will also be reported.
We cannot guarantee that explicit conversion will succeed though. 

If explicit conversion is untended it's better for user to perform it manually.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
maxkoshevoi: This is C# only analyzer for now.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
maxkoshevoi: `ForEachStatementInfo` is a C# class and an entire `AnalyzeLoop` method depends on it

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
maxkoshevoi: Tuples, right! I couldn't figure out what else could be there except `IVariableDeclaratorOperation`. Thank you =)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
maxkoshevoi: Updated to `Cannot implicitly convert type 'type1' to 'type2'.` like in [CS0266](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0266)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
maxkoshevoi: Updated to `Cannot implicitly convert type 'type1' to 'type2'.` like in [CS0266](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0266)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
maxkoshevoi: This analyzer only checks foreach. For other places there are already [CS0266](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0266) and [CS0030](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0030). Any they usually work, just not here.

Anyway, I think analyzer name should include `foreach` somewhere.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
maxkoshevoi: Thanks for explanation, refactored the tests

maxkoshevoi: Thanks for suggestion. Replaced `VerifyAnalyzerAsync` with `VerifyCodeFixAsync`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
maxkoshevoi: What about `Use implicitly assignable foreach variable type`? It conveys the meaning pretty well.

maxkoshevoi:
I think re-requesting a review wasn't the right move. Could someone please add `dotnet/roslyn-analysis` back to reviewers list?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
Evangelink: You are right, that will force adding too many methods on base.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,47 @@
...
+                return;
+            }
+
+            CommonConversion conversion = compilation.ClassifyCommonConversion(collectionElementType, variableType);
+            if (!conversion.Exists || conversion.IsImplicit)
+            {
+                return;
+            }
+
+            context.ReportDiagnostic(syntax.ForEachKeyword.CreateDiagnostic(Rule, collectionElementType.Name, variableType!.Name));
```
Evangelink, line 44: I am surprised that you need the bang after `variableType`. Could you test if you can remove it? If not, maybe using `RoslynDebug.Assert(variableType != null);` would be better.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
+                compilationContext.RegisterOperationAction(
+                    context => AnalyzeLoop(context, context.Compilation),
+                    OperationKind.Loop)
+            );
```
Evangelink, line 37: Sorry I missed this one on the first review, you don't need the `RegisterCompilationStartAction`:

```suggestion
            context.RegisterOperationAction(
                context => AnalyzeLoop(context, context.Compilation),
                OperationKind.Loop);
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
```diff
@@ -0,0 +1,533 @@
...
+using System.Collections.Generic;
+
+namespace ConsoleApplication1
+{
+    class Program
+    {   
+        void Main()
+        {
+            var x = new List<A>();
+            {|#0:foreach|} (B item in x)
```
Evangelink, line 50: Thank you for the changes, that's easier to review where is the issue reported :)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
```diff
@@ -0,0 +1,533 @@
...
+        {
+            var x = new List<string>();
+            foreach (string item in x)
+            {
+            }
+        }
+    }
+}";
+
+            await VerifyCS.VerifyCodeFixAsync(test, test);
```
Evangelink, line 34: You have no code fix so you could use the `VerifyAnalyzerAsync` method instead but I am fine with keeping the current code.

Evangelink (commented):
Overall LGTM, @mavasani to decide on the warning level.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariableTests.cs
Youssef1313: @Evangelink Using VerifyCodeFixAsync is better in case there is a codefix in the future.

Evangelink: That's a good remark, maybe we could (in a separate PR) ban the `VerifyAnalyzerAsync` and replace all calls. I am not sure which quantity of work that represents.

mavasani (changes requested):
New analyzer needs to target https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx-preview2

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
mavasani: Yes, this should not be a build warning. Only critical API analyzers which are approved by .NET runtime team to be build warnings are allowed to be warnings. This would also be a blocking change.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
maxkoshevoi: Bang was needed earlier, didn't notice it's still there. Strange that VS didn't tell that I don't need it anymore.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
maxkoshevoi: Thank you! Didn't noticed `compilation` was present in `context`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
maxkoshevoi: @mavasani Should it be `IdeSuggestion` then? But it says that `IdeSuggestion` is `not something to be enforced in CI`.
This analyzer catches unobvious explicit cast that'll likely result in runtime error. Could you please ping someone from .NET runtime team to verify that it shouldn't be enforced.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
Evangelink: I don't think there is any analyzer for that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
mavasani: > But it says that IdeSuggestion is not something to be enforced in CI. This analyzer catches unobvious explicit cast that'll likely result in runtime error. 

@maxkoshevoi IDE Suggestion seems fine. The doc for that field states it is not to be enforced in CI _by default_, end users can bump it to a warning/error if they feel appropriate.

> This analyzer catches unobvious explicit cast that'll likely result in runtime error. 

We only want issues that are critical .NET API issues to be warnings by default. To give more context, if this is a warning, then the next .NET SDK release which contains this analyzer will start breaking everyone's code bases when they just move to a newer SDK version - it is essentially a build breaking change. If you still want to push this as a warning by default, please file a separate issue on dotnet/runtime repo and once you have an approval, we can do the necessary change. Until then, please keep this as an IDE suggestion or Hidden/Bulk configurable diagnostic.

maxkoshevoi: @mavasani Changed it to IDE Suggestion for now, and created an issue https://github.com/dotnet/runtime/issues/48529 to discuss if it's appropriate to set it to Warning.

Wouldn't build break only if users have `TreatWarningsAsErrors` set?

mavasani: > Wouldn't build break only if users have TreatWarningsAsErrors set?

Majority of enterprise customers have warnaserror. All warnings shipping in the .NET SDK analyzers are considered build breaking. For example, note that all the CA diagnostics with default severity warning are listed under the build breaking changes in .NET 5.0: https://docs.microsoft.com/dotnet/core/compatibility/5.0#code-analysis

maxkoshevoi: Didn't know that. Thanks for the explanation

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
Youssef1313: @Evangelink @mavasani Do we want a tracking issue to follow-up with VB?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1510,4 +1510,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseAssignableTypeForForeachVariableDescription" xml:space="preserve">
+    <value>There is no implicit conversion between collection element and iterator variable, this will likely result in runtime exception.</value>
```
Youssef1313, line 1514: "will likely" seems stronger than it should be.

Youssef1313 (approved):
Looks good :tada:

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
maxkoshevoi: Changed it to "can"

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAssignableTypeForForeachVariable.cs
Evangelink: See #4875

Evangelink (approved):
Thank you for your contribution @maxkoshevoi

stephentoub:
I appreciate the efforts here, but I'm skeptical that this is a rule we should be adding.
https://github.com/dotnet/runtime/issues/48529#issuecomment-782724984

Evangelink:
Oops sorry didn't read through all the ticket thread. Let's block the PR until some decision is made on the direction.

maxkoshevoi:
> I appreciate the efforts here, but I'm skeptical that this is a rule we should be adding.
> [dotnet/runtime#48529 (comment)](https://github.com/dotnet/runtime/issues/48529#issuecomment-782724984)

Excluded collections with `object` element type from analysis, so it wouldn't report on non-generic collection like `ArrayList` and generic collection that are intentionally created with `object` type, since it's obvious that cast will be performed there, and most likely it's intended (better to be false negative than false positive on this one).

Should not complain that much anymore.

stephentoub:
> Should not complain that much anymore.

Can you find any case in any dotnet/* repo for example where this actually finds a real bug?

stephentoub:
> Should not complain that much anymore.

Ran it on dotnet/runtime: 868 Warning(s).

maxkoshevoi:
> Ran it on dotnet/runtime: 868 Warning(s).

I'm trying to do that right now via standalone tool, but if fails to load around 1/3 of solutions with this error: `Cannot modify an evaluated object originating in an imported file`. 
Trying 5.0.0 branch now. 

From warnings that I've got, there are lot of duplicated ones (because of multitargeting), so it's actually around 5 times less warnings if we count only unique ones. 

- Could you please share how you run the analysis. Are you also creating standalone tool using template from VS?
- Added a commit that excludes all types that don't implement `IEnumerable<T>`. This should remove most of the warngs that I saw. I'm doing another run with latest version and will investigate what's left.

stephentoub:
> Could you please share how you run the analysis.

Built the dlls from your PR, overwrote the corresponding ones in my .nuget cache, did a runtime build with `.\build.cmd clr+libs -rc Release -warnaserror 0`

maxkoshevoi:
> Ran it on dotnet/runtime: 868 Warning(s)

Run it on `release/5.0.0` branch of `dotnet/runtime`. ([using standalone tool](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxASwDZICYgNQA+AAgEwCMAsAFBEAMABEWQCwDc1djZAdAMID2GDHADGMNPwB2UbgEkwYAK4wIwYexoMmcgPIbO2gDJpJAR31aeAFQAWCOBGwmA5ha4BWNwEFJEDAE8AL0Q2DgYAWTQRBH4ofgAzGG4AIUVMbG5DfhEIGH4EN0jo2ISkgWw4Hz9/KDQoQqiYuMS+fgqqgNqZABE0CGdJWPERerD6IqbS1vbfTrrucIBlVPSNDjIANkYSLgB2agBvanoTri2iAA5GLfCIEwAKAEpj06OqU4/6ADcIBHoY/gwXr2MT5fwABVyNnoAF5GEQAES8EAAHQAqlBEFAUZEUXFFAgRHAUfYAA6xFHYQGSOAwAC0CEUknEYGJUEJCI0n1O8iUKjUlQQCAg/gAPL1+oNYFEOkFEAA+ehgfwdNDBBBQWH0XnKVTCLxCkV8ey5ODivoDIYy2ZyhDy+40gDu9Fl6uYrsQTyQ9CdLptbo9CAA+joMNgno81u9ufQAPSxl0weBgUkweh5eiYtMwGxwb5YiSSegJcbLNJh7gvbk/P5fOqKPyLGCKJz8WTSFSSImauFLFZhrI5PIIbgARUUiH8ADV643m6327AIF24FAntwrPwDcL/E8uTGa/QTEuV1q61AGxgmy2JIvO93MnBJM4c7C4ZRozHTvHtfF07n7CPTUpACYsaSPDtlyJYs/z7ctsDA/86iVCARBsEw4G9LN/1yehoH/PMpDzDNFExStPy/egAH5vlnK951vSCVygABtWgAF0q0on8AAll2wYRMzgYQxBcAj8w1QsYNLfsEP8fhFHoR1l2zfh6FIuByMo04QHoRZhNEGAZwvOcbzbJiiXuc9L2vBcLNXSNqC4j4mAATnuAASBEMTEuD0jwtMAHIDmPe9NL8sNIRzABfQL0zUjB+EcehSRiAArQyZARRyqGc78EwAOR0KwAFFdOSPML0AjN7GcOp4D+ZcIJPaDHTQV8czzCLsEHXJ8noPKTh/blgDgeJ8jzHIhF8st0gAdXyABrKBSVQzTeBNeAnkGuME25QEALazFvW6haEGW1bWqkQK03CEqADF6BEfgU1iTSdu63rh24AAlOA6tgRA7yguB7lCkGcp2zhD0dJaVrWrVTrhy71s20Gcq/H9wQQEw01ZKAoH6PNxr+M6LrW+67mEBC4C+J9VPoXMMFJegnElWIxNS/gMrEehEuS+IqYJVctJjWHzvholuDJyW4EpzA4AQvA4XufhvTgR5YUVNz7k0iVLWlEQFlXQnnA1qMv2MWBRSYWhFX5ipsEWQRlELHsfTgR09x2q2YBtsg7foQWFadl3xCkd2nW93KKI+X2bY2UUdX5fVDTFfWpWGeVs/TaBlq1KOMZjXgQaEXJC03Ran2dgloJyFcy/DyQa8JPM4UL/duRJhw0PoKzfkzMPCyi6ETHoYFDLBbgAHFaXuhW1wBIE0BBYcIShb0EQAKm4KAMEkBFvX0340J0VNC24LwhAn0EcYc55Y9eHaXLIdyvJRSQskcMS4gwV2pHoMFX+/9JAj1itlC22ldqJhUL3Lmzh7AE3+HAckCAGqDyUnmVKuMUrpUynhTUnV/B4UAg7RWottLOz/k3Qe1DCyd0ojABA/hn5P0ftpYBNC4REFckpZGa1uBn2rkPKQXgoD+C7PcThw8N4e2dAIaQgg4DY34Agk2f1UENQjAwr80VWEnCHL3e4JUAAeRJz4ALgCYh+UD6BvFsR8BRv9NL3QmgghSkhsACESn8OETilHePyL9RWOioE6y8vdLwshDAlW6HzJKThny0JAYAg40ipBgMCrpA4VjjYEyJtFCB+jPj+OEMErMgSEDRwcacYO1MqEgJkF4bA4Z0mgKhEXGpz1mQmAnKE7kej2ElIjko7grjaoxCZF4wQ/U/EjOEJUme9gnz9M+OEhE89JB1FzAhB2P8RFFiAQczJRShmOPmZpP6FSZlVM6ZRMhoc6ER0vi0qRxyOmQO0j+KwOhug6F0t0NSy5qhdEQp1XB3N8Fj3BQ85JTdikqCgMtF52AfrLjNm8p5khuAqJ5jAJpUBITCiEMJdc+kRIwHuJcCFeKtbFPsTU+gAgUyYHLgA56LKMBsqLNw3hXM8Uz1pMy0krKm5iIkSIapjLhWisLHNdqNhAyag5SKrlTd5U5iVVqFVsqpDSwVUq+4ypVTqigJDM5B4B5OCgAbYYWoeFPReqq7lGrFX+ixIKmAgYM5WhGOKyRxrZhqixN6euRJG6FhblLSuT47m2KILsVmdRbVRFWR8aKjw42nEGbYqGvCiAbGlrmSQV8MD3ERctLNJwfWGxYuxPCQgyT5HxVqCtMhyWGVuJIXcCEYSKgyFcxQGAYCPA3FuNO0coav08giBpTdNTAoCMEBC9wDgPLnW7VoTIYDRUzIoEQ3YoCxjXQkxWG7nkCG3fQPAQcqZnoOTIS9zJd15BUBgR4pydqHncYoUkistR+AwE2tBMhp6TNJMkHtWtWZyGwFWoOE1ULQn7n8H9LMx5ocVjYmMDLtKHmwL+jAURy7PinH4Cc7s0Nju3CKKVeGB4/D/quLUBHSREaHC4MjTGej1RMGISD9xnDQecJkbI3LuBRrgIsVakhR2bho7uc18bp3v0kLIXAdiqMAGk4D+EKfBl+b8kQeNQHYxjFHHzPhikm6UXY0yrtY+xkjzguMWcME+F8NgX2Aj8B+gzMZiloD/IJ8D3AdPEJhHCQM7p3XBlDBkGtwx1PYe0rhhx3TxCSD6cUwZxTu5Ib7vhi0mcogQW+ORhy9LikfEPIlDjAC4RsxTUbL6F9Z4wGMDSaTy46MOPWSe+r2KwF2Lq+J7rkh9NpuzTtXLQzSmXIcD1DCvXZsfGwT8eAZx6DGRsgxcyLU8wdrEDt0ydkDv3BO/RMywMVx1toiZK7Z2wpmq4mlwz9wEThCHeINjeZwZCGArBWaYZWa0kMv+370BKoGV5kREApyvwkz7jgtAWpaCsCPPQUU93dvXfsjIdzVmbAY7QHgPAKXPhvZjOsm70EQrXvoGQSbxT1kfAKhAVk2TrKncYgd1iaB2LcHZ6yZnFqTis9OFOAsUgud0Vsrz57LEBfcCl5JKQovlNGY+N1egI9ZcPfl/txXyvupgIRzGVbYt0KCXLYyDWO0qfVgHhpBAVzyTSDbkyi5wTHCdfRlNk4QXkfMg3MwolmJ7gu7dxHTCxZlAQTTODFcBVFBgFGggTWAAyTP1XThJ6JCntPiB6CKgYNn3Pgf7KF/T1juE3PHsK5BgTjzOYKdfkd5RBNOOedG6b0rqvqea90kZ+xAP9Ac1hOnQidspJ4+DDTKhcx8BsDeiYcQ/odxsXm4GVxS3Jx1umm2F7xRwgVFqIJho5txfdKyDP4gqAopcWGS/tgO/Jt5SvahgAZkYMwegl+0F7gn8xAX838kEHYwCXshkO8asB5+VDJegVouUItdcoRPV55hBhdQYICYhz8ZAMDlEPlAtgscDVF78NxfgzYYB7phRWRxZFp6AABCduIdd9KrMXT4eAsQRAtjEUa9OELyPuE9RwSAighAKgmgjnT2JaDNTkHLJyDg/rUgvAwRP9YUJuJAOkMgdwXdYQ1/XA8gkqLlUkTEbAKwNAVkTQ7Q8AFEFEEAAmWw7geIZwlw+IXQrg5eJAkUDXC3XfBQlyH/Jgc4MgL/UUKwSTRURAzLfjMUcIhSVub0KwcLe0HMZCJgUIuI2uOARUfEBIxgdwMIyTRI5I+gKufwI7YcCnDvPiKAGwfSP2JI3THIuAJ8cLSOT2XrD4fLIxQ8AyVkZkMrXIokNvNhWxIPKRFoyQNolFe4Moio/IXWYQfokdTNdgxlLgMgRgRNPo+mMfCfHfCiQZaKIAA==))

```
Solutions analyzed (237 success/243 total)

Id: CA2250_Old (all collections are checked)
Count: 933 distinct (1341 total)

Id: CA2250 (only generic non-object collections are checked)
Count: 0 distinct (0 total)
```

Looks like explicit cast in `foreach` in production is used only for non-generic collections. I'll run it on some more repositories to see if it finds anything.

stephentoub:
Closing per discussion in dotnet/runtime#48529. Thanks!

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            if (context.Operation is not IForEachLoopOperation loop ||
+                loop.Syntax is not CommonForEachStatementSyntax syntax)
+            {
+                return;
+            }
+
+            ForEachStatementInfo loopInfo = loop.SemanticModel.GetForEachStatementInfo(syntax);
+            if (!loopInfo.CurrentConversion.Exists ||
+                !loopInfo.CurrentConversion.IsImplicit ||
```
RikkiGibson, line 26: When is `loopInfo.CurrentConversion.IsImplicit` false?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
```diff
@@ -0,0 +1,54 @@
...
+                return;
+            }
+
+            CommonConversion conversion = context.Compilation.ClassifyCommonConversion(collectionElementType, variableType);
+            if (!conversion.Exists || conversion.IsImplicit)
+            {
+                return;
+            }
+
+            context.ReportDiagnostic(syntax.ForEachKeyword.CreateDiagnostic(Rule, collectionElementType.Name, variableType.Name));
```
RikkiGibson, line 51: It feels like it would make sense to put the diagnostic on the loop variable type syntax, since that's what the diagnostic is suggesting the user should change.

maxkoshevoi: There're similar rules that generate errors when explicit case would definitely fail (like `foreach (string item in new[] { 1 })`). Those rules locate diagnostic on `ForEachKeyword`. That's why I also put it here.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpUseAssignableTypeForForeachVariable.cs
maxkoshevoi: I think, I'd copied this condition from somewhere, and this line may be removed

maxkoshevoi:
@stephentoub Could you please reopen this PR, sine the issue was reopened? It can be converted to draft, for example.
This way, it can be retargeted to `main`, and I could continue to work on this analyzer (if some other issues will arise)

maxkoshevoi:
Should this be closed since https://github.com/dotnet/roslyn/pull/60120 is merged (triple hooray for that)?

RikkiGibson:
I think so.

PR #4878, opened 2021-02-22
CA5360: Show potential invocations to reach potentially dangerous method

### LINKED ISSUE #4809: CA5390: The call stack to the dangerous method should be part of the message
### Analyzer

**Diagnostic ID**: [CA5360](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca5360)

### Describe the improvement

Right now, the message only includes two methods: the method called by deserialization and the dangerous method. It would be better to include the intermediate methods potentially invoked that lead from the first method to the last method.

dotpaul:
Fixes #4809.


nathanstocking (approved):

dotpaul:
Thank you @nathanstocking !

Issue #4772, opened 2021-02-03
AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.'


notnotchris:
### Analyser

AD0001: `Analyzer 'Microsoft.NetCore.Analyzers.Security.UseAutoValidateAntiforgeryToken' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.'`

### Analyser source

**SDK**: [Built-in CA analysers in .NET 5 SDK or later](https://docs.microsoft.com/dotnet/fundamentals/productivity/code-analysis)

**Version**: [SDK 5.0.102](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug

An `ArgumentNullException` occurs, causing an `AD0001` warning, when building a project that registers an asynchronous authorization filter (implements `IAsyncAuthorizationFilter`) that derives from an abstract class and all rules are enabled by default (`<AnalysisMode>AllEnabledByDefault</AnalysisMode>` in the project file).

### Steps to Reproduce

1. Create an abstract asynchronous authorization filter.
2. Create a concrete class that derives from the abstract type.
3. Register the concrete filter in `AddControllers`.
4. Enable all rules by default.
5. Build the solution/project.

### Expected behaviour

Either the project builds successfully without warnings, or the relevant code analysis warning is displayed.

### Actual behaviour

An `AD0001` warning is generated.

## Additional context

Adding the latest [NuGet package](https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers/5.0.3) version (5.0.3) of the analysers makes no difference. The `AD0001` warning is still generated.

The `AD0001` warning does not occur when all rules are not enabled by default (i.e. `<AnalysisMode>AllEnabledByDefault</AnalysisMode>` is omitted from the project file).

A minimum repro is available [here](https://github.com/notnotchris/ad0001-issue).

Youssef1313:
@notnotchris Thanks for the detailed repro steps. The issue happens when the filter used doesn't directly contain "OnAuthorizationAsync" method.

The problematic part of the analyzer implementation is the following:

[...]

PR #4880, opened 2021-02-22
Use SymbolEqualityComparer

dotpaul:
Part of #4772.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Security/UseAutoValidateAntiforgeryToken.cs
```diff
@@ -202,7 +202,7 @@ public override void Initialize(AnalysisContext context)
                                         .FirstOrDefault(
                                             s =>
                                                 s.Name == "OnAuthorizationAsync" &&
-                                                s.ReturnType.Equals(taskTypeSymbol) &&
+                                                s.ReturnType.Equals(taskTypeSymbol, SymbolEqualityComparer.Default) &&
```
sharwell, line 205: 💭 Why is this not:

```suggestion
                                                SymbolEqualityComparer.Default.Equals(s.ReturnType, taskTypeSymbol) &&
```

dotpaul:
@sharwell Lemme know if there's anythiing else

nathanstocking (approved):

dotpaul:
Thank you @sharwell and @nathanstocking !

PR #4890, opened 2021-02-23
Cache hash sets for TaintedDataConfig

sharwell:
Saves over 100GB allocations when analyzing dotnet/roslyn-analyzers.

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataConfig.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataConfig.cs
@@ -26,6 +26,18 @@ namespace Analyzer.Utilities.FlowAnalysis.Analysis.TaintedDataAnalysis
         private static readonly BoundedCacheWithFactory<Compilation, TaintedDataConfig> s_ConfigsByCompilation = new();
 
         /// <summary>
+        /// Caches the results for <see cref="GetSourceInfos(SinkKind)"/>.
+        /// </summary>
+        private static ImmutableDictionary<SinkKind, ImmutableHashSet<SourceInfo>> s_sinkKindToSourceInfo
+            = ImmutableDictionary.Create<SinkKind, ImmutableHashSet<SourceInfo>>();
+
+        /// <summary>
+        /// Caches the results for <see cref="GetSanitizerInfos(SinkKind)"/>.
+        /// </summary>
+        private static ImmutableDictionary<SinkKind, ImmutableHashSet<SanitizerInfo>> s_sinkKindToSanitizerInfo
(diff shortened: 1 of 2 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataConfig.cs
```diff
@@ -194,42 +210,71 @@ private static ImmutableHashSet<SourceInfo> GetSourceInfos(SinkKind sinkKind)
                 case SinkKind.XPath:
                 case SinkKind.Xml:
                 case SinkKind.Xaml:
-                    return WebInputSources.SourceInfos.AddRange(StringTranferSources.SourceInfos);
+                    // All of these use WebInputSources.SourceInfos.AddRange(StringTranferSources.SourceInfos), which is
+                    // the same set as SinkKind.Sql. Delegate the call to SinkKind.Sql to avoid computing separate hash
+                    // sets for identical cases.
+                    sourceInfo = GetSourceInfos(SinkKind.Sql);
```
genlu, line 216: How much this saves given the caching we are doing now? Is it worth the potential maintainability issue in the future?

genlu (approved):
:shipit:

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/TaintedDataAnalysis/TaintedDataConfig.cs
sharwell: It's used as the key for another dictionary for the compilation, so as best I can tell it was always meant to share the value.

genlu: But does keying all those different SinkKind to Sql actually have impact on allocation? Unless the benchmark result shows noticeable difference, I'd rather avoid optimization like this so we don't make the code more complicate (which is already not easy to follow to begin with)

sharwell: Testing this is a bit complicated, but the scale of the observed problems suggests data sharing provides significant value. Is there another way I can improve clarity without reintroducing data duplication?

genlu: I certainly don't want to block this great change for this. But maybe we can get rid of the cache and move the creation of those SourceInfos and SanitizerInfos to be part of XXXSinks static classes (e.g. XmlSinks, SqlSinks, etc.). So in additiona to XXXSinks.SinkInfos, we will also have XXXSink.SourceInfos and XXXSinks.SanitizerInfos.

dotpaul (approved):
:ship:

PR #4892, opened 2021-02-23
Reduce allocations

### LINKED ISSUE #4893: WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName allocates even when already cached
`WellKnownTypeProvider.TryGetOrCreateTypeByMetadataName` creates a capturing delegate every time the method is called, even if the delegate is not used (value was already cached).

sharwell:
Reduce hot path allocations.

Fixes #4893


src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -86,6 +86,18 @@ public static WellKnownTypeProvider GetOrCreate(Compilation compilation)
         public bool TryGetOrCreateTypeByMetadataName(
             string fullTypeName,
             [NotNullWhen(returnValue: true)] out INamedTypeSymbol? namedTypeSymbol)
+        {
+            if (_fullNameToTypeMap.TryGetValue(fullTypeName, out namedTypeSymbol))
+            {
+                return namedTypeSymbol is not null;
+            }
+
+            return TryGetOrCreateTypeByMetadataNameSlow(fullTypeName, out namedTypeSymbol);
```
genlu, line 95: Comment here would be nice, since the new method name doesn't really capture the intent of the change

src/Utilities/FlowAnalysis/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <paramref name="predicate"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of element in the list.</typeparam>
+        /// <typeparam name="TArg">A state argument to pass to <paramref name="predicate"/>.</typeparam>
+        /// <param name="list">The list.</param>
+        /// <param name="predicate">A predicate matching elements to remove from <paramref name="list"/>.</param>
+        /// <param name="argument">An additional state argument to pass to <paramref name="predicate"/>.</param>
+        /// <returns>A collection of elements removed from <paramref name="list"/>, in the order they were removed. If
+        /// no elements were removed, this method returns <see cref="ImmutableArray{T}.Empty"/>.</returns>
+        public static ImmutableArray<T> Extract<T, TArg>(List<T> list, Func<T, TArg, bool> predicate, TArg argument)
```
genlu, line 22: nit: can we make this an extension method?

src/Utilities/Compiler/WellKnownTypeProvider.cs
sharwell: Added `[PerformanceSensitive]`

src/Utilities/FlowAnalysis/Extensions/ListExtensions.cs
sharwell: Done

src/Utilities/Compiler/Analyzer.Utilities.projitems
```diff
@@ -41,6 +41,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)CodeMetrics\CodeMetricsAnalysisContext.cs" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)..\..\..\nuget\PerformanceSensitiveAnalyzers\PerformanceSensitiveAttribute.cs" Link="PerformanceSensitiveAttribute.cs" />
```
genlu, line 44: Hmm, can we actually dogfood the analyzer package instead of linked file like this?

sharwell: I'm not sure since it's part of a shared project. I can try...

sharwell: Updated to reference PerformanceSensitiveAnalyzers instead

src/PublicApiAnalyzers/Core/Analyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.csproj
```diff
@@ -8,6 +8,9 @@
     -->
     <PackageId>*$(MSBuildProjectFile)*</PackageId>
     <MicrosoftCodeAnalysisVersion>1.2.1</MicrosoftCodeAnalysisVersion>
+
+    <!-- Work around PerformanceSensitiveAttribute.cs cannot resolve ValueTask<TResult>. -->
+    <NoWarn>$(NoWarn);CS1574</NoWarn>
```
sharwell, line 13: 📝 Submitted #4895 as a draft to remove all these as soon as this PR is published

src/Utilities/Compiler/Analyzer.Utilities.projitems
genlu: Thanks!

genlu (approved):

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -2691,25 +2691,21 @@ public override TAbstractAnalysisValue Visit(IOperation operation, object? argum
...
-                            PostProcessEscapedArgument(argumentOperation);
-                            isEscaped = true;
-                        }
-                        else
-                        {
-                            isEscaped = false;
-                        }
-
-                        PostProcessArgument(argumentOperation, isEscaped);
-                        _pendingArgumentsToPostProcess.Remove(argumentOperation);
```
sharwell, line 2711: @mavasani In order to use the `ExtractAll` method (which has clear semantics), I needed to hoist this call to `Remove` outside the `foreach` (where it combined with the `Where` method). Considering there is no other code that checks `_pendingArgumentsToPostProcess.Contains`, it seemed like a safe change.

mavasani (approved):

PR #4896, opened 2021-02-23
Support 'await using' in CA2007 (DoNotDirectlyAwaitATask)

### LINKED ISSUE #4888: ConfigureAwait analzyer needs to be updated for IAsyncDisposable
Found in https://github.com/dotnet/roslyn/pull/51328

The appropriate pattern for IAsyncDisposable (with ConfigureAwait) is to do the following:

```c#
                var storage = await persistService.GetStorageAsync(solution, cancellationToken).ConfigureAwait(false);
                await using var _ = storage.ConfigureAwait(false);
```

Note the *two* `.ConfigureAwait(false)` calls.  One on the Task to get the IAsyncDisposable instance, then one on the IAsyncDisposable in the `using` statement to get an appropriate configured awaiter for the `DisposeAsync()` call.

Right now, you can write: 

```c#
                await using var storage = await persistService.GetStorageAsync(solution, cancellationToken).ConfigureAwait(false);
```

Which is correct for getting the value, but not for appropriately calling `DisposeAsync()`.

sharwell:
Fixes #4888

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.Fixer.cs
@@ -33,14 +33,14 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                 context.RegisterCodeFix(
                     new MyCodeAction(title,
                         async ct => await GetFix(context.Document, expression, argument: false, cancellationToken: ct).ConfigureAwait(false),
-                        equivalenceKey: title),
+                        equivalenceKey: nameof(MicrosoftCodeQualityAnalyzersResources.AppendConfigureAwaitFalse)),
                     context.Diagnostics);
 
                 title = MicrosoftCodeQualityAnalyzersResources.AppendConfigureAwaitTrue;
                 context.RegisterCodeFix(
                     new MyCodeAction(title,
                         async ct => await GetFix(context.Document, expression, argument: true, cancellationToken: ct).ConfigureAwait(false),
-                        equivalenceKey: title),
+                        equivalenceKey: nameof(MicrosoftCodeQualityAnalyzersResources.AppendConfigureAwaitTrue)),
                     context.Diagnostics);
             }
         }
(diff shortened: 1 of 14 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.Fixer.cs
```diff
@@ -71,8 +78,26 @@ private class MyCodeAction : DocumentChangeAction
                 base(title, createChangedDocument, equivalenceKey)
             {
             }
+        }
 
-            public override string EquivalenceKey => Title;
+        private sealed class CustomFixAllProvider : DocumentBasedFixAllProvider
```
mavasani, line 83: Is this change related to this PR (support 'await using')?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.Fixer.cs
```diff
@@ -33,14 +33,14 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
                 context.RegisterCodeFix(
                     new MyCodeAction(title,
                         async ct => await GetFix(context.Document, expression, argument: false, cancellationToken: ct).ConfigureAwait(false),
-                        equivalenceKey: title),
+                        equivalenceKey: nameof(MicrosoftCodeQualityAnalyzersResources.AppendConfigureAwaitFalse)),
```
mavasani, line 36: None of the changes in this file seem to be related to this PR. If so, can you please revert?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.Fixer.cs
sharwell: Yes, the test was not able to produce correct results for the Fix All cases without this change.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.Fixer.cs
sharwell: This change gives a predicable equivalence key for the actions that we can use in testing. I might be able to submit it as a separate pull request, but the change is still needed.

mavasani: Sounds good, ideally we should keep these changes separate in future - fine to have them as-is in this PR. Thanks.

mavasani (approved):

PR #4897, opened 2021-02-24
Add source generator for resource files

sharwell:
Add new package Microsoft.CodeAnalysis.ResxSourceGenerator. When installed, this package replaces the need for Resources.designer.cs files in the repository by providing the same functionality through a source generator.

This feature is derived from the RESX generator in dotnet/arcade, but enhanced to support additional project configurations that arcade was able to overlook due to lack of need.


jmarolf (approved):
:shipit:

src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.UnitTests/Verifiers/CSharpVerifierHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        internal static ImmutableDictionary<string, ReportDiagnostic> NullableWarnings { get; } = GetNullableWarningsFromCompiler();
+
+        private static ImmutableDictionary<string, ReportDiagnostic> GetNullableWarningsFromCompiler()
+        {
+            string[] args = { "/warnaserror:nullable" };
+            var commandLineArguments = CSharpCommandLineParser.Default.Parse(args, baseDirectory: Environment.CurrentDirectory, sdkDirectory: Environment.CurrentDirectory);
+            var nullableWarnings = commandLineArguments.CompilationOptions.SpecificDiagnosticOptions;
+
+            // Workaround for https://github.com/dotnet/roslyn/issues/41610
```
Youssef1313, line 27: The issue was fixed and this should no longer be needed. It can be removed from roslyn-sdk repo as well.

Youssef1313:
> When installed, this package replaces the need for Resources.designer.cs files in the repository by providing the same functionality through a source generator.

This repo (as an example) already doesn't include .designer.cs files without source generators. Not sure how this is done in the repo.

sharwell:
> Not sure how this is done in the repo.

dotnet/arcade has MSBuild integration that generates these files as part of a build ("old style" source generation). It's buggy and using it in general projects would be difficult (if it's possible at all).

Youssef1313:
Sounds good. One thing to check is:

- Does this work with the descriptor creation analyzer?
- Is there a way to test both of them working together?

sharwell:
> Does this work with the descriptor creation analyzer?

This feature should work the same for analyzers as any other generation mechanism (all generation strategies run before analyzers).

> Is there a way to test both of them working together?

It's probably possible but I haven't tried to write an analyzer test with a source generator dependency.

src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator/AbstractResxGenerator.cs
```diff
@@ -0,0 +1,709 @@
...
+            if (!context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.RootNamespace", out var rootNamespace))
+            {
+                rootNamespace = context.Compilation.AssemblyName;
+            }
+
+            var resourceName = Path.GetFileNameWithoutExtension(resourceFile.Path);
+            if (options.TryGetValue("build_metadata.AdditionalFiles.RelativeDir", out var relativeDir))
+            {
+                resourceName = relativeDir.Replace(Path.DirectorySeparatorChar, '.').Replace(Path.AltDirectorySeparatorChar, '.') + resourceName;
+            }
```
ericstj, line 75: IIRC the name of the type is actually more nuanced than this.  I believe this is the default but other things can influence the actual value.  Doesn't the resx item itself have this on it?

sharwell: > Doesn't the resx item itself have this on it?

I was going based on what I saw in Arcade. If there's a better way to achieve this let me know.

nuget/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.Package.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
+    <IsPackable>true</IsPackable>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NuspecPackageId>Microsoft.CodeAnalysis.ResxSourceGenerator</NuspecPackageId>
```
Youssef1313, line 9: If there are any plans to add more source generators in the same package in future, it could be renamed to be more general.

src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.UnitTests/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.CSharp.Testing;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+
+namespace Microsoft.CodeAnalysis.ResxSourceGenerator.Test
+{
+    public static partial class CSharpAnalyzerVerifier<TAnalyzer>
```
Youssef1313, line 9: Can we keep only the source generator verifiers?

nuget/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.Package.csproj
sharwell: No immediate plans. Even if additional source generators were implemented, it seems this package could stand on its own.

src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.UnitTests/Verifiers/CSharpVerifierHelper.cs
sharwell: ➡️ Removed workaround from this location

src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.UnitTests/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
sharwell: ➡️ Removed the others

src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator/AbstractResxGenerator.cs
ericstj: Did you check out the targets there and in MSBuild? https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/GenerateResxSource.targets#L75

I don't imagine this source generator is going to have / require targets?

PR #4903, opened 2021-02-26
Update dogfood analyzers to 3.3.2

sharwell:


genlu (approved):

eng/Versions.props
```diff
@@ -31,7 +31,7 @@
     <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
     <MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>3.7.0</MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>
     <MicrosoftCodeAnalysisVersionForTests>3.8.0</MicrosoftCodeAnalysisVersionForTests>
-    <DogfoodAnalyzersVersion>3.3.1</DogfoodAnalyzersVersion>
+    <DogfoodAnalyzersVersion>3.3.2</DogfoodAnalyzersVersion>
```
Evangelink, line 34: Shouldn't we use the `5.0.3` version instead (and of course change the analyzer)?

mavasani: @Evangelink This is the version of rest of the packages, there is a separate property for dogfood version of Net Analyzers package.

Evangelink: Ah my bad got mislead by the name (no longer have the names in my head now).

PR #4906, opened 2021-02-27
Additional performance work from compilation traces collected for dotnet/roslyn-analyzers

### LINKED ISSUE #4905: Heavy allocations in AbstractCategorizedAnalyzerConfigOptions.TryGetOptionValue
Performance trace reveals heavy allocations in `AbstractCategorizedAnalyzerConfigOptions.TryGetOptionValue`, centered around two things that occur on the fast path:

1. String concatenation for the key
2. Capturing the parameters

sharwell:
These changes have less impact from the previous round, but address some low-hanging fruit.

Fixes #4905

### DIFF
--- a/src/Utilities/Compiler/Options/OptionKey.cs
+++ b/src/Utilities/Compiler/Options/OptionKey.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Analyzer.Utilities
+{
+    internal readonly struct OptionKey : IEquatable<OptionKey>
+    {
+        private static readonly ConcurrentDictionary<(string ruleId, string optionName), OptionKey> s_keys = new();
+        private static int s_lastOrdinal;
+
+        private readonly int _ordinal;
+
+        private OptionKey(string name)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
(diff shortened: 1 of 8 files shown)


genlu (approved):
:shipit:

src/Utilities/Compiler/Options/OptionKey.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        public override int GetHashCode()
+        {
+            return _ordinal;
+        }
+
+        public bool Equals(OptionKey other)
+        {
+            return _ordinal == other._ordinal;
```
Youssef1313, line 54: Is depending on this integer for equality and hashcode instead of the `Name` property more performant?

sharwell: Yes

src/Utilities/Compiler/Options/OptionKey.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Analyzer.Utilities
+{
+    internal readonly struct OptionKey : IEquatable<OptionKey>
+    {
+        private static readonly ConcurrentDictionary<(string ruleId, string optionName), OptionKey> s_keys = new();
+        private static int _lastOrdinal;
```
Evangelink, line 12: Do you use the `s_` prefix only for readonly static fields?

sharwell: Supposed to be all static fields I think.

sharwell: Fixed

PR #4908, opened 2021-02-28
Add Analyzer & Codefix - CA1847 -  Use String.Contains(Char) instead of String.Contains(String) with single characters

MeikTranel:
Resolves https://github.com/dotnet/runtime/issues/47180

Outstanding Work:

- [x] roslyn-analyzer violations: Addressed here https://github.com/dotnet/roslyn-analyzers/pull/4908#issuecomment-790182541
  - https://github.com/dotnet/roslyn-analyzers/blob/802b2a7bd94757418e555ddb509de09dccd5f2e2/src/Utilities/Compiler/Options/SymbolNamesWithValueOption.cs#L112
  - https://github.com/dotnet/roslyn-analyzers/blob/802b2a7bd94757418e555ddb509de09dccd5f2e2/src/Utilities/Compiler/RulesetToEditorconfigConverter.cs#L163
- [x] runtime PR (two small violations of the rule): https://github.com/dotnet/runtime/pull/49094

- [x] docs PR: https://github.com/dotnet/docs/pull/23377

The roslyn-analyzer violations are tricky -  they're part of shared code that is used by both frameworks that have String.Contains(char) and those that don't - is it really worth the #if extravaganza here or should i just ignore them?


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,16 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersDescription" xml:space="preserve">
+    <value>String.Contains(String) internally uses an obsolete inner loop when searching for a single character. String.Contains(char) will internally use String.IndexOf(char).</value>
```
Youssef1313, line 1523:
```suggestion
    <value>'string.Contains(string)' internally uses an obsolete inner loop when searching for a single character. 'string.Contains(char)' will internally use 'string.IndexOf(char)'.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,16 @@
...
+  </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersMessage" xml:space="preserve">
+    <value>Consider using String.Contains(char) instead of String.Contains(String) when searching for a single character</value>
```
Youssef1313, line 1526:
```suggestion
    <value>Consider using 'string.Contains(char)' instead of 'string.Contains(string)' when searching for a single character</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,16 @@
...
+  </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersTitle" xml:space="preserve">
+    <value>Use String.Contains(char) instead of String.Contains(String) with single characters</value>
```
Youssef1313, line 1529:
```suggestion
    <value>Use 'string.Contains(char)' instead of 'string.Contains(string)' with single characters</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var violatingNode = root.FindNode(context.Span, getInnermostNodeForTie: true);
+
+            context.RegisterCodeFix(new ReplaceStringLiteralWithCharLiteralCodeAction(context.Document, violatingNode), context.Diagnostics);
+        }
+
+        public override FixAllProvider GetFixAllProvider()
+        {
+            return base.GetFixAllProvider();
```
Youssef1313, line 31: I think this should be BatchFixer:

```suggestion
            return WellKnownFixAllProviders.BatchFixer;
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Performance
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class UseStringContainsCharOverloadWithSingleCharactersAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string CA1839 = "CA1839";
```
Youssef1313, line 16: nit:

```suggestion
        internal const string CA1839 = nameof(CA1839);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+            RuleLevel.IdeSuggestion,
+            description: s_localizableDescription_CA1839,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(s_rule_CA1839);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```
Youssef1313, line 35:
```suggestion
            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+            context.EnableConcurrentExecution();
+            context.RegisterCompilationStartAction(CheckIfRuleIsApplicableAndRegister);
+        }
+
+        private void CheckIfRuleIsApplicableAndRegister(CompilationStartAnalysisContext ctx)
+        {
+            var stringType = ctx.Compilation.GetSpecialType(SpecialType.System_String);
+            var charType = ctx.Compilation.GetSpecialType(SpecialType.System_Char);
+            var stringMembers = stringType.GetMembers();
```
Youssef1313, line 44: This can throw in some edge cases (I think when corlib isn't included in the compilation).
You may want to bail-out if stringType is null, or charType is null.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+            var ruleIsApplicable = stringType.GetMembers("Contains").OfType<IMethodSymbol>().Any(m =>
+            {
+                var firstArgument = m.Parameters.First();
+                if (firstArgument is null)
+                    return false;
```
Youssef1313, line 49: I don't think it can be null. You may want to check that parameters length > 1 and check firstArgument[0]. Or alternatively: `return SymbolEqualityComparer.Default.Equals(m.Parameters.FirstOrDefault(), charType)`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                return firstArgument.Type.Equals(charType);
```
Youssef1313, line 51: Symbol comparisons should use SymbolEqualityComparer

src/Test.Utilities/CSharpCodeFixVerifier`2.cs
```diff
@@ -23,28 +24,39 @@ public static DiagnosticResult Diagnostic(string diagnosticId)
         public static DiagnosticResult Diagnostic(DiagnosticDescriptor descriptor)
             => CSharpCodeFixVerifier<TAnalyzer, TCodeFix, XUnitVerifier>.Diagnostic(descriptor);
 
-        public static async Task VerifyAnalyzerAsync(string source, params DiagnosticResult[] expected)
+        public static async Task VerifyAnalyzerAsync(string source, ReferenceAssemblies referenceAssemblies, params DiagnosticResult[] expected)
```
Youssef1313, line 27: @sharwell Can you take a look?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
```diff
@@ -0,0 +1,193 @@
...
+    {
+        [Fact]
+        public async Task CSharp_RegularStringLiteralWithoutCharOverloadAvailable_NoDiagnostic()
+        {
+            var sourceCode = @"using System;
+public class TestClass
+{
+    public bool TestMethod(string input)
+    {
+        return ""testString"".Contains({|#0:""a""|});
```
Youssef1313, line 26: I'm not sure why you have markup here given that there is no diagnostic.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}";
+            var expectedDiagnostic = VerifyCS.Diagnostic(UseStringContainsCharOverloadWithSingleCharactersAnalyzer.s_rule_CA1839)
+                .WithLocation(0);
+            await VerifyCS.VerifyAnalyzerAsync(violatingSourceCode, ReferenceAssemblies.NetStandard.NetStandard21, expectedDiagnostic);
```
Youssef1313, line 45: Any tests with expected diagnostics should use VerifyCodeFixAsync

src/Test.Utilities/CSharpCodeFixVerifier`2.cs
sharwell: This should not change. Tests that need a custom `ReferenceAssemblies` instance should use the full form:

```csharp
await new VerifyCS.Test
{
  ReferenceAssemblies = referenceAssemblies,
  ...
}.RunAsync();
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
sharwell: > Any tests with expected diagnostics should use VerifyCodeFixAsync.

Yes, specifically for the reason that all calls to `VerifyCodeFixAsync` _automatically_ call `VerifyAnalyzerAsync`, so it's literally twice the work for the same result.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
MeikTranel: What can i say - i am lazy man. 🤣

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
MeikTranel: Its just a relic of the process. I wrote tests to verify my analyzer first and then moved on with the codefixer, but you're right it looks pretty obsolete given the other tests.

src/Test.Utilities/CSharpCodeFixVerifier`2.cs
MeikTranel: I expected this looking at this with fresh eyes today - i can move these helpers back into the test class tho, right? I assume the concern is with bloating the class with all kinds of unique use cases?

MeikTranel:
@sharwell / @Youssef1313 
two questions while digging into fixing the outstanding issues:
- Should i create an issue before opening up a PR on the runtime or can i just throw them the PR if the amount of violations are small?
- Same question basically for the docs PR

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
MeikTranel: Isnt it safe to call the regular `.Equals` if the Symbol Interface (`IMethodSymbol` in this case) implements `IEquatable<ISymbol>` or am i completely off?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
MeikTranel: I like it :D

MeikTranel:
Two outstanding violations remain - in the issue we decided that the information severity is probably better suited.

The specific violations in question are part of the shproj compile items, which have lots of different frameworks they're compiled against (See comment for reference https://github.com/dotnet/roslyn-analyzers/pull/4908#issue-581615628) - much like the runtime repo i dont think the change would have been worth it, but i'm not firm on that stance - if y'all prefer the ol' `#if NETCOREAPP3_1` shuffle i'd be happy to do it.

As for the other repos:
- runtime - PR (https://github.com/dotnet/runtime/pull/49094)
- roslyn - no violations found by the test tool (performance impact recorded was also neglible IIRC - i can generate another report if y'all are interested)
- roslyn-analyzers - as said above
- aspnetcore - wasn't able to finish the compile yet

Docs PR: https://github.com/dotnet/docs/pull/23101

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
Youssef1313: > If i get this right this would just plain disable the analyzer for generated code

@MeikTranel This is correct. We usually don't report on generated code unless for security rules.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
Youssef1313: Using Equals without specifying a SymbolEqualityComparer is usually okay. But after the introduction of nullable reference types, SymbolEqualityComparer was added so the intent is clear on whether the developer wants to include nullability or not.

There is a rule (RS1024) that flags calls without SymbolEqualityComparer, it's unfortunately currently disabled due to a codefix crash.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        [Fact]
+        public async Task VB_RegularStringLiteral_Fixed()
+        {
+            var test = new VerifyVB.Test()
+            {
+                TestCode = $@"Public Class Program
+    Public Sub M()
+        Dim a = ""testString"".Contains({{|{UseStringContainsCharOverloadWithSingleCharactersAnalyzer.CA1839}:""t""|}})
```
Youssef1313, line 144: For simplicity I think you can use `[|` `|]` instead of `{|CA1839:` `|}`. This will make things shorter and avoids the need of interpolation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
MeikTranel: I was wondering what that was about - thanks for the explanation 😊

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
MeikTranel: Will the former notation default to the analyzer listed in the type variable or would this allow any diagnostic to fulfill the implicit assertion?
On another note i was looking for a documentation regarding the marker syntax but was ultimately unable so i just went off usage in the repo - is there anything to refer to?

Youssef1313: > Will the former notation default to the analyzer listed in the type variable or would this allow any diagnostic to fulfill the implicit assertion?

This allows only diagnostics from the analyzer under tested (i.e, compile errors for example won't be fulfilled). It also requires that the analyzer under test has only one rule id.

Tagging @sharwell for enhancing the documentation in [this README](https://github.com/dotnet/roslyn-sdk/blob/master/src/Microsoft.CodeAnalysis.Testing/README.md).

AraHaan:
Personally I think this can be done, on TFM's that does not have ``String.Contains(char)`` overload of contains on string could in fact do ``String.IndexOf(char) > -1`` to check if it contains that character on those TFM's only.

In fact I would like it if that was added to the resource description on this rule that is being added.

First of all we need to calculate what TFMs do not contain the ``String.Contains(char)`` or for security have to codefix basically do:
```cs
#if /*Frameworks that supply the String.Contains(char) overload. */
// the actual fix to change that here.
#else
// the rest of the code but instead replace it with "String.IndexOf(char) > -1".
#endif
```

MeikTranel:
@AraHaan Like in this PR? https://github.com/dotnet/runtime/pull/49094

> In fact I would like it if that was added to the resource description on this rule that is being added.

I am unsure - are you proposing we add this as an addendum to the `How to fix` section of the documentation or you proposing adding this to the codefix automation itself?

I can see value in the documentation part, but i'm unsure if this is worth setting the precedent for nor am sure if its even possible to detect the context beyond the single target framework compilation.

AraHaan:
Yes but does that also apply to netcoreapp3.1+ as well too since .NET Core 3.1 implements stardard2.1?

MeikTranel:
@sharwell can you take a look at this again - i think a resolved all issues raised so far and it feels like the re-request for review went under your radar 😄

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private static async Task VerifyCSCodeFixWithGivenReferenceAssemblies(string source, ReferenceAssemblies referenceAssemblies, string fixedSource)
+        {
+            var test = new VerifyCS.Test()
+            {
+                TestCode = source,
+                ReferenceAssemblies = referenceAssemblies,
+                FixedCode = fixedSource
+            };
+            await test.RunAsync();
```
sharwell, line 165:
```suggestion
            await new VerifyCS.Test()
            {
                ReferenceAssemblies = referenceAssemblies,
                TestCode = source,
                FixedCode = fixedSource,
            }.RunAsync();
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,87 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(s_rule_CA1839);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+            context.RegisterCompilationStartAction(CheckIfRuleIsApplicableAndRegister);
+        }
+
+        private void CheckIfRuleIsApplicableAndRegister(CompilationStartAnalysisContext ctx)
```
sharwell, line 40: 💡 All context parameters are typically named `context`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,87 @@
...
+                                && invokedMethod.ContainingType.SpecialType.Equals(SpecialType.System_String)
+                                && invokedMethod.Name == "Contains"
+                                && invokedMethod.Parameters[0].Type.SpecialType.Equals(SpecialType.System_String);
+            }
+
+            static bool ViolatesRule(IArgumentOperation primaryArgument)
+            {
+                return primaryArgument.Value is ILiteralOperation literalOperation
+                    && literalOperation.ConstantValue.HasValue
+                    && literalOperation.ConstantValue.Value is string constantString
```
sharwell, line 82: 💭 I believe `GetValueOrDefault()` will work here as well

```suggestion
                    && literalOperation.ConstantValue.GetValueOrDefault() is string constantString
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            // Bail out of further evaluations in special cases where string/char types arent even available
+            if (stringType is null || charType is null)
+                return;
+
+            var ruleIsApplicable = stringType.GetMembers("Contains").OfType<IMethodSymbol>().Any(m =>
+            {
+                if (m.Parameters.Length > 0)
+                {
+                    return SymbolEqualityComparer.Default.Equals(m.Parameters[0].Type, charType);
```
sharwell, line 53: 💡 This can also use `SpecialType`

```suggestion
                    return m.Parameters[0].Type.SpecialType == SpecialType.System_Char;
```

Youssef1313 (changes requested):
Can you target the PR to `release/6.0.1xx-preview3` branch?

MeikTranel:
Sure - and rebase for good measure? saw that the CodeCov bot was throwing shade because the target branch received some updates to coverage 😄 ?

MeikTranel:
Weeeee... that cant be good. @Youssef1313 how do we proceed here? theres a conflict with the rule id - someone already took that on the release preview 3 branch. Should i just go and move mine to 1841? (will have to update docs as well)

sharwell (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,16 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersDescription" xml:space="preserve">
+    <value>'string.Contains(string)' internally uses an obsolete inner loop when searching for a single character. 'string.Contains(char)' will internally use 'string.IndexOf(char)'.</value>
```
sharwell, line 1523: ❓ Why not update `string.Contains(string)` to use `string.IndexOf(char)` automatically? It seems like this description is prone to quickly becoming obsolete.

Youssef1313:
> Should i just go and move mine to 1841? (will have to update docs as well)

Yes move to the next unused id.

The conflicts should (hopefully) not be so hard.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
MeikTranel: I asked this question in the issue since it wasnt clear from the spec and i was wondering the same:

https://github.com/dotnet/runtime/issues/47180#issuecomment-779481857

Basically the idea is that the Analyzers are still used by older frameworks where it is unlikely that this state will change anytime soon. I agree with your reasoning though i'm not sure how to word it better. 😄

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
```diff
@@ -0,0 +1,67 @@
...
+                _sourceStringLiteral = sourceStringLiteral;
+            }
+
+            protected override async Task<Document> GetChangedDocumentAsync(CancellationToken cancellationToken)
+            {
+                var editor = await DocumentEditor.CreateAsync(_document, cancellationToken).ConfigureAwait(false);
+                var semanticModel = await _document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+                if (semanticModel.GetOperation(_sourceStringLiteral, cancellationToken) is ILiteralOperation literalOperation && literalOperation.ConstantValue.HasValue)
+                {
+                    if (literalOperation.ConstantValue.Value is string stringValue && char.TryParse(stringValue, out char value))
```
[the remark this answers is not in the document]
MeikTranel: Haha i'm absolutely not following here - Are you saying i could somehow avoid loading the semantic model by passing the fixed literal node together with analyzer?

Anyhow... can you go into detail - maybe link another CodeFix where this happens?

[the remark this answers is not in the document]
Youssef1313: @MeikTranel In the analyzer, when you report a diagnostic, you can pass a dictionary which can retrieved in the codefix. See the `properties` parameter in `context.ReportDiagnostic`.

In the fixer, you can access it through `Properties` property of the diagnostic.

It's like you're caching things you already did in the analyzer to not do it again in the fixer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
MeikTranel: `IOperation.ConstantValue` sadly is `Optional<object>`  instead of a nullable object, so no `GetValueOrDefault` for me if theres no established helper that i dont know of 😄

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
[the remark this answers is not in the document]
MeikTranel: Haha nice. That feels a little sketchy, but i'll give it a try. Itll probably be another day though since the rebase turned out to be a little more sketchy than expected (autogenerated files are fun :smile: )

MeikTranel:
I updated the docs PR (messed up the PR had to start a new one 🤦‍♂️) https://github.com/dotnet/docs/pull/23377 with the new id.
Will try to get that performance improvement done y'all suggested tomorrow.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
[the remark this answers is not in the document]
MeikTranel: @Youssef1313 i think i've got though it feels sketchy NGL

MeikTranel:
@sharwell i updated the rebased the branch, but i cant seem to get rid of that seemingly unrelated test failure. Can you take a look at this? This PR has been in limbo for a while and changing the analyzer ids really isnt fun enough warrant a second time 😄

AraHaan:
I do not see which specific test failed, perhaps it failed because you overwritten another diagnostic id?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
sharwell: That doesn't seem like a reason that would warrant an analyzer to fix. Are the performance problems with the current code bad enough that we _need_ to put in an analyzer as an urgent patch for users who can't update to a version which is already fixed?

sharwell: I _really_ don't like this description. `string.Contains(string)` is a well-defined and reasonably-performing method when used with a single character. It is not obsolete for this usage, and will not be obsolete for this usage in the future.

sharwell: I'll send @buyaa-n a message and try to get this resolved.

mavasani: I tend to agree with @sharwell here.

buyaa-n: Well seems i did not exactly suggest this message, sorry if I misguide you, i agree we shouldn't say that it is "internally uses an obsolete inner loop". My point was to mention that there is 'string.Contains(char)' overload is available which performs better for single char lookup, and that should be enough description even for experienced developers as there was no char overload before.

MeikTranel: I'll update the string tonight:
Does `'string.Contains(char)' is available as a better performing overload for single char lookup.`  work for y'all?

BTW: the obsolete in the original description wasnt saying the overload itself was obsolete. i was saying running that extra loop was obsolete if you know you're searching for a single char.

Youssef1313:
> I do not see which specific test failed, perhaps it failed because you overwritten another diagnostic id that recently been added again?
> 
> But it does look unrelated enough could this be merged, then the tests fixed in a follow up pr just to get this in to preview 3?

Is that about the current CI failure? If so, this was a bug causing flakiness on Roslyn's side that @sharwell has fixed recently. Restarting the CI run will pass normally.

sharwell:
This build failure:

```
System.AggregateException : One or more errors occurred. (Unexpected Null)\n---- System.InvalidOperationException : Unexpected Null
   at Microsoft.CodeAnalysis.Host.RecoverableWeakValueSource`1.GetValueAsync(CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.RecoverableTextAndVersion.GetValueAsync(CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.TextDocumentState.GetTextAndVersionAsync(CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.TextDocumentState.GetTextAsync(CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.Testing.CodeFixTest`1.VerifyFixAsync(String language, ImmutableArray`1 analyzers, ImmutableArray`1 codeFixProviders, SolutionState oldState, SolutionState newState, Int32 numberOfIterations, Func`10 getFixedProject, IVerifier verifier, CancellationToken cancellationToken) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs:line 427
   at Microsoft.CodeAnalysis.Testing.CodeFixTest`1.VerifyFixAsync(SolutionState testState, SolutionState fixedState, SolutionState batchFixedState, IVerifier verifier, CancellationToken cancellationToken) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs:line 372
   at Microsoft.CodeAnalysis.Testing.CodeFixTest`1.RunAsync(CancellationToken cancellationToken) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs:line 248
   at Microsoft.CodeAnalysis.Analyzers.UnitTests.MetaAnalyzers.ReleaseTrackingAnalyzerTests.TestCodeFixToEnableAnalyzerReleaseTracking() in /_/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/ReleaseTrackingAnalyzerTests.cs:line 100
----- Inner Stack Trace -----
   at Roslyn.Utilities.Contract.Fail(String message)
   at Roslyn.Utilities.Contract.ThrowIfNull[T](T value, String message)
   at Microsoft.CodeAnalysis.RecoverableTextAndVersion.RecoverableText.RecoverAsync(CancellationToken cancellationToken)
```

was fixed by https://github.com/dotnet/roslyn/pull/51244.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1550,4 +1550,16 @@
   <data name="UseValidPlatformStringNoVersion" xml:space="preserve">
     <value>Version '{0}' is not valid for platform '{1}'. Do not use versions for this platform.</value>
   </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersDescription" xml:space="preserve">
+    <value>'string.Contains(char)' is available as a better performing overload for single char lookup.</value>
```
AraHaan, line 1554: should this read as ``'string.Contains(char)' is available as a better performing overload for a single character lookup.``

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1550,4 +1550,16 @@
...
+  </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersMessage" xml:space="preserve">
+    <value>Consider using 'string.Contains(char)' instead of 'string.Contains(string)' when searching for a single character</value>
+  </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersTitle" xml:space="preserve">
+    <value>Use 'string.Contains(char)' instead of 'string.Contains(string)' with single characters</value>
```
AraHaan, line 1560: Should this read as ``Use 'string.Contains(char)' instead of 'string.Contains(string)' with an single character.``?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1550,4 +1550,16 @@
...
+  </data>
+  <data name="ReplaceStringLiteralWithCharLiteralCodeActionTitle" xml:space="preserve">
+    <value>Replace string literal with char literal</value>
```
AraHaan, line 1563: Should this end with an ``.``?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1550,4 +1550,16 @@
   <data name="UseValidPlatformStringNoVersion" xml:space="preserve">
     <value>Version '{0}' is not valid for platform '{1}'. Do not use versions for this platform.</value>
   </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersDescription" xml:space="preserve">
+    <value>'string.Contains(char)' is available as a better performing overload for single char lookup.</value>
+  </data>
+  <data name="UseStringContainsCharOverloadWithSingleCharactersMessage" xml:space="preserve">
+    <value>Consider using 'string.Contains(char)' instead of 'string.Contains(string)' when searching for a single character</value>
```
AraHaan, line 1557: Should this end with an ``.`` as well?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: > Does 'string.Contains(char)' is available as a better performing overload for single char lookup. work for y'all?

Sounds fine, another version could be `For a single char lookup the 'string.Contains(char)' overload would perform better than  'string.Contains(string)'` 

> BTW: the obsolete in the original description wasnt saying the overload itself was obsolete. i was saying running that extra loop was obsolete if you know you're searching for a single char.

Understand your intention, but for a quick glance it does sound like it has an obsolete loop 😄

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n:
```suggestion
    <value>Use 'string.Contains(char)' overload for a single character lookup</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: No, as far as I only description need `.`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+    public sealed class UseStringContainsCharOverloadWithSingleCharactersAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string CA1841 = nameof(CA1841);
+        internal const string SourceStringPropertyKey = "SourceString";
+        private static readonly LocalizableString s_localizableTitle_CA1841 = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseStringContainsCharOverloadWithSingleCharactersTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage_CA1841 = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseStringContainsCharOverloadWithSingleCharactersMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString? s_localizableDescription_CA1841 = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseStringContainsCharOverloadWithSingleCharactersDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
```
buyaa-n, line 21:
```suggestion
        private static readonly LocalizableString s_localizableDescription_CA1841 = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseStringContainsCharOverloadWithSingleCharactersDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: no

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n, line 1560:
```suggestion
    <value>Use char literal for a single character lookup</value>
```

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -1332,6 +1332,18 @@ Marshalling of 'StringBuilder' always creates a native buffer copy, resulting in
 |CodeFix|True|
 ---
 
+## [CA1841](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1841): Use 'string.Contains(char)' instead of 'string.Contains(string)' with single characters
```
buyaa-n, line 1335:
```suggestion
## [CA1841](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1841): Use char literal for a single character lookup
```

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif
```diff
@@ -2485,6 +2485,26 @@
             ]
           }
         },
+        "CA1841": {
+          "id": "CA1841",
+          "shortDescription": "Use 'string.Contains(char)' instead of 'string.Contains(string)' with single characters",
```
buyaa-n, line 2490:
```suggestion
          "shortDescription": "Use char literal for a single character lookup",
```

src/NetAnalyzers/RulesMissingDocumentation.md
```diff
@@ -5,3 +5,4 @@ Rule ID | Missing Help Link | Title |
 CA1418 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1418> | Use valid platform string |
 CA1839 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1839> | Use 'Environment.ProcessPath' |
 CA1840 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1840> | Use 'Environment.CurrentManagedThreadId' |
+CA1841 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1841> | Use 'string.Contains(char)' instead of 'string.Contains(string)' with single characters |
```
buyaa-n, line 8:
```suggestion
CA1841 | <https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1841> | Use char literal for a single character lookup |
```

buyaa-n (approved):
Thanks, @MeikTranel, i left 2 main comments:
1. change the title from `Use 'string.Contains(char)' instead of 'string.Contains(string)' with single characters`  => `Use char literal for a single character lookup` (because it is used in many places as doc commented on all of them)
2. There is unneeded nullable annotation (`?`) in one of the `LocalizableString` 

Except that looks great!

MeikTranel:
Hey @buyaa-n the issues should be resolved now. 
Quick heads up: Unless you're willing to provide review suggestions for like 16 files you're probably wasting your time - you'd have to build locally to sync and generate new XLF files anyways so providing this once would have been enough 😆

buyaa-n:
Thank you @MeikTranel, looks good to me, I will let the repo owners finish their review CC @sharwell @mavasani

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+            });
+
+            if (ruleIsApplicable)
+                context.RegisterOperationAction(AnalyseContainsUsage, OperationKind.Invocation);
+        }
+
+        private void AnalyseContainsUsage(OperationAnalysisContext context)
+        {
+            var invocationOperation = (IInvocationOperation)context.Operation;
+            if (ShouldAnalyzeInvocations(invocationOperation) && CheckForViolationAndTransformLiteral(invocationOperation.Arguments[0], out var sourceString))
```
mavasani, line 67: `invocationOperation.Arguments[0]` for below invocation to `M` will give you first named argument that is passed to the second parameter `p2`:

```csharp
void M(char p1, int p2);
M(p2: 1, 'c');

Comment (mavasani) [reply src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs:67]: I think you need to iterate through `invocationOperation.Arguments` and find the argument whose `.Parameter` has index `0` OR parameter symbol matches.

Comment (mavasani) [reply src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs:67]: Actually, you can just use this extension method that @buyaa-n recently added: https://github.com/dotnet/roslyn-analyzers/blob/04883a7d9343d5e757f2c1a5401b74d9e8d51eaa/src/Utilities/Compiler/Extensions/IOperationExtensions.cs#L787-L789

Comment (mavasani) [reply src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs:67]: Please do add a unit test for this case.

File: src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs [hunk 29]
```diff
@@ -0,0 +1,97 @@
...
+                propertyBuilder.Add(SourceStringPropertyKey, sourceString);
+                context.ReportDiagnostic(invocationOperation.Arguments[0].CreateDiagnostic(s_rule_CA1841, propertyBuilder.ToImmutable()));
+            }
+
+            static bool ShouldAnalyzeInvocations(IInvocationOperation invocationOperation)
+            {
+                return invocationOperation.TargetMethod is IMethodSymbol invokedMethod
+                                && invokedMethod.ContainingType.SpecialType.Equals(SpecialType.System_String)
+                                && invokedMethod.Name == "Contains"
+                                && invokedMethod.Parameters[0].Type.SpecialType.Equals(SpecialType.System_String);
```
mavasani, line 79: This will throw an exception if the `Contains` method has no parameters (possible for say a malformed or user-defined corlib). I think we should add a defensive length check.
```suggestion
                                && invokedMethod.Parameters.Length > 0
                                && invokedMethod.Parameters[0].Type.SpecialType.Equals(SpecialType.System_String);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani, line 1557: Message should be more prescriptive.
```suggestion
    <value>Use 'string.Contains(char)' instead of 'string.Contains(string)' when searching for a single character</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
```diff
@@ -0,0 +1,65 @@
...
+    public sealed class UseStringContainsCharOverloadWithSingleCharactersCodeFix : CodeFixProvider
+    {
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(
+            UseStringContainsCharOverloadWithSingleCharactersAnalyzer.CA1841);
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var violatingNode = root.FindNode(context.Span, getInnermostNodeForTie: true);
+            if (context.Diagnostics[0].Properties.TryGetValue(UseStringContainsCharOverloadWithSingleCharactersAnalyzer.SourceStringPropertyKey, out var sourceString))
```
mavasani, line 24: Please add a defensive null check for `Properties`
```suggestion
            if (context.Diagnostics[0].Properties?.TryGetValue(UseStringContainsCharOverloadWithSingleCharactersAnalyzer.SourceStringPropertyKey, out var sourceString) == true)
```

mavasani (commented):
Overall, looks great to me. Kindly address the review comments and also retarget to preview4 branch. Should be ready to merge then. Thanks again for your contribution!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
MeikTranel: oh hell yeah that could've went south real fast :D

MeikTranel: Is it legal for me to use both CSharp and VB Syntax Types in a language agnostic analyzer or do i have to split this off into separate analyzers? Because theres no IOperation way to detect named parameters and i need to detect them to be able to tell the fixer that it needs to generate a named argument.

mavasani (changes requested):
Marking as request changes as the PR needs to be retargeted to [release/6.0.1xx](https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx) branch.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -5,6 +5,7 @@ Rule ID | Category | Severity | Notes
 CA1418 | Interoperability | Warning | UseValidPlatformString, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1418)
 CA1839 | Performance | Info | UseEnvironmentMembers, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1839)
 CA1840 | Performance | Info | UseEnvironmentMembers, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1840)
+CA1841 | Performance | Info | UseStringContainsCharOverloadWithSingleCharactersAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1841)
```
mavasani, line 8: This ID is already taken up: https://github.com/dotnet/roslyn-analyzers/blob/release/6.0.1xx/src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
mavasani: Please update.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
mavasani: > use both CSharp and VB Syntax Types in a language agnostic analyzer

No, please don't do so. Instead make the language agnostic type abstract and add non-abstract deriving types in C# and VB projects in this layer for language specific functionality.

MeikTranel: Okay - will do so after work today.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
MeikTranel: Had to go a little different route since there were changes to that section and the nullability analyzer was really not having it today with the exact change you proposed (honestly dont know why - it kept accusing that sourceString may be null even if i double checked it with a shortcircuit null check after the TryGet call) - hope this is ok now

MeikTranel:
Has anyone ever had this luck of being pushed from 5 numbers in a single PR 🤣🤣

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Performance
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpUseStringContainsCharOverloadWithSingleCharactersAnalyzer : global::Microsoft.NetCore.Analyzers.Performance.UseStringContainsCharOverloadWithSingleCharactersAnalyzer
```
Youssef1313, line 11:
```suggestion
    public sealed class CSharpUseStringContainsCharOverloadWithSingleCharactersAnalyzer : UseStringContainsCharOverloadWithSingleCharactersAnalyzer
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharacters.Fixer.cs
```diff
@@ -0,0 +1,73 @@
...
+                _document = document;
+                _sourceStringLiteral = sourceStringLiteral;
+                _sourceString = sourceString;
+                _argumentName = argumentName;
+            }
+
+            protected override async Task<Document> GetChangedDocumentAsync(CancellationToken cancellationToken)
+            {
+                var editor = await DocumentEditor.CreateAsync(_document, cancellationToken).ConfigureAwait(false);
+                if (char.TryParse(_sourceString, out char value))
```
Youssef1313, line 59: Can this condition be false? If not, can you add an assertion?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Performance/BasicUseStringContainsCharOverloadWithSingleCharactersAnalyzer.vb
```diff
@@ -0,0 +1,29 @@
...
+
+Namespace Microsoft.NetCore.VisualBasic.Analyzers.Performance
+
+    <DiagnosticAnalyzer(LanguageNames.VisualBasic)>
+    Public NotInheritable Class BasicUseStringContainsCharOverloadWithSingleCharactersAnalyzer
+        Inherits UseStringContainsCharOverloadWithSingleCharactersAnalyzer
+
+        Protected Overrides Function IsNamedArgument(argumentOperation As IArgumentOperation, ByRef name As String) As Boolean
+            name = ""
+            If argumentOperation.Syntax.GetType() IsNot GetType(SimpleArgumentSyntax) Then
```
Youssef1313, line 17:
```suggestion
            If TypeOf argumentOperation.Syntax IsNot SimpleArgumentSyntax Then
```

mavasani:
> Has anyone ever had this luck of being pushed back 5 numbers in a single PR

Yes, that is very common. We have large number of community contributors, and bunch of WIP new analyzer PRs. If you prefer, you can hold off bumping up the rule ID until you have received a complete review/approval and then do a final merge resolution, which should require bumping the rule ID

mavasani:
@MeikTranel Just FYI: There are 3 more new analyzer PRs opened prior to yours that need a review/approval. I am slowly getting through them, but please expect a bit of delay/merge conflicts as we wade through them.

MeikTranel:
Haha no issue. My ID update chops have been honed by now 🤣

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+            });
+
+            if (ruleIsApplicable)
+                context.RegisterOperationAction(AnalyseContainsUsage, OperationKind.Invocation);
+        }
+
+        private void AnalyseContainsUsage(OperationAnalysisContext context)
+        {
+            var invocationOperation = (IInvocationOperation)context.Operation;
+            var argumentOperation = invocationOperation.Arguments.GetArgumentForParameterAtIndex(0);
```
mavasani, line 67: This method will throw an invalid operation exception for invocations without arguments. 
```suggestion
            if (invocationOperation.Arguments.Length == 0)
            {
                return;
            }
            
            var argumentOperation = invocationOperation.Arguments.GetArgumentForParameterAtIndex(0);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (ShouldAnalyzeInvocations(invocationOperation) && CheckForViolationAndTransformLiteral(argumentOperation, out var sourceString))
+            {
+                var propertyBuilder = ImmutableDictionary.CreateBuilder<string, string?>(StringComparer.Ordinal);
+                propertyBuilder.Add(SourceStringPropertyKey, sourceString);
+                if (IsNamedArgument(argumentOperation, out var name))
+                    propertyBuilder.Add(ArgumentNamePropertyKey, name);
+                context.ReportDiagnostic(argumentOperation.CreateDiagnostic(s_rule_CA1846, propertyBuilder.ToImmutable()));
+            }
+
+            static bool ShouldAnalyzeInvocations(IInvocationOperation invocationOperation)
```
mavasani, line 77: Nit: rename to `ShouldAnalyzeInvocation`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
mavasani: Alternatively, you can move the `ShouldAnalyzeInvocation(invocationOperation)` check before `GetArgumentForParameterAtIndex` call

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+        private void AnalyseContainsUsage(OperationAnalysisContext context)
+        {
+            var invocationOperation = (IInvocationOperation)context.Operation;
+            var argumentOperation = invocationOperation.Arguments.GetArgumentForParameterAtIndex(0);
+            if (ShouldAnalyzeInvocations(invocationOperation) && CheckForViolationAndTransformLiteral(argumentOperation, out var sourceString))
+            {
+                var propertyBuilder = ImmutableDictionary.CreateBuilder<string, string?>(StringComparer.Ordinal);
+                propertyBuilder.Add(SourceStringPropertyKey, sourceString);
+                if (IsNamedArgument(argumentOperation, out var name))
+                    propertyBuilder.Add(ArgumentNamePropertyKey, name);
```
mavasani, line 73: Seems very weird to have the `IsNamedArgument` check in the analyzer, as the analyzer does not need this information at all. Can you instead move this code to the fixer and also remove `ArgumentNamePropertyKey`?

sharwell:
@mavasani We resolved the numbering problem in StyleCop Analyzers by assigning IDs at the time analyzer proposals were approved. We sometimes end up with a gap if a later proposal gets implemented first, but it avoids the conflicts.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersAnalyzer.cs
mavasani: Sounds good, thanks!

MeikTranel:
@mavasani i was able to move this out successfully like we talked about - theres still some outstanding conflicts but those are ID conflicts again - i can fix those once the regular review stuff has been resolved. Just make sure to tag me so my phone gives me a notification 🙃

mavasani (approved):

mavasani:
@MeikTranel this is ready to merge once you resolve the merge conflicts.

MeikTranel:
should be good to go now ;)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        [Fact]
+        public async Task CSharp_RegularStringLiteral_Fixed()
+        {
+            var violatingSourceCode = @"using System;
+public class TestClass
+{
+    public bool TestMethod(string input)
+    {
+        return ""testString"".Contains([|""a""|]);
```
Youssef1313, line 25: Probably not very important, but consider adding a test for constant interpolated strings, i.e `return "testString".Contains($"a");` or `"testString.Contains($"{CONST_CHAR_DEFINED_ABOVE}");`

MeikTranel: That use case was specifically excluded in the source issue.

MeikTranel: Unless roslyn somehow optimizes constant values in a way that the interpolation with a constant value would end up classifying as a `LiteralSyntax`?

mavasani (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseStringContainsCharOverloadWithSingleCharactersTests.cs
Youssef1313: @MeikTranel I couldn't see anything about it in the comments. Was that said in the API review video?

PR #4909, opened 2021-03-01
Add analyzers/fixers for Environment.ProcessPath and CurrentManagedThreadId

stephentoub:
Fixes https://github.com/dotnet/runtime/issues/43257
Fixes https://github.com/dotnet/runtime/issues/42948

cc: @carlossanlop, @mavasani, @jkotas


src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,6 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA1839 | Performance | Info | UseEnvironmentMembers, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1839)
+CA1840 | Performance | Info | UseEnvironmentMembers, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1840)
```
Youssef1313, line 6: Shouldn't CA1837 be reused based on https://github.com/dotnet/runtime/issues/43257#issuecomment-713061245?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+                    return;
+                }
+
+                // Everything we're looking for is a property, so find all property references.
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    var initialPropRef = (IPropertyReferenceOperation)operationContext.Operation;
+
+                    // Process.GetCurrentProcess().Id
+                    if (lookForProcessId && processIdSymbol!.Equals(initialPropRef.Property))
```
Youssef1313, line 94: I think this should use SymbolEqualityComparer.Default

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembersFixer.cs
```diff
@@ -32,13 +35,31 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+
+                    case UseEnvironmentMembers.EnvironmentProcessPathRuleId:
+                        title = MicrosoftNetCoreAnalyzersResources.UseEnvironmentProcessPathFix;
+                        memberName = "ProcessPath";
+                        break;
+
+                    default:
+                        title = MicrosoftNetCoreAnalyzersResources.UseEnvironmentCurrentManagedThreadIdFix;
+                        memberName = "CurrentManagedThreadId";
+                        break;
```
Youssef1313, line 54: It may be better to add a third case, and make the default case throw.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Linq;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>CA1837: Use Environment.ProcessId.</summary>
```
Youssef1313, line 13: nit: Include the other rules as well.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
```diff
@@ -3,15 +3,15 @@
 using System.Threading.Tasks;
```
Youssef1313, line 3: nit: rename test file name.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                // Require that Environment / Process / ProcessModule / Thread exist.
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemEnvironment, out var environmentType) ||
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcess, out var processType) ||
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcessModule, out var processModuleType) ||
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingThread, out var threadType))
```
Youssef1313, line 63: I think `threadType` isn't required by all rules. I know it's unlikely that it doesn't exist, but just noting this anyway.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>CA1837: Use Environment.ProcessId.</summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class UseEnvironmentMembers : DiagnosticAnalyzer
```
Youssef1313, line 15: I'm not sure if it makes sense to extract an abstract class, and implement each rule separately?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
```diff
@@ -79,27 +107,28 @@ public static class Environment
 
 class C
 {
-    int M()
+    int HandleProcessId()
     {
-        int pid = [|Process.GetCurrentProcess().Id|];
-        pid = [|Process.GetCurrentProcess()/*willberemoved*/.Id|];
-        Use([|Process.GetCurrentProcess().Id|]);
+        int pid = {|CA1837:Process.GetCurrentProcess().Id|};
```
Youssef1313, line 112: Should we have tests for something like:

```csharp
if (Process.GetCurrentProcess() is { Id: some_value })
{
    // do something
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+                // Everything we're looking for is a property, so find all property references.
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    var initialPropRef = (IPropertyReferenceOperation)operationContext.Operation;
+
+                    // Process.GetCurrentProcess().Id
+                    if (lookForProcessId && processIdSymbol!.Equals(initialPropRef.Property))
+                    {
+                        if (initialPropRef.Instance is IInvocationOperation getCurrentProcess &&
+                            processGetCurrentProcessSymbol!.Equals(getCurrentProcess.TargetMethod))
```
Youssef1313, line 97: nit: `SymbolEqualityComparer.Default.Equals(symbol1, symbol2)` will avoid the suppression.

Youssef1313 (commented):
I think this should target .NET 6 branch (`6.0.1xx-preview3`).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
stephentoub: > I think this should use SymbolEqualityComparer.Default

Why is that better than calling Equals?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
stephentoub: It's not required, but it also exists everywhere that's relevant, so it didn't seem worthwhile to special-case.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
stephentoub: I don't think it does.  From my perspective this is straightforward and less overhead.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
Youssef1313: @stephentoub  I don't have full context to be honest. But that's what RS1024 analyzer recommends. The analyzer is currently disabled due to a crash bug, and will be re-enabled after it's fixed. If you enabled it, I believe you'll get a warning here.

Probably @sharwell or @mavasani have more context on why this is recommended.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
stephentoub: I don't think using the same diagnostic ID is a good idea here.  The target APIs were released in different versions of the libraries, and I think someone who's multitargeting will want to be able to suppress diagnostics for the APIs that don't exist in all of the platforms they're targeting while still getting diagnostics for those that do exist everywhere.  I'll need to rewatch the video to understand why we said that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
stephentoub: Happy to change it if it's the right thing to do. But would like to understand. @sharwell ?

sharwell: `SymbolEqualityComparer` allows the author to be explicit on whether `string` and `string?` should be considered the same type (`SymbolEqualityComparer.Default`) or different (`SymbolEqualityComparer.IncludeNullability`). Code written before Nullable Reference Types used `Equals` but frequently failed to account for the difference here. The analyzer is a disambiguation aid for future maintainers who would otherwise be unsure whether the analyzer was written with NRT in mind.

stephentoub: So I should use it to avoid a future analyzer yelling at me, but there's no functional difference in this case nor is there an ambiguity since none of the symbols in question are impacted by NRT. Correct?

sharwell: Yes, that is the action item here

jmarolf: as @sharwell says, using the default equals is fine is nullability isn't a part of symbol identity in your case. The analyzer warns here because we want analyzer authors to be aware of this and make it explicit which behavior they indend.

jmarolf:
@stephentoub as @Youssef1313 says, master is currently being used to service nuget packages. This PR should target https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx-preview3

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
stephentoub: We could, but I don't think it's necessary.  There are lots of different ways to express the same thing, and all we really care about right now is the one that we've actually seen devs write.

stephentoub:
Feedback addressed and rebased on release/6.0.1xx-preview3

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(UseEnvironmentProcessIdRule, UseEnvironmentProcessPathRule, UseEnvironmentCurrentManagedThreadIdRule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
```
sharwell, line 57: 💡 In general, we use `context` as the name for all context parameters. The intentional aliasing prevents the invalid use of a context object from an enclosing scope.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembersTests.cs
```diff
@@ -0,0 +1,381 @@
...
+using System.Threading;
+
+namespace System
+{
+    public static class Environment
+    {
+        public static int ProcessId => 0;
+        public static string ProcessPath => """";
+        public static int CurrentManagedThreadId => 0;
+    }
```
sharwell, line 62: ❔ What reference assemblies do these first appear in?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentProcessIdTests.cs
Youssef1313: @stephentoub I didn't mean to necessarily support this, but at least to make sure that this doesn't cause the analyzer or fixer to crash.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembersTests.cs
stephentoub: CurrentManagedThreadId:
image

ProcessId:
image

ProcessPath:
Hasn't shipped yet, so .NET 6.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
stephentoub: Okey dokey

Youssef1313:
Test failure is unrelated and was fixed in https://github.com/dotnet/roslyn/pull/51244

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
Youssef1313, line 57:
```suggestion
            context.RegisterCompilationStartAction(context =>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                // Require that Environment / Process / ProcessModule / Thread exist.
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemEnvironment, out var environmentType) ||
```
Youssef1313, line 60:
```suggestion
                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemEnvironment, out var environmentType) ||
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcess, out var processType) ||
```
Youssef1313, line 61:
```suggestion
                    !context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcess, out var processType) ||
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
Youssef1313, line 57: Was this intentionally not renamed to context, or it was just missed?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcessModule, out var processModuleType) ||
```
Youssef1313, line 62:
```suggestion
                    !context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemDiagnosticsProcessModule, out var processModuleType) ||
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+                    !compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingThread, out var threadType))
```
Youssef1313, line 63:
```suggestion
                    !context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingThread, out var threadType))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembers.cs
```diff
@@ -0,0 +1,134 @@
...
+                bool lookForProcessId = processGetCurrentProcessSymbol != null && processIdSymbol != null && environmentProcessIdSymbol != null;
+                bool lookForProcessPath = processGetCurrentProcessSymbol != null && processMainModuleSymbol != null && processModuleFileNameSymbol != null && environmentProcessPathSymbol != null;
+                bool lookForCurrentManagedThreadId = threadCurrentThreadSymbol != null && threadManagedThreadId != null && environmentCurrentManagedThreadIdSymbol != null;
+                if (!lookForProcessId && !lookForProcessPath && !lookForCurrentManagedThreadId)
+                {
+                    return;
+                }
+
+                // Everything we're looking for is a property, so find all property references.
+                compilationContext.RegisterOperationAction(context =>
```
Youssef1313, line 89:
```suggestion
                context.RegisterOperationAction(context =>
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseEnvironmentMembersTests.cs
```diff
@@ -0,0 +1,390 @@
...
+
+    void Use(int id) {}
+    void Use(string something, int id) {}
+    void Use(string something, int id, double somethingElse) { }
+}
+",
+@"
+using System;
+using System.Diagnostics;
+using System.Threading;
```
Youssef1313, line 284: The codefix *could* cleanup the now-unused usings. This is not necessary but probably good to have as a follow-up. Should I open an issue tracking this?

stephentoub: Do we do that with other analyzers? Removing usings seems orthogonal and a style preference,  not something we should be doing automatically just because we happened to remove the last use of something in that namespace. We could actually be creating more work and confusion for the developer.

stephentoub:
@mavasani, any feedback before it's merged? Thanks.

mavasani (approved):
Looks great to me!

PR #4922, opened 2021-03-03
Update DesiredFileMissing_GetFile_ReturnsNull to work with Test Explorer

sharwell:
One of these tests failed in Test Explorer without this change. Not sure why, but this fixes it.

### DIFF
--- a/src/Utilities.UnitTests/AdditionalFileProviderTests.cs
+++ b/src/Utilities.UnitTests/AdditionalFileProviderTests.cs
@@ -19,10 +19,10 @@ namespace Analyzers.Utilities.UnitTests
     public sealed class AdditionalFileProviderTests
     {
         [Theory]
-        [InlineData]
-        [InlineData("b.txt")]
-        [InlineData("a.bat")]
-        public void DesiredFileMissing_GetFile_ReturnsNull(params string[] fileNames)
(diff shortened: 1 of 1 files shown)


src/Utilities.UnitTests/AdditionalFileProviderTests.cs
```diff
@@ -19,10 +19,10 @@ namespace Analyzers.Utilities.UnitTests
...
     {
         [Theory]
-        [InlineData]
-        [InlineData("b.txt")]
-        [InlineData("a.bat")]
-        public void DesiredFileMissing_GetFile_ReturnsNull(params string[] fileNames)
+        [InlineData(new object[] { new string[] { } })]
+        [InlineData(new object[] { new string[] { "b.txt" } })]
+        [InlineData(new object[] { new string[] { "a.bat" } })]
+        public void DesiredFileMissing_GetFile_ReturnsNull(string[] fileNames)
```
Youssef1313, line 25: I'm surprised that passing object array to string array parameter works.

sharwell: The first parameter of InlineData is `params object[]`.

mavasani (approved):

PR #4926, opened 2021-03-04
Add main to yml trigger list

allisonchou:
Adds main to yml trigger list while we switch from master -> main. Hopefully keeping master in the trigger list for now avoids the build breakage seen yesterday in the roslyn repo. Once the transition is complete, master will be removed from the list and I will fix up the rest of the references in the repo.


jmarolf:
I believe we want to do this for all yml files in the repo root

jmarolf (approved):

PR #4393, opened 2020-10-30
Handle multiple analyses passes of the visitor

mavasani:
Fixes #4372

Issue #4372, opened 2020-10-27
PlatformCompatibilityAnalyzer throws System.ArgumentException


davidkarlsson:
### Analyzer

**Diagnostic ID**: [CA1416](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/CA1416): `Validate platform compatibility`

### Analyzer source

**SDK**: [Built-in CA analyzers in .NET 5 SDK or later](https://docs.microsoft.com/dotnet/fundamentals/productivity/code-analysis)

**Version**: [5.0.100-rc.2.20479.15](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug

If you are on Windows and call a method with a `[SupportedOSPlatform("windows")]` attribute in a loop that checks that you can call it with a `OperatingSystem.IsWindows()` condition a `ArgumentException` exception will be thrown with the message: `'An element with the same key but a different value already exists. Key: 'Microsoft.CodeAnalysis.Operations.FlowCaptureReferenceOperation''`. If you instead call the method outside of the loop the exception is not thrown.

### Steps To Reproduce
Here's a small repro I think:

```csharp
using System;
using System.Runtime.Versioning;

namespace AnalyzerTest
{
    class Program
    {
        static void Main(string[] args)
        {
            for (var i = 0; i < 2; i++)
            {
                if (OperatingSystem.IsWindows())
                {
                    WindowsSupported();
                }
            }
        }

        [SupportedOSPlatform("windows")]
        public static void WindowsSupported()
        {
            Console.WriteLine("Windows");
        }
    }
}
```
### Expected behavior
I didn't expect this to throw an `ArgumentException` just because the method is called from a loop.

### Actual behavior
`PlatformCompatibilityAnalyzer` throws an `ArgumentException` which get logged as a AD0001 warning on build.

## Additional context
Here's a stack trace of exception that PlatformCompatibilityAnalyzer throws in this scenario:
```
Warning	AD0001	Analyzer 'Microsoft.NetCore.Analyzers.InteropServices.PlatformCompatibilityAnalyzer' threw an exception of type 'System.ArgumentException' with message 'An element with the same key but a different value already exists. Key: 'Microsoft.CodeAnalysis.Operations.FlowCaptureReferenceOperation''.
Exception occurred with following context:
Compilation: Test
ISymbol: TestAsync (Method)

[...]

PR #4928, opened 2021-03-05
Fix the global flow state analysis merge algorithm to handle duplicat…

### LINKED ISSUE #4920: dotnet build doesn't terminate in 5.0.200, same code compiles in previous version.
**Version Used**: 5.0.200

**Steps to Reproduce**:

1.  Create an empty C# project (dotnet new console)
2.  Run ` dotnet add package Microsoft.Win32.Registry`
3.  Replace Program.cs with the following code:
```cs
using Microsoft.Win32;

if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Registry.GetValue("", "", null) != null)
{
    foreach (var x in new string[0])
    {
        if ("".ToString() == "")
        {
            try
            {
            }
            catch
            {
            }
        }
    }
}
```

**Expected Behavior**:
`dotnet build` compiles the project

**Actual Behavior**:
`dotnet build` hangs indefinitely.

mavasani:
### Related Issues:

[dotnet build doesn't terminate in 5.0.200](https://github.com/dotnet/roslyn-analyzers/issues/4920)
[C# compiler freezes in certain platform analysis scenarios](https://github.com/dotnet/roslyn/issues/51652)
 

### Context: 
The DFA for Platform Compatibility Analyzer enters an infinite loop in certain edge case code patterns where platform checks are performed inside/outside a loop and there is a try/catch within the loop.  This was untested edge case scenario was hidden under [another bug](https://github.com/dotnet/roslyn-analyzers/issues/4372) that is [fixed in 5.0.2](https://github.com/dotnet/roslyn-analyzers/pull/4393) but revealed this bug. Since 5.0.2 release 2 customers reported above issues. As this analyzer is enabled by default with a build warning level, it runs as part of every build with .NET5 SDK. We think we should fix it for 5.0.2 servicing
 
### Risk: 
Low risk. The PR adds couple of bail out checks in DFA merge algorithm which prevents infinite analysis loop. This will also likely improve the overall performance of the analysis in other cases

CC @jaredpar @jeffhandley @danmoseley @marcpopMSFT  @jmarolf

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
@@ -43,7 +43,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         for (var i = 0; i < infosBuilder.Count; i++)
                         {
                             var newValue = new GlobalFlowStateAnalysisValueSet(infosBuilder[i]);
-                            value = i == 0 ? newValue : new GlobalFlowStateAnalysisValueSet(value, newValue);
+                            value = (i == 0 || value == newValue) ? newValue : new GlobalFlowStateAnalysisValueSet(value, newValue);
                         }
 
                         return value;
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
@@ -43,7 +43,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         for (var i = 0; i < infosBuilder.Count; i++)
                         {
                             var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
-                            value = i == 0 ? newValue : GlobalFlowStateAnalysisValueSet.CreateWithParents(value, newValue);
+                            value = i == 0 ? newValue : GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
                         }
 
                         return value;
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
@@ -42,8 +42,8 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                     {
                         for (var i = 0; i < infosBuilder.Count; i++)
                         {
-                            var newValue = new GlobalFlowStateAnalysisValueSet(infosBuilder[i]);
-                            value = (i == 0 || value == newValue) ? newValue : new GlobalFlowStateAnalysisValueSet(value, newValue);
+                            var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
+                            value = i == 0 ? newValue : GlobalFlowStateAnalysisValueSet.CreateWithParents(value, newValue);
                         }
 
                         return value;
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
@@ -3466,6 +3466,36 @@ if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Registry.GetValue("""
             }.RunAsync();
         }
 
+        [Fact, WorkItem(51652, "https://github.com/dotnet/roslyn/issues/51652")]
+        public async Task TestGuardedCheckInsideLoopWithTryCatch()
+        {
+            var source = @"
+using System;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+using System.Threading.Tasks;
+
+class TestType
+{
+    static async Task Main(string[] args) {
+	while (true)
+		try {
+			await Task.Delay(1000);
+			if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+				Test();
+		} catch {
+			if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+				Test();
+			throw;
+		}
+    }
+
+    [SupportedOSPlatform(""windows"")]
+    static void Test() { }
+}";
+            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms);
+        }
+
         private string GetFormattedString(string resource, params string[] args) =>
             string.Format(CultureInfo.InvariantCulture, resource, args);
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
@@ -3426,7 +3426,7 @@ class Test
             await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms);
         }
 
-        [Fact(Skip = "https://github.com/dotnet/roslyn-analyzers/issues/4920")]
+        [Fact]
         [WorkItem(4920, "https://github.com/dotnet/roslyn-analyzers/issues/4920")]
         public async Task TestTimelyTermination()
         {
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.cs
@@ -79,12 +79,7 @@ namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.GlobalFlowStateAnalysis
                 Debug.Assert(value1.Kind == GlobalFlowStateAnalysisValueSetKind.Known);
                 Debug.Assert(value2.Kind == GlobalFlowStateAnalysisValueSetKind.Known);
 
-                if (value1 == value2)
-                {
-                    return value1;
-                }
-
-                return new GlobalFlowStateAnalysisValueSet(value1, value2);
+                return GlobalFlowStateAnalysisValueSet.CreateWithParents(value1, value2);
             }
 
             public static GlobalFlowStateAnalysisValueSet Intersect(GlobalFlowStateAnalysisValueSet value1, GlobalFlowStateAnalysisValueSet value2)
@@ -220,7 +215,7 @@ namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.GlobalFlowStateAnalysis
                     }
                     else
                     {
-                        result = new GlobalFlowStateAnalysisValueSet(sets, value1.Parents, value1.Height, GlobalFlowStateAnalysisValueSetKind.Known);
+                        result = GlobalFlowStateAnalysisValueSet.Create(sets, value1.Parents, value1.Height);
                     }
 
                     return true;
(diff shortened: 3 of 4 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -3426,7 +3426,7 @@ void MethodHasNoAttribute()
             await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms);
         }
 
-        [Fact(Skip = "https://github.com/dotnet/roslyn-analyzers/issues/4920")]
+        [Fact]
         [WorkItem(4920, "https://github.com/dotnet/roslyn-analyzers/issues/4920")]
         public async Task TestTimelyTermination()
```
Youssef1313, line 3431: Can you add a test for https://github.com/dotnet/roslyn/issues/51652 as well to make sure this PR fixes both?

mavasani: Doesn't seem to fix that second issue, its still hanging. Will continue to investigate

Youssef1313: @mavasani Should it be re-opened and moved in this repo for investigation?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
```diff
@@ -43,7 +43,7 @@ private sealed class OperationVisitor : GlobalFlowStateDataFlowOperationVisitor
                         for (var i = 0; i < infosBuilder.Count; i++)
                         {
                             var newValue = new GlobalFlowStateAnalysisValueSet(infosBuilder[i]);
-                            value = i == 0 ? newValue : new GlobalFlowStateAnalysisValueSet(value, newValue);
+                            value = (i == 0 || value == newValue) ? newValue : new GlobalFlowStateAnalysisValueSet(value, newValue);
```
sharwell, line 46: ❔ Why `==` instead of this?

```suggestion
                            value = (i == 0 || Equals(value, newValue)) ? newValue : new GlobalFlowStateAnalysisValueSet(value, newValue);
```

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
```diff
@@ -51,6 +51,7 @@ public GlobalFlowStateAnalysisValueSet(GlobalFlowStateAnalysisValueSet parent1,
                    height: Math.Max(parent1.Height, parent2.Height) + 1,
                    GlobalFlowStateAnalysisValueSetKind.Known)
         {
+            Debug.Assert(parent1 != parent2);
```
genlu, line 54: Consider replacing public constrictors with static factory method, so the duplication handling can be done in a single location.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
mavasani: Test added and now passes. I think it is fine to keep the issue closed.

mavasani:
@genlu @sharwell can you take another look? This will likely need to go through .NET 5 servicing.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
```diff
@@ -40,17 +40,49 @@ internal sealed class GlobalFlowStateAnalysisValueSet : CacheBasedEquatable<Glob
...
 
-        public GlobalFlowStateAnalysisValueSet(GlobalFlowStateAnalysisValueSet parent1, GlobalFlowStateAnalysisValueSet parent2)
-            : this(ImmutableHashSet<IAbstractAnalysisValue>.Empty,
-                   ImmutableHashSet.Create(parent1, parent2),
-                   height: Math.Max(parent1.Height, parent2.Height) + 1,
-                   GlobalFlowStateAnalysisValueSetKind.Known)
+        public static GlobalFlowStateAnalysisValueSet Create(IAbstractAnalysisValue analysisValue)
+            => new(ImmutableHashSet.Create(analysisValue), ImmutableHashSet<GlobalFlowStateAnalysisValueSet>.Empty, height: 0, GlobalFlowStateAnalysisValueSetKind.Known);
+
+        public static GlobalFlowStateAnalysisValueSet CreateWithParents(GlobalFlowStateAnalysisValueSet parent1, GlobalFlowStateAnalysisValueSet parent2)
```
sharwell, line 55: 💭 It seems like `Merge` would be a better fit for the naming conventions in this library.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
```diff
@@ -40,17 +40,49 @@ internal sealed class GlobalFlowStateAnalysisValueSet : CacheBasedEquatable<Glob
             Kind = kind;
         }
 
-        public GlobalFlowStateAnalysisValueSet(IAbstractAnalysisValue analysisValue)
-            : this(ImmutableHashSet.Create(analysisValue), ImmutableHashSet<GlobalFlowStateAnalysisValueSet>.Empty, height: 0, GlobalFlowStateAnalysisValueSetKind.Known)
+        public static GlobalFlowStateAnalysisValueSet Create(
+            ImmutableHashSet<IAbstractAnalysisValue> analysisValues,
+            ImmutableHashSet<GlobalFlowStateAnalysisValueSet> parents,
```
sharwell, line 45: 😕 Why do we have `CreateWithParents`, and also have `Create` with `parents`?

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
```diff
@@ -40,17 +40,49 @@ internal sealed class GlobalFlowStateAnalysisValueSet : CacheBasedEquatable<Glob
...
+            if (parent2.Height == 0 && parent2.AnalysisValues.IsSubsetOf(parent1.AnalysisValues))
+            {
+                return parent1;
+            }
+
+            if (parent1.Height == parent2.Height &&
+                parent1.AnalysisValues.Count == parent2.AnalysisValues.Count &&
+                Equals(parent1, parent2.GetNegatedValue()))
+            {
+                return Empty;
```
sharwell, line 79: ❔ Why this path?

sharwell:
@mavasani Can you point out the specific change which produced this regression?

mavasani:
> Can you point out the specific change which produced this regression?

I don't believe this was a regression, the bugs existed since the first implementation.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
mavasani: Ah, I wanted to rename `CreateWithParents` to `Create` but forgot to do that change, I can rename.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
mavasani: That would be misleading terminology. `Merge` is only used for standard DFA value set merge at basic blocks with multiple predecessors where incoming data sets from each control flow branch are merged. This is a different operation.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
mavasani: This was leading to unnecessary analysis iterations for the added test cases

sharwell: Please add a comment on the return statement with a description of what `parent1` and `parent2` might represent in order to return this, and why `Empty` is the right combined result.

mavasani: Sure, will add.

genlu (approved):
:shipit:

mavasani:
@sharwell PR feedback addressed. @buyaa-n, I can port the PR to release/5.0.2xx once this is merged. Please let me know if you need anything else apart from this from me.

buyaa-n:
@mavasani just confirmed with @jmarolf, or servicing it should always go into a release branch could you retarget it to `release/5.0.2x` ? I will ask for servicing approval

mavasani:
@buyaa-n I plan to first check in the change to master and just cherry-pick it to release/5.0.2xx with a separate PR. Primarily because the test added by Sam for this bug doesn't compile on release/5.0.2xx branch and it seems non-trivial to enable the functionality required by it. So, I plan to only cherry-pick the product changes + the second test added in my second commit to 5.0.2xx branch.

buyaa-n:
> @buyaa-n I plan to first check in the change to master and just cherry-pick it to release/5.0.2xx with a separate PR. Primarily because the test added by Sam for this bug doesn't compile on release/5.0.2xx branch and it seems non-trivial to enable the functionality required by it. So, I plan to only cherry-pick the product changes + the second test added in my second commit to 5.0.2xx branch.

Ah sorry, didn't think about that, makes sense

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.cs
```diff
@@ -79,7 +80,36 @@ public override GlobalFlowStateAnalysisValueSet Merge(GlobalFlowStateAnalysisVal
...
+                if (value2.Height == 0 && value2.AnalysisValues.IsSubsetOf(value1.AnalysisValues))
+                {
+                    return value1;
+                }
+
+                // Check if value1 and value2 are negations of each other.
+                // If so, the analysis values nullify each other and we return an empty set.
+                if (value1.Height == value2.Height &&
+                    value1.AnalysisValues.Count == value2.AnalysisValues.Count &&
+                    Equals(value1, value2.GetNegatedValue()))
```
sharwell, line 103: 💡 Can you file a follow-up issue to investigate using a negating comparer here to avoid constructing a negated value every time?

jeffhandley (approved):

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysisValueSet.cs
```diff
@@ -40,18 +40,17 @@ internal sealed class GlobalFlowStateAnalysisValueSet : CacheBasedEquatable<Glob
             Kind = kind;
         }
 
-        public GlobalFlowStateAnalysisValueSet(IAbstractAnalysisValue analysisValue)
-            : this(ImmutableHashSet.Create(analysisValue), ImmutableHashSet<GlobalFlowStateAnalysisValueSet>.Empty, height: 0, GlobalFlowStateAnalysisValueSetKind.Known)
+        public static GlobalFlowStateAnalysisValueSet Create(
```
sharwell, line 43: 💡 This might as well be a constructor at this point

sharwell (approved):
Looking at the values coming from `Intersect`, I would not be surprised if we need more fixes in the future. These seem fine for now though.

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/GlobalFlowStateAnalysis/GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.cs
mavasani: Sure will do

PR #4942, opened 2021-03-11
Delete dead links

Youssef1313:
There are very few dead links of the deleted FxCopAnalyzers solution. This GH Action will fail on them. I'll then fix them in the second commit. Let me know if you want to have the action or only the link fixes.


.github/workflows/markdown-links-verifier.yml
```diff
@@ -0,0 +1,14 @@
...
+  validate_links:
+    name: Markdown links verifier
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout the repository
+      uses: actions/checkout@v1
+
+    - name: Validate links
+      uses: Youssef1313/markdown-links-verifier@v0.0.5
```
Youssef1313, line 14: I'm not sure of your policy of using such GH Actions. Let me know if this shouldn't be used or if it requires any action to be used.

Youssef1313:
Well, actually I forgot that GH Action isn't enabled in this repo. I'm going to revert this commit and only fix the links

mavasani (approved):

Issue #4932, opened 2021-03-08
How to fix CA1416 for net5-windows10.0.17763.0 ?


bgavrilMS:
### Analyzer

**Diagnostic ID**:  CA1416 

### Analyzer source

**SDK**: [Built-in CA analyzers in .NET 5 SDK or later](https://docs.microsoft.com/dotnet/fundamentals/productivity/code-analysis)

**Version**: 5.0.200-preview.21077.7 

### Describe the bug

In my library I have
```csharp
    public class MSAL
    {
        [SupportedOSPlatform("windows10.0.17763.0")]
        public static void UseWAM()
        {
            WebTokenRequest webTokenRequest = new WebTokenRequest(null);
        }
    }
```
A calling app does :

```csharp
    class Program
    {
        static void Main(string[] args)
        {
            if (OperatingSystem.IsWindowsVersionAtLeast(10, 0, 17763, 0))
            {
                MSAL.UseWAM(); // gets a warning despite the platform guard
            }
       }
   }
```

### Expected behavior
no CA1416 warning in app

### Actual behavior
warning

## Additional context

If I change it to [SupportedOSPlatform("windows")] and OperatingSystem.IsWindows, the warning goes away.

buyaa-n:
hm, `OperatingSystem.IsWindowsVersionAtLeast(10, 0, 17763, 0)` is evaluated to version `10.0.17763` the comparison operators of the Version type is a bit strange
```
Console.WriteLine(Version.Parse("10.0.17763.0") < Version.Parse("10.0.17763")); // False
Console.WriteLine(Version.Parse("10.0.17763.0") > Version.Parse("10.0.17763")); // True
Console.WriteLine(Version.Parse("10.0.17763.0") == Version.Parse("10.0.17763")); // False
```
as a workaround @bgavrilMS could you remove leading `0` from the attribute ([SupportedOSPlatform("windows10.0.17763")]) until it gets resolved?

buyaa-n:
Fixed with https://github.com/dotnet/roslyn-analyzers/pull/4943

PR #4943, opened 2021-03-11
CA1416 Fix Version comparison ambiguity

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4932
The comparison operations of the Version type is a bit strange:
```
Console.WriteLine(Version.Parse("10.0.17763.0") < Version.Parse("10.0.17763")); // False
Console.WriteLine(Version.Parse("10.0.17763.0") > Version.Parse("10.0.17763")); // True
Console.WriteLine(Version.Parse("10.0.17763.0") == Version.Parse("10.0.17763")); // False
```
Because of that now `[SupportedOSPlatform("windows10.0.17763.0")]` is not equal to `[SupportedOSPlatform("windows10.0.17763")]`  because `Version.Revision` will be defaulted to `-1`.
Accordingly API with `[SupportedOSPlatform("windows10.0.17763.0")]` attribute  cannot be guarded with `OperatingSystem.IsWindowsVersionAtLeast(10, 0, 17763)`

Added extension method for version comparison

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -406,7 +406,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                             if (info.Negated)
                             {
                                 if (attribute.UnsupportedFirst != null &&
-                                    attribute.UnsupportedFirst >= info.Version)
+                                    attribute.UnsupportedFirst.IsGreaterThanOrEqualTo(info.Version))
                                 {
                                     if (DenyList(attribute))
                                     {
@@ -418,7 +418,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                 }
 
                                 if (attribute.UnsupportedSecond != null &&
-                                    attribute.UnsupportedSecond <= info.Version)
+                                    info.Version.IsGreaterThanOrEqualTo(attribute.UnsupportedSecond))
                                 {
                                     attribute.UnsupportedSecond = null;
                                 }
@@ -434,21 +434,21 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                 {
                                     if (attribute.UnsupportedFirst != null &&
                                         capturedVersions.TryGetValue(info.PlatformName, out var version) &&
-                                        attribute.UnsupportedFirst >= version)
+                                        attribute.UnsupportedFirst.IsGreaterThanOrEqualTo(version))
                                     {
                                         attribute.UnsupportedFirst = null;
                                     }
 
                                     if (attribute.UnsupportedSecond != null &&
                                         capturedVersions.TryGetValue(info.PlatformName, out version) &&
-                                        attribute.UnsupportedSecond <= version)
+                                        version.IsGreaterThanOrEqualTo(attribute.UnsupportedSecond))
                                     {
                                         attribute.UnsupportedSecond = null;
                                     }
                                 }
 
                                 if (attribute.SupportedFirst != null &&
-                                    attribute.SupportedFirst <= info.Version)
+                                    info.Version.IsGreaterThanOrEqualTo(attribute.SupportedFirst))
                                 {
                                     attribute.SupportedFirst = null;
                                     RemoveUnsupportedWithLessVersion(info.Version, attribute);
@@ -456,7 +456,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                 }
 
                                 if (attribute.SupportedSecond != null &&
-                                    attribute.SupportedSecond <= info.Version)
+                                    info.Version.IsGreaterThanOrEqualTo(attribute.SupportedSecond))
                                 {
                                     attribute.SupportedSecond = null;
                                     RemoveUnsupportedWithLessVersion(info.Version, attribute);
@@ -549,8 +549,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
 
             static void RemoveUnsupportedWithLessVersion(Version supportedVersion, Versions attribute)
             {
-                if (attribute.UnsupportedFirst != null &&
-                    attribute.UnsupportedFirst <= supportedVersion)
+                if (supportedVersion.IsGreaterThanOrEqualTo(attribute.UnsupportedFirst))
                 {
                     attribute.UnsupportedFirst = null;
                 }
@@ -887,14 +886,14 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         var supportedVersion = attribute.SupportedSecond ?? attribute.SupportedFirst;
                         if (supportedVersion != null)
                         {
-                            version = version == null || supportedVersion >= version ? supportedVersion : version;
+                            version = supportedVersion.IsGreaterThanOrEqualTo(version) ? supportedVersion : version;
                         }
                         else
                         {
                             var unsupportedVersion = attribute.UnsupportedSecond ?? attribute.UnsupportedFirst;
                             if (unsupportedVersion != null)
                             {
-                                version = version == null || unsupportedVersion >= version ? unsupportedVersion : version;
+                                version = unsupportedVersion.IsGreaterThanOrEqualTo(version) ? unsupportedVersion : version;
                             }
                             else
                             {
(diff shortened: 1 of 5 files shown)


Youssef1313 (commented):
There are still few comparisons left:

https://github.com/dotnet/roslyn-analyzers/blob/6628ba691c886657f6267a9956e16e1f82cbf4d3/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L724

https://github.com/dotnet/roslyn-analyzers/blob/6628ba691c886657f6267a9956e16e1f82cbf4d3/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L727

https://github.com/dotnet/roslyn-analyzers/blob/6628ba691c886657f6267a9956e16e1f82cbf4d3/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L801

https://github.com/dotnet/roslyn-analyzers/blob/6628ba691c886657f6267a9956e16e1f82cbf4d3/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L823

https://github.com/dotnet/roslyn-analyzers/blob/6628ba691c886657f6267a9956e16e1f82cbf4d3/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L1159

(and few more)


Was that intended?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1324,25 +1323,25 @@ private static PlatformAttributes CopyAttributes(PlatformAttributes copyAttribut
 
             static bool SuppressedByCallSiteUnsupported(Versions callSiteAttribute, Version unsupporteAttribute) =>
                 DenyList(callSiteAttribute) && callSiteAttribute.SupportedFirst != null ?
-                callSiteAttribute.UnsupportedSecond != null && unsupporteAttribute >= callSiteAttribute.UnsupportedSecond :
-                callSiteAttribute.UnsupportedFirst != null && unsupporteAttribute >= callSiteAttribute.UnsupportedFirst;
+                callSiteAttribute.UnsupportedSecond != null && unsupporteAttribute.IsGreaterThanOrEqualTo(callSiteAttribute.UnsupportedSecond) :
```
Youssef1313, line 1326: Is null check now redundant here?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -549,8 +549,7 @@ static void RemoveUnsupportsOnDifferentPlatforms(SmallDictionary<string, Version
 
             static void RemoveUnsupportedWithLessVersion(Version supportedVersion, Versions attribute)
             {
-                if (attribute.UnsupportedFirst != null &&
-                    attribute.UnsupportedFirst <= supportedVersion)
+                if (supportedVersion.IsGreaterThanOrEqualTo(attribute.UnsupportedFirst))
                 {
                     attribute.UnsupportedFirst = null;
                 }
```
Youssef1313, line 555: Behavior when UnsupportedFirst is null:

- Old code, condition not met.
- New code, condition is met.

Should the null check stays?

Youssef1313 (commented):
Is there a way to ban the comparison operators for `Version`?

buyaa-n:
> There are still few comparisons left:
> ...
> Was that intended?

Yes the comparisons with `>` or `<` has no issue, only related to `>=` or `<=`

> Is there a way to ban the comparison operators for Version?

That should be runtime related question, anyway, i don't think so because I heard there was a long discussion (idk when/where though) about the Version comparisons and if this behavior is correct or should be fixed etc, and ended up not fixing. I guess it didn't fix it because the existing behavior was made sense for some other scenarios

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: It is not redundant for the `IsGreaterThanOrEqualTo` it is optional but in this exact case it is required to be non null

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: Here the null check was not required (added that because it was causing NRE)

Youssef1313:
> That should be runtime related question

I meant using the BannedApiAnalyzer to warn if we use `>=` or `<=` for `Version`. If there is (currently) no way for BannedApiAnalyzer to ban operators, it may be a reasonable feature request.

@mavasani @Evangelink Do you know whether BannedApiAnalyzer supports this?

mavasani:
> Do you know whether BannedApiAnalyzer supports this?

Yes, we can ban usage of specific APIs in NetAnalyzers projects by adding the documentation comment IDs of the APIs to https://github.com/dotnet/roslyn-analyzers/blob/dc515b0412abf13d9d2b8e34543c61c9abbee87b/src/NetAnalyzers/BannedSymbols.txt

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -120,19 +120,17 @@ void M1()
...
     }
 }
 ";
             await VerifyAnalyzerAsyncCs(csSource, s_msBuildPlatforms,
-#pragma warning disable RS0030 // Do not used banned APIs
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.OnlySupportedCsAllPlatforms).WithLocation(24, 13).WithArguments("BrowserOnlyType", "'browser'"),
-#pragma warning restore RS0030 // Do not used banned APIs
-#pragma warning disable RS0030 // Do not used banned APIs
-                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.OnlySupportedCsAllPlatforms).WithLocation(25, 13).WithArguments("WindowsOnlyType", "'windows'"));
-#pragma warning restore RS0030 // Do not used banned APIs
```
carlossanlop, line 135: The 'browser' diagnostic was only showing up for the first invocation, and the 'windows' diagnostic was only showing up for the second invocation?

The fix makes sense (both diagnostics should show up for both invocations) but I am not understanding why is it related to the problem this PR is trying to address.

carlossanlop (commented):
@buyaa-n I left a couple of questions.

Can you please address @Youssef1313 's [question](https://github.com/dotnet/roslyn-analyzers/pull/4943#pullrequestreview-610741144) about the other comparisons that were not modified?

Also, let's try to explore the suggestion to ban `<=` and `>=` for `Version` instances. Maybe we can find a way to ban them only inside the `PlatformCompatibilityAnalyzer` class.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: > The fix makes sense (both diagnostics should show up for both invocations) but I am not understanding why is it related to the problem this PR is trying to address.

This doesn't relate to the problem, as you can see `WithLocation(int, int)` is banned, previously i have had updated all such usages with `WithLocation(int)` but somehow this one left, i guess it is caused by many different branches merge conflicts. Anyway, this change just to replace the deprecated API usage and remove the suppression, no any functional/test change here

src/Utilities/Compiler/Extensions/VersionExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+                return current.Major > compare.Major;
+            }
+
+            if (current.Minor != compare.Minor)
+            {
+                return current.Minor > compare.Minor;
+            }
+
+            // For build or revision value of 0 equals to -1
+            if (current.Build != compare.Build && (current.Build > 0 || compare.Build > 0))
```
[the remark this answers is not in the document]
buyaa-n: It mean we only accept/allow inequality for values `> 0` (which makes sure 0 and-1 treated as equal)

buyaa-n:
> Can you please address @Youssef1313 's question about the other comparisons that were not modified?

That question is answered [here](https://github.com/dotnet/roslyn-analyzers/pull/4943#issuecomment-799573911), in short, `<` or `>` comparisons unrelated, only `>=` or `<=` related to this change

> Also, let's try to explore the suggestion to ban <= and >= for Version instances. Maybe we can find a way to ban them only inside the PlatformCompatibilityAnalyzer class.

From what I have seen and what @mavasani answered seems we cannot do that for only `PlatformCompatibilityAnalyzer`, correct me if I am wrong, i am in doubt if we should do that for the entire NetAnalyzers project, existing Version comparisons might be useful for other analyzers

carlossanlop (approved):
LGTM. All questions answered.

buyaa-n:
Followed up with @mavasani, we cannot ban version comparison only for PlatformCompatibilityAnalyzer, i don't think we should ban Version comparison in the entire project, so gonna merge the PR

PR #4952, opened 2021-03-15
OSX is alias for MacOS

buyaa-n:
Related to https://github.com/dotnet/runtime/issues/49323

While testing the Known Platform names analyzer in runtime repo we discovered that runtime repo is using OSX in TFM which is  alias (old name) of MacOS. As OSX can be used instead if MacOS we should add it as known platform in case MacOS exist in the known platforms list

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
@@ -21,6 +21,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         private static readonly ImmutableArray<string> methodNames = ImmutableArray.Create("IsOSPlatform", "IsOSPlatformVersionAtLeast");
         private const string IsPrefix = "Is";
         private const string VersionSuffix = "VersionAtLeast";
+        private const string macOS = nameof(macOS);
 
         private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseValidPlatformStringTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
         private static readonly LocalizableString s_localizableUnknownPlatform = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseValidPlatformStringUnknownPlatform), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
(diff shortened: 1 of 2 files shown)


jeffhandley (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
```diff
@@ -21,6 +21,7 @@ public sealed class UseValidPlatformString : DiagnosticAnalyzer
         private static readonly ImmutableArray<string> methodNames = ImmutableArray.Create("IsOSPlatform", "IsOSPlatformVersionAtLeast");
         private const string IsPrefix = "Is";
         private const string VersionSuffix = "VersionAtLeast";
+        private const string macOS = nameof(macOS);
```
carlossanlop, line 24: Is there a place somewhere in the roslyn-analyzers repo where we have all the platform names predefined? Seems strange to have only this one defined here.

Also, wouldn't it be better to use a `const string` instead of a `nameof` that points to its own name? This pattern is a bit strange.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
```diff
@@ -376,6 +376,31 @@ public class Test
...
+
+    [SupportedOSPlatform(""MacOS1.2"")]
+    public void SupportedMacOs2PartValid() { }
+
+    [{|#1:SupportedOSPlatform(""OSX1.2.3.4"")|}] // Version '1.2.3.4' is not valid for platform 'OSX'. Use a version with 2-3 parts for this platform.
+    public void SupportedOSPlatformOSX4PartsInvalid() { }
+
+    [SupportedOSPlatform(""Osx1.2"")]
+    public void SupportedOsx2PartValid() { }
+}";
```
carlossanlop, line 398: What would happen if you add a method topped with two `[SupportedOSPlatform]` attributes, one with "OSX" and another one with "MacOS"? Maybe we should have a test for that edge case too.

carlossanlop (approved):
LGTM. Left a couple of non-blocking comments. Consider adding the extra unit test I suggested.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformString.cs
buyaa-n: > Is there a place somewhere in the roslyn-analyzers repo where we have all the platform names predefined? Seems strange to have only this one defined here.

Do not know, anyway for the analyzer the known names loaded from MSBuild known platforms list

> Also, wouldn't it be better to use a const string instead of a nameof that points to its own name? This pattern is a bit strange.

It can, but i wouldn't say that would be better. In this case, as you can see the variable name is the same as the string literal (no need to be different or no need to be more descriptive), in fact, it is not just a random string literal, it illustrates that it is a known OS platform name from MSBuild list used with the exact same value and casing from the list.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/UseValidPlatformStringTest.cs
buyaa-n: Nothing should happen, the analyzer should not distinguish them as same nor should require to remove one of them.

I can add a test to show nothing happening

PR #4955, opened 2021-03-16
Handle conditional access in PreferIsKind analyzer

### LINKED ISSUE #4946: Prefer 'IsKind' should handle null conditional
### Analyzer

RS1034 (Prefer 'IsKind' for checking syntax kinds)

### Describe the improvement

Currently the following pattern is not recognized by the analyzer:

```
node?.Kind() == SyntaxKind.Kind
```

This pattern should be reported by the analyzer, and converted to this:

```
node.IsKind(SyntaxKind.Kind)
```

Youssef1313:
Fixes #4946

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
@@ -36,6 +36,8 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
 
             context.RegisterCompilationStartAction(context =>
             {
+                var nullableOfT = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNullable1);
+
                 // Kind() methods
                 //  Microsoft.CodeAnalysis.CSharp.CSharpExtensions.Kind
                 //  Microsoft.CodeAnalysis.VisualBasic.VisualBasicExtensions.Kind
@@ -59,12 +61,12 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
 
                 if (containingTypeMap.Count > 0)
                 {
-                    context.RegisterOperationAction(context => HandleBinaryOperation(context, containingTypeMap), OperationKind.Binary);
+                    context.RegisterOperationAction(context => HandleBinaryOperation(context, containingTypeMap, nullableOfT), OperationKind.Binary);
                 }
             });
         }
 
-        private static void HandleBinaryOperation(OperationAnalysisContext context, Dictionary<INamedTypeSymbol, INamedTypeSymbol> containingTypeMap)
+        private static void HandleBinaryOperation(OperationAnalysisContext context, Dictionary<INamedTypeSymbol, INamedTypeSymbol> containingTypeMap, INamedTypeSymbol? nullableOfT)
         {
             var operation = (IBinaryOperation)context.Operation;
             if (operation.OperatorKind is not (BinaryOperatorKind.Equals or BinaryOperatorKind.NotEquals))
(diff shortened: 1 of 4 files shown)


src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
```diff
@@ -419,5 +419,63 @@ End Class
...
+        [Fact]
+        public async Task TestSingleNullConditionalAccess_CS()
+        {
+            var source =
+@"using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+class C
+{
+    bool Method(SyntaxNode node)
```
sharwell, line 432: 💡 Can you add a test for `SyntaxToken?` as well?

Youssef1313: @sharwell The codefix (currently) produces:

```csharp
    bool Method(SyntaxToken? token)
    {
        return token.IsKind(SyntaxKind.None); // <--  error CS1929: 'SyntaxToken?' does not contain a definition for 'IsKind' and the best extension method overload 'CSharpExtensions.IsKind(SyntaxToken, SyntaxKind)' requires a receiver of type 'SyntaxToken'
    }
```

Going to look how I'll fix that.

Youssef1313: I found it easier to not report a diagnostic rather than complicating the logic inside the codefix. The usage of `Nullable<SyntaxToken>` is a little bit rare. But let me know if you want me to go the other route and make the codefix behave correctly.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -75,6 +77,12 @@ private static void HandleBinaryOperation(OperationAnalysisContext context, Dict
             var possibleInvocation = operation.LeftOperand.WalkDownConversion();
             if (possibleInvocation is IConditionalAccessOperation conditionalAccess)
             {
+                // We don't currently report on Nullable<SyntaxToken>. If we'll report that in the future, the codefix must behave correctly.
+                if (SymbolEqualityComparer.Default.Equals(conditionalAccess.Operation.Type.OriginalDefinition, nullableOfT))
```
Youssef1313, line 81: IOperation APIs are not fully nullable annotated. I'm not sure if `Operation` or `Type` can ever be null or not.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/PreferIsKindAnalyzerTests.cs
sharwell: Skipping it is what I expected. Totally fine.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -36,6 +36,8 @@ public override void Initialize(AnalysisContext context)
 
             context.RegisterCompilationStartAction(context =>
             {
+                var nullableOfT = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNullable1);
```
sharwell, line 39: 💡 Use `GetSpecialType`?

```suggestion
                var nullableOfT = context.Compilation.GetSpecialType(SpecialType.System_Nullable_T);
```

sharwell (approved):

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
Youssef1313: @sharwell Probably not even needed? Could check directly for:

```csharp
conditionalAccess.Operation.Type.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T
```

instead of retrieving an INamedTypeSymbol and using SymbolEqualityComparer.

sharwell: Nice 👍

PR #4959, opened 2021-03-19
Don't report RS1024 on non-symbol GetHashCode call

### LINKED ISSUE #4956: Incorrect RS1024 analysis when calling GetHashCode(ReadOnlySpan<char>, StringComparison)
### Analyzer

RS1024

### Analyzer source

**Version**: 3.10.0-2.21166.52

### Steps To Reproduce

1. Reference the 3.10.0-2.21166.52 of Roslyn binaries in a console app
2. Code:
```C#
ReadOnlySpan<char> foo = default;
int hashCode = string.GetHashCode(foo, StringComparison.OrdinalIgnoreCase);
```

### Actual behavior

```
error RS1024: Compare symbols correctly
```

Youssef1313:
Fixes #4956

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
@@ -149,14 +149,14 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
             switch (method.Name)
             {
                 case WellKnownMemberNames.ObjectGetHashCode:
-                    if (IsNotInstanceInvocationOrNotOnSymbol(invocationOperation, symbolType))
+                    if (IsSymbolInstance(invocationOperation, symbolType))
                     {
                         context.ReportDiagnostic(invocationOperation.CreateDiagnostic(GetHashCodeRule));
                     }
                     break;
 
                 case s_symbolEqualsName:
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -180,8 +180,8 @@ private static void HandleBinaryOperator(in OperationAnalysisContext context, IN
                     break;
             }
 
-            static bool IsNotInstanceInvocationOrNotOnSymbol(IInvocationOperation invocationOperation, INamedTypeSymbol symbolType)
-                => invocationOperation.Instance is null || IsSymbolType(invocationOperation.Instance, symbolType);
+            static bool IsSymbolInstance(IInvocationOperation invocationOperation, INamedTypeSymbol symbolType)
+                => invocationOperation.Instance is not null && IsSymbolType(invocationOperation.Instance, symbolType);
```
sharwell, line 184: 💭 I believe the old method is the one we want for `Equals`, while the new one is the one we want for `GetHashCode`.

Youssef1313: > I believe the old method is the one we want for Equals

Won't old method will always return true for the static `object.Equals(obj1, obj2)`, which isn't desired?

Youssef1313:
Closing as this is being fixed in #5057

PR #4965, opened 2021-03-21
Additional source build cleanup

sharwell:
This change removes most remaining build customization performed as part of the Source Build work.


eng/GenerateAnalyzerNuspec.targets
```diff
@@ -86,8 +86,16 @@
...
     </PropertyGroup>
-    
-    <MSBuild Projects="$(RepoRoot)src\Tools\GenerateDocumentationAndConfigFiles\GenerateDocumentationAndConfigFiles.csproj" Targets="Build">
+
+    <PropertyGroup>
+      <!--
+        Source build's runtime environment is incomplete, so we call through a stub project to force assembly resolution
+        to choose versions from the available subset.
+      -->
+      <GenerateDocumentationAndConfigFilesProjectSuffix Condition="'$(DotNetBuildFromSource)' == 'true'">ForBrokenRuntime</GenerateDocumentationAndConfigFilesProjectSuffix>
```
jmarolf, line 95: Is `ForBrokenRuntime` the best description of this? Perhaps we want to indicate that this is related to source build? Also, what is broken? Can we file a bug?

eng/Versions.props
```diff
@@ -52,4 +52,15 @@
...
+
+  <PropertyGroup>
+    <!--
+      Binaries that need to be executable during source build are restricted to versions available within source build.
+      This section defines executable versions of packages which are referenced via lower-version reference assemblies
+      during the build.
+    -->
+    <SystemCollectionsImmutableExecutableVersion>5.0.0</SystemCollectionsImmutableExecutableVersion>
+    <SystemReflectionMetadataExecutableVersion>5.0.0</SystemReflectionMetadataExecutableVersion>
+    <MicrosoftCodeAnalysisExecutableVersion>3.8.0</MicrosoftCodeAnalysisExecutableVersion>
```
jmarolf, line 64: Is there a way we can have arcade auto-update these dependencies so we don't need to be responsible for updating them? Logically these _should_ be whatever compiler is being produced as part of source build right?

jmarolf (approved):

eng/Versions.props
sharwell: Arcade should not be responsible for this. If this is a necessary concept for tooling, it should be pushed all the way up to dotnet CLI and/or nuget to allow execution of a project using a different (newer) set of dependencies than the build used.

eng/GenerateAnalyzerNuspec.targets
sharwell: I already renamed this once. It's not inaccurate, but perhaps could be better.

> Also, what is broken?

Source build omits some stable public releases of some dependencies in violation of the backwards compatibility expectations of those projects. From the perspective of those projects, this is a broken runtime design.

> Can we file a bug?

We should file a feature request for allowing a different set of build/runtime dependencies without creating a new project: https://github.com/dotnet/roslyn-analyzers/pull/4965#discussion_r598843475

PR #4967, opened 2021-03-22
Remove unused build properties

sharwell:
This independent change was extracted from features/source-build.


jmarolf (approved):

jmarolf:
Odd, thanks for removing these we really don't want to import VSSDK stuff if we can avoid it.

sharwell:
@jmarolf Note that these are _unused_ properties (as opposed to _unnecessary_). Binlogs show that there are no uses of `$(ImportVSSDKTargets)`.

PR #4971, opened 2021-03-22
Merge features/source-build to main

sharwell:
@omajid This is now ready for review.

* All patches and the patching infrastructure have been removed.
* Exclusion of projects from source build is achieved through `ExcludeFromSourceBuild`.
* All projects build against the original set of references. In one case, a new project was added which effectively allows a tool to be published/run using a newer set of dependencies than it was built against.


jmarolf (approved):

MichaelSimons (approved):

omajid (approved):
Thanks!

Issue #4870, opened 2021-02-19
CA2016: Code fix removes generic type parameter from message


DavidBoike:
### Analyzer

**Diagnostic ID**: [CA2016](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2016): Forward the CancellationToken parameter to methods that take one

### Analyzer source

**SDK**: [Built-in CA analyzers in .NET 5 SDK or later](https://docs.microsoft.com/dotnet/fundamentals/productivity/code-analysis)

**Version**: [SDK 5.0.103](https://dotnet.microsoft.com/download/dotnet/5.0)

### Describe the bug

Consider the following code snippet, which will highlight CA2016 on the `reader.GetFieldValueAsync` method:

```
static async Task<Guid> Foo(SqlDataReader reader, CancellationToken cancellationToken)
{
    return await reader.GetFieldValueAsync<Guid>(0)
        .ConfigureAwait(false);
}
```

Using the fixer will result in this code, with the generic type `<Guid>` stripped from the method call:

```
static async Task<Guid> Foo(SqlDataReader reader, CancellationToken cancellationToken)
{
    return await reader.GetFieldValueAsync(0, cancellationToken)
        .ConfigureAwait(false);
}
```

Because of the removal of the generic type from the method call, the code now has this compile error:

> CS0411 The type arguments for method 'SqlDataReader.GetFieldValueAsync<T>(int, CancellationToken)' cannot be inferred from the usage. Try specifying the type arguments explicitly.

### Steps To Reproduce

1. Create a project targeting `net5.0`.
1. Add a package reference to `Microsoft.Data.SqlClient`.
1. Add the code above.
1. Use the code fix on `reader.GetFieldValueAsync`.
1. See error.

### Expected behavior

Should add the cancellation token to the method arguments without modifying the type argument.

### Actual behavior

Type argument is removed.

eskye:
I will like to work on this issue

mavasani:
@eskye Feel free to pick it up. @carlossanlop can guide you if you need any help.

eskye:
> @eskye Feel free to pick it up. @carlossanlop can guide you if you need any help.

Okay, thanks. I think I'll surely need a guide @carlossanlop

eskye:
@carlossanlop Please I need help.

DavidBoike:
[...]

PR #4974, opened 2021-03-23
Fix CA2016 when used on instance method with generic types

DavidBoike:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4870

* Adds a failing test case for the problem matching the description of https://github.com/dotnet/roslyn-analyzers/issues/4870
* Adds a test case for a static generic method, which was green before applying the fix, to guard against regressions in that behavior

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Fixer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Fixer.cs
@@ -6,7 +6,6 @@ using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using Analyzer.Utilities;
-using Analyzer.Utilities.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.Editing;
@@ -104,7 +103,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 context.Diagnostics);
         }
 
-        private SyntaxNode TryGenerateNewDocumentRoot(
+        private static SyntaxNode TryGenerateNewDocumentRoot(
             Document doc,
             SyntaxNode root,
             IInvocationOperation invocation,
@@ -128,30 +127,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             newArguments = newArguments.Add(cancellationTokenArgument);
 
-            SyntaxNode newInstance;
-            // The instance is null when calling a static method from another type
-            switch (invocation.Instance)
-            {
-                case null:
-                    newInstance = expression;
-                    break;
-                case IConditionalAccessInstanceOperation:
-                    newInstance = GetConditionalOperationInvocationExpression(invocation.Syntax);
-                    break;
-                default:
-                    if (invocation.Instance.IsImplicit)
-                    {
-                        newInstance = invocation.GetInstanceSyntax()!;
-                    }
-                    // Calling a method from an object, we must include the instance variable name
-                    else
-                    {
-                        newInstance = generator.MemberAccessExpression(invocation.GetInstanceSyntax(), invocation.TargetMethod.Name);
-                    }
-                    break;
-            }
             // Insert the new arguments to the new invocation
-            SyntaxNode newInvocationWithArguments = generator.InvocationExpression(newInstance, newArguments).WithTriviaFrom(invocation.Syntax);
+            SyntaxNode newInvocationWithArguments = generator.InvocationExpression(expression, newArguments).WithTriviaFrom(invocation.Syntax);
 
             return generator.ReplaceNode(root, invocation.Syntax, newInvocationWithArguments);
         }
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
@@ -2250,6 +2250,156 @@ class C
         }
 
         [Fact]
+        [WorkItem(4870, "https://github.com/dotnet/roslyn-analyzers/issues/4870")]
+        public Task CS_Diagnostic_GenericTypeParamOnInstanceMethod()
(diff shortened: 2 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Fixer.cs
```diff
@@ -128,30 +127,8 @@ Task<Document> CreateChangedDocumentAsync(CancellationToken _)
...
-                    newInstance = expression;
-                    break;
-                case IConditionalAccessInstanceOperation:
-                    newInstance = GetConditionalOperationInvocationExpression(invocation.Syntax);
-                    break;
-                default:
-                    if (invocation.Instance.IsImplicit)
-                    {
-                        newInstance = invocation.GetInstanceSyntax()!;
-                    }
```
carlossanlop, line 145: > _Adds a test case for a static generic method, which was green before applying the fix, to guard against regressions in that behavior_

The CI is still passing because apparently I forgot to add unit tests for all three cases you removed.

- It seems we don't have a conditional operation case in the unit tests. Can you please add one?
- I'm trying to remember what `IsImplicit` is referring to. It's either:
    - An invocation of a method from an explicit interface implementation, or
    - An invocation of a method passed as a callback to another method (a `Func` or an `Action`).
  Independently if it's one or the other, we don't have unit tests for any of them. Would you mind adding one to make sure that functionality is not regressed?

Thanks for adding a method for the static case.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -2229,6 +2229,78 @@ void LocalMethod()
             return CS8VerifyCodeFixAsync(originalCode, fixedCode);
         }
 
+        [Fact]
+        [WorkItem(4870, "https://github.com/dotnet/roslyn-analyzers/issues/4870")]
+        public Task CS_Diagnostic_GenericTypeParamOnInstanceMethod()
```
carlossanlop, line 2234: Can you please also add VB unit tests?

carlossanlop (commented):
Thank you for the change, @DavidBoike . I left a couple of requests.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Fixer.cs
DavidBoike: From the [IConditionalAccessInstanceOperation Interface docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.operations.iconditionalaccessinstanceoperation?view=roslyn-dotnet) it sounds like you're referring to a call like `someExpression?.Method()`. Isn't that covered by [CS_Diagnostic_DereferencePossibleNullReference](https://github.com/dotnet/roslyn-analyzers/blob/main/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs#L1974-L2018)?

`IsImplicit` seems to refer to the "this" object passed to the method, either the instance or null if it's a static method. In any case, when I tried adding this block to the fixer:

```csharp
if (invocation.Instance != null && invocation.Instance.IsImplicit)
{
    throw new Exception("Implicit");
}
```

…the result was that 73 tests failed. First one was [CS_Diagnostic_ActionDelegateAwait](https://github.com/dotnet/roslyn-analyzers/blob/main/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs#L1146-L1180) where `MethodAsync()` is being called without `this.`, which seems to track. So I think it's safe to say that test case is covered.

While the removal of that code looks scary, I think it makes sense philisophically. The fixer is only trying to adjust what's going on in the ArgumentList. So trying to rebuild the part of the invocation "outside" the parens from scratch based on a number of conditions doesn't really make sense when `expression` is already going to contain whatever that outside part was originally highlighted by the red squiggly.

To illustrate that point further, I added 2 additional tests `CS_Diagnostic_NullCoalescedDelegates` and `CS_Diagnostic_NullCoalescedDelegatesWithInvoke`, both of which have 2 delegates and call either `(f1 ?? f2)()` or `(f1 ?? f2).Invoke()`. The current version on `main` will rewrite `(f1 ?? f2)()` to `(f1 ?? f2).Invoke(c)`, adding the explicit `Invoke` where none was there before. I assume it's preferable for a fixer to do the bare minimum to implement the fix without changing any other code, and this PR respects whether or not the explicit `Invoke` is used in the expression.

DavidBoike (commented):
@carlossanlop ready for another review.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
carlossanlop: We all use Telerik 😄 .

DavidBoike:
Anything else I can do here?

mavasani:
> Anything else I can do here?

@carlossanlop Can you please help review the PR again? We already have multiple users reporting this bug.

DavidBoike:
I rebased the PR to make it easier to review.

DavidBoike:
Checking in again. What can I do here?

DavidBoike:
@mavasani @carlossanlop @buyaa-n  Please let me know how I can help drive this forward. I'm willing to add more test cases or get on a call or whatever, just let me know.

adamralph:
@mavasani @carlossanlop is there anything that I, @DavidBoike, or anyone else can do to help move this forward? It would be a shame to abandon it after the non-trivial effort that's already been made by @DavidBoike and @carlossanlop. And as mentioned above, multiple users have reported this problem.

buyaa-n (approved):
The fix looks good to me, I will merge it on Monday if @carlossanlop will not add any other comments/concerns

CC @mavasani

jeffhandley:
@carlossanlop won't be available to review on Monday. LGTM.

adamralph:
🎉 which SDK version is this scheduled for?

jeffhandley:
This will be included in .NET 6 Preview 7.

Issue #4985, opened 2021-03-29
CA2016 ignores the return type of the method overloads


jmarolf:
Version: 6.0.0-preview1.21054.10

In the following Code CA2016 is suggested even though the return types are completely different. 
```csharp
using System;
using System.Threading;
using System.Threading.Tasks;

class P
{
    static void M1(string s, CancellationToken cancellationToken)
    {
        var result = M2(s); // CA2016
    }

    static Task M2(string s) { throw new NotImplementedException(); }

    static int M2(string s, CancellationToken cancellationToken) { throw new NotImplementedException(); }
}
```

CA2016 should at least consider the return type of the method it is examining to decide if the cancellation token can be passed in.

PR #4986, opened 2021-03-29
Do Not consider overloads with different return types for CA2016

jmarolf:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4985

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
@@ -386,7 +386,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 // Overload is  valid if its return type is implicitly convertable
                 var toCompareReturnType = methodToCompareWithAllParameters.ReturnType;
                 var originalReturnType = originalMethodWithAllParameters.ReturnType;
-                if (!compilation.ClassifyCommonConversion(toCompareReturnType, originalReturnType).IsImplicit)
+                if (!toCompareReturnType.IsAssignableTo(originalReturnType, compilation))
                 {
                     // Generic Task-like types are special since awaiting them essentially erases the task-like type.
                     // If both types are Task-like we will warn if their generic arguments are convertable to each other.
@@ -431,7 +431,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
                     var leftTypeArgument = left.TypeArguments[0];
                     var rightTypeArgument = right.TypeArguments[0];
-                    if (!compilation.ClassifyCommonConversion(rightTypeArgument, leftTypeArgument).IsImplicit)
+                    if (!leftTypeArgument.IsAssignableTo(rightTypeArgument, compilation))
                     {
                         return false;
                     }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
@@ -77,6 +77,10 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 return;
             }
 
+            // We don't care if these symbols are not defined in our compilation. They are used to special case the Task<T> <-> ValueTask<T> logic
+            context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTask1, out INamedTypeSymbol? genericTask);
+            context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksValueTask1, out INamedTypeSymbol? genericValueTask);
+
             context.RegisterOperationAction(context =>
             {
                 IInvocationOperation invocation = (IInvocationOperation)context.Operation;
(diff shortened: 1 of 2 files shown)


jmarolf:
cc: @carlossanlop

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
```diff
@@ -336,6 +337,12 @@ private static bool AnyArgument(ImmutableArray<IArgumentOperation> arguments, Fu
             // Checks if the parameters of the two passed methods only differ in a ct.
             static bool HasSameParametersPlusCancellationToken(ITypeSymbol cancellationTokenType, IMethodSymbol originalMethod, IMethodSymbol methodToCompare)
             {
+                // Overload is not valid if it has a different return type
+                if (!originalMethod.ReturnType.Equals(methodToCompare.ReturnType))
```
stephentoub, line 341: I think this is too specific a constraint. The common case will be that the method result is directly awaited,  in which case any awaitable return value is fine. That's probably harder to validate,  but when directly awaited,  we should at least allow for a ValueTask- returning overload where the original is Task- returning.  This is not uncommon,  with a cancellation token being added to a new overload that then also returns value task instead.

jmarolf: I think you could amend this to say that if a type is convertible to the resultant type it should be allowed

So we would not warn for something like:

```csharp
static void M1(string s, CancellationToken cancellationToken)
{
    var result = M2(s); // CA2016
}

static long M2(string s) { throw new NotImplementedException(); }

static int M2(string s, CancellationToken cancellationToken) { throw new NotImplementedException(); }
```

but not here:
```csharp
static void M1(string s, CancellationToken cancellationToken)
{
    var result = M2(s); // CA2016
}

static (int value1, int value2) M2(string s) { throw new NotImplementedException(); }

static int M2(string s, CancellationToken cancellationToken) { throw new NotImplementedException(); }
```

I am mostly concerned about cases where the return types are wildly divergent. In most of these cases it is because we are calling a helper type _within_ a method or the methods are simply poorly named and are not semantically equivalent.

stephentoub: > I am mostly concerned about cases where the return types are wildly divergent.

Yup, the concern is valid.  Just wanted to avoid losing out on the cases that aren't problematic and that we know do arise. Thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
```diff
@@ -336,6 +337,12 @@ private static bool AnyArgument(ImmutableArray<IArgumentOperation> arguments, Fu
...
+                {
+                    return false;
+                }
+
                 // Avoid comparing to itself, or when there are no parameters, or when the last parameter is not a ct
                 if (originalMethod.Equals(methodToCompare) ||
```
carlossanlop, line 347: I agree with your change. However, I feel like the cases that were already being checked inside this `if` are more common than the case you're adding. Can you please consider putting your new condition at the end of the existing `if`?

carlossanlop (commented):
Looks good as it is, but I left a small suggestion for you to consider.

The amend you suggested to Stephen (overloads with convertible return types will get flagged) sounds like a reasonable check to include in this PR as well.

jmarolf:
Thanks @carlossanlop. I'll add the check to see if the types are implicitly convertible to each other so things like `Task` and `ValueTask` will continue to work and I'll add a test for that

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
jmarolf: I've changed the rules to now consider it a valid overload if:
1. return types are implicitly convertible to each other
2. In the case of generic Task-Like types consider them compatible if their type arguments are implicitly convertable to each other

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocations.Analyzer.cs
```diff
@@ -365,7 +383,61 @@ static bool HasSameParametersPlusCancellationToken(ITypeSymbol cancellationToken
                     }
                 }
 
+                // Overload is  valid if its return type is implicitly convertable
+                var toCompareReturnType = methodToCompareWithAllParameters.ReturnType;
+                var originalReturnType = originalMethodWithAllParameters.ReturnType;
+                if (!compilation.ClassifyCommonConversion(toCompareReturnType, originalReturnType).IsImplicit)
```
mavasani, line 389: We have a helper IsAssignableTo in language specific layers that you might be able to use.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ForwardCancellationTokenToInvocationsTests.cs
```diff
@@ -2229,6 +2249,86 @@ void LocalMethod()
...
+    static Task<int> M2(string s) { throw new NotImplementedException(); }
+
+    static ValueTask<int> M2(string s, CancellationToken cancellationToken) { throw new NotImplementedException(); }
+}";
+            return VerifyCS.VerifyCodeFixAsync(originalCode, fixedCode);
+        }
+
         #endregion
 
         #region No Diagnostic - VB
```
carlossanlop, line 2334: Can you please also add VB tests?

mavasani (approved):

PR #4992, opened 2021-04-02
Track changes between public and protected in PublicApiAnalyzers

### LINKED ISSUE #4991: PublicAPI Analyzer treats `abstract class` implicit constructors as public
https://github.com/dotnet/roslyn/pull/51330/files#diff-9a6b818860122e38724a80cb5258e64ed6a82c7e1c81887a3440ab19b12c9758R25-R29

This change should have been flagged by the public API analyzer, as `abstract class` implicit constructors are protected, and this commit changed the accessibility of the constructor to public. However, the analyzer treats these constructors as public apis, not protected: https://github.com/sharwell/roslyn/blob/2aa1c3fa73c343f0f6b0444a8d46850fffbb2b98/src/Workspaces/Core/Portable/PublicAPI.Shipped.txt#L884.

Youssef1313:
### TODO

- [x] Do we need an option for this to allow users to upgrade without having to make additional (possibly large) changes?

    No (or not now at least). From comments below:
    > Let's wait for feedback before adding one. Theoretically, one can hold off the upgrade to new public API analyzers package if they are not ready for this (reasonable) breaking change.

- [ ] We need to test the package produced by this build in dotnet/roslyn to see if there are any problems.
- [x] Is the approach taken good enough? I'm open to any alternatives.
- [x] Write more tests.

@333fred @sharwell Can you take a look?

Fixes #4991


sharwell:
This syntax seems like the least intrusive way to add this feature considering it was never present in the past.

Youssef1313:
@sharwell Any alternatives in mind? That was the main reason I asked:

> Do we need an option for this to allow users to upgrade without having to make additional (possibly large) changes?

:confused:

mavasani:
I think the change looks fine to me as-is.

> Do we need an option for this to allow users to upgrade without having to make additional (possibly large) changes?

Let's wait for feedback before adding one. Theoretically, one can hold off the upgrade to new public API analyzers package if they are not ready for this (reasonable) breaking change.

mavasani:
@333fred @sharwell do you have any concerns here? Otherwise, @Youssef1313 can go ahead with addressing other TODOs.

mavasani:
@Youssef1313 Can you please proceed with other TODOs? We haven't heard any pushback for your chosen approach.

Youssef1313:
@mavasani What would be the easiest way to run this against Roslyn?

The instructions [here](https://github.com/dotnet/roslyn-analyzers/blob/a0522afd52b25ad211298e308fd5bbc7734865e0/docs/NetCore_GettingStarted.md#testing-against-the-roslyn-repo) will only run the analyzer and show the violations. I want to experiment with the codefix experience inside Visual Studio and see how easy is it to deal with the new change.

mavasani:
> What would be the easiest way to run this against Roslyn?

1. Build the analyzer package locally with `msbuild /t:pack`
2. Checkout roslyn main
3. Update [NuGet.config](https://github.com/dotnet/roslyn/blob/main/NuGet.config) to add a local feed pointing to the roslyn-analyzers folder on your machine that has the built analyzer package.
4. Update the version below (note `Roslyn.Diagnostics.Analyzers` brings in `Microsoft.CodeAnalysis.Analyzers`) https://github.com/dotnet/roslyn/blob/bc8b7b806521bb5febd05e392cc47925cac1ef0b/eng/targets/Settings.props#L124

Youssef1313:
@mavasani Thanks. I tried that now, but `build.cmd` seems to complete successfully. I even tried to hand-edit one of the `PublicAPI.Shipped.md`, and `build.cmd` still shows 0 errors 0 warnings. (Did that after a `git clean -xdf).

Is the package enabled by a specific property that's disabled by default for `build.cmd`? or do I need to pass some flag to it to enable it?

Here are my changes in case I'm doing it wrong:

image

mavasani:
Ah yes, you need to run ‘build.cmd -a’

Youssef1313:
Still can't get a diagnostic for the removed entry (I'm expecting diagnostics due to this change as well)

Also wanted to point that commenting the local entry in `NuGet.config` doesn't fail the build despite I'm using 3.3.3-dev (I expect it to fail as this version isn't found - but not sure if it's related to some caching)

mavasani:
@Youssef1313 Did you make sure to delete the NuGet package cache for this version of the analyzer package?

mavasani:
Also, I would suggest doing a rebuild instead of a build. There is a known msbuild issue that skips repeated build if the first build succeeded with warnings, the second build does not produce any.

sharwell:
Here's the full command:

```
.\Build.cmd -restore -rebuild -a
```

Evangelink:
ping @mavasani

mavasani:
Ping @Youssef1313

Youssef1313:
@mavasani Sorry for the delay.

The codefix looks like this:

image

I'm not sure this is an acceptable behavior given this is a breaking change. I think we should introduce a new code action that updates existing entries with modifiers. What do you think?

Youssef1313:
Also note that the codefix used to be make the text file ordered alphabetically. Do the new code action I suggested above have to keep the file ordered (produce noisy diff), or just update existing entries in their place?

Youssef1313:
Converting to a draft since this is not ready. But need to hear feedback about latest changes and the plan for a new diagnostic with fix to ease dealing with this breaking change.

PR #5012, opened 2021-04-11
OSX needs to treated  as the same platform as MacOS, everywhere

buyaa-n:
OSX needs to treated  as the same platform as MacOS, everywhere

Somehow the code handling this logic lost 🙅 (i remember had it handled in my first version, I guess it is lost in the design change refactoring result).

The brute force way for treating `OSX` as `MacOS` would add additional check for each OS platform check and platform lookup, which don't like to do. So instead:
- Whenever an OSPlatform attribute with OSX found for an API adding it as `MacOs` in the internal cache, so no need to add `OSX` specific special handling anymore
- The only issue with this solution can be seen in reporting the incompatible APIs, i.e. when there is `OSX` specific API used from an incompatible call site then developers would get a warning that the API is supported on `MacOS` instead of `OSX`. Handling this issue with reporting the platform as `MacOS/OSX` - meaning it can be `MacOS` or `OSX`

Example:
```cs
class Test
{
    public void Api_Usage()
    {
        if (OperatingSystem.IsWindows() ||
            OperatingSystem.IsLinux() ||
            OperatingSystem.IsMacOS())
        {
            Api();
            Api2(); // All supported platforms guarded
        }

        [|Api()|]; // This call site is reachable on all platforms. 'Test.Api()' is only supported on: 'MacOS/OSX', 'Linux', 'windows'.
        [|Api2()|]; // This call site is reachable on all platforms. 'Test.Api2()' is only supported on: 'MacOS/OSX', 'Linux', 'windows'.
    }

    void Api() { }

    [SupportedOSPlatform("Osx")]
    void Api2() { }
}
```

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -56,7 +56,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
         private const string OptionalSuffix = "VersionAtLeast";
         private const string Net = "net";
         private const string macOS = nameof(macOS);
-        private const string MacSlashOSX = "MacOS/OSX";
+        private const string MacSlashOSX = "macOS/OSX";
 
         internal static DiagnosticDescriptor OnlySupportedCsReachable = DiagnosticDescriptorHelper.Create(RuleId,
                                                                                       s_localizableTitle,
@@ -650,7 +650,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                     platformNames.Add(GetFormattedString(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityAllVersions, pName));
                                     continue;
                                 }
-                                platformNames.Add(EncloseWithQuota(pName));
+                                platformNames.Add(EncloseWithQuotes(pName));
                             }
                             else
                             {
@@ -666,7 +666,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                     platformNames.Add(GetFormattedString(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityAllVersions, pName));
                                     continue;
                                 }
-                                platformNames.Add(EncloseWithQuota(pName));
+                                platformNames.Add(EncloseWithQuotes(pName));
                             }
                             else
                             {
@@ -757,7 +757,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                         platformNames.Add(GetFormattedString(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityAllVersions, pName));
                                         continue;
                                     }
-                                    platformNames.Add(EncloseWithQuota(pName));
+                                    platformNames.Add(EncloseWithQuotes(pName));
                                 }
                                 else
                                 {
@@ -770,7 +770,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                             unsupportedRule = false;
                             if (IsEmptyVersion(supportedVersion))
                             {
-                                platformNames.Add(EncloseWithQuota(pName));
+                                platformNames.Add(EncloseWithQuotes(pName));
                             }
                             else
                             {
@@ -818,7 +818,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                     platformNames.Add(GetFormattedString(MicrosoftNetCoreAnalyzersResources.PlatformCompatibilityAllVersions, pName));
                                     continue;
                                 }
-                                platformNames.Add(EncloseWithQuota(pName));
+                                platformNames.Add(EncloseWithQuotes(pName));
                             }
                             else
                             {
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -205,6 +205,103 @@ class Target
...
+    public void Api_Usage()
+    {
+        if (!OperatingSystem.IsWindows() &&
+            !OperatingSystem.IsLinux() &&
+            !OperatingSystem.IsMacOS())
+        {
+            Api();
+        }
+
+        [|Api()|]; // This call site is reachable on all platforms. 'Test.Api()' is unsupported on: 'MacOS/OSX', 'windows'.
```
mavasani, line 263: Do we want at least one test that verifies the diagnostic message/argument for these cases? I see that all the added test have annnotation based verification, and will pass even if diagnostic message does not have the expected terminology.

mavasani (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: For this test better not add message/argument verification because for multiple platforms message the order of the platforms appears in the list not consistent. But i will add such verification for tests flag only on platform

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -55,6 +55,8 @@ public sealed partial class PlatformCompatibilityAnalyzer : DiagnosticAnalyzer
         private const string IsPrefix = "Is";
         private const string OptionalSuffix = "VersionAtLeast";
         private const string Net = "net";
+        private const string macOS = nameof(macOS);
+        private const string MacSlashOSX = "MacOS/OSX";
```
jeffhandley, line 59: The `"m"` should be lower-case in `"macOS"`.

```suggestion
        private const string MacSlashOSX = "macOS/OSX";
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -870,7 +872,13 @@ static bool GetPlatformNames(SmallDictionary<string, Versions> attributes, Small
...
             }
 
-            static string GetFormattedString(string resource, params object[] args) => string.Format(CultureInfo.InvariantCulture, resource, args);
+            static string GetFormattedString(string resource, string platformName, object? arg1 = null, object? arg2 = null) =>
+                string.Format(CultureInfo.InvariantCulture, resource, AddOsxIfNeeded(platformName), arg1, arg2);
+
+            static string AddOsxIfNeeded(string platformName) =>
+                platformName.Equals(macOS, StringComparison.OrdinalIgnoreCase) ? MacSlashOSX : platformName;
+
+            static string EncloseWithQuota(string pName) => $"'{AddOsxIfNeeded(pName)}'";
```
jeffhandley, line 881: This was probably meant to be named `EncloseWithQuotes`. This should also use a resource string since `"` characters are not universal and should be localizable. French uses `«{0}»`, for example.

jeffhandley (changes requested):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: > This was probably meant to be named EncloseWithQuotes. 

Right

> This should also use a resource string since " characters are not universal and should be localizable. French uses «{0}», for example.

I am not sure that is necessary, in this repo we every use single quotes for enclosing APIs  `'{0}'` everywhere. https://github.com/dotnet/roslyn-analyzers/blob/fde0c6667d5af4b8cf1c7b135325748f2c0fb226/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx#L124

buyaa-n: i think `'{0}'` is convention even in runtime

jeffhandley: OK fair enough. It looks like our [French translation of that message](https://github.com/dotnet/roslyn-analyzers/blob/fde0c6667d5af4b8cf1c7b135325748f2c0fb226/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/xlf/MicrosoftNetCoreAnalyzersResources.fr.xlf#L1870-L1874) does not change the quotation marks. Thanks for checking it.

jeffhandley (approved):

PR #4825, opened 2021-02-10
Update the release analyzer fixer to add a blank line after headers

Evangelink:
Fixes #4131

@mavasani this PR could be considered as a breaking-change so we probably don't want to target master but I am not sure which of the preview branch to target.

PR #5033, opened 2021-04-18
Include AnalyzerReleases in markdownlint

Youssef1313:


src/Microsoft.CodeAnalysis.Analyzers/Core/AnalyzerReleases.Shipped.md
```diff
@@ -1,6 +1,7 @@
 ## Release 3.0.0
 
 ### New Rules
+
```
mavasani, line 4: You should also fix the code fixer that generates these tables, probably in a separate PR though. I am presuming the analyzer already ignores blank lines and hence is unaffected.

Youssef1313: @mavasani It was already fixed by @Evangelink.

https://github.com/dotnet/roslyn-analyzers/pull/4825

mavasani (approved):

PR #5055, opened 2021-04-22
iconUrl -> icon

Youssef1313:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5054

### DIFF
--- a/eng/GenerateAnalyzerNuspec.targets
+++ b/eng/GenerateAnalyzerNuspec.targets
@@ -144,7 +144,7 @@
       <_NuspecMetadata Include="copyright=$(Copyright)" />
       <_NuspecMetadata Include="license=$(PackageLicenseExpression)" />
       <_NuspecMetadata Include="projectUrl=$(PackageProjectUrl)" />
-      <_NuspecMetadata Include="iconUrl=$(PackageIconUrl)" />
+      <_NuspecMetadata Include="icon=$(PackageIconUrl)" />
       <_NuspecMetadata Include="releaseNotes=$(PackageReleaseNotes)" />
       <_NuspecMetadata Include="tags=$(PackageTags)" />
       <_NuspecMetadata Include="serviceable=$(Serviceable)" />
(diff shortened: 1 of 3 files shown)


mmitche:
@Youssef1313 The icon element is to an embedded icon, rather than a url.

eng/GenerateAnalyzerNuspec.targets
```diff
@@ -144,7 +144,7 @@
       <_NuspecMetadata Include="copyright=$(Copyright)" />
       <_NuspecMetadata Include="license=$(PackageLicenseExpression)" />
       <_NuspecMetadata Include="projectUrl=$(PackageProjectUrl)" />
-      <_NuspecMetadata Include="iconUrl=$(PackageIconUrl)" />
+      <_NuspecMetadata Include="icon=$(PackageIconUrl)" />
```
Youssef1313, line 147:
```suggestion
      <_NuspecMetadata Include="icon=$(PackageIcon)" />
```

Youssef1313:
Thanks @mmitche. I updated the property name. Since no packages in the repository uses an icon, I believe no other renames are needed. But will let @mavasani confirm.

mmitche:
I think you'll also need to ensure that you package up the icon too. There is one being brought in from arcade (PackageIconFullPath), but the build and packaging logic in this repo is different enough from arcade that it might take some doing to get it to work.

src/Tools/GenerateAnalyzerNuspec/Program.cs
```diff
@@ -85,6 +85,7 @@
 result.AppendLine(@"  </metadata>");
 
 result.AppendLine(@"  <files>");
+result.AppendLine(@"    $CommonFileElements$");
```
Youssef1313, line 88: Added this based on the description in https://github.com/dotnet/arcade/pull/4078.

mmitche:
@Youssef1313 Did you check the package content to verify it has the expected nuspec and icon file?

Youssef1313:
@mmitche I ran `build.cmd -pack`, changed one of the produced nupkg to `.zip` and extracted it, then verified `Icon.png`:

image

Anything else to check?

mmitche:
LGTM then!

mmitche (approved):

PR #4841, opened 2021-02-12
Add WhenAll and WaitAll analyzer for single task argument

ryzngard:
Fixes https://github.com/dotnet/runtime/issues/33806 and https://github.com/dotnet/runtime/issues/33807

Definition of Done Checklist 
* [x] Analyzer and Code Fix implemented for C# and VB
* [ ] Run against dotnet/runtime
* [ ] Run against dotnet/roslyn-analyzers
* [ ] Run against dotnet/roslyn
* [ ] Document for review: severity, default, categorization, numbering, titles, messages, and descriptions.

PR #5060, opened 2021-04-23
Cleanup DoNotUseWhenAllOrWaitAllWithSingleArgument

Youssef1313:
Quick follow-up to #4841

- Removed unnecessary derived classes.
- Used non-localized equivalence key.
- Renamed all contexts to `context.
- Used SymbolEqualityComparer.Default

cc: @ryzngard


Youssef1313:
@mavasani I'm unable to debug the failure.

Locally, the tests fail with: `NuGet.Packaging.Core.PackagingException : The package is missing the required nuspec file. Path: C:\Users\Youssef\AppData\Local\Temp\test-packages\Microsoft.NETCore.App.Ref.3.1.0`

I have no clue how to fix this. :/

Youssef1313:
> System.AggregateException : One or more errors occurred. (Unexpected Null)\n---- System.InvalidOperationException : Unexpected Null

:confused:

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/DoNotUseWhenAllOrWaitAllWithSingleArgument.cs
```diff
@@ -42,28 +42,28 @@ public sealed override void Initialize(AnalysisContext context)
         {
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
-            context.RegisterCompilationStartAction(compilationContext =>
+            context.RegisterCompilationStartAction(context =>
```
ryzngard, line 45: Is there any particular reason to rename this to just `context`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/DoNotUseWhenAllOrWaitAllWithSingleArgumentFixer.cs
```diff
@@ -12,7 +13,8 @@
 
 namespace Microsoft.NetCore.Analyzers.Tasks
 {
-    public abstract class DoNotUseWhenAllOrWaitAllWithSingleArgumentFixer : CodeFixProvider
+    [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic), Shared]
```
ryzngard, line 16: oops 😅 I really thought I had this change in already. Thanks for fixing!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/DoNotUseWhenAllOrWaitAllWithSingleArgument.cs
```diff
@@ -112,7 +112,8 @@ private static bool IsSingleTaskArgument(IInvocationOperation invocation, INamed
                 return false;
             }
 
-            return namedTypeSymbol.Equals(taskType) || namedTypeSymbol.ConstructedFrom.Equals(genericTaskType);
+            return namedTypeSymbol.Equals(taskType, SymbolEqualityComparer.Default) ||
```
ryzngard, line 115: I thought `symbol.Equals(otherSymbol)` implicitly used the `SymbolEqualityComparer.Default` if available?

ryzngard:
> @mavasani I'm unable to debug the failure.
> 
> ~Locally, the tests fail with: `NuGet.Packaging.Core.PackagingException : The package is missing the required nuspec file. Path: C:\Users\Youssef\AppData\Local\Temp\test-packages\Microsoft.NETCore.App.Ref.3.1.0`~
> 
> ~I have no clue how to fix this. :/~ Deleting test-packages folder fixed it.
> 
> But the behavior of IOperation API looks very strange:
> 
> image
> 
> Is that a bug in the operation tree?

It looks like you figured this out, but the Value being an implicitly generated IArrayCreationOperation changes the behavior slightly.

ryzngard:
Thanks for doing this!

ryzngard (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/DoNotUseWhenAllOrWaitAllWithSingleArgument.cs
Youssef1313: @ryzngard It does. But there is an analyzer (RS1024) to enforce using SymbolEqualityComparer explicitly to make the intent clear.

Currently the analyzer is disabled due to few bugs and will be re-enabled once they are fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Tasks/DoNotUseWhenAllOrWaitAllWithSingleArgument.cs
Youssef1313: @ryzngard It's almost always incorrect to use a context other than the most inner one. So using the same name ensures this doesn't happen accidentally.

PR #5062, opened 2021-04-23
Remove dead condition

### LINKED ISSUE #3023: UseNameofInPlaceOfString: Add coverage for early return

### LINKED ISSUE #3115: Coverage - UseNameofInPlaceOfString early exit is not tested
On PR #3005 we have noticed that the early exit is not tested. We need to add a test that would cover this case.

Youssef1313:
Closes #3023
Closes #3115 (they look like an exact dupe)

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/UseNameofInPlaceOfString.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/UseNameofInPlaceOfString.cs
@@ -60,16 +60,11 @@ namespace Microsoft.CodeQuality.Analyzers.Maintainability
                 return;
             }
 
-            if (argument.Parameter == null)
-            {
-                return;
-            }
-
             var stringText = (string)argument.Value.ConstantValue.Value;
 
-            var matchingParameter = argument.Parameter;
-
-            switch (matchingParameter.Name)
+            // argument.Parameter will not be null here. The only case for it to be null is an __arglist argument, for which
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/UseNameofInPlaceOfString.cs
```diff
@@ -60,16 +60,11 @@ private void AnalyzeArgument(OperationAnalysisContext context)
...
-            {
-                return;
-            }
-
             var stringText = (string)argument.Value.ConstantValue.Value;
 
-            var matchingParameter = argument.Parameter;
-
-            switch (matchingParameter.Name)
+            // argument.Parameter will not be null here. The only case for it to be null is an __arglist argument, for which
```
genlu, line 65: I think removing the check is fine. That said, I'm also not sure there's any harm leaving this simple check in as is :)

genlu (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/UseNameofInPlaceOfString.cs
Youssef1313: Yeah there is no harm in keeping it. But the linked issue stated that it should either be deleted or a test coverage is added.

Issue #5035, opened 2021-04-19
CA1710 Clearing default suffixes doesn't work


erichiller:
### Analyzer

**Diagnostic ID**: [CA1710](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1710): `Identifiers should have correct suffix`
### Analyzer source

Versions:
```
.NET SDK (reflecting any global.json):
 Version:   5.0.300-preview.21180.15
 Commit:    19ce2535ee

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19042
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.300-preview.21180.15\

Host (useful for support):
  Version: 5.0.4
  Commit:  f27d337295

.NET SDKs installed:
  3.1.102 [C:\Program Files\dotnet\sdk]
  3.1.302 [C:\Program Files\dotnet\sdk]
  3.1.401 [C:\Program Files\dotnet\sdk]
  5.0.100 [C:\Program Files\dotnet\sdk]
  5.0.200-preview.20601.7 [C:\Program Files\dotnet\sdk]
  5.0.300-preview.21180.15 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
[...]

PR #5064, opened 2021-04-23
Correctly give higher priority to user's suffix map

Youssef1313:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5035

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
@@ -176,14 +176,15 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
             foreach (var type in typeSymbols)
             {
                 // User specific mapping has higher priority than hardcoded one
-                if (userMap.TryGetValue(type.OriginalDefinition, out var suffix) &&
-                    !RoslynString.IsNullOrWhiteSpace(suffix))
+                if (userMap.TryGetValue(type.OriginalDefinition, out var suffix))
                 {
-                    suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
-                    return true;
+                    if (!RoslynString.IsNullOrWhiteSpace(suffix))
+                    {
+                        suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
+                        return true;
+                    }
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
```diff
@@ -176,14 +176,15 @@ private static void AnalyzeCompilationStart(CompilationStartAnalysisContext cont
...
-                    return true;
+                    if (!RoslynString.IsNullOrWhiteSpace(suffix))
+                    {
+                        suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
+                        return true;
+                    }
                 }
-
-                if (hardcodedMap.TryGetValue(type.OriginalDefinition, out suffixInfo))
+                else if (hardcodedMap.TryGetValue(type.OriginalDefinition, out suffixInfo))
```
Youssef1313, line 187: This can be seen as breaking change :confused:

Old behavior was that defining a user map *extends* the hardcoded one.
New behavior is that defining a user map *replaces* the hardcoded one.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/IdentifiersShouldHaveCorrectSuffix.cs
```diff
@@ -176,14 +176,15 @@ private static void AnalyzeCompilationStart(CompilationStartAnalysisContext cont
             foreach (var type in typeSymbols)
             {
                 // User specific mapping has higher priority than hardcoded one
-                if (userMap.TryGetValue(type.OriginalDefinition, out var suffix) &&
-                    !RoslynString.IsNullOrWhiteSpace(suffix))
+                if (userMap.TryGetValue(type.OriginalDefinition, out var suffix))
                 {
-                    suffixInfo = SuffixInfo.Create(suffix, canSuffixBeCollection: false);
-                    return true;
+                    if (!RoslynString.IsNullOrWhiteSpace(suffix))
```
genlu, line 181: So we are still going through all symbols even after we found an empty match in userMap? The behavior makes sense, just consider adding a test to capture this intention. (i.e. some type implement two interfaces, one with user overriding the other without)

genlu (approved):
:shipit:

Youssef1313:
Sorry for the delay @genlu. Added the test.

genlu:
Thanks! @Youssef1313

PR #5065, opened 2021-04-24
Use AnalyzerConfigFiles instead of AnalyzerConfigDocument

Youssef1313:
This was recently added in the testing library in https://github.com/dotnet/roslyn-sdk/pull/734.
~~The main goal of this is to (later) remove the logic of parsing editorconfig files and leave that up to the compiler (https://github.com/dotnet/roslyn-sdk/pull/734#discussion_r581165302)~~ This comment is about `AdditionalFiles`, not `AnalyzerConfigDocument`.

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
@@ -925,7 +925,7 @@ public class C
     }}
 }}"
                     },
-                    AnalyzerConfigFiles = { ("/.editorconfig", $"[*]{Environment.NewLine}{editorConfigText}") }
+                    AnalyzerConfigFiles = { ("/.editorconfig", $"[*]{Environment.NewLine}{editorConfigText}") },
                 },
             };
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
@@ -868,7 +868,7 @@ public class C
 
 [*]
 {editorConfigText}
-") }
+") },
                 }
             }.RunAsync();
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
@@ -948,7 +948,7 @@ Public Class C
     End Sub
 End Class"
                     },
-                    AnalyzerConfigFiles = { ("/.editorconfig", $"[*]{Environment.NewLine}{editorConfigText}") }
+                    AnalyzerConfigFiles = { ("/.editorconfig", $"[*]{Environment.NewLine}{editorConfigText}") },
                 },
             };
 
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
@@ -1,11 +1,9 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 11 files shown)


Youssef1313:
@sharwell @mavasani With this test-only change, the product code no longer gets the correct value for `GetMSBuildPropertyValue`.

It looks like `GetMSBuildPropertyValue` tries to read from `AdditionalFiles`. Does that indicate a product bug? I'm uncertain how to proceed with working on this.

src/Test.Utilities/CSharpCodeFixVerifier`2+Test.cs
```diff
@@ -40,20 +39,6 @@ static Test()
...
-                    {
-                        solution = solution.AddAnalyzerConfigDocument(
-                            DocumentId.CreateNewId(projectId, debugName: ".editorconfig"),
-                            ".editorconfig",
-                            SourceText.From($"is_global = true" + Environment.NewLine + AnalyzerConfigDocument),
-                            filePath: @"/.editorconfig");
-                    }
-
-                    return solution;
-                });
```
Youssef1313, line 56: AnalyzerConfigDocument doesn't work with named sections despite it looks like it should parse from the compiler?
This is a point of confusion to me.

While I'm deleting `AnalyzerConfigDocument` altogether, the behavior of it not working when its value is like `"[*]\r\nprop=value` but works when it's `"prop=value"` is very very strange to me.

I'm either missing something or this may indicate a bug from roslyn side.

sharwell:
I need to look and see how this relates to #4921

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
```diff
@@ -926,14 +924,14 @@ public class C
     {{
     }}
 }}"
-                    }
+                    },
+                    AnalyzerConfigFiles = { ("/.editorconfig", $"[*]{Environment.NewLine}{editorConfigText}") }
```
mavasani, line 928: IMO, this deserves a static helper method to return `AnalyzerConfigFiles` given `string editorConfigText` as the input.

Youssef1313: @mavasani For this specific test file, I used it only twice (in the same test). Did you mean to introduce it in a new helper class and use it from all tests?

Youssef1313:
> I need to look and see how this relates to #4921

@sharwell Have you got the time to look? Thanks :)

sharwell:
😕 I thought I already implemented this as part of another change

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
sharwell: The form used here is the form I generally expect to see.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -143,21 +146,24 @@ public void M()
...
                 // /0/Test0.cs(9,13): warning IL3001: Assemblies embedded in a single-file app cannot have additional files in the manifest.
-                VerifyCS.Diagnostic(AvoidAssemblyLocationInSingleFile.IL3001).WithSpan(9, 13, 9, 25).WithArguments("System.Reflection.Assembly.GetFiles()")
+                VerifyCS.Diagnostic(IL3001).WithSpan(9, 13, 9, 25).WithArguments("System.Reflection.Assembly.GetFiles()")
                 );
         }
 
         private Task VerifyDiagnosticsAsync(string source, params DiagnosticResult[] expected)
         {
-            const string singleFilePublishConfig = @"
+            const string singleFilePublishConfig = @"[*]
```
sharwell, line 157: 📝 Build properties will always appear in a global config

```suggestion
            const string singleFilePublishConfig = @"is_global = true
```

src/Test.Utilities/CSharpCodeFixVerifier`2+Test.cs
sharwell: Global config files don't use `[*]`. They just have `is_global=true` as the first line followed by the properties themselves.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/ReviewUnusedParametersTests.cs
sharwell, line 928: 💡 A new line should be added at the end of this, specifically to avoid the case like this file encountered. Line 927 should not have changed as part of this pull request, but a change was forced because the line failed to include the trailing comma.

sharwell (approved):

PR #1952, opened 2018-12-28
Add .editorconfig based configuration support for analyzers

mavasani:
Fixes #1850

editorconfigbasedoptions

**Additions in this PR:**
1. `EditorConfigParser`: Simple .editorconfig based parser to read "key = value" property values from an additional file with name _.editorconfig_. Note that this is a temporary parser that will be deleted once the compiler support for .editorconfig comes online.
2. `CategorizedAnalyzerConfigOptions`: Analyzer configuration options from an .editorconfig file that are parsed into general and specific configuration options.
  .editorconfig format:
   1. General configuration option:
      1. `dotnet_code_quality.OptionName = OptionValue`
   2. Specific configuration option:
      1. `dotnet_code_quality.RuleId.OptionName = OptionValue`
      2. `dotnet_code_quality.RuleCategory.OptionName = OptionValue`
   .editorconfig examples to configure API surface analyzed by analyzers:
   1. General configuration option:
      1. `dotnet_code_quality.api_surface = all`
   2. Specific configuration option:
      1. `dotnet_code_quality.CA1040.api_surface = public, internal`
      2. `dotnet_code_quality.Naming.api_surface = public`
3. `AnalyzerOptionsExtensions`: Extension methods on `AnalyzerOptions` to fetch .editorconfig based configuration options. This PR adds support for the most commonly requested configuration option: _analyzed API surface_.
4. Update all the analyzers that defaulted to `IsExternallyVisible` default visibility for FxCop compat to operate based on .editorconfig configuration, with `IsExternallyVisible` being the fallback check in absence of .editorconfig configuration for corresponding rule.
5. Update the default props file included with each analyzer NuGet package to tag .editorconfig at the root of the project directory as an additional file. This functionality is enabled by default, but is guarded by a property to enable end users to specify different custom .editorconfig. With this change, end users just need to drop a .editorconfig file at the root of their project directory, and the analyzers will automatically respect the configuration setting within it.

To support rule category based configuration, I also fixed #1468 as part of this PR.
Fixes #1468

**Follow-up work:**
1. ~~Add an analyzer that reports additional file diagnostics for unknown key and/or values in the .editorconfig file. Currently, we just drop them on the floor.~~ On second thought, we cannot really do this as the editorconfig is shared with other analyzer packages, for which the key/value pair might be an expected one.
2. Add support for more configuration options. For example, #1826 for dataflow analysis, #1859 for security analyzers, etc.
[...]

PR #5066, opened 2021-04-24
Delete EditorConfigParser

Youssef1313:


Youssef1313:
CI is green.

Based on the initial implementation in https://github.com/dotnet/roslyn-analyzers/pull/1952:

> EditorConfigParser: Simple .editorconfig based parser to read "key = value" property values from an additional file with name .editorconfig. Note that this is a temporary parser that will be deleted once the compiler support for .editorconfig comes online.

Is this the time to delete it? Any extra work is needed? (Keeping in mind that some analyzers are still using an old version of MS.CA packages - I have no historical background on when the compiler support for editorconfig was added)

src/Utilities/Compiler/Options/AggregateCategorizedAnalyzerConfigOptions.cs
```diff
@@ -95,23 +91,17 @@ public T GetOptionValue<T>(string optionName, SyntaxTree? tree, DiagnosticDescri
...
         {
+            value = defaultValue;
+
             if (ReferenceEquals(this, Empty))
             {
-                value = defaultValue;
                 return false;
             }
 
-            // Prefer additional file based options for back compat.
```
Youssef1313, line 104: This comment is the only thing that worries me about this change.
@mavasani Can you provide more details on this? How this is a back compat concern if the compiler should be able to parse the editorconfig correctly whether it's provided as an additional file or not?

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -644,38 +603,14 @@ SymbolNamesWithValueOption<TValue> GetDefaultValue()
...
-            var createValueCallback = new ConditionalWeakTable<AnalyzerOptions, ICategorizedAnalyzerConfigOptions>.CreateValueCallback(_ => ComputeCategorizedAnalyzerConfigOptions());
-            return s_cachedOptions.GetValue(options, createValueCallback);
-
-            // Local functions.
-            ICategorizedAnalyzerConfigOptions ComputeCategorizedAnalyzerConfigOptions()
-            {
-                var categorizedOptionsFromAdditionalFiles = ComputeCategorizedAnalyzerConfigOptionsFromAdditionalFiles();
-
+            var createValueCallback = new ConditionalWeakTable<AnalyzerOptions, ICategorizedAnalyzerConfigOptions>.CreateValueCallback(_ =>
 #if CODEANALYSIS_V3_OR_BETTER
```
Youssef1313, line 607: Projects where this is not true:

- PublicAPIAnalyzers
- PerformanceSensitiveAnalyzers
- BannedAPIAnalyzers

Need to give a closer look to these since I believe there will be behavioral change if they come to this code path.

If they don't come to this code path, it makes more sense to move this conditional directive further up.

src/Utilities/Compiler/Options/OptionKind.cs
```diff
@@ -3,7 +3,7 @@
 namespace Analyzer.Utilities
 {
     /// <summary>
-    /// Kind of option to fetch from <see cref="AnalyzerOptionsExtensions"/>.
+    /// Kind of option to fetch from <see cref="ICategorizedAnalyzerConfigOptions"/>.
```
Youssef1313, line 6: `cref="AnalyzerOptionsExtensions"` generates a compiler warning since it's conditionally compiled. But `ICategorizedAnalyzerConfigOptions` is still accurate too (it's the high level part that uses OptionKind)

src/Utilities/Compiler/Options/AggregateCategorizedAnalyzerConfigOptions.cs
mavasani: @Youssef1313 This repo added support for parsing and respecting editorconfig options even prior to compiler adding support in VS2019 16.3. This was done via passing editorconfig files as additional files, as msbuild/compiler did not implicitly discover or respect editorconfig in build. Your change will break anyone on a VS2019 build prior to 16.3 OR if they had editorconfig outside the project directory cone and passed it as an additional file. We can recommend the users to switch to latest VS2019 builds + move from additional file based editorconfig to global config files, so the breaking change has a path forward towards resolution. However, it will be a breaking change indeed.

Youssef1313: @mavasani Thanks for clarifying. So this suggests that a doc update will be needed to clarify this breaking change?

cc: @gewarren

gewarren: Please file a breaking change issue at https://github.com/dotnet/docs/issues/new?assignees=&labels=&template=dotnet-breaking-change.md&title= once this is merged.

mavasani: I don't expect the original approach for passing editorconfig as additional file would be used that widely, so I feel like we can take the following route:

1. @Youssef1313 - alter your PR slightly so that if we encounter an editorconfig as an additional file, instead of silently ignoring it, we produce a graceful diagnostic message that helps in migration. Implementation wise, you can throw an exception with a clear message that users no longer need to pass editorconfig as an additional file. Either it will be implicitly discovered (if the file is in the project's directory or parent) OR it should be converted into a global config - also recommend moving to latest Dev16 update. The exception message can also point to https://docs.microsoft.com/dotnet/fundamentals/code-analysis/configuration-files. Users will see a special AD0001 analyzer diagnostic with this exception message when they move to this new analyzer package.
2. If someone files an issue on either the analyzers repo or doc repo about them encountering this, then we can consider adding a section at the end of the above docs link about this breaking change and the migration steps. I have a feeling that it might not be required.

Thoughts?

src/Utilities/Compiler/Extensions/CompilationExtensions.cs
```diff
@@ -1,12 +1,13 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+#pragma warning disable IDE0005 // Using directive is unnecessary - used in conditional compilation.
```
mavasani, line 3: Instead of applying this pragma to the entire set of usings, can you instead narrow it to only the usings used in conditional compilation?

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -208,35 +193,21 @@ static bool TryParseValue(string value, out ImmutableHashSet<TEnum> result)
...
-            bool defaultValue,
-            CancellationToken cancellationToken)
+            bool defaultValue)
         {
-            var analyzerConfigOptions = options.GetOrComputeCategorizedAnalyzerConfigOptions(compilation, cancellationToken);
+            var analyzerConfigOptions = options.GetOrComputeCategorizedAnalyzerConfigOptions(compilation);
             return analyzerConfigOptions.GetOptionValue(optionName, tree, rule, bool.TryParse, defaultValue);
         }
 
-        public static uint GetUnsignedIntegralOptionValue(
```
mavasani, line 218: FYI: I kept these unused overloads for consistency with other APIs in this file (one symbol overload and one syntax tree overload). It is fine to delete them though.

src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs
```diff
@@ -626,56 +577,26 @@ SymbolNamesWithValueOption<TValue> GetDefaultValue()
...
-#endif
-            }
-
-            CompilationCategorizedAnalyzerConfigOptions ComputeCategorizedAnalyzerConfigOptionsFromAdditionalFiles()
-            {
-                foreach (var additionalFile in options.AdditionalFiles)
-                {
-                    var fileName = Path.GetFileName(additionalFile.Path);
-                    if (fileName.Equals(".editorconfig", StringComparison.OrdinalIgnoreCase))
-                    {
```
mavasani, line 668: This might be a good place to throw an exception with the migration message.

src/Utilities/Compiler/Options/EmptyCategorizedAnalyzerConfigOptions.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Analyzer.Utilities
+{
+    internal sealed class EmptyCategorizedAnalyzerConfigOptions : ICategorizedAnalyzerConfigOptions
```
mavasani, line 7: Is this a new type or existing type being moved to a separate file?

Youssef1313: @mavasani New :(

Youssef1313: Actually I think I introduced it, used it, then removed usage in a later commit. Will double check and delete it.

docs/Analyzer Configuration.md
```diff
@@ -46,6 +46,8 @@ End users can enable `.editorconfig` based configuration for individual document
 
 Note that this additional file based approach is also supported on VS2019 16.3 and later releases for backwards compatibility.
 
+**The additional file based approach is no longer supported starting in Microsoft.CodeAnalysis.NetAnalyzers v(TODO). It will be implicitly discovered (if the file is in the project's directory or any ancestor directory), or it should be converted into a 'globalconfig'. See [Configuration files for code analysis rules](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/configuration-files).**
```
Youssef1313, line 49: @mavasani What version should I write in place of TODO? We likely want to include two versions, one for 5.x and one for 6.x?

mavasani: 5.0.4 or later. Which reminds me that we should release the 5.0.4 NuGet package from main branch soon as we have accumulated quite a few bug fixes.

Youssef1313:
```suggestion
**The additional file based approach is no longer supported starting in Microsoft.CodeAnalysis.NetAnalyzers v5.0.4. It will be implicitly discovered (if the file is in the project's directory or any ancestor directory), or it should be converted into a 'globalconfig'. See [Configuration files for code analysis rules](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/configuration-files).**
```

src/Utilities/Compiler/Options/AggregateCategorizedAnalyzerConfigOptions.cs
mavasani: Ping @Youssef1313

Youssef1313: @mavasani I already did that in 4753fd32fe94a1441a76c63ea3cf1e7732f804aa

mavasani: Ah sorry, I forgot. Let me take another final look at the PR.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/InterproceduralAnalysisConfiguration.cs
```diff
@@ -43,39 +43,40 @@ public struct InterproceduralAnalysisConfiguration : IEquatable<InterproceduralA
             ControlFlowGraph cfg,
             Compilation compilation,
             InterproceduralAnalysisKind defaultInterproceduralAnalysisKind,
+#pragma warning disable CA1801 // Review unused parameters - Public API
```
mavasani, line 46: Its fine to make a breaking public change for this package. I'd remove the unused parameter.

mavasani (approved):
LGTM. Just a minor suggestion.

src/Utilities.UnitTests/FlowAnalysis/Analysis/PropertySetAnalysis/PropertySetAnalysisTests.cs
```diff
@@ -74,80 +74,77 @@ public PropertySetAnalysisParameters(string typeToTrack, ConstructorMapper const
             Debug.Assert(success);
             Debug.Assert(cfg != null);
 
-            using (var cancellationSource = new CancellationTokenSource())
```
Youssef1313, line 77: Hide whitespace changes for easier diff.

Youssef1313:
> I don't see codefix is being offered locally :(

Closed and reopened the code file in VS and the codefix is shown now. Fixing.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Publish/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -8,7 +8,6 @@
     Microsoft.NetCore.Analyzers.Publish.AvoidAssemblyLocationInSingleFile,
     Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
 using static Analyzer.Utilities.MSBuildPropertyOptionNames;
-using static Microsoft.NetCore.Analyzers.Publish.AvoidAssemblyLocationInSingleFile;
```
Youssef1313: @mavasani

- IDE0005 (remove unnecessary imports) seems to be enabled for product code but disabled for tests. I ran a fix all so all files with unused usings got fixed (the number isn't large), I'm assuming this is okay?
- For this specific using in this test file, it's reported as unused. But removing it results in a compile-time error. I don't understand why there is a compile-time error after removing this `using static`. Note that this error occurs only in CI. Removing the using statement doesn't result in any errors on my machine.

mavasani: Can you please revert unrelated changes from this PR? Also the primary reason to disable this rule for test files is to enable us to have stub test files with required usings to help quick start writing tests. Let’s move this change and discussion to a separate PR

Youssef1313: Fixed

Issue #4133, opened 2020-09-05
Circular removing-adding of record's parameter


WhiteBlackGoose:
PublicApiAnalyzer v3.3.0.

### Description

When adding a parameter to a record, it suggests adding its Initter to the public API. After you you add the Setter, it suggests you to add a Getter **instead** of the Initter.
![image](https://user-images.githubusercontent.com/31178401/92299300-df8e6380-ef59-11ea-8ebb-be1d063056a1.png)

### Steps To Reproduce

Within a C# 9 project create an empty record with one parameter. Then, having your cursor on the property hit `ctrl + .` and add  a getter to the PublicAPI file. As the error remains, hit it again and see, that it suggests removing the getter and adding an initter

### Expected behavior

Adding the property's getter (initter)

### Actual behavior

Removes the previosly added initter (getter)  and adds getter (initter)

WhiteBlackGoose:
Is this issue irrelevant? I don't get any answers so far... 😞

mavasani:
Need to confirm if this is a dupe.

WhiteBlackGoose:
@mavasani it is still relevant! It still works so with records.

WhiteBlackGoose:
Have been half a year since .net5 release, still nothing changed.

WhiteBlackGoose:
![image](https://user-images.githubusercontent.com/31178401/115970830-b99bea80-a54d-11eb-9932-c57acd85fbfb.png)

PR #5067, opened 2021-04-25
Fix incorrect removal of implicit property from PublicApi

wzchua:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4133

I'm not sure what is the best way to test `record` but I targeted `ReferenceAssemblies.Net.Net50`

### DIFF
--- a/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
+++ b/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
@@ -536,7 +536,7 @@ namespace Microsoft.CodeAnalysis.PublicApiAnalyzers
                         {
                             if (sibling.IsImplicitlyDeclared)
                             {
-                                if (!sibling.IsConstructor())
+                                if (!((sibling as IMethodSymbol)?.MethodKind is MethodKind.Constructor or MethodKind.PropertyGet or MethodKind.PropertySet))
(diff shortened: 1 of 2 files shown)


src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.Impl.cs
```diff
@@ -536,7 +536,7 @@ private string GetSiblingNamesToRemoveFromUnshippedTextCore(ISymbol symbol)
                         {
                             if (sibling.IsImplicitlyDeclared)
                             {
-                                if (!sibling.IsConstructor())
+                                if (!((sibling as IMethodSymbol)?.MethodKind is MethodKind.Constructor or MethodKind.PropertyGet or MethodKind.PropertySet))
```
Youssef1313, line 539: Something like this would be easier to read.

```suggestion
                                if (sibling is not IMethodSymbol { MethodKind: MethodKind.Constructor or MethodKind.PropertyGet or MethodKind.PropertySet })
```

Youssef1313:
> I'm not sure what is the best way to test `record` but I targeted `ReferenceAssemblies.Net.Net50`

Either using Net50 as you did, or explicitly declare `IsExternalInit` dummy class should be okay. (My personal opinion is using ReferenceAssemblies as you did).

mavasani (approved):
Thank you for the contribution!

Issue #5061, opened 2021-04-23
AD0001 Analyzer 'Microsoft.CodeAnalysis.PublicApiAnalyzers.DeclarePublicApiAnalyzer' threw an exception of type 'System.NullReferenceException'


mayerj:
### Analyzer

RS0016: Public API analyzer

### Analyzer source

**NuGet Package**: Microsoft.CodeAnalysis.PublicApiAnalyzers

**Version**: 3.3.2, and tip of main branch

### Describe the bug

Unexpected AD0001 when you have a csproj file that includes the "PublicAPI.Shipped.txt" or "PublicAPI.Unshipped.txt" explicitly via:

```
  <ItemGroup>
    <AdditionalFiles Include="PublicAPI.Shipped.txt" />
    <AdditionalFiles Include="PublicAPI.Unshipped.txt" />
  </ItemGroup>
```

### Steps To Reproduce

Create a new project file using the content below:
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <AdditionalFiles Include="PublicAPI.Shipped.txt" />
    <AdditionalFiles Include="PublicAPI.Unshipped.txt" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.2">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>
</Project>
```

Do not create "PublicAPI.Shipped.txt" or "PublicAPI.Unshipped.txt" on disk.

Compile csproj

### Expected behavior

Either a warning saying the files couldn't be read, or RS0048 "file missing"

### Actual behavior

AD0001 NullReferenceException here:

https://github.com/dotnet/roslyn-analyzers/blob/90660a6b270c5b9a46be6bea6f83885de829f4d2/src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs#L264

(sourceText.Lines was null).

Edit: sourceText was null.

shippedText.GetText(cancellationToken) can return null.

## Additional context

```
System.NullReferenceException
  HResult=0x80004003
[...]

PR #5070, opened 2021-04-26
Prevent AD0001 on missing analyzer file

mayerj:


mavasani (commented):
Do we have an issue tracking this? Are you able to write a regression test?

mayerj:
> Do we have an issue tracking this? Are you able to write a regression test?

This is issue [5061](https://github.com/dotnet/roslyn-analyzers/issues/5061)

I tried writing a test, but have run into some issues.  The "AdditionalFiles" in the unit tests don't support a null value for the text (just not present, or string.empty).  However, the issue is reproducible.

https://github.com/dotnet/roslyn-analyzers/blob/a0522afd52b25ad211298e308fd5bbc7734865e0/src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs#L62

I am still attempting to write the test differently.

mayerj:
@mavasani I'm not sure it's straightforward to write a test.  I need to generate an AdditionalText which returns null from GetText, but it seems like the unit test framework seems to deal in SourceText objects themselves.  Any hints?

Youssef1313 (commented):
Can you add a test? Specifically for https://github.com/dotnet/roslyn-analyzers/issues/5061#issuecomment-825695032.

---

> I'm not sure it's straightforward to write a test. I need to generate an AdditionalText which returns null from GetText, but it seems like the unit test framework seems to deal in SourceText objects themselves. Any hints?

I'm not 100% sure, but I think `build_property.AdditionalFiles` may be helpful to represent an `<AdditionalFiles>` in csproj that doesn't actually exist on disk.

It may look something like:


```csharp

await new VerifyCS.Test
{
    TestState =
    {
        AdditionalFiles = {  }, // Empty that they are not existing on disk.
        AnalyzerConfigFiles = { ("/.globalconfig", @"
is_global = true

build_property.AdditionalFiles = PublicAPI.Shipped.txt
"),
    }
}
```

mayerj:
> Can you add a test? Specifically for [#5061 (comment)](https://github.com/dotnet/roslyn-analyzers/issues/5061#issuecomment-825695032).
> 
> > I'm not sure it's straightforward to write a test. I need to generate an AdditionalText which returns null from GetText, but it seems like the unit test framework seems to deal in SourceText objects themselves. Any hints?
> 
> I'm not 100% sure, but I think `build_property.AdditionalFiles` may be helpful to represent an `<AdditionalFiles>` in csproj that doesn't actually exist on disk.
> 
> It may look something like:
> 
> ```cs
> await new VerifyCS.Test
> {
>     TestState =
>     {
>         AdditionalFiles = {  }, // Empty that they are not existing on disk.
>         AnalyzerConfigFiles = { ("/.globalconfig", @"
> is_global = true
> 
> build_property.AdditionalFiles = PublicAPI.Shipped.txt
> "),
>     }
> }
> ```

I don't think so?  I thought that that functionality was to move the CSPROJ data into the analyzers, but I need the file provided as an AdditionalFile in analyzerOptions.AdditionalFiles.  The analyzer isn't looking for additional files in the editorconfig data (and at least my brief testing didn't translate the snippet above into an analyzerOptions.AdditionalFiles entry, is it supposed to?

Youssef1313:
I'll defer to @sharwell if you tried that and it didn't work. My understanding is that `build_property.*` translates as if the property is present in csproj.

sharwell:
Writing a test requires a workspace implementation that believes a file exists only to find that it doesn't actually exist when you try to read it. It might be possible through a solution transform, but I wouldn't count on it.

mayerj:
> Writing a test requires a workspace implementation that believes a file exists only to find that it doesn't actually exist when you try to read it. It might be possible through a solution transform, but I wouldn't count on it.

I'm certainly willing to do it, but I will need some pointers if that is the desire.  I did play with the solution transform, and it doesn't seem like there is a way of getting at the AdditionalFiles list to add directly to it.  I can only add SourceText objects, which are checked for null prior to the test running.

sharwell:
If you wanted to try it, you could do something like this:

```csharp
await new VerifyCS.Test
{
  // other properties ...
  SolutionTransforms =
  {
    (solution, projectId) =>
    {
      var id = DocumentId.CreateNewId(projectId, "non-existent file");
      var fileName = Path.GetRandomFileName();
      var filePath = Path.Combine(Path.GetTempPath(), fileName);
      return solution.AddAdditionalDocument(DocumentInfo.Create(id, fileName, loader: new FileTextLoader(filePath, Encoding.UTF8), filePath: filePath));
    },
  },
}.RunAsync();
```

mayerj:
> ```cs
> await new VerifyCS.Test
> {
>   // other properties ...
>   SolutionTransforms =
>   {
>     (solution, projectId) =>
>     {
>       var id = DocumentId.CreateNewId(projectId, debugName: "non-existent file");
>       var fileName = Path.GetRandomFileName();
>       var filePath = Path.Combine(Path.GetTempPath(), fileName);
>       return solution.AddAdditionalDocument(DocumentInfo.Create(id, fileName, loader: new FileTextLoader(filePath, Encoding.UTF8), filePath: filePath));
>     },
>   },
> }.RunAsync();
> ```

I thought I had it, but this doesn't throw the NullReferenceException.  It seems like in the test this results in an "AdditionalTextWithState" object to be returned in AdditionalFiles.  However, in VS2019/the compiler, it returns an "AdditionalTextFile".

The AdditionalTextWithState has some additional logic that returns a "failed" text if it can't  read the document.

https://github.com/dotnet/roslyn/blob/315c2e149ba7889b0937d872274c33fcbfe9af5f/src/Workspaces/Core/Portable/Workspace/Solution/TextLoader.cs#L60

The AdditionalTextFile does seem to hint that it might return null.

https://github.com/dotnet/roslyn/blob/3aa88b7957cbb4dedcc03f2fa58f79bf9a68538a/src/Compilers/Core/Portable/AdditionalTextFile.cs#L53

sharwell:
💭 It almost seems like this should be fixed by updating the compiler to avoid exposing these null values to the analyzers.

mayerj:
> I don't see a way to bypass this difference without private reflection.
> 
> 💭 It almost seems like this should be fixed by updating the compiler to avoid exposing these null values to the analyzers.

Agreed, though it does seem intentional that it can return null.

https://github.com/dotnet/roslyn/blob/aa91123c53d9a84daf6053c69e88fff255c236d4/src/Compilers/Core/Portable/DiagnosticAnalyzer/AdditionalText.cs#L24

What is next for the PR?

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
```diff
@@ -318,8 +318,8 @@ private static bool TryGetApiData(AnalyzerOptions analyzerOptions, Compilation c
                 return false;
             }
 
-            shippedData = ReadApiData(shippedText.Path, shippedText.GetText(cancellationToken), isShippedApi: true);
-            unshippedData = ReadApiData(unshippedText.Path, unshippedText.GetText(cancellationToken), isShippedApi: false);
+            shippedData = ReadApiData(shippedText.Value.path, shippedText.Value.text, isShippedApi: true);
+            unshippedData = ReadApiData(unshippedText.Value.path, unshippedText.Value.text, isShippedApi: false);
```
sharwell, line 322: ❔ Is this better than just updating `ReadApiData` to check the argument for null?

sharwell:
@mayerj I have no objection to taking this change without tests. I wasn't really expecting them to start with, but we still gave it a solid shot. I did have question about the value of this "more complicated" approach over a simpler approach that involves a null check. If you have a reason to prefer the current form feel free to respond with an explanation and not feel obligated to make a change.

mavasani:
Agree with @sharwell - thank you @mayerj for putting in the hard yards to create a test. We should be good to merge once @sharwell approves.

src/PublicApiAnalyzers/Core/Analyzers/DeclarePublicApiAnalyzer.cs
mayerj: The reason I went with this approach was that I wanted to treat the file being configured but not available on disk the same as if it was unconfigured (IE, it would respect silentlyBailOutOnMissingApiFiles if they were both missing, but give you PublicApiFileMissing if it was just 1).  So TryGetApiText needs to return false if shippedText.GetText(...) is going to return null.  At that point I just figured there was no reason to call it twice, so I just returned the path and content.

mavasani (approved):

Issue #4428, opened 2020-11-13
CA1416: Consider supporting guard clauses like IsThreadStartSupported


buyaa-n:
Consider supporting guard clauses like IsThreadStartSupported in Platform compatibility analyzer

I'd have hoped the analyzer would be able to understand the guard clauses like IsThreadStartSupported just as the linker can.  Is that not possible?

_Originally posted by @stephentoub in https://github.com/dotnet/runtime/pull/43363#discussion_r520598687_

Evangelink:
@buyaa-n could you please give a little more detail about the issue? As far as I understand the idea is to fix a false positive with the rule CA1416, is it more global? If not, maybe the title should state the rule and the label might be `False-Positive` instead of `Enhancement`.

buyaa-n:
> @buyaa-n could you please give a little more detail about the issue? As far as I understand the idea is to fix a false positive with the rule CA1416, is it more global? If not, maybe the title should state the rule and the label might be False-Positive instead of Enhancement.

It is not a fix for a false positive, it would be an enhancement for the analyzer to support this scenario. Yes, it is related to CA1416 only, updated the title, thanks!

buyaa-n:
Fixed with https://github.com/dotnet/roslyn-analyzers/pull/5087

PR #5087, opened 2021-05-05
CA1416 Support custom guard members annotated with guard attributes

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/4428 Related to https://github.com/dotnet/runtime/issues/51541 and https://github.com/dotnet/runtime/issues/44922#issuecomment-789212369

The CA1416 Platform Compatibility analyzer already recognizes platform guards using the methods on OperatingSystem/RuntimeInformation, such as OperatingSystem.IsWindows and OperatingSystem.IsWindowsVersionAtLeast. However, the analyzer does not recognize other guard options like when a field, property, or helper methods cached the platform check result and used as platform guards. 

Expanding this support involved adding new attributes `SupportedOSPlatformGuard` and `UnsupportedOSPlatformGuard` and annotated the custom guards with those attributes with the corresponding platform and/or version. Now this annotation needs to be recognized and respected in the Platform Compatibility analyzer flow analysis

## Usage Examples
`Thread.IsThreadStartSupported` indicates whether or not the current platform supports calls to `Thread.Start()`, which is not supported on `browser`. So in this `[UnsupportedOSPlatformGuard]` attributes can be used for this field and the analyzer will recognize it as a platform guard.

```cs 
#if !TARGET_BROWSER
    [UnsupportedOSPlatformGuard("browser")] // The platform guard atribute
    internal const bool IsThreadStartSupported = true;
#endif

protected internal override void QueueTask(Task task)
{
    if (Thread.IsThreadStartSupported) // This will be evaluated same as `!OperatingSystem.IsBrowser()`
    {
        new Thread(s_longRunningThreadWork)
        {
            IsBackground = true,
            Name = ".NET Long Running Task"
        }.UnsafeStart(task);      // Call unsupported on browser APIs safely (without the attribute it would warn here)
    }
}
```

Example with helper method:

```cs
[SupportedOSPlatformGuard("windows10.0")]
public bool  IsWindows10OrAbove()
{ 
     return OperatingSystem.IsWindowsVersionAtLeast(10);
}

[SupportedOSPlatform("windows10.0")]
public bool  Windows10OnlyAPI() { }

void Sample ()
{
    if (IsWindows10OrAbove()) // work same as OperatingSystem.IsWindowsVersionAtLeast(10)
    {
        Windows10OnlyAPI();
    }
}
```
For guarding multiple platforms need to apply the attribute for each platform
```cs
class Test
{
    [SupportedOSPlatformGuard(""linux"")]
    [SupportedOSPlatformGuard(""macOS"")]
    [SupportedOSPlatformGuard(""Windows"")]
    private readonly bool _http3Enabled = OperatingSystem.IsLinux() || OperatingSystem.IsWindows() || OperatingSystem.IsMacOS();

    void M1()
    {
        if (_http3Enabled)
        {
            ApiWorkOnWindowsLinuxMac();  // Not warn
        }
        else
        {
            ApiWorkOnWindowsLinuxMac();  // Warns 
        }
    }

    void ApiWorkOnWindowsLinuxMac() { }
}
```

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Immutable;
 using System.Linq;
+using Analyzer.Utilities.Extensions;
 using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.GlobalFlowStateAnalysis;
@@ -26,25 +27,31 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 _osPlatformType = osPlatformType;
             }
 
-            internal static bool TryParseGuardAttributes(ImmutableArray<AttributeData> attributes, ref GlobalFlowStateAnalysisValueSet value)
+            internal static bool TryParseGuardAttributes(ISymbol symbol, ref GlobalFlowStateAnalysisValueSet value)
             {
-                if (HasAnyGuardAttribute(attributes))
-                {
-                    using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
-                    if (PlatformMethodValue.TryDecode(attributes, infosBuilder))
-                    {
-                        for (var i = 0; i < infosBuilder.Count; i++)
-                        {
-                            var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
-                            value = i == 0 ? newValue : infosBuilder[i].Negated ? value.WithAdditionalAnalysisValues(newValue, false) :
-                                GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
-                        }
+                var attributes = symbol.GetAttributes();
 
-                        return true;
+                if (symbol.GetMemberType()!.SpecialType != SpecialType.System_Boolean ||
+                    !HasAnyGuardAttribute(attributes))
+                {
+                    return false;
+                }
+
+                using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                if (PlatformMethodValue.TryDecode(attributes, infosBuilder))
+                {
+                    for (var i = 0; i < infosBuilder.Count; i++)
+                    {
+                        var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
+                        // if the incoming value is negated it should be merged with AND logic, else with OR. 
+                        value = i == 0 ? newValue : infosBuilder[i].Negated ? value.WithAdditionalAnalysisValues(newValue, false) :
+                            GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
                     }
 
-                    value = GlobalFlowStateAnalysisValueSet.Unknown;
+                    return true;
                 }
+
+                value = GlobalFlowStateAnalysisValueSet.Unknown;
 
                 return false;
 
@@ -62,28 +69,25 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             {
                 var value = base.VisitInvocation_NonLambdaOrDelegateOrLocalFunction(method, visitedInstance, visitedArguments, invokedAsDelegate, originalOperation, defaultValue);
 
-                if (method.ReturnType.SpecialType == SpecialType.System_Boolean)
+                if (_platformCheckMethods.Contains(method.OriginalDefinition))
                 {
-                    if (_platformCheckMethods.Contains(method.OriginalDefinition))
+                    using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                    if (PlatformMethodValue.TryDecode(method, visitedArguments, DataFlowAnalysisContext.ValueContentAnalysisResult, _osPlatformType, infosBuilder))
                     {
-                        using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
-                        if (PlatformMethodValue.TryDecode(method, visitedArguments, DataFlowAnalysisContext.ValueContentAnalysisResult, _osPlatformType, infosBuilder))
+                        for (var i = 0; i < infosBuilder.Count; i++)
                         {
-                            for (var i = 0; i < infosBuilder.Count; i++)
-                            {
-                                var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
-                                value = i == 0 ? newValue : GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
-                            }
-
-                            return value;
+                            var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
+                            value = i == 0 ? newValue : GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
                         }
 
-                        return GlobalFlowStateAnalysisValueSet.Unknown;
-                    }
-                    else if (TryParseGuardAttributes(method.GetAttributes(), ref value))
-                    {
                         return value;
                     }
+
+                    return GlobalFlowStateAnalysisValueSet.Unknown;
+                }
+                else if (TryParseGuardAttributes(method, ref value))
+                {
+                    return value;
                 }
 
                 return value;
@@ -93,8 +97,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             {
                 var value = base.VisitFieldReference(operation, argument);
 
-                if (operation.Type.SpecialType == SpecialType.System_Boolean &&
-                    TryParseGuardAttributes(operation.Field.GetAttributes(), ref value))
+                if (TryParseGuardAttributes(operation.Field, ref value))
                 {
                     return value;
                 }
@@ -105,8 +108,8 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
             public override GlobalFlowStateAnalysisValueSet VisitPropertyReference(IPropertyReferenceOperation operation, object? argument)
             {
                 var value = base.VisitPropertyReference(operation, argument);
-                if (operation.Type.SpecialType == SpecialType.System_Boolean &&
-                    TryParseGuardAttributes(operation.Property.GetAttributes(), ref value))
+
+                if (TryParseGuardAttributes(operation.Property, ref value))
                 {
                     return value;
                 }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System.Collections.Immutable;
+using System.Linq;
 using Analyzer.Utilities.PooledObjects;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.GlobalFlowStateAnalysis;
@@ -25,6 +26,32 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 _osPlatformType = osPlatformType;
             }
 
+            internal static bool TryParseGuardAttrbiutes(ISymbol symbol, ref GlobalFlowStateAnalysisValueSet value)
+            {
+                if (HasAnyGuardAttribute(symbol))
+                {
+                    using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                    if (PlatformMethodValue.TryDecode(symbol.GetAttributes(), infosBuilder))
+                    {
+                        for (var i = 0; i < infosBuilder.Count; i++)
+                        {
+                            var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
+                            value = i == 0 ? newValue : infosBuilder[i].Negated ? value.WithAdditionalAnalysisValues(newValue, false) :
+                                GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
+                        }
+
+                        return true;
+                    }
+
+                    value = GlobalFlowStateAnalysisValueSet.Unknown;
+                }
+
+                return false;
+
+                static bool HasAnyGuardAttribute(ISymbol symbol) =>
+                    symbol.GetAttributes().Any(a => a.AttributeClass.Name is SupportedOSPlatformGuardAttribute or UnsupportedOSPlatformGuardAttribute);
+            }
+
             public override GlobalFlowStateAnalysisValueSet VisitInvocation_NonLambdaOrDelegateOrLocalFunction(
                 IMethodSymbol method,
                 IOperation? visitedInstance,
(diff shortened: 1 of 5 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
```diff
@@ -22,21 +22,19 @@ public sealed partial class PlatformCompatibilityAnalyzer
...
-            private PlatformMethodValue(string invokedPlatformCheckMethodName, string platformPropertyName, Version version, bool negated)
+            internal PlatformMethodValue(string platformPropertyName, Version version, bool negated)
             {
-                InvokedMethodName = invokedPlatformCheckMethodName ?? throw new ArgumentNullException(nameof(invokedPlatformCheckMethodName));
                 PlatformName = platformPropertyName ?? throw new ArgumentNullException(nameof(platformPropertyName));
                 Version = version ?? throw new ArgumentNullException(nameof(version));
                 Negated = negated;
             }
 
-            public string InvokedMethodName { get; }
```
buyaa-n, line 33: This property no really used in any logic except `ToString(), Equals()` overloads

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -4315,6 +4315,58 @@ void M1()
...
+using System;
+using System.Runtime.Versioning;
+
+class Test
+{
+    void M1()
+    {
+        [|UnsupportedBrowserType.M1()|];
+        [|UnsupportedBrowserType.M2()|];
+        if (UnsupportedBrowserType.IsSupported)
```
buyaa-n, line 4331: Handling  https://github.com/dotnet/runtime/issues/49755, when a type is annotated with platform attribute and it has a boolean member used for platform check, we can apply the guard attribute for that member so that it can be used as a guard plus we know that we don't need to warn for this member reference

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
```diff
@@ -109,6 +107,26 @@ public IAbstractAnalysisValue GetNegatedValue()
...
+                foreach (var attribute in attributes)
+                {
+                    if (attribute.AttributeClass.Name is SupportedOSPlatformGuardAttribute or UnsupportedOSPlatformGuardAttribute &&
+                        HasNonEmptyStringArgument(attribute, out var argument) &&
+                        TryParsePlatformNameAndVersion(argument, out string platformName, out Version? version))
+                    {
+                        if (platformName.Equals(OSX, StringComparison.OrdinalIgnoreCase))
+                        {
+                            platformName = macOS;
+                        }
```
jeffhandley, line 121: I recommend moving this into `TryParsePlatformNameAndVersion` or the `PlatformMethodValue` constructor. We should limit the number of places this logic exists.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1590,16 +1623,11 @@ static Versions NormalizeAttribute(Versions attributes)
...
         {
-            if (!attribute.ConstructorArguments.IsEmpty &&
-                                attribute.ConstructorArguments[0] is { } argument &&
-                                argument.Kind == TypedConstantKind.Primitive &&
-                                argument.Type.SpecialType == SpecialType.System_String &&
-                                !argument.IsNull &&
-                                !argument.Value.Equals(string.Empty) &&
-                                TryParsePlatformNameAndVersion(argument.Value.ToString(), out string platformName, out Version? version))
+            if (HasNonEmptyStringArgument(attribute, out var argument) &&
+                TryParsePlatformNameAndVersion(argument, out var platformName, out var version))
```
jeffhandley, line 1627: Is it worth creating a helper for `HasNonEmptyStringArgument(attribute, out var argument) && TryParsePlatsformNameAndVersion(argument, out var platformName, out var version)` since it's used in a couple places?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -3945,6 +3945,454 @@ bool IsWindows11OrLater()
...
+        if (IsLunuxSupported()) // one of the support guarded, so no warning
+        {
+            SupportedOnWindowsLinuxOsx();    
+            SupportedOnLinux();
+        }
+
+        if (_http3Enabled)
+        {
+            //SupportedOnWindowsLinuxOsx();  
+            [|SupportedOnLinux()|]; // only supported on linux but call site is reachable on 3, linus, windows, macos
```
jeffhandley, line 3993: Nit: typo
```suggestion
            [|SupportedOnLinux()|]; // only supported on linux but call site is reachable on 3, linux, windows, macos
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -3945,6 +3945,454 @@ bool IsWindows11OrLater()
...
+        }
+        else
+        {
+            [|SupportedOnWindowsLinuxOsx()|]; // This call site is reachable on all platforms. 'Test.SupportedOnWindowsLinuxOsx()' is only supported on: 'macOS/OSX', 'Linux', 'windows'.
+            {|#0:SupportedOnLinux()|}; // This call site is reachable on all platforms. 'Test.SupportedOnLinux()' is only supported on: 'linux'.
+        }
+
+        if (IsHttp3PlusAndroidEnabled()) // Android is not in the support list
```
jeffhandley, line 4001: `IsHttp3PlusAndroidEnabled` _does_ have a support attribute for Android. I don't understand this test.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -3945,6 +3945,454 @@ bool IsWindows11OrLater()
...
+    [SupportedOSPlatform(""linux"")]
+    void SupportedOnLinux() { }
+}" + MockAttributesCsSource;
+
+            await VerifyAnalyzerAsyncCs(source, s_msBuildPlatforms,
+                VerifyCS.Diagnostic(PlatformCompatibilityAnalyzer.OnlySupportedCsAllPlatforms).WithLocation(0).
+                WithArguments("Test.SupportedOnLinux()", "'linux'"));
+        }
+
+        [Fact]
```
jeffhandley, line 4032: This is where my review has left off so far...

jeffhandley (commented):
I'm still reviewing, but I wanted to go ahead and post the comments I have so far.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
```diff
@@ -25,6 +26,32 @@ private sealed class OperationVisitor : GlobalFlowStateDataFlowOperationVisitor
                 _osPlatformType = osPlatformType;
             }
 
+            internal static bool TryParseGuardAttrbiutes(ISymbol symbol, ref GlobalFlowStateAnalysisValueSet value)
+            {
+                if (HasAnyGuardAttribute(symbol))
+                {
+                    using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                    if (PlatformMethodValue.TryDecode(symbol.GetAttributes(), infosBuilder))
```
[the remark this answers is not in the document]
buyaa-n: sure

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
jeffhandley: To make sure I understand the behavior here.

* This call site is reachable on all platforms, including browser
* `UnsupportedBrowserType` is unsupported on browser
* `UnsupportedBrowserType.IsSupported` is a static method that indicates whether or not `UnsupportedBrowserType` is supported on the current platform
    * Its return value would be intended to return `true` for all platforms except browser
    * That could be done either through `#ifdef` code blocks in the source, or by having `public static bool IsSupported { get => !OperatingSystem.IsBrowser(); }`
* The analyzer now recognizes that the member you're accessing is a guard member, and it respects that guard's platforms while evaluating the types referenced to reach that guard
* The result is that you can:
    * Access the guard member itself without getting any analyzer warnings
    * Access members within the guard, with the guard's assertions applying
    * And therefore not get any diagnostic messages in the scenario captured here

Is that all correct?

buyaa-n: Right, all correct

jeffhandley: Awesome; thanks! This feels like a powerful pattern we can promote. Platform-specific types can have static `IsSupported` methods/properties annotated with the guards that match the platforms the type is annotated with. Then code referencing the APIs only has to call into the `IsSupported` member, and all other guarding is taken care of.

/cc @marek-safar @terrajobst

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: I think the helper would need a long name, something like `TryParsePlatsformNameAndVersionForNonEmptyStringArgument`, isn't the existing condition is more readable/understandable?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: Yes, meant the Android support is extra for guarding `SupportedOnWindowsLinuxOsx()` which has not Android support

jeffhandley: Got it; thanks.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
jeffhandley: It could just be an overload of `TryParsePlatformNameAndVersion`.

```csharp
        private static bool TryParsePlatformNameAndVersion(AttributeData attribute, out string osPlatformName, [NotNullWhen(true)] out Version? version)
        {
            if (HasNonEmptyStringArgument(attribute, out string? argument))
            {
                return TryParsePlatformNameAndVersion(argument, out osPlatformName, out version);
            }

            version = null;
            osPlatformName = string.Empty;

            return false;
        }
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -3945,6 +3945,454 @@ bool IsWindows11OrLater()
...
+
+    [SupportedOSPlatformGuard(""linux"")]
+    [SupportedOSPlatformGuard(""macOS"")]
+    [SupportedOSPlatformGuard(""Windows"")]
+    [UnsupportedOSPlatformGuard(""Android"")]
+    private readonly bool _http3EnabledNotAndroid;
+
+    void M1()
+    {
+        if (IsLunuxSupported()) // one of the support guarded, so no warning
```
mavasani, line 3984: Just trying to understand the desired semantics here - so we are not actually performing flow analysis on the called method, `IsLunuxSupported` here, but just decoding the new platform guard attributes on the method to drive the analysis, correct? Basically, if I were to change this test case so `IsLunuxSupported` returned `false` instead of `true` and made no other changes, will the analysis + diagnostics stay exactly the same?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
```diff
@@ -25,6 +26,32 @@ private sealed class OperationVisitor : GlobalFlowStateDataFlowOperationVisitor
...
+            {
+                if (HasAnyGuardAttribute(attributes))
+                {
+                    using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                    if (PlatformMethodValue.TryDecode(attributes, infosBuilder))
+                    {
+                        for (var i = 0; i < infosBuilder.Count; i++)
+                        {
+                            var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
+                            value = i == 0 ? newValue : infosBuilder[i].Negated ? value.WithAdditionalAnalysisValues(newValue, false) :
```
mavasani, line 39: > infosBuilder[i].Negated ? value.WithAdditionalAnalysisValues(newValue, false)

Can you explain the reasoning behind this logic? This makes me feel uncomfortable, as `WithAdditionalAnalysisValues` was only designed to add to current global state when a new runtime platform guard is encountered. I am not able to figure out why we use `WithAdditionalAnalysisValues` for negated value but do the merge for non-negated value.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
```diff
@@ -35,25 +62,57 @@ private sealed class OperationVisitor : GlobalFlowStateDataFlowOperationVisitor
...
-                        for (var i = 0; i < infosBuilder.Count; i++)
+                        using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
+                        if (PlatformMethodValue.TryDecode(method, visitedArguments, DataFlowAnalysisContext.ValueContentAnalysisResult, _osPlatformType, infosBuilder))
                         {
-                            var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
-                            value = i == 0 ? newValue : GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
+                            for (var i = 0; i < infosBuilder.Count; i++)
+                            {
+                                var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
+                                value = i == 0 ? newValue : GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
```
mavasani, line 75: The inconsistency between the code here and the newly added one above which uses `WithAdditionalAnalysisValues` for negated values makes me cringe a bit more. IMO, the analysis should be identical when decoding method attributes at callsites, regardless of whether it is a well-known platform check method OR non-well known method with the new guard attributes.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
buyaa-n: `UnsupportedOSPlatformGuard` attributes parsed to `Negated` value when a member has a list of 
Unsupported guard attributes those need to be merged as AND logic, for example:
```cs
    [UnsupportedOSPlatformGuard("linux")]
    [UnsupportedOSPlatformGuard("Windows")]
    bool LinuxAndWindowsNotSupported { get; } = !OperatingSystem.IsLinux() && !OperatingSystem.IsWindows()
``` 
As you can see It should work same as checking condition `!OperatingSystem.IsLinux() && !OperatingSystem.IsWindows()`, for that emerging with `WithAdditionalAnalysisValues` worked well 

For Supported platform guards they need to be merged with OR, for `Merge` worked as needed
```cs
    [SupportedOSPlatformGuard("linux")]
    [SupportedOSPlatformGuard("Windows")]
    bool LinuxAndWindowsSupported { get; set; } = OperatingSystem.IsLinux() || OperatingSystem.IsWindows()
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: > Just trying to understand the desired semantics here - so we are not actually performing flow analysis on the called method, IsLunuxSupported here, but just decoding the new platform guard attributes on the method to drive the analysis, correct

Correct

> Basically, if I were to change this test case so IsLunuxSupported returned false instead of true and made no other changes, will the analysis + diagnostics stay exactly the same?

Right, no matter what value the method returns explicitly, only matter what attribute it has with what platform and or version, then it can guard corresponding platform-specific APIs on its `true` side
```cs
[SupportedOSPlatformGuard("linux")]
private bool IsLunuxSupported() => false; // same guard
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
buyaa-n: It could be identical if annotation made same as guard methods, i, e, only check if a platform is supported. But in this case we have guards for Unsupported which i don't think we can efficiently express same as existing Guard methods, well we could somehow express them same us guard methods but that would be much more complicated

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
mavasani: Sounds good, thanks!

jeffhandley (approved):

Issue #5023, opened 2021-04-15
[CA1416] Allow child APIs attribute override assembly attribute in targeted builds of cross-platform assemblies


ManickaP:
Internally used `[SupportedOSPlatform]` will not raise any warnings due to build automatically adding assembly-wide `[SupportedOSPlatform("...")]` for the currently build platform.

Repro:
```C#
[SupportedOSPlatform("windows")]
[SupportedOSPlatform("linux")]
[SupportedOSPlatform("macos")]
private async ValueTask<(HttpConnectionBase connection, bool isNewConnection)>
    GetHttp3ConnectionAsync(HttpRequestMessage request, HttpAuthority authority, CancellationToken cancellationToken) { ... }
    
public async ValueTask<(HttpConnectionBase connection, bool isNewConnection)>
    GetConnection(...)
{
   ...
   // This line should cause a warning on Android/tvOS/iOS etc. but doesn't
   return GetHttp3ConnectionAsync(request, authority, cancellationToken);
}
```

## Potential solution:
- the nested `SupportedOS` should behave as intersection --> supported are only OSes named on all levels
- the nested `UnsupportedOS` should behave as union --> unsupported are all OSes from all levels
In other words, narrowing the lists of supported OSes with nested attributes.

**Also consider that the list of operating systems in the attributes is behaving differently from the platform guards.** Example:
```C#
if (OperatingSystem.IsLinux())
{
    // true for Linux distributions as well as Android, see: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/eng/targetframeworksuffix.props
}

[SupportedOSPlatform("Linux")]
public class A
{
    // true only for Linux distributions, not for Android.
}
```
This inconsistency confused me quite a bit since the guards used in the code must use different list of OSes than what is used in the attributes to achieve the same. This might cause some headaches to our customers in the future.

This is a spin off from conversation in https://github.com/dotnet/runtime/pull/49261, see this comment https://github.com/dotnet/runtime/pull/49261#issuecomment-819829388

cc: @marek-safar @terrajobst @jeffhandley @buyaa-n

buyaa-n:
[...]

PR #5094, opened 2021-05-10
CA1416 Add 'CrossPlatform' as special case allowing override parent support

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5023
### Allow child APIs attribute override assembly attribute in targeted builds of cross-platform assemblies

By the design, [child APIs should not extend parent support](https://github.com/dotnet/runtime/issues/43971). SupportedOSPlatform illustrates that the API/assembly only reachable for that platform and it should not have any API supported on other platforms inside it. We have planned to warn if it happens by https://github.com/dotnet/runtime/issues/43971, but this scenario is happening in runtime for targeted cross-platform builds. Ideally, all platform-specific implementation should be separated from cross-platform src. Though that causing code duplication at some level, so most developers would not like to apply strict separation of code. 

Even though it is an expected scenario by design it is unfortunate to not get any warning about this inconsistency or possible reference of incompatible APIs within the same assembly

After some discussion, we decided to add a special case `[SupportedOSPlatform("CrossPlatofrm")]` attribute for targeted builds of cross-platform assembly to allow child member attributes to override the parent. To distinguish between real platform-specific build and cross-platform with platform-specific builds. (We might change `"CrossPlatofrm"` into `"Neutral"` or something like that)

CC @jeffhandley @terrajobst


jeffhandley (approved):

buyaa-n:
Using `SupportedOSPlatform` attribute for handling this is causing side effects in further analysis and needs more hack like fixes to get rid of it. I realize we can use just an MSBuild property for this and raised a [new PR](https://github.com/dotnet/roslyn-analyzers/pull/5117) with implementation, so closing this PR

PR #5102, opened 2021-05-14
#5072 ca1305 ignore nullable invariant types

### LINKED ISSUE #5072: Don't require IFormatProvider for nullable versions of invariant types
### Analyzer

**Diagnostic ID**: [CA1305](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1305)

### Describe the improvement

Should expand the special cases on where the rule doesn't need to be fired to include Nullable versions of the types in play.

There is a list of types at 
https://github.com/dotnet/roslyn-analyzers/blob/a31de86f7752846dceeb1064e07fe187592e090e/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs#L90-L95 for which the rule knows not to fire.  It should also not fire for Nullable versions of those types.

### Additional context

https://github.com/dotnet/aspnetcore/pull/32143

NewellClark:
Fix #5072


AndrewZu1337:
@NewellClark  the your tests have been passed without fix. Looks like there isn't problem in this analyzer.

NewellClark:
> @NewellClark the your tests have been passed without fix. Looks like there isn't problem in this analyzer.

I did fix the issue brought up in #5072. I made two commits: one where I added failing tests, and another where I fixed the problem.

AndrewZu1337:
> > @NewellClark the your tests have been passed without fix. Looks like there isn't problem in this analyzer.
> 
> @AndrewZu1337 I did fix the issue brought up in #5072. I made two commits: one where I added failing tests, and another where I fixed the problem.

I mean I ran the your failing tests without your fix and it's passed. So the analyzer handles `Nullable` versions of  types correctly without this fix.

NewellClark:
@AndrewZu1337 I ran the tests before implementing a fix, and they failed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -86,12 +86,19 @@ protected override void InitializeWorker(CompilationStartAnalysisContext context
             var charType = context.Compilation.GetSpecialType(SpecialType.System_Char);
             var boolType = context.Compilation.GetSpecialType(SpecialType.System_Boolean);
             var guidType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemGuid);
+            var nullableTType = context.Compilation.GetSpecialType(SpecialType.System_Nullable_T);
+            INamedTypeSymbol? GetNullableType(INamedTypeSymbol? typeSymbol) => typeSymbol is not null ? nullableTType?.Construct(typeSymbol) : null;
```
mavasani, line 90: Kindly move the local function towards the end of the method.

mavasani (approved):
Please retarget to `main` branch.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProviderTests.cs
```diff
@@ -945,6 +945,36 @@ End Class
 ");
         }
 
+        [Fact]
+        public async Task CA1305_NullableInvariantTypes_NoDiagnostic()
```
Youssef1313, line 949: Can you add a workitem here?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
```diff
@@ -86,12 +86,19 @@ protected override void InitializeWorker(CompilationStartAnalysisContext context
...
 
             var builder = ImmutableHashSet.CreateBuilder<INamedTypeSymbol>();
+
             builder.AddIfNotNull(charType);
+            builder.AddIfNotNull(GetNullableType(charType));
             builder.AddIfNotNull(boolType);
+            builder.AddIfNotNull(GetNullableType(boolType));
             builder.AddIfNotNull(stringType);
             builder.AddIfNotNull(guidType);
+            builder.AddIfNotNull(GetNullableType(guidType));
```
Youssef1313, line 100: Would it be better to do the fix in `IsValidToStringCall`? (i.e, instead of constructing `char?`, `bool?`, and `Guid?`, just try to find `Nullable<SomeType>` and see if `SomeType` is in the hashset.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
NewellClark: I ended up removing the local function.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyIFormatProvider.cs
NewellClark: Good idea. I went ahead and did that.

mavasani (approved):

PR #5104, opened 2021-05-15
#4775 Use case-insensitive identifier comparisons in VB

### LINKED ISSUE #4775: Comparing identifiers should be done case insensitively in VB
Two fixers are comparing identifiers case-sensitively in VB, which can lead to wrong results. These should be fixed and tested.

https://github.com/dotnet/roslyn-analyzers/blob/33582d0369adb7657731b611f4811e981dcb2f7f/src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb#L30

https://github.com/dotnet/roslyn-analyzers/blob/33582d0369adb7657731b611f4811e981dcb2f7f/src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb#L49

https://github.com/dotnet/roslyn-analyzers/blob/33582d0369adb7657731b611f4811e981dcb2f7f/src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicPreferStreamAsyncMemoryOverloads.Fixer.vb#L99

https://github.com/dotnet/roslyn-analyzers/blob/b148fa89c974a72c28184931767835709f34d1fa/src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/InteropServices/BasicSpecifyMarshalingForPInvokeStringArguments.Fixer.vb#L25

NewellClark:
Fix #4775

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/SpecifyMarshalingForPInvokeStringArgumentsTests.Fixer.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/SpecifyMarshalingForPInvokeStringArgumentsTests.Fixer.cs
@@ -144,7 +144,7 @@ Class C
     Private Shared Sub SomeMethod2(s As String)
     End Sub
 
-    <{|CA2101:DllImport(""user32.dll"", CharSet:=CharSet.Unicode)|}>
+    <DllImport(""user32.dll"", CharSet:=CharSet.Unicode)>
     Private Shared Sub SomeMethod3(s As String)
     End Sub
 End Class
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/SpecifyMarshalingForPInvokeStringArgumentsTests.Fixer.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/SpecifyMarshalingForPInvokeStringArgumentsTests.Fixer.cs
@@ -126,6 +126,10 @@ Class C
     <{|CA2101:DllImport(""user32.dll"", CharSet:=CharSet.Ansi)|}>
     Private Shared Sub SomeMethod2(s As String)
     End Sub
+
+    <{|CA2101:DllImport(""user32.dll"", CHARSET:=CharSet.Ansi)|}>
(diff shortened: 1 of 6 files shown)


mavasani (approved):

mavasani:
@NewellClark Can you retarget to main branch? Note the branch requirements:
1. Bug fixes: Main branch is okay if the underlying analyzer is already in the main branch. Otherwise, if this is a fix for a new .NET6 analyzer, please target to `release\6.0.1xx`.
2. New .NET6 analyzer: Needs to target `release\6.0.1xx` branch.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/SpecifyMarshalingForPInvokeStringArgumentsTests.Fixer.cs
```diff
@@ -139,6 +143,10 @@ End Sub
     <DllImport(""user32.dll"", CharSet:=CharSet.Unicode)>
     Private Shared Sub SomeMethod2(s As String)
     End Sub
+
+    <{|CA2101:DllImport(""user32.dll"", CharSet:=CharSet.Unicode)|}>
+    Private Shared Sub SomeMethod3(s As String)
+    End Sub
```
Youssef1313, line 149: I don't see why there is an expected CA2101 warning here? The above `SomeMethod2` looks exactly the same but don't expect violation.

Youssef1313:
If the test passes both before and after 3552e99, then this is (to me at least) an unexpected behavior from the testing library.

cc @sharwell

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/SpecifyMarshalingForPInvokeStringArgumentsTests.Fixer.cs
NewellClark: @Youssef1313 oops, that is a copy-paste error. Good catch. There shouldn't be any markup at all there as that's in the fixed-code. I've gone ahead and fixed it.

PR #5112, opened 2021-05-18
Analyzer to detect RequiresPreviewFeaturesAttribute

pgovind:
Fixes https://github.com/dotnet/runtime/issues/52726

I think this is a good point to get the implementation reviewed. I've borrowed ideas from the PlatformCompatilityAnalyzer implementation.

Pending work (possibly in follow up PRs):
1. Update unit tests to include VB once all the C# test cases have been defined comprehensively
2. Update the implementation to detect:

- Operators such as `==, !=, <, <=, >, >=, +, -, *, /, ++, --, %, &, |, ^, ~, <<, >>`

3. Add a unit test to test reading MSBuild properties.


src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### Changed Rules
```
pgovind, line 2: This is potentially wrong? Maybe it should've been New Rules here instead of Changed Rules

src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md
```diff
@@ -157,6 +157,7 @@ CA2246 | Usage | Info | AssigningSymbolAndItsMemberInSameStatement, [Documentati
 CA2247 | Usage | Warning | DoNotCreateTaskCompletionSourceWithWrongArguments, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2247)
 CA2248 | Usage | Info | ProvideCorrectArgumentToEnumHasFlag, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2248)
 CA2249 | Usage | Info | PreferStringContainsOverIndexOfAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2249)
+CA2250 | Usage | Warning | DetectPreviewFeatureAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2250)
```
mavasani, line 160: Please do no hand edit shipped releases file, this is the reason that the code fix added to Changed Rules table in the unshipped file. This file should be reverted.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,177 @@
...
+                OperationKind.EventReference
+                );
+
+            context.RegisterOperationBlockEndAction(context =>
+            {
+                if (thisAssemblyUsesPreviewFeatures)
+                {
+                    foreach (var symbolAndOperation in requiresPreviewFeaturesSymbols)
+                    {
+                        ISymbol symbol = symbolAndOperation.Key;
```
pgovind, line 77: Delete this line

mavasani:
New .NET6 analyzers should target [release/6.0.1xx](https://github.com/dotnet/roslyn-analyzers/tree/release/6.0.1xx) branch, not `main`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -0,0 +1,558 @@
...
+    {
+        [RequiresPreviewFeatures]
+        public Program()
+        {
+
+        }
+
+        static void Main(string[] args)
+        {
+            [|new Program()|];
```
pgovind, line 161: I'm having some trouble with the unit tests here. The analyzer seems to be invoked multiple times => multiple diagnostics are being triggered on the same line. I'm not sure if this is expected or if I'm doing something wrong?

cc @mavasani

src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md
pgovind: Yup yup, I figured. Fixed now

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+                                                                                      RuleLevel.BuildWarning,
+                                                                                      s_localizableDescription,
+                                                                                      isPortedFxCopRule: false,
+                                                                                      isDataflowRule: false);
+
+        private const string RequiresPreviewFeaturesAttribute = nameof(RequiresPreviewFeaturesAttribute);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        private bool thisAssemblyUsesPreviewFeatures;
```
Youssef1313, line 38: A private instance field doesn't seem correct for diagnostic analyzers.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(context =>
+            {
+                IAssemblySymbol? thisAssembly = context.Compilation.Assembly;
+                ImmutableArray<AttributeData> attributes = thisAssembly.GetAttributes();
+
+                AttributeData? attribute = attributes.FirstOrDefault(x => x.AttributeClass.ToString() == RequiresPreviewFeaturesAttribute);
```
Youssef1313, line 48: Should probably be symbol comparison instead.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
Youssef1313: I believe it's not expected to have multiple diagnostics on the same line.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+                if (thisAssemblyUsesPreviewFeatures)
+                {
+                    foreach (var symbolAndOperation in requiresPreviewFeaturesSymbols)
+                    {
+                        IOperation? value = symbolAndOperation.Value;
+                        if (value != null)
+                        {
+                            context.ReportDiagnostic(value.CreateDiagnostic(Rule));
+                        }
+                    }
```
Youssef1313, line 88: I'm not sure if this is related to the test issue. But could you try breaking after reporting a diagnostic in the loop.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
Youssef1313: @pgovind I ran the test locally, and I'm getting one diagnostic only.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
```
mavasani, line 2: Slightly confused here - https://github.com/dotnet/roslyn-analyzers/blob/release/6.0.1xx/src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md file in `release/6.0.1xx` already has bunch of rules. Did you pull in the latest changes from `release/6.0.1xx` branch?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="DetectPreviewFeaturesDescription" xml:space="preserve">
+    <value>An assembly has to opt into preview features before using them.</value>
+  </data>
+  <data name="DetectPreviewFeaturesMessage" xml:space="preserve">
+    <value>Enable preview features for this assembly</value>
```
mavasani, line 1526: This needs a much more informative message, similar to the descriptive messages that @buyaa-n has for platform compat analyzer diagnostics.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
mavasani: Yes, this field _must_ be removed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            IOperation? value = null;
+            ImmutableArray<AttributeData> attributes = symbol.GetAttributes();
+            foreach (AttributeData attribute in attributes)
+            {
+                string attributeName = attribute.AttributeClass.Name;
+                if (attributeName == RequiresPreviewFeaturesAttribute)
+                {
+                    value = operation;
+                    thisAssemblyUsesPreviewFeatures = true;
```
mavasani, line 170: Yep, this field is the reason for all race conditions that you might be seeing

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+                OperationKind.FieldReference,
+                OperationKind.DelegateCreation,
+                OperationKind.EventReference
+                );
+
+            context.RegisterOperationBlockEndAction(context =>
+            {
+                if (thisAssemblyUsesPreviewFeatures)
+                {
+                    foreach (var symbolAndOperation in requiresPreviewFeaturesSymbols)
```
mavasani, line 81: This is a problematic approach. This dictionary is created in the compilation start action and passed to every operation block start action so it will have new entries whenever a new method is analyzed. The dictionary's key-value pairs are then iterated in each operation block end, which is problematic as an entry added to this dictionary for the first operation block will be reported for every operation block end callback for every method in the compilation. Iterating a compilation wide mutable dictionary in an operation end block callback is not the right approach.

I think you should sync offline with @buyaa-n on the approach she used in platform compat analyzer for similar problem, similar approach might work for you.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
pgovind: Yup yup, the message/title/description will change in the future. We need to work with the docs team to come up with a resource that explains how preview features work. Then we can update the message here to point to that resource. At the moment, I want to get a working implementation up so we can iterate on the final shape for `[RequiresPreviewFeatures]` better.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: I have discussed this analyzer before with @pgovind, that time I thought it will be very similar to PCA but from this implementation and reading the design doc it seems not quite similar to PCA, at least it doesn't look like has any guard operation, therefore, no need flow analysis. If it doesn't need flow analysis i don't think it needs to register an action for `OperationBlockStartAction` and `rOperationBlockEndAction`, seems it only need to analyze `OperationAction`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: Seems its better to make this method return boolean instead which will be passed all the way to the `CheckOperationAttributes` to signal that it found an API having the attribute

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: Looks like you can report diagnostics immediately whenever it is found in the `OperationAction` as no further analysis  needed for the findings while using the cache only for searching an operation or adding a new one

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
pgovind: I hadn't. Fixed now

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: It was actually in multiple tests because I was using fields :) This is fixed now

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,185 @@
...
+                IPropertyReferenceOperation pOperation => pOperation.Property,
+                IFieldReferenceOperation fOperation => fOperation.Field,
+                IDelegateCreationOperation dOperation => dOperation.Type,
+                IEventReferenceOperation eOperation => eOperation.Member,
+                _ => null,
+            };
+
+        private bool TryGetOrCachePreviewFeaturesAttributesOnSymbol(IOperation operation, ISymbol symbol, ConcurrentDictionary<ISymbol, IOperation?> requiresPreviewFeaturesSymbols, bool checkParents)
+        {
+            if (!requiresPreviewFeaturesSymbols.TryGetValue(symbol, out IOperation? existing))
```
Youssef1313, line 105: It looks like the dictionary is only used to check whether we encountered a symbol or not. Consider a HashSet for that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                if (checkParents)
+                {
+                    var containingSymbol = symbol.ContainingSymbol;
+                    // Namespaces do not have attributes
+                    while (containingSymbol is INamespaceSymbol)
+                    {
+                        containingSymbol = containingSymbol.ContainingSymbol;
+                    }
```
Youssef1313, line 114: Can a namespace be contained in a symbol that's not a namespace?
i.e, if we have a namespace here, won't we be getting ContainingSymbol and eventually will end up with null?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -0,0 +1,558 @@
...
+        public virtual void PreviewMethod()
+        {
+
+        }
+
+        static void Main(string[] args)
+        {
+            var prog = new Program();
+            [|prog.PreviewMethod()|];
+        }
```
Youssef1313, line 57: A few extra cases to consider:

- Attribute on a user-defined operator (`+`) for example, and call-site does `prog1 + prog2`.
- Attribute on a type (class/struct/interface), and call-site does some cast to this class, or a method receives this type as parameter.
- Attribute on a type, and call-site does `new MyType[] { }`
- Attribute on a class derives from `Exception, and call-site does `catch (ThisException e)`
- Attribute on a `Deconstruct` method, and call-site use deconstruction.

Youssef1313 (commented):
For a case like:


```csharp
public class A
{
    [RequiresPreviewFeatures]
    public A() {}
}

public class B
{
    [RequiresPreviewFeatures] // or remove from here and move on class B
    public static A M() => new A();
}
```

There shouldn't be a warning on `new A()` I think? Consider adding a test for this.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Yup, good catch. I missed doing that in the refactoring.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: >eventually will end up with null?

Indeed. That is the behavior I want here. The [RequiresPreviewFeature] attribute can be set as an assembly level attribute, so I need to go up the symbol chain all the way to the root. So a method like `System.Text.SomeMethod()`, will have the chain `SomeMethod -> Text -> System (this is the namespace) -> Current Module -> Current Assembly -> null.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,185 @@
...
+                if (symbol is ITypeSymbol typeSymbol)
+                {
+                    ImmutableArray<INamedTypeSymbol> interfaces = typeSymbol.AllInterfaces;
+                    foreach (INamedTypeSymbol? anInterface in interfaces)
+                    {
+                        if (TryGetOrCachePreviewFeaturesAttributesOnSymbol(operation, anInterface, requiresPreviewFeaturesSymbols, checkParents))
+                        {
+                            return true;
+                        }
+                    }
```
Youssef1313, line 133: Any case that could fail if `AllInterfaces` is replaced with `Interfaces`?

My thought here is that looping the direct interfaces will eventually end up checking the parent interfaces when we recursively call the method with `anInterface`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,185 @@
...
+                IInvocationOperation iOperation => iOperation.TargetMethod,
+                IObjectCreationOperation cOperation => cOperation.Constructor,
+                IPropertyReferenceOperation pOperation => pOperation.Property,
+                IFieldReferenceOperation fOperation => fOperation.Field,
+                IDelegateCreationOperation dOperation => dOperation.Type,
+                IEventReferenceOperation eOperation => eOperation.Member,
+                _ => null,
+            };
+
+        private bool TryGetOrCachePreviewFeaturesAttributesOnSymbol(IOperation operation, ISymbol symbol, ConcurrentDictionary<ISymbol, IOperation?> requiresPreviewFeaturesSymbols, bool checkParents)
```
Youssef1313, line 103: Looks like `checkParents` is always true?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: All good suggestions :) I called out in the PR description that more test cases will be added. The exact cases we want to tag with the analyzer are not defined properly yet. For ex:

``` csharp
// We already have 
public struct Int32: ...
{
}
```
 Now let's say we define a new `public static interface IAddable` type
``` csharp
[RequiresPreviewFeature]
public static interface IAddable
{ }

public struct Int32: IAddable
{

}

Int32 foo = new Int32(5);
```

We don't want to tag `foo` here with a diagnostic, so we're still discussing how to best approach this. So I expect that the analyzer implementation and test cases will evolve as the rules around `RequiresPreviewFeatures` become more defined.

pgovind:
@mavasani @buyaa-n, can you go over the changes I made now and review them? The expectation here is not that the PR is ready to go in. Rather, I want feedback about whether the overall approach is:
1. Correct?
2. Efficient?

Once I get sign off that I can keep this design, I can go about expanding the test matrix and iterating over the analyzer behavior.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: For now. I'm keeping it around because I expect we will decide to stop checking all the way up to the root when we start expanding the test cases.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: In CA1416 we used string comparison with attribute name so that it could be used in environments lower than 5.0. I guess this attribute would newer be used in lower envs as it is about a new feature, so probably better to load the type from the compilation, which also would allow failing fast if the symbol is not found 
```cs
if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeVersioningRequiresPreviewFeaturesAttribute, out var previewFeaturesAttribute))
{
    return;
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -0,0 +1,185 @@
...
+                    OperationKind.FieldReference,
+                    OperationKind.DelegateCreation,
+                    OperationKind.EventReference
+                    );
+            });
+        }
+
+        private void BuildSymbolInformation(OperationAnalysisContext context, ConcurrentDictionary<ISymbol, IOperation?> requiresPreviewFeaturesSymbols)
+        {
+            if (OperationUsesPreviewFeatures(context.Operation, requiresPreviewFeaturesSymbols))
```
buyaa-n, line 74: Seems like we should not warn if the consumer (call-site) has the attributed, example:
```cs
public class Program
{    
    [RequiresPreviewFeatures]    
    public virtual void PreviewMethod()  {        } 
   
    [RequiresPreviewFeatures]  
    void M1()     
   {
        PreviewMethod();  // no warning  
    }
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: I thought the cache will be used to determine if the symbol has the attribute or not, from this comment looks like it is only caching symbols that have the PreviewFeature attribute, which means for a non attributed symbol the cache is useless. I think the dictionary should be `<ISymbol, bool>` type then we could add all symbols into the cache with the value if it has the attribute

buyaa-n:
> @mavasani @buyaa-n, can you go over the changes I made now and review them? The expectation here is not that the PR is ready to go in. Rather, I want feedback about whether the overall approach is:
> 
> 1. Correct?
> 2. Efficient?
> 
> Once I get sign off that I can keep this design, I can go about expanding the test matrix and iterating over the analyzer behavior.

Thanks, @pgovind, overall looks good, left 3 comments to consider.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -12,6 +12,7 @@ CA1842 | Performance | Info | DoNotUseWhenAllOrWaitAllWithSingleArgument, [Docum
 CA1843 | Performance | Info | DoNotUseWhenAllOrWaitAllWithSingleArgument, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1843)
 CA1844 | Performance | Info | ProvideStreamMemoryBasedAsyncOverrides, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1844)
 CA1845 | Performance | Info | UseSpanBasedStringConcat, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1845)
+CA2250 | Usage | Warning | DetectPreviewFeatureAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2250)
```
jeffhandley, line 15: Did we decide on Warning, or were we still considering making this an Error by default?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
jeffhandley: For the time-being, while we prepare to start demoing the feature, would the following message be easy to get in place?

`Using {api} requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1611,4 +1611,13 @@
...
     <value>Replace 'WhenAll' call with argument</value>
   </data>
+  <data name="DetectPreviewFeaturesDescription" xml:space="preserve">
+    <value>An assembly has to opt into preview features before using them.</value>
+  </data>
+  <data name="DetectPreviewFeaturesMessage" xml:space="preserve">
+    <value>Enable preview features for this assembly</value>
+  </data>
+  <data name="DetectPreviewFeaturesTitle" xml:space="preserve">
+    <value>Enable preview features for this assembly</value>
```
jeffhandley, line 1621:
```suggestion
    <value>This API requires opting into preview features</value>
```

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
pgovind: We should make this error by default I think. @terrajobst

pgovind:
Thank you for the reviews everyone. I'll put up a complete PR once I am done with the full implementation

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
mavasani: Analyzers should never report an error by default. Error severity is technically reserved for compiler errors and extremely critical security analyzers.

Tagging @sharwell as well.

jeffhandley: This scenario might justify an exception to that rule since we need users to take explicit action to enable preview features if they want to use them. Not everyone has warnings as errors enabled, and this message could be lost in a sea of other warnings. This logic will be implemented by the compiler for F# too. The fact that this is an analyzer is a factoring/implementation choice--not a key aspect of the user experience, and we're leaving the door open that this could get "promoted" into the compiler once the feature and its details are well proven.

pgovind: > Analyzers should never report an error by default. Error severity is technically reserved for compiler errors and extremely critical security analyzers.

Is this a guideline or a rule? In either case, do you know why analyzers prefer warnings vs errors?

jeffhandley: Here's some background discussion that might also be helpful:
https://github.com/dotnet/csharplang/blob/main/meetings/2021/LDM-2021-05-12.md#discussion

mavasani: @pgovind It’s a guideline. We normally don’t want analyzer diagnostics to prevent building the assembly, which blocks users from doing operations like Run Tests, Start Debugging, etc.

@jeffhandley Thanks for the context. It seems clear that this is a very special analyzer, and analyzer implementation is an interim step towards making it a compiler feature. So, this deserves an exception. I’ll let you all go ahead with warning vs error choice.

Issue #782, opened 2016-01-09
Incorrect report of RS1008 (storing per-compilation data)


sharwell:
I believe the following will reproduce the issue:

```csharp
private static readonly Action<SyntaxTreeAnalysisContext, Compilation, StyleCopSettings> SyntaxTreeAction = HandleSyntaxTree;

private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context, Compilation compilation, StyleCopSettings settings)
{
}
```

The warning is reported in the declaration of `SyntaxTreeAction`, on the token `Compilation`.

Evangelink:
The code currently looks for outer and inner types of all fields. So fields of type `Action` or `Func`  with any inner type being `Compilation` will be reporting an issue. 

We could ignore inner types for those cases. I am wondering if this exclusion should be more restrictive and only be applicable for readonly fields. Because it could be possible to capture some compilation through a local lambda declaration in which case the message would still make sense.

It could be interesting to have 2 messages, one for confident cases and one where we would want the user to review the actual usage.

WDYT @sharwell?

Full test:

```csharp
[Fact]
[WorkItem(782, "https://github.com/dotnet/roslyn-analyzers/issues/782")]
public void CSharp_ActionWithCompilation_NoDiagnostic()
{
	var source = @"
using System;
using System.Collections.Generic;
using System.Collections.Immutable;
using Microsoft.CodeAnalysis;
using Microsoft.CodeAnalysis.CSharp;
using Microsoft.CodeAnalysis.Diagnostics;

class StyleCopSettings { }

[DiagnosticAnalyzer(LanguageNames.CSharp)]
class MyAnalyzer : DiagnosticAnalyzer
{
	private static readonly Action<SyntaxTreeAnalysisContext, Compilation, StyleCopSettings> SyntaxTreeAction = HandleSyntaxTree;

	private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context, Compilation compilation, StyleCopSettings settings)
	{
	}

	public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
	{
		get
		{
			throw new NotImplementedException();
		}
	}

	public override void Initialize(AnalysisContext context)
	{
	}
}";

	VerifyCSharp(source);
}
```

themightyzek:
This warning is also reported on nested types that that contain a variable of type INamedTypeSymbol.
I believe that this code resembles correct usage, but the warning is thrown nonetheless.
[...]

PR #5116, opened 2021-05-19
#45552 use string equals over string compare

NewellClark:
Fix [#45552](https://github.com/dotnet/runtime/issues/45552).

Note: I've gone ahead and added a `GetFirstOrDefaultMemberWithParameterTypes` extension method. I've found the existing `GetFirstOrDefaultMemberWithParameterInfos` method to be clunky while implementing other analyzers. Also, the `ParameterInfo` class that the older method uses only accepts `INamedTypeSymbol`, and not `ITypeSymbol`, which has caused problems on a couple of my other analyzers.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,322 @@
...
+            protected static bool IsNotEqualsOperation(IOperation operation)
+            {
+                return operation is IBinaryOperation binaryOperation &&
+                    binaryOperation.OperatorKind is BinaryOperatorKind.NotEquals;
+            }
+
+            [DoesNotReturn]
+#pragma warning disable CS8763 // A method marked [DoesNotReturn] should not return.
+            protected static void Fail() => Debug.Fail($"'{nameof(GetReplacementExpression)}' must only be called when '{nameof(IsMatch)}' is 'true'.");
+#pragma warning restore CS8763 // A method marked [DoesNotReturn] should not return.
```
Youssef1313, line 184: If the failure is intended for debug only, you can add `[Conditional("DEBUG")]` attribute.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,322 @@
...
+                if (!TryGetInvocationFromComparisonWithLiteralZero(compareResultToLiteralZero, out var invocation))
+                    Fail();
+
+                return invocation;
+            }
+
+            protected static bool IsNotEqualsOperation(IOperation operation)
+            {
+                return operation is IBinaryOperation binaryOperation &&
+                    binaryOperation.OperatorKind is BinaryOperatorKind.NotEquals;
```
Youssef1313, line 178:
```suggestion
                return operation is IBinaryOperation { OperatorKind: BinaryOperatorKind.NotEquals };
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,322 @@
...
+    public sealed class UseStringEqualsOverStringCompareFixer : CodeFixProvider
+    {
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var document = context.Document;
+            var token = context.CancellationToken;
+            var semanticModel = await document.GetSemanticModelAsync(token).ConfigureAwait(false);
+
+            if (!RequiredSymbols.TryGetSymbols(semanticModel.Compilation, out var symbols))
+                return;
```
Youssef1313, line 33: How can this fail? This was already checked in the analyzer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,322 @@
...
+
+                //  Local functions
+
+                static bool TryConvertOperands(IOperation first, IOperation second, [NotNullWhen(true)] out IInvocationOperation? result)
+                {
+                    if (first is IInvocationOperation invocation &&
+                        second is ILiteralOperation literal &&
+                        literal.ConstantValue.HasValue &&
+                        literal.ConstantValue.Value is int integer &&
+                        integer is 0)
```
Youssef1313, line 156: Would that work?

```suggestion
                        literal.ConstantValue.Value is 0)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
```diff
@@ -0,0 +1,128 @@
...
+                var compareStringString = compareMethods.GetFirstOrDefaultMemberWithParameterTypes(stringType, stringType);
+                var compareStringStringBool = compareMethods.GetFirstOrDefaultMemberWithParameterTypes(stringType, stringType, boolType);
+                var compareStringStringStringComparison = compareMethods.GetFirstOrDefaultMemberWithParameterTypes(stringType, stringType, stringComparisonType);
+
+                if (compareStringString is null ||
+                    compareStringStringBool is null ||
+                    compareStringStringStringComparison is null)
+                {
+                    return false;
+                }
```
Youssef1313, line 90: This would be a false negative if one of the overloads exists but not the others.
However, it doesn't worth fixing that IMO since all these overloads are expected to almost always exist.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,322 @@
...
+                {
+                    var codeAction = CodeAction.Create(
+                        Resx.UseStringEqualsOverStringCompareCodeFixTitle,
+                        async token =>
+                        {
+                            var editor = await DocumentEditor.CreateAsync(document, token).ConfigureAwait(false);
+                            var replacementNode = selector.GetReplacementExpression(operation, editor.Generator);
+                            editor.ReplaceNode(operation.Syntax, replacementNode);
+                            return editor.GetChangedDocument();
+                        }, Resx.UseStringEqualsOverStringCompareCodeFixTitle);
```
Youssef1313, line 54:
```suggestion
                        }, nameof(Resx.UseStringEqualsOverStringCompareCodeFixTitle));
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,322 @@
...
+            }
+
+            protected RequiredSymbols Symbols { get; }
+
+            /// <summary>
+            /// Indicates whether the specified <see cref="IOperation"/> matches the current <see cref="OperationSelector"/>
+            /// </summary>
+            /// <param name="compareResultToLiteralZero"></param>
+            /// <returns></returns>
+            public abstract bool IsMatch(IOperation compareResultToLiteralZero);
```
Youssef1313, line 115: Is it possible to change the parameter type to `IBinaryOperation` (and delegate the parameter type change to other methods as well?).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
NewellClark: @Youssef1313 Good catch. I'm having to basically rewrite the analyzer due to architecture flaws (I didn't realize analyzers couldn't reference anything in the code fix), but I will make sure to rewrite it so it is tolerant of some overloads being absent.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: I will do that when I restructure the analyzer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: Do we know for a fact that it is impossible for this to fail in the fixer after it succeeded in the analyzer? I figured better safe than sorry.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: Yup, that works.

NewellClark:
I ended up having to completely rewrite the analyzer because I wasn't aware that analyzers are not supposed to have any reference whatsoever to certain APIs, and originally both the analyzer and fixer had access to a type that used `SyntaxGenerator` which kept triggeringRS1022. I rewrote the analyzer so it no longer has any dependencies on the fixer.

mavasani:
Tagging @ryzngard for review.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
ryzngard: Agree this should be an assertion. If the symbols somehow disappear between the analyzer and the fixer something has gone horribly wrong.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,218 @@
...
+            var document = context.Document;
+            var token = context.CancellationToken;
+            var semanticModel = await document.GetSemanticModelAsync(token).ConfigureAwait(false);
+
+            if (!RequiredSymbols.TryGetSymbols(semanticModel.Compilation, out var symbols))
+                return;
+
+            var root = await document.GetSyntaxRootAsync(token).ConfigureAwait(false);
+
+            if (semanticModel.GetOperation(root.FindNode(context.Span, getInnermostNodeForTie: true), token) is not IBinaryOperation violation)
```
ryzngard, line 36: nit: consider multiple lines for clarity

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,218 @@
...
+                return;
+
+            //  Get the replacer that applies to the reported violation. Bail out if no replacer 
+            //  matches the reported violation.
+            var replacer = GetOperationReplacers(symbols).FirstOrDefault(x => x.IsMatch(violation));
```
ryzngard, line 41: Assert or throw here. The replacers are supposed to be strongly tied with the analyzer right?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,218 @@
...
+            if (replacer is null)
+                return;
+
+            var codeAction = CodeAction.Create(
+                Resx.UseStringEqualsOverStringCompareCodeFixTitle,
+                CreateChangedDocument,
+                nameof(Resx.UseStringEqualsOverStringCompareCodeFixTitle));
+            context.RegisterCodeFix(codeAction, context.Diagnostics);
+            return;
```
ryzngard, line 50: return is not needed here

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,218 @@
...
+            {
+                var editor = await DocumentEditor.CreateAsync(document, token).ConfigureAwait(false);
+                var replacementNode = replacer.CreateReplacementExpression(violation, editor.Generator);
+                editor.ReplaceNode(violation.Syntax, replacementNode);
+
+                return editor.GetChangedDocument();
+            }
+        }
+
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(UseStringEqualsOverStringCompare.RuleId);
```
ryzngard, line 64: nit: Public properties usually go before public methods. It also makes it easier to tell what diagnostics this fixes (in case you couldn't figure it out from the name)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
Youssef1313: @ryzngard It's a convention used in the repository (and dotnet/roslyn as well) to put an explicit return just before local functions

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            public override SyntaxNode CreateReplacementExpression(IBinaryOperation violation, SyntaxGenerator generator)
+            {
+                RoslynDebug.Assert(IsMatch(violation));
+
+                var compareInvocation = GetInvocation(violation);
+
+                //  We know that the 'ignoreCase' argument in 'string.Compare(string, string, bool)' is a boolean literal
+                //  because we've asserted that 'IsMatch' returns true.
+                var ignoreCaseLiteral = (ILiteralOperation)compareInvocation.Arguments.GetArgumentForParameterAtIndex(2).Value;
```
ryzngard, line 171: Is this also true for cases where a boolean variable is used? I haven't checked `UseStringEqualsOverStringCompare.IsStringStringBoolCase` to know if it handles that as well

```c# 
bool CompareNames(Person p, bool ignoreCase) => string.Compare(Name, p.Name, ignoreCase)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                //  If the violation contains a call to 'string.Compare(x, y, true)' then we
+                //  replace it with a call to 'string.Equals(x, y, StringComparison.CurrentCultureIgnoreCase)'.
+                //  If the violation contains a call to 'string.Compare(x, y, false)' then we
+                //  replace it with a call to 'string.Equals(x, y, StringComparison.CurrentCulture)'. 
+                var stringComparisonEnumMemberName = (bool)ignoreCaseLiteral.ConstantValue.Value ?
```
ryzngard, line 177: This should likely be `Ordinal` and `OrdinalIgnoreCase` based on [guidelines](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
```diff
@@ -0,0 +1,248 @@
...
+        /// <returns></returns>
+        internal static bool IsStringStringCase(IBinaryOperation binaryOperation, RequiredSymbols symbols)
+        {
+            var invocation = GetInvocationFromEqualityCheckWithLiteralZero(binaryOperation);
+
+            return invocation is not null &&
+                //  Don't report a diagnostic if either the string.Compare overload or the
+                //  corrasponding string.Equals overload is missing.
+                symbols.CompareStringString is not null &&
+                symbols.EqualsStringString is not null &&
```
ryzngard, line 185: Check these before calling `GetInvocationFromEqualityCheckWithLiteralZero` for clarity and reducing calls when not needed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
```diff
@@ -0,0 +1,248 @@
...
+            context.EnableConcurrentExecution();
+            context.RegisterCompilationStartAction(OnCompilationStart);
+        }
+
+        private static void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            if (!RequiredSymbols.TryGetSymbols(context.Compilation, out var symbols))
+                return;
+            context.RegisterOperationAction(AnalyzeOperation, OperationKind.Binary);
+            return;
```
ryzngard, line 59:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
```diff
@@ -0,0 +1,248 @@
...
+                    }
+                }
+            }
+        }
+
+        internal sealed class RequiredSymbols
+        {
+            //  Named-constructor 'TryGetSymbols' inits all properties or doesn't construct an instance.
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+            private RequiredSymbols() { }
```
ryzngard, line 81: This is a clarity/style concern for me. It's likely easier to just have a constructor with paramters needed than to rely on all properties being init only. Or convert this to a record?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
```diff
@@ -0,0 +1,248 @@
...
+                symbols = new RequiredSymbols
+                {
+                    StringType = stringType,
+                    BoolType = boolType,
+                    StringComparisonType = stringComparisonType,
+                    CompareStringString = compareStringString,
+                    CompareStringStringBool = compareStringStringBool,
+                    CompareStringStringStringComparison = compareStringStringStringComparison,
+                    EqualsStringString = equalsStringString,
+                    EqualsStringStringStringComparison = equalsStringStringStringComparison
```
ryzngard, line 119: We should check symbol requirements here for pairings. `CompareStringString` requires `EqualsStringString` for example. Return false if no valid configuration is available

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
```diff
@@ -0,0 +1,248 @@
...
+            return invocation is not null &&
+                //  Don't report a diagnostic if either the string.Compare overload or the
+                //  corrasponding string.Equals overload is missing.
+                symbols.CompareStringStringStringComparison is not null &&
+                symbols.EqualsStringStringStringComparison is not null &&
+                invocation.TargetMethod.Equals(symbols.CompareStringStringStringComparison, SymbolEqualityComparer.Default);
+        }
+
+        //  No IOperation instances are being stored here.
+#pragma warning disable RS1008 // Avoid storing per-compilation data into the fields of a diagnostic analyzer
```
ryzngard, line 240: Is there a bug for this false positive on RS1008?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompareTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        Return {fixedExpression}
+    End Function
+End Class";
+
+            return VerifyVB.VerifyCodeFixAsync(testCode, VerifyVB.Diagnostic(Rule).WithLocation(0), fixedCode);
+        }
+
+        [Theory]
+        [MemberData(nameof(CS_StringCompareExpressionsTestData))]
+        public Task StringCompareResult_CompareToNonLiteralZero_NoDiagnostic_CS(string expression)
```
ryzngard, line 155: For const values we could also provide a diagnostic. I'd consider this an area for future improvement and not required for this PR

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
ryzngard: o.o who decided that? Because I definitely haven't been

ryzngard: Apparently this is @CyrusNajmabadi preferrence... I won't block a PR on style preference that isn't documented. You can keep if you like, and I'll just grumble in my corner alone :P

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
ryzngard: Feel free to ignore per https://github.com/dotnet/roslyn-analyzers/pull/5116#discussion_r637182218

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
Youssef1313: @ryzngard string.Compare is using CurrentCulture by default.

https://source.dot.net/#System.Private.CoreLib/String.Comparison.cs,200

So probably the codefix would want to keep the *same* semantics.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
ryzngard: I've added https://github.com/dotnet/roslyn-analyzers/issues/5122 for further discussion

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
ryzngard: I don't fully agree. In general we don't want to change explicit semantics for code, but there's no way to know if the user intended to use `CurrentCulture` or not. When it's ambiguous, I think suggesting the public documented guidelines would be a good thing.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: The analyzer only flags cases where the boolean argument is a literal.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompareTests.cs
NewellClark: If I write an expression with a constant in it, I expect that changing the value initially assigned to the constant will change the value of the expression. The fixer removes the constant from the expression entirely, which could break if the constant is given a different value.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: I'll replace `FirstOrDefault` with `First`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: @mavasani has asked me to have a `return` before local the local function definitions for clarity.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: The examples in [the issue](https://github.com/dotnet/runtime/issues/45552) specified that we use `CurrentCulture` and `CurrentCultureIgnoreCase` so we don't change semantics.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
NewellClark: Same as above. @mavasani has asked that I return before local function declarations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
NewellClark: Agreed. I'll change it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
NewellClark: Good idea. I'll add that.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
ryzngard: Yea, I followed up offline with him and a few others. I was unaware that was a practice we started. Hopefully we can get an analyzer to help the underlying issue, but for now this is correct.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
NewellClark: Fixed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.cs
NewellClark: [Bug report](https://github.com/dotnet/roslyn-analyzers/issues/782)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
ryzngard: I don't see any explicit discussion in the issue about whether the replacement for `string.Compare(s1, s2)` should use `CurrentCulture` or `Ordinal`. That said, I'm not strong enough of the opinion that we should change the functional meaning for potential correction. Especially if others think it should be exact parity.  

I will expand on my thought process for discussion, but consider this non-blocking: 

https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings specifies
> "Use overloads that explicitly specify the string comparison rules for string operations. Typically, this involves calling a method overload that has a parameter of type StringComparison." 

If I were a user trying to update code to follow that guideline, how would I consider doing it? I think most cases I would add a `Ordinal` or `OrdinalIgnoreCase` as the intent for string comparison unless context dictates I should be using culture aware comparison. That also follows the guidance of 
> "Use StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase for comparisons as your safe default for culture-agnostic string matching." 

and 

> "Use comparisons with StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase for better performance."

This fix is listed as `Usage`, but the key findings in the issue are centered around performance. Probably because nobody disagreed that `string.Equals(...)` is more readable than `string.Compare(...) == 0`. 

If we are encouraging users to have mostly correct and performant code, or even for what I would consider to give as guidance as "probably correct" in most cases, we would use `Ordinal` or `OrdinalIgnoreCase`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompareTests.cs
ryzngard: While true, it's an indicator of what I would consider hard to read code and potentially bad design. I don't know of a strong use case where we would want to encourage that, and analyzers are opinionated by design.

ryzngard (approved):
Looks good to me other than discussion points, but nothing blocking. Thanks for adding this!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseStringEqualsOverStringCompare.Fixer.cs
NewellClark: In my opinion, a code fix should only change the issue that its analyzer is reporting on. This analyzer is not flagging incorrect uses of `CurrentCulture`, it is flagging uses of a total-order comparison when a simple equality check was desired. I would not be opposed to an analyzer that flags usages of any `Compare` or `Equals` overload that doesn't specify the kind of comparison, but I don't think this analyzer should be changing semantics.

PR #5117, opened 2021-05-19
[CA1416] Targeted build attributes can be overriden in cross platform assemblies

### LINKED ISSUE #5023: [CA1416] Allow child APIs attribute override assembly attribute in targeted builds of cross-platform assemblies

buyaa-n:
Fixes #5023

### Allow overriding assembly-level attribute in targeted builds of cross-platform assemblies
By the design, child APIs should not extend parent support. `SupportedOSPlatform` attributes illustrate that the API/assembly is only reachable for that platform and it should not have any API supported on other platforms inside it. We have planned to warn if it happens by dotnet/runtime#43971, but this scenario is happening in runtime for targeted cross-platform builds. Ideally, all platform-specific implementation should be separated from cross-platform src. Though that causing code duplication at some level, so most developers would not like to apply strict separation of code.

Even though it is an expected scenario by design it is unfortunate to not get any warning about this inconsistency or possible reference of incompatible APIs within the same assembly

Initially, we decided to add a special case `[SupportedOSPlatform("CrossPlatofrm")]` attribute for targeted builds of cross-platform assembly to allow child member attributes to override the parent. For distinguishing real platform-specific assemblies from cross-platform assembly with targeted builds. But using `SupportedOSPlatform` attribute for handling this is causing side effects in further analysis and needs more tweak to get rid of it.

So i decided to use MSBuild property `<CompilerVisibleProperty Include="CrossPlatform" />` instead, which we will add into targeted builds of cross platform assembly.

(We might change "CrossPlatform" into "PlatformNeutral" or something like that)

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -1025,7 +1025,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 return;
             }
 
-            CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, symbol, checkParents: true);
+            CheckOperationAttributes(symbol, checkParents: true);
 
             if (symbol is IPropertySymbol property)
             {
@@ -1033,7 +1033,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 {
                     if (accessor != null)
                     {
-                        CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor, checkParents: false);
+                        CheckOperationAttributes(accessor, checkParents: false);
                     }
                 }
             }
@@ -1043,32 +1043,26 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
 
                 if (accessor != null)
                 {
-                    CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, accessor, checkParents: false);
+                    CheckOperationAttributes(accessor, checkParents: false);
                 }
             }
             else if (symbol is IMethodSymbol method && method.IsGenericMethod)
             {
-                CheckTypeArguments(method.TypeArguments, operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms);
+                CheckTypeArguments(method.TypeArguments);
             }
 
             if (symbol.ContainingSymbol is INamedTypeSymbol namedType && namedType.IsGenericType)
             {
-                CheckTypeArguments(namedType.TypeArguments, operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms);
+                CheckTypeArguments(namedType.TypeArguments);
             }
 
-            void CheckTypeArguments(ImmutableArray<ITypeSymbol> typeArguments, IOperation operation, OperationAnalysisContext context,
-                PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>, (SmallDictionary<string, Versions> attributes,
-                SmallDictionary<string, Versions>? csAttributes)> platformSpecificOperations,
-                ConcurrentDictionary<ISymbol, PlatformAttributes> platformSpecificMembers, ImmutableArray<string> msBuildPlatforms)
+            void CheckTypeArguments(ImmutableArray<ITypeSymbol> typeArguments)
             {
                 using var workingSet = PooledHashSet<ITypeSymbol>.GetInstance();
-                CheckTypeArgumentsCore(typeArguments, operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, workingSet);
+                CheckTypeArgumentsCore(typeArguments, workingSet);
             }
 
-            void CheckTypeArgumentsCore(ImmutableArray<ITypeSymbol> typeArguments, IOperation operation, OperationAnalysisContext context,
-                PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>, (SmallDictionary<string, Versions> attributes,
-                SmallDictionary<string, Versions>? csAttributes)> platformSpecificOperations, ConcurrentDictionary<ISymbol,
-                PlatformAttributes> platformSpecificMembers, ImmutableArray<string> msBuildPlatforms, PooledHashSet<ITypeSymbol> workingSet)
+            void CheckTypeArgumentsCore(ImmutableArray<ITypeSymbol> typeArguments, PooledHashSet<ITypeSymbol> workingSet)
             {
                 foreach (var typeArgument in typeArguments)
                 {
@@ -1077,20 +1071,18 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         workingSet.Add(typeArgument);
                         if (typeArgument.SpecialType == SpecialType.None)
                         {
-                            CheckOperationAttributes(operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, typeArgument, checkParents: true);
+                            CheckOperationAttributes(typeArgument, checkParents: true);
 
                             if (typeArgument is INamedTypeSymbol nType && nType.IsGenericType)
                             {
-                                CheckTypeArgumentsCore(nType.TypeArguments, operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, workingSet);
+                                CheckTypeArgumentsCore(nType.TypeArguments, workingSet);
                             }
                         }
                     }
                 }
             }
 
-            void CheckOperationAttributes(IOperation operation, OperationAnalysisContext context, PooledConcurrentDictionary<KeyValuePair<IOperation, ISymbol>,
-                (SmallDictionary<string, Versions> attributes, SmallDictionary<string, Versions>? csAttributes)> platformSpecificOperations,
-                ConcurrentDictionary<ISymbol, PlatformAttributes> platformSpecificMembers, ImmutableArray<string> msBuildPlatforms, ISymbol symbol, bool checkParents)
+            void CheckOperationAttributes(ISymbol symbol, bool checkParents)
             {
                 if (TryGetOrCreatePlatformAttributes(symbol, checkParents, crossPlatform, platformSpecificMembers, out var operationAttributes))
                 {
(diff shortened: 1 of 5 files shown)


jeffhandley (approved):
Looks good to me. And I like this approach much better than the magic string platform name!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -199,6 +200,11 @@ static ImmutableArray<IMethodSymbol> GetOperatingSystemGuardMethods(IMethodSymbo
 
             static bool NameAndParametersValid(IMethodSymbol method) => method.Name.StartsWith(IsPrefix, StringComparison.Ordinal) &&
                     (method.Parameters.Length == 0 || method.Name.EndsWith(OptionalSuffix, StringComparison.Ordinal));
+
+            static bool HasCrossPlatformProperty(AnalyzerOptions options, Compilation compilation, CancellationToken cancellationToken)
+            {
+                return null != options.GetMSBuildPropertyValue(MSBuildPropertyOptionNames.PlatformNeutralAssembly, compilation, cancellationToken);
```
mavasani, line 206: We normally have the constant (null) on the right hand side of the comparison operator. I don't have strong feelings though, so this is fine.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -1058,7 +1065,7 @@ private static ISymbol GetEventAccessor(IEventSymbol iEvent, IOperation operatio
                 CheckTypeArgumentsCore(typeArguments, operation, context, platformSpecificOperations, platformSpecificMembers, msBuildPlatforms, workingSet);
             }
 
-            static void CheckTypeArgumentsCore(ImmutableArray<ITypeSymbol> typeArguments, IOperation operation, OperationAnalysisContext context,
```
mavasani, line 1061: We should either retain this as static and pass the new bool parameter to the required methods OR remove all the extra parameters that are shared between these local functions and containing method (`AnalyzeOperation`)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -128,6 +128,86 @@ class Target
...
+    [UnsupportedOSPlatform(""linux"")]
+    public string UnsupportedLinuxProgram => program;
+
+    void CrossPlatformCallSite()
+    {   
+        // should warn as WindowsOnlyProgram is windows only
+        var a = {|#0:WindowsOnlyProgram|};  // This call site is reachable on: 'linux'. 'Test.WindowsOnlyProgram' is only supported on: 'windows'.
+        a = {|#1:UnsupportedLinuxProgram|}; // This call site is reachable on: 'linux'. 'Test.UnsupportedLinuxProgram' is unsupported on: 'linux'.
+        a = WindowsIosLinuxOnlyProgram;
+        a = [|AndroidBrowserOnlyProgram|]; // This call site is reachable on: 'linux'. 'Test.AndroidBrowserOnlyProgram' is only supported on: 'android', 'browser'.
```
mavasani, line 164: Any reason why this diagnostic is not annotated with `#2`? I would expect either all diagnostics to have no `#` prefix or all of them to have it for consistency.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
```diff
@@ -128,6 +128,86 @@ class Target
...
+            var source = @"
+using System.Collections.Generic;
+using System.Runtime.Versioning;
+
+[assembly:SupportedOSPlatform(""linux"")] 
+public class Test
+{
+    private string program;
+
+    [SupportedOSPlatform(""windows"")] // Can overwrite parent browser support as cross platform
```
mavasani, line 143: > Can overwrite parent browser support as cross platform

Is the parent browser or linux here? Was that a typo?

mavasani (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
buyaa-n: It is because the diagnostic parameter has a list of platforms: "'android', 'browser'" and the order of the platforms are not consistent and cause a test failure

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
buyaa-n: yeah that is typo, initially i used browser, later changed to linux to repro a bug found in runtime test

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: Good point, removing all extra parameters from local functions

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs
mavasani: > order of the platforms are not consistent

We always try to keep the analyzer diagnostic messages deterministic between builds. Can you order the platform names in the message?

buyaa-n: > Can you order the platform names in the message

I guess I need to update the [Enumerator of the SmallDictionary](https://github.com/dotnet/roslyn-analyzers/blob/5bd4ce968d457bcc42edf8838af79b937d69cbeb/src/Utilities/Compiler/SmallDictionary.cs#L739) for that, to loop through left branches first, but do not see an easy solution for that.

buyaa-n: I gave it try, updated the enumerator to always return elements in order so that the smallest elements come first, but the key hashcode populated by `StringComparer.GetHashCode(string)` is not consistent for each run, therefore the order of keys also not consistent. So I cannot keep the platform names in order efficiently, I could sort the list before diagnostics but not sure that overhead is worth it.

mavasani: You shouldn’t order this in the dictionary, just call the Order linq routine for the argument passed to the diagnostic create. We do this all the time, it should be fine to call Order on arguments passed to diagnostic.

buyaa-n: > You shouldn’t order this in the dictionary, just call the Order linq routine for the argument passed to the diagnostic create. 

Yes, as i mentioned in my previous comment i can call the `List.Sort()` method on the platforms list, my point is i do not really like to add such addition unless necessary, even it has a very minor impact.

>  We do this all the time, it should be fine to call Order on arguments passed to diagnostic.

After adding the sort but it is causing some tests fail where we are appending `and all other platforms` to build more meaningful message, of course i can tweak that somehow to fix it, but still i don't think it is worth. 

```log
Expected: This call site is reachable on: 'windows' 9.0 and before, and all other platforms. 'AllowList.Windows10Only()' is only supported on: 'windows' 10.0 and later.
Actual:   This call site is reachable on: and all other platforms, 'windows' 9.0 and before. 'AllowList.Windows10Only()' is only supported on: 'windows' 10.0 and later.
```

buyaa-n:
@mavasani @jmarolf I could not understand why the Spanish tests are failing, seems it is not sorting the platform names at all but do not know how could that happen:

```log
Expected: Se puede acceder a este sitio de llamada en todas las plataformas. "Test.SupportedOnWindows10LinuxMacOS14()" solo se admite en 'linux', "macOS/OSX" 14.0 y versiones posteriores, "windows" 10.0 y versiones posteriores.
Actual:   Se puede acceder a este sitio de llamada en todas las plataformas. "Test.SupportedOnWindows10LinuxMacOS14()" solo se admite en "macOS/OSX" 14.0 y versiones posteriores, "windows" 10.0 y versiones posteriores, 'linux'.
``

PR #5124, opened 2021-05-23
Tiny Refactoring CA1824

### LINKED ISSUE #5123: GetOrCreateTypeByMetadataName is called on every attribute syntax callback
https://github.com/dotnet/roslyn-analyzers/blob/5968beca1c5c6c12d4fa81a08b7b4561a3c05d6e/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Resources/CSharpMarkAssembliesWithNeutralResourcesLanguage.cs#L20-L34

https://github.com/dotnet/roslyn-analyzers/blob/5968beca1c5c6c12d4fa81a08b7b4561a3c05d6e/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Resources/MarkAssembliesWithNeutralResourcesLanguage.cs#L92-L99

Should see if this impacts performance and move the call to `GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCodeDomCompilerGeneratedCodeAttribute);` to the compilation start action

Youssef1313:
Closes #5123

**First commit**:

Main idea is to move the line:

```csharp
            INamedTypeSymbol? generatedCode = model.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCodeDomCompilerGeneratedCodeAttribute);
```

from `CheckResxGeneratedFile` to `RegisterCompilationStartAction` and bail-out early if needed.

Extra cleanups:

- Avoid unneeded type checking and cast the node directly to AttributeSyntax. The cast can't fail because the registered node action is for `SyntaxKind.Attribute`.
- Rename `nc` to `context`

---

**Second commit**:

Move `TryCheckNeutralResourcesLanguageAttribute` to compilation start to avoid unnecessary callbacks when we're sure there is no diagnostics to be produced.

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Resources/CSharpMarkAssembliesWithNeutralResourcesLanguage.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Resources/CSharpMarkAssembliesWithNeutralResourcesLanguage.cs
@@ -15,16 +15,17 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Resources
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     public sealed class CSharpMarkAssembliesWithNeutralResourcesLanguageAnalyzer : MarkAssembliesWithNeutralResourcesLanguageAnalyzer
     {
-        protected override void RegisterAttributeAnalyzer(CompilationStartAnalysisContext context, Action onResourceFound)
+        protected override void RegisterAttributeAnalyzer(CompilationStartAnalysisContext context, Action onResourceFound, INamedTypeSymbol generatedCode)
         {
-            context.RegisterSyntaxNodeAction(nc =>
+            context.RegisterSyntaxNodeAction(context =>
             {
-                if (!CheckAttribute(nc.Node))
+                var attributeSyntax = (AttributeSyntax)context.Node;
+                if (!CheckAttribute(attributeSyntax))
                 {
                     return;
                 }
 
-                if (!CheckResxGeneratedFile(nc.SemanticModel, nc.Node, ((AttributeSyntax)nc.Node).ArgumentList.Arguments[0].Expression, nc.CancellationToken))
+                if (!CheckResxGeneratedFile(context.SemanticModel, attributeSyntax, attributeSyntax.ArgumentList.Arguments[0].Expression, generatedCode, context.CancellationToken))
                 {
                     return;
                 }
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Resources/CSharpMarkAssembliesWithNeutralResourcesLanguage.cs
```diff
@@ -15,16 +15,17 @@ namespace Microsoft.NetCore.CSharp.Analyzers.Resources
...
     public sealed class CSharpMarkAssembliesWithNeutralResourcesLanguageAnalyzer : MarkAssembliesWithNeutralResourcesLanguageAnalyzer
     {
-        protected override void RegisterAttributeAnalyzer(CompilationStartAnalysisContext context, Action onResourceFound)
+        protected override void RegisterAttributeAnalyzer(CompilationStartAnalysisContext context, Action onResourceFound, INamedTypeSymbol generatedCode)
         {
-            context.RegisterSyntaxNodeAction(nc =>
+            context.RegisterSyntaxNodeAction(context =>
             {
-                if (!CheckAttribute(nc.Node))
+                var attributeSyntax = (AttributeSyntax)context.Node;
```
Evangelink, line 22: Previous code was having the possibility that the `Node` isn't a `AttributeSyntax`, it also doesn't look possible but I am wondering if that was done for a specific reason or just by "node is generic so let's work in the generic case".

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Resources/MarkAssembliesWithNeutralResourcesLanguage.cs
```diff
@@ -53,9 +53,26 @@ public override void Initialize(AnalysisContext context)
 
             context.RegisterCompilationStartAction(cc =>
             {
-                var hasResource = false;
+                INamedTypeSymbol? generatedCode = cc.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCodeDomCompilerGeneratedCodeAttribute);
```
Evangelink, line 56: [cosmetic][minor] You could use the `TryGetXXX` instead of doing get then null check.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Resources/MarkAssembliesWithNeutralResourcesLanguage.cs
```diff
@@ -53,9 +53,26 @@ public override void Initialize(AnalysisContext context)
 
             context.RegisterCompilationStartAction(cc =>
```
Evangelink, line 54: As you have done some extra refactorings, it might be worth also renaming `cc` into `context`.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Resources/CSharpMarkAssembliesWithNeutralResourcesLanguage.cs
Youssef1313: @Evangelink It should always be `AttributeSyntax` since we do `RegisterSyntaxNodeAction(...., SyntaxKind.Attribute)`

Youssef1313:
Thanks for review @Evangelink. I addressed your feedback.

mavasani (approved):

PR #5129, opened 2021-05-24
Implement analyzer and code-fixers for DynamicInterfaceCastableImplementation analyzers

jkoritzinsky:
Implement analyzers and code fixes as per https://github.com/dotnet/runtime/issues/41529

Fixes https://github.com/dotnet/runtime/issues/41529


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Collections.Immutable;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    internal class DynamicInterfaceCastableImplementationAnalyzer : DiagnosticAnalyzer
```
Youssef1313, line 12:
```suggestion
    internal sealed class DynamicInterfaceCastableImplementationAnalyzer : DiagnosticAnalyzer
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,133 @@
...
+            MethodsDeclaredOnImplementationTypeMustBeSealed);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+            context.RegisterSymbolAction(context => AnalyzeType(context), SymbolKind.NamedType);
+        }
+
+        private const string DynamicInterfaceCastableImplementationAttributeTypeName = "System.Runtime.InteropServices.DynamicInterfaceCastableImplementationAttribute";
```
Youssef1313, line 65: Could be moved to `WellKnownTypeNames`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,133 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+            DynamicInterfaceCastableImplementationUnsupported,
+            InterfaceMethodsMissingImplementation,
+            MethodsDeclaredOnImplementationTypeMustBeSealed);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+            context.RegisterSymbolAction(context => AnalyzeType(context), SymbolKind.NamedType);
```
Youssef1313, line 62: Should probably work?

```suggestion
            context.RegisterSymbolAction(AnalyzeType, SymbolKind.NamedType);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            if (targetType.TypeKind != TypeKind.Interface)
+            {
+                return;
+            }
+
+            bool isDynamicInterfaceImplementation = false;
+            foreach (var attribute in targetType.GetAttributes())
+            {
+                if (attribute.AttributeClass.ToDisplayString(SymbolDisplayFormats.QualifiedTypeAndNamespaceSymbolDisplayFormat) == DynamicInterfaceCastableImplementationAttributeTypeName)
```
Youssef1313, line 79: I think it's better to retrieve the attribute type in compilation start action, bail-out early if it doesn't exist, and perform symbol equality here (using the HasAttribute extension):

https://github.com/dotnet/roslyn-analyzers/blob/0f20ec1f3877f114800347a873c86d21f377f01b/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L595

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementationTests.cs
```diff
@@ -0,0 +1,1076 @@
...
+        }
+
+        private static async Task VerifyCSAnalyzerAsync(string source)
+        {
+            await new VerifyCS.Test
+            {
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                TestCode = source
+            }.RunAsync();
```
Youssef1313, line 1053:
```suggestion
            await VerifyCSCodeFixAsync(source, source);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            else if (diagnostic.Id == DynamicInterfaceCastableImplementationAnalyzer.MethodsDeclaredOnImplementationTypeMustBeSealedRuleId)
+            {
+                context.RegisterCodeFix(
+                    new MyCodeAction(
+                        MicrosoftNetCoreAnalyzersResources.SealMethodDeclaredOnImplementationType,
+                        async ct => await SealMemberDeclaredOnImplementationType(declaration, context.Document, context.CancellationToken).ConfigureAwait(false),
+                        equivalenceKey: nameof(MicrosoftNetCoreAnalyzersResources.SealMethodDeclaredOnImplementationType)),
+                    diagnostic);
+            }
```
Youssef1313, line 57: else throw NotSupportedException or something that ensures the code path isn't reachable?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,203 @@
...
+                        }
+                }
+
+                return declaration;
+            }
+        }
+
+        protected override bool CodeFixSupportsDeclaration(SyntaxNode declaration)
+        {
+            return !declaration.IsKind(SyntaxKind.VariableDeclarator);
```
Youssef1313, line 200: Any test shows this returning false (i.e, a variable declarator)?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,203 @@
...
+            SyntaxNode SealMemberDeclaration(SyntaxNode declaration, ISymbol symbol)
+            {
+                switch (symbol.Kind)
+                {
+                    case SymbolKind.Method:
+                        {
+                            DeclarationModifiers modifiers = generator.GetModifiers(declaration)
+                               .WithIsAbstract(false)
+                               .WithIsVirtual(false)
+                               .WithIsSealed(true);
```
Youssef1313, line 153: It looks like all the 3 switch cases are using this. Could it be moved outside the switch?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementationTests.cs
```diff
@@ -0,0 +1,1076 @@
...
+        [Fact]
+        public async Task DynamicInterfaceCastableImplementation_NoPropertiesImplementated_CS_Diagnostic()
+        {
+            string source = @"
+using System;
+using System.Runtime.InteropServices;
+
+interface I
+{
+    int Property { get; set; }
```
Youssef1313, line 247: A test with `init`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1662,4 +1662,37 @@
...
+    <value>Providing a 'DynamicInterfaceCastableImplementation' interface in Visual Basic is unsupported</value>
+  </data>
+  <data name="DynamicInterfaceCastableImplementationUnsupportedTitle" xml:space="preserve">
+    <value>Providing a 'DynamicInterfaceCastableImplementation' interface in Visual Basic is unsupported</value>
+  </data>
+  <data name="ImplementInterfacesOnDynamicCastableImplementation" xml:space="preserve">
+    <value>Implement inherited interfaces</value>
+  </data>
+  <data name="InterfaceMethodsMissingImplementationDescription" xml:space="preserve">
+    <value>Types attributed with 'DynamicInterfaceCastableImplementationAttribute' act as an interface implementation for a type that implements the 'IDynamicInterfaceCastable' type. As a result, it must provide an implementation of all of the methods defined in the inherited interfaces, because the type that implements 'IDynamicInterfaceCastable' will not provide them otherwise.</value>
```
Youssef1313, line 1678: Is `IDynamicInterfaceCastable` is a must to use `DynamicInterfaceCastableImplementationAttribute`? Any separate analyzer work for that?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,203 @@
...
+            CancellationToken ct)
+        {
+            var editor = await DocumentEditor.CreateAsync(document, ct).ConfigureAwait(false);
+
+            INamedTypeSymbol type = (INamedTypeSymbol)editor.SemanticModel.GetDeclaredSymbol(declaration, ct);
+            var generator = editor.Generator;
+
+            var defaultMethodBodyStatements = generator.DefaultMethodBody(editor.SemanticModel.Compilation).ToArray();
+            List<SyntaxNode> generatedMembers = new List<SyntaxNode>();
+            foreach (var iface in type.AllInterfaces)
```
Youssef1313, line 33: Are there any tests that will fail if `AllInterfaces` is replaced with `Interfaces`? If not, considering adding one.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,133 @@
...
+                new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.InterfaceMethodsMissingImplementationMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources)),
+                DiagnosticCategory.Usage,
+                RuleLevel.BuildWarning,
+                new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.InterfaceMethodsMissingImplementationDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources)),
+                isPortedFxCopRule: false,
+                isDataflowRule: false);
+
+        internal const string MethodsDeclaredOnImplementationTypeMustBeSealedRuleId = "CA2254";
+
+        private static readonly DiagnosticDescriptor MethodsDeclaredOnImplementationTypeMustBeSealed =
```
Youssef1313, line 42: It seems it's not only methods, so probably say "Members"?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
jkoritzinsky: It will work, but it ends up actually being slightly worse performance (the delegate instance isn't cached), so I prefer doing the explicit lambda.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDynamicInterfaceCastableImplementation.Fixer.cs
jkoritzinsky: This is tested by `DynamicInterfaceCastableImplementation_Events_MultipleInOneDeclaration_NoBody_CS_Diagnostic`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.Fixer.cs
jkoritzinsky: I think silently falling through here is ok unless the pattern in the repo otherwise is to throw.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
jkoritzinsky: `IDynamicInterfaceCastable` is a must-use type related to the `DynamicInterfaceCastableImplementationAttribute` type. But since `IDynamicInterfaceCastable` will be implemented by a different type, so there's no analyzer work for that.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDynamicInterfaceCastableImplementation.Fixer.cs
jkoritzinsky: Yes, the `DynamicInterfaceCastableImplementation_GrandparentInterfaceMethodsNotImplementated_CS_Diagnostic` test would fail if we used `Interfaces` instead of `AllInterfaces`.

jkoritzinsky:
@dotnet/roslyn-analysis just want to make sure that reviewing this is on the backlog as it's been green in CI and waiting for review for 27 days.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            if (missingMethodImplementations)
+            {
+                context.ReportDiagnostic(targetType.CreateDiagnostic(InterfaceMembersMissingImplementation, targetType.ToDisplayString()));
+            }
+
+            foreach (var member in targetType.GetMembers())
+            {
+                if (member.IsVirtual || member.IsAbstract)
```
ryzngard, line 115: I don't see tests for virtual or abstract members. Did I miss something? (GH search on large files can miss sometimes)

ryzngard (commented):
Some style problems. In general I think this repo uses `var` when possible. I marked a few but realized a lot of the same comment would get tedious. 

There also may be some missing paths for testing.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,255 @@
...
+                methodDecl = generator.WithModifiers(methodDecl, generator.GetModifiers(declaration).WithIsAbstract(false));
+                return generator.WithStatements(methodDecl, defaultMethodBodyStatements);
+            }
+
+            SyntaxNode GeneratePropertyImplementation(IPropertySymbol property)
+            {
+                SyntaxNode propertyDecl = property.IsIndexer
+                    ? generator.IndexerDeclaration(property)
+                    : generator.PropertyDeclaration(property);
+                propertyDecl = generator.WithModifiers(propertyDecl, generator.GetModifiers(declaration).WithIsAbstract(false));
```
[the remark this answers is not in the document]
jkoritzinsky: No. The implementations must be non-virtual, otherwise the runtime will attempt interface dispatch and fail.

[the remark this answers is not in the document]
ryzngard: Do you have a test case covering this? Might be easier for me to check what this is covering so I understand fully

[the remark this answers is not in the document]
jkoritzinsky: The `DynamicInterfaceCastableImplementation_NoPropertiesImplementated_CS_Diagnostic` test should cover this scenario.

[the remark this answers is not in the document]
ryzngard: `DynamicInterfaceCastableImplementation_NoPropertiesImplementated_CS_Diagnostic ` doesn't seem to have anything abstract. Is it even possible for the declaration to be abstract in this case? I guess that's my bigger question. The code suggests yes, but I'm not sure why. It's technically right, but if there's a reason `declaration` could be abstract here a comment is probably warranted as to why.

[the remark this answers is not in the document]
jkoritzinsky: The property declaration in an interface is implicitly abstract and (as part of Default Interface Members) it can be explicitly declared abstract as well (see https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQGYACLEgYRIG8CT6zT0AWEgWQAoBKGuhgXwKD8BWABcApgCcAZhADGEkgEkCtfAxIRgAZzFSFYkuJIAxAPbmaAcwliA3Drv3h/IA).

I'll add some tests around this.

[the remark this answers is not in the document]
ryzngard: I see. In this case abstract properties are the same as other properties on interface declarations?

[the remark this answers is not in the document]
jkoritzinsky: Yep. Any property without a body declared on an interface is considered `abstract`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
jkoritzinsky: There's a few tests for virtual members and I'm adding some tests for explicitly abstract members (in addition to the implicitly-abstract members tests that already exist)

jkoritzinsky:
@sharwell is there an ETA on when this will be reviewed? Last review pass was two weeks ago and the only changes since then were  resolving merge conflicts.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1662,6 +1662,39 @@
...
+  </data>
+  <data name="InterfaceMembersMissingImplementationMessage" xml:space="preserve">
+    <value>Type '{0}' has the 'DynamicInterfaceCastableImplementationAttribute' applied to it but does not provide an implementation of all interface members defined in inherited interfaces</value>
+  </data>
+  <data name="InterfaceMembersMissingImplementationTitle" xml:space="preserve">
+    <value>All members declared in parent interfaces must have an implementation in a DynamicInterfaceCastableImplementation-attributed interface</value>
+  </data>
+  <data name="MembersDeclaredOnImplementationTypeMustBeSealedDescription" xml:space="preserve">
+    <value>Since a type that implements 'IDynamicInterfaceCastable' will conventionally not specify that it implements a 'DynamicInterfaceCastableImplementationAttribute'-attributed type, only the public interface type, virtual interface method lookup will fail even if the method has an implementation. As a result, all members on 'DynamicInterfaceCastableImplementation'-attributed types should be 'sealed'.</value>
```
sharwell, line 1687: > ... will conventionally not specify that it implements a 'DynamicInterfaceCastableImplementationAttribute'-attributed type, only the public interface type, virtual interface method lookup will fail even if the method has an implementation ...

I don't understand this wording. Can you suggest a rephrasing to iterate on it?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,127 @@
...
+            foreach (var iface in targetType.AllInterfaces)
+            {
+                foreach (var member in iface.GetMembers())
+                {
+                    if (!member.IsStatic
+                        && context.Compilation.IsSymbolAccessibleWithin(member, targetType)
+                        && targetType.FindImplementationForInterfaceMember(member) is null)
+                    {
+                        missingMethodImplementations = true;
+                        break;
```
sharwell, line 103: 💡 This only breaks out of the inner loop. It looks like we could break outside the outer loop too for efficiency.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            if (missingMethodImplementations)
+            {
+                context.ReportDiagnostic(targetType.CreateDiagnostic(InterfaceMembersMissingImplementation, targetType.ToDisplayString()));
+            }
+
+            foreach (var member in targetType.GetMembers())
+            {
+                if (member.IsVirtual || member.IsAbstract)
+                {
```
sharwell, line 116: 💭 Would this be more readable if it focused on `sealed`?

```suggestion
            {
                if (!member.IsSealed)
                {
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,80 @@
...
+            SyntaxGenerator generator = SyntaxGenerator.GetGenerator(context.Document);
+            SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            SyntaxNode enclosingNode = root.FindNode(context.Span, getInnermostNodeForTie: true);
+            SyntaxNode declaration = generator.GetDeclaration(enclosingNode);
+            if (declaration == null || !CodeFixSupportsDeclaration(declaration))
+            {
+                return;
+            }
+            Diagnostic diagnostic = context.Diagnostics.First();
```
sharwell, line 39: 💡 This should `foreach` over `context.Diagnostics`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,80 @@
...
+                    diagnostic);
+            }
+        }
+
+        protected abstract bool CodeFixSupportsDeclaration(SyntaxNode declaration);
+
+        protected abstract Task<Document> ImplementInterfacesOnDynamicCastableImplementation(
+            SyntaxNode declaration,
+            Document document,
+            CancellationToken ct);
```
sharwell, line 65: 💡 All `CancellationToken` parameters should be named `cancellationToken`

```suggestion
            CancellationToken cancellationToken);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.Fixer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        private class MyCodeAction : DocumentChangeAction
+        {
+            public MyCodeAction(string title, Func<CancellationToken, Task<Document>> createChangedDocument, string equivalenceKey)
+                : base(title, createChangedDocument, equivalenceKey)
+            {
+            }
+        }
```
sharwell, line 78: 💡 Use `CodeAction.Create` instead of a custom class

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
jkoritzinsky: We can't use `IsSealed` here. `IsSealed` is based on the rules for emitting the `final` bit for the member in metadata, but `sealed` DIM members don't set that bit, they just don't set the `virtual` or `abstract` bits.

jkoritzinsky: SharpLab link for reference: https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyDMABGAHbACmATgGYCGAxmQQJIEDeOBnB+BAzmTQhkY3ACwEAsgAoAlGw5cAvjmW5CaAgCF52Lt0IA3MBWABXQWMmyde1ap4aAwgRBabXHgHsDlCmBiM/ILCltJy7LpKKkA==

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.Fixer.cs
Youssef1313: @sharwell fwiw, I've seen fixers with the 3 patterns: `.Diagnostics.First()`, `.Diagnostics[0]`, and `foreach` across the repository.

So if the `foreach` is always the recommended approach, this could be cleaned up in couple other fixers.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
jkoritzinsky:
```suggestion
    <value>Since a type that implements 'IDynamicInterfaceCastable' will not specify in source or in the `IDynamicInterfaceCastable` implementation that it implements a 'DynamicInterfaceCastableImplementationAttribute'-attributed type, virtual interface method lookup will fail even if the method has an implementation. As a result, all members on 'DynamicInterfaceCastableImplementation'-attributed types must be 'sealed'.</value>
```

sharwell (approved):

jkoritzinsky:
CI failure is https://github.com/dotnet/roslyn-analyzers/issues/5381 and is unrelated to this PR (it's failing in the release/6.0.1xx branch without this change).

PR #5165, opened 2021-06-22
adding performance tests

jmarolf:
merging initial set of performance tests to main (after a lot of validation behind the scenes)

For now
- We are only taking a look at analyzers that will run on the commandline, meaning they must have a default severity of warning or higher
- We have a "baseline" tests that tells us how long this scenario would take if we had an analyzer that does nothing
- We are testing against .NET 6


src/PerformanceTests/Utilities/Common/CompilationHelper.cs
```diff
@@ -0,0 +1,156 @@
...
+            foreach (var (newFileName, source) in primaryProject.AnalyzerConfigFiles)
+            {
+                var documentId = DocumentId.CreateNewId(projectId, debugName: newFileName);
+                solution = solution.AddAnalyzerConfigDocument(documentId, newFileName, source, filePath: newFileName);
+            }
+
+            return solution.GetProject(projectId)!;
+        }
+
+        private static async Task<Solution> CreateSolutionAsync(
```
sharwell, line 80: 💭 This is a bunch of duplication of a library we already reference. Is there a reason we can't just use the code from the test library?

src/PerformanceTests/VisualBasic/Program.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using BenchmarkDotNet.Running;
+
+namespace PerformanceTests
+{
+    public class Program
```
sharwell, line 7: 😕 Does this _need_ to be a separate project?

src/PerformanceTests/VisualBasic/Enabled/VisualBasic_IL3001.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            if (diagnostics.Length != 2 * Constants.Number_Of_Code_Files)
+            {
+                throw new InvalidOperationException($"Expected '2,000' analyzer diagnostics but found '{diagnostics.Length}'");
+            }
+        }
+
+        [Benchmark(Baseline = true)]
+        public void IL3001_Baseline()
```
sharwell, line 66: BenchmarkDotNet supports `async` benchmarks right?

src/PerformanceTests/VisualBasic/Enabled/VisualBasic_IL3001.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            if (analysisResult.CompilationDiagnostics.Count != 0)
+            {
+                throw new InvalidOperationException($"Expected no compilation diagnostics but found '{analysisResult.CompilationDiagnostics.Count}'");
+            }
+
+            if (diagnostics.Length != 2 * Constants.Number_Of_Code_Files)
+            {
+                throw new InvalidOperationException($"Expected '2,000' analyzer diagnostics but found '{diagnostics.Length}'");
```
sharwell, line 61:
```suggestion
                throw new InvalidOperationException($"Expected '{2 * Constants.Number_Of_Code_Files}' analyzer diagnostics but found '{diagnostics.Length}'");
```

src/PerformanceTests/CSharp/Enabled/CSharp_CA1416.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (diagnostics.Length != 1 * Constants.Number_Of_Code_Files)
+            {
+                throw new InvalidOperationException($"Expected '1,000' analyzer diagnostics but found '{diagnostics.Length}'");
+            }
+        }
+
+        [Benchmark(Baseline = true)]
+        public void CA1416_Baseline()
+        {
+            var analysisResult = BaselineCompilationWithAnalyzers.GetAnalysisResultAsync(default).GetAwaiter().GetResult();
```
sharwell, line 87: ❕ Use `CancellationToken.None` for constants. `default` reads as "unknown type which is definitely not a cancellation token".

src/PerformanceTests/VisualBasic/Program.cs
jmarolf: nope, these could be combined.

Issue #3871, opened 2020-07-16
CA1069: Indeterministic false positive


davkean:
#### Analyzer package

?

#### Diagnostic ID

Severity	Code	Description	Project	File	Line	Suppression State
Warning	CA1069	The enum member 'NIL' has the same constant value '-1' as member 'NIL'.	Microsoft.VisualStudio.ProjectSystem.Managed.VS	C:\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\Interop\IVsAppId.cs	52	Active

#### Repro steps

This one comes and goes and I cannot pin down why/when it appears.

In this file: https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Interop/IVsAppId.cs, I've seen this warning appear for both https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Interop/IVsAppId.cs#L54 and https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Interop/IVsAppId.cs#L283, it fires for no other enum values in that file.

mavasani:
@Evangelink Any ideas?

PR #5170, opened 2021-06-25
CA1069: Fix FP caused by concurrency issue

### LINKED ISSUE #4900: [.NET 5 built in Analyzer] CA1069 False Positive when scanning enum values.
When analyzing the following enum it choked up saying the value is equal to itself and must not be duplicated 😂.

```cs
        /// <summary>Values to pass to the GetDCEx method.</summary>
        [Flags]
        public enum DeviceContextValues : uint
        {
            /// <summary>DCX_WINDOW: Returns a DC that corresponds to the window rectangle rather
            /// than the client rectangle.</summary>
            Window = 0x00000001,

            /// <summary>DCX_CACHE: Returns a DC from the cache, rather than the OWNDC or CLASSDC
            /// window. Essentially overrides CS_OWNDC and CS_CLASSDC.</summary>
            Cache = 0x00000002,

            /// <summary>DCX_NORESETATTRS: Does not reset the attributes of this DC to the
            /// default attributes when this DC is released.</summary>
            NoResetAttrs = 0x00000004,

            /// <summary>DCX_CLIPCHILDREN: Excludes the visible regions of all child windows
            /// below the window identified by hWnd.</summary>
            ClipChildren = 0x00000008,

            /// <summary>DCX_CLIPSIBLINGS: Excludes the visible regions of all sibling windows
            /// above the window identified by hWnd.</summary>
            ClipSiblings = 0x00000010,

            /// <summary>DCX_PARENTCLIP: Uses the visible region of the parent window. The
            /// parent's WS_CLIPCHILDREN and CS_PARENTDC style bits are ignored. The origin is
            /// set to the upper-left corner of the window identified by hWnd.</summary>
            ParentClip = 0x00000020,

            /// <summary>DCX_EXCLUDERGN: The clipping region identified by hrgnClip is excluded
            /// from the visible region of the returned DC.</summary>
            ExcludeRgn = 0x00000040,

            /// <summary>DCX_INTERSECTRGN: The clipping region identified by hrgnClip is
            /// intersected with the visible region of the returned DC.</summary>
            IntersectRgn = 0x00000080,

            /// <summary>DCX_EXCLUDEUPDATE: Unknown...Undocumented.</summary>
            ExcludeUpdate = 0x00000100,

            /// <summary>DCX_INTERSECTUPDATE: Unknown...Undocumented.</summary>
            IntersectUpdate = 0x00000200,

            /// <summary>DCX_LOCKWINDOWUPDATE: Allows drawing even if there is a LockWindowUpdate
            /// call in effect that would otherwise exclude this window. Used for drawing during
            /// tracking.</summary>
            LockWindowUpdate = 0x00000400,

            /// <summary>DCX_VALIDATE When specified with DCX_INTERSECTUPDATE, causes the DC to
            /// be completely validated. Using this function with both DCX_INTERSECTUPDATE and
            /// DCX_VALIDATE is identical to using the BeginPaint function.</summary>
            Validate = 0x00200000,

            /// <summary>Unknown... Undocumented.</summary>
            Unknown = 0x10000,
        }
```
I usually have this enum inside of an ``NativeMethods`` class for Windows that basically then uses these flags to make a screenshot of the entire monitors (and then try to also screenshot any other monitors it finds and then combine them together).

You should notice it would either: 1. warn, 2. error with warn as error, or 3. suggest to fix the "issue" in the code.

Evangelink:
Fixes #4900 .

I don't remember how to generate a new VSIX to test, could you remind me please?

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
@@ -56,7 +56,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     return;
                 }
 
-                var membersByValue = new ConcurrentDictionary<object, ConcurrentBag<(SyntaxNode fieldSyntax, string fieldName)>>();
+                var membersByValue = new ConcurrentDictionary<object, ConcurrentDictionary<(SyntaxNode fieldSyntax, string fieldName), Unit>>();
 
                 // Workaround to get the value of all enum fields that don't have an explicit initializer.
                 // We will start with all of the enum fields and remove the ones with an explicit initializer.
@@ -86,18 +86,18 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
 
                     var onlyReferencesOneField = GetFilteredDescendants(fieldInitializerOperation, op => op.Kind != OperationKind.Binary)
                         .OfType<IFieldReferenceOperation>()
-                        .Where(fro => fro.Field.ContainingType.Equals(enumSymbol))
-                        .Any();
+                        .Any(fro => fro.Field.ContainingType.Equals(enumSymbol));
                     if (onlyReferencesOneField)
                     {
                         return;
                     }
 
-                    membersByValue.AddOrUpdate(fieldInitializerOperation.Value.ConstantValue.Value,
-                        new ConcurrentBag<(SyntaxNode, string)> { (fieldInitializerOperation.Syntax.Parent, currentField.Name) },
+                    membersByValue.AddOrUpdate(
+                        fieldInitializerOperation.Value.ConstantValue.Value,
+                        new ConcurrentDictionary<(SyntaxNode fieldSyntax, string fieldName), Unit> { [(fieldInitializerOperation.Syntax.Parent, currentField.Name)] = Unit.Default },
                         (key, value) =>
                         {
-                            value.Add((fieldInitializerOperation.Syntax.Parent, currentField.Name));
+                            value.AddOrUpdate((fieldInitializerOperation.Syntax.Parent, currentField.Name), Unit.Default, (key, value) => value);
                             return value;
                         });
                 }, OperationKind.FieldInitializer);
@@ -112,19 +112,20 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
 
                         if (fieldSyntax != null && value != null)
                         {
-                            membersByValue.AddOrUpdate(value,
-                                new ConcurrentBag<(SyntaxNode, string)> { (fieldSyntax, field.Name) },
-                                (k, v) =>
+                            membersByValue.AddOrUpdate(
+                                value,
+                                new ConcurrentDictionary<(SyntaxNode, string), Unit> { [(fieldSyntax, field.Name)] = Unit.Default },
+                                (key, value) =>
                                 {
-                                    v.Add((fieldSyntax, field.Name));
-                                    return v;
+                                    value.AddOrUpdate((fieldSyntax, field.Name), Unit.Default, (key, value) => value);
+                                    return value;
                                 });
                         }
                     }
 
                     foreach (var kvp in membersByValue)
                     {
-                        var orderedItems = kvp.Value.OrderBy(x => x.fieldSyntax.GetLocation().SourceSpan);
+                        var orderedItems = kvp.Value.Keys.OrderBy(x => x.fieldSyntax.GetLocation().SourceSpan);
                         var duplicatedMemberName = orderedItems.FirstOrDefault().fieldName;
 
                         foreach ((var fieldSyntax, var fieldName) in orderedItems.Skip(1))


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -56,7 +56,7 @@ public sealed override void Initialize(AnalysisContext context)
                     return;
                 }
 
-                var membersByValue = new ConcurrentDictionary<object, ConcurrentBag<(SyntaxNode fieldSyntax, string fieldName)>>();
+                var membersByValue = new ConcurrentDictionary<object, ConcurrentDictionary<(SyntaxNode fieldSyntax, string fieldName), Unit>>();
```
Evangelink, line 59: I don't recall if `GetLocation` is an expensive call, if not maybe we could directly store the location span rather than the `SyntaxNode`.

Youssef1313:
> Fixes #4900 .

Also https://github.com/dotnet/roslyn-analyzers/issues/3871.

Evangelink:
@Youssef1313 Do you remember how to generate an experimental version to be able to test in VS directly? I'd like to ensure the issue is correctly fixed.

Youssef1313:
> @Youssef1313 Do you remember how to generate an experimental version to be able to test in VS directly? I'd like to ensure the issue is correctly fixed.

If you mean VSIX, then I don't know :(
But you could probably do `msbuild /t:pack` and get the package from artifacts and test it in a normal Visual Studio instance.

sharwell (changes requested):
Marking Request Changes only so I have more time to validate this before merge.

Evangelink:
@sharwell Do you think we could update the test framework to allow a run in parallel so that we could catch more easily such kind of issue? I mean yes of course there is a design problem in the implementation and with more experience we could avoid these but maybe we could have a little help from the test infra to prevent or notice these problems earlier.

sharwell:
@Evangelink The test framework already does run in parallel.

Evangelink:
Oh ok, that's weird then. I am thinking it might be a bug from roslyn that would callback multiple time for the same node/operation. But being honest I have no idea how to get any deeper and have any proof of whatsoever. Sorry to drop the baby on you like this. If you find out something and are willing to share how you did the exploration/debugging, I'd be super interested (willing to screen-share with you if you are ok with it :)).

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -86,18 +86,18 @@ public sealed override void Initialize(AnalysisContext context)
...
-                        new ConcurrentBag<(SyntaxNode, string)> { (fieldInitializerOperation.Syntax.Parent, currentField.Name) },
+                    membersByValue.AddOrUpdate(
+                        fieldInitializerOperation.Value.ConstantValue.Value,
+                        new ConcurrentDictionary<(SyntaxNode fieldSyntax, string fieldName), Unit> { [(fieldInitializerOperation.Syntax.Parent, currentField.Name)] = Unit.Default },
                         (key, value) =>
                         {
-                            value.Add((fieldInitializerOperation.Syntax.Parent, currentField.Name));
+                            value.AddOrUpdate((fieldInitializerOperation.Syntax.Parent, currentField.Name), Unit.Default, (key, value) => value);
                             return value;
                         });
```
sharwell, line 102: I'm suspicious of this code as the cause of the problem. In general, `AddOrUpdate` allows `updateValueFactory` to get called more than once (in which case the `Add` call would clearly occur too many times). However, it's not clear that this would be possible in the specific case where all calls to `AddOrValue` return `value` unchanged (still trying to verify this).

If I'm correct regarding the underlying problem, perhaps a better fix would be:

* Alter the call to `AddOrUpdate` to `GetOrAdd`
* After obtaining the `ConcurrentBag` from the dictionary, make a single call to `Add` which is guaranteed to only run once

sharwell: 🔗 You can see the loop which calls `updateValueFactory` (potentially) multiple times here:
https://github.com/dotnet/runtime/blob/91599fff26619cb8bccb7b39c39590e7cd9a4379/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1276-L1284

The pending verification is whether `TryUpdateInternal` (or `TryUpdate` on .NET Framework) is capable of returning false if values are never modified in the collection after addition.

Evangelink: @sharwell How shall we proceed for this PR? Keep `ConcurrentBag` and have a more complex update function or keep going with the current proposed API?

sharwell: I'd prefer the code change to use:

```
ConcurrentDictionary<object, ImmutableHashSet<(SyntaxNode fieldSyntax, string fieldName)>>
```

Evangelink: Couple of questions:
- `ImmutableHashSet` is enough because the concurrency is already handled by the `ConcurrentDictionary`?
- Any reason to have immutable data for the value while keeping a mutable data for the outer type?

Youssef1313:
Another fix was made in #5243

PR #5181, opened 2021-06-29
New rule: prefer Span.Clear over Span.Fill(default)

huoyaoyuan:
Resolves https://github.com/dotnet/runtime/issues/33813

I'm taking CA1849 since another PR is already taking CA1848.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1674,4 +1674,16 @@
...
+    <value>Use 'Clear()'</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillDescription" xml:space="preserve">
+    <value>It is more efficient to use 'Clear', instead of 'Fill' with default value.</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillMessage" xml:space="preserve">
+    <value>Use 'Clear' instead of 'Fill' with default value for spans</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillTitle" xml:space="preserve">
+    <value>Use Clear instead of Fill with default value</value>
```
Youssef1313, line 1687:
```suggestion
    <value>Use 'Clear' instead of 'Fill' with default value</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (node is null)
+            {
+                return;
+            }
+
+            var invocationTarget = GetInvocationTarget(node);
+            if (invocationTarget == null)
+            {
+                return;
+            }
```
Youssef1313, line 39: `node is null` check is redundant. `GetInvocationTarget` will already return `null` if the `node` is `null`. Consider making the node parameter of `GetInvocationTarget` nullable, and remove the node null check above.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,62 @@
...
+                        DocumentEditor editor = await DocumentEditor.CreateAsync(context.Document, cancellationToken).ConfigureAwait(false);
+                        SyntaxGenerator generator = editor.Generator;
+
+                        var memberAccess = generator.MemberAccessExpression(invocationTarget, UseSpanClearInsteadOfFillAnalyzer.ClearMethod);
+                        var invocation = generator.InvocationExpression(memberAccess);
+
+                        editor.ReplaceNode(node, invocation);
+                        return editor.GetChangedDocument();
+                    },
+                    equivalenceKey: diagnostic.Id),
```
Youssef1313, line 56:
```suggestion
                    equivalenceKey: nameof(MicrosoftNetCoreAnalyzersResources.UseSpanClearInsteadOfFillCodeFixTitle)),
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        private void OnCompilationStart(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+            var spanType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSpan1);
+
+            if (spanType == null)
+            {
+                return;
+            }
```
Youssef1313, line 56:
```suggestion
            if (!compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSpan1, out var spanType))
            {
                return;
            }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,155 @@
...
+                    {
+                        return;
+                    }
+
+                    if (IsDefaultValue(invocation.Arguments[0]))
+                    {
+                        operationContext.ReportDiagnostic(Diagnostic.Create(s_Rule, invocation.Syntax.GetLocation()));
+                    }
+                },
+                OperationKind.Invocation);
```
Youssef1313, line 83: Renamed `operationContext` to `context`.

```suggestion
                context =>
                {
                    var invocation = (IInvocationOperation)context.Operation;

                    if (!SymbolEqualityComparer.Default.Equals((invocation.Instance.Type as INamedTypeSymbol)?.ConstructedFrom, spanType))
                    {
                        return;
                    }

                    if (invocation.TargetMethod.Name != FillMethod)
                    {
                        return;
                    }

                    if (invocation.Arguments.Length != 1)
                    {
                        return;
                    }

                    if (IsDefaultValue(invocation.Arguments[0]))
                    {
                        context.ReportDiagnostic(Diagnostic.Create(s_Rule, invocation.Syntax.GetLocation()));
                    }
                },
                OperationKind.Invocation);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,155 @@
...
+
+                    if (invocation.TargetMethod.Name != FillMethod)
+                    {
+                        return;
+                    }
+
+                    if (invocation.Arguments.Length != 1)
+                    {
+                        return;
+                    }
```
Youssef1313, line 76: I'm not sure. But if these checks are cheaper than symbol comparison, consider moving them up.

Youssef1313: Actually, I'd try getting the method symbol at compilation start, then do symbol comparison against `invocation.TargetMethod`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,155 @@
...
+
+                    if (IsDefaultValue(invocation.Arguments[0]))
+                    {
+                        operationContext.ReportDiagnostic(Diagnostic.Create(s_Rule, invocation.Syntax.GetLocation()));
```
Youssef1313, line 80: Use `CreateDiagnostic` extensions. It's available on `IOperation`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
```diff
@@ -0,0 +1,250 @@
...
+            var test = new VerifyCS.Test
+            {
+                TestCode = source,
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                LanguageVersion = LanguageVersion.Preview,
+            };
+
+            test.ExpectedDiagnostics.AddRange(expected);
+            return test.RunAsync();
+        }
```
Youssef1313, line 248: I don't see that you're using the `expected` params.

```suggestion
        private static Task TestCS(string source, string corrected)
        {
            var test = new VerifyCS.Test
            {
                TestCode = source,
                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
                LanguageVersion = LanguageVersion.Preview,
                FixedCode = corrected,
            };

            return test.RunAsync();
        }

        private static Task TestCS(string source)
        {
            var test = new VerifyCS.Test
            {
                TestCode = source,
                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
                LanguageVersion = LanguageVersion.Preview,
            };

            return test.RunAsync();
        }
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Performance
+{
+    /// <summary>
+    /// CA1849: C# implementation of use Span.Clear instead of Span.Fill(default)
+    /// Implements the <see cref="CodeFixProvider" />
+    /// </summary>
+    /// <seealso cref="UseSpanClearInsteadOfFillFixer"/>
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public class CSharpUseSpanClearInsteadOfFillFixer : UseSpanClearInsteadOfFillFixer
```
Youssef1313, line 17: VB fixer and tests are missing.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,62 @@
...
+                    createChangedDocument: async cancellationToken =>
+                    {
+                        DocumentEditor editor = await DocumentEditor.CreateAsync(context.Document, cancellationToken).ConfigureAwait(false);
+                        SyntaxGenerator generator = editor.Generator;
+
+                        var memberAccess = generator.MemberAccessExpression(invocationTarget, UseSpanClearInsteadOfFillAnalyzer.ClearMethod);
+                        var invocation = generator.InvocationExpression(memberAccess);
+
+                        editor.ReplaceNode(node, invocation);
+                        return editor.GetChangedDocument();
```
Youssef1313, line 54: Instead of that, why not replace `GetInvocationTarget` with another method that returns `MemberAccessExpressionSyntax.Name` instead of `MemberAccessExpressionSyntax.Expression`, and just replace this node with `SyntaxFactory.IdentifierName("Clear")`

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
huoyaoyuan: VB can't use `Span`. It's actually not applicable to VB.

Youssef1313: @huoyaoyuan The analyzer is marked as supporting VB. That should be fixed then if it's only applicable to C#.

huoyaoyuan: I'm doing the same thing with CA1831. It's providing a analyzer for both language, and only a fixer for C#.

Youssef1313: @huoyaoyuan Can a diagnostic be reported in VB? If yes, then tests are needed. If no, then I don't see why providing the analyzer for VB is needed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
```diff
@@ -0,0 +1,250 @@
...
+
+class C
+{{
+    void M(Span<{type}> span)
+    {{
+        [|span.Fill({value})|];
+    }}
+}}
+";
+            await TestCS(source);
```
Youssef1313, line 74: Codefix should be tested here.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
```diff
@@ -0,0 +1,250 @@
...
+
+class C
+{
+    void M<T>(Span<T> span)
+    {
+        [|span.Fill(default)|];
+    }
+}
+";
+            await TestCS(source);
```
Youssef1313, line 119: Codefix should be tested here.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
```diff
@@ -0,0 +1,250 @@
...
+
+class C
+{
+    void M<T>(Span<T?> span) where T : class
+    {
+        [|span.Fill(null)|];
+    }
+}
+";
+            await TestCS(source);
```
Youssef1313, line 138: Codefix should be tested here.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
huoyaoyuan: It should be able to report diagnostic. The test framework isn't capable to test analyzer with fixer.
Anyway, I'm aligning with UseAsSpanInsteadOfRangeIndexer.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
huoyaoyuan: The fixed result is too simple that I don't want to bother bloating the test code.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
Youssef1313: > The test framework isn't capable to test analyzer with fixer.

Assuming you meant to say "without fixer", then no, the test framework can test analyzers without fixers. You only need to use `Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
Youssef1313: > The fixed result is too simple that I don't want to bother bloating the test code.

I personally disagree. But deferring to @mavasani or @sharwell.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,155 @@
...
+                        return BitConverter.DoubleToInt64Bits(f) == 0;
+
+                    case double d:
+                        return BitConverter.DoubleToInt64Bits(d) == 0;
+
+                    case decimal d:
+                        return decimal.GetBits(d).All(b => b == 0);
+
+                    default:
+                        return false;
```
Youssef1313, line 129: Does this handle the following case?

```
    public void M(Span<nint> span)
    {
        span.Fill(0);
    }
```

Consider adding a test.

Youssef1313: Considering testing `Span<char>` with provided value `\0`

Youssef1313: Consider testing `Span<bool>` as well.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,155 @@
...
+                }
+            }
+
+            if (value is IDefaultValueOperation)
+            {
+                return SymbolEqualityComparer.Default.Equals(value.Type, type);
```
Youssef1313, line 135: Probably will cause false negative for the following code?

```csharp
using System;
public class Base {}
public class Derived : Base {}

public class C
{
    public void M(Span<Base> span)
    {
        span.Fill(default(Derived));
    }
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFIll.Fixer.cs
huoyaoyuan: I don't understand what you mean. The method and arguments both need to be updated, so updating the whole invocation expression is natural.

Youssef1313: Yeah my comment doesn't make sense at all 😄

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
huoyaoyuan: This will get `null` in constant value and handled earlier.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public sealed override FixAllProvider GetFixAllProvider()
+        {
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var node = root.FindNode(context.Span);
```
Youssef1313, line 28: It might be safer to pass `getInnermostNodeForTie: true`

I cannot think of any current scenario where this can make a difference (except an unlikely error scenario where a user does `FunctionCall(span.Fill(..))`. But it's safer to have `getInnermostNodeForTie`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,149 @@
...
+                return SymbolEqualityComparer.Default.Equals(defaultValue.Type, type);
+            }
+
+            if (value is IObjectCreationOperation objectCreation)
+            {
+                return objectCreation.Type.IsValueType
+                    && objectCreation.Arguments.IsEmpty
+                    && (objectCreation.Initializer?.Initializers.IsEmpty ?? true)
+                    && SymbolEqualityComparer.Default.Equals(value.Type, type);
+            }
```
Youssef1313, line 144: If I understand this correctly. I think this doesn't play nice with struct parameterless constructors feature?

```csharp
using System;

C c = default;
C c2 = new C();

Console.WriteLine(c.X); // 0
Console.WriteLine(c2.X); // 5

public struct C
{
    public int X;
    public C() => X = 5;
}
```

huoyaoyuan: Yes. I wrote the code when struct parameterless constructor was not a thing. It may be easier to test when ReferenceAssemblies.Net60 comes?

Youssef1313: @huoyaoyuan I think it's not related to .NET 6 RefenceAssemblies. Just a compiler version that has the feature is enough. If there is an available version already, it should be just upgrading `MicrosoftCodeAnalysisVersionForTests` to that version.

Also Net60 was actually added in https://github.com/dotnet/roslyn-sdk/commit/de9af89015295d30088cf95477c9079eb3c3d1b4, but not sure if a version is out or not.

huoyaoyuan: The feature was merged in roslyn by 17.0p3, however this repo is still referencing 4.0.0-2 . So I can't test it right now.

Youssef1313: @huoyaoyuan Did you try upgrading to `4.0.0-3.final` and encountered any issues?

huoyaoyuan:
Since this is included in 7.0 meta issue, should I retarget the PR to 7.0.1xx branch?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
huoyaoyuan: It should be fine, but I think the version should be updated by a separated PR like #5274 . /cc @sharwell

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
mavasani: > I'd want to remove the declaration of VB support, since there's always error BC30668. Is it OK?

I think it is reasonable to move VB support and tests to a separate PR. However, if we want to completely drop VB support for this analyzer, a confirmation is needed from the team on https://github.com/dotnet/runtime/issues/33813.
@buyaa-n @carlossanlop what do you think?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,150 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class UseSpanClearInsteadOfFillAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "CA1851";
+        internal const string FillMethod = "Fill";
+        internal const string ClearMethod = "Clear";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseSpanClearInsteadOfFillTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseSpanClearInsteadOfFillMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseSpanClearInsteadOfFillDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
```
mavasani, line 26: You will need to replace calls to `new LocalizableResourceString` with `CreateLocalizableResourceString` like below:

https://github.com/dotnet/roslyn-analyzers/blob/50d35c8d1c50a8e4dc6d0327be904495c23c5e19/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/ImmutableCollections/DoNotCallToImmutableCollectionOnAnImmutableCollectionValue.cs#L28

This was a repo wide change that was done after your PR was first created.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,150 @@
...
+                return;
+            }
+
+            context.RegisterOperationAction(
+                context =>
+                {
+                    var invocation = (IInvocationOperation)context.Operation;
+
+                    if (SymbolEqualityComparer.Default.Equals(invocation.TargetMethod.OriginalDefinition, spanFillMethod)
+                        && IsDefaultValue(invocation.Arguments[0]))
```
mavasani, line 71: Please also test error code where no arguments are provided to `Fill` method call, such as `span.Fill();`. I think this case will not generate an `IInvocationOperation`, but good to have a test to verify it. A more defensive implementation would be to just add an explicit check `invocation.Arguments.Length == 1` before accessing `invocation.Arguments[0]`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,150 @@
...
+                    // 0 isn't default value for T?
+                    return false;
+                }
+
+                // enum/nint are treated as integers
+                // This is missing default value of DateTime literal for VB,
+                // but since VB doesn't properly support Span, just don't consider it
+                switch (constantOpt.Value)
+                {
+#pragma warning disable IDE0004 // Remove Unnecessary Cast - false positive
```
mavasani, line 104: Can we file an issue on Roslyn repo for this false positive? I wonder if the false positive is now resolved as @CyrusNajmabadi has completely rewritten remove unnecessary cast analyzer.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
mavasani: Please do add a test for this case, if you haven't already done so. Thanks!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,150 @@
...
+                return SymbolEqualityComparer.Default.Equals(value.Type, type);
+            }
+
+            if (value is IConversionOperation { Operand: IDefaultValueOperation defaultValue })
+            {
+                // handle the conversion of default literal only
+                return SymbolEqualityComparer.Default.Equals(defaultValue.Type, type);
+            }
+
+            if (value is IObjectCreationOperation objectCreation)
```
mavasani, line 138: Can you add a comment for the scenario handled inside this if statement?

mavasani (approved):
Overall the C# analyzer/fixer/tests look good to me.

For VB, please confirm on the dotnet/runtime issue whether VB support is desired or not? If desired, and you wish to postpone it to a separate PR, please file a tracking issue in this repo for VB part.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
stephentoub: I think it makes sense to drop VB support. As noted, VB doesn't have anything beyond bare minimum span support.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
```diff
@@ -0,0 +1,330 @@
...
+        public async Task TestGeneric_ValueType_DefaultT()
+        {
+            string source = @"
+using System;
+
+class C
+{
+    void M<T>(Span<T?> span) where T : struct
+    {
+        span.Fill(default(T));
```
buyaa-n, line 248: Why it is not flagging this?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1834,4 +1834,16 @@
...
+    <value>Use 'Clear()'</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillDescription" xml:space="preserve">
+    <value>It is more efficient to use 'Clear', instead of 'Fill' with default value.</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillMessage" xml:space="preserve">
+    <value>Use 'Clear' instead of 'Fill' with default value for spans</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillTitle" xml:space="preserve">
+    <value>Use 'Clear' instead of 'Fill' with default value</value>
```
buyaa-n, line 1847: NIT: The messaging looks need a bit more improvement
```suggestion
    <value>Prefer 'Clear' over 'Fill'</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1834,4 +1834,16 @@
...
     <value>'{0}' uses the preview type '{1}' and needs to opt into preview features. See {2} for more information.</value>
   </data>
+  <data name="UseSpanClearInsteadOfFillCodeFixTitle" xml:space="preserve">
+    <value>Use 'Clear()'</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillDescription" xml:space="preserve">
+    <value>It is more efficient to use 'Clear', instead of 'Fill' with default value.</value>
+  </data>
+  <data name="UseSpanClearInsteadOfFillMessage" xml:space="preserve">
+    <value>Use 'Clear' instead of 'Fill' with default value for spans</value>
```
buyaa-n, line 1844: NIT:
```suggestion
    <value>Prefer 'Span<T>.Clear()' instead of 'Span<T>.Fill(default)'</value>
```

buyaa-n:
> For VB, please confirm on the dotnet/runtime issue whether VB support is desired or not? If desired, and you wish to postpone it to a separate PR, please file a tracking issue in this repo for VB part.

@mavasani [its confirmed on the issue](https://github.com/dotnet/runtime/issues/33813#issuecomment-1027888278) that we don't need VB support

Overall the PR looks good to me, @huoyaoyuan could you address a few remaining comments and merge with latest so that can merge your PR. Please let us know we contribute to the PR if needed, thanks!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFillTests.cs
huoyaoyuan: Because `default(T)` is not `default(Nullable<T>)` which is `null`. Notice the struct constraint.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
huoyaoyuan: The test version has been updated now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
huoyaoyuan: It still repros at VS 17.4P1. I remember the rewritten was delayed?

Youssef1313: @huoyaoyuan https://github.com/dotnet/roslyn/pull/56932

Youssef1313: The rewrite was shipped in 17.1. But the code style packages used in this repo are 4.0

https://github.com/dotnet/roslyn-analyzers/blob/793113a41d7f21b03470521bf48438f2abd9b12f/eng/Versions.props#L42-L44

So we are not yet consuming the fix from https://github.com/dotnet/roslyn/pull/56932

huoyaoyuan:
I've resolved most of the comments. Sorry for the delay because I don't have much time.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Performance
+{
+    using static MicrosoftNetCoreAnalyzersResources;
+
+    /// <summary>
+    /// CA1851: Use Span.Clear instead of Span.Fill(default)
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```
Youssef1313, line 19: Wasn't it decided to drop VB support?

huoyaoyuan: It was not clear what "support" means to me. There should never be diagnostics reported for VB, but I want to be consistent with other analyzers too.

Youssef1313: > There should never be diagnostics reported for VB

In that case, there shouldn't be `LanguageNames.VisualBasic`.

@mavasani Is adding a suppression like the following good enough?

https://github.com/dotnet/roslyn-analyzers/blob/793113a41d7f21b03470521bf48438f2abd9b12f/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotDeclareEventFieldsAsVirtual.cs#L13-L15

Or is it better to have an abstract analyzer in `Core` and a concrete one in `CSharp`?

mavasani: @Youssef1313 Either way sounds good to me. Generally we follow the latter approach.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFill.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Performance
+{
+    /// <summary>
+    /// CA1855: C# implementation of use Span.Clear instead of Span.Fill(default)
+    /// Implements the <see cref="UseSpanClearInsteadOfFillAnalyzer" />
+    /// </summary>
+    /// <seealso cref="UseSpanClearInsteadOfFillFixer"/>
+    [DiagnosticAnalyzer(LanguageNames.CSharp), Shared]
```
Youssef1313, line 14:
```suggestion
    [DiagnosticAnalyzer(LanguageNames.CSharp)]
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.NetCore.Analyzers.Performance;
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Performance
+{
+    /// <summary>
+    /// CA1855: C# implementation of use Span.Clear instead of Span.Fill(default)
+    /// Implements the <see cref="CodeFixProvider" />
+    /// </summary>
+    /// <seealso cref="UseSpanClearInsteadOfFillFixer"/>
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
```
buyaa-n, line 16: NIT
```suggestion
    [ExportCodeFixProvider(LanguageNames.CSharp)]
```

buyaa-n (approved):
LGTM, thanks!

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Performance/CSharpUseSpanClearInsteadOfFIll.Fixer.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using System.Composition;
```
buyaa-n, line 3:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseSpanClearInsteadOfFill.cs
buyaa-n: @mavasani do we want to update this version with this PR? Seems it is only open question in this PR

Youssef1313: @buyaa-n Updating requires more changes. I'm handling that in another PR.

buyaa-n: OK, just making sure all comments resolved

PR #5210, opened 2021-07-05
Update docs on how to record removed API

RussKie:

### DIFF
--- a/src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
+++ b/src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
@@ -161,7 +161,7 @@
   </data>
   <data name="RemoveDeletedApiDescription" xml:space="preserve">
     <value>When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
-    <note>{Locked="*REMOVED*"}</note>
+    <comment>{Locked="*REMOVED*"}</comment>
   </data>
   <data name="ExposedNoninstantiableTypeMessage" xml:space="preserve">
     <value>Constructor makes its noninheritable base class inheritable, thereby exposing its protected members</value>
--- a/src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
+++ b/src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
@@ -160,7 +160,8 @@
     <value>Symbol '{0}' is part of the declared API, but is either not public or could not be found</value>
   </data>
   <data name="RemoveDeletedApiDescription" xml:space="preserve">
-    <value>When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
+    <value>When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
+    <note>{Locked="*REMOVED*"}</note>
   </data>
   <data name="ExposedNoninstantiableTypeMessage" xml:space="preserve">
     <value>Constructor makes its noninheritable base class inheritable, thereby exposing its protected members</value>
--- a/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
+++ b/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
@@ -14,7 +14,7 @@ All public types and members should be declared in PublicAPI.txt. This draws att
 
 ## [RS0017](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md): Remove deleted types and members from the declared API
 
-When removing a public type or member put that entry in PublicAPI.Unshipped.txt with *REMOVED* prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
+When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
 
 |Item|Value|
 |-|-|
--- a/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
+++ b/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
@@ -14,7 +14,7 @@ All public types and members should be declared in PublicAPI.txt. This draws att
 
 ## [RS0017](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md): Remove deleted types and members from the declared API
 
-When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
+When removing a public type or member put that entry in PublicAPI.Unshipped.txt with *REMOVED* prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
 
 |Item|Value|
 |-|-|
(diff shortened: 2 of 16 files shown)


src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
```diff
@@ -160,7 +160,7 @@
     <value>Symbol '{0}' is part of the declared API, but is either not public or could not be found</value>
   </data>
   <data name="RemoveDeletedApiDescription" xml:space="preserve">
-    <value>When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
+    <value>When removing a public type or member put that entry in PublicAPI.Unshipped.txt with *REMOVED* prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
```
Youssef1313, line 163: Also it's safer to lock `*REMOVED*` from localization. Syntax for locking localization:

https://github.com/dotnet/roslyn-analyzers/blob/affa88c580a9cffc5b016ad892c5bcd3971e8456/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx#L353

```suggestion
    <value>When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
```

src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
```diff
@@ -14,7 +14,7 @@ All public types and members should be declared in PublicAPI.txt. This draws att
 
 ## [RS0017](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md): Remove deleted types and members from the declared API
 
-When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
+When removing a public type or member put that entry in PublicAPI.Unshipped.txt with *REMOVED* prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
```
Youssef1313, line 17: 🐛 This is a bug in markdown generation. The asterisks should be escaped. I'll file a follow-up issue on that.

RussKie: Good catch, it didn't occur to me.

drewnoakes (commented):
Is the REMOVED string a symbol that should not be translated?

Youssef1313:
> Is the REMOVED string a symbol that should not be translated?

Yeah I think it shouldn't. I requested to use "Locked" in https://github.com/dotnet/roslyn-analyzers/pull/5210#discussion_r665960432, but I think this part of the commend was unintentionally dropped.

RussKie:
> I requested to use "Locked" in [#5210 (comment)](https://github.com/dotnet/roslyn-analyzers/pull/5210#discussion_r665960432), but I think this part of the commend was unintentionally dropped.

Totally missed it! Thank you for pointing it out.

Youssef1313:
@RussKie No nees to manually update doc files. You can run `msbuild /t:pack`, they will get updates automatically.

RussKie:
I have... It does something very bizarre like replacing the string I modified with a completely different string. I'll try again later...

src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
```diff
@@ -14,7 +14,7 @@ All public types and members should be declared in PublicAPI.txt. This draws att
 
 ## [RS0017](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md): Remove deleted types and members from the declared API
 
-When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
+When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '\*REMOVED\*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
```
Youssef1313, line 17: This suggestion removes the "correct" escaping. This should be fixed in the tool generating the doc file.

```suggestion
When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.
```

src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.sarif
```diff
@@ -31,7 +31,7 @@
         "RS0017": {
           "id": "RS0017",
           "shortDescription": "Remove deleted types and members from the declared API",
-          "fullDescription": "When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.",
+          "fullDescription": "When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '\*REMOVED\*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.",
```
Youssef1313, line 34:
```suggestion
          "fullDescription": "When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.",
```

src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md
RussKie: Thank you.

Youssef1313:
> I have... It's doing something very bizarre like replacing the string I modified with a completely different string. I'll try again later...

If it's about the escaping backslash, then it's okay, fixing the tool is tracked by #5219.
If it's doing some other weird thing, it could be a another unknown bug with the tool.

RussKie:
I have rebased my branch on top of the latest main, `git clean -xdf`, followed by `build.cmd -pack`. As a result I get this changes:
image

This makes no sense to me...

Youssef1313:
@RussKie I can reproduce. For some reason, the description of the diagnostic descriptor is empty:

image

I don't really get why :(

Youssef1313:
The weird thing is that this problem doesn't exist in CI (ie, the CI is failing and is most likely expecting the correct result, at the same time, both CI and local development relies on the same generation code)

Youssef1313:
I decompiled the produced artifacts using Telerik JustDecompile. Here is the resources I see in the decompiler:

image

The dll I decompiled is located at `<repo-root>\artifacts\bin\Microsoft.CodeAnalysis.PublicApiAnalyzers\Debug\netstandard1.3\Microsoft.CodeAnalysis.PublicApiAnalyzers.dll`

I have no idea how to dig further. But I hope you find this information useful.

src/PublicApiAnalyzers/Core/Analyzers/PublicApiAnalyzerResources.resx
```diff
@@ -160,7 +160,8 @@
     <value>Symbol '{0}' is part of the declared API, but is either not public or could not be found</value>
   </data>
   <data name="RemoveDeletedApiDescription" xml:space="preserve">
-    <value>When removing a public type or member the corresponding entry in PublicAPI.txt should also be removed. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
+    <value>When removing a public type or member, put that entry in PublicAPI.Unshipped.txt with '*REMOVED*' prefix. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.</value>
+    <note>{Locked="*REMOVED*"}</note>
```
Youssef1313, line 164: Ohhh, this should have been `<comment>` not `<note>`

I haven't confirmed if that fixes the generation, but it should, I can't imagine there is any other issue.

RussKie: 🤦‍♂️ I looked at a wrong file.... in xlf files the "comment" is turned into a "note"

RussKie:
Yay! It is finally green!

Thank you @Youssef1313 so much for your help!

jmarolf (approved):

PR #5220, opened 2021-07-09
Escape asterisks when generating markdown doc files

### LINKED ISSUE #5219: 🐛 Asterisk should be escaped in generated markdown files
🐛 This is a bug in markdown generation. The asterisks should be escaped. I'll file a follow-up issue on that.

_Originally posted by @Youssef1313 in https://github.com/dotnet/roslyn-analyzers/pull/5210#r665961235_

Youssef1313:
Fixes #5219

FYI @RussKie

### DIFF
--- a/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
+++ b/src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
@@ -412,15 +412,9 @@ $@"<Project>
                 }
             }
 
-            string escapeMarkdown(string text)
-            {
-                // Escape generic arguments to ensure they are not considered as HTML elements
-                text = Regex.Replace(text, "(<.+?>)", "\\$1");
-
-                // Escape asterisks.
-                text = text.Replace("*", @"\*");
-                return text;
-            }
+            // Escape generic arguments to ensure they are not considered as HTML elements, and also escape asterisks.
+            static string escapeMarkdown(string text)
+                => Regex.Replace(text, "(<.+?>)", "\\$1").Replace("*", @"\*");
 
             // based on https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/CommandLine/ErrorLogger.cs
             void createAnalyzerSarifFile()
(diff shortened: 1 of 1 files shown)


src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -413,6 +412,16 @@ void createAnalyzerDocumentationFile()
                 }
             }
 
+            string escapeMarkdown(string text)
+            {
+                // Escape generic arguments to ensure they are not considered as HTML elements
+                text = Regex.Replace(text, "(<.+?>)", "\\$1");
+
+                // Escape asterisks.
+                text = text.Replace("*", @"\*");
```
Youssef1313, line 421: I kept it simple and escaped asterisks only.

RussKie (approved):
:shipit:

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -413,6 +412,16 @@ void createAnalyzerDocumentationFile()
                 }
             }
 
+            string escapeMarkdown(string text)
```
Evangelink, line 415: Could be made `static`.

Evangelink, line 415: Probably better to stick with pascal case.

Evangelink: Ah nevermind the other functions are with camel case...

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -413,6 +412,16 @@ void createAnalyzerDocumentationFile()
...
+            {
+                // Escape generic arguments to ensure they are not considered as HTML elements
+                text = Regex.Replace(text, "(<.+?>)", "\\$1");
```
Evangelink, line 418: It's a personal taste but you could rewrite the function just as a chain of function calls:

```csharp
 string escapeMarkdown(string text) =>
	return text
		// Escape generic arguments to ensure they are not considered as HTML elements
		.Replace("(<.+?>)", "\\$1")
		// Escape asterisks.
        .Replace("*", @"\*");
```

jmarolf (approved):

PR #5221, opened 2021-07-09
Use a MS.CA version containing source generator stuff in tests

Youssef1313:
@ryzngard I haven't confirmed, but I think that's why the tests were using 3.3 incorrectly.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/Microsoft.CodeAnalysis.Analyzers.UnitTests.csproj
+++ b/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/Microsoft.CodeAnalysis.Analyzers.UnitTests.csproj
@@ -4,6 +4,11 @@
     <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
     <ServerGarbageCollection>true</ServerGarbageCollection>
   </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersionForTests)" />
+  </ItemGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\..\Test.Utilities\Test.Utilities.csproj" />
     <ProjectReference Include="..\Core\Microsoft.CodeAnalysis.Analyzers.csproj" />
--- a/src/Test.Utilities/AdditionalMetadataReferences.cs
+++ b/src/Test.Utilities/AdditionalMetadataReferences.cs
@@ -89,7 +89,7 @@ namespace Test.Utilities
             referenceAssemblies = referenceAssemblies.AddAssemblies(ImmutableArray.Create("System.Xml.Data"));
 #endif
 
-            referenceAssemblies = referenceAssemblies.AddPackages(ImmutableArray.Create(new PackageIdentity("Microsoft.CodeAnalysis", "3.0.0")));
+            referenceAssemblies = referenceAssemblies.AddPackages(ImmutableArray.Create(new PackageIdentity("Microsoft.CodeAnalysis", "3.6.0")));
 
 #if NETCOREAPP
             referenceAssemblies = referenceAssemblies.AddPackages(ImmutableArray.Create(
--- a/src/Test.Utilities/Test.Utilities.csproj
+++ b/src/Test.Utilities/Test.Utilities.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.VisualBasic" Version="$(MicrosoftVisualBasicVersion)" />
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersionForTests)" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit" Version="$(MicrosoftCodeAnalysisTestingVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing.XUnit" Version="$(MicrosoftCodeAnalysisTestingVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.XUnit" Version="$(MicrosoftCodeAnalysisTestingVersion)" />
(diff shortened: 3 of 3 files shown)


ryzngard:
That would be great! I haven't looked yet, but it was odd that we were using such an old version. I crashed after the stream 😅 . I'll pull this and check if it fixes the issue I was having

Youssef1313:
I have moved the version from `Test.Utilities` to `MS.CA.Analyzers.UnitTests` due to NuGet errors (detected version downgrade). If this doesn't fix the issue, I'll get it back to `Test.Utilities` and try fix the NuGet errors.

ryzngard:
@Youssef1313 that should work. It should be based off whatever the test library is including, not the Utilities version. Works for me locally, but I'll do a full rebuild and test tomorrow just to make sure.

Youssef1313:
Just tried locally. It has issues :( :(

1. It only works when explicitly referencing `CodeAnalysisReference` via `AdditionalMetadataReferences`.
2. That does seem to work for `Generator` and types existing in newer version. But when referencing old types, it generates an error `CS0433: The type 'DiagnosticAnalyzer' exists in both 'Microsoft.CodeAnalysis, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' and 'Microsoft.CodeAnalysis, Version=3.9.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'`.

I'll try few more things to see if I can fix both issues, or at least the second.

Youssef1313:
@jmarolf Performance check is failing with:

```
Global total time: 00:34:01 (2041.4 sec), executed benchmarks: 36
// * Artifacts cleanup *
Done with performance run
Comparing performance results baseline: 'D:\a\1\s\artifacts\performance\perfResults\baseline\results' 
 - baseline: 'D:\a\1\s\artifacts\performance\perfResults\baseline\results' 
 - results: 'D:\a\1\s\artifacts\performance\perfResults\perfTest\results' 
  ' worse: 5, geomean: 1.244%
  ' test: 'VisualBasicPerformanceTests.Enabled.VisualBasic_CA2247.CA2247_DiagnosticsProduced' took '1.275' times longer
  ' test: 'VisualBasicPerformanceTests.Enabled.VisualBasic_CA2247.CA2247_Baseline' took '1.207' times longer
  ' test: 'VisualBasicPerformanceTests.Enabled.VisualBasic_IL3000.IL3000_DiagnosticsProduced' took '1.270' times longer
  ' test: 'VisualBasicPerformanceTests.Enabled.VisualBasic_IL3000.IL3000_Baseline' took '1.259' times longer
  ' test: 'VisualBasicPerformanceTests.Enabled.VisualBasic_IL3001.IL3001_DiagnosticsProduced' took '1.212' times longer
  We detected a regression in BMDN and there is no ETL info.

```

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -89,7 +89,7 @@ private static ReferenceAssemblies CreateDefaultReferenceAssemblies()
             referenceAssemblies = referenceAssemblies.AddAssemblies(ImmutableArray.Create("System.Xml.Data"));
 #endif
 
-            referenceAssemblies = referenceAssemblies.AddPackages(ImmutableArray.Create(new PackageIdentity("Microsoft.CodeAnalysis", "3.0.0")));
+            referenceAssemblies = referenceAssemblies.AddPackages(ImmutableArray.Create(new PackageIdentity("Microsoft.CodeAnalysis", "3.6.0")));
```
Youssef1313, line 92: Note: I tried using 3.8.0, but I got version conflicts with System.Collections.Immutable, so I went with the lowest version containing source generator stuff.

ryzngard:
@sharwell might have a better idea of what it takes to update the roslyn packages for tests.

jmarolf:
> @jmarolf Performance check is failing with:

We may need to re-baseline if the newer test framework has regressions. I would have thought that you just needed to reference.

I am a bit confused by this PR. 3.8 is the minimum version we should reference in our tests for things to work with source generators. There were several API changes between 3.6 and 3.8. I would expect referencing [Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.XUnit](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.XUnit/) would do the right thing.

ryzngard:
I don't think this is needed. Tests can reference specific versions from nuget

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/Microsoft.CodeAnalysis.Analyzers.UnitTests.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
     <ServerGarbageCollection>true</ServerGarbageCollection>
   </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersionForTests)" />
```
sharwell, line 9: 😕 Why is this changing?

src/Test.Utilities/AdditionalMetadataReferences.cs
sharwell: 💡 It might be better to just have a separate `ReferenceAssemblies` instance for source generator testing.

ryzngard: This seems the best path forward imo. When I was originally hitting the issue I didn't realize you could just specify the version and it would retrieve it.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/Microsoft.CodeAnalysis.Analyzers.UnitTests.csproj
Youssef1313: @sharwell This is done in multiple unit test projects. The effect AFAIK is using a recent compiler version for compiling the code. It's unnecessary currently, but I think it's desired. Let me know if you want to revert that.

src/Test.Utilities/AdditionalMetadataReferences.cs
Youssef1313: I see that @ryzngard already doing that in https://github.com/dotnet/roslyn-analyzers/pull/5235. Going to close the PR.

Youssef1313:
Closed per https://github.com/dotnet/roslyn-analyzers/pull/5221#discussion_r670200868.

PR #5222, opened 2021-07-09
Add static to partial definitions

### LINKED ISSUE #5171: CA1822 (Mark members as static) codefix doesn't add static to partial definitions

Youssef1313:
Fixes #5171


Youssef1313:
Looks like I'm messing up with things here 😄 
Will see if I can get back to it later.

Issue #5050, opened 2021-04-21
New analyzer: Missing GeneratorAttribute


Youssef1313:
Similar to https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingDiagnosticAnalyzerAttributeRuleTests.cs, but for source generators, which require [GeneratorAttribute](https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/SourceGeneration/GeneratorAttribute.cs).

mavasani:
Tagging @chsienki

Issue #192, opened 2015-07-24
Remove duplicated types


tmeschter:
The AsyncPackage has a number of types that derive from `CodeAction`. These are identical, except for minor differences in the implementation of the `EquivalenceKey` property.

These types should be replaced with a single type that takes the value of the `EquivalenceKey` as a constructor parameter.

sharwell:
Is there any reason to not just use `CodeAction.Create`?

srivatsn:
The reason we create separate types here is because we use the type name for telemetry purposes and CodeAction.Create would just have the same type name for different fixers. Using type name for telemetry is in itself quite odd and we need a different mechanism for that. We need to create an issue on dotnet\roslyn for such a mechanism and once we have that get rid of these types.

srivatsn:
Blocked by https://github.com/dotnet/roslyn/issues/4919

mavasani:
Stale issue...

PR #5235, opened 2021-07-12
Add missing [Generator] attribute analyzer and code fix

ryzngard:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5050

Adds an analyzer and code fix to ensure that types that implement `ISourceGenerator` also have the `[Generator]` attribute. Works in both C# and Visual Basic.


src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,13 @@
...
   </data>
+  <data name="MissingSourceGeneratorAttributeDescription" xml:space="preserve">
+    <value>'{0}' inherits from ISourceGenerator but does not have the [Generator] attribute</value>
+  </data>
+  <data name="MissingSourceGeneratorAttributeMessage" xml:space="preserve">
+    <value>Missing [Generator] Attribute</value>
+  </data>
+  <data name="MissingSourceGeneratorAttributeTitle" xml:space="preserve">
+    <value>Missing [Generator] Attribute</value>
+  </data>
```
Youssef1313, line 540: nit: since those applies to both languages, I wouldn't use the language-specific syntax in them. Probably just say `GeneratorAttribute` instead of `[Generator] attribute`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                return;
+            }
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                context.RegisterCodeFix(new MyCodeAction(
+                    "Fix stuff",
+                    (c) => FixDocumentAsync(document, node, c),
+                    "Fix stuff"), diagnostic);
```
Youssef1313, line 37: Localize and update codefix title. Also the parentheses around `c` looks redundant.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        public override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        private static async Task<Document> FixDocumentAsync(Document document, SyntaxNode node, CancellationToken cancellationToken)
+        {
+            var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            var generator = editor.Generator;
+            var generatorAttribute = generator.Attribute(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute);
```
Youssef1313, line 47: Note: The parameterless attribute constructor makes the generator **C#-only**. I'd instead make the codefix add both languages, and let the user delete an extra language if needed.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+            isEnabledByDefault: true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(DiagnosticRule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+
+            context.RegisterCompilationStartAction(compilationContext =>
```
Youssef1313, line 34: Rename `compilationContext` to `context`. This ensures that we're always using the inner most `context`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            if (!symbol.AllInterfaces.Any(i => sourceGenerator.Equals(i, SymbolEqualityComparer.Default)))
+            {
+                return;
+            }
+
+            if (symbol.GetApplicableAttributes(null).Any(a => a.AttributeClass.Inherits(sourceGeneratorAttribute)))
+            {
+                return;
+            }
```
Youssef1313, line 64: Is there any test hitting this condition?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
```diff
@@ -0,0 +1,406 @@
...
+    {
+    }
+}";
+
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources = { code , SourceGeneratorStub_CSharp },
+                    AdditionalReferences = { AdditionalMetadataReferences.CodeAnalysisReference}
```
Youssef1313, line 89: I doubt this is needed since you're using those stubs.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+public class CustomGenerator : CustomGeneratorBase
+{
+    public override void Initialize(GeneratorInitializationContext context) {}
+
+    public override void Execute(GeneratorExecutionContext context)
+    {
+    }
+}";
```
Youssef1313, line 359: Can you add another test with `CustomGenerator` being in a new document, and the overrides are not there (just the class)?

This will have compile errors, but what I'm intending to test here is to trigger the analyzer and codefix without `using Microsoft.CodeAnalysis;` being there, to see if the generator will be added as fully-qualified name or the using will get added by the codefix.

src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -44,6 +44,8 @@ internal static class WellKnownTypeNames
         public const string MicrosoftCodeAnalysisDiagnosticsCompilationStartAnalysisContext = "Microsoft.CodeAnalysis.Diagnostics.CompilationStartAnalysisContext";
         public const string MicrosoftCodeAnalysisDiagnosticsDiagnosticAnalyzer = "Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer";
         public const string MicrosoftCodeAnalysisDiagnosticsDiagnosticAnalyzerAttribute = "Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzerAttribute";
+        public const string MicrosoftCodeAnalysisISourceGenerator = "Microsoft.CodeAnalysis.ISourceGenerator";
+        public const string MicrosoftCodeAnalysisGeneratorAttribute = "Microsoft.CodeAnalysis.GeneratorAttribute";
```
Youssef1313, line 48: The file was intended to be sorted alphabetically.

https://github.com/dotnet/roslyn-analyzers/blob/d1fbafacc364f3c32168478df0b060b6f541d4c8/src/Utilities/Compiler/WellKnownTypeNames.cs#L5

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,13 @@
   <data name="PreferIsKindFix" xml:space="preserve">
     <value>Use 'IsKind' instead of 'Kind'</value>
   </data>
+  <data name="MissingSourceGeneratorAttributeDescription" xml:space="preserve">
+    <value>'{0}' inherits from ISourceGenerator but does not have the [Generator] attribute</value>
```
Youssef1313, line 533: Enclose `ISourceGenerator` between single quotes, and possibly use "implements" instead of "inherits". Inheritance is used more with classes.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(DiagnosticRule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                var sourceGenerator = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisISourceGenerator);
```
Youssef1313, line 36: nit: rename to `iSourceGenerator`

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+                var sourceGeneratorAttribute = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute);
```
Youssef1313, line 37: nit: rename to `generatorAttribute`

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                if (sourceGenerator == null || sourceGeneratorAttribute == null)
+                    // We don't need to check assemblies unless they're referencing Microsoft.CodeAnalysis which defines DiagnosticAnalyzer.
+                    return;
+
+                compilationContext.RegisterSymbolAction(c => AnalyzeSymbol(c, sourceGenerator, sourceGeneratorAttribute), SymbolKind.NamedType);
+            });
+        }
+
+        private static void AnalyzeSymbol(SymbolAnalysisContext c, INamedTypeSymbol sourceGenerator, INamedTypeSymbol sourceGeneratorAttribute)
```
Youssef1313, line 47: If you decided to take the renaming suggestion, reflect it here too.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
sharwell: 😬eww I prefer current form

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
```diff
@@ -0,0 +1,406 @@
...
+        Sub Initialize(context As GeneratorInitializationContext)
+        Sub Execute(context As GeneratorExecutionContext)
+    End Interface
+
+    public Class GeneratorInitializationContext
+    End Class
+
+    public Class GeneratorExecutionContext
+    End Class
+End Namespace";
```
sharwell, line 65: ❗ Do not attempt to reinvent a public API in tests. Rather, use the `ReferenceAssemblies` property to add a reference to  version of the library which contains these definitions (in this case, Microsoft.CodeAnalysis.Common 3.10.0). This ensures the analyzer works as expected in production scenarios (we've had bugs in the test definition manifest in broken production features with no test failures).

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
sharwell: 📝 As part of https://github.com/dotnet/roslyn-analyzers/pull/5235/files#r668021528, you'll want to remove `AdditionalReferences` anyway. `ReferenceAssemblies` is preferred for all cases where the definition comes from another pipeline (e.g. tests in dotnet/roslyn can use `AdditionalReferences`, but tests in dotnet/roslyn-analyzers would use `ReferenceAssemblies`).

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
```diff
@@ -0,0 +1,406 @@
...
+                TestState =
+                {
+                    Sources = { code , SourceGeneratorStub_CSharp },
+                    AdditionalReferences = { AdditionalMetadataReferences.CodeAnalysisReference}
+                }
+            }.RunAsync();
+        }
+
+        [Fact]
+        public async Task TestMissing_VisualBasic()
```
sharwell, line 95: 📝 This test isn't really necessary. It's already tested in exactly the same form by `TestSimpleClass_VisualBasic` (the test library tests diagnostics reported in both the `TestState` and the `FixedState`, and the latter happens to match this test.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
sharwell: > to see if the generator will be added as fully-qualified name or the using will get added by the codefix.

I came looking for this, but by the time I reached this reply I realized that if a user fully-qualified `ISourceGenerator` in code, they probably want `GeneratorAttribute` to be fully-qualified as well.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: Probably worth to add single quotes also around `[Generator]`.

Evangelink, line 533: Looks like the resource analyzer is broken, nothing was raised to add the punctuation to the description.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: Or if you keep the current form, please ensure `attribute` is lower-cased.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            editor.ReplaceNode(node, generator.AddAttributes(node, generatorAttribute));
+
+            return editor.GetChangedDocument();
+        }
+
+        private class MyCodeAction : DocumentChangeAction
+        {
+            public MyCodeAction(string title, Func<CancellationToken, Task<Document>> createChangedDocument, string equivalenceKey) : base(title, createChangedDocument, equivalenceKey)
```
Evangelink, line 56: nit: split line before the base definition

```suggestion
            public MyCodeAction(string title, Func<CancellationToken, Task<Document>> createChangedDocument, string equivalenceKey)
                : base(title, createChangedDocument, equivalenceKey)
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: Nevermind, the issue is raised.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                var sourceGenerator = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisISourceGenerator);
+                var sourceGeneratorAttribute = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute);
+
+                if (sourceGenerator == null || sourceGeneratorAttribute == null)
```
Evangelink, line 39: [minor] You could use the `TryGet` function and bail-out in case of `false`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+        private static void AnalyzeSymbol(SymbolAnalysisContext c, INamedTypeSymbol sourceGenerator, INamedTypeSymbol sourceGeneratorAttribute)
+        {
+            var symbol = (INamedTypeSymbol)c.Symbol;
+
+            if (symbol.IsAbstract || symbol.IsAnonymousType)
+            {
+                return;
+            }
+
+            if (!symbol.AllInterfaces.Any(i => sourceGenerator.Equals(i, SymbolEqualityComparer.Default)))
```
Evangelink, line 56: I don't recall the name but I think there is an helper function for this that reads better.

Evangelink, line 56: Any reason not to merge the various ifs together?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: The generator attribute is an inherited attribute, so still technically correct afaik if it's in an ancestor class. If we really wanted to enforce otherwise we could make that change at the attribute definition. 

Correct that `TestHierarchy_InheritedAttribute` tests this code path.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
ryzngard: Good to know. Removed

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
ryzngard: Agree on adding test, but I think leaving fully qualified is ideal. Sam did a good job of capturing my initial reasoning about not adding an imports annotation.

src/Utilities/Compiler/WellKnownTypeNames.cs
ryzngard: Next analyzer... `WellKnownTypesIsSortedAnalyzer`

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                return;
+            }
+
+            if (symbol.GetApplicableAttributes(null).Any(a => a.AttributeClass.Inherits(sourceGeneratorAttribute)))
```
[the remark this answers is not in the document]
ryzngard: Correct. Removed

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
ryzngard: What's the correct way to reference a specific min version of an assembly? I thought we were restricted by what assemblies the tests reference, which is currently 3.3 and doesn't have ISourceGenerator etc.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
ryzngard: Good call. I'll remove this and csharp

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: Removed this code action altogether and just used `CodeAction.Create`

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: Personal taste more than anything. They're testing different logical points of the analyzer, so it's easier to read.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
sharwell: `ReferenceAssemblies.Default.AddPackages`

ryzngard: That's awesome, I didn't realize it would work with NuGet packages!!! Fixing

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: I'm not actually sure if that's enforced. I didn't find anywhere it was referenced in roslyn. @chsienki might know more. 

I don't think we want to add both, I doubt that's correct in most cases. If anything, maybe use the document language

ryzngard: Ah nm, it's used indirectly https://sourceroslyn.io/#Microsoft.CodeAnalysis/DiagnosticAnalyzer/AnalyzerFileReference.cs,60

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: Updated wording. Going to resolve this, but feel free to open more comments on newer

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: I like as is :)

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: Fixed, or at least ready to be flagged by the analyzer again :)

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
Youssef1313: @sharwell The case I'm referring to involves a document that doesn't contain `ISourceGenerator`, so the attribute will be the first thing from `Microsoft.CodeAnalysis` namespace.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (node is null)
+            {
+                return;
+            }
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
```
Youssef1313, line 35: I'm not sure if this comment (https://github.com/dotnet/roslyn-analyzers/issues/192#issuecomment-136806431) about `CodeAction.Create` still applies.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,56 @@
...
+                        CodeAnalysisDiagnosticsResources.AddGeneratorAttribute,
+                        c => FixDocumentAsync(document, node, c),
+                        equivalenceKey: nameof(SourceGeneratorAttributeAnalyzerFix)),
```
Youssef1313, line 38: I think it doesn't matter, but this should be more consistent with other usages of equivalence key.

```suggestion
                        equivalenceKey: nameof(CodeAnalysisDiagnosticsResources.AddGeneratorAttribute)),
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
Youssef1313: @ryzngard Using the document language isn't ideal IMO.

There are few options I can come up with:

- Offer 3 codefixes (C# - VB - Both)
- Heuristic 1:
    - Look at the namespaces referenced in `using` statements if they are contain C#-specific, VB-specific, or both. In case there are none, I'd fall back to "both" and let the developer adjust the attribute.
- Heuristic 2:
    - Look at the type name if it contains "CSharp" or "VB" (or "VisualBasic"), and fall back to both languages
- Heuristic 3:
    - Combines both above.

Simplest is probably the first option to offer 3 codefixes.

chsienki: I think for now it will be fine to leave it with the default value. Ultimately a C# generator targeting a C# only project is the most likely scenario. We could put the lang parameter in and default it to C# as a hint that you can change it if people feel strongly (or as @ryzngard said, use the document language), but this feels like we're probably just over engineering something that we can address down the line if we get customer feedback about it.

ryzngard: While true, I think that's overengineering for the problem space right now. We also don't catch cases where the user puts the wrong language with the analyzer... we just don't know. During compilation I'm sure the generator will fail if the attribute is configured incorrectly. Hopefully whatever error happens is enough to help the author figure out next steps. Having the code fix as is does exactly what the analyzer catches, no more and no less. Having some hard to follow heuristics is unwarranted until we have a stronger user opinion on the matter.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: hehe, I pulled this from another fixer :P if this is the normal key that works for me

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: hmm, if that's the case there's at least one other place that needs to be fixed. Let me look into how telemetry id gets calculated

ryzngard: So... yes and no (ish) 

https://github.com/dotnet/roslyn/blob/main/src/EditorFeatures/Core.Wpf/Suggestions/SuggestedActions/SuggestedAction.cs#L192 

I think that's the function that's getting telemetry for suggested code fixes. It has `TryGetTelemetryId(out var telemetryId)` which does have the problem described in that comment. It only uses the type as an indicator for telemetry. 

If the diagnostic id is also being saved (https://github.com/dotnet/roslyn/blob/main/src/EditorFeatures/Core.Wpf/Suggestions/SuggestedActions/SuggestedAction.cs#L208) then we have enough information to distinguish this code fix. The downside would be if we added more code fixes we might lose which one is being applied. 

In short, I think adding a class is best... even if I don't like it :(

sharwell: Use `CodeAction.Create` to match the rest of this package.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
sharwell: We should use the approach from `ApplyDiagnosticAnalyzerAttributeFix` for consistency.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,16 @@
...
+    <value>'{0}' implements 'ISourceGenerator' but does not have the 'Generator' attribute.</value>
+  </data>
+  <data name="MissingSourceGeneratorAttributeMessage" xml:space="preserve">
+    <value>Missing 'Generator' attribute</value>
+  </data>
+  <data name="MissingSourceGeneratorAttributeTitle" xml:space="preserve">
+    <value>Missing 'Generator' Attribute</value>
+  </data>
+  <data name="AddGeneratorAttribute" xml:space="preserve">
+    <value>Add 'Generator' Attribute</value>
```
Youssef1313, line 542: Consider making attribute lowercase.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,62 @@
...
+            return editor.GetChangedDocument();
+        }
+
+        private class MyCodeAction : DocumentChangeAction
+        {
+            public MyCodeAction(Func<CancellationToken, Task<Document>> func)
+                : base(CodeAnalysisDiagnosticsResources.AddGeneratorAttribute, func, nameof(CodeAnalysisDiagnosticsResources.AddGeneratorAttribute))
+            {
+            }
+        }
```
sharwell, line 60: 📝 This package uses `CodeAction.Create` for simplicity and for matching the public APIs used by the majority of analyzer authors.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,62 @@
...
+    public sealed class SourceGeneratorAttributeAnalyzerFix : CodeFixProvider
+    {
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(
+            DiagnosticIds.MissingSourceGeneratorAttributeId);
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var document = context.Document;
+            var root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var node = root.FindNode(context.Span, getInnermostNodeForTie: true);
```
Youssef1313, line 26: In case `getInnermostNodeForTie: true` is necessary, is there a test that can fail if it's false?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
...
+            {
+                if (context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisISourceGenerator, out var sourceGenerator) &&
+                    context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute, out var generatorAttribute))
+                {
+                    context.RegisterSymbolAction(c => AnalyzeSymbol(c, sourceGenerator, generatorAttribute), SymbolKind.NamedType);
+                }
+            });
+        }
+
+        private static void AnalyzeSymbol(SymbolAnalysisContext c, INamedTypeSymbol sourceGenerator, INamedTypeSymbol generatorAttribute)
```
sharwell, line 44: ❓ Can we adopt the same logic from `DiagnosticAnalyzerAttributeAnalyzer.AttributeAnalyzer`?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/MissingGeneratorAttributeRuleTests.cs
```diff
@@ -0,0 +1,352 @@
...
+                TestState =
+                {
+                    Sources = { code },
+                    ReferenceAssemblies = ReferenceAssemblies
+                },
+                FixedState =
+                {
+                    Sources = { fixedCode },
+                    ReferenceAssemblies = ReferenceAssemblies
+                },
```
sharwell, line 54: 💡 `FixedState` inherits unspecified items from `TestState`. Also, this form is typically used for cases where a code fix is adding/removing references. For these tests, you can set `Test.ReferenceAssemblies`:

```suggestion
                ReferenceAssemblies = ReferenceAssemblies,
                TestState =
                {
                    Sources = { code },
                },
                FixedState =
                {
                    Sources = { fixedCode },
                },
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
Youssef1313: Would that work?

```suggestion
            if (!symbol.AllInterfaces.Contains(sourceGenerator))
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: Sure, changing to `Apply Generator attribute for '{0}'.`

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: There is no test for this currently, I'll just remove this part. I don't think it's necessary

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: There's some things that I think are different in this case. Can you explain the benefit in the complexity that the `DiagnosticAnalyzerAttributeAnalyzer.AttributeAnalyzer` has? It looks like it has logic to determine what languages are supported based on referenced assemblies.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,21 @@
...
+  </data>
+  <data name="MissingSourceGeneratorAttributeMessage" xml:space="preserve">
+    <value>Missing 'Generator' attribute</value>
+  </data>
+  <data name="MissingSourceGeneratorAttributeTitle" xml:space="preserve">
+    <value>Missing 'Generator' Attribute</value>
+  </data>
+  <data name="AddGeneratorAttribute_1" xml:space="preserve">
+    <value>Apply Generator attribute for '{0}'.</value>
+    <comment>"Generator" is a named type and should not be changed</comment>
```
Youssef1313, line 543: There is a special syntax for locking localization. e.g:

https://github.com/dotnet/roslyn-analyzers/blob/6f7420ab11fb2fb639fa2242657885c8a614f97e/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx#L380

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers.Fixers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic, Name = nameof(SourceGeneratorAttributeAnalyzerFix))]
+    [Shared]
+    public sealed class SourceGeneratorAttributeAnalyzerFix : CodeFixProvider
+    {
+        public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(
```
Youssef1313, line 20: This creates a new array every time the property is accessed. Consider assigning it only one time.

```suggestion
        public override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            if (node is null)
+            {
+                return;
+            }
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                AddFix(
+                    string.Format(CultureInfo.CurrentCulture, CodeAnalysisDiagnosticsResources.AddGeneratorAttribute_1, LanguageNames.CSharp),
```
Youssef1313, line 37: I recall an issue in dotnet/roslyn where `CurrentCulture` has caused language-divergence compared to other UI elements. Unless I'm missing something, this should be `CurrentUICulture`

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: Interesting. I've always just left a comment like this and translators worked it out. Wonder what benefit `{Locked="Shared"}` has

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,21 @@
...
+  </data>
+  <data name="AddGeneratorAttribute_2" xml:space="preserve">
+    <value>Apply Generator attribute for both '{0}' and '{1}'.</value>
```
Youssef1313, line 546: The placeholders here are always going to be the same. Do we need them?

ryzngard: Technically no, but this follows the previous pattern so I kept it. Seems easier to prevent typos in language names too

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: Likely true. I'm actually surprised we have a warning for this and no fix to get it right... :/ 

Has this analyzer been turned off? I remember when I changed to string.Format I was getting a warning about including a culture. Wonder if that was a bug.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: Sure, but this is such a miniscule win for less readability imo :P

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,109 @@
...
+                }
+
+                var arguments = new SyntaxNode[languageNames.Length];
+                for (var i = 0; i < languageNames.Length; i++)
+                {
+                    RoslynDebug.Assert(languageNames[i] == LanguageNames.CSharp || languageNames[i] == LanguageNames.VisualBasic);
+
+                    var language = languageNames[i] == LanguageNames.CSharp
+                        ? nameof(LanguageNames.CSharp)
+                        : nameof(LanguageNames.VisualBasic);
```
Youssef1313, line 95: I don't get how `language` is going to be different than `languageNames[i]`

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class SourceGeneratorAttributeAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.MissingSourceGeneratorAttributeTitle), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.MissingSourceGeneratorAttributeTitle), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.MissingSourceGeneratorAttributeDescription), CodeAnalysisDiagnosticsResources.ResourceManager, typeof(CodeAnalysisDiagnosticsResources));
```
Youssef1313, line 16: The use of those private fields was eliminated in #5406

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: It definitely is not. This was prior to me adding the assert above, which is just a nice to have to formally state what languages are expected in this fix. Changing

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Youssef1313: @ryzngard I have no idea of the internal localization process 😄

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: I see. Thanks for the heads up :). Nice little win there

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
Youssef1313: I don't know about that analyzer 😄

ryzngard: *shrug* it's gone now, but `CurrentUICulture` is definitely what we want here.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: hmm... looks like @DustinCampbell  missed the MetadataAnalyzers for adding this? They're not currently ported.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: I'm going to leave now and follow up tomorrow for what's going on. Likely you're right, but I want to know what the difference is.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -90,9 +90,7 @@ private static async Task<Document> FixDocumentAsync(Document document, SyntaxNo
                 {
                     RoslynDebug.Assert(languageNames[i] == LanguageNames.CSharp || languageNames[i] == LanguageNames.VisualBasic);
 
-                    var language = languageNames[i] == LanguageNames.CSharp
-                        ? nameof(LanguageNames.CSharp)
-                        : nameof(LanguageNames.VisualBasic);
+                    var language = languageNames[i];
```
Youssef1313: Sorry I think I was incorrect here. The constant value is "C#", it doesn't match the field constant name ("CSharp").

ryzngard: You know, I trusted you Youssef. 😑

I'll get this tomorrow as well when it's not bedtime 😉

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
Youssef1313: @ryzngard They're not missed as far as I can tell:

https://github.com/dotnet/roslyn-analyzers/blob/6f7420ab11fb2fb639fa2242657885c8a614f97e/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticAnalyzerAttributeAnalyzer.cs#L16-L24

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
Youssef1313: Sorry again.

Actually, even the UICulture suggestion is failing CI (I don't quite understand why). But the failure is for the good this time. If it's wrong to use UICulture for `string.Format`, then this should get updated and the analyzer should be enabled in roslyn:

https://github.com/dotnet/roslyn/blob/b88948f10a1085980df7080f0b765b1a8c614436/src/Compilers/CSharp/Portable/Compiler/DocumentationCommentCompiler.cs#L289

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
ryzngard: Yea, I don't know what my tired brain was doing last night that couldn't get this to resolve. Got it now

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: I was just jesting, it's easy to make that mistake. I often forget that "C#" is what the string is. 

As for UICulture, I'll look into that. I would expect it to be correct, but it's possible not formatting for culture at all is more correct in these cases.

ryzngard: > src\Microsoft.CodeAnalysis.Analyzers\Core\MetaAnalyzers\Fixers\SourceGeneratorAttributeAnalyzerFix.cs(38,21): error CA1305: (NETCORE_ENGINEERING_TELEMETRY=Build) The behavior of 'string.Format(string, object)' could vary based on the current user's locale settings. Replace this call in 'SourceGeneratorAttributeAnalyzerFix.RegisterCodeFixesAsync(CodeFixContext)' with a call to 'string.Format(IFormatProvider, string, params object[])'

Here it is... This seems to be disabled during build in VS though. @sharwell is the correct thing to do is pass `CurrentCulture` here?

ryzngard:
Going through my old PRs. Apologies for dropping this on the floor :( PTAL if you get a moment.

src/Microsoft.CodeAnalysis.Analyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,8 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+RS1034 | MicrosoftCodeAnalysisPerformance | Warning | PreferIsKindAnalyzer
```
Youssef1313, line 7: This shouldn't be here I think, but I'd like to wait to see if the CI is going to fail before removing this line. It should fail IMO.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            var codeAction = CodeAction.Create(
+                title,
+                (cancellationToken) => FixDocumentAsync(document, node, languageNames, cancellationToken),
+                equivalenceKey: nameof(SourceGeneratorAttributeAnalyzerFix));
+
+            context.RegisterCodeFix(codeAction, diagnostic);
+        }
+
+        public override FixAllProvider? GetFixAllProvider() => null;
```
Youssef1313, line 61: First time I see a null value returned from `GetFixAllProvider`, I'm curious to know what does it mean and why we're not returning `WellKnownFixAllProviders.BatchFixer`

src/Microsoft.CodeAnalysis.Analyzers/Core/AnalyzerReleases.Unshipped.md
Youssef1313: CI is indeed reporting this:

> src/Microsoft.CodeAnalysis.Analyzers/Core/AnalyzerReleases.Unshipped.md(7,1): error RS2006: (NETCORE_ENGINEERING_TELEMETRY=Build) Rule 'RS1034' has duplicate entry between release 'unshipped' and release '3.3.3'

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,21 @@
   <data name="PreferIsKindFix" xml:space="preserve">
     <value>Use 'IsKind' instead of 'Kind'</value>
   </data>
+  <data name="MissingSourceGeneratorAttributeDescription" xml:space="preserve">
+    <value>'{0}' implements 'ISourceGenerator' but does not have the 'Generator' attribute.</value>
```
Evangelink, line 533: It's been some time but I think @Youssef1313 told me we should not be using placeholders in descriptions.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Evangelink: I have seen the same pattern described by @Youssef1313 on this repo.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -529,4 +529,21 @@
...
+  </data>
+  <data name="MissingSourceGeneratorAttributeMessage" xml:space="preserve">
+    <value>Missing 'Generator' attribute</value>
+  </data>
+  <data name="MissingSourceGeneratorAttributeTitle" xml:space="preserve">
+    <value>Missing 'Generator' Attribute</value>
+  </data>
+  <data name="AddGeneratorAttribute_1" xml:space="preserve">
```
Evangelink, line 541: We are not really consistent for naming of the code fix titles but I would avoid `_X`.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/SourceGeneratorAttributeAnalyzer.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class SourceGeneratorAttributeAnalyzer : DiagnosticAnalyzer
+    {
+        public static readonly DiagnosticDescriptor DiagnosticRule = new(
+            DiagnosticIds.MissingSourceGeneratorAttributeId,
+            CodeAnalysisDiagnosticsResources.CreateLocalizableResourceString(nameof(CodeAnalysisDiagnosticsResources.MissingSourceGeneratorAttributeTitle)),
```
Evangelink, line 16: Pattern is usually to include `using static CodeAnalysisDiagnosticsResources;` inside the namespace.

src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md
```diff
@@ -424,6 +424,18 @@ Prefer 'syntax.IsKind(kind)' to 'syntax.Kind() == kind' when checking syntax kin
 |CodeFix|True|
 ---
 
+## RS1035: Missing 'Generator' Attribute
+
+'{0}' implements 'ISourceGenerator' but does not have the 'Generator' attribute.
```
Evangelink, line 429: Should be fixed when the description will be updated.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Youssef1313: @Evangelink Yes this is correct. Good catch.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,111 @@
...
+            var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            var generator = editor.Generator;
+
+            SyntaxNode? generatorAttribute;
+
+            if (languageNames.Length == 1 && languageNames[0] == LanguageNames.CSharp)
+            {
+                // CSharp is the only language, which is the default paramterless 
+                // constructor for the Generator attribute
+                generatorAttribute = generator.Attribute(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute);
```
Youssef1313, line 74: See https://github.com/dotnet/roslyn-analyzers/issues/5797:

```suggestion
                generatorAttribute = generator.Attribute(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute).WithAddImportsAnnotation();
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,111 @@
...
+                        LanguageNames.CSharp => nameof(LanguageNames.CSharp),
+                        LanguageNames.VisualBasic => nameof(LanguageNames.VisualBasic),
+                        _ => throw new InvalidOperationException()
+                    };
+
+                    var finalExpression = generator.MemberAccessExpression(baseLanguageNameExpression, language);
+                    arguments[i] = finalExpression;
+                }
+
+                generatorAttribute = generator.Attribute(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute, arguments);
```
Youssef1313, line 103:
```suggestion
                generatorAttribute = generator.Attribute(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute, arguments).WithAddImportsAnnotation();
```

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: Is the preference to use "A type implements 'ISourceGenerator'"... ?

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: I'll switch to locked here

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
Youssef1313: @ryzngard Yes I think the wording looks good.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
ryzngard: Let me add it and some tests for fix all

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
ryzngard: What is the way to distinguish between this and `AddGeneratorAttribute_2`?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/SourceGeneratorAttributeAnalyzerFix.cs
```diff
@@ -0,0 +1,114 @@
...
+            var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            var generator = editor.Generator;
+
+            SyntaxNode? generatorAttribute;
+
+            if (languageNames.Length == 1 && languageNames[0] == LanguageNames.CSharp)
+            {
+                // CSharp is the only language, which is the default paramterless 
+                // constructor for the Generator attribute
+                generatorAttribute = generator.Attribute(WellKnownTypeNames.MicrosoftCodeAnalysisGeneratorAttribute).WithAddImportsAnnotation();
```
Youssef1313, line 77: Looks like imports annotation isn't working as I was expecting (`TestSimpleClass_FullyQualified_CSharp` should have failed).

I think Roslyn can improve this by assuming the annotation is meant to any descendant nodes, but for now we can generate the type syntax ourselves and pass it to `generator.Attribute` similar to


https://github.com/dotnet/roslyn-analyzers/blob/c3145214555afda60a74da31f9096a788348e4fc/src/Roslyn.Diagnostics.Analyzers/Core/ImportingConstructorShouldBeObsoleteCodeFixProvider.cs#L86-L91

https://github.com/dotnet/roslyn-analyzers/blob/c3145214555afda60a74da31f9096a788348e4fc/src/Roslyn.Diagnostics.Analyzers/Core/ImportingConstructorShouldBeObsoleteCodeFixProvider.cs#L103-L104

Youssef1313 (commented):
I think we need to extend this to `IIncrementalGenerator`

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
sharwell: Maybe `_OneLanguage` and `_TwoLanguages`? Otherwise you could use the naming style from Roslyn here.

ryzngard:
Closing old PR. This could be used for anyone moving forward to implement this

Issue #3267, opened 2020-02-05
Report a diagnostic in case of invalid user option


Evangelink:
Currently, any invalid user option will be discarded without any notification to the end-user. It would be nice to have some diagnostic to let them know about the incorrect configuration(s).

> We may want to eventually report a diagnostic here so user can fix their configuration. It is fine to not do it here as it seems applicable for all our option values.

_Originally posted by @mavasani in https://github.com/dotnet/roslyn-analyzers/pull/3258_

Evangelink:
I am not sure which project and rule ID naming shall be used for this rule as it seems like a new rule "type".

mavasani:
We used CA1509 under `DiagnosticCategory.Maintainability` for invalid entry in Code metrics configuration file: https://github.com/dotnet/roslyn-analyzers/blob/acc3c12a1597f0e003c4e98a0067c117d50358d1/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsAnalyzer.cs#L24

I am wondering if we can continue using the same category and use `CA1510`? Does not feel the most appropriate category, but I'd like to avoid adding a new category just for this one rule. Note that we should not flag for unrecognized `key`, just unrecognized `value` for a key that we understand.

Evangelink:
I am not sure to understand if you mean `CA1510` for invalid value for the key `additional_required_generic_interfaces` or as a analyzer for all keys. If the former, what about other keys?

mavasani:
I'd say lets use a single diagnostic ID with customized message for invalid editorconfig values for all keys that we recognize.

PR #5240, opened 2021-07-13
Allow multiple input files for SymbolIsBannedAnalyzer

### LINKED ISSUE #5143: Support loading banned APIs from multiple additional files
### Analyzer

**Diagnostic ID**: [RS0030](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Microsoft.CodeAnalysis.BannedApiAnalyzers.md)

### Describe the improvement

Support loading banned API rules from multiple files.

Consider updating [ReadBannedApis](https://github.com/dotnet/roslyn-analyzers/blob/a0522afd52b25ad211298e308fd5bbc7734865e0/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs#L180) to load multiple txt files. For example, any additional files with names that match the format BannedSymbols*.txt or end with BannedSymbols.txt. 

### Additional context

Our team has a very large code base and we'd like to use this analyzer for multiple purposes but the set of banned APIs won't be the same across the whole code base. Instead of having a bunch of different BannedSymbols.txt with overlapping rules we have to maintain, it would be nicer to have one file per type/area of banned APIs and we can include them  to different areas of the code as needed.

I'm happy to code up this change if this is something that would be considered.

ltrzesniewski:
Fixes #5143.

`SymbolIsBannedAnalyzer` will now accept the following file patterns: `BannedSymbols.txt` and `BannedSymbols.*.txt`.

Note that a duplicate pattern in different files will still generate RS0031. I'm not sure if we should allow that or not, but that can go into a separate issue.

### DIFF
--- a/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
@@ -181,6 +181,7 @@ namespace Microsoft.CodeAnalysis.BannedApiAnalyzers
                     from additionalFile in compilationContext.Options.AdditionalFiles
                     let fileName = Path.GetFileName(additionalFile.Path)
                     where fileName != null && fileName.StartsWith("BannedSymbols.", StringComparison.Ordinal) && fileName.EndsWith(".txt", StringComparison.Ordinal)
+                    orderby additionalFile.Path // Additional files are sorted by DocumentId (which is a GUID), make the file order deterministic
                     let sourceText = additionalFile.GetText(compilationContext.CancellationToken)
                     where sourceText != null
                     from line in sourceText.Lines
--- a/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
@@ -18,8 +18,6 @@ namespace Microsoft.CodeAnalysis.BannedApiAnalyzers
 {
     internal static class SymbolIsBannedAnalyzer
     {
-        public const string BannedSymbolsFileName = "BannedSymbols.txt";
-
         public static readonly DiagnosticDescriptor SymbolIsBannedRule = new(
             id: DiagnosticIds.SymbolIsBannedRuleId,
             title: new LocalizableResourceString(nameof(BannedApiAnalyzerResources.SymbolIsBannedTitle), BannedApiAnalyzerResources.ResourceManager, typeof(BannedApiAnalyzerResources)),
@@ -181,7 +179,8 @@ namespace Microsoft.CodeAnalysis.BannedApiAnalyzers
             {
                 var query =
                     from additionalFile in compilationContext.Options.AdditionalFiles
-                    where StringComparer.Ordinal.Equals(Path.GetFileName(additionalFile.Path), SymbolIsBannedAnalyzer.BannedSymbolsFileName)
+                    let fileName = Path.GetFileName(additionalFile.Path)
+                    where fileName != null && fileName.StartsWith("BannedSymbols.", StringComparison.Ordinal) && fileName.EndsWith(".txt", StringComparison.Ordinal)
                     let sourceText = additionalFile.GetText(compilationContext.CancellationToken)
                     where sourceText != null
                     from line in sourceText.Lines
--- a/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
+++ b/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
@@ -126,6 +126,32 @@ End Class");
         }
 
         [Fact]
+        public async Task DiagnosticReportedForDuplicateBannedApiLinesInDifferentFiles()
+        {
+            var test = new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources = { @"" },
+                    AdditionalFiles =
+                    {
+                        ("BannedSymbols.txt", @"{|#0:T:System.Console|}") ,
+                        ("BannedSymbols.Other.txt", @"{|#1:T:System.Console|}")
+                    },
+                },
+                ExpectedDiagnostics =
+                {
+                    new DiagnosticResult(SymbolIsBannedAnalyzer.DuplicateBannedSymbolRule)
+                        .WithLocation(0)
+                        .WithLocation(1)
+                        .WithArguments("System.Console")
+                }
+            };
+
+            await test.RunAsync();
+        }
+
+        [Fact]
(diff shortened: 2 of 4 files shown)


src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
```diff
@@ -181,7 +179,8 @@ private void OnCompilationStart(CompilationStartAnalysisContext compilationConte
             {
                 var query =
                     from additionalFile in compilationContext.Options.AdditionalFiles
-                    where StringComparer.Ordinal.Equals(Path.GetFileName(additionalFile.Path), SymbolIsBannedAnalyzer.BannedSymbolsFileName)
+                    let fileName = Path.GetFileName(additionalFile.Path)
+                    where fileName != null && fileName.StartsWith("BannedSymbols.", StringComparison.Ordinal) && fileName.EndsWith(".txt", StringComparison.Ordinal)
```
Evangelink, line 183: Shall we still use consts instead of hardcoded strings?

Evangelink, line 183: Do we want to allow `BannedSymbols..txt`? (which is currently allowed with your implementation).

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
```diff
@@ -208,6 +210,46 @@ void M()
...
+
+            var test = new VerifyCS.Test
+            {
+                TestState =
+                {
+                    Sources = { source },
+                    AdditionalFiles =
+                    {
+                        ("BannedSymbols.txt", @"T:N.BannedA") ,
+                        ("BannedSymbols.Other.txt", @"T:N.BannedB"),
```
Evangelink, line 241: I think it would be worth to add a test where you have some symbols duplicated in the various files.

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/Core/SymbolIsBannedAnalyzer.cs
ltrzesniewski: I could introduce `FileNamePrefix`/`FileNameSuffix` consts, but these would be tied to the implementation (since they would include the periods) and would only be referenced once. I don't think that would be a good design.

ltrzesniewski: I'd say yes, since it matches the `BannedSymbols.*.txt` pattern. I don't see a good reason to special-case it.

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
ltrzesniewski: I noticed something strange while adding the test: the `AdditionalFiles` collection is sorted by `DocumentId`, which is essentially a `Guid`, which means the order of the additional files is random.

This _may_ be a test environment artifact, but I don't think so. It looks like a real issue, since random file order doesn't make the build reproducible. Also, I found an issue which mentions this: https://github.com/dotnet/roslyn/issues/8605

In the case of this analyzer, it randomly swaps the diagnostic location with the additional location. Note that this PR doesn't really introduce this issue, since you can already reproduce it if you provide several additional files named `BannedSymbols.txt`, but from different directories.

How would you handle that?

Evangelink: That's a really good remark. Pinging @mavasani here

In case of the banned symbols we should not end-up with contradictory configurations so a union/merge of all banned symbols would/should work.

ltrzesniewski: Yes, in any case the analyzer would take a merge of the rules, except that right now:
 - If the same symbol is present in multiple config files with a different description text, the RS0030 diagnostic (*Do not use banned APIs*) for this symbol will pick the description text randomly from one of the files.
 - The RS0031 diagnostic (*The list of banned symbols contains a duplicate*) will pick its location and additional location randomly.

I could add an `orderby additionalFile.Path` clause after the `where fileName != null ...` line in the analyzer to work around this and make it deterministic, I'm just not sure if I *should* since it looks like a Roslyn issue.

Evangelink: Thanks, I didn't considered these.

I cannot make a call here so let's wait the input from a MS folk.

> * If the same symbol is present in multiple config files with a different description text, the RS0030 diagnostic (_Do not use banned APIs_) for this symbol will pick the description text randomly from one of the files.

We could "merge" the description, pick first (current) assuming the ordering issue will get fixed and maybe add this case to the analyzer I need to implement for invalid configurations https://github.com/dotnet/roslyn-analyzers/issues/3267.

jmarolf: I would prefer that some diagnostic is given to the user indicating that they have duplicate entries.

ltrzesniewski: @jmarolf Ok, I added a test which validates that RS0031 is emitted in that case.

As stated previously, I had to add an `orderby` clause to make the behavior deterministic, and this was actually an existing issue since you already could provide multiple `BannedSymbols.txt` files from different directories.

jmarolf: >  and this was actually an existing issue since you already could provide multiple BannedSymbols.txt files from different directories.

True, thanks for cleaning this up.

jmarolf (approved):

src/Microsoft.CodeAnalysis.BannedApiAnalyzers/UnitTests/SymbolIsBannedAnalyzerTests.cs
RikkiGibson: I'm coming across this much later but... if the `AnalyzerOptions.AdditionalFiles` are coming in a nondeterministic order, that is a bug that we should file on the analyzer host/compiler.

ltrzesniewski: @RikkiGibson yes, indeed, it breaks determinism. I didn't open another issue since I considered it's already filled under #8605.

PR #5242, opened 2021-07-14
Buffer.BlockCopy Analyzer

mahdiva:
This PR adds the Analyzer for the `Buffer.BlockCopy` method, verifying the `count` argument.
More info: https://github.com/dotnet/runtime/issues/33765

Fixes https://github.com/dotnet/runtime/issues/33765

cc @pgovind


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=5242) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=5242) before we can accept your contribution.<br/><br/>:x: mahdiva [sign now](https://cla.dotnetfoundation.org/dotnet/roslyn-analyzers?pullRequest=5242)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/roslyn-analyzers?pullRequest=5242) it.</sub>

Youssef1313:
New analyzers shouldn't target main. I'm not sure if they should target `release/6.0.1xx` or `release/6.0.1xx-preview7`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                                                                                      isPortedFxCopRule: false,
+                                                                                      isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterCompilationStartAction(compilationContext =>
```
Youssef1313, line 40: Rename `compilationContext` to just `context`. It's intentional to match the name of the `AnalysisContext` parameter.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemByte, out INamedTypeSymbol? byteType))
+                {
+                    return;
+                }
+
+                if (!compilationContext.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemSByte, out INamedTypeSymbol? sByteType))
+                {
+                    return;
+                }
```
Youssef1313, line 80: Use `GetSpecialType` instead.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+
+                compilationContext.RegisterOperationAction(context =>
+                {
+                    var invocationOperation = (IInvocationOperation)context.Operation;
+                    if (invocationOperation.Arguments.IsEmpty)
+                    {
+                        return;
+                    }
```
Youssef1313, line 88: This looks redundant since we're checking the length below?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+
+                    if (!invocationOperation.TargetMethod.Equals(blockCopyMethod))
+                    {
+                        return;
+                    }
+
+                    IArgumentOperation sourceArgument = arguments[0];
+                    IArgumentOperation destinationArgument = arguments[2];
+                    IArgumentOperation lastArgument = arguments[4];
```
Youssef1313, line 103: Arguments order may not match parameter order.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="BufferBlockCopyLengthMessage" xml:space="preserve">
+    <value>Buffer.BlockCopy expects the number of bytes to be copied as the last parameter. Using .Length on an array returns the length of the array.</value>
```
Youssef1313, line 1523: Add single quotes around `Buffer.BlockCopy` and `.Length`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.BufferBlockCopyLengthMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.BufferBlockCopyDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                                      s_localizableTitle,
+                                                                                      s_localizableMessage,
+                                                                                      DiagnosticCategory.Usage,
```
pgovind, line 28: Just noticing this now. In https://github.com/dotnet/runtime/issues/33765 Immo has marked this as Reliability, so we should change it here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.BufferBlockCopyLengthMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.BufferBlockCopyDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                                      s_localizableTitle,
+                                                                                      s_localizableMessage,
+                                                                                      DiagnosticCategory.Reliability,
+                                                                                      RuleLevel.IdeSuggestion,
```
buyaa-n, line 29: Seems approved at warning level
```suggestion
                                                                                      RuleLevel.BuildWarning,
```

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA2017 | Reliability | Info | BufferBlockCopyLengthAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2017)
```
buyaa-n, line 5:
```suggestion
CA2017 | Reliability | Warning | BufferBlockCopyLengthAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2017)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1519,4 +1519,13 @@
     <value>and all other platforms</value>
     <comment>This call site is reachable on: 'windows' 10.0.2000 and later, and all other platforms</comment>
   </data>
+  <data name="BufferBlockCopyLengthMessage" xml:space="preserve">
+    <value>'Buffer.BlockCopy' expects the number of bytes to be copied for the 'count' argument. Using '.Length' on an array returns the length of the array.</value>
```
buyaa-n, line 1523: NIT: The message needs improvement, `Using '.Length' on an array returns the length of the array.` doesn't really explain/address the issue. Please change something like:
```suggestion
    <value>'Buffer.BlockCopy' expects the number of bytes to be copied for the 'count' argument. Using 'Array.Length' may not match the number of bytes needs to be copied.</value>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                IPropertySymbol arrayLengthProperty = arrayType
+                    .GetMembers("Length")
+                    .OfType<IPropertySymbol>()
+                    .FirstOrDefault();
+
+                if (arrayLengthProperty is null)
+                {
+                    return;
+                }
```
buyaa-n, line 70: For me loading Length property and this check is redundant, i would not imagine an array type exist without Length property, I would just check the property by the name

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                INamedTypeSymbol byteType = context.Compilation.GetSpecialType(SpecialType.System_Byte);
+                INamedTypeSymbol sByteType = context.Compilation.GetSpecialType(SpecialType.System_SByte);
+
+                if (byteType is null || sByteType is null)
+                {
+                    return;
+                }
```
buyaa-n, line 78: This is definitely redundant, we don't need to load these symbols, for type check you can just use the SpeicalType enums https://sourceroslyn.io/#Microsoft.CodeAnalysis/SpecialType.cs,78

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+                    ImmutableArray<IArgumentOperation> arguments = IOperationExtensions.GetArgumentsInParameterOrder(invocationOperation.Arguments);
+
+                    if (arguments.Length != 5)
+                    {
+                        return;
+                    }
+
+                    IArgumentOperation sourceArgument = arguments[0];
+                    IArgumentOperation destinationArgument = arguments[2];
+                    IArgumentOperation lastArgument = arguments[4];
```
buyaa-n, line 98: NIT: follow same pattern for naming
```suggestion
                    IArgumentOperation countArgument = arguments[4];
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                    bool CheckArrayLengthLocalReference(IArgumentOperation targetArgument, IPropertyReferenceOperation lengthPropertyArgument)
+                    {
+                        if (targetArgument.Value is IConversionOperation targetArgValue)
+                        {
+                            if (lengthPropertyArgument.Instance.GetReferencedMemberOrLocalOrParameter() == targetArgValue.Operand.GetReferencedMemberOrLocalOrParameter())
+                            {
+                                IArrayTypeSymbol lastArgumentArrayTypeSymbol = (IArrayTypeSymbol)lengthPropertyArgument.Instance.Type;
+                                if (!lastArgumentArrayTypeSymbol.ElementType.Equals(byteType) && !lastArgumentArrayTypeSymbol.ElementType.Equals(sByteType))
```
buyaa-n, line 107:
```suggestion
                                if (lastArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_Byte && lastArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_SByte)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+                            return;
+                        }
+
+                        SemanticModel semanticModel = lastArgument.SemanticModel;
+                        CancellationToken cancellationToken = context.CancellationToken;
+
+                        if (semanticModel is null)
+                        {
+                            return;
+                        }
```
buyaa-n, line 143: NIT: doesn't seem to be null for source

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                        ILocalSymbol localArgumentDeclaration = localReferenceOperation.Local;
+                        if (localArgumentDeclaration is null)
```
buyaa-n, line 146: I would not expect it would ever be null, does it @mavasani ?

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
```diff
@@ -14,11 +14,11 @@ Globalization: CA2101, CA1300-CA1310
 Mobility: CA1600-CA1601
 Performance: HA, CA1800-CA1838
 Security: CA2100-CA2153, CA2300-CA2330, CA3000-CA3147, CA5300-CA5403
-Usage: CA1801, CA1806, CA1816, CA2200-CA2209, CA2211-CA2249
+Usage: CA1801, CA1806, CA1816, CA2200-CA2209, CA2211-CA2250
```
buyaa-n, line 17: Please revert unrelated change

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthTests.cs
```diff
@@ -0,0 +1,511 @@
...
+Module Program
+    Sub Main(args As String())
+        Dim src = New Integer() {1, 2, 3, 4}
+        Dim dst = New Integer() {0, 0, 0, 0}
+
+        Buffer.BlockCopy(srcOffset:=0, src:=src, count:=[|src.Length|], dstOffset:=0, dst:=dst)
+    End Sub
+End Module
+");
+        }
```
buyaa-n, line 509: What about non-local arrays used in `BlockCopy`? Shouldn't we cover them?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+                    ImmutableArray<IArgumentOperation> arguments = IOperationExtensions.GetArgumentsInParameterOrder(invocationOperation.Arguments);
+
+                    if (arguments.Length != 5)
+                    {
+                        return;
+                    }
+
+                    IArgumentOperation sourceArgument = arguments[0];
+                    IArgumentOperation destinationArgument = arguments[2];
+                    IArgumentOperation countArgument = arguments[4];
```
jeffhandley, line 85: Suggestion: Add a comment here that shows the method signature that we are identifying

```suggestion
					// BlockCopy(Array src, int srcOffset, Array dst, int dstOffset, int count)
                    IArgumentOperation sourceArgument = arguments[0];
                    IArgumentOperation destinationArgument = arguments[2];
                    IArgumentOperation countArgument = arguments[4];
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+                    var invocationOperation = (IInvocationOperation)context.Operation;
+
+                    if (!invocationOperation.TargetMethod.Equals(blockCopyMethod))
+                    {
+                        return;
+                    }
+
+                    ImmutableArray<IArgumentOperation> arguments = IOperationExtensions.GetArgumentsInParameterOrder(invocationOperation.Arguments);
+
+                    if (arguments.Length != 5)
```
jeffhandley, line 78: Is it worth ensuring the types of the arguments match what we're looking for, in case a different overload with 5 arguments ever gets added?

(I'm not sure how defensive we tend to be in these scenarios)

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -18,6 +18,7 @@ CA1847 | Performance | Info | UseStringContainsCharOverloadWithSingleCharactersA
 CA2250 | Usage | Info | UseCancellationTokenThrowIfCancellationRequested, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2250)
 CA2251 | Usage | Hidden | UseStringEqualsOverStringCompare, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2251)
 CA2252 | Usage | Info | DetectPreviewFeatureAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2252)
+CA2017 | Reliability | Warning | BufferBlockCopyLengthAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2017)
```
jeffhandley, line 21: Nit: Move this up to be in numeric order

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthTests.cs
```diff
@@ -0,0 +1,549 @@
...
+        Dim numOfBytes = 8
+
+        Buffer.BlockCopy(src, 0, dst, 0, numOfBytes)
+    End Sub
+End Module
+");
+        }
+
+        [Fact]
+        public async Task NumOfBytesAsClassConstProperty()
```
jeffhandley, line 377: Should this be named "NumOfBytesAsClassConstField()"? And then have a test that uses a property getting?

mahdiva:
@Youssef1313 @buyaa-n @jeffhandley Thanks for the comments! The issues have been fixed. Please review the changes once again.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemBuffer, out INamedTypeSymbol? bufferType))
+                {
+                    return;
+                }
+
+                IMethodSymbol blockCopyMethod = bufferType
+                   .GetMembers("BlockCopy")
+                   .OfType<IMethodSymbol>()
+                   .FirstOrDefault();
```
Youssef1313, line 50: nit:

```suggestion
                   .FirstOrDefault(m => m is IMethodSymbol);
```

or

```suggestion
                   .FirstOrDefault(m => m.Kind == SymbolKind.Method);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemArray, out INamedTypeSymbol? arrayType))
+                {
+                    return;
+                }
+
+                IPropertySymbol arrayLengthProperty = arrayType
+                    .GetMembers("Length")
+                    .OfType<IPropertySymbol>()
+                    .FirstOrDefault();
```
Youssef1313, line 65: Same suggestion

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+                   .GetMembers("BlockCopy")
+                   .OfType<IMethodSymbol>()
+                   .FirstOrDefault();
+
+                if (blockCopyMethod is null)
+                {
+                    return;
+                }
+
+                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemArray, out INamedTypeSymbol? arrayType))
```
Youssef1313, line 57: I'd suggest using context.Compilation.GetSpecialType instead. It's a little bit faster.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
Youssef1313: > in case a different overload with 5 arguments ever gets added?

If we want to be defensive for this, `GetFirstOrDefaultMemberWithParameterInfos` extension method can be used.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+                    }
+
+                    if (arguments[0].Parameter.Type.SpecialType != SpecialType.System_Array ||
+                        arguments[1].Parameter.Type.SpecialType != SpecialType.System_Int32 ||
+                        arguments[2].Parameter.Type.SpecialType != SpecialType.System_Array ||
+                        arguments[3].Parameter.Type.SpecialType != SpecialType.System_Int32 ||
+                        arguments[4].Parameter.Type.SpecialType != SpecialType.System_Int32)
+                    {
+                        return;
+                    }
```
Youssef1313, line 90: This looks fragile. If we hit this, then we're in trouble. That means another overload is added, in that case a diagnostic may be valid but we're returning here because we didn't find the correct overload (because we're using FirstOrDefault for finding the method by name).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
jeffhandley: > `GetFirstOrDefaultMemberWithParameterInfos` extension method.

Oh, nice!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
jeffhandley: Good catch. The `GetFirstOrDefaultMemberWithParameterInfos` extension method approach will address what was intended here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+                    IArgumentOperation countArgument = arguments[4];
+
+                    bool CheckArrayLengthLocalReference(IArgumentOperation targetArgument, IPropertyReferenceOperation lengthPropertyArgument)
+                    {
+                        if (targetArgument.Value is IConversionOperation targetArgValue)
+                        {
+                            if (lengthPropertyArgument.Instance.GetReferencedMemberOrLocalOrParameter() == targetArgValue.Operand.GetReferencedMemberOrLocalOrParameter())
+                            {
+                                IArrayTypeSymbol countArgumentArrayTypeSymbol = (IArrayTypeSymbol)lengthPropertyArgument.Instance.Type;
+                                if (countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_Byte && countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_SByte)
```
buyaa-n, line 104: I think bool type should also be excluded

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+                        }
+
+                        if (semanticModel.GetOperationWalkingUpParentChain(declaringSyntaxReference.GetSyntax(cancellationToken), cancellationToken) is not IVariableDeclaratorOperation variableDeclaratorOperation)
+                        {
+                            return;
+                        }
+
+                        IVariableInitializerOperation variableInitializer = variableDeclaratorOperation.Initializer;
+
+                        if (variableInitializer is not null && variableInitializer.Value is IPropertyReferenceOperation variableInitializerPropertyReference && CheckLengthPropertyOnByteOrSByteArrays(variableInitializerPropertyReference))
```
buyaa-n, line 152:
```suggestion
                        if (variableInitializer is not null && variableInitializer.Value is IPropertyReferenceOperation variableInitializerPropertyReference &&
                         CheckLengthPropertyOnByteOrSByteArrays(variableInitializerPropertyReference))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+                        {
+                            context.ReportDiagnostic(countArgument.Value.CreateDiagnostic(Rule));
+                        }
+                        else if (variableDeclaratorOperation.Parent is IVariableDeclarationOperation variableDeclarationOperation && variableDeclarationOperation.Initializer is not null && variableDeclarationOperation.Initializer.Value is IPropertyReferenceOperation variableInitializerPropertyReferenceVB && CheckLengthPropertyOnByteOrSByteArrays(variableInitializerPropertyReferenceVB))
```
buyaa-n, line 156:
```suggestion
                        else if (variableDeclaratorOperation.Parent is IVariableDeclarationOperation variableDeclarationOperation &&
                        variableDeclarationOperation.Initializer is not null &&
                        variableDeclarationOperation.Initializer.Value is IPropertyReferenceOperation variableInitializerPropertyReferenceVB &&
                        CheckLengthPropertyOnByteOrSByteArrays(variableInitializerPropertyReferenceVB))
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthTests.cs
```diff
@@ -0,0 +1,537 @@
...
+    End Sub
+
+    Function GetSrcArray()
+        Return New Integer() {1, 2, 3, 4}
+    End Function
+End Module
+
+");
+        }
+    }
```
buyaa-n, line 536: Please add a test with a global array
```cs
class Program
{
    private char[] arr1;
    private char[] arr2;

    public void Test()
    {
        Buffer.BlockCopy(arr1, 0, arr2, 0, arr1.Length);
    }
}
```

buyaa-n: and another test with Array type:
```cs
Array arr1 = Array.CreateInstance(typeof(byte), 2), arr2 = Array.CreateInstance(typeof(byte), 2);
Buffer.BlockCopy(arr1, 0, arr2, 0, arr1.Length);
```

pgovind: Let me know if you need help with these. AFAICT these are the last 2 comments needing to be addressed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
pgovind: You're right that `bool` can be excluded here. https://github.com/dotnet/runtime/issues/33765 mostly talked about byte and `sbyte`, that's why @mahdiva dealt with those first. What do you think about filing an issue to fix the implementation here to exclude bool and getting this merged in the meanwhile?

pgovind: Filed https://github.com/dotnet/roslyn-analyzers/issues/5353 optimistically :) Please feel free to disagree if you think we should fix it in this PR itself!

jeffhandley: > What do you think about filing an issue to fix the implementation here to exclude bool and getting this merged in the meanwhile?

I support that approach

buyaa-n: > What do you think about filing an issue to fix the implementation here to exclude bool and getting this merged in the meanwhile?

I am OK with that, but i assumed it needs very little change like just need another condition for bool type, no? 
```if (countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_Byte && countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_SByte && countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_Bool)```

File: src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs (hunk 23 above)
Comment (mahdiva) [reply src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs:71]: Using the lambda filter here would return a `ISymbol` type, compared to `IPropertySymbol` when using `.OfType<IPropertySymbol>` which I think is more specific, but I believe they achieve the same result if I'm not mistaken

**Comment by mahdiva** [comment]
@Youssef1313 @buyaa-n @jeffhandley I've addressed your comments. Would appreciate it if you could review them again. Thanks!

File: src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs [hunk 30]
```diff
@@ -0,0 +1,169 @@
...
+                IPropertySymbol arrayLengthProperty = arrayType
+                    .GetMembers("Length")
+                    .OfType<IPropertySymbol>()
+                    .FirstOrDefault();
+
+                context.RegisterOperationAction(context =>
+                {
+                    var invocationOperation = (IInvocationOperation)context.Operation;
+
+                    if (!invocationOperation.TargetMethod.Equals(blockCopyMethod))
```
Youssef1313, line 77:
```suggestion
                    if (!invocationOperation.TargetMethod.Equals(blockCopyMethod, SymbolEqualityComparer.Default))
```

src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -97,10 +97,12 @@ internal static class WellKnownTypeNames
         public const string SystemAppContext = "System.AppContext";
         public const string SystemAppDomain = "System.AppDomain";
         public const string SystemArgumentException = "System.ArgumentException";
+        public const string SystemArray = "System.Array";
```
Youssef1313, line 100:
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
Youssef1313: It should be okay if the variable type is changed to `ISymbol`. The usage will still work.

jeffhandley:
The check failures indicate that `msbuild /t:pack` needs to be run (via the Visual Studio Developer Command Prompt).

jeffhandley (approved):
Approved with the condition of running `msbuild /t:Pack` and pushing the changes from that to resolve the failing checks.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
```diff
@@ -0,0 +1,168 @@
...
+                                if (countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_Byte &&
+                                countArgumentArrayTypeSymbol.ElementType.SpecialType != SpecialType.System_SByte)
+                                {
+                                    return true;
+                                }
+                            }
+                        }
+                        else if (targetArgument.Value is ILocalReferenceOperation local)
+                        {
+                            if (lengthPropertyArgument.Instance.GetReferencedMemberOrLocalOrParameter() == local.Local)
```
buyaa-n, line 103: Do not need to check the element type of the array?

buyaa-n (approved):
Added one more comment, overall the PR LGTM by assuming that the https://github.com/dotnet/roslyn-analyzers/issues/5353 will be resolved within 6.0

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
jeffhandley: I defer to @buyaa-n and @pgovind here.

buyaa-n: We should not warn for such cases, not sure if we can report different diagnostic with lower severity like`DiagnosticSeverity.Hidden` for this case (I know we can with different diagnostic ID but not sure if we can with same diagnostic id) CC @mavasani @jmarolf @sharwell

mahdiva: Opened #5371 to track this issue. The unit test for this case is being skipped for now.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
buyaa-n: > Anyway I am OK with resolving it later too

But this needs to be resolved within 6.o as it would cause false positive

pgovind:
CI passed. Merging. Thanks for the reviews everyone!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/BufferBlockCopyLengthAnalyzer.cs
mahdiva: Created PR #5373 that fixes this issue

PR #5243, opened 2021-07-14
Fix concurrency issue in CA1069 analyzer

### LINKED ISSUE #3871: CA1069: Indeterministic false positive

RikkiGibson:
Fixes #3871

I ended up revising the way this works pretty heavily.

Initially I tried to fix this by gathering up all the initializer operations into a `ConcurrentDictionary<IFieldSymbol, IFieldInitializerOperation>`, then enumerating the enum members at the end and looking up the initializers.

It occurred to me that most enum values are probably *not* duplicates, though, and it would probably take less memory and fewer modifications on concurrent collections by instead using:
1. a `membersByValue` dictionary which is populated single-threaded and read concurrently
2. a `duplicates` set which is read and written concurrently

@jmarolf @sharwell

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
@@ -47,9 +47,9 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
 
             context.RegisterSymbolStartAction(visitEnumSymbol, SymbolKind.NamedType);
 
-            void visitEnumSymbol(SymbolStartAnalysisContext symbolStartAnalysisContext)
+            void visitEnumSymbol(SymbolStartAnalysisContext context)
             {
-                if (symbolStartAnalysisContext.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Enum } enumSymbol)
+                if (context.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Enum } enumSymbol)
                 {
                     return;
                 }
@@ -80,11 +80,12 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                     }
                 }
 
-                symbolStartAnalysisContext.RegisterOperationAction(visitFieldInitializer, OperationKind.FieldInitializer);
+                context.RegisterOperationAction(visitFieldInitializer, OperationKind.FieldInitializer);
+                context.RegisterSymbolEndAction(endVisitEnumSymbol);
 
-                void visitFieldInitializer(OperationAnalysisContext operationAnalysisContext)
+                void visitFieldInitializer(OperationAnalysisContext context)
                 {
-                    var initializer = (IFieldInitializerOperation)operationAnalysisContext.Operation;
+                    var initializer = (IFieldInitializerOperation)context.Operation;
                     if (initializer.InitializedFields.Length != 1)
                     {
                         return;
@@ -97,7 +98,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                         if (initializer.Value is not IConversionOperation { Operand: IFieldReferenceOperation { Field: IFieldSymbol referencedField } }
                             || !SymbolEqualityComparer.Default.Equals(referencedField, duplicatedField))
                         {
-                            operationAnalysisContext.ReportDiagnostic(field.CreateDiagnostic(RuleDuplicatedValue, field.Name, field.ConstantValue, duplicatedField.Name));
+                            context.ReportDiagnostic(field.CreateDiagnostic(RuleDuplicatedValue, field.Name, field.ConstantValue, duplicatedField.Name));
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


Youssef1313:
There was another PR attempting to fix the same issue https://github.com/dotnet/roslyn-analyzers/pull/5170.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
```diff
@@ -246,6 +309,27 @@ public enum MyEnum
...
 
+        [Fact]
+        public async Task EnumValueReferenceExplicitDuplicateMember_Diagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public enum MyEnum
+{
+    Value1 = 1,
+    Value2 = Value1,
+    Value3 = Value2,
```
RikkiGibson, line 320: This is a behavior change. The idea here is that if you are going to explicitly include a duplicate, its initializer should reference the same "original". To fix this code the user should change `Value3 = Value2` to `Value3 = Value1`. I felt like this was in the spirit of the [docs](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1069) but I'm also fine with changing the behavior back so this test produces no diagnostics.

jmarolf (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValues.cs
```diff
@@ -48,136 +45,105 @@ public sealed override void Initialize(AnalysisContext context)
...
-                }, OperationKind.Binary);
-            }, SymbolKind.NamedType);
-        }
+                }
 
-        private static IEnumerable<IOperation> GetFilteredDescendants(IOperation operation, Func<IOperation, bool> descendIntoOperation)
-        {
-            using var stack = ArrayBuilder<IEnumerator<IOperation>>.GetInstance();
-            stack.Add(operation.Children.GetEnumerator());
+                symbolStartAnalysisContext.RegisterSymbolEndAction(endVisitEnumSymbol);
```
sharwell, line 134: ❗ Place all statements in a block before the first local function in that block

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
sharwell: I think this makes sense here, but might not for bitwise enums. I'll try to find an example.

sharwell: Consider [MethodAttributes](https://github.com/dotnet/runtime/blob/4c92aef2b08f9c4374c520e7e664a44f1ad8ce56/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodAttributes.cs) for example. The members `PrivateScope` and `ReuseSlot` have the same value; if another member were added to the end with the intended value `Family | ReuseSlot`, it would be annoying to have a warning appearing saying you need to put `Family | PrivateScope` instead (which is technically the same value, but clearly less meaningful).

RikkiGibson: Have added a test for this scenario in 37d8ae9159e0cb1e4b430e259840c2c2c6dfd0e6. It appears the diagnostics given have not changed from before this PR.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/EnumShouldNotHaveDuplicatedValuesTests.cs
```diff
@@ -425,6 +538,50 @@ End Enum
...
+
+        [Fact]
+        public async Task EnumMemberNullConstantValue_NoDiagnostic()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+public enum MyEnum
+{
+    Value1 = null
+}",
+                DiagnosticResult.CompilerError("CS0037").WithSpan(4, 14, 4, 18).WithArguments("int"));
```
sharwell, line 576: 💡 For this case I would normally just use markup syntax `{|CS0037:...|}` since it's an expected part of the input.

sharwell (approved):

PR #5252, opened 2021-07-17
General cleanup

### LINKED ISSUE #5251: Unreachable null check after call to GetSpecialType
https://github.com/dotnet/roslyn-analyzers/blob/cab4a924d6e23560508eabdd802455d0eae18752/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SpecifyStringComparison.cs#L61

It looks like `stringType` can never be null here. It can be an error type.

Youssef1313:
Closes #5251


src/Utilities/Compiler/FxCopWellKnownDiagnosticTags.cs
```diff
@@ -16,12 +13,5 @@ internal static class FxCopWellKnownDiagnosticTags
 
         public static readonly string[] PortedFxCopDataflowRule = new string[] { PortedFromFxCop, WellKnownDiagnosticTagsExtensions.Dataflow, WellKnownDiagnosticTags.Telemetry };
         public static readonly string[] PortedFxCopDataflowRuleEnabledInAggressiveMode = new string[] { PortedFromFxCop, WellKnownDiagnosticTagsExtensions.Dataflow, WellKnownDiagnosticTags.Telemetry, WellKnownDiagnosticTagsExtensions.EnabledRuleInAggressiveMode };
-
-        public static bool IsPortedFxCopRule(DiagnosticDescriptor diagnosticDescriptor)
-        {
-            var result = diagnosticDescriptor.CustomTags.Any(t => t == PortedFromFxCop);
-            Debug.Assert(!result || diagnosticDescriptor.Id.StartsWith("CA", StringComparison.OrdinalIgnoreCase));
-            return result;
-        }
```
Youssef1313, line 25: This is not currently used.

It may have been used in the deleted `FxCopAnalyzers.sln`.

src/Utilities/Compiler/WellKnownTypeNames.cs
```diff
@@ -53,13 +53,13 @@ internal static class WellKnownTypeNames
         public const string MicrosoftCodeAnalysisDiagnosticsSyntaxNodeAnalysisContext = "Microsoft.CodeAnalysis.Diagnostics.SyntaxNodeAnalysisContext";
         public const string MicrosoftCodeAnalysisDiagnosticsSyntaxTreeAnalysisContext = "Microsoft.CodeAnalysis.Diagnostics.SyntaxTreeAnalysisContext";
         public const string MicrosoftCodeAnalysisHostMefMefConstruction = "Microsoft.CodeAnalysis.Host.Mef.MefConstruction";
-        public const string MicrosoftCodeAnalysisLocalizableString = "Microsoft.CodeAnalysis.LocalizableString";
```
Evangelink, line 56: @mavasani Do you think the code style repo would accept a suggestion for an analyzer that sorts members by name? Shall we do something specific to `WellKnownTypeNames`?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
```diff
@@ -145,7 +145,7 @@ private static async Task<Document> ConvertToEqualsAsync(Document document, Sema
                 false =>
                     generator.InvocationExpression(
                         generator.MemberAccessExpression(
-                            generator.TypeExpression(semanticModel.Compilation.GetSpecialType(SpecialType.System_Object)),
+                            generator.TypeExpression(SpecialType.System_Object),
```
Evangelink, line 148: I guess that's a new API on roslyn side? I don't know why we would have gone through this more complex syntax otherwise.

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -286,6 +286,14 @@ public static IEnumerable<IParameterSymbol> GetParametersOfType(this IEnumerable
             return parameters.Where(p => p.Type.Equals(type));
         }
 
+        /// <summary>
+        /// Get parameters which type is the given special type
```
Evangelink, line 290: Suggestion

```suggestion
        /// Gets the parameters whose type is equal to the given special type.
```

Evangelink (commented):
Nice cleanup @Youssef1313

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
Youssef1313: @Evangelink Both APIs were introduced in 3.0.0. I think this approach was taken since this is the first overload shown by IntelliSense?

Youssef1313:
*Important before merging:* This may regress respecting user preferences for `CodeStyleOptions.PreferIntrinsicPredefinedTypeKeywordInMemberAccess`.

Converting to a draft for now.

mavasani:
Ping @Youssef1313

src/Utilities/Compiler/WellKnownTypeNames.cs
Evangelink: Can we enable FxCop on this repo?

Evangelink:
Ping @Youssef1313

Youssef1313:
> _Important before merging:_ This may regress respecting user preferences for `CodeStyleOptions.PreferIntrinsicPredefinedTypeKeywordInMemberAccess`.
> 
> Converting to a draft for now.

I haven't yet looked at this unfortunately.

Evangelink:
@Youssef1313 Thanks for the heads-up!

Youssef1313:
For reference, https://github.com/dotnet/roslyn/pull/55004 was what revealed the potential problem with this PR.

We can still take the small parts of this PR that doesn't use `TypeExpression(SpecialType)` overload. However, I think Roslyn should have respected user preferences for this overload. cc @sharwell @CyrusNajmabadi

Youssef1313:
@mavasani @Evangelink Sorry for the delay. I reverted the potentially problematic changes.

Evangelink (approved):

PR #5265, opened 2021-07-20
readme: nearly vertically center badges with `<sub>`

ZacharyPatten:
This is just a small update to the readme. If you wrap badges in `<sub>`_badge_`</sub>` it _nearly_ vertically centers them as opposed to them to being offset above the text line.

Before:
image

After:
image

I just wanted to make the suggestion as I think it looks nicer, but feel free to disregard this PR if not desired.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -3,6 +3,9 @@
   "MD013": false,
   "MD024": false,
   "MD026": false,
+  "MD033": {
+    "allowed_elements": [ "sub" ]
+  },
   "MD041": false,
   "MD046": {
     "style": "fenced"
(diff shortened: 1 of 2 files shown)


ZacharyPatten:
`roslyn-analyzers-CI (Markdownlint)` intentionally fails on inline HTML. Does that mean inline HTML isn't allowed? If yes, this can obviously be closed.

Youssef1313:
@ZacharyPatten I believe you can configure that in https://github.com/dotnet/roslyn-analyzers/blob/main/.markdownlint.json as follows:

```json
  "MD033": {
    "allowed_elements": [ "sub" ]
  }
```

Evangelink:
I think that the forbid html rule might not make much sense for the readme but we could probably start with some includes rather than simply disabling the rule for this file.

PR #5266, opened 2021-07-20
CA1416 MacCatalyst is a superset of iOS

buyaa-n:
Related to https://github.com/dotnet/runtime/issues/53084
## Acceptance Criteria

To consider the proposal of using `SupportedOSPlatformGuard` attributes on the `OperatingSystem` methods to achieve a relationship of MacCatalyst as a superset of IOS, we will utilize the following acceptance criteria.

1. APIs annotated as supported on IOS are available on MacCatalyst without warning
    * This allows existing APIs annotated as supported on IOS to be automatically supported on MacCatalyst
2. Guarding those calls only requires checking a single platform
    * `OperatingSystem.IsIOS()` matches both IOS and MacCatalyst
3. APIs can still be annotated as being supported on one but not both IOS and MacCatalyst
    * `OperatingSystem.IsIOS() && !OperatingSystem.IsMacCatalyst()` matches only IOS but not MacCatalyst
    * `OperatingSystem.IsMacCatalyst()` matches only MacCatalyst but not IOS
4. New usage of `[SupportedOSPlatformGuard]` attributes is consistent with existing behavior
5. `[UnsupportedOSPlatform]` and `[UnsupportedOSPlatformGuard]` attributes must continue to work as expected
    * `[UnsupportedOSPlatform]` attributes must respect the platform relationship defined
    * This allows existing APIs annotated as unsupported on IOS to be automatically unsupported on MacCatalyst
6. The performance impact of the change must be negligible

## Proposal Summary

1. Update `OperatingSystem.IsIOS` to return `true` when the platform is "ios" *OR* "maccatalyst"
2. Update the Platform Compatibility Analyzer to respect `[SupportedOSPlatformGuard]` attributes on `System.OperatingSystem`
    * Load these into a map during analyzer initialization
    * Use fallback logic of inferring the method's platform guard from the method name if no attributes exist
3. Annotate `OperatingSystem.IsIOS` with [`SupportedOSPlatformGuard("maccatalyst")]`
4. Update the Platform Compatibility Analyzer such that when identifying an API's supported/unsupported platforms, the platform set is expanded to include all platforms indicated through `[SupportedOSPlatformGuard]` attributes on that platform's `Is{Platform}` method

### Illustration of Acceptance Criteria
```csharp
public class OperatingSystem
{
    // Guard methods that simulate the OperatingSystem members
    // [SupportedOSPlatformGuard("ios")] this annotation is not needed as the method name express that it is iOS guard 
    [SupportedOSPlatformGuard("maccatalyst")]
    public static bool IsIOS() => IsIOS() || IsMacCatalyst();

    // [SupportedOSPlatformGuard("maccatalyst")] for same reason annotation doesn't need
    public static bool IsMacCatalyst() => IsMacCatalyst();
    ...
}

public class AnnotationScenarios    // APIs for the scenarios
{
    // The [SupportedOSPlatform("maccatalyst")] attribute would not exist in the code
    // but it would be inferred by the analyzer.
    [SupportedOSPlatform("ios")]
    void SupportedOnIOSAndMacCatalyst() { }

    // Having the [SupportedOSPlatform("maccatalyst")] explicitly
    // has no effect, it is same as above annotation
    [SupportedOSPlatform("maccatalyst")]
    [SupportedOSPlatform("ios")]
    void SupportedOnIOSAndMacCatalyst() { }

    [SupportedOSPlatform("ios")]
    // The following attribute could be used
    // to cancel the inferred "maccatalyst" support
    [UnsupportedOSPlatform("maccatalyst")]
    void SupportedOnIOS_ButNotMacCatalyst() { }

    // Unlike the [SupportedOSPlatform("ios")] 
    // this will not infer iOS support, only support  maccatalyst
    [SupportedOSPlatform("maccatalyst")]
    void SupportedOnMacCatalyst() { }

    [UnsupportedOSPlatform("ios")]
    // The following attribute would not exist in the code
    // but it would be inferred by the analyzer.
    // that also [UnsupportedOSPlatform("maccatalyst")]
    void UnsupportedOnIOSAndMacCatalyst() { }

    // only unsupported on maccatalyst
    [UnsupportedOSPlatform("maccatalyst")]
    void UnsupportedOnMacCatalyst() { }

    void Main()
    {
        // This block is reachable on both IOS and MacCatalyst
        if (OperatingSystem.IsIOS())
        {
            SupportedOnIOSAndMacCatalyst();      // Scenario 1a - No warning
            SupportedOnMacCatalyst();           // Scenario 1b - Warning
            SupportedOnIOS();                   // Scenario 1c - No warning
            SupportedOnIOS_ButNotMacCatalyst(); // Scenario 1d - Warning
            UnsupportedOnIOS();                 // Scenario 1e - Warning
            UnsupportedOnMacCatalyst();         // Scenario 1f - Warning
        }

        // This block is reachable on MacCatalyst but not IOS
        if (OperatingSystem.IsMacCatalyst())
        {
            SupportedOnIOSAndMacCatalyst();      // Scenario 2a - No warning
            SupportedOnMacCatalyst();           // Scenario 2b - No warning
            SupportedOnIOS();                   // Scenario 2c - No warning
            SupportedOnIOS_ButNotMacCatalyst(); // Scenario 2d - Warning
            UnsupportedOnIOS();                 // Scenario 2e - Warning
            UnsupportedOnMacCatalyst();         // Scenario 2f - Warning
        }

        // This code block is reachable on IOS but not MacCatalyst
        if (OperatingSystem.IsIOS() && !OperatingSystem.IsMacCatalyst())
        {
            SupportedOnIOSAndMacCatalyst();      // Scenario 3a - No warning
            SupportedOnMacCatalyst();           // Scenario 3b - Warning
            SupportedOnIOS();                   // Scenario 3c - No warning
            SupportedOnIOS_ButNotMacCatalyst(); // Scenario 3d - UPDATE: NO WARNING
            UnsupportedOnIOS();                 // Scenario 3e - Warning
            UnsupportedOnMacCatalyst();         // Scenario 3f - No warning
        }
    }
}
```

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
@@ -121,24 +121,6 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 return infosBuilder.Any();
             }
 
-            private static bool TryExtractPlatformName(string methodName, [NotNullWhen(true)] out string? platformName)
-            {
-                if (!methodName.StartsWith(IsPrefix, StringComparison.Ordinal))
-                {
-                    platformName = null;
-                    return false;
-                }
-
-                if (methodName.EndsWith(OptionalSuffix, StringComparison.Ordinal))
-                {
-                    platformName = methodName.Substring(2, methodName.Length - 2 - OptionalSuffix.Length);
-                    return true;
-                }
-
-                platformName = methodName[2..];
-                return true;
-            }
-
             private static bool TryDecodeRuntimeInformationIsOSPlatform(
                 IOperation argumentValue,
                 INamedTypeSymbol? osPlatformType,
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
@@ -107,20 +107,6 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 return false;
             }
 
-            public static bool TryDecode(ImmutableArray<AttributeData> attributes, ArrayBuilder<PlatformMethodValue> infosBuilder)
-            {
-                foreach (var attribute in attributes)
-                {
-                    if (attribute.AttributeClass.Name is SupportedOSPlatformGuardAttribute or UnsupportedOSPlatformGuardAttribute &&
-                        TryParsePlatformNameAndVersion(attribute, out var platformName, out var version))
-                    {
-                        var info = new PlatformMethodValue(platformName, version, negated: attribute.AttributeClass.Name == UnsupportedOSPlatformGuardAttribute);
-                        infosBuilder.Add(info);
-                    }
-                }
-                return infosBuilder.Any();
-            }
-
             private static bool TryDecodeRuntimeInformationIsOSPlatform(
                 IOperation argumentValue,
                 INamedTypeSymbol? osPlatformType,
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -1540,7 +1540,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                     return;
                 }
 
-                CheckAttributesConsistency(childAttributes, relatedPlatforms);
+                CheckAttributesConsistency(childAttributes);
                 var pAttributes = parentAttributes.Platforms;
                 if (pAttributes != null && !pAttributes.IsEmpty)
                 {
@@ -1633,6 +1633,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                         }
                     }
 
+                    CheckAttributesConsistency(pAttributes);
                     if (crossPlatform && parentAttributes.IsAssemblyAttribute)
                     {
                         foreach (var childAttribute in childAttributes)
@@ -1675,23 +1676,21 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
 
                 return;
 
-                static void CheckAttributesConsistency(SmallDictionary<string, Versions> childAttributes,
-                    SmallDictionary<string, (string relatedPlatform, bool isSubset)> relatedPlatforms)
+                static void CheckAttributesConsistency(SmallDictionary<string, Versions> childAttributes)
                 {
                     bool allowList = false;
                     using var unsupportedList = PooledHashSet<string>.GetInstance();
 
                     foreach (var (platform, attributes) in childAttributes)
                     {
+                        NormalizeAttribute(attributes);
                         if (AllowList(attributes))
                         {
-                            NormalizeSupportedList(attributes, childAttributes, platform, relatedPlatforms, unsupportedList);
                             allowList = true;
                         }
                         else
                         {
                             Debug.Assert(DenyList(attributes));
-                            NormalizeUnsupportedList(attributes);
                             unsupportedList.Add(platform);
                         }
                     }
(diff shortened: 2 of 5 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
```diff
@@ -121,24 +121,6 @@ public static bool TryDecode(ImmutableArray<AttributeData> attributes, ArrayBuil
                 return infosBuilder.Any();
             }
 
-            private static bool TryExtractPlatformName(string methodName, [NotNullWhen(true)] out string? platformName)
```
buyaa-n, line 124: Moved to the main class for reuse

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -509,7 +563,20 @@ private static bool ComputeNeedsValueContentAnalysis(IOperation operationBlock,
...
+                            foreach (var val in value.AnalysisValues)
+                            {
+                                if (parentEnumerator.MoveNext() && !parentEnumerator.Current.Equals(val.GetNegatedValue()))
+                                {
+                                    goto label;
+                                }
+                            }
+                            continue;
+                        }
+
```
buyaa-n, line 578: This is for merging `OperatingSystem.IsIOS() && !OperatingSystem.IsMacCatalyst()` conditional which flow analysis value is `(OperatingSystem.IsIOS() || OperatingSystem.IsMacCatalyst()) && !OperatingSystem.IsMacCatalyst()`. Ideally it should be handled/merged by the flow analisys into just `OperatingSystem.IsIOS()` CC @mavasani

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -175,23 +174,58 @@ public override void Initialize(AnalysisContext context)
...
+                static IEnumerable<IMethodSymbol> FilterPlatformCheckMethods(INamedTypeSymbol symbol, SmallDictionary<string, string> relatedPlatforms)
+                {
+                    return symbol.GetMembers().OfType<IMethodSymbol>().Where(m =>
+                    {
+                        if (m.IsStatic &&
+                            m.ReturnType.SpecialType == SpecialType.System_Boolean &&
+                            (IsOSPlatform == m.Name) || NameAndParametersValid(m))
+                        {
+                            CheckDependentPlatofmrs(m, relatedPlatforms);
```
jeffhandley, line 207: Typo in method name `CheckDependentPlatofmrs`.

```suggestion
                            CheckDependentPlatforms(m, relatedPlatforms);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -175,23 +174,58 @@ public override void Initialize(AnalysisContext context)
...
+                            return true;
+                        }
+
+                        return false;
+                    });
+                }
+            }
+
+            static void CheckDependentPlatofmrs(IMethodSymbol m, SmallDictionary<string, string> relatedPlatforms)
```
jeffhandley, line 216:
```suggestion
            static void CheckDependentPlatforms(IMethodSymbol m, SmallDictionary<string, string> relatedPlatforms)
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -509,7 +563,20 @@ private static bool ComputeNeedsValueContentAnalysis(IOperation operationBlock,
...
 
-                        if (!IsKnownValueGuarded(parentAttributes, ref csAttributes, parent, parentCapturedVersions))
+                        if (value.AnalysisValues.Count != 0 && value.AnalysisValues.Count == parent.AnalysisValues.Count)
+                        {
+                            var parentEnumerator = parent.AnalysisValues.GetEnumerator();
+                            foreach (var val in value.AnalysisValues)
+                            {
+                                if (parentEnumerator.MoveNext() && !parentEnumerator.Current.Equals(val.GetNegatedValue()))
+                                {
+                                    goto label;
```
jeffhandley, line 573: Suggestion: Use `break` instead of `goto`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
```diff
@@ -139,15 +139,14 @@ public override void Initialize(AnalysisContext context)
...
-
-                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(typeName, out var operatingSystemType))
+                if (!context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemOperatingSystem, out var operatingSystemType))
                 {
                     return;
                 }
 
                 var osPlatformType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesOSPlatform);
                 var runtimeInformationType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeInteropServicesRuntimeInformation);
+                var mockOSType = context.Compilation.GetOrCreateTypeByMetadataName("System.MockOperatingSystem");
```
jeffhandley, line 149: Is `mockOSType` (`System.MockOperatingSystem`) needed only for allowing the tests to override the annotations on `System.OperatingSystem`? If so, we shouldn't keep this code in here; instead, we could request to have the runtime updated in this repo to have the annotated `System.OperatingSystem` and comment out the tests that need them until that update occurs.

jeffhandley (changes requested):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: No, it was breaking the outer loop, anyway used a function instead

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
buyaa-n: > Is mockOSType (System.MockOperatingSystem) needed only for allowing the tests to override the annotations on System.OperatingSystem

Yes and `OperatingSystem.IsMacCatalys()` is also not available, we used mock attributes/APIs for the previous versions until the repo could use new APIs. Anyway, moved the mock API and tests into Debug mode, hope the CI would pass

jeffhandley (commented):
I see a lot of test cases that have comments indicating the message states the code is "reachable on all platforms" but the code is within some sort of a guard. I wasn't sure if those were accurate messages or copy/paste errors in the comments.

buyaa-n:
> I see a lot of test cases that have comments indicating the message states the code is "reachable on all platforms" but the code is within some sort of a guard. I wasn't sure if those were accurate messages or copy/paste errors in the comments.

That is accurate, if the diagnostic is found inside a platform check method and the issue is caused by platform check then the message shows `This call site is reachable/unreachable on: 'Platform'`. But If the platform check is not causing the problem, for example, if the condition is `!OperatingSystem.IsWindows()` and if the calling method is Linux only API then the message `This call site is reachable on all platforms. 'LinuxOnlyAPI()' is only supported on : 'Linux'` was more meaningful instead of `This call site is unreachable on: 'Windows'. The 'LinuxOnlyAPI()' is only supported on : 'Linux'`

jeffhandley (approved):

buyaa-n:
Seems @mavasani is out, if he gives a feedback later i will apply them with different PR, gonna merge

PR #5268, opened 2021-07-21
Remove Eula.rtf from NuGet packages

### LINKED ISSUE #3347: NuGet package claims Apache-2.0 license but contains EULA.rtf with different terms
#### Analyzer package

#### Package Version

v2.9.8 (Latest)

#### Diagnostic ID

N/A

#### Repro steps

1. Open <https://www.nuget.org/packages/Microsoft.CodeQuality.Analyzers/2.9.8> in a browser and see that it says "License: Apache-2.0".
2. Download `microsoft.codequality.analyzers.2.9.8.nupkg` from <https://www.nuget.org/api/v2/package/Microsoft.CodeQuality.Analyzers/2.9.8>.
3. Unzip `microsoft.codequality.analyzers.2.9.8.nupkg`.
4. Note that the unzipped `Microsoft.CodeQuality.Analyzers.nuspec` likewise contains `<license type="expression">Apache-2.0</license>` and `<licenseUrl>https://licenses.nuget.org/Apache-2.0</licenseUrl>`.
5. Look at the other extracted files.

#### Expected behavior

The NuGet package can be used and redistributed under the terms of the Apache-2.0 license, although the package could contain third-party dependencies with other compatible licenses.

#### Actual behavior

The package also contains `EULA.rtf` with "MICROSOFT PRE-RELEASE SOFTWARE LICENSE TERMS", "MICROSOFT CODE ANALYSIS 2017". These terms appear to forbid distributing the software as is, without adding primary functionality.

It is not clear which license terms I have to comply with if I redistribute the NuGet package. Was `EULA.rtf` intended to apply to VSIX packages only?

On a related note, the source repository contains both `THIRD-PARTY-NOTICES.txt` and `assets/ThirdPartyNotices.rtf`, the latter of which is also in the NuGet package. It is not entirely clear which of these is the "NOTICE" text file referenced by clause 4. (d) of the Apache-2.0 license.

Evangelink:
Fixes #3347

### DIFF
--- a/src/Tools/GenerateAnalyzerNuspec/Program.cs
+++ b/src/Tools/GenerateAnalyzerNuspec/Program.cs
@@ -265,7 +265,6 @@ if (analyzerConfigurationFileDir.Length > 0 && Directory.Exists(analyzerConfigur
     }
 }
 
-result.AppendLine(FileElement(Path.Combine(assetsDir, "EULA.rtf"), ""));
 result.AppendLine(FileElement(Path.Combine(assetsDir, "ThirdPartyNotices.rtf"), ""));
 result.AppendLine(@"  </files>");
 


src/Tools/GenerateAnalyzerNuspec/Program.cs
```diff
@@ -265,7 +265,6 @@
     }
 }
 
-result.AppendLine(FileElement(Path.Combine(assetsDir, "EULA.rtf"), ""));
```
Evangelink, line 268: I haven't removed the file itself because it is still referenced by the `.vsix` files.

Youssef1313: Any reason to not remove all its references and delete it?

Evangelink: It's still used by vsix and as we keep generating them, I don't know what the team wants to do. But that's why I did called it out.

Evangelink:
@mavasani Seems like there is a problem with the stability of the performance check. This PR should have no impact on performance.

jmarolf (approved):

Evangelink:
@jmarolf I have rebased on latest main and now the build is green, PR is ready to be merged.

Evangelink:
Thank you @jmarolf

PR #5271, opened 2021-07-21
Use StringBuilder instead of string concat in loops

Evangelink:


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_IdRangeAndCategoryValidation.cs
```diff
@@ -351,34 +354,39 @@ private static bool TryParseIdRangeEntry(string entry, out string prefix, out in
...
                 {
-                    prefix += ch;
+                    prefixBuilder.Append(ch);
                 }
             }
 
-            return prefix.Length > 0 && suffixStr.Length > 0 && int.TryParse(suffixStr, out suffix);
+            prefix = prefixBuilder.ToString();
+            return prefix.Length > 0 &&
+                suffixStr.Length > 0 &&
```
Youssef1313, line 388: This check looks redundant given the next `TryParse` check?

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_IdRangeAndCategoryValidation.cs
```diff
@@ -118,19 +119,21 @@ static bool ShouldValidateRange((string? prefix, int start, int end) range)
...
                 foreach (var range in allowedIdsInfoList)
                 {
                     if (arg3.Length != 0)
                     {
-                        arg3 += ", ";
+                        arg3.Append(", ");
                     }
 
-                    arg3 += !ShouldValidateRange(range) ? range.prefix + "XXXX" : $"{range.prefix}{range.start}-{range.prefix}{range.end}";
+                    arg3.Append(!ShouldValidateRange(range)
```
Therzok, line 130: Should probably use append format or split it into multiple append calls for the interpolated string.

Evangelink: You are absolutely right!

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer_IdRangeAndCategoryValidation.cs
Evangelink: Yes, I will double check quickly the impl of `TryParse` just to make sure that's not some kind of optimization linked to the knowledge of how `TryParse` works.

Evangelink: I had a look and I'd say both options are fine by me. The length check seems relevant given the implementation of `TryParse` but isn't functionally required.

PR #5273, opened 2021-07-21
Bug fix: Only tag types implementing empty Preview interfacees

pgovind:
Just a small change. Currently the DetectPreviewFeatures analyzer tags all types implementing empty interfaces with a diagnostic. This is a bug. We should only tag types implementing empty Preview interfaces.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
@@ -112,7 +112,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             foreach (INamedTypeSymbol anInterface in interfaces)
             {
                 var interfaceMembers = anInterface.GetMembers();
-                if (interfaceMembers.Length == 0)
+                if (interfaceMembers.Length == 0 && ProcessPreviewAttribute(anInterface, requiresPreviewFeaturesSymbols))
                 {
                     // Only tag empty interfaces to prevent breaking changes in the future
                     requiresPreviewFeaturesSymbolsToUsageType.GetOrAdd(symbol, PreviewFeatureUsageType.EmptyInterface);
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
@@ -584,6 +584,33 @@ namespace Preview_Feature_Scratch
         }
 
         [Fact]
+        public async Task TestNonPreviewEmptyInterface()
+        {
+            var csInput = @" 
+using System.Runtime.Versioning; using System;
+namespace Preview_Feature_Scratch
+{" +
+@"
+
+    class Program : IProgram
+    {
+        static void Main(string[] args)
+        {
+            new Program();
+        }
+    }
+
+    public interface IProgram
+    {
+    }
+}
+
+    ";
+
+            await TestCS(csInput);
+        }
+
+        [Fact]
         public async Task TestInterfaceMethodInvocation()
         {
             var csInput = @" 


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -112,7 +112,7 @@ private static bool ProcessTypeSymbolAttributes(ITypeSymbol symbol, ConcurrentDi
             foreach (INamedTypeSymbol anInterface in interfaces)
             {
                 var interfaceMembers = anInterface.GetMembers();
-                if (interfaceMembers.Length == 0)
+                if (interfaceMembers.Length == 0 && ProcessPreviewAttribute(anInterface, requiresPreviewFeaturesSymbols))
```
stephentoub, line 115: Why does it matter if there are 0 members?  I see the comment below, but I'm not understanding what's special about 0 vs non-0 members... ?

pgovind: So we (myself, Tanner, Immo and Jeff) had a discussion a while back went through this thought process:

Assume we had the following code:
```csharp
    class Program : IProgram
    {
        {
            new Program();
        }
    }

    [RequiresPreviewFeatures]
    public interface IProgram
    {
    }
```
where IProgram is an interface defined in version 1 of some dependency. Currently, the DetectPreviewFeature analyzer won't report any diagnostics in this code. We made an explicit choice not to report diagnostics on `Program`'s class definition. Instead we report diagnostics whenever a preview method/property/field/event etc is used/invoked. Since `IProgram` has no members, no diagnostics are reported here. 
1. So, in effect, `Program`'s author has opted into preview features here without knowing about it (even though no preview features are used). 
2. The other concern was if `IProgram` was updated in V2 to now have preview methods, and those preview methods are now called by `Program`, the analyzer will flag a diagnostic. `Program`'s author now has to opt into PreviewFeatures (or go back to a previous version), and we considered that to be a breaking change. (tagging @jeffhandley / @terrajobst  in case he can explain more clearly here)

TLDR: We don't want to surprise folks. We're trying to be as thorough as we can in reporting when an assembly is using PreviewFeatures

stephentoub: > We made an explicit choice not to report diagnostics on Program's class definition.

Why? Why not warn at any mention of something preview?

pgovind: Because, Byte (for example) is currently defined like below and we can't mark `Byte`s definition with the `RequiresPreviewFeatures` attribute (or all uses of Byte will now be considered Preview):

```csharp
    [Serializable]
    [StructLayout(LayoutKind.Sequential)]
    [TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
    public readonly struct Byte : IComparable, IConvertible, ISpanFormattable, IComparable<byte>, IEquatable<byte>
#if FEATURE_GENERIC_MATH
#pragma warning disable SA1001
        , IBinaryInteger<byte>,
          IMinMaxValue<byte>,
          IUnsignedNumber<byte>
#pragma warning restore SA1001
```

stephentoub: Byte isn't annotated as preview.

buyaa-n: >  There would be a warning where Byte implements the interface, not on uses of Byte.

Yes and for suppressing the warning we will need to add `RequiresPreviewFeatures` attribute on the Byte which we don't want to

buyaa-n (approved):
LGTM

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
stephentoub: > Yes and for suppressing the warning we will need to add RequiresPreviewFeatures attribute on the Byte which we don't want to

We'd just use #pragma warning disable.

buyaa-n: Right but i guess we want to propagate the warning by adding the attribute

stephentoub: We don't want to propagate the warning in that case, because Byte isn't preview.

It's very strange to me that we're special-casing a particular shape of interface and only warning on inheritance for some but not others.  Either we want to warn on the definition or we don't.

> The other concern was if IProgram was updated in V2 to now have preview methods, and those preview methods are now called by Program, the analyzer will flag a diagnostic. Program's author now has to opt into PreviewFeatures (or go back to a previous version), and we considered that to be a breaking change. (tagging @jeffhandley / @terrajobst in case he can explain more clearly here)

If someone adds something to an interface, that's already a breaking change, forcing the implementer to implement those new methods.  And even if it's a DIM, when Program is changed to call it, yes, they're using a new preview feature, they should get a warning.

buyaa-n: > We don't want to propagate the warning in that case, because Byte isn't preview.

Yes, therefore by my understanding when it is not empty interface it will warn on the implemented methods not on the interface implementation of `Byte`, so that the implementations can have the attribute and propagate the warning

pgovind: > There would be a warning where Byte implements the interface
> Either we want to warn on the definition or we don't.

I'm a little bit confused here about whether you mean 1) `Byte`'s type definition where it implements `IBinaryInteger<byte>` or 2) The implicit/explicit implementations of `IBinaryInteger<byte>`'s methods/properties in the `Byte` type.

If you meant 1, that would mean either marking the `Byte` type itself as Preview (this is a no-go, this would make existing APIs in Byte preview) or suppressing the warning on `Byte`'s type definition. We considered suppressing the warning, but @tannergooding preferred not touching the type definitions and instead just wanted to mark the preview method implementations. He had a good reason related to behavior when adding a new preview API to `IBinaryInteger<byte>` for the next release (something about code readability/breaking change here, but I can't remember it exactly right now :/). 

> If someone adds something to an interface, that's already a breaking change, forcing the implementer to implement those new methods.

We considered that. @terrajobst , I think you can answer this better than me?

tannergooding: Once we start having to suppress the warnings, we run the risk of not actually understanding what is/isn't preview and the risk of suppressing an actual issue and missing it.

Differentiating on "empty" vs "non-empty" interface isn't great, but it helps avoid that issue by ensuring that we have a valid location to say "this must be preview"
* For non-empty interfaces, you must mark the entire implementing type or each method (whether implicitly or explicitly implemented)
* For empty interfaces, there is no method to annotate it on so its assumed to be a "marker" interface and the type needs the annotation

terrajobst: @pgovind

> > If someone adds something to an interface, that's already a breaking change, forcing the implementer to implement those new methods.
> 
> We considered that. @terrajobst , I think you can answer this better than me?

The argument was: it's a preview interface, adding or removing members is fine. The problem is if we decide to warn on implementing interface members, rather than implementing the interface because now it means that an interface with 0 members now wouldn't produce any diagnostics.

@stephentoub 

> > Yes and for suppressing the warning we will need to add RequiresPreviewFeatures attribute on the Byte which we don't want to
> 
> We'd just use #pragma warning disable.

My thinking was similar to that of Stephen and I believe warning on the interface implementation is appropriate (regardless of how many members there are). However, we had a meeting where @tannergooding pushed back and provided the following rationale:

1. If you implement an interface you don't want a single warning for the interface, you want a warning per member and decide individually which ones you're going to suppress vs. allow, rather than doing it only for the interface implementation syntax at the type header level.
2. And once you report a diagnostic per member, it's a bit noisy for the common case where the interface has any members.
3. So we decided to compromise and only produce a diagnostic for the interface implementation if the interface has zero members.

@tannergooding @jeffhandley @pgovind does this summary sound accurate?

tannergooding: That is, while suppressing warnings does exists, it is very much my view that the actual suppression should only occur if there is a legitimate reason (often a bug or edge case in the analyzer itself).

We shouldn't be required to suppress something that is a de-facto/normal/default use case (e.g. implementing a preview interface on a non-preview type, which is our own primary use-case).

terrajobst: > We shouldn't be required to suppress something that is a de-facto/normal/default use case (e.g. implementing a preview interface on a non-preview type, which is our own primary use-case).

I'd be OK with saying that it's always an edge case in the sense that it highly depends on the semantics of the interface and whether that's considered OK. For our feature, it wouldn't be viable if implementing the interface would "infect" the implementing type as preview too, but I believe this to be a special case. I'd think in most cases where we add preview interface the customer can, and probably should, implement those on net-new types. And if that's not desirable, then I think it's OK to say "suppress, but buyer beware".

stephentoub: Taking a step back, my concern here is two-fold:

a) We're creating an inconsistent experience around what warns and what doesn't.  Saying "any utterance of a preview type/member requires enabling previews" is super simple.  Trying to special-case this but not that but only when this and so on is much more complicated, hard for a user to understand, ...

b) And error prone.  I expect an analyzer that flagged such an "any utterance" rule would be striaghtforward and fairly airtight.  In contrast, we leave open a lot of gaps by trying to catch each individual thing.  Just as some quick examples, I'd expect all of these to warn, and none of them do:

```C#

Console.WriteLine(typeof(IFoo));

[RequiresPreviewFeatures]
interface IFoo { }
```

```C#

Console.WriteLine(nameof(IFoo));

[RequiresPreviewFeatures]
interface IFoo { }
```

```C#

Console.WriteLine("hello");

interface IZoo : IFoo
{
}

[RequiresPreviewFeatures]
interface IFoo
{
    void Bar();
}
```

```C#

Console.WriteLine("hello");

class A<T> where T : IFoo
{
}

[RequiresPreviewFeatures]
interface IFoo
{
    void Bar();
}
```

```C#

Console.WriteLine(new A<Foo>());

class A<T> where T : IFoo, new()
{
    public A()
    {
        new T().Bar();
    }
}

[RequiresPreviewFeatures]
class Foo : IFoo
{
    public void Bar() { }
}

interface IFoo
{
    void Bar();
}
```

```C#

Console.WriteLine("hello");

[My]
class A
{
}

[RequiresPreviewFeatures]
[AttributeUsage(AttributeTargets.All)]
class MyAttribute : Attribute
{
}
```

```C#

Console.WriteLine(A.B());

class A
{
    [RequiresPreviewFeatures]
    public static bool B() => true;
}
```

and so on.  The purpose of the analyzer is to help folks know "you're depending on something that's not supported and might change or be renamed or be deleted entirely", and so not flagging every use of such a preview thing is a bug.  Keep in mind RequiresPreviewFeatures isn't just about numerics or statics in interfaces; it's a feature all on its own that anyone should be able to use to annotate anything that's preview, not just interfaces or interface members.

On top of all that, the user experience of special-casing empty interfaces is inconsistent.  Why does this warn twice:
```C#

Console.WriteLine("hello");

abstract class A : IZoo
{
    void IFoo.Bar() { }
}

interface IZoo : IFoo { }

[RequiresPreviewFeatures]
interface IFoo
{
    void Bar();
}
```
but this only once:
```C#

Console.WriteLine("hello");

abstract class A : IFoo
{
    void IFoo.Bar() { }
}

interface IZoo : IFoo { }

[RequiresPreviewFeatures]
interface IFoo
{
    void Bar();
}
```
? Etc.

pgovind: All good test cases. I can some of these, but I feel like it'd be better to get on a call instead. I opened https://github.com/dotnet/roslyn-analyzers/issues/5284 now to capture these test cases. How about I flush out the analyzer implementation more, and set up a meeting next week to go over the edge cases/design choices?

jeffhandley: I agree with @stephentoub here. When we had the design discussion referenced above, we called out the risks of adding this special-casing logic and how that introduces risk of a long bug tail with endless edge cases. Learning from past experience with analyzers, I think we should pivot here back to the simple/straightforward design of "any utterance of a preview type/member requires enabling previews" (as @stephentoub described it).

terrajobst: Agreed. I like this framing:

> Saying "any utterance of a preview type/member requires enabling previews" is super simple. 

Which is precisely the reason why I expected ` class Foo : IPreviewThingier` to warn in the first place.

tannergooding: I am not a fan of that approach. I think it is the exact opposite of how this will get used in practice and will result in a longer tail of bugs due to users having to suppress the analyzer for basic usages in practice.

tannergooding: The entire purpose of this feature is to allow shipping preview features alongside existing "stable" features. This, by definition, means we will (and users will) be defining preview types and methods that live side by side with their non preview equivalents.

One such scenario involves preview interfaces that are implemented by non-preview types (this is our core scenario in generic math).

If the expected behavior of the analyzer is that the following diff requires that `System.Complex` be annotated as `[RequiredPreviewFeatures]` then I do not believe the analyzer is of any use:
```diff
  public struct Complex : IEquatable<Complex>
+ , IAdditionOperators<Complex>
  {
  }
```

I believe this is a core scenario and we cannot expect users to make the analyzer work by either:
A) Marking existing non-preview types `RequiresPreviewFeatures`
B) Pragma suppressing the analyzer

In the former case (A), this means users cannot implement preview interfaces on existing types, which defeats the purpose of the attribute.
In the latter case (B), this means that users have to suppress the analyzer to target a core scenario we have.

jeffhandley: I agree that option (A) there cannot be considered--we cannot apply the attribute to existing types.

Option (B), pragma suppressing the analyzer is the explicit gesture of "I know I'm referencing a preview type here" though, and we need explicit gestures. We have plenty of dials for the suppressions too. They can be suppressed around individual references, blocks of code, entire files, projects, or solutions. And, unless I'm mistaken, through config can be dialed down from error to warning or even info.  With all of these options, we have to trust our users to choose the level of suppression they are most comfortable with.

> I think it is the exact opposite of how this will get used in practice and will result in a longer tail of bugs due to users having to suppress the analyzer for basic usages in practice.

I believe you're referring to bugs of over-suppressing, and later finding that we've allowed preview API usage to leak through in an implementation detail, or worse, an API definition; is that correct?

Having logic that attempts to minimize the suppressing users do requires _us_ to essentially auto-suppress scenarios on users' behalf.  As is evidenced by the quickly-identified false-negatives that @stephentoub called out, this logic incurs risk of users allowing preview API usage to leak through in implementation details as well as in API definitions--and because of logic bugs in the analyzer, vs. over-suppressing by users. I believe that to be a far worse situation for multiple reasons--one of which being that plugging a false-negative hole would be a build-time breaking change since we'd be introducing new build errors.

Altogether, I think we have _room for error_ in suppressing in the codebase, but we have _zero room for error_ suppressing within the analyzer logic.

tannergooding: >  pragma suppressing the analyzer is the explicit gesture of "I know I'm referencing a preview type here" though,
> Having logic that attempts to minimize the suppressing users do requires us to essentially auto-suppress scenarios on users' behalf.

I don't agree. The analyzer is meant to tell users: "You are using or exposing a preview feature and are not yourself annotated as preview". The expected user action here is to annotate the relevant parts as preview using the `RequiresPreviewFeatures` attribute.

Suppression, in general, is a user saying something like "this warning is wrong" or "I know what I'm doing". It should not be a normal or expected action and should not be required to use the feature in "core" scenarios.

> They can be suppressed around individual references, blocks of code, entire files, projects, or solutions

I believe this only works when you have logical hierarchies where a given level has clear semantics on how it impacts types. I don't think suppressions fit this bill.

For example, when you look at attributes:
* Placing it at the assembly level impacts everything in the assembly
* Placing it at the type level impacts everything in the type
* Placing it at the member level impacts the member

Therefore, if you want to treat the entire assembly as preview you put it at the assembly level. If you want to treat the entire type as preview you put it at the type level, etc.

If you take the same logic with regards to suppressions then suppressing at a given level means that the warning is invalid or should be ignored at that level. This means that suppressing at the type level, such as because you implement a preview interface, means that nothing in the type should be considered "preview" which adds risk and can lead to bugs on many levels. This, to me, means it is explicitly negative value to do so and should not be expected outside of extreme scenarios and is worse than having an analyzer that gets improved over time and which may introduce additional warnings in future SDKs. We already have plenty of analyzers that do change from release to release and which catch new issues. Likewise, it is realistically only around preview features in which case users should already be expected potential breaks when changing SDK versions or dependencies (which includes build time dependencies, such as analyzers or source generators).

---------------

For cases like `Int32`, this means that we need a mechanism through which users can successfully implement a preview interface without also annotating or suppressing the entire type. The normal scenario, as I'd expect, it is that the user can explicitly annotate each member coming from the interface. This works well, gives users a normal/expected scenario for the type, and avoids any issues.

If you have an interface which defines some members that are already public/exposed, you can avoid suppression by explicitly implementing the interface.

The remaining case is that of marker interfaces. Marker interfaces are already fairly rare but a decision has to be made around whether they need the type to be marked preview or only if usages/casts to that type need to be preview. In the former case (the type must be preview because no methods exist to be annotated) users have to suppress if they implement a marker interface. In the latter, you can freely implement a marker interface without being preview but need to be preview anywhere you directly handle such an interface. I believe the latter most closely matches the behavior of the first case (non marker interfaces).

We then have a couple remaining scenarios:
* Preview members
* Abstract classes

For preview members, its relatively straightforward. They are marked preview and you must be preview to invoke/access them.

However, abstract classes are a bit different due to how they are represented in the type hierarchy and how they are exposed to the user. They can contain fields, have members that are directly accessible/exposed, etc. They are distinct from interfaces and so it makes sense, IMO, for these to require the inheriting type to be marked as preview if they themselves are preview.

All of this also fits in with the normal "breaking change" rules for types and adding new members. If the entire type is preview, its safe, otherwise you have to consider the break as "normal" and that things like adding new abstract methods to an existing abstract class is "breaking", even if that single member is marked `Preview` (and same with going from non-preview to `Preview` for an existing type).

jeffhandley: Thanks, @tannergooding. This is good input, and I appreciate where you're coming from and what your goals are, which I believe include the following:

1. Minimize the likelihood of consumers of preview APIs from inadvertently taking dependencies on them in non-preview features
2. Balance the annoyance vs. value of the analyzer
3. Ensure that we ourselves are responsible users of the analyzer/attributes/suppression and that we protect ourselves from leaking preview features

> Suppression, in general, is a user saying something like "this warning is wrong" or "I know what I'm doing". It should not be a normal or expected action and should not be required to use the feature in "core" scenarios.

I believe most of the disagreement here circles around the "I know what I'm doing" category and where that intersects with a "core scenario." For ourselves, we have preview feature usage in core scenarios where naïve analyzer would lead us to suppress, because _we know what we're doing_. A for suppressing (`#pragma warning disable` style) is when we are indicating that a type (`Int32`) implements a preview interface. It boils down to whether or not we should need a `#pragma warning disable` around the interface name.

Interestingly, those exact same code blocks already have `#pragma warning disable SA1001` around them because our commas are formatted differently from usual. That's a case of "I know what I'm doing" and it's OK. Suppressing diagnostics from the preview feature analyzer is perfectly acceptable when we know what we're doing and it's a core scenario for us.

> I believe this only works when you have logical hierarchies where a given level has clear semantics on how it impacts types. I don't think suppressions fit this bill.

I understand your remarks here, that suppressing a broader scopes leads to more likelihood of over-suppressing and introducing issues in the future. I don't discount that. My point with suppression being available at all of those scopes is that _users can choose for themselves what is most appropriate for them_; they have dials to turn. If they are getting tired of wrapping lots of tiny code blocks in `#pragma warning disable`, then they can choose to rearrange their code to reduce the number of `#pragma` statements. They can find the balance that best suits their code cleanliness and protection against leaking preview features.  Ourselves, we'd likely opt for very granular `#pragma` statements to keep our risk minimized, as we have a high tolerance for the statements in our codebase.

The breaking change point I was trying to make is that if the following occurs:

1. We "auto-suppress" a diagnostic because we think it would be too noisy for our users
2. We later find that there are cases that this auto-suppression can lead to usage of preview APIs without diagnostics ever occurring
3. We remove the auto-suppression
4. This is now a source breaking change that must be treated as a breaking change -- we do this with other analyzers too

In that case, we've harmed the user first by letting them leak a preview API, and then again by introducing a source breaking change.  This is a troublesome direction to need to go.  We can also become more lenient in the future if enough customer reports come in that the analyzer is too noisy, but we can't go in the other direction.  We should start off by over-diagnosing though, to make sure we're protecting our users and giving them control over how they want to best suppress for their scenarios.

I also want to echo @stephentoub's remarks about the predictability.  If a user references a preview marker interface and gets a diagnostic, they'll recognize they need to handle that scenario (mark themselves or suppress). And then when they reference a non-marker preview interface but they don't get a diagnostic, they won't know why and this can lead to confusion and doubt. The behavior needs to be as simple and predictable as possible--the less nuance the better.  Again, we can also introduce nuance in the future if it seems justified by customer feedback, but I don't think we should start there.

tannergooding: >  And then when they reference a non-marker preview interface but they don't get a diagnostic, they won't know why and this can lead to confusion and doubt

This isn't a scenario. They will still get a diagnostic, just in a different location. It can even be in the same location. The main point is that you should not be required to suppress to deal with this scenario.

I believe that implementing a non-marker preview interface on a non-preview type is a "core scenario". Not just for us, but for generic math and the `RequiresPreviewFeatures` attribute in general. As such, suppressing should not be the only solution to this problem. It should be equally valid to deal with the warning by explicitly placing `RequiresPreviewFeatures` on every member implemented for the interface (whether implicit or explicitly implemented).

terrajobst: I suggest we stop this threadzilla and chat about this on Thursday in our meeting. Otherwise this will keep going :-)

pgovind:
Closing this draft. The discussion here is not relevant anymore. We now tag all implementations of preview interfaces at the type definition itself.

PR #5275, opened 2021-07-22
Split throw and yield parts of a function to avoid laziness issue

Evangelink:


src/Utilities/Compiler/Extensions/IEnumerableExtensions.cs
```diff
@@ -18,12 +18,17 @@ public static IEnumerable<T> Concat<T>(this IEnumerable<T> source, T value)
                 throw new ArgumentNullException(nameof(source));
             }
 
-            foreach (T v in source)
+            return ConcatImpl(source, value);
```
Evangelink, line 21: With the previous implementation the null check wasn't performed before any actual iteration of the enumerable was performed.

Evangelink:
Tagging @sharwell for a feedback.

Evangelink:
Ping @mavasani

sharwell (commented):
Can we add tests for this?

Evangelink:
Sure @sharwell, I will add it tomorrow.

PR #5277, opened 2021-07-22
Simplify linq expressions

Evangelink:


Youssef1313 (commented):
Is there an analyzer for that to enable?

Evangelink:
In this case I was using one of the rule I wrote back at [SonarSource](https://github.com/SonarSource/sonar-dotnet). I probably need to rewrite some of the "helpful" rules in roslyn-analyzer to have more potential adoption and also to be able to use operations and other features like this.

Youssef1313:
@Evangelink Actually there is IDE0120 for that. https://github.com/dotnet/roslyn/blob/main/src/Analyzers/CSharp/Tests/SimplifyLinqExpression/CSharpSimplifyLinqExpressionTests.cs

Evangelink:
@Youssef1313 interesting! I need to check how the rule is enabled (maybe it raises and I simply missed it).

Evangelink:
@Youssef1313 In VS 2019 I can't have the issue triggering but it works well in VS 2022.

Ping @mavasani for review

PR #5281, opened 2021-07-23
Fix a bunch of RS1024 on Equals

Evangelink:
Not sure why but I have started to get hundreds of warning related to `RS1024` in the codebase so I decided to start tackling them.

In this PR, I am focusing on the `Equals` because for some reason most of the complains on collections don't get fixed after applying the comparer.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpRegisterActionAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpRegisterActionAnalyzer.cs
@@ -77,10 +77,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Analyzers.MetaAnalyzers
             }
 
             protected override bool IsSyntaxKind(ITypeSymbol type)
-            {
-                return (_csharpSyntaxKind != null && SymbolEqualityComparer.Default.Equals(type, _csharpSyntaxKind)) ||
-                    (_basicSyntaxKind != null && SymbolEqualityComparer.Default.Equals(type, _basicSyntaxKind));
-            }
+                => SymbolEqualityComparer.Default.Equals(type, _csharpSyntaxKind)
+                    || SymbolEqualityComparer.Default.Equals(type, _basicSyntaxKind);
         }
     }
 }
--- a/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpRegisterActionAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpRegisterActionAnalyzer.cs
@@ -78,8 +78,8 @@ namespace Microsoft.CodeAnalysis.CSharp.Analyzers.MetaAnalyzers
 
             protected override bool IsSyntaxKind(ITypeSymbol type)
             {
-                return (_csharpSyntaxKind != null && type.Equals(_csharpSyntaxKind)) ||
-                    (_basicSyntaxKind != null && type.Equals(_basicSyntaxKind));
+                return (_csharpSyntaxKind != null && SymbolEqualityComparer.Default.Equals(type, _csharpSyntaxKind)) ||
+                    (_basicSyntaxKind != null && SymbolEqualityComparer.Default.Equals(type, _basicSyntaxKind));
             }
         }
     }
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
@@ -240,7 +240,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
                     {
                         AnalyzeFixerWithFixAll(fixer, context);
                     }
-                    else if (fixer.BaseType != null && fixer.BaseType.Equals(_codeFixProviderSymbol))
+                    else if (fixer.BaseType != null && SymbolEqualityComparer.Default.Equals(fixer.BaseType, _codeFixProviderSymbol))
                     {
                         Diagnostic diagnostic = fixer.CreateDiagnostic(OverrideGetFixAllProviderRule, fixer.Name);
                         context.ReportDiagnostic(diagnostic);
@@ -254,7 +254,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
                 {
                     foreach (INamedTypeSymbol type in fixer.GetBaseTypesAndThis())
                     {
-                        if (!type.Equals(_codeFixProviderSymbol))
+                        if (!SymbolEqualityComparer.Default.Equals(type, _codeFixProviderSymbol))
                         {
                             IMethodSymbol getFixAllProviderMethod = type.GetMembers(GetFixAllProviderMethodName).OfType<IMethodSymbol>().FirstOrDefault();
                             if (getFixAllProviderMethod != null && getFixAllProviderMethod.IsOverride)
@@ -278,7 +278,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
 
                     foreach (var argument in invocation.Arguments)
                     {
-                        if (argument.Parameter.Equals(param))
+                        if (SymbolEqualityComparer.Default.Equals(argument.Parameter, param))
                         {
                             return argument.Value.ConstantValue.HasValue && argument.Value.ConstantValue.Value == null;
                         }
@@ -323,13 +323,13 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
                     // Local functions
                     bool IsCodeActionWithOverriddenEquivalenceKey(INamedTypeSymbol namedType)
                     {
-                        if (namedType == null || namedType.Equals(_codeActionSymbol))
+                        if (namedType == null || SymbolEqualityComparer.Default.Equals(namedType, _codeActionSymbol))
                         {
                             return false;
                         }
 
                         // We are already tracking CodeActions with equivalence key in this compilation.
-                        if (namedType.ContainingAssembly.Equals(_sourceAssembly))
+                        if (SymbolEqualityComparer.Default.Equals(namedType.ContainingAssembly, _sourceAssembly))
                         {
                             return _codeActionsWithEquivalenceKey != null && _codeActionsWithEquivalenceKey.Contains(namedType);
                         }
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
@@ -240,7 +240,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
                     {
                         AnalyzeFixerWithFixAll(fixer, context);
                     }
-                    else if (fixer.BaseType != null && SymbolEqualityComparer.Default.Equals(fixer.BaseType, _codeFixProviderSymbol))
+                    else if (SymbolEqualityComparer.Default.Equals(fixer.BaseType, _codeFixProviderSymbol))
                     {
                         Diagnostic diagnostic = fixer.CreateDiagnostic(OverrideGetFixAllProviderRule, fixer.Name);
                         context.ReportDiagnostic(diagnostic);
@@ -323,7 +323,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.FixAnalyzers
                     // Local functions
                     bool IsCodeActionWithOverriddenEquivalenceKey(INamedTypeSymbol namedType)
                     {
-                        if (namedType == null || SymbolEqualityComparer.Default.Equals(namedType, _codeActionSymbol))
+                        if (SymbolEqualityComparer.Default.Equals(namedType, _codeActionSymbol))
                         {
                             return false;
                         }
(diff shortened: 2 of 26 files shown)


src/Utilities/Compiler/SymbolEqualityComparer.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Microsoft.CodeAnalysis
+{
+#if !CODEANALYSIS_V3_OR_BETTER
```
Evangelink, line 5: I had some exceptions saying cannot find `SymbolEqualityComparer` which made me realize that it's available only since v3. Adding a class matching the symbol to fallback to object equality seems the simplest solution as it will prevent pragmas everywhere in the code and make it simple when we drop support for older version to "bump" to the new version.

I am not entirely sure why but this symbol is not giving me what I would except so I end up with some warnings telling me that the symbol exists in 2 locations.

Any idea which compilation symbol to use to have the "proper" fix?

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -30,6 +30,9 @@ private WellKnownTypeProvider(Compilation compilation)
                 () =>
                 {
                     return ImmutableHashSet.Create<IAssemblySymbol>(
+#if CODEANALYSIS_V3_OR_BETTER
```
Evangelink, line 33: Need to revert this change to use the added class.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -200,7 +203,11 @@ internal bool IsTaskOfType([NotNullWhen(returnValue: true)] ITypeSymbol? typeSym
         {
             return typeSymbol != null
                 && typeSymbol.OriginalDefinition != null
-                && typeSymbol.OriginalDefinition.Equals(GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTask1))
+                &&
+#if CODEANALYSIS_V3_OR_BETTER
```
Evangelink, line 207: Need to revert this change to use the added class.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -496,7 +500,7 @@ public static bool IsCollectionAddMethod(this IMethodSymbol method, ImmutableHas
         /// <param name="taskType">Task type.</param>
         public static bool IsTaskFromResultMethod(this IMethodSymbol method, [NotNullWhen(returnValue: true)] INamedTypeSymbol? taskType)
             => method.Name.Equals("FromResult", StringComparison.Ordinal) &&
-               method.ContainingType.Equals(taskType);
+               SymbolEqualityComparer.Default.Equals(method.ContainingType, taskType);
```
Evangelink, line 503: Shall we check the `OriginalDefinition` here?

src/Microsoft.CodeAnalysis.Analyzers/CSharp/MetaAnalyzers/CSharpRegisterActionAnalyzer.cs
```diff
@@ -78,8 +78,8 @@ protected override SyntaxNode GetInvocationExpression(InvocationExpressionSyntax
 
             protected override bool IsSyntaxKind(ITypeSymbol type)
             {
-                return (_csharpSyntaxKind != null && type.Equals(_csharpSyntaxKind)) ||
-                    (_basicSyntaxKind != null && type.Equals(_basicSyntaxKind));
+                return (_csharpSyntaxKind != null && SymbolEqualityComparer.Default.Equals(type, _csharpSyntaxKind)) ||
+                    (_basicSyntaxKind != null && SymbolEqualityComparer.Default.Equals(type, _basicSyntaxKind));
```
Youssef1313, line 82: I believe the null check is redundant

src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
```diff
@@ -248,7 +248,7 @@ internal void CompilationEnd(CompilationAnalysisContext context)
                     {
                         AnalyzeFixerWithFixAll(fixer, context);
                     }
-                    else if (fixer.BaseType != null && fixer.BaseType.Equals(_codeFixProviderSymbol))
+                    else if (fixer.BaseType != null && SymbolEqualityComparer.Default.Equals(fixer.BaseType, _codeFixProviderSymbol))
```
Youssef1313, line 251: The null check was previously needed, but now it's redundant.

src/Microsoft.CodeAnalysis.Analyzers/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs
```diff
@@ -335,13 +335,13 @@ bool IsViolatingCodeActionObjectCreation(IObjectCreationOperation objectCreation
                     // Local functions
                     bool IsCodeActionWithOverriddenEquivalenceKey(INamedTypeSymbol namedType)
                     {
-                        if (namedType == null || namedType.Equals(_codeActionSymbol))
+                        if (namedType == null || SymbolEqualityComparer.Default.Equals(namedType, _codeActionSymbol))
```
Youssef1313, line 338: Null check is now redundant

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/CompareSymbolsCorrectlyAnalyzer.cs
```diff
@@ -269,7 +269,7 @@ private static bool IsSymbolType(ITypeSymbol typeSymbol, INamedTypeSymbol symbol
                 return false;
             }
 
-            if (typeSymbol.Equals(symbolType))
+            if (SymbolEqualityComparer.Default.Equals(typeSymbol, symbolType))
```
Youssef1313, line 272: null check above is now redundant.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -213,7 +214,9 @@ public static bool IsAsyncDisposeImplementation([NotNullWhen(returnValue: true)]
 
             // Identify the implementor of IAsyncDisposable.Dispose in the given method's containing type and check
             // if it is the given method.
-            return method.ReturnType != null && method.ReturnType.Equals(valueTaskType) && method.Parameters.IsEmpty &&
+            return method.ReturnType != null &&
```
Youssef1313, line 217: null check is now redundant.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
Youssef1313: > Shall we check the `OriginalDefinition` here?

This is the non-generic task type. So I think the way it's currently written is okay.

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -713,7 +713,7 @@ public static IEnumerable<AttributeData> GetAttributes(this ISymbol symbol, INam
                 return Enumerable.Empty<AttributeData>();
             }
 
-            return symbol.GetAttributes().Where(attr => attr.AttributeClass.Equals(attributeType));
+            return symbol.GetAttributes().Where(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, attributeType));
```
Youssef1313, line 716: You can use HasAttribute extension method here instead.

src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
```diff
@@ -53,10 +53,16 @@ static bool IsSingleStatementBody(IBlockOperation body)
                     descendants[0] is IThrowOperation throwOperation &&
                     throwOperation.GetThrownExceptionType() is ITypeSymbol createdExceptionType)
                 {
-                    if (Equals(context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNotImplementedException), createdExceptionType.OriginalDefinition) ||
-                        Equals(context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNotSupportedException), createdExceptionType.OriginalDefinition) ||
+                    if (SymbolEqualityComparer.Default.Equals(
+                            context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemNotImplementedException),
```
Youssef1313, line 57: Out of scope for this PR, but I'm worried about the performance when `GetOrCreateTypeByMetadataName` is called multiple times in this extension.

I think it may be better to get the symbols once at compilation start for the analyzers using this extension, and pass them to the method.

src/Utilities/Compiler/WellKnownTypeProvider.cs
```diff
@@ -200,7 +203,11 @@ internal bool IsTaskOfType([NotNullWhen(returnValue: true)] ITypeSymbol? typeSym
...
+                    SymbolEqualityComparer.Default.
+#endif
+                    Equals(typeSymbol.OriginalDefinition, GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemThreadingTasksTask1))
```
Youssef1313, line 210: Same concern about performance if the extension is called multiple times in the same analyzer.

src/Utilities/Compiler/SymbolEqualityComparer.cs
Youssef1313: > I am not entirely sure why but this symbol is not giving me what I would except so I end up with some warnings telling me that the symbol exists in 2 locations.


@Evangelink I'm not seeing anything about that in CI.

src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -18,7 +18,7 @@
 
 namespace Analyzer.Utilities.Extensions
 {
-    internal static class IMethodSymbolExtensions
+    internal static partial class IMethodSymbolExtensions
```
Youssef1313, line 21: Why this change?

src/Utilities/Compiler/SymbolEqualityComparer.cs
```diff
@@ -0,0 +1,21 @@
...
+    internal class SymbolEqualityComparer
```
Youssef1313, line 6:
```suggestion
    internal sealed class SymbolEqualityComparer : IEqualityComparer<ISymbol?>
```

src/Utilities/Compiler/SymbolEqualityComparer.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        private SymbolEqualityComparer()
+        {
+
+        }
+
+#pragma warning disable CA1822 // Mark members as static
+        public bool Equals(ISymbol? symbol1, ISymbol? symbol2) =>
+#pragma warning restore CA1822 // Mark members as static
+            object.Equals(symbol1, symbol2);
```
Youssef1313, line 16: To be more matching with roslyn implementation

```suggestion
        public bool Equals(ISymbol? x, ISymbol? y)
        {
            if (x is null)
            {
                return y is null;
            }

            return x.Equals(y);
        }

        public int GetHashCode(ISymbol? obj)
        {
            return obj?.GetHashCode() ?? 0;
        }
```

Evangelink: Thanks!

src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
Evangelink: I think that perf should be ok because we are doing a dictionary call (if it exists we return the value otherwise we create, store and return). We implemented this helper especially to help with performance issue.

src/Utilities/Compiler/SymbolEqualityComparer.cs
Evangelink: I can no longer reproduce the issue. I am wondering if it was just some kind of non-refresh from VS.

Evangelink:
Ping @mavasani

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
Evangelink: Just had a look and I can't make this replacement here.

src/Utilities/Compiler/Extensions/OperationBlockAnalysisContextExtension.cs
mavasani: Both of you are right to an extent - perf is much better even inside operation callbacks because it is a dictionary, however we still prefer to collect all well known symbols once at compilation start and pass them around, primary for consistency in our code base and also avoid large number of operation callbacks trying to access the same dictionary when multiple analyzers are executing in parallel.

mavasani: Agreed this is our of scope for this PR though.

src/Utilities/Compiler/SymbolEqualityComparer.cs
```diff
@@ -0,0 +1,26 @@
...
+#if !CODEANALYSIS_V3_OR_BETTER
+namespace Microsoft.CodeAnalysis
+{
+    using System.Collections.Generic;
+
+    internal sealed class SymbolEqualityComparer : IEqualityComparer<ISymbol?>
+    {
+        private SymbolEqualityComparer()
+        {
+
```
mavasani, line 12:
```suggestion
```

mavasani (approved):

src/Utilities/Compiler/SymbolEqualityComparer.cs
mavasani: @Evangelink going to commit this minor style change and merge your PR.

Evangelink:
Thanks!

PR #5295, opened 2021-07-26
Remove unused generic types

Evangelink:


src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.Fixer.cs
```diff
@@ -4,15 +4,14 @@
...
 namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
 {
     /// <summary>
     /// CA2207: Initialize value type static fields inline
     /// </summary>
     [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
-    public class CSharpInitializeStaticFieldsInlineFixer : InitializeStaticFieldsInlineFixer<SyntaxKind>
+    public class CSharpInitializeStaticFieldsInlineFixer : InitializeStaticFieldsInlineFixer
     {
     }
```
Youssef1313, line 16: I think this can be completely deleted and have only one fixer in `Core` layer.

mavasani:
@Evangelink Did you use any analyzer to identify the unused generic type parameters? If so, I am wondering if the analyzer can be enabled for the repo as a build warning that is caught in CI.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpInitializeStaticFieldsInline.Fixer.cs
mavasani: Agreed.

Evangelink:
@mavasani I haven't but I assume we could probably write such analyzer. Want me to have a look at it?

PR #5296, opened 2021-07-26
Create non-generic base class for static fields

Evangelink:
Static fields in a generic type are actually not shared across instances. To remedy this issue, I have introduced an upper, non-generic, class to properly share them.


sharwell (changes requested):
I don't think we need to take this change. Only one instantiation of these types ever exists for a given language, and in the worst case there are only two supported languages in use.

Evangelink:
Do you think it will bring un-needed complexity?

sharwell:
I do. Currently there is one base type in one file, but after this change there will be two base types in two different files¹ making it more difficult to figure out which state goes in which location. We also don't have any static analysis or testing support to avoid regressions related to this change.

¹ I didn't ask for the files to be split because splitting them is unnecessary work if the plan is to close the PR anyway.

Evangelink:
Explanation makes sense, I will close the PR.

Out of curiosity, would it make sense to suggest the analyzer (raises on `static` fields of generic types)?

sharwell:
We have [CA1000](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1000) but it was noisy and I believe it's either disabled by default or a candidate for removal.

PR #5297, opened 2021-07-26
Make known tags readonly

Evangelink:


src/Utilities/Compiler/Extensions/WellKnownDiagnosticTagsExtensions.cs
```diff
@@ -7,11 +7,11 @@ internal static class WellKnownDiagnosticTagsExtensions
         public const string EnabledRuleInAggressiveMode = nameof(EnabledRuleInAggressiveMode);
         public const string Dataflow = nameof(Dataflow);
         public const string CompilationEnd = nameof(CompilationEnd);
-        public static string[] DataflowAndTelemetry = new string[] { Dataflow, WellKnownDiagnosticTags.Telemetry };
-        public static string[] DataflowAndTelemetryEnabledInAggressiveMode = new string[] { Dataflow, WellKnownDiagnosticTags.Telemetry, EnabledRuleInAggressiveMode };
+        public static readonly string[] DataflowAndTelemetry = new string[] { Dataflow, WellKnownDiagnosticTags.Telemetry };
```
Youssef1313, line 10: Unrelated: I'm not a big fan of exposing public arrays. There is no guarantee that it cannot be modified 😄 

That's totally out of scope of this PR. Just writing my thoughts 😄

Youssef1313 (approved):

src/Utilities/Compiler/Extensions/WellKnownDiagnosticTagsExtensions.cs
Evangelink: That's a shared feeling, afaik `ImmutableArray` has similar perfs as array so might be worth to redo this.

drewnoakes: Interestingly, if `ImmutableArray` is not yet a `readonly struct` then making these fields `readonly ImmutableArray<string>` would lead to defensive copies of the struct.

Evangelink: @drewnoakes thanks! It seems it is not yet https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.cs#L15

Youssef1313: @Evangelink corefx repo is archived and the codebase is now maintained in dotnet/runtime. It's marked `readonly` as of https://github.com/dotnet/runtime/pull/44640. I don't know which version is used in the repo though.

Evangelink: Stupid me, got fooled by first google result.

Evangelink:
Ping @mavasani

mavasani (approved):

PR #5302, opened 2021-07-26
Use vs-threading analyzers to enforce Async naming convention

sharwell:


src/PerformanceTests/Tests/Enabled/CSharp_CA1416.cs
```diff
@@ -1,5 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+#pragma warning disable VSTHRD200 // Use "Async" suffix for async methods
```
Evangelink, line 3: I am wondering if it would be worth to have a specific `.editorconfig` in the `PerformanceTests` section.

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1685,6 +1685,7 @@ public C(int value)
             await VerifyCSharpAdditionalFileFixAsync(source, shippedText, unshippedText, fixedUnshippedText);
         }
 
+        [Fact]
```
Evangelink, line 1688: Nice!

Shall it be moved to a separate PR?

src/Roslyn.Diagnostics.Analyzers/Core/RelaxTestNamingSuppressor.cs
```diff
@@ -27,7 +27,9 @@ public sealed class RelaxTestNamingSuppressor : DiagnosticSuppressor
 
         public override void ReportSuppressions(SuppressionAnalysisContext context)
         {
-            if (context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitFactAttribute) is not { } factAttribute)
+            context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.XunitFactAttribute, out var factAttribute);
+            context.Compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.BenchmarkDotNetAttributesBenchmarkAttribute, out var benchmarkAttribute);
```
Evangelink, line 31: Shall it be moved to a separate PR?

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
sharwell: No, this fixes the VSTHRD200 violation on the method.

src/PerformanceTests/Tests/Enabled/CSharp_CA1416.cs
sharwell: ➡️ These will become unnecessary after Roslyn.Diagnostics.Analyzers is updated and the new RelaxNamingSuppressor rules take effect.

src/Roslyn.Diagnostics.Analyzers/Core/RelaxTestNamingSuppressor.cs
sharwell: ➡️ This one _could_ be moved to a different PR, but this non-production package is the analyzers specifically created for dealing with scenarios like this.

Evangelink (approved):

src/Roslyn.Diagnostics.Analyzers/Core/RelaxTestNamingSuppressor.cs
sharwell: @jmarolf do you have an order preference here?

jmarolf: I am fine with this change being a part of this PR

jmarolf (approved):
:shipit:

src/Roslyn.Diagnostics.Analyzers/Core/RelaxTestNamingSuppressor.cs
sharwell: Thanks, I'll revert 07c9509 as part of the analyzer upgrade after the signed build is published.

PR #5303, opened 2021-07-26
Clarify that RuleLevel.BuildError is only valid for source generators

sharwell:


Youssef1313:
@sharwell It was added specifically for an analyzer. So I think it was left unused *unintentionally*

src/Utilities/Compiler/RuleLevel.cs
```diff
@@ -11,9 +11,15 @@ internal enum RuleLevel
...
         /// <summary>
-        /// Correctness rule which should have <b>no false positives</b>, and is required to be fixed by users.
-        /// This rule will be <b>enabled in CI and IDE live analysis</b> by default with severity <see cref="DiagnosticSeverity.Error"/>.
+        /// Correctness rule which prevents the compiler from producing a well-defined output binary, and must have
+        /// <b>no false positives</b>. Violations of this rule must be fixed by users before any other testing work can
+        /// continue. This rule will be <b>enabled in CI and IDE live analysis</b> by default with severity
+        /// <see cref="DiagnosticSeverity.Error"/>.
         /// </summary>
+        /// <remarks>
+        /// Since analyzers cannot directly influence output binaries, this value is only valid in the implementation of
```
pgovind, line 20: Only thing I'd note here is that the `DetectPreviewFeatures` analyzer will end up using `BuildError` (it is not a source generator) eventually. I don't think we need to call it out here or make any changes to the wording. As long as everyone is aware of it I'm fine with the wording here.

sharwell: This section was specifically intended to clarify that `DetectPreviewFeaturesAnalyzer` will use `BuildWarning`, not `BuildError`.

Youssef1313: > This section was specifically intended to clarify that `DetectPreviewFeaturesAnalyzer` will use `BuildWarning`, not `BuildError`.

@sharwell In that case, should it still report diagnostics for generated code?

sharwell: There is no direct impact here on decisions to analyze and/or report diagnostics on generated code.

terrajobst:
I do not believe that disallowing analyzers from using `error` as the severity is wise. There are examples, like the [preview feature analyzer](https://github.com/dotnet/designs/blob/main/accepted/2021/preview-features/preview-features.md) where we plan on using `error`.

I believe it's fair to say that the bar for `error` needs to be very high because it prevents people from producing a successful build. But ultimately the point of analyzers is to allow domain-specific knowledge to be injected into the compiler pipeline, so what constitutes a correctness issue may result outside the compilation (for example, consider analyzers that flag the use of configuration settings that would block deployment; marking those as errors might still be valid because for all intents and purposes they are preventing the application to actually work).

mavasani (changes requested):
We now do have a CA analyzer with default severity Error, so the documentation would need to be different.

sharwell:
@mavasani The working has already been updated to account for that analyzer (see the second commit of this pull request).

jmarolf (approved):

PR #5312, opened 2021-07-28
Ensure double cast in int-int division

Evangelink:
Just noticed a couple of places where we are doing divisions of 2 `int` members.

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
@@ -503,7 +503,7 @@ class C4 { public static C2 MyC2 = null; }
 ";
 
             var expectedMetricsText = @"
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
```diff
@@ -503,7 +503,7 @@ class C4 { public static C2 MyC2 = null; }
 ";
 
             var expectedMetricsText = @"
-Assembly: (Lines: 12, ExecutableLines: 3, MntIndex: 95, CycCxty: 5, CoupledTypes: {C1, C2, C3, C4}, DepthInherit: 2)
+Assembly: (Lines: 12, ExecutableLines: 3, MntIndex: 96, CycCxty: 5, CoupledTypes: {C1, C2, C3, C4}, DepthInherit: 2)
```
Evangelink, line 506: I have replaced with the value given but I have no idea if the value makes more sense or not.

Given these changes, it may be better to wait for a major bump before merging except if we consider this was an important bug.

Evangelink:
Ping @mavasani

mavasani (approved):

PR #5315, opened 2021-07-28
Remove un-needed interface declaration

Evangelink:
`IReadonlyCollection` is inherited by `IReadonlyLsit` so there is no need for the extra declaration.


Youssef1313 (commented):
Same change can be done in roslyn repo as well:

https://github.com/dotnet/roslyn/blob/main/src/Dependencies/PooledObjects/ArrayBuilder.cs#L14

- Worth an analyzer reporting such redundant interface implementations?
- Are there cases when this actually can have meaningful value?

Evangelink:
> Same change can be done in roslyn repo as well:

Would it be worth sharing some utilities between roslyn and roslyn-analyzers rather than having duplicates that can go out of sync?

> * Worth an analyzer reporting such redundant interface implementations?

That would be easy to implement but I am not convinced that the value is high.

> * Are there cases when this actually can have meaningful value?

Don't know.

Evangelink:
Ping @mavasani

mavasani (approved):

PR #5327, opened 2021-07-29
Refactor tests to use non-obsolete location API

Evangelink:
Simple tests refactoring


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStaticTests.cs
```diff
@@ -26,49 +25,49 @@ public class MembersTests
     internal static int s_field;
     public const int Zero = 0;
 
-    public int Method1(string name)
+    public int {|#0:Method1|}(string name)
```
Youssef1313, line 28: It might be simpler to use `[|...|]` instead.

Evangelink: Yes but we should make sure to keep some where we test the arg.

sharwell (approved):

PR #5172, opened 2021-06-25
Fix CA1822 false positive with partial methods

Youssef1313:
Fixes #5168
Fixes #4733

PR #5333, opened 2021-07-30
Update CA1822 to handle local functions

### LINKED ISSUE #5331: Update CA1822 to analyze local functions
### Analyzer

**Diagnostic ID**: [CA1822](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1822)

### Describe the improvement

Extend the rule to analyze local functions and report a diagnostic on the ones that can be made `static`.

Evangelink:
Fixes #5331

### DIFF
--- a/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpMarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpMarkMembersAsStatic.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeQuality.Analyzers.QualityGuidelines;
+
+namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpMarkMembersAsStaticAnalyzer : MarkMembersAsStaticAnalyzer
+    {
+        protected override bool SupportStaticLocalFunctions(ParseOptions parseOptions)
+            => ((CSharpParseOptions)parseOptions).LanguageVersion >= LanguageVersion.CSharp8;
+    }
+}
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.Linq;
 using Analyzer.Utilities;
 using Analyzer.Utilities.Extensions;
@@ -15,8 +16,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
     /// <summary>
     /// CA1822: Mark members as static
     /// </summary>
-    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
-    public sealed class MarkMembersAsStaticAnalyzer : DiagnosticAnalyzer
+    public abstract class MarkMembersAsStaticAnalyzer : DiagnosticAnalyzer
     {
         internal const string RuleId = "CA1822";
 
@@ -34,9 +34,11 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                                                                              isPortedFxCopRule: true,
                                                                              isDataflowRule: false);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+        public sealed override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
 
-        public override void Initialize(AnalysisContext context)
+        protected abstract bool SupportStaticLocalFunctions(ParseOptions parseOptions);
+
+        public sealed override void Initialize(AnalysisContext context)
         {
             context.EnableConcurrentExecution();
 
@@ -59,7 +61,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
 
             return;
 
-            static void OnSymbolStart(
+            void OnSymbolStart(
                 SymbolStartAnalysisContext symbolStartContext,
                 WellKnownTypeProvider wellKnownTypeProvider,
                 ImmutableArray<INamedTypeSymbol> skippedAttributes,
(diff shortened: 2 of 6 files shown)


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -59,7 +61,7 @@ public override void Initialize(AnalysisContext context)
 
             return;
 
-            static void OnSymbolStart(
+            void OnSymbolStart(
```
Evangelink, line 64: Can no longer be static because of a call to `SupportStaticLocalFunctions`.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -91,57 +93,61 @@ void OnMethodReference(OperationAnalysisContext operationContext)
...
-                    }, OperationKind.InstanceReference);
+                        CheckHasInstanceReference(operationContext.Operation, ref isInstanceReferenced)
+                        , OperationKind.InstanceReference);
 
-                    // Workaround for https://github.com/dotnet/roslyn/issues/27564
                     blockStartContext.RegisterOperationAction(operationContext =>
+                        CheckHasReferenceInXml(operationContext.Operation, ref isInstanceReferenced)
+                        , OperationKind.None);
+
+                    if (SupportStaticLocalFunctions(blockStartContext.OperationBlocks[0].Syntax.SyntaxTree.Options))
```
Evangelink, line 114: Maybe this check should be done at the symbol level instead of at each block.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -273,21 +279,32 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
...
+        private static bool ShouldAnalyzeLocalFunction(IMethodSymbol methodSymbol, WellKnownTypeProvider wellKnownTypeProvider, OperationAnalysisContext context)
+        {
+            Debug.Assert(methodSymbol.MethodKind == MethodKind.LocalFunction);
+
+            if (methodSymbol.IsStatic || IsEventHandlerLike(methodSymbol, wellKnownTypeProvider))
+            {
+                return false;
+            }
+
+            var hasCorrectVisibility = context.Options.MatchesConfiguredVisibility(Rule, methodSymbol, wellKnownTypeProvider.Compilation,
```
Evangelink, line 307: Not sure if that's worth to handle. My idea was to say that if `private` are excluded we may want to also exclude local functions but that's debatable.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -273,21 +279,32 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
...
+            if (!hasCorrectVisibility)
             {
                 return false;
             }
 
-            var hasCorrectVisibility = blockStartContext.Options.MatchesConfiguredVisibility(Rule, methodSymbol, wellKnownTypeProvider.Compilation,
+            return true;
+        }
+
+        private static bool ShouldAnalyzeLocalFunction(IMethodSymbol methodSymbol, WellKnownTypeProvider wellKnownTypeProvider, OperationAnalysisContext context)
```
Evangelink, line 298: I started from the `ShouldAnalyze` function and removed all cases that can't (or at least I think) be applied to local functions.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Youssef1313: @Evangelink It actually can be done at compilation start, then pass the boolean value around

src/NetAnalyzers/CSharp/Microsoft.CodeQuality.Analyzers/QualityGuidelines/CSharpMarkMembersAsStatic.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeQuality.Analyzers.QualityGuidelines;
+
+namespace Microsoft.CodeQuality.CSharp.Analyzers.QualityGuidelines
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class CSharpMarkMembersAsStaticAnalyzer : MarkMembersAsStaticAnalyzer
+    {
+        protected override bool SupportStaticLocalFunctions(ParseOptions parseOptions)
+            => ((CSharpParseOptions)parseOptions).LanguageVersion >= LanguageVersion.CSharp8;
```
Youssef1313, line 14: Out of scope for the PR, but I'd love to have `ILanguageFeatureSupportService` with language specific implementations, and do something like `compilation.GetService<ILanguageFeatureSupportService>().IsFeatureSupported(LanguageFeature.StaticLocalFunctions);`

Benefits are:

- This will eliminate the need for language-specific classes for some analyzers.
- One central place to decide whether a feature is supported.

But I'm not sure if this is easily approachable, or if it's worth doing.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
```diff
@@ -273,21 +279,32 @@ void OnSymbolEnd(SymbolAnalysisContext symbolEndContext)
...
+        {
+            Debug.Assert(methodSymbol.MethodKind == MethodKind.LocalFunction);
+
+            if (methodSymbol.IsStatic || IsEventHandlerLike(methodSymbol, wellKnownTypeProvider))
+            {
+                return false;
+            }
```
Youssef1313, line 305: I'm surprised this works. The compiler *often* produces statics even if the local function wasn't declared static in source.

See [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEDMACdAmbAYQFgAoAbzO2tx3QBZsBZACgEoqbLSbfdGAMgHsAxhAA27bOWwBfTtXmlZQA=).

I was expecting that `IsStatic` is going to be `true`. But no action is needed if the test passes.

Youssef1313 (commented):
This is already handled by an IDE analyzer.

I think we should consider removing the IDExxxx in favor of this one.

https://github.com/dotnet/roslyn/blob/main/src/Analyzers/CSharp/Analyzers/MakeLocalFunctionStatic/MakeLocalFunctionStaticDiagnosticAnalyzer.cs

Youssef1313 (commented):
Consider adding a test for recursive local function (test stolen from roslyn):

```csharp
class C
{
    void M()
    {
        int [|fibonacci|](int n)
        {
            return n <= 1 ? n : fibonacci(n - 1) + fibonacci(n - 2);
        }
    }
}
```

Youssef1313 (commented):
Consider adding a test for the case in https://github.com/dotnet/roslyn/issues/46858

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/MarkMembersAsStatic.cs
Evangelink: Yes, I was just wondering if it could cause some problem with partial definitions.

Youssef1313: @Evangelink Partial definitions should never be analyzed, only partial implementations does. https://github.com/dotnet/roslyn-analyzers/pull/5172

Evangelink:
> This is already handled by an IDE analyzer.
> 
> I think we should consider removing the IDExxxx in favor of this one.

It's not the first time I see duplication between the analyzer and roslyn repos. Without blaming one side or the other, I am just wondering if we could find a way to have everything centralized in one or a couple of places:

1. Everything in `roslyn`
2. Compiler itself and maybe the integration into one repo and everything related to refactoring, issues, code-fixes... in another.

Youssef1313:
> > This is already handled by an IDE analyzer.
> > I think we should consider removing the IDExxxx in favor of this one.
> 
> It's not the first time I see duplication between the analyzer and roslyn repos. Without blaming one side or the other, I am just wondering if we could find a way to have everything centralized in one or a couple of places:
> 
> 1. Everything in `roslyn`
> 2. Compiler itself and maybe the integration into one repo and everything related to refactoring, issues, code-fixes... in another.

@mavasani @CyrusNajmabadi for thoughts.

mavasani:
I tend to agree with @Youssef1313 here. Even though both the analyzers have similar analysis, CA1822 is aimed at performance improvement and the IDE analyzer aimed at purely a style/functional improvement. I can see the other side of the argument as well, there seems to be a very thin line separating these analyzers and one may want a single analyzer/code fix/fix all to just ensure adding static wherever applicable.
Personally, I am fine if we take this PR or don't, I don't see too much push back from customers either way. @Evangelink thoughts?

Evangelink:
> The compiler _often_ produces statics even if the local function wasn't declared static in source.

Do we have any idea of how _often_ this is? 

My point being that if it's really likely (I don't know 80%+) of cases where the IL will be declared as _static_ then it's probably useless and the `IDE` rule would be enough. 
If the bar is low/lower then I would say, is there any chance to produce a less optimized code by having people adding the `static` keyword? If not, I don't see any reason not to want to enforce this.

I will wait for a decision to be taken here before fixing/changing target.

Youssef1313:
> Do we have any idea of how often this is?

I don't know, but I assume the compiler will do that whenever it can.


> is there any chance to produce a less optimized code by having people adding the static keyword? 

Adding static should never produce a less optimized code as far as I know. The whole point is the analyzer is complaining about "performance" issue while it's not a performance issue.

Could we have both analyzers here, but under different rules with different categories, and retire the IDExxxx one?

Evangelink:
> The whole point is the analyzer is complaining about "performance" issue while it's not a performance issue.

Totally fine with the fact that the compiler is doing its best to optimize but if there is no drawback and the compiler is not right enough cases then it is a mainly a performance improvement to add the `static` keyword.

Youssef1313:
@Evangelink Have you found any scenario where the compiler didn't add "static" when it could? In that case, I totally agree with the change as is. Otherwise, I lean towards having a different category for the local functions (I'm not sure if we can reuse the ID or not)

Evangelink:
> Have you found any scenario where the compiler didn't add "static" when it could

I have made many tests and I couldn't find a scenario where the `static` wasn't generated so I lean toward your suggestion.

Closing this PR and the ticket as won't fix.

> > This is already handled by an IDE analyzer.
> > I think we should consider removing the IDExxxx in favor of this one.
> 
> It's not the first time I see duplication between the analyzer and roslyn repos. Without blaming one side or the other, I am just wondering if we could find a way to have everything centralized in one or a couple of places:
> 
> 1. Everything in `roslyn`
> 2. Compiler itself and maybe the integration into one repo and everything related to refactoring, issues, code-fixes... in another.

@mavasani On that topic, could you discuss it?

mavasani:
> It's not the first time I see duplication between the analyzer and roslyn repos. 

I don't think there is an easy answer here or a clear set of work items to resolve this. As long as style and quality analyzers are in different repos, analyzers that have some overlap between style preference and quality improvement are likely to get some duplication, we try to minimize it, but it is not avoidable unless we can move both bucket of analyzers into a single repo. That has been discussed few times in the past but has huge costs and bunch of blocking issues, it is unlikely to be picked up anytime soon.

PR #5340, opened 2021-08-02
Look for PreviewFeatures on type definitions

pgovind:
Currently, we only report PreviewFeature diagnostics on method/property/field invocations on Preview types. There was some discussion that we want to move over to reporting preview feature usage everywhere, including type definitions. This PR implements that. Changes include:
1. Getting rid of the check for types implementing empty interfaces => All code/resources related to empty preview interfaces have been removed
2. A diagnostic for typeof(PreviewType) calls
3. Improvements to type definitions. In particular, if type definitions a) return or b) accept a preview type parameter, we now report a diagnostic.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -261,15 +329,15 @@ private static bool OperationUsesPreviewFeatures(OperationAnalysisContext contex
         {
             IOperation operation = context.Operation;
             ISymbol containingSymbol = context.ContainingSymbol;
-            if (OperationUsesPreviewFeatures(containingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol) ||
-                OperationUsesPreviewFeatures(containingSymbol.ContainingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
+            if (ProcessPreviewAttribute(containingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol) ||
```
pgovind, line 332: This is just a rename of OperationUsesPreviewFeatures -> ProcessPreviewAttribute. No functional changes

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
+        [Fact]
+        public async Task TestDerivedInterface()
+        {
+            var csInput = @" 
+using System.Runtime.Versioning; using System;
+namespace Preview_Feature_Scratch
+{" +
+@"
+
+    class Program
```
jmarolf, line 930: What is this class for in this test?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
+    {
+        static void Main(string[] args)
+        {
+            A<Foo> aFooInstance = {|#0:new A<Foo>()|}; // Can we have a property?
+        }
+    }
+
+class A<T> where T : IFoo, new()
```
jmarolf, line 972: Why no warning here?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
+}";
+
+            var test = TestCS(csInput);
+            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(0).WithArguments("B"));
+            await test.RunAsync();
+
+        }
+
+        [Fact]
+        public async Task TestDebug()
```
jmarolf, line 1110: I assume we want to take this out once everything is working?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
+            ((IZoo)b).Bar();
+        }
+    }
+
+    public class BImplementation : BAbstract
+    {
+
+    }
+
+    public abstract class BAbstract : IZoo
```
jmarolf, line 1158: Why not warn for every inheritor?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -210,15 +263,22 @@ private static bool SymbolIsStaticAndAbstract(ISymbol symbol, ConcurrentDictiona
         {
             ISymbol symbol = context.Symbol;
 
-            // Don't report diagnostics on symbols that are marked as Preview
-            if (ProcessPreviewAttribute(symbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
+            // Don't report diagnostics on symbols that are marked as Preview or if the ContainingType is Preview. We'll miss 2+ level nested type/method definitions, but that's ok. Only users who suppress the analyzer will run into this case.
```
jmarolf, line 266: While suppression of this analyzer is fine. I would expect the correct thing to do on the users part is also mark their types as preview. Becasue of this I would expect us to want to always find all locations

jmarolf:
Considering the change in scope I would expect the analyzer to operate like this
- find all preview annotated symbols in a compilation
- build a list of these "banned" symbols
- use IOperation to examine the compilation for violations.

This is essentially how the banned api analyzer operates: https://github.com/dotnet/roslyn-analyzers/tree/086d1cc888ca0d8c96b94444935f488fcaf3fe98/src/Microsoft.CodeAnalysis.BannedApiAnalyzers

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -285,21 +341,63 @@ private static bool OperationUsesPreviewFeatures(OperationAnalysisContext contex
...
+                {
+                    var attributeClassSymbol = attribute.AttributeClass;
+                    if (attributeClassSymbol.HasAttribute(previewFeatureAttribute))
+                    {
+                        requiresPreviewFeaturesSymbols.GetOrAdd(symbol, true);
+                        return true;
+                    }
+                }
+
+                requiresPreviewFeaturesSymbols.GetOrAdd(symbol, false);
```
buyaa-n, line 399: Isn't this and below should be `attributeClassSymbol` instead `symbol`?
```suggestion
                        requiresPreviewFeaturesSymbols.GetOrAdd(attributeClassSymbol, true);
                        return true;
                    }
                }

                requiresPreviewFeaturesSymbols.GetOrAdd(attributeClassSymbol, false);
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
         }
+
+        [Fact]
+        public async Task TestTypeOf()
+        {
+            var csInput = @" 
+using System.Runtime.Versioning; using System;
+namespace Preview_Feature_Scratch
+{" +
+@"
```
buyaa-n, line 902: Here and everywhere else why need to concatenate strings?
```suggestion
{
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            A aObject = {|#0:new A()|};
+        }
+    }
+
+[My]
+class {|#1:A|}
```
buyaa-n, line 1063: it shouldn't warn on the attribute usage?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +891,655 @@ static void Main(string[] args)
...
+namespace Preview_Feature_Scratch
+{" +
+@"
+
+    [RequiresPreviewFeatures]
+    class Program
+    {
+        class NestedClass
+        {
+            public bool {|#1:GenericMethod|}<T>()
```
buyaa-n, line 1472: Isn't it false positive

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: There used to be another variable being appended between the 2 parts. I got rid of that variable in the last PR, but just kept this around. Fixed now. Removed the `+`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: Because `IFoo` is not preview. Only `Foo` is

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: Sorry, looks like the last commit (that deleted this) didn't get pushed. This is fixed now.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: I think the rationale was that `IZoo` is not marked as Preview on purpose, so we assume that types deriving/implementing `IZoo` are not Preview. The design calls for 1 level look ups, so it's up to the author to get the attributes applied correctly at every level.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: I don't think so. We only 1 level look up for methods/properties, so any other level of nesting will produce the diagnostic. I also didn't spend too much time trying to optimize this since either we expect users to just opt into preview features at the assembly level.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: It should, but I don't know if it is even possible to create a diagnostic around `[My]` here. Created https://github.com/dotnet/roslyn-analyzers/issues/5346 to track it

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: I think this does find all locations right? All this (and the next) if block is doing is:

```csharp
    [RequiresPreviewFeatures]
    public class MarkedClass // Don't report a diagnostic here
    {
    }
```

When `MarkedClass` is being processed, we detect that it is already marked with `[RequiresPreviewAttribute]` and return. The next if block check the immediate parent when type members of Preview classes are processed. SOmething like:

```csharp
    [RequiresPreviewFeatures]
    public class MarkedPreviewClass
    {

        public MarkedPreviewClass() { } // Don't report a diagnostic here

        public void UnmarkedMethodInPreviewClass() { } // Don't report a diagnostic here

    }
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: This code block has been deleted now :)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
buyaa-n: For me this is a false positive, the parent class annotation should suppress this. Though I am not sure if suppressing could only one level up is by design cc @terrajobst @jeffhandley

jeffhandley: I do think we should look "all the way up" the hierarchy. If an API is contained in a preview feature scope, it too should be recognized as a preview feature, regardless of how deep into the hierarchy it is.

pgovind:
> Considering the change in scope I would expect the analyzer to operate like this
> 
> find all preview annotated symbols in a compilation
> build a list of these "banned" symbols
> use IOperation to examine the compilation for violations.
> This is essentially how the banned api analyzer operates: https://github.com/dotnet/roslyn-analyzers/tree/086d1cc888ca0d8c96b94444935f488fcaf3fe98/src/Microsoft.CodeAnalysis.BannedApiAnalyzers


Hmmm I don't understand where exactly you were going with this. The implementation is currently very very similar to this? It has a RegisterCompilationStartAction that register an operationAction and SymbolAction. Instead of reading a list of preview symbols from a txt file (like how the BannedApi analyzer works), it populates a dictionary on the fly in the `SymbolAction` and `OperationAction` call backs. Are you suggesting another approach here?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
jeffhandley: I _think_ we'd need to analyze everything that can have attributes on it, use `GetAttributes`, and then check the attribute types. That would be incredibly expensive with low value.  It could be optimized by keeping a list of preview attribute types, checking what their targets are, and only evaluating targets that could possibly be in scope.  But I think we can wait for a preview attribute scenario to emerge.

buyaa-n: >  don't know if it is even possible to create a diagnostic around [My] here

That is possible, i think if you were reporting diagnostics on the `attribute symbol` instead of https://github.com/dotnet/roslyn-analyzers/pull/5340#discussion_r682178250 reporting it on the `symbol` as i mentioned here it would work

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
jeffhandley: Follow-up... we should look all the way up to the assembly even, such that if an assembly is marked, everything in the assembly can call preview features.

jeffhandley (commented):
If I'm understanding the test syntax correctly, it looks like diagnostics are reported differently from what I expected.

I expected diagnostics to be reported on any API that is marked as preview and used within a non-preview context. It looks to me like diagnostics are reported on the element that references a preview API instead.

Since it's possible for a member to reference multiple preview APIs as well as non-preview APIs, it's important for the diagnostic to get reported on the preview API symbol instead of where it's used. That would allow for granular suppression in cases like we have for Generic Math--we don't want to suppress around `Int32`, we need to suppress around `IAdditionOperator`.

To aide in the illustration of the diagnostic placement and the suppression behavior, I recommend adding some tests that apply granular `#pragma warning disable` statements as we expect to see them used in our scenarios.

I also left a comment that I'm unsure why need to do anything to warn on static abstracts in interfaces in this analyzer--I thought the language would prevent usage; but I could be missing something.

For the unit tests, I suggest a deliberate organization, maybe even into regions. You could group them by OperationKind and then have the positive/negative/edge cases for each operation kind all together. That would make coverage/gaps more apparent.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: Still not certain we can raise a diagnostic on `[My]`. I'll deal with this as the last thing in the PR. Using attributeSymbol like Buyaa suggested, it would tag the definition of [My] i.e. `class MyAttribute : ...` and not the instance of `[My]`. Will test it out just to double check I didn't miss anything before reporting back

jeffhandley: Here's how I achieved it in the `[ModuleInitializer]` analyzer: https://github.com/dotnet/roslyn-analyzers/pull/5347/files#diff-a571ea1f340976888ed2fc96d3e57695fffcfc84f9fc93cc6acb6881f0647922R74-R80

```csharp
AttributeData? initializerAttribute = context.Symbol.GetAttributes(moduleInitializerAttribute).FirstOrDefault();
SyntaxReference? attributeReference = initializerAttribute?.ApplicationSyntaxReference;

if (attributeReference is not null)
{
    context.ReportDiagnostic(attributeReference.GetSyntax(context.CancellationToken).CreateDiagnostic(Rule));
}
```

pgovind: I didn't know about ApplicationSyntaxReference. It's undocumented too. Nice find!. This is fixed now and #5346 is now closed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -108,7 +98,8 @@ public override void Initialize(AnalysisContext context)
                     OperationKind.Unary,
                     OperationKind.Binary,
                     OperationKind.ArrayCreation,
-                    OperationKind.CatchClause
+                    OperationKind.CatchClause,
+                    OperationKind.TypeOf
                     );
 
                 // Handle library side definitions of preview features
```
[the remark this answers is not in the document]
pgovind: Just a mental model of how the analyzer works :) There are 2 registered call backs:

``` csharp
                // Handle user side invocation/references to preview features
                context.RegisterOperationAction(context => BuildSymbolInformationFromOperations(context, requiresPreviewFeaturesSymbols, previewFeaturesAttribute),
                    OperationKind.Invocation,
                    OperationKind.ObjectCreation,
                    OperationKind.PropertyReference,
                    OperationKind.FieldReference,
                    OperationKind.DelegateCreation,
                    OperationKind.EventReference,
                    OperationKind.Unary,
                    OperationKind.Binary,
                    OperationKind.ArrayCreation,
                    OperationKind.CatchClause,
                    OperationKind.TypeOf
                    );

                // Handle library side definitions of preview features
                context.RegisterSymbolAction(context => AnalyzeSymbol(context, requiresPreviewFeaturesSymbols, requiresPreviewFeaturesSymbolsToUsageType, virtualStaticsInInterfaces, previewFeaturesAttribute), s_symbols);
```

`AnalyzeSymbol` deals with type/method/property/field/event etc definitions. It doesn't track usage, it only track definitions, so I named it "library side".

`BuildSymbolInformationFromOperations` tracks the usage of preview types, akin to a client using a library, so I named it "user side".

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -210,15 +263,22 @@ private static bool SymbolIsStaticAndAbstract(ISymbol symbol, ConcurrentDictiona
...
+            if (containingSymbol != null && ProcessPreviewAttribute(containingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
             {
                 return;
             }
 
             var reportDiagnostic = symbol switch
             {
-                ITypeSymbol typeSymbol => ProcessTypeSymbolAttributes(typeSymbol, requiresPreviewFeaturesSymbols, requiresPreviewFeaturesSymbolsToUsageType, previewFeatureAttributeSymbol),
+                ITypeSymbol typeSymbol => ProcessTypeSymbolAttributes(typeSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol),
                 IPropertySymbol or IMethodSymbol => ProcessPropertyOrMethodAttributes(symbol, requiresPreviewFeaturesSymbols, requiresPreviewFeaturesSymbolsToUsageType, virtualStaticsInInterfaces, previewFeatureAttributeSymbol),
```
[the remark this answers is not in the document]
pgovind: Ah my very first PR had similar names, and the feedback I got was to use "Process*" because `TypeSymbolReferencesPreviewFeatures` was also updating a dictionary within in => it had a side effect that is not obvious from the name. I personally prefer the non `Process*` names, so I'll take your suggestion

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
jeffhandley: @Youssef1313 taught it to me. 😃

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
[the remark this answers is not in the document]
jeffhandley: Ah, makes sense--thanks. I wonder if there's better terminology that could be used. Perhaps "symbol definition references" and "symbol operation references"?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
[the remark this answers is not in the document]
jeffhandley: Cool; thanks. I think it's OK for verb-named methods to have side-effects (building caches essentially).

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -116,8 +114,7 @@ public async Task TestArrayOrPreviewTypes()
             var csInput = @" 
         using System.Runtime.Versioning; using System;
         namespace Preview_Feature_Scratch
-        {" +
-@"
+        {
             public class Program
             {
                 static void Main(string[] args)
```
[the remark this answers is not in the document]
pgovind: Actually I tried that already. I expected it to work like `Lib[] array1 = {|#1:new() { }|};` since we're listening for an `OperationKind.ObjectCreation`, but it led to a compile error: `error CS8752: The type 'Lib[]' may not be used as the target type of new()`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -382,8 +372,7 @@ public async Task TestAbstractClass()
             var csInput = @" 
         using System.Runtime.Versioning; using System;
         namespace Preview_Feature_Scratch
-        {" +
-@"
+        {
 
             class {|#0:Program|} : AbClass
```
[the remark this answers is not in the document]
pgovind: This is the approach I tried first, but I wasn't sure how we can use the IOperation based APIs to mark `AbClass` here. This is what I have currently:
``` csharp
            INamedTypeSymbol? baseType = symbol.BaseType; // This will return AbClass. symbol is still Program
            if (baseType != null)
            {
                if (SymbolIsAnnotatedAsPreview(baseType, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
                {
                    context.ReportDiagnostic(symbol.CreateDiagnostic(DerivesFromPreviewClassRule, baseType.Name));
                }
            }
```

So the diagnostic will say `'Program' derives from preview class 'AbClass' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -541,10 +527,9 @@ public async Task TestMarkedPreviewInterface()
             var csInput = @" 
         using System.Runtime.Versioning; using System;
         namespace Preview_Feature_Scratch
-        {" +
-@"
+        {
 
-            class Program : IProgram
+            class {|#1:Program|} : IProgram
```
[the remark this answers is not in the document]
pgovind: Same as Program deriving from AbClass above. The diagnostic here will say `'Program' implements a preview interface 'IProgram' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jeffhandley: Ah, interesting. Do you have a test case elsewhere that allows `new()` to be used?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jeffhandley: I wonder if @Youssef1313 knows how we can access the `AbClass` symbol here. As mentioned on another comment, it's important that we find a way to get the diagnostic to paint on the base class / interface symbol instead of the type symbol as that is needed to give `#pragma warning disable` around the preview symbol instead of the consuming symbol.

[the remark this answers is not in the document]
pgovind: What is the problem with having the diagnostic on the base class though? Suppressing the diagnostic on the base class is no different than suppressing it on the interface list right?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -831,8 +810,7 @@ public async Task TestEnum()
             var csInput = @" 
         using System.Runtime.Versioning; using System;
         namespace Preview_Feature_Scratch
-        {" +
-@"
+        {
 
             [RequiresPreviewFeatures]
             enum AnEnum
```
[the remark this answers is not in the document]
pgovind: The message was saying `Foo`. Improved it now to say `AnEnum` instead.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jeffhandley: It's a difference between:

**Desired:**

```csharp
    public readonly struct Int32 : IComparable, IConvertible, ISpanFormattable, IComparable<int>, IEquatable<int>
#if FEATURE_GENERIC_MATH
#pragma warning disable SA1001, CA2252 // comma formatting and preview features
        , IBinaryInteger<int>,
          IMinMaxValue<int>,
          ISignedNumber<int>
#pragma warning restore SA1001, CA2252
#endif // FEATURE_GENERIC_MATH
    {
```

And **Current:**

```csharp
#pragma warning disable CA2252 // preview features
    public readonly struct Int32 : IComparable, IConvertible, ISpanFormattable, IComparable<int>, IEquatable<int>
#pragma warning restore CA2252
#if FEATURE_GENERIC_MATH
#pragma warning disable SA1001 // comma formatting
        , IBinaryInteger<int>,
          IMinMaxValue<int>,
          ISignedNumber<int>
#pragma warning restore SA1001
#endif // FEATURE_GENERIC_MATH
    {
```

With the current behavior, we don't have a granularity safeguard around _which_ interfaces we're opting into using as preview.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -865,8 +843,7 @@ public async Task TestEvent()
             var csInput = @" 
         using System.Runtime.Versioning; using System;
         namespace Preview_Feature_Scratch
-        {" +
-@"
+        {
```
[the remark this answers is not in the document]
pgovind: Hmm what do you mean by event add/remove methods? I've only ever added and removed method using the `+/-` operators. Are there methods too? This is the best I could come up with to your question:

```csharp
    // Class that publishes an event
    class Publisher
    {
        [RequiresPreviewFeatures]
        public event EventHandler<EventArgs> RaiseCustomEvent;

        public void DoSomething()
        {
            OnRaiseCustomEvent(new EventArgs());
        }

        protected virtual void OnRaiseCustomEvent(EventArgs e)
        {
            EventHandler<EventArgs> raiseEvent = [|RaiseCustomEvent|];

            if (raiseEvent != null)
            {
                raiseEvent(this, e);
            }
        }
    }

    //Class that subscribes to an event
    class Subscriber
    {
        private readonly string _id;

        public Subscriber(string id, Publisher pub)
        {
            _id = id;

            // Subscribe to the event
            [|pub.RaiseCustomEvent|] += HandleCustomEvent;
            [|pub.RaiseCustomEvent|] -= HandleCustomEvent;
        }

        // Define what actions to take when the event is raised.
        void HandleCustomEvent(object sender, EventArgs e)
        {
            Console.WriteLine($"Hi");
        }
    }
```

These turn out to be nothing more than FieldReferenceOperations. Maybe you meant something else?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -505,8 +492,7 @@ public async Task TestPreviewLanguageFeatures()
             var csInput = @" 
                 using System.Runtime.Versioning; using System;
                 namespace Preview_Feature_Scratch
-                {" +
-@"
+                {
 
                     class Program : IProgram
```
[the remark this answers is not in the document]
pgovind: > Does our analyzer actually need to raise diagnostics on using statics in interfaces?

I just double checked with Immo and he mentioned that it was decided early on that the analyzer would check for `static abstract` and report a diagnostic iff `RuntimeFeature.VirtualStaticsInInterfaces` is set to true. If `LangVersion` isn't set to Preview, there will be 2 errors unfortunately. @terrajobst for visibility.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
terrajobst: Right, the message is fine, but it seems the location should be in the interface list, not the identifier `Program`.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
terrajobst: Right. Originally, we to split this analyzer and let the compiler handle the language portion but we felt like couldn't define the desired semantics well enough, especially because we want to allow 'fuzzy' rules where the author can selectively suppress.

The problem that we need to think through is the double reporting when the feature is turned off:

1. Compiler will complain about LangVersion needing to be Preview
2. Analyzer will complain about missing `[RequiresPreviewFeature]`

We generally don't want developers to put `<LangVersion>Preview</LangVersion>` into the project file.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -890,5 +867,606 @@ static void Main(string[] args)
...
+        public async Task TestPreviewParametersToMethods()
+        {
+            var csInput = @" 
+using System.Runtime.Versioning; using System;
+namespace Preview_Feature_Scratch
+{
+
+    class Program
+    {
+        public Foo {|#0:Getter|}(Foo foo)
```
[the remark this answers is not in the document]
pgovind: Actually I'm not sure we can get to `Foo` easily. What about a diagnostic on `foo` instead? (the parameter)

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
pgovind: Fixed now.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jeffhandley: Related to the interface type name reference, I think the best experience would be to always get a diagnostic on the reference to the API that is preview.

I'm not seeing anything in the Symbol APIs to get us there either though. I wonder if we're going to have to drop down to Syntax-specific APIs to traverse the syntax tree and get to the right node.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jeffhandley: Here's an example where the Platform Compatibility analyzer tests show annotating the `add` and `remove` methods of an event handler separately: https://github.com/dotnet/roslyn-analyzers/blob/main/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.cs#L1331-L1336

```csharp
    public static event EventHandler WindowsOnlyEventRemove
    {
        add { }
        [SupportedOSPlatform(""windows"")]
        remove { }
    }
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jeffhandley: It feels like a violation of the single responsibility principle to me. If the preview feature analyzer happens to also analyze usage of statics in interfaces, then that sets the precedent that it should also analyze usage of any other new language feature that is a dependency of a preview feature. But that won't hold true--especially since anyone can define a preview feature.

I don't think we'll need to block this PR on potentially revisiting this, but I suspect we'll want to iterate on this.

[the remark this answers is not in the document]
terrajobst: > If the preview feature analyzer happens to also analyze usage of statics in interfaces, then that sets the precedent that it should also analyze usage of any other new language feature that is a dependency of a preview feature.

Happy to do discuss this, but this wasn't decided on a wimp -- it was the output of the various constraints that we had. The general direction is what you're describing though, that is, if there are language features that are tight to `RuntimeFeature` fields that are marked with `[RequiresPreviewFeatures]` then yes, our analyzer would flag the use of those.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
pgovind: Separately, @sharwell, @Youssef1313 , @jmarolf, using the IOperation based APIs, is there a way to mark `AbClass` with a diagnostic in the following example:
``` csharp
class Program : [|AbClass|] //AbClass is an abstract class defined elsewhere
{

}
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
pgovind: Added 1 now to this test case.

jmarolf:
> 
> Hmmm I don't understand where exactly you were going with this. The implementation is currently very very similar to this? It has a RegisterCompilationStartAction that register an operationAction and SymbolAction. Instead of reading a list of preview symbols from a txt file (like how the BannedApi analyzer works), it populates a dictionary on the fly in the `SymbolAction` and `OperationAction` call backs. Are you suggesting another approach here?

populating a dictionary on-the-fly seems unnecessary to me. you can discover all the apis that are marked as preview in the compilation start action

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -435,8 +424,7 @@ public async Task TestUnmarkedPreviewProperty()
             var csInput = @" 
         using System.Runtime.Versioning; using System;
         namespace Preview_Feature_Scratch
-        {" +
-@"
+        {
```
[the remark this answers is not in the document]
pgovind: Sigh, I'm hoping this is not another spot where the IOperation APIs are insufficient. If I have the following test case:
```csharp
            class Program : IProgram
            {
                {
                    Program program = new Program();
                    bool getter = program.[|AGetter|];
                }

                public bool AGetter
                {
                    [RequiresPreviewFeatures]
                    get
                    {
                        return true;
                    }
                }
            }
```

Currently we're registered to inspect OperationKind.PropertyReference, so we get a callback when `program.AGetter` is encountered. I inspect the operation and perform `(IPropertyReferenceOperation)operation.Property which returns the `AGetter` symbol. There's no way to know if the current operation is a getter or a setter it looks like? :/  Is that true @jmarolf @sharwell @mavasani ? I searched for something like OperationKind.GetProperty, but that doesn't exist, so I suspect the IOperation APIs don't contain specific information for getters/setters.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
jmarolf: thanks, I misread it. ignore me

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
jmarolf: sounds good

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
pgovind: Ah cool, I had no idea this existed. Added a test now

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
buyaa-n: Not sure what exactly you are asking for or trying to do, but this might what you are asking https://github.com/dotnet/roslyn-analyzers/blob/d0663cca11722045b9aeabb5d0d5335042084e13/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L929

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jmarolf: Since each language expresses the concept of inheritance differently I believe you will need to get the [Syntax](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.ioperation.syntax?view=roslyn-dotnet-3.10.0#Microsoft_CodeAnalysis_IOperation_Syntax) from the operation node and then walk the syntax tree for each language to get this location.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
pgovind: Yup, this is what I was looking for! Thanks!!!!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
pgovind: @jmarolf : Just to double check, this means I need to abandon the current implementation and start over for a C# one right? AFAIK, there's no way to do 95% of the work in IOperation and drop down to C#/VB for the last 5%?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: It's not. We now examine the containing symbol chain all the way to the assembly. Deleted the comment

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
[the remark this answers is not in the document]
jmarolf: You can make your current class abstract with single abstract helper method `FindInheritanceLocation` that a C# and VB impl can implement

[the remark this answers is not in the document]
Youssef1313: cc @333fred for potential enhancements to IOperation APIs.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -108,138 +170,243 @@ public override void Initialize(AnalysisContext context)
...
-            ConcurrentDictionary<ISymbol, PreviewFeatureUsageType> requiresPreviewFeaturesSymbolsToUsageType,
             INamedTypeSymbol previewFeatureAttributeSymbol)
         {
+            ISymbol symbolType = symbol.Type;
+            if (symbolType is IArrayTypeSymbol arrayType)
+            {
+                symbolType = arrayType.ElementType;
+            }
+
+            if (SymbolIsAnnotatedAsPreview(symbolType, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
```
pgovind, line 193: Refactor

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1683,10 +1683,31 @@
...
+    <value>'{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+  </data>
+  <data name="ImplementsPreviewMethodMessage" xml:space="preserve">
+    <value>'{0}' implements the preview method '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+  </data>
+  <data name="MethodReturnsPreviewTypeMessage" xml:space="preserve">
+    <value>'{0}' returns the preview type '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+  </data>
+  <data name="MethodUsesPreviewTypeAsParamaterMessage" xml:space="preserve">
+    <value>`{0}` takes in a preview parameter of type `{1}` and needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
```
Youssef1313, line 1702: Change backtick to single quote for consistency

pgovind: Can you hold off on reviewing for a few minutes? I'm just going through the diff on GH myself. I'll ping you once I'm done!!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -285,21 +517,95 @@ private static bool OperationUsesPreviewFeatures(OperationAnalysisContext contex
...
+                    return true;
+                }
+            }
+
+            previewSymbol = null;
+            return false;
         }
 
-        private static bool ProcessPreviewAttribute(ISymbol symbol, ConcurrentDictionary<ISymbol, bool> requiresPreviewFeaturesSymbols, INamedTypeSymbol? previewFeatureAttribute)
+        private static bool SymbolIsAnnotatedAsPreview(ISymbol symbol, ConcurrentDictionary<ISymbol, bool> requiresPreviewFeaturesSymbols, INamedTypeSymbol previewFeatureAttribute)
```
pgovind, line 559: This is the most significant change. Previously I was only caching each symbol and a bool value. Now, I've changed this to cache each symbol and it's children in this call with a true value. I'm doing this as an optimization for deeply nested heirarachies where the Preview attribute is added close to the top.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
pgovind: Ok done, can be reviewed now @Youssef1313

pgovind:
Alright so I pushed some changes to this branch to address the concerns that @buyaa and @jeffhandley had. The most significant changes since the last review are:
1. New messages to handle various cases. The analyzer code largely remains the same. There's just some changes now to make the diagnostics sound better.
2. The `SymolIsAnnotatedAsPreview` method now caches a true value for a preview symbol and all it's children in a given call. This is required to handle deeply heirarchies where the Preview attribute is applied high up in the heirarchy.
3. I also added support for fields and events containing preview types within them.

I'm still hoping that we can get this in for RC1. I'd super appreciate another round of reviews! @buyaa-n @Youssef1313 @jmarolf

pgovind:
> populating a dictionary on-the-fly seems unnecessary to me. you can discover all the apis that are marked as preview in the compilation start action

I still don't get why this might be better. What is the exact approach in this suggestion? AFAICT, we need 2 steps in this approach:
1. We'd need to populate the dictionary in the CompilationStartAction first, and 
2. Then run another RegisterOperationAction and RegisterSymbolAction to go over the symbols in the compilation and report diagnostics on them.

Instead I combined step 1 and 2 into a single step. I don't see how having 2 steps is better than combining them? We still need to process all the symbols in the compilation right? And having 2 steps would mean going over the AST twice? Maybe I'm missing something?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1683,10 +1683,31 @@
...
   </data>
-  <data name="ImplementsEmptyPreviewInterfaceMessage" xml:space="preserve">
-    <value>'{0}' implements a preview interface. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
-  </data>
   <data name="StaticAndAbstractRequiresPreviewFeatures" xml:space="preserve">
     <value>Using both 'static' and 'abstract' modifiers requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
   </data>
+  <data name="DerivesFromPreviewClassMessage" xml:space="preserve">
+    <value>'{0}' derives from preview class '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+  </data>
```
Youssef1313, line 1691: Note: All analyzers in this package get a help link:

https://github.com/dotnet/roslyn-analyzers/blob/a497be39fc800a3436287e9e0f1987c9284ed27d/src/Utilities/Compiler/DiagnosticDescriptorHelper.cs#L36

The analyzer should eventually be documented in dotnet/docs. So likely we won't need the link in the message itself.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -43,25 +121,10 @@ public sealed class DetectPreviewFeatureAnalyzer : DiagnosticAnalyzer
...
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(GeneralPreviewFeatureAttributeRule, StaticAbstractIsPreviewFeatureRule, ImplementsEmptyPreviewInterfaceRule);
-
-        private enum PreviewFeatureUsageType
-        {
-            General = 0,
-            StaticAbstract,
-            EmptyInterface
-        }
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(GeneralPreviewFeatureAttributeRule, StaticAbstractIsPreviewFeatureRule);
```
Youssef1313, line 127: Looks outdated.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -108,138 +170,243 @@ public override void Initialize(AnalysisContext context)
...
-        private static bool ProcessTypeSymbolAttributes(ITypeSymbol symbol,
+        private static void ProcessFieldSymbolAttributes(SymbolAnalysisContext context, IFieldSymbol symbol,
+            ConcurrentDictionary<ISymbol, bool> requiresPreviewFeaturesSymbols,
+            INamedTypeSymbol previewFeatureAttributeSymbol)
+        {
+            ISymbol symbolType = symbol.Type;
+            if (symbolType is IArrayTypeSymbol arrayType)
+            {
+                symbolType = arrayType.ElementType;
+            }
```
Youssef1313, line 191: Do we need to iteratively apply this until the symbolType is no longer array symbol?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Ah yeap. Fixed now

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Nice catch. Fixed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
buyaa-n: Right, each diagnostic has its doc page, i thought the URL here for some kind of conceptual for preview feature

buyaa-n (commented):
I am not sure if the call-site guard logic is covered completely, especially for symbol analysis

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Ok fixed this locally. Will push it up in the next commit.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -257,19 +433,84 @@ private static void BuildSymbolInformationFromOperations(OperationAnalysisContex
...
+                {
+                    symbol = eventAssignment.Adds ? eventSymbol.AddMethod : eventSymbol.RemoveMethod;
+                }
+
+                if (symbol is IMethodSymbol methodSymbol && methodSymbol.IsConstructor())
+                {
+                    if (SymbolIsAnnotatedOrUsesPreviewTypes(methodSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol, out referencedPreviewSymbol))
+                    {
+                        // Constructor symbols have the name .ctor. Return the containing type instead so we get meaningful names in the diagnostic message
+                        referencedPreviewSymbol = referencedPreviewSymbol.ContainingType;
```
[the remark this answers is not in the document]
pgovind: Actually no. The diagnostic is thrown on the IOperation. `referencedPreviewSymbol` is only used to populate the name of the type in the diagnostic message.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -257,19 +433,84 @@ private static void BuildSymbolInformationFromOperations(OperationAnalysisContex
...
         {
             IOperation operation = context.Operation;
             ISymbol containingSymbol = context.ContainingSymbol;
-            if (OperationUsesPreviewFeatures(containingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol) ||
-                OperationUsesPreviewFeatures(containingSymbol.ContainingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
+            if (SymbolIsAnnotatedAsPreview(containingSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
             {
-                symbol = null;
+                referencedPreviewSymbol = null;
                 return false;
```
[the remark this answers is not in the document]
pgovind: > Is this the logic for guarding the call site?

Yup.

> Doesn't look like work for all cases, but I might not understand completely

Why not? Can you think of an example where this might not work? `containinSymbol` here is `context.Operation.ContainingSymbol`, so it is the parent to the current operation. So I'm checking to see if the parent to the current location is marked as preview, and returning false if it is. I don't see what I'm missing :/

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
jeffhandley: > I thought the URL here for some kind of conceptual for preview feature

I do think the conceptual documentation link is justified here--it can be a document that describes Preview Features, their level of support, etc. The diagnostic id help link can point to the standard documentation about the _analyzer_ itself.

/cc @gewarren

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -108,138 +179,243 @@ public override void Initialize(AnalysisContext context)
...
             INamedTypeSymbol previewFeatureAttributeSymbol)
         {
             ISymbol symbol = context.Symbol;
 
-            // Don't report diagnostics on symbols that are marked as Preview
-            if (ProcessPreviewAttribute(symbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
+            if (SymbolIsAnnotatedAsPreview(symbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol) ||
+                (symbol is IMethodSymbol method && method.AssociatedSymbol != null && SymbolIsAnnotatedAsPreview(method.AssociatedSymbol, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol)))
             {
                 return;
```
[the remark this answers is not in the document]
pgovind: Because this is the symbol definition site. So I'm guarding against reporting a diagnostic on a type/method/property/etc definition when it is marked with the preview attribute.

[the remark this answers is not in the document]
buyaa-n: I see, thanks

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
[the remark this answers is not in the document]
buyaa-n: > Can you think of an example where this might not work?

No, thinking again about the scenario i was thinking has no issue. This guard looks good for Operation analysis

buyaa-n:
> I am not sure if the call-site guard logic is covered completely, especially for symbol analysis

The symbol analysis guard looks good, i see there are a few more unresolved comments, other than that the PR looks good to me

buyaa-n (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -285,21 +526,95 @@ private static bool OperationUsesPreviewFeatures(OperationAnalysisContext contex
...
+                {
+                    requiresPreviewFeaturesSymbols.GetOrAdd(symbol, true);
+                    return true;
+                }
+
+                if (ContainingSymbolHeirarchyIsAnnotatedAsPreview(symbol, requiresPreviewFeaturesSymbols, previewFeatureAttribute, out ISymbol? containingPreviewSymbol))
+                {
+                    ISymbol loopSymbol = symbol.ContainingSymbol;
+                    requiresPreviewFeaturesSymbols.GetOrAdd(loopSymbol, true);
+                    while (loopSymbol != containingPreviewSymbol)
```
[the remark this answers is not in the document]
pgovind: > i think you can do it with one loop inside ContainingSymbolHeirarchyIsAnnotatedAsPreview

How though? For ex: let's say we have

```csharp
    [RequiresPreviewFeatures]
    public class NestedClass0
    {
        public class NestedClass1
        {
            public class NestedClass2
            {
                public class NestedClass3
                {
                    public bool AMethod() => false;
                    public bool AProperty => false;
                    public bool AField = true;
                }
            }
        }
    }
```
and we are processing `NestedClass3`. I'm going up the chain in `ContainingSymbolHeirarchyIsAnnotatedAsPreview` until we see the `[RequiresPreviewFeatures]` attribute on `NestedClass0` and return that as the containingPreviewSymbol. We have to traverse the heirarchy again from `NestedClass3` to `NestedClass0` to set the bool flag for all the symbols right? I don't see how we can do it in 1 loop inside `ContainingSymbolHeirarchyIsAnnotatedAsPreview` because we don't know which symbol above the current symbol has a preview attribute?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.cs
```diff
@@ -135,114 +164,213 @@ public class Lib
...
+
+    class Program
+    {
+        static void Main(string[] args)
         {
-                [RequiresPreviewFeatures]
-                public static bool SomeStaticProperty => false;
+            Program prog = new Program();
         }
+    }
```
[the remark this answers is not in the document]
pgovind: Removed.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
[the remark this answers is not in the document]
buyaa-n: > How though?

Recursively, see how it is done in PCA
https://github.com/dotnet/roslyn-analyzers/blob/51ac1b599127833c7da4622b419c869daab54136/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L1408-L1435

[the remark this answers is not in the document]
pgovind: Oh the recursion is just using an implied stack, so it's very similar. I fixed mine to use recursion too!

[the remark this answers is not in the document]
buyaa-n: Thanks, yes it is similar but i believe recursive will be more efficient

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -108,138 +177,300 @@ public override void Initialize(AnalysisContext context)
...
             });
         }
 
-        private static bool ProcessTypeSymbolAttributes(ITypeSymbol symbol,
+        /// <summary>
+        /// Returns null if the inner types are not preview
+        /// </summary>
+        /// <param name="typeArguments"></param>
+        /// <returns></returns>
+        private static ISymbol? GetInnerPreviewSymbolForGenericTypes(ImmutableArray<ITypeSymbol> typeArguments,
```
pgovind, line 195: Rename to GetPreviewSymbolForGenericTypesFromTypeArguments

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -283,23 +579,66 @@ private static bool OperationUsesPreviewFeatures(OperationAnalysisContext contex
...
+                {
+                    requiresPreviewFeaturesSymbols.GetOrAdd(symbol, true);
+                    return true;
+                }
+
+                INamedTypeSymbol? parent = symbol.ContainingType;
+                if (parent is INamespaceSymbol)
+                {
+                    parent = parent.ContainingType;
+                }
```
pgovind, line 630: new line

jmarolf (approved):

jmarolf:
@pgovind I've retargeted you to rc1

PR #5347, opened 2021-08-05
Implement the ModuleInitializerAttributeShouldNotBeUsedInLibraries analyzer

jeffhandley:
Fixes dotnet/runtime#43328 by raising a diagnostic any time the `[ModuleInitializer]` attribute is applied to a method (in conformance with the C# language feature that supports this attribute).

Per [the API review](https://github.com/dotnet/runtime/issues/43328#issuecomment-717452492) of this analyzer, the analyzer will be defaulted to on as a warning.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1689,4 +1689,13 @@
...
     <value>Using both 'static' and 'abstract' modifiers requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
   </data>
+  <data name="ModuleInitializerAttributeShouldNotBeUsedInLibrariesDescription" xml:space="preserve">
+    <value>Module initializers are intended to be used by application code to ensure an application's components are initialized before the application code begins executing. If library code declares a [ModuleInitializer] method, it can interfere with application initialization and also lead to limitations in that application's trimming abilities. Library code should therefore not utilize the [ModuleInitializer] attribute, but instead expose methods that can be used to initialize any components within the library and allow the application to invoke the method during application initialization.</value>
+  </data>
+  <data name="ModuleInitializerAttributeShouldNotBeUsedInLibrariesMessage" xml:space="preserve">
+    <value>The [ModuleInitializer] attribute is only intended to be used in application code or advanced source generator scenarios</value>
```
Youssef1313, line 1696: I'd replace the `[...]` with `'..'` so that we don't use C#-specific syntax.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+                                                                    isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(compilationContext =>
```
Youssef1313, line 48:
```suggestion
            context.RegisterCompilationStartAction(context =>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                INamedTypeSymbol? moduleInitializerAttributeType = compilationContext.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeCompilerServicesModuleInitializerAttribute);
```
Youssef1313, line 50:
```suggestion
                INamedTypeSymbol? moduleInitializerAttributeType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeCompilerServicesModuleInitializerAttribute);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                if (moduleInitializerAttributeType is null) return;
+
+                compilationContext.RegisterSymbolAction(context =>
```
Youssef1313, line 54:
```suggestion
                context.RegisterSymbolAction(context =>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+                {
+                    if (context.Symbol is IMethodSymbol method)
+                    {
+                        if (method.IsPrivate() ||
```
Youssef1313, line 58: Should this use `GetResultantVisibility`? The difference is when the method is declared `public`, but contained in `private` nested class.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+                            method.Parameters.Length > 0 ||
+                            method.IsGenericMethod ||
```
Youssef1313, line 60: Should we check if the containing type is generic too?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+                        )
+                        {
+                            return;
+                        }
+
+                        AttributeData? initializerAttribute = context.Symbol.GetAttributes(moduleInitializerAttributeType).FirstOrDefault();
+
+                        if (initializerAttribute is not null)
+                        {
+                            context.ReportDiagnostic(initializerAttribute.ApplicationSyntaxReference.GetSyntax().CreateDiagnostic(Rule));
```
Youssef1313, line 72: nit:

```suggestion
                            context.ReportDiagnostic(initializerAttribute.ApplicationSyntaxReference.CreateDiagnostic(Rule));
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,80 @@
...
+                            method.Parameters.Length > 0 ||
+                            method.IsGenericMethod ||
+                            !method.IsStatic ||
+                            !method.ReturnsVoid
+                        )
+                        {
+                            return;
+                        }
+
+                        AttributeData? initializerAttribute = context.Symbol.GetAttributes(moduleInitializerAttributeType).FirstOrDefault();
```
Youssef1313, line 68: `AttributeData.ApplicationSyntaxReference` is annotated as nullable, and I see other analyzers check for null. Can this happen for this analyzer?

Youssef1313 (commented):
The analyzer doesn't really check whether we're in a library or not.
So the false positive rate can be high.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                    Sources = {
+                        @$"
+                        {classModifier} class CA2255ModuleInitializerOnMethod
+                        {{
+                            [{{|CA2255:System.Runtime.CompilerServices.ModuleInitializer|}}]
```
Youssef1313, line 39: Simpler markup syntax.

```suggestion
                            [[|System.Runtime.CompilerServices.ModuleInitializer|]]
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+                            {methodModifier} static {(useAsync ? "async" : "")} void Initialize() {{ }}
+                        }}
+                        "
+                    }
+                },
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                MarkupOptions = MarkupOptions.UseFirstDescriptor
```
Youssef1313, line 46: Is this necessary?

```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [Theory]
+        [InlineData("public", "public", false)]
+        [InlineData("public", "internal", false)]
+        [InlineData("internal", "public", false)]
+        [InlineData("internal", "internal", false)]
+        [InlineData("public", "public", true)]
+        [InlineData("public", "internal", true)]
+        [InlineData("internal", "public", true)]
+        [InlineData("internal", "internal", true)]
+        public async Task CA2255ModuleInitializerOnMethod(string classModifier, string methodModifier, bool useAsync = false)
```
Youssef1313, line 28:
```suggestion
        public async Task CA2255ModuleInitializerOnMethod(string classModifier, string methodModifier, bool useAsync)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [Theory]
+        [InlineData("public", "public", false)]
+        [InlineData("public", "internal", false)]
+        [InlineData("internal", "public", false)]
+        [InlineData("internal", "internal", false)]
+        [InlineData("public", "public", true)]
+        [InlineData("public", "internal", true)]
+        [InlineData("internal", "public", true)]
+        [InlineData("internal", "internal", true)]
+        public async Task CA2255ModuleInitializerOnMethod_WithParens(string classModifier, string methodModifier, bool useAsync = false)
```
Youssef1313, line 59:
```suggestion
        public async Task CA2255ModuleInitializerOnMethod_WithParens(string classModifier, string methodModifier, bool useAsync)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [Fact]
+        public async Task CA2255DoesNotApply_ToInstanceMethods()
+        {
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                    Sources = {
+                        @$"
```
Youssef1313, line 169: Interpolation not needed here.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [Theory]
+        [InlineData("public", "public", false)]
+        [InlineData("public", "internal", false)]
+        [InlineData("internal", "public", false)]
+        [InlineData("internal", "internal", false)]
+        [InlineData("public", "public", true)]
+        [InlineData("public", "internal", true)]
+        [InlineData("internal", "public", true)]
+        [InlineData("internal", "internal", true)]
+        public async Task CA2255ModuleInitializerOnMethod_Suppressed(string classModifier, string methodModifier, bool useAsync = false)
```
Youssef1313, line 90:
```suggestion
        public async Task CA2255ModuleInitializerOnMethod_Suppressed(string classModifier, string methodModifier, bool useAsync)
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+                        {{
+                            [System.Runtime.CompilerServices.ModuleInitializer]
+                            {methodModifier} static {(useAsync ? "async" : "")} void Initialize() {{ }}
+                        }}
+                        #pragma warning restore CA2255
+                        "
+                    }
+                },
+                LanguageVersion = CodeAnalysis.CSharp.LanguageVersion.CSharp9,
+                MarkupOptions = MarkupOptions.UseFirstDescriptor
```
Youssef1313, line 110:
```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [Fact]
+        public async Task CA2255DoesNotApply_ToNonVoidMethods()
+        {
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                    Sources = {
+                        @$"
```
Youssef1313, line 146: interpolation not needed. This (and other test feedback) applies to other places. I won't add extra repeated comments to keep things cleaner and easier to track feedback.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
jeffhandley: Good catch. According to the `ModuleInitializer` compiler feature's [accessibility tests](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Test/Symbol/Symbols/ModuleInitializers/AccessibilityTests.cs), `CS8814` will kick in on a public method within a private nested class.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
jeffhandley: `SyntaxReference` doesn't have `CreateDiagnostic` on it and I couldn't find an extension method available for this shortcut. But I am going to edit this to pass the cancellation token into `GetSyntax` at least.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
jeffhandley: Thanks; I'll guard it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
Youssef1313: @jeffhandley Oh, sorry. It's defined in Roslyn.Diagnostics.Analyzers project, but not for NetAnalyzers

https://github.com/dotnet/roslyn-analyzers/blob/main/src/Roslyn.Diagnostics.Analyzers/Core/DiagnosticExtensions.cs

jeffhandley: Ah, gotcha. OK if I leave it as-is then, or would you prefer I make that extension method available?

jeffhandley:
> The analyzer doesn't really check whether we're in a library or not.
> So the false positive rate can be high.

We expect the `[ModuleInitializer]` attribute to be used very rarely, so we can tolerate a level of false positives.

Is it possible to check if the type is in a class library assembly?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
jeffhandley: Thanks. I had consciously kept the interpolation in place because it kept the code copy-pastable across the tests. But I'll update it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
jeffhandley: Since the tests (have to) illustrate places where `CS88xx` diagnostics show up, I wanted to explicitly illustrate where `CA2255` is raised to be clear which diagnostic shows up in which tests and avoid a false positive match on the wrong diagnostic.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,207 @@
...
+using Xunit;
+
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.ModuleInitializerAttributeShouldNotBeUsedInLibraries,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.ModuleInitializerAttributeShouldNotBeUsedInLibraries,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
```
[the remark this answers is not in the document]
jeffhandley: Oops. :-)
I had VB tests in place before I realized the feature didn't apply to VB.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
Youssef1313: I think it's okay to leave it.

Youssef1313:
> > The analyzer doesn't really check whether we're in a library or not.
> > So the false positive rate can be high.
> 
> We expect the `[ModuleInitializer]` attribute to be used very rarely, so we can tolerate a level of false positives.
> 
> Is it possible to check if the type is in a class library assembly?

Yeah I think it's possible: `if (context.Compilation.Options.OutputKind != OutputKind.DynamicallyLinkedLibrary) return;`

jeffhandley:
> Yeah I think it's possible: `if (context.Compilation.Options.OutputKind != OutputKind.DynamicallyLinkedLibrary) return;`

Thanks! This will still allow the analyzer to run for web and mobile apps I presume, but it will remove false positives from console apps and windows apps, so that's still an improvement.

Youssef1313:
> This will still allow the analyzer to run for web and mobile apps I presume

I think it won't run for web apps. Not sure about mobile.

jeffhandley:
Thanks for the thorough review, @Youssef1313! I've addressed all of your feedback if you want to take another look.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1689,4 +1689,13 @@
...
     <value>Using both 'static' and 'abstract' modifiers requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
   </data>
+  <data name="ModuleInitializerAttributeShouldNotBeUsedInLibrariesDescription" xml:space="preserve">
+    <value>Module initializers are intended to be used by application code to ensure an application's components are initialized before the application code begins executing. If library code declares a [ModuleInitializer] method, it can interfere with application initialization and also lead to limitations in that application's trimming abilities. Library code should therefore not utilize the [ModuleInitializer] attribute, but instead expose methods that can be used to initialize any components within the library and allow the application to invoke the method during application initialization.</value>
+  </data>
+  <data name="ModuleInitializerAttributeShouldNotBeUsedInLibrariesMessage" xml:space="preserve">
+    <value>The ModuleInitializer attribute is only intended to be used in application code or advanced source generator scenarios</value>
```
Youssef1313, line 1696: Usually we enclose type names between single quotes (applies to `Message`, `Title`, and `Description`)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,84 @@
...
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.ModuleInitializerAttributeShouldNotBeUsedInLibrariesDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+
+        internal static DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                    s_localizableTitle,
+                                                                    s_localizableMessage,
+                                                                    DiagnosticCategory.Usage,
+                                                                    RuleLevel.BuildWarning,
+                                                                    s_localizableDescription,
+                                                                    isPortedFxCopRule: true,
```
Youssef1313, line 38: I think this should be `false`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
Youssef1313: @jeffhandley I think `[| |]` isn't supposed to match CS88xx, it matches the supported diagnostic from the analyzer under test, and (by default) will fail if the analyzer supports multiple diagnostics or if the diagnostic is from the compiler. but I'm not opposing leaving it as is. It's more of style.

Youssef1313 (approved):

jeffhandley:
I forgot to run `msbuild /t:Pack` on that commit. I'm correcting the resource strings and running that command, which should fix the build.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Linq;
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    /// <summary>
+    /// CS2255: ModuleInitializer attributes should not be used in libraries.
```
Youssef1313, line 13: very minor nit :smile:

```suggestion
    /// CA2255: ModuleInitializer attributes should not be used in libraries.
```

jeffhandley: pffft. not minor at all. 🤦‍♂️

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
jeffhandley: Ah, I found that I can specify `CompilerDiagnostics = CompilerDiagnostics.None` to disable those diagnostics from showing up in the verification.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,254 @@
...
+        [Theory]
+        [InlineData("public", "public", false)]
+        [InlineData("public", "internal", false)]
+        [InlineData("internal", "public", false)]
+        [InlineData("internal", "internal", false)]
+        [InlineData("public", "public", true)]
+        [InlineData("public", "internal", true)]
+        [InlineData("internal", "public", true)]
+        [InlineData("internal", "internal", true)]
+        public async Task CA2255ModuleInitializerOnMethod_WithParens(string classModifier, string methodModifier, bool useAsync)
```
buyaa-n, line 54: NIT: I might miss something but don't see any difference between this and the previous test, what is `Parens`?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            await new VerifyCS.Test
+            {
+                TestState =
+                {
+                    ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                    Sources = {
+                        @$"
+                        {classModifier} class AccessibleClass
+                        {{
+                        #pragma warning disable CA2255
```
buyaa-n, line 95: NIT: we don't really need to test suppression, but i am ok with keeping it

src/PerformanceTests/Tests/Enabled/CSharp_CA2255.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            var sources = new List<(string name, string content)>();
+            for (var i = 0; i < Constants.Number_Of_Code_Files; i++)
+            {
+                var name = "TypeName" + i;
+                sources.Add((name, @$"
+class {name}
+{{
+    [System.Runtime.CompilerServices.ModuleInitializer]
+    void Initialize() {{ }}
```
buyaa-n, line 30: isn't it should be static?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibraries.cs
```diff
@@ -0,0 +1,88 @@
...
+                    {
+                        // Eliminate methods that would fail the CS8814, CS8815, and CS8816 checks
+                        // for what can have [ModuleInitializer] applied
+                        if (method.GetResultantVisibility() == SymbolVisibility.Private ||
+                            method.Parameters.Length > 0 ||
+                            method.IsGenericMethod ||
+                            method.ContainingType.IsGenericType ||
+                            !method.IsStatic ||
+                            !method.ReturnsVoid
+                        )
```
buyaa-n, line 70: NIT:
```suggestion
                            !method.ReturnsVoid)
```

src/PerformanceTests/Tests/Enabled/CSharp_CA2255.cs
buyaa-n: Seems you wanted to include CS88XY diagnostics, so i making the comment as NIT up to you, but i don't think that is needed, plus it is not clear if one of the  captured 3 diagnostics is actually from this analyzer, i might be wrong but though when there is a compiler error kind of analyzers will not run

buyaa-n (approved):
Left few NITs overall looks good, thanks!

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
jeffhandley: It's subtle:

```csharp
[[|System.Runtime.CompilerServices.ModuleInitializer|]]
// vs
[[|System.Runtime.CompilerServices.ModuleInitializer()|]]
```

I mostly added this test because I wasn't sure (and was curious) if the diagnostic would highlight just the attribute type name or the entire attribute; it highlights the entire attribute, parens/arguments included.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/ModuleInitializerAttributeShouldNotBeUsedInLibrariesTests.cs
jeffhandley: I added it for 2 reasons:

1. To validate the diagnostic is `CA2255` as expected (that doesn't get validated anywhere else)
2. Since the analyzer first identifies the method and then adds the diagnostic to the attribute, I wanted to illustrate the suppression only needs to be around the attribute (not the method)

src/PerformanceTests/Tests/Enabled/CSharp_CA2255.cs
jeffhandley: You're right; thanks! I had created the perf test before adding the logic to enforce public/static, and I failed to revisit this.

PR #5351, opened 2021-08-06
Update top-level checks implementations

Youssef1313:
Needed for https://github.com/dotnet/roslyn/issues/54877.

### DIFF
--- a/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
@@ -709,9 +709,7 @@ namespace Analyzer.Utilities.Extensions
         /// Check if the given <paramref name="methodSymbol"/> is an implicitly generated method for top level statements.
         /// </summary>
         public static bool IsTopLevelStatementsEntryPointMethod([NotNullWhen(true)] this IMethodSymbol? methodSymbol)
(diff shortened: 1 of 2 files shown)


src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -709,9 +709,7 @@ public static bool IsAutoPropertyAccessor(this IMethodSymbol methodSymbol)
         /// Check if the given <paramref name="methodSymbol"/> is an implicitly generated method for top level statements.
         /// </summary>
         public static bool IsTopLevelStatementsEntryPointMethod([NotNullWhen(true)] this IMethodSymbol? methodSymbol)
-            => methodSymbol?.ContainingType.IsTopLevelStatementsEntryPointType() == true &&
-               methodSymbol.IsStatic &&
```
mavasani, line 713: I think we should still retain the `IsStatic` check.

Youssef1313: @mavasani It looks redundant to me. If the name is `<Main>$`, then the compiler must have emitted it as static. Any particular reason to keep it?

(As it's not a big deal - I'm going to add it now, but curious about what difference does it make)

PR #5352, opened 2021-08-06
Set analyzers `global_level` to `-99`/`-100`.

alexrp:
This allows MSBuild SDKs that extend `Microsoft.NET.Sdk` to inject config files between Roslyn's config files and arbitrary files provided by the user or NuGet packages.

I think this change should be safe since `global_level` hasn't shipped with a final .NET build yet.


alexrp:
Related: https://github.com/dotnet/roslyn/pull/55472

jmarolf (approved):

PR #5355, opened 2021-08-07
Remove unneeded `Shared` attribute

Youssef1313:
This was discussed recently on Discord.


Evangelink:
@Youssef1313 Could elaborate a bit more? Or give a link to the discord chat message?

Youssef1313:
@Evangelink `Shared` attribute is meant for MEF-exported components. It causes the component to be instantiated only once (used for codefixes as they're stateless classes). For analyzers, they don't use MEF at all, so the `Shared` attribute does nothing. You can find that most (if not all) analyzers in the repo don't use `Shared` for analyzers (but use it for codefixes)

mavasani (approved):

sharwell (approved):

PR #5356, opened 2021-08-07
CA1011: Consider passing base types as parameters

### LINKED ISSUE #375: Port FxCop rule CA1011: ConsiderPassingBaseTypesAsParameters
**Title:** Consider passing base types as parameters

**Description:**

When a base type is specified as a parameter in a method declaration, any type that is derived from the base type can be passed as the corresponding argument to the method. If the additional functionality that is provided by the derived parameter type is not required, use of the base type enables wider use of the method.

**Proposed analyzer:** Microsoft.ApiDesignGuidelines

**Notes:**

Evangelink:
Fixes #375


src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,227 @@
...
+                        methodSymbol.HasEventHandlerSignature(eventArgsTypeSymbol) ||
+                        !context.Options.MatchesConfiguredVisibility(Rule, methodSymbol, context.Compilation,
+                            defaultRequiredVisibility: SymbolVisibilityGroup.All) ||
+                        context.Options.IsConfiguredToSkipAnalysis(Rule, methodSymbol,
+                            context.Compilation))
+                    {
+                        return;
+                    }
+
+                    var typesUsedPerParameter = PooledConcurrentDictionary<IParameterSymbol, PooledConcurrentSet<ITypeSymbol>>.GetInstance(SymbolEqualityComparer.Default);
```
Evangelink, line 76: I am absolutely not sure that's the right type to use, feedback welcomed.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,227 @@
...
+                    foreach (var param in methodSymbol.Parameters)
+                    {
+                        typesUsedPerParameter.AddOrUpdate(param, PooledConcurrentSet<ITypeSymbol>.GetInstance(SymbolEqualityComparer.Default), (param, value) => value);
+                    }
+
+                    context.RegisterOperationAction(context =>
```
Evangelink, line 82: I could have registered multiple operations and checked that the target was of type `IParameterReferenceOperation` but it felt easier this way.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,227 @@
...
+
+                            case OperationKind.Argument:
+                                AddType(paramReference.Type);
+                                break;
+
+                            case OperationKind.Conversion:
+                                AddType(paramReference.Parent.Type);
+                                break;
+
+                            default:
```
Evangelink, line 112: Not sure if you want to keep the fallback case and be explicit about the fact we do nothing?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,227 @@
...
+            var containingType = symbol.ContainingType;
+            var query = from iface in containingType.AllInterfaces
+                        from interfaceMember in iface.GetMembers(symbol.Name)
+                        let impl = containingType.FindImplementationForInterfaceMember(interfaceMember)
+                        where symbol.Equals(impl)
+                        select interfaceMember;
+            return query.ToImmutableArray();
+        }
+
+        private static ITypeSymbol FindMostGenericType(ITypeSymbol originalType, ImmutableArray<ITypeSymbol> constraints)
```
Evangelink, line 209: The ordering in this method always favors interfaces over types, this could be revisited if wanted.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1266,4 +1266,13 @@
   <data name="OperatorOverloadsHaveNamedAlternatesCodeFixTitle" xml:space="preserve">
     <value>Add operator overload named alternate</value>
   </data>
-</root>
+  <data name="ConsiderPassingBaseTypesAsParametersDescription" xml:space="preserve">
```
Evangelink, line 1269: I have tried to copy the description from fxcop but the title and message are customized. Let me know what's your preference.

Evangelink, line 1269: I have tried to copy the description from fxcop but the title and message are customized. Let me know what's your preference.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,227 @@
...
+                // When reaching this point, either we found no interface member, which means the
+                // member is declared directly on the type OR more than one interface was found
+                // as declaring this type so let's assume it is defined on the containing type.
+                return symbol.ContainingType;
+            }
+        }
+
+        // Borrowed from Microsoft.CodeAnalysis.Workspaces/ISymbolExtensions.cs with a small tweak
+        // which consist in passing the member name.
+        private static ImmutableArray<ISymbol> ExplicitOrImplicitInterfaceImplementations(ISymbol symbol)
```
Evangelink, line 191: I am surprised I couldn't find such helper in the codebase, have I missed something?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
```diff
@@ -0,0 +1,418 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.ConsiderPassingBaseTypesAsParameters,
+    Microsoft.CodeAnalysis.Testing.EmptyCodeFixProvider>;
+
+namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.UnitTests
+{
+    public class ConsiderPassingBaseTypesAsParametersTests
```
Evangelink, line 12: Potential additional tests:

- Method is marked as `override`
- Method is interface implementation
- Visibility user option
- Skip analysis user option
- Generic interfaces
- Struct
- Record

All the VB.NET tests

Youssef1313 (commented):
Should retarget .NET 6 release branch I think

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1266,4 +1266,13 @@
...
+    <value>Consider changing the type of parameter '{0}' from '{1}' to its base type '{2}'. This method appears to only require base class members in its implementation.</value>
```
Youssef1313, line 1270: Description shouldn't contain placeholders

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,227 @@
...
+                context.RegisterOperationBlockStartAction(context =>
+                {
+                    if (context.OwningSymbol is not IMethodSymbol methodSymbol ||
+                        methodSymbol.IsOverride ||
+                        methodSymbol.Parameters.Length == 0 ||
+                        methodSymbol.IsImplementationOfAnyInterfaceMember() ||
+                        methodSymbol.HasEventHandlerSignature(eventArgsTypeSymbol) ||
+                        !context.Options.MatchesConfiguredVisibility(Rule, methodSymbol, context.Compilation,
+                            defaultRequiredVisibility: SymbolVisibilityGroup.All) ||
+                        context.Options.IsConfiguredToSkipAnalysis(Rule, methodSymbol,
```
Youssef1313, line 70: Should we instead always default to reporting public APIs? The reasoning behind the rule is "allowing broader API usage", which doesn't make much sense for internal/private APIs.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
```diff
@@ -0,0 +1,418 @@
...
+    public void M(B b)
+    {
+    }
+}");
+        }
+
+        [Fact]
+        public async Task MethodCall_TypeWithBase()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
```
Youssef1313, line 32: Use VerifyCodeFixAsync and pass the same source as fixedSource. The benefits are:

- Whenever a codefix is implemented, the test will fail and require update to add the fixedCode
- Even if there is a case when a codefix shouldn't be offered, it will check that it's indeed not offered, instead of a code action that does nothing is offered.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
```diff
@@ -0,0 +1,418 @@
...
+{
+    public int GetSomethingElse() => 42;
+}
+
+public class Z
+{
+    public void UsesOnlyBaseType(B {|#0:b|})
+    {
+        b.GetSomething();
+    }
```
Youssef1313, line 48: Can we get to a scenario where changing `B` to `A` can cause `GetSomething` to be statically (at compile-time) or dynamically (at run-time) bound to something different?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I don't know, for me it makes sense to still report on private functions if I want to because it allows for a better chance of reusability.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
Evangelink: Will do! I have been a little lazy with the implementation of a code-fix but if I manage to save some time this week I will try to work on it.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,5 @@
 ; Please do not edit this file manually, it should only be updated through code fix application.
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+CA1011 | Design | Hidden | ConsiderPassingBaseTypesAsParameters, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1011)
```
mavasani, line 5: New analyzers added to NetAnalyzers project should not be targeting main branch, but a specific .NET release branch (`release/7.0.1xx` I believe): https://github.com/dotnet/roslyn-analyzers/blob/main/docs/NetCore_GettingStarted.md#branch-definitions

@buyaa-n @carlossanlop - this PR is porting an old FxCop CA rule to .NET Analyzers. Does this need to go through dotnet/runtime triage?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -4,6 +4,7 @@
 
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
+CA1011 | Design | Hidden | ConsiderPassingBaseTypesAsParameters, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1011)
```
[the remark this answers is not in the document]
Evangelink: See https://github.com/dotnet/runtime/issues/61916

AraHaan:
I am unsure if I like this rule or not, after all it just depends on what the base type is for. An example against is for when you have cpu specific structures that get selected at runtime where you have a blank base type for it that all of the cpu specific versions derive from and have separate functions for each of them that takes each specific one specially to do special stuff (aka one does avx2 specific optimizations where the cpu supports it, sse2 otherwise). So to me I feel like this rule is a catch 21 between "I like it." and "I hate it." depending on use case, likewise I think it should not default as hidden, but to disabled for those who do not want to use it by default.

Evangelink:
I understand the feeling, I also have and love/hate relation to this one. When developing syntax based analyzers, this is often more of a pain because although the property or method is accessible on all `SyntaxNode` I want the function to be focus on a specific node type because the shape of parent/children... is expected only in this context. On the opposite side, I have seen many more junior dev using specific collections as input while a more abstract one would be sufficient.

Maybe the rule should be less wide (i.e. more focused on certain types: collections?...).

carlossanlop:
> this PR is porting an old FxCop CA rule to .NET Analyzers. Does this need to go through dotnet/runtime triage?

Yep. And it was approved today: https://github.com/dotnet/runtime/issues/61916#issuecomment-982973839

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,210 @@
...
+                });
+            });
+        }
+
+        private static ITypeSymbol FindDeclaringType(ISymbol symbol)
+        {
+            Debug.Assert(
+                symbol.Kind != SymbolKind.Method ||
+                symbol.Kind != SymbolKind.Property ||
+                symbol.Kind != SymbolKind.Event);
```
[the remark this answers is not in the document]
Evangelink: Just that this method is only used in a context where the `ISymbol` is only one of these types. I am fine to remove it if this feels un-necessary.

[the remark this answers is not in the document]
Youssef1313: @Evangelink This should be `==` then right?
The assertion doesn't make much sense with `!=` and `||`. The assertion will pass for all symbol kinds currently.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I wasn't, it's now fixed. If you do a global search in the codebase, it seems that we are not doing it consistently or even not always at all.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,210 @@
...
+                                AddDeclaringTypeSymbol(((IPropertyReferenceOperation)paramReference.Parent).Property);
+                                break;
+
+                            case OperationKind.EventReference:
+                                AddDeclaringTypeSymbol(((IEventReferenceOperation)paramReference.Parent).Event);
+                                break;
+
+                            case OperationKind.FieldReference:
+                                AddType(((IFieldReferenceOperation)paramReference.Parent).Field.ContainingType);
+                                break;
```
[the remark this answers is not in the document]
Evangelink: Done.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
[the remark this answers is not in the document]
Evangelink: Thanks!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
```diff
@@ -1313,5 +1313,13 @@
   </data>
   <data name="CollectionsShouldImplementGenericInterfaceMultipleMessage" xml:space="preserve">
     <value>Type '{0}' directly or indirectly inherits '{1}' without implementing any of '{2}'. Publicly-visible types should implement the generic version to broaden usability.</value>
+  <data name="ConsiderPassingBaseTypesAsParametersDescription" xml:space="preserve">
```
Youssef1313, line 1316: This was dropped when conflicts were fixed and is why the sarif file and markdown file produces empty descriptions currently.

```suggestion
  </data>
  <data name="ConsiderPassingBaseTypesAsParametersDescription" xml:space="preserve">
```

You'll need to re-run pack after making this change.

Feel free to make the changes in a single commit locally instead of accepting the suggestion from GitHub then running pack in a separate commit.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,239 @@
...
+                        AddDeclaringTypeSymbol(((IEventReferenceOperation)paramReference.Parent).Event);
+                        break;
+
+                    case OperationKind.FieldReference:
+                        AddType(((IFieldReferenceOperation)paramReference.Parent).Field.ContainingType);
+                        break;
+
+                    case OperationKind.MethodReference:
+                        AddType(((IMethodReferenceOperation)paramReference.Parent).Method.ContainingType);
+                        break;
```
Youssef1313, line 94: All those `IOperation`s are `IMemberReferenceOperation`, could we group them under a single `case` and use `AddDeclaringTypeSymbol(((IMemberReferenceOperation)paramReference.Parent).Member)`. I'm not sure if `AddDeclaringTypeSymbol` would do the right thing with this suggestion. But something to consider if it can be done.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/MicrosoftCodeQualityAnalyzersResources.resx
Evangelink: Thank you @Youssef1313!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: @mavasani What's your opinion?

Evangelink:
@Youssef1313 I am trying to run the pack option but I keep getting an empty description, is there any open related issue you are aware of?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+        internal const string RuleId = "CA1011";
+
+        internal static readonly DiagnosticDescriptor Rule =
+            DiagnosticDescriptorHelper.Create(
+                RuleId,
+                CreateLocalizableResourceString(ConsiderPassingBaseTypesAsParametersTitle),
+                CreateLocalizableResourceString(ConsiderPassingBaseTypesAsParametersMessage),
+                DiagnosticCategory.Design,
+                RuleLevel.IdeSuggestion,
+                description: CreateLocalizableResourceString(ConsiderPassingBaseTypesAsParametersDescription),
```
Youssef1313, line 30: I think this should fix it :)

```suggestion
                CreateLocalizableResourceString(nameof(ConsiderPassingBaseTypesAsParametersTitle)),
                CreateLocalizableResourceString(nameof(ConsiderPassingBaseTypesAsParametersMessage)),
                DiagnosticCategory.Design,
                RuleLevel.IdeSuggestion,
                description: CreateLocalizableResourceString(nameof(ConsiderPassingBaseTypesAsParametersDescription)),
```

Evangelink: I am so stupiiiidd! Thank you so much :D

Youssef1313: This took time first time I hit it. I wish a "constant parameter" feature in C#, which would have avoided this and I think helps occasionally. Not sure if there is already a proposal.

Evangelink: I saw something about it a couple of days ago

AraHaan:
odd the 2 ci's are experiencing failures for some reason with some sort of brownout.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
mavasani: Note that all the FxCop Design rules always operated on just the public API surface by default. So, it makes sense to default to public, but support the `api_surface` option to enable other visibilities.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+    {
+        internal const string RuleId = "CA1011";
+
+        internal static readonly DiagnosticDescriptor Rule =
+            DiagnosticDescriptorHelper.Create(
+                RuleId,
+                CreateLocalizableResourceString(nameof(ConsiderPassingBaseTypesAsParametersTitle)),
+                CreateLocalizableResourceString(nameof(ConsiderPassingBaseTypesAsParametersMessage)),
+                DiagnosticCategory.Design,
+                RuleLevel.IdeSuggestion,
```
mavasani, line 29: According to https://github.com/dotnet/runtime/issues/61916#issuecomment-977195070, this should be Hidden, not an IDE suggestion.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        private static void AnalyzeOperationBlockStart(OperationBlockStartAnalysisContext context, INamedTypeSymbol? eventArgsTypeSymbol)
+        {
+            if (context.OwningSymbol is not IMethodSymbol methodSymbol ||
+                methodSymbol.IsOverride ||
+                methodSymbol.Parameters.Length == 0 ||
+                methodSymbol.IsImplementationOfAnyInterfaceMember() ||
+                methodSymbol.HasEventHandlerSignature(eventArgsTypeSymbol) ||
+                !context.Options.MatchesConfiguredVisibility(Rule, methodSymbol, context.Compilation,
+                    defaultRequiredVisibility: SymbolVisibilityGroup.All) ||
```
mavasani, line 57: Lets keep parity with FxCop CA1011 implementation and only report on public APIs by default
```suggestion
                    defaultRequiredVisibility: SymbolVisibilityGroup.Public) ||
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+                        }
+                        else if (usedTypes.Length > 1)
+                        {
+                            ReportIfTypesAreDifferent(parameter, FindMostGenericType(parameter.Type, usedTypes), overloads);
+                        }
+                    }
+                }
+                finally
+                {
+                    typesUsedPerParameter.Free(context.CancellationToken);
```
mavasani, line 136: Does this also free the PooledConcurrentSet values in this dictionary?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+                        parameter.Type.Name, suggestedBaseTypeSymbol.Name));
+                }
+            });
+        }
+
+        private static ITypeSymbol FindDeclaringType(ISymbol symbol)
+        {
+            if (symbol.IsOverride)
+            {
+                return FindDeclaringType(symbol.GetOverriddenMember());
```
mavasani, line 170: I believe there are some error cases where `GetOverriddenMember` can return null even if `IsOverride` is true. I think you may want to add a null check somewhere.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
mavasani: Can this be moved to ISymbolExtensions.cs?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+
+                    case OperationKind.Argument:
+                        AddType(paramReference.Type);
+                        break;
+
+                    case OperationKind.Conversion:
+                        AddType(paramReference.Parent.Type);
+                        break;
+
+                    default:
```
mavasani, line 95: How about the following?
```cs
void M(Derived d)
{
   Derived d2 = d;
   d2.SomeDerivedTypeMethod();
}
```

We probably need to identify cases where the parameter _escapes_ our tracking logic and we have to bail out on it.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
Youssef1313: @Evangelink Consider generic method test as well. I'm concerned by this comment in the compiler:

https://github.com/dotnet/roslyn/blob/10f51778f79b90cb0b7cdd39355584cef7fa7059/src/Compilers/CSharp/Portable/Symbols/MethodSymbol.cs#L480-L481

You might need `.OriginalDefinition` after `.TargetMethod` to fix it.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+
+                    foreach (var (parameter, usedTypesPool) in typesUsedPerParameter)
+                    {
+                        var usedTypes = usedTypesPool.ToImmutableArray();
+                        // If type is unused, we ignore it. There is another rule to detect
+                        // unused parameters.
+
+                        if (usedTypes.Length == 1)
+                        {
+                            ReportIfTypesAreDifferent(parameter, usedTypes[0], overloads);
```
mavasani, line 126: What if the used type is even less generic?

```cs
// Base => Derived => FurtherDerived
void M(Derived d)
{
   d.MethodOnDerived();
   if (...) // is always FurtherDerived?
   {
      ((FurtherDerived)d).MethodOnFurtherDerived();
   }  
}
```

I think we should not suggest changing the parameter type to `FurtherDerived`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,225 @@
...
+                        select interfaceMember;
+            return query.ToImmutableArray();
+        }
+
+        private static ITypeSymbol FindMostGenericType(ITypeSymbol originalType, ImmutableArray<ITypeSymbol> constraints)
+        {
+            Debug.Assert(constraints.Length >= 2);
+
+            var mostGenericInterface =
+                originalType.AllInterfaces.FirstOrDefault(type => constraints.All(x => type.Inherits(x)));
```
mavasani, line 212: Can we make this method more performant by ordering or reducing the `constraints` types set? For example, if it contains types A and B, such that A derives from B, then we can remove A from the constraints right?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
Youssef1313: Can also consider adding `.OriginalDefinition` inside `GetOverriddenMember` extension itself.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: Thanks!

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I don't know. `Free()` is calling `Clear` which I assume calls `Dispose` which would call the `Free` but I am not sure so I will add a loop.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
buyaa-n: Yes, we proposed to be hidden with that comment, but it is approved with IdeSuggestion https://github.com/dotnet/runtime/issues/61916#issuecomment-982973839, no need to change

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: The various properties followed by `GetOverriddenMember`  are not annotated with null, if you know it can return null maybe our helper should be annotated so we avoid such mistake.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I am adding a test for that but I believe we should know about this change of reference (maybe not).

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I am not sure to follow. Do you mean we should not suggest to replace `Derived` by `Base` in the method signature? For me it would makes sense to still suggest so.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: Ah nullable are not fully enable on roslyn...

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I confirm that we do not report on this case.

mavasani: How about following:

```csharp
void M(Derived d1, Derived d2)
{
   d1.SomeBaseTypeMethod();
   d2.SomeDerivedTypeMethod();
}
```

Do we suggest changing the parameter type to Base for d1?

mavasani: > I confirm that we do not report on this case.

Hmm I guess that is because there are no other parameter references. How about the following slightly modified example:

```csharp
void M(Derived d)
{
   d.SomeBaseTypeMethod();

   Derived d2 = d;
   d2.SomeDerivedTypeMethod();
}
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
mavasani: > Do you mean we should not suggest to replace Derived by Base

I meant we should not suggest changing `Derived` to `FurtherDerived` here. I don't think anything inside `ReportIfTypesAreDifferent` checks if the suggested type is a more generic or less generic type and we skip invoking `FindMostGenericType` when `usedTypes.Length == 1`

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: We also do not report here.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: I have added a loop.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParametersTests.cs
Evangelink: I can't think of anything.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
Evangelink: Is it going to be faster? I mean to make the reduction of the constraints I will have to make a lot of iterations.

Also, for your example, I think we want the opposite. If `A` derives from `B` then we want to keep `B` (i.e. the less generic type).

Evangelink: Just in case, I was thinking about the following implementation:

```csharp
			var reducedConstraints = constraints.ToList();
            for (int i = 1; i < reducedConstraints.Count; i++)
            {
                if (reducedConstraints[i].Inherits(reducedConstraints[i - 1]))
                {
                    reducedConstraints.RemoveAt(i - 1);
                }
                else if (reducedConstraints[i - 1].Inherits(reducedConstraints[i]))
                {
                    reducedConstraints.RemoveAt(i);
                }
            }
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/ConsiderPassingBaseTypesAsParameters.cs
```diff
@@ -0,0 +1,214 @@
...
+            return interfaceMembers.Length switch
+            {
+                // Member is not overridden nor interface implementation so it must be
+                // implemented on the containing type.
+                0 => symbol.ContainingType,
+                1 => interfaceMembers[0].ContainingType,
+                // False negative: more than one interface were found as declaring this
+                // type so let's assume it is defined on the containing type. Ideally,
+                // we would want to collect the type and make multiple offers to users.
+                _ => symbol.ContainingType,
```
Evangelink, line 192: @mavasani It'd be nice to handle this case but I am not sure how to best handle this. For example if the code uses `.Count` from `List<T>` we have `ICollection<T>.Count`, `IReadonlyCollection<T>.Count` and `ICollection.Count`.

PR #5357, opened 2021-08-07
Replace bitwise operator to boolean

flash2048:
With the bitwise operator, it's possible to get `NullReferenceException`, because the right-hand operand will execute.
With the boolean operator if the `attributeInstance.ConstructorArguments[0].Value != null` will false the next operand will  not run.


drewnoakes (approved):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/MarkAssembliesWithComVisible.cs
```diff
@@ -67,7 +67,7 @@ private static void AnalyzeCompilation(CompilationAnalysisContext context)
                 {
                     if (!attributeInstance.ConstructorArguments.IsEmpty &&
                         attributeInstance.ConstructorArguments[0].Kind == TypedConstantKind.Primitive &&
-                        attributeInstance.ConstructorArguments[0].Value != null &
+                        attributeInstance.ConstructorArguments[0].Value != null &&
                         attributeInstance.ConstructorArguments[0].Value.Equals(true))
```
sharwell, line 71:
```suggestion
                        Equals(attributeInstance.ConstructorArguments[0].Value, true))
```

sharwell (approved):

PR #5367, opened 2021-08-11
Use Async Methods when in Async Context Analyzer

mahdiva:
This PR adds the analyzer for preferring async methods when inside a `Task`-returning async method.

Fixes dotnet/runtime#45661

This analyzer was already implemented in VS-Threading [VSTHRD103](https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD103.md), but was based on C# specific APIs.
This PR is implemented using the `IOperation` based APIs instead with a similar functionality, supporting both C# and VB tests.

[Old analyzer implementation](https://github.com/microsoft/vs-threading/blob/main/src/Microsoft.VisualStudio.Threading.Analyzers.CSharp/VSTHRD103UseAsyncOptionAnalyzer.cs) for reference

cc @pgovind


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1698,4 +1698,16 @@
...
+    <value>'{0}' synchronously blocks. Await '{1}' instead.</value>
+  </data>
+  <data name="UseAsyncMethodInAsyncContextTitle" xml:space="preserve">
+    <value>Call async methods when in an async method</value>
+  </data>
+  <data name="UseAsyncMethodInAsyncContextMessage_NoAlternative" xml:space="preserve">
+    <value>'{0}' synchronously blocks. Use await instead.</value>
+  </data>
+  <data name="UseAsyncMethodInAsyncContextDescription" xml:space="preserve">
+    <value>When inside a Task-returning method, use the async version of methods, if they exist.</value>
```
Youssef1313, line 1711: The localization team will likely attempt to translate "Task" here.

Localization can be prevented like this:

https://github.com/dotnet/roslyn-analyzers/blob/affa88c580a9cffc5b016ad892c5bcd3971e8456/src/Roslyn.Diagnostics.Analyzers/Core/RoslynDiagnosticsAnalyzersResources.resx#L378-L381

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,215 @@
...
+        internal static DiagnosticDescriptor DescriptorNoAlternativeMethod = DiagnosticDescriptorHelper.Create(RuleId,
+                                                                              s_localizableTitle,
+                                                                              s_localizableMessageNoAlternative,
+                                                                              DiagnosticCategory.Reliability,
+                                                                              RuleLevel.IdeSuggestion,
+                                                                              s_localizableDescription,
+                                                                              isPortedFxCopRule: false,
+                                                                              isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Descriptor);
```
Youssef1313, line 50: The other descriptor is missing here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,215 @@
...
+                            // An async alternative exists.
+                            ImmutableDictionary<string, string>? properties = ImmutableDictionary<string, string>.Empty
+                                .Add(AsyncMethodKeyName, asyncMethodName);
+
+                            Diagnostic diagnostic = Diagnostic.Create(
+                                Descriptor,
+                                invokedMethodLocation,
+                                properties,
+                                invokedMethodName.ToString(),
+                                asyncMethodName);
```
Youssef1313, line 122: Use `invocationOperation.CreateDiagnostic`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
```diff
@@ -0,0 +1,201 @@
...
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    internal static class CommonInterests
+    {
+        internal static class JoinableTask
+        {
+            internal const string TypeName = "JoinableTask";
+            internal const string Join = "Join";
+            internal const string JoinAsync = "JoinAsync";
+        }
```
stephentoub, line 21: I don't think all of this JoinableTask stuff should be here.  That's a specific library with its own set of analyzers.  This analyzer shouldn't need to know anything about JTF.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// </remarks>
+        internal static bool IsAsyncCompatibleReturnType([NotNullWhen(true)] this ITypeSymbol? typeSymbol)
+        {
+            if (typeSymbol is null)
+            {
+                return false;
+            }
+
+            // ValueTask and ValueTask<T> have the AsyncMethodBuilderAttribute
+            return (typeSymbol.Name == nameof(Task) && typeSymbol.BelongsToNamespace(Namespaces.SystemThreadingTasks))
```
stephentoub, line 187: The project already has definitions for all the task types in WellKnownTypes.  The analyzer should fetch those once at the beginning and then just compare the symbols.  You can look at other task-related analyzers in the repo to see how they do it, e.g.
https://github.com/dotnet/roslyn-analyzers/blob/086d1cc888ca0d8c96b94444935f488fcaf3fe98/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs#L144

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
```diff
@@ -0,0 +1,201 @@
...
+                || IsIAsyncEnumerable(typeSymbol) || typeSymbol.AllInterfaces.Any(IsIAsyncEnumerable)
+                || typeSymbol.GetAttributes().Any(ad => ad.AttributeClass?.Name == AsyncMethodBuilderAttribute.TypeName && ad.AttributeClass.BelongsToNamespace(AsyncMethodBuilderAttribute.Namespace));
+
+            static bool IsIAsyncEnumerable(ITypeSymbol symbol)
+                => symbol.Name == "IAsyncEnumerable"
+                && symbol.BelongsToNamespace(Namespaces.SystemCollectionsGeneric);
```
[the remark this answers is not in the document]
stephentoub: Yup. We should fetch all relevant symbols once at the beginning of the analyzer, and then just use the built-in symbol equality checking to perform these operations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            if (typeSymbol is null)
+            {
+                return false;
+            }
+
+            // ValueTask and ValueTask<T> have the AsyncMethodBuilderAttribute
+            return (typeSymbol.Name == nameof(Task) && typeSymbol.BelongsToNamespace(Namespaces.SystemThreadingTasks))
+                || IsIAsyncEnumerable(typeSymbol) || typeSymbol.AllInterfaces.Any(IsIAsyncEnumerable)
+                || typeSymbol.GetAttributes().Any(ad => ad.AttributeClass?.Name == AsyncMethodBuilderAttribute.TypeName && ad.AttributeClass.BelongsToNamespace(AsyncMethodBuilderAttribute.Namespace));
```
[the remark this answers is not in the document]
stephentoub: Can't we instead check whether the method is "async", rather than checking whether the return type is attributed?  The compiler already checks all the conditions (which is more than just the attribute) necessary to know whether the return type is valid for use with an async method.  We should be able to then just check whether the method is "async".

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.MicrosoftVisualStudioShellInterop, "IVsTask"), "Wait"), extensionMethodNamespace: Namespaces.MicrosoftVisualStudioShell),
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.MicrosoftVisualStudioShellInterop, "IVsTask"), "GetResult"), extensionMethodNamespace: Namespaces.MicrosoftVisualStudioShell),
+        });
+
+        internal static readonly IReadOnlyList<SyncBlockingMethod> SyncBlockingProperties = new[]
+        {
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.SystemThreadingTasks, nameof(Task)), nameof(Task<int>.Result)), null),
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.SystemThreadingTasks, nameof(ValueTask)), nameof(ValueTask<int>.Result)), null),
+        };
```
stephentoub, line 62: As with the other comments, we should get these as well-known symbols and just compare.  We also shouldn't restrict this to just `Task<int>` but handle `Task<T>` more generally.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
```diff
@@ -0,0 +1,201 @@
...
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.SystemThreadingTasks, nameof(Task)), nameof(Task.WaitAny)), null),
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.SystemRuntimeCompilerServices, nameof(TaskAwaiter)), nameof(TaskAwaiter.GetResult)), null),
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.SystemRuntimeCompilerServices, nameof(ValueTaskAwaiter)), nameof(ValueTaskAwaiter.GetResult)), null),
+        };
+
+        internal static readonly IEnumerable<SyncBlockingMethod> SyncBlockingMethods = JTFSyncBlockers.Concat(ProblematicSyncBlockingMethods).Concat(new[]
+        {
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.MicrosoftVisualStudioShellInterop, "IVsTask"), "Wait"), extensionMethodNamespace: Namespaces.MicrosoftVisualStudioShell),
+            new SyncBlockingMethod(new QualifiedMember(new QualifiedType(Namespaces.MicrosoftVisualStudioShellInterop, "IVsTask"), "GetResult"), extensionMethodNamespace: Namespaces.MicrosoftVisualStudioShell),
+        });
```
stephentoub, line 56: We shouldn't need any of this VS-specific stuff and can delete all of it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// </summary>
+        /// <param name="candidateMethod">The candidate method.</param>
+        /// <param name="baselineMethod">The baseline method.</param>
+        /// <returns>
+        ///   <c>true</c> if <paramref name="candidateMethod"/> has a superset of parameter types found in <paramref name="baselineMethod"/>; otherwise <c>false</c>.
+        /// </returns>
+        private static bool HasSupersetOfParameterTypes(IMethodSymbol candidateMethod, IMethodSymbol baselineMethod)
+        {
+            return candidateMethod.Parameters.All(candidateParameter => baselineMethod.Parameters.Any(baselineParameter => baselineParameter.Type?.Equals(candidateParameter.Type) ?? false));
+        }
```
stephentoub, line 143: This is a fine way to start for now.  Subsequently, though, we might want to consider augmenting it to handle common patterns between sync and async variants.  For example, a sync method might take `byte[] array, int offset, int count`, but the newer async variant might take `Memory<byte>`.

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -18,6 +18,7 @@ CA1846 | Performance | Info | PreferAsSpanOverSubstring, [Documentation](https:/
 CA1847 | Performance | Info | UseStringContainsCharOverloadWithSingleCharactersAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1847)
 CA1848 | Performance | Hidden | LoggerMessageDefineAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1848)
 CA2017 | Reliability | Warning | LoggerMessageDefineAnalyzer, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2017)
+CA2018 | Reliability | Info | UseAsyncMethodInAsyncContext, [Documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2018)
```
jeffhandley, line 21: Is `Reliability` the correct category? Or would `Performance` be more appropriate?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,219 @@
...
+                    // Don't return double-diagnostics.
+                    return;
+                }
+
+                // Also consider all method calls to check for Async-suffixed alternatives.
+                var semanticModel = context.Operation.SemanticModel;
+                SymbolInfo symbolInfo = semanticModel.GetSymbolInfo(context.Operation.Syntax, context.CancellationToken);
+
+                if (symbolInfo.Symbol is IMethodSymbol methodSymbol &&
+                    !methodSymbol.Name.EndsWith(MandatoryAsyncSuffix, StringComparison.CurrentCulture) &&
```
Youssef1313, line 91:
```suggestion
                    !methodSymbol.Name.EndsWith(MandatoryAsyncSuffix, StringComparison.Ordinal) &&
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
```diff
@@ -0,0 +1,1208 @@
...
+        Dim t As Task = Nothing
+        [|t.Wait()|]
+        Return CType(Nothing, ValueTask)
+    End Function
+End Module
+";
+            await VerifyVB.VerifyAnalyzerAsync(testVB);
+        }
+
+#if NETCOREAPP
```
Youssef1313, line 84: What issue happens if this is removed?

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
mahdiva: https://github.com/dotnet/runtime/issues/45661#issuecomment-765702030
@carlossanlop had suggested `Reliability` as the category in the api review

jeffhandley: Thanks. @carlossanlop / @stephentoub -- What do you two think about _Reliability_ vs. _Performance_ on this analyzer? The category was not discussed during the API review.

stephentoub: It's a bit of a gray area, but I'd go with performance.  Typically the negative impact of using a sync method from an async method is reduced scalability, which is closer to performance than reliability.  In a bad version of this, that sync method is actually doing async work and blocking waiting for it to complete, which brings it back closer to reliability (as that can lead to starvation and the like), but the problem there is the sync-over-async nature of the sync operation, not the fact that the async method is using the sync method.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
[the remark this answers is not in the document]
mahdiva: @stephentoub The purpose of this is to check for methods that have an async-compatible return type, such as `Task` or `ValueTask`, which means it may or may not be an async method (i.e a Task-returning sync method)

Using `typeSymbol.IsAsync` checks whether the method has the `async` keyword, which would produce a different result here.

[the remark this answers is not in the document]
stephentoub: > Using typeSymbol.IsAsync checks whether the method has the async keyword, which would produce a different result here.

Why isn't IsAsync desirable here though? i.e. what would break if we just use IsAsync instead?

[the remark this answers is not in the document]
mahdiva:
```csharp
Task Test() {
    Task t = null;
    t.Wait(); // Diagnostic: Use `await t` instead
    return Task.FromResult(1);
}
```
In the above example, using `IsAsync` on method `Test` would return false since it's a sync method. 
The goal of this analyzer is to find sync-blocking invocations (such as `t.Wait()` in this case) while also checking that the parent method or delegate (i.e. `Test`) returns a type that is Async compatible, which has the potential to be made async (such as `Task`, `ValueTask`, `IAsyncEnumberable`, etc...), regardless of whether the method is `async` or not.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -41,93 +42,162 @@ public sealed class UseAsyncMethodInAsyncContext : DiagnosticAnalyzer
...
+                ConcurrentDictionary<string, string> qualifyingTypes = new();
+                qualifyingTypes.AddOrUpdate("Task", WellKnownTypeNames.SystemThreadingTasksTask, (k, v) => v);
+                qualifyingTypes.AddOrUpdate("TaskGeneric", WellKnownTypeNames.SystemThreadingTasksTask1, (k, v) => v);
+                qualifyingTypes.AddOrUpdate("ValueTask", WellKnownTypeNames.SystemThreadingTasksValueTask, (k, v) => v);
+                qualifyingTypes.AddOrUpdate("TaskAwaiter", WellKnownTypeNames.SystemRuntimeCompilerServicesTaskAwaiter, (k, v) => v);
+                qualifyingTypes.AddOrUpdate("ValueTaskAwaiter", WellKnownTypeNames.SystemRuntimeCompilerServicesValueTaskAwaiter, (k, v) => v);
+                qualifyingTypes.AddOrUpdate("IAsyncEnumberableGeneric", WellKnownTypeNames.SystemCollectionsGenericIAsyncEnumerable1, (k, v) => v);
+                qualifyingTypes.AddOrUpdate("AsyncMethodBuilderAttribute", WellKnownTypeNames.SystemRuntimeCompilerServicesAsyncMethodBuilderAttribute, (k, v) => v);
+
+                ConcurrentDictionary<string, ISymbol> syncBlockingTypes = GetSyncBlockingTypes(context.Compilation, qualifyingTypes);
```
pgovind, line 69: Any reason you're populating a `string-string` dictionary first and then using that to populate another `string-ISymbol` dictionary here? Wouldn't it be more efficient to just create a single `string-ISymbol` dictionary here and populate it? I'm thinking you can instantiate an empty `syncBlockingTypes` here first and then create a function like below:

```csharp
private void GetTypeAndAddToDictionary(string key, string typeName, ConcurrentDictionary<string, ISymbol> syncBlockingTypes, compilation)
{
    if (compilation.TryGetOrCreateTypeByMetadataName(typeName, out INamedTypeSymbol? typeValue))
                {
                    syncBlockingTypes.AddOrUpdate(typeName, typeValue, (k, v) => v);
                }
```

Then you can just call `GetTypeAndAddToDictionary` with `Task, TaskGeneric, ValueTask, etc`

You'll save on allocating and populating `qualifyingTypes` this way.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/CommonInterests.cs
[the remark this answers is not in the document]
stephentoub: > using IsAsync on method Test would return false since it's a sync method.

Yes

> which has the potential to be made async

I didn't realize that was a goal for this analyzer as well.  I suspect this is going to hurt more than it helps, in terms of false positives incurred vs real issues found.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
```diff
@@ -0,0 +1,1542 @@
...
+
+            await vbTestVerify.RunAsync();
+        }
+
+        [Fact]
+        public async Task GenericMethodName()
+        {
+            var testCS = @"
+using System.Threading.Tasks;
+using static FruitUtils;
```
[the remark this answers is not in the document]
mahdiva: I tried removing it and the tests failed saying `Foo` doesn't exist in the current context

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,293 @@
...
+            ISymbol? memberSymbol = context.Operation.SemanticModel.GetSymbolInfo(context.Operation.Syntax, context.CancellationToken).Symbol;
+
+            if (memberSymbol is object)
+            {
+                foreach (SyncBlockingSymbol? symbol in syncBlockingSymbols)
+                {
+                    if (symbol is null || symbol.Kind != kind) continue;
+                    if (symbol.Value.Equals(memberSymbol.OriginalDefinition))
+                    {
+                        ImmutableDictionary<string, string?>? properties = ImmutableDictionary<string, string?>.Empty;
```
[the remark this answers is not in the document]
mahdiva: You are right, the nullability was not necessary so I removed it.
Not sure if `.Empty` is better, I just saw this pattern used in other analyzers so I used it too

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
```diff
@@ -0,0 +1,1542 @@
...
+    End Function
+End Module
+";
+            var vbTestVerify = new VerifyVB.Test
+            {
+                TestCode = testVB,
+                ExpectedDiagnostics = { VerifyVB.Diagnostic(UseAsyncMethodInAsyncContext.DescriptorNoAlternativeMethod).WithLocation(0).WithArguments("Wait") }
+            };
+
+            await vbTestVerify.RunAsync();
```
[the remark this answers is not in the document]
Youssef1313: @carlossanlop Both patterns exist in the repository. @sharwell Is one of the approaches preferred over the other?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
```diff
@@ -0,0 +1,1542 @@
...
+        {|#0:Task.Delay(TimeSpan.FromSeconds(5)).Wait()|};
+        yield return 1;
+    }
+}
+";
+            var csTestVerify = new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.Default.AddPackages(
+                    ImmutableArray.Create(new PackageIdentity("Microsoft.Bcl.AsyncInterfaces", "5.0.0"))
+                ),
```
Youssef1313, line 131: Would this work?

```suggestion
                ReferenceAssemblies = AdditionalMetadataReferences.DefaultWithAsyncInterfaces,
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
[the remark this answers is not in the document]
carlossanlop: Interesting. I'm trying to understand why, I would not have expected it. This is not C++ 😄 .

jeffhandley:
> @jeffhandley I am comfortable with this if it is shipped as off by default and we experiment with enabling it in the .NET 7.0 timeframe

I'm OK with that too. I forget how to mark the analyzer as off by default. @carlossanlop, can you guide @mahdiva on that please?

jeffhandley:
@jmarolf Heads-up that we are still targeting getting this into RC2.

@mahdiva -- @buyaa-n is going to take a look at this on Monday to address the remaining feedback, resolve the conflicts, and change the analyzer to be off by default.

jeffhandley (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,296 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
```
jeffhandley, line 7:
```suggestion
```
Per CI, this directive is unnecessary. I'm going to apply this and let the CI re-run.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,295 @@
...
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterOperationBlockStartAction(context =>
+            {
+                ConcurrentDictionary<string, INamedTypeSymbol> syncBlockingTypes = new();
+                GetTypeAndAddToDictionary("Task", WellKnownTypeNames.SystemThreadingTasksTask, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("TaskGeneric", WellKnownTypeNames.SystemThreadingTasksTask1, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("ValueTask", WellKnownTypeNames.SystemThreadingTasksValueTask, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("TaskAwaiter", WellKnownTypeNames.SystemRuntimeCompilerServicesTaskAwaiter, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("ValueTaskAwaiter", WellKnownTypeNames.SystemRuntimeCompilerServicesValueTaskAwaiter, syncBlockingTypes, context.Compilation);
```
buyaa-n, line 63: These two looks not used therefore not supported, should i remove them, or do they need to be supported?

buyaa-n: As per discussion with @stephentoub it's not needed for return type check, so removed

buyaa-n:
> @jeffhandley I am comfortable with this if it is shipped as off by default and we experiment with enabling it in the .NET 7.0 timeframe

@jmarolf the severity updated to `disabled` and all other comments are addressed with some refactoring, the PR is ready for final review, as you know we are still planning to add it in RC2 

CC @carlossanlop @jeffhandley @pgovind @stephentoub

src/NetAnalyzers/Core/AnalyzerReleases.Unshipped.md
```diff
@@ -4,5 +4,6 @@
 
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
+CA1849 | Performance | Disabled | UseAsyncMethodInAsyncContext, [Documentation]
```
Youssef1313, line 7: Was this file manually edited?? The codefix should have produced a link.

buyaa-n: Ah, it is just bad merging, good catch!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAsyncMethodInAsyncContextTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAsyncMethodInAsyncContextMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAsyncMethodInAsyncContextDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessageNoAlternative = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.UseAsyncMethodInAsyncContextMessage_NoAlternative), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
```
Youssef1313, line 30: This pattern was eliminated in #5406

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,282 @@
...
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.RegisterOperationBlockStartAction(context =>
+            {
+                ConcurrentDictionary<string, INamedTypeSymbol> syncBlockingTypes = new();
+                GetTypeAndAddToDictionary("Task", WellKnownTypeNames.SystemThreadingTasksTask, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("TaskGeneric", WellKnownTypeNames.SystemThreadingTasksTask1, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("ValueTask", WellKnownTypeNames.SystemThreadingTasksValueTask, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("IAsyncEnumerableGeneric", WellKnownTypeNames.SystemCollectionsGenericIAsyncEnumerable1, syncBlockingTypes, context.Compilation);
+                GetTypeAndAddToDictionary("AsyncMethodBuilderAttribute", WellKnownTypeNames.SystemRuntimeCompilerServicesAsyncMethodBuilderAttribute, syncBlockingTypes, context.Compilation);
```
Youssef1313, line 63: Why to get types in operation block start? Isn't compilation start better as this needs to be invoked only once for compilation?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,282 @@
...
+                                foreach (IMethodSymbol method in methodSymbols)
+                                {
+                                    if (!method.HasAttribute(systemObsoleteAttribute)
+                                        && HasSupersetOfParameterTypes(method, methodSymbol)
+                                        && method.Name != containingMethodName
+                                        && HasAsyncCompatibleReturnType(method, syncBlockingTypes))
+                                    {
+                                        Diagnostic diagnostic = invocationOperation.CreateDiagnostic(
+                                            Descriptor,
+                                            invokedMethodName.ToString(),
```
Youssef1313, line 129: This produces a string from SyntaxNode of the `IInvocationOperation`, which seems to be including the arguments. Consider confirming the message produce by a test, and see if the following suggestion fixes it:

```suggestion
                                            invocationOperation,TargetMethod.Name,
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        internal class SyncBlockingSymbol
+        {
+            public SyncBlockingSymbol(string Name, string Namespace, SymbolKind Kind, ISymbol Value)
+            {
+                this.Name = Name;
+                this.Namespace = Namespace;
+                this.Kind = Kind;
+                this.Value = Value;
+            }
```
Youssef1313, line 156: I'm not sure if a struct will be better or not, But was it considered?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
```diff
@@ -0,0 +1,1258 @@
...
+class Test {
+    Task T() {
+        {|#0:Foo(10, 15)|};
+        return Task.FromResult(1);
+    }
+    internal static void Foo(int x, int y) { }
+    internal static Task FooAsync(int x, int y) => null;
+}
+";
+            await CreateCSTestAndRunAsync(testCS, VerifyCS.Diagnostic(UseAsyncMethodInAsyncContext.Descriptor).WithLocation(0).WithArguments("Foo(10, 15)", "FooAsync"));
```
Youssef1313, line 473: This is what I was suggesting to test in a previous comment 😄 
I think the message feels weird with the arguments being there, esp that they can get too long and also the other argument is just a method name.

jmarolf:
To be honest I am skeptical that the general approach that this analyzer is taking can work. There are just so many places where just because two methods are named XYZAsync() and XYZ() does not mean they are semantically the same except for one being an async implementation. Like the Dispose() pattern there has been 10+ years of people creating async methods with no enforcement that the semantics make sense.

While I can understand the argument of "Well they _should_ make sense" this just isn't the reality in any large codebase (including Visual Studio). 

imho a better approach here would be to have this analyzer only operate on a known set of blessed framework methods. `File.ReadAllText` and `File.ReadAllTextAsync` are obviously equivalent to each other and we can know this. We cannot know this for XYZAsync() and XYZ(). 

this analyzer in its current form is something I am always going to recommend be off by default because of the number of false positives this approach will find. A more targeted analyzer that operates on a set of methods that we know the semantics for I can vote to turn on by default.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
buyaa-n: Good point, did not notice that

jeffhandley:
Thanks for the feedback and input, @jmarolf. I'll plan for us to spend more time iterating on this, getting usage feedback, and determining how to move the analyzer forward.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
buyaa-n: That also doesn't look that good, I am updating to use `Symbol.ToDisplayString()` instead

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContext.cs
buyaa-n: i am not sure either, so leaving as is

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
buyaa-n: I did not see your previous comment, though agree with you, now after using `symbol.ToDisplayString()` it would look better

Youssef1313: @buyaa-n If I recall correctly, I think @mavasani was preferring `symbol.Name`, but I'm not sure, so will let him confirm

carlossanlop (approved):
LGTM.

@jmarolf the analyzer is currently Disabled. Can we merge it as is? We can continue improving it afterwards if needed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/UseAsyncMethodInAsyncContextTests.cs
buyaa-n: We had some discussion in PCA messaging and `symbol.ToDisplayString()` is what we end up using

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -1452,6 +1452,18 @@ For improved performance, use the LoggerMessage delegates.
...
 
+## [CA1849](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1849): Call async methods when in an async method
+
+When inside a Task-returning method, use the async version of methods, if they exist.
+
+|Item|Value|
+|-|-|
+|Category|Performance|
+|Enabled|False|
+|Severity|Warning|
```
buyaa-n, line 1463: Hm, this is strange, i did not update it manually `msbuild /t:pack` updated this, but i guess Severity should no be as a Warning for disabled analyzers, should I update the severity manually? @mavasani @jmarolf

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.sarif
```diff
@@ -2701,6 +2701,26 @@
             ]
           }
         },
+        "CA1849": {
+          "id": "CA1849",
+          "shortDescription": "Call async methods when in an async method",
+          "fullDescription": "When inside a Task-returning method, use the async version of methods, if they exist.",
+          "defaultLevel": "warning",
```
buyaa-n, line 2708: huh, this one looks wrong too @mavasani @jmarolf

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
buyaa-n: oh i see all other `Disabled` analyzers has same value, guess it is expected then

buyaa-n:
> To be honest I am skeptical that the general approach that this analyzer is taking can work. There are just so many places where just because two methods are named XYZAsync() and XYZ() does not mean they are semantically the same except for one being an async implementation. Like the Dispose() pattern there has been 10+ years of people creating async methods with no enforcement that the semantics make sense.
> 
> While I can understand the argument of "Well they should make sense" this just isn't the reality in any large codebase (including Visual Studio).

As you mentioned this exists in VS and was requested by internal partners to be added into SDK which I think shows it was useful. We all agree that it will be noisy and we are planning to make improvements in 7.0, `Dispose() pattern` example you mentioned could be covered there

> imho a better approach here would be to have this analyzer only operate on a known set of blessed framework methods.

For me, its might be better to exclude known exceptions from the diagnostics, but i have no strong opinion here 

> this analyzer in its current form is something I am always going to recommend be off by default because of the number of false positives this approach will find. A more targeted analyzer that operates on a set of methods that we know the semantics for I can vote to turn on by default.

Yes, we disabled the analyzer as you suggested, please take another look

jeffhandley:
> Thanks for the feedback and input, @jmarolf. I'll plan for us to spend more time iterating on this, getting usage feedback, and determining how to move the analyzer forward.

I realized this comment might have been unclear on the intent.  I would still like to include this analyzer in RC2, and then proactively seek early adopters of it who can provide feedback to us to measure the true-positive vs. false-positive ratio with the current design.  We can then iterate on the implementation during the 7.0 cycle based on feedback.

jmarolf (approved):

PR #5373, opened 2021-08-13
CA2018: Do not throw a diagnostic for Boolean arrays

### LINKED ISSUE #5353: Improve CA2017 to detect bool[]
https://github.com/dotnet/roslyn-analyzers/pull/5242/files implements CA2017 to tag calls to `Buffer.BlockCopy` with a warning about the last `count` parameter. Currently the analyzer excludes `Buffer.BlockCopy` calls that pass in `byte[]` or `sbyte[]` as the source and destination parameters because `count` makes sense for these calls. Source and destination parameters of type `bool[]` must also be excluded from this check.

cc @mahdiva @jeffhandley @buyaa-n

mahdiva:
Fixes #5353 

CA2018 currently only checks for `Byte` and `SByte` array types. `Boolean` should also be included, since it has a size of 1 byte.
This PR adds support for `bool[]` arrays to prevent a false positive.

cc @pgovind @Youssef1313


buyaa-n (approved):
LGTM, thanks!

Youssef1313 (approved):

pgovind:
Thanks for the reviews @buyaa-n  and @Youssef1313

PR #5377, opened 2021-08-16
Detect missing .ConfigureAwait in 'await foreach'

### LINKED ISSUE #3703: ConfigureAwait diagnostic + code fix for IAsyncEnumerable (and other types)
Could the existing ConfigureAwait refactoring be loosened to offer this refactoring on any type that has a .ConfigureAwait(bool) instance or extension method in scope?

IAsyncEnumerable and IAsyncDisposable are two that were recently added to the BCL. It seems like all future types and user-defined types could be treated uniformly.

#### Analyzer package

#### Package Version

v3.0.0 (Latest)

#### Diagnostic ID

Example: [CA2007](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2007)

manfred-brands:
Fixes #3703 

Added detection of missing `.ConfigureAwait` when using `await foreach`

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
@@ -20,7 +20,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
     public sealed class DoNotDirectlyAwaitATaskAnalyzer : DiagnosticAnalyzer
     {
         internal const string RuleId = "CA2007";
-
+        private const string ConfigureAwait = "ConfigureAwait";
         public static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
             RuleId,
             CreateLocalizableResourceString(nameof(DoNotDirectlyAwaitATaskTitle)),
@@ -53,6 +53,7 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                 }
 
                 var configuredAsyncDisposable = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeCompilerServicesConfiguredAsyncDisposable);
+                var configuredCancelableAsyncEnumerable = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemRuntimeCompilerServicesConfiguredCancelableAsyncEnumerable);
 
                 context.RegisterOperationBlockStartAction(context =>
                 {
@@ -77,6 +78,10 @@ namespace Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines
                             context.RegisterOperationAction(context => AnalyzeUsingOperation(context, configuredAsyncDisposable), OperationKind.Using);
                             context.RegisterOperationAction(context => AnalyzeUsingDeclarationOperation(context, configuredAsyncDisposable), OperationKindEx.UsingDeclaration);
                         }
+                        if (configuredCancelableAsyncEnumerable is not null)
+                        {
+                            context.RegisterOperationAction(context => AnalyzeForEachOperation(context, configuredCancelableAsyncEnumerable), OperationKind.Loop);
+                        }
                     }
                 });
             });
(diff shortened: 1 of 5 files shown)


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATaskTests.cs
```diff
@@ -741,5 +741,202 @@ async Task CoreAsync()
...
+            sum += item;
+        return sum;
+    }
+}
+";
+
+            await new VerifyCS.Test
+            {
+                ReferenceAssemblies = ReferenceAssemblies.Default.AddPackages(
+                    ImmutableArray.Create(new PackageIdentity("Microsoft.Bcl.AsyncInterfaces", "5.0.0"))),
```
sharwell, line 936: ❔  Can we use `ReferenceAssemblies.NetCore.NetCoreApp31` instead?

manfred-brands: Microsoft.Bcl.AsyncInterfaces is the smallest additional package to support IAsyncEnumerable, and even works on .net framework.
It also matches the package you used for the 'await using' unit tests.

MartyIX:
Will this be available in .NET 6+?

Thanks.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
```diff
@@ -96,6 +101,70 @@ private static void AnalyzeAwaitOperation(OperationAnalysisContext context, Immu
...
+
+            if (!operation.IsAsynchronous())
+            {
+                return;
+            }
+
+            // Get the type of the expression being iterated over.
+            IOperation collectionOperation = operation.Collection;
+            ITypeSymbol? typeOfCollection = collectionOperation.Type;
+            if (typeOfCollection.OriginalDefinition == configuredCancelableAsyncEnumerable)
```
sharwell, line 118:
```suggestion
            if (Equals(typeOfCollection.OriginalDefinition, configuredCancelableAsyncEnumerable))
```

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
```diff
@@ -96,6 +101,70 @@ private static void AnalyzeAwaitOperation(OperationAnalysisContext context, Immu
...
+                                                               .FindNode(location.SourceSpan);
+                        if (node is not null)
+                        {
+                            // Checking VariableDeclaratorSyntax.Initializer is CSharp specific
+                            // Do we need to move rule to CSharp.NetAnalyzers?
+                            if (node.ToString().Contains(ConfigureAwait))
+                            {
+                                return;
+                            }
+                        }
```
sharwell, line 137: 💭 This check should be possible using the operation tree directly.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
```diff
@@ -96,6 +101,70 @@ private static void AnalyzeAwaitOperation(OperationAnalysisContext context, Immu
...
+                        // the TaskAsyncEnumerableExtensions extension method.
+                        if (invocationOperation.TargetMethod.Name == "WithCancellation")
+                        {
+                            invocationOperation = invocationOperation.Instance as IInvocationOperation;
+                        }
+
+                        if (invocationOperation?.TargetMethod.Name == ConfigureAwait)
+                        {
+                            return;
+                        }
```
sharwell, line 160: ❓ Is it not possible to just check the return types?

sharwell (changes requested):

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs
manfred-brands: Yes, found a way to obtain `IVariableDeclaratorOperation` and from there the `IVariableInitializerOperation`.

manfred-brands:
Rebased on main to fix merge conflict.

mavasani:
@sharwell Can you please review?

PR #5383, opened 2021-08-17
CA1848: Avoid const arrays

steveberdy:
## CA1848: Avoid `const` arrays

This analyzer is for performance improvements requested in https://github.com/dotnet/runtime/issues/33794. The purpose of the analyzer and code fix is to extract constant array literals, that are passed into methods, as `static readonly` fields.

Fixes in .NET repos:
- [ ] runtime - `In progress`
- [ ] roslyn - `In progress`
- [ ] roslyn-analyzers - `In progress`
- [ ] aspnetcore - `In progress`

Documentation: https://github.com/dotnet/docs/pull/25659

Fixes https://github.com/dotnet/runtime/issues/33794
cc @jeffhandley @terrajobst


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -246,6 +246,17 @@
   <data name="AvoidUnsealedAttributesMessage" xml:space="preserve">
     <value>Avoid unsealed attributes</value>
   </data>
+  <data name="AvoidConstArraysTitle" xml:space="preserve">
+    <value>Avoid const arrays</value>
+  </data>
+  <data name="AvoidConstArraysDescription" xml:space="preserve">
+    <value>Having a constant array passed as an argument is not ideally performant. Extract constant arrays as 'static readonly'.</value>
+    <comment>{Locked="static", "readonly"}</comment>
```
Youssef1313, line 254: I think it doesn't work like that. The syntax for multiple words is `{Locked="FirstWord"}{Locked="SecondWord"}. But there is no need to separate them.

```suggestion
    <comment>{Locked="static readonly"}</comment>
```

steveberdy: Ok, done. I'll be working on reformatting the specific node rather than the whole document:
https://github.com/dotnet/roslyn-analyzers/blob/6af52ad4d3067b9a0237cc5993033ba682266082/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs#L102
That's the only other comment I did not address, except for the `s_` prefix for `static` fields. I believe that, while that is a .NET source code style, it doesn't seem right to enforce a snake case formatting for field extractions.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,338 @@
...
+
+    public void B()
+    {
+        Console.WriteLine(value);
+    }
+}
+");
+
+            // Explicit initialization check
+            await VerifyCS.VerifyCodeFixAsync(@"
```
Evangelink, line 47: [suggestion] Could we group the test by "function" instead of having these group comments? Also, merging C# and VB.NET in the same functional test.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,338 @@
...
+    Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer>;
+
+namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
+{
+    public class AvoidConstArraysTests
+    {
+        #region C# Tests
+
+        [Fact]
+        public async Task CA1849_CSharp_IdentifyConstArrays()
```
Evangelink, line 19: Test name and inline diagnostic checks are usually agnostic of the analyzer ID. I don't know if that matters much to be honest.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,338 @@
...
+        {
+            // Implicit initialization check
+            await VerifyCS.VerifyCodeFixAsync(@"
+using System;
+
+public class A
+{
+    public void B()
+    {
+        Console.WriteLine({|CA1849:new[]{ 1, 2, 3 }|});
```
Evangelink, line 29: Could you please add a test for C# and VB.NET where there are some specific trivias (comment, maybe line break) to see if the fixer is doing what we would expect.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,338 @@
...
+using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysAnalyzer,
+    Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer>;
+using VerifyVB = Test.Utilities.VisualBasicCodeFixVerifier<
+    Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysAnalyzer,
+    Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer>;
+
+namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
+{
+    public class AvoidConstArraysTests
```
Evangelink, line 14: Here are a couple of extra tests you could add:

- specific trivias (comment, maybe line break) to see if the fixer is doing what we would expect.
- array using constants defined in variable (local and fields)
- test where values are `static readonly` (so not const)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,94 @@
...
+                    }
+                    arrayCreationOperation = (IArrayCreationOperation)conversionOperation.Operand;
+                }
+                else
+                {
+                    return;
+                }
+
+                // Must be literal array
+                if (!arrayCreationOperation.Children.First(x => x is IArrayInitializerOperation).Children.All(x => x is ILiteralOperation))
```
Evangelink, line 78: @sharwell Is it fine to "manually" look up for children of some type?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
steveberdy: Sure. I'm on it.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
steveberdy: Done. Thanks for the feedback @Evangelink

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
steveberdy: I'll update this. I didn't look closely about the properties that were included, and I can clean this up and make it more performant.

steveberdy: Updated to this:
https://github.com/dotnet/roslyn-analyzers/blob/9f0e5ac03e33c510231389e396d24f6ce812178a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs#L78

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
steveberdy: @Evangelink Done! I added a test for your first point. For your third point I already tested non-literal arrays in the NoDiagnostic method. For the third, the only values we look at are local -- those included in local invocations or arguments -- so I added no test for that. I'll make sure that my tests pass.

steveberdy: @Evangelink all of the tests I added according to your feedback passed.

steveberdy:
@Youssef1313 I finished work on all feedback, and even included more tests to cover various scenarios. Do you have any more feedback for me? Is this mergeable?

Evangelink:
@steveberdy You will need to run `msbuild /t:pack` as stated in the build error (and PR comments):

`One or more auto-generated documentation files were either edited manually, or not updated. Please revert changes made to the following files (if manually edited) and run `msbuild /t:pack` for each solution at the root of the repo to automatically update them:`

We will then need to wait for a review from someone at MS. Thanks for the hardwork in the meantime :)

jeffhandley:
I went ahead and ran `dotnet msbuild /t:Pack` and pushed to the branch.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -246,6 +246,17 @@
...
+    <value>Avoid constant arrays as arguments</value>
+  </data>
+  <data name="AvoidConstArraysDescription" xml:space="preserve">
+    <value>Having a constant array passed as an argument is not ideally performant. Extract constant arrays as 'static readonly'.</value>
+    <comment>{Locked="static readonly"}</comment>
+  </data>
+  <data name="AvoidConstArraysMessage" xml:space="preserve">
+    <value>Make any constant arrays passed as arguments into 'static readonly' fields</value>
+    <comment>{Locked="static readonly"}</comment>
+  </data>
```
stephentoub, line 259: Thanks for working on this.

I've not reviewed the PR in depth, just enough to see that I think something important to cover hasn't been covered yet.  I'll be curious to see what gets flagged in the dotnet/runtime repo as a result of running this.

If someone has this code:
```C#
ReadOnlySpan<byte> span = new byte[] { 1, 2, 3, 4, 5 };
```
the C# compiler will optimize that to not actually creating an array, but instead storing the data in the binary and just creating a span over that data's address/length.  In such situations, recommending it be extracted to a static readonly byte[] would actually be a deoptimization.  So at least for certain types of arrays, the analyzer would need to to target type and check whether it's being assigned to a `ReadOnlySpan<T>`.

Today those `T`s can be `byte`, `sbyte`, and `bool`, but I'm hopeful that for .NET 7 it'll accommodate any `T` primitive type.  So, it might make sense to avoid firing this suggestion if the target type is `ReadOnlySpan<T>` and `T` is a primitive.

steveberdy: I added a check for this. When there is a parameter taking in a `ReadOnlySpan` argument, it ignores the argument.

buyaa-n:
@steveberdy  please retarget to [release/7.0.1xx](https://github.com/dotnet/roslyn-analyzers/tree/release/7.0.1xx) branch, release `release/6.0.1xx` now closed for new analyzers.
 
[More info](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/NetCore_GettingStarted.md#branch-definitions)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            // Analyzes an argument operation
+            context.RegisterOperationAction(operationContext =>
+            {
+                IArgumentOperation? argumentOperation;
+                string readonlySpanTypeString = WellKnownTypeProvider.GetOrCreate(operationContext.Compilation)
+                    .GetOrCreateTypeByMetadataName(s_readonlySpanTypeName)!.Name;
```
Youssef1313, line 49: Better to keep it as a symbol instead of getting the name, and do a symbol equality instead of string equality.

steveberdy: https://github.com/dotnet/roslyn-analyzers/blob/124afdcbb69d45e95794c8f289d3fc1f44c26518/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs#L96

When I check the type of the parameter, it's `INamedTypeSymbol` doesn't always match up with the well-known type, but the name of it always should.

Youssef1313: @stevenbrix I think that means you're somehow doing the comparison incorrectly. Probably you need `argumentOperation.Parameter.Type.OriginalDefinition.Equals(readOnlSpanSymbol)`. (The `OriginalDefinition` here gets the type before it was constructed, ie, if the type is for example `ReadOnlySpan<string>`, the original definition will be `ReadOnlySpan<T>`, which is what `GetOrCreateTypeByMetadataName` returns.

The problem with comparing type names is that the user can define a type with the same name. So symbol equality is always more reliable.

steveberdy: Thanks. I'll work on those changes.

steveberdy:
@jeffhandley Hi, could you or someone from your team look this PR over for merging? Thanks

jeffhandley:
Thanks, @steveberdy! We will review as soon as possible, but it might be a week or two more while we focus on remaining 6.0 deliverables and duties. Thanks for your contribution and patience. If any conflicts emerge because of our delay, we will help resolve them.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            // Analyzes an argument operation
+            context.RegisterOperationAction(operationContext =>
+            {
+                IArgumentOperation? argumentOperation;
+                INamedTypeSymbol readonlySpanType = WellKnownTypeProvider.GetOrCreate(operationContext.Compilation)
+                    .GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemReadOnlySpan1)!;
```
Youssef1313, line 47: The null suppression isn't safe here. Also it would be better to retrieve the type in compilation start instead of every operation.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// CA1849: Avoid constant arrays as arguments. Replace with static readonly arrays.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AvoidConstArraysAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1849";
+
+        private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableMessage = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysMessage), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
+        private static readonly LocalizableString s_localizableDescription = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysDescription), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
```
Youssef1313, line 24: This pattern was changed in https://github.com/dotnet/roslyn-analyzers/pull/5406.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,181 @@
...
+            SemanticModel model = await document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            DocumentEditor editor = await DocumentEditor.CreateAsync(document, context.CancellationToken).ConfigureAwait(false);
+            SyntaxGenerator generator = editor.Generator;
+
+            context.RegisterCodeFix(
+                new MyCodeAction(
+                    MicrosoftNetCoreAnalyzersResources.AvoidConstArraysTitle,
+                    async c => await ExtractConstArrayAsync(node, model, editor, generator, context.Diagnostics.First().Properties, c).ConfigureAwait(false),
+                    equivalenceKey: nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysTitle)),
+                context.Diagnostics);
```
Youssef1313, line 46: Codefix titles should usually be different than analyzer titles. The codefix title should say what does this fixer do (e.g, Replace constant array with a static readonly array) or something like that.

Youssef1313 (commented):
For empty arrays this could conflict with CA1825 suggestion. I'm not sure if we should skip zero length arrays for this analyzer or not.

steveberdy:
@Youssef1313 there wouldn't be any conflicts. Any occurrences of empty arrays as arguments would be extracted. Then the CA1825 analyzer may apply once that is extracted, and it will become `Array.Empty<T>`. There shouldn't be conflicts, as this stays strictly in its own scope.

Youssef1313:
> @Youssef1313 there wouldn't be any conflicts. Any occurrences of empty arrays as arguments would be extracted. Then the CA1825 analyzer may apply once that is extracted, and it will become `Array.Empty<T>`. There shouldn't be conflicts, as this stays strictly in its own scope.

Yeah indeed it will still apply after it's extracted, my bad.

steveberdy:
Do you know why this was rebased to main?

Youssef1313:
@steveberdy I believe `main` is now used for the .NET 7 released. The `release/7.0.1xx` was actually deleted.

steveberdy:
Alright thanks. I'll resolve the conflicts.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,182 @@
...
+            if (lastFieldSymbol is not null)
+            {
+                // Insert after last field if any fields are present
+                SyntaxNode lastFieldNode = await lastFieldSymbol.DeclaringSyntaxReferences.First().GetSyntaxAsync(cancellationToken).ConfigureAwait(false);
+                editor.InsertAfter(generator.GetDeclaration(lastFieldNode), newMember);
+            }
+            else
+            {
+                // Insert before first method if no fields are present, as a method already exists
+                editor.InsertBefore(containingMethodBody.Syntax, newMember);
```
Youssef1313, line 86: How is this "first method"?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,125 @@
...
+    /// CA1850: Avoid constant arrays as arguments. Replace with static readonly arrays.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class AvoidConstArraysAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1850";
+
+        private static readonly LocalizableString s_localizableTitle = CreateLocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysTitle));
+        private static readonly LocalizableString s_localizableMessage = CreateLocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysMessage));
+        private static readonly LocalizableString s_localizableDescription = CreateLocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysDescription));
```
Youssef1313, line 26: This pattern was changed in https://github.com/dotnet/roslyn-analyzers/pull/5406.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,125 @@
...
+        internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            s_localizableTitle,
+            s_localizableMessage,
+            DiagnosticCategory.Performance,
+            RuleLevel.IdeSuggestion,
+            s_localizableDescription,
+            isPortedFxCopRule: false,
+            isDataflowRule: false);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
```
Youssef1313, line 37: I think this allocates a new array every time the property is accessed. Should be:

```suggestion
        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(compilationContext =>
```
Youssef1313, line 44: Rename `compilationContext` and `operationContext` to just `context`.
The shadowing is intentional since we never need to access the outer `context`, and it sometimes can be wrong to do so (e.g, we should use cancellation token from the most inner context)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
steveberdy: Hmmm, I have no idea what I was thinking. That feature is still important, so I'll have to get all method symbols, unless there is a better way to do it that you can think of.

Youssef1313: @steveberdy actually we're looking for methods here? Why not put it at the top of the class? ie, before the first class member?

steveberdy: Well, members can refer to fields, properties, and methods. I think there is a way to get all members, but I'll have to check. My logic here, as I remember, was that by that block there would be no class fields, and therefore I'd have to look for some other type of member. It isn't perfect logic looking back, so I'll jump in and see what I can do.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,125 @@
...
+                }
+
+                // Analyzes an argument operation
+                compilationContext.RegisterOperationAction(operationContext =>
+                {
+                    IArgumentOperation? argumentOperation;
+
+                    if (operationContext.Operation is IArrayCreationOperation arrayCreationOperation) // For arrays passed as arguments
+                    {
+                        argumentOperation = arrayCreationOperation.GetAncestor<IArgumentOperation>(OperationKind.Argument);
```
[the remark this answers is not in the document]
steveberdy: @Youssef1313 did you mean to ping me instead?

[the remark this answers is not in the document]
Youssef1313: @steveberdy Indeed 😄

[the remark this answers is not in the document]
steveberdy: > What if, for example, the argument is a lambda, and the array creation is somewhere in the lambda? Is it intentional to report a diagnostic for this?

It will be reporting this. If the argument is just a function that always returns a constant array, it should be reported, but it should be noted that the actual argument type is a function.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,133 @@
...
+    public sealed class AvoidConstArraysAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA1850";
+
+        internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(RuleId,
+            CreateLocalizableResourceString(AvoidConstArraysTitle),
+            CreateLocalizableResourceString(AvoidConstArraysMessage),
+            DiagnosticCategory.Performance,
+            RuleLevel.IdeSuggestion,
+            CreateLocalizableResourceString(AvoidConstArraysDescription),
```
Youssef1313, line 29: You're missing a `nameof(..)` for `CreateLocalizableResourceString` calls. That's why MSBuild didn't find these strings :)

steveberdy: Thanks for catching that!

buyaa-n (commented):
Thanks, @steveberdy I have two main questions:
1. ~~From the [comment ](https://github.com/dotnet/runtime/issues/33794#issuecomment-634272979) `The hardest part of this one will be coming up with the list of APIs we want this to target--using a heuristic is the recommended approach` seems we want to flag it only for specific APIs, is this requirement changed @jeffhandley @stephentoub?~~ EDIT: This requirement outdated, please ignore
2. Did you run the analyzer in runtime and/or any other repo? Did it catch anything? Do you need help with that?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -247,7 +247,7 @@
     <value>Avoid unsealed attributes</value>
   </data>
   <data name="AvoidConstArraysTitle" xml:space="preserve">
-    <value>Avoid constant arrays as arguments</value>
+    <value>Extract to static readonly field</value>
   </data>
```
buyaa-n: I meant we need to use a separate title for the fixer, lets leave `Avoid constant arrays as arguments` as is for analyzer title, please add another resource for fixer title
```xml
  <data name="AvoidConstArraysCodeFixTitle" xml:space="preserve">
    <value>Extract to static readonly field</value>
  </data>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,133 @@
...
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(context =>
+            {
+                INamedTypeSymbol? readonlySpanType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemReadOnlySpan1);
+                INamedTypeSymbol? functionType = context.Compilation.GetOrCreateTypeByMetadataName("System.Func`2");
+                if (readonlySpanType is null || functionType is null)
```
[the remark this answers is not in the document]
steveberdy: I'll ignore the types, meaning I won't check if they match, if they are null. That should work instead of bailing out.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
steveberdy: I made that change. Thank you

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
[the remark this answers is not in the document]
steveberdy: @Youssef1313 checkout https://github.com/dotnet/roslyn-analyzers/pull/5383/commits/47a0232a861dbe02243e8c9198dd94f3a0659bff and tell me what you think.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,444 @@
...
+        {
+            // Implicit initialization check
+            await VerifyCS.VerifyCodeFixAsync(@"
+using System;
+
+public class A
+{
+    public void B()
+    {
+        Console.WriteLine({|CA1850:new[]{ 1, 2, 3 }|});
```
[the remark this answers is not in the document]
steveberdy: It is needed to specify a span. I removed it and got test failures.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,444 @@
...
+
+            // Explicit initialization check
+            await VerifyCS.VerifyCodeFixAsync(@"
+using System;
+
+public class A
+{
+    public void B()
+    {
+        Console.WriteLine({|CA1850:new int[]{ 1, 2, 3 }|});
```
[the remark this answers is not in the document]
steveberdy: The API does accept it, it just prints out the type of the argument, `System.Int32[]`, rather than the contents of the array. I just wanted an example that would accept an array as an argument.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,178 @@
...
+                return arrayCreation;
+            }
+            // Otherwise, we'll get the IArrayCreationOperation from the argument node's child
+            isInvoked = false;
+            return (IArrayCreationOperation)model.GetOperation(node.ChildNodes().First(), cancellationToken);
+        }
+
+        private static string GetExtractedMemberName(IEnumerable<string> memberNames, string parameterName)
+        {
+            bool hasCollectionEnding = s_collectionMemberEndings.Any(x => parameterName.EndsWith(x, true, null));
```
buyaa-n, line 122:
```suggestion
            bool hasCollectionEnding = s_collectionMemberEndings.Any(x => parameterName.EndsWith(x, true, CultureInfo.InvariantCulture));
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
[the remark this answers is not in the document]
buyaa-n: Normally you could be able to use square brackets to mark the location `[|new[]{ 1, 2, 3 }|]`, but somehow it seems doesn't work in your VB tests

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
steveberdy: Looks like I fixed that a while back so I'm resolving. Thanks

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        Console.WriteLine({ str })
+    End Sub
+End Class
+");
+        }
+
+        [Fact]
+        public async Task IgnoreReadonlySpan_NoDiagnostic()
+        {
+
```
Evangelink, line 451: [cosmetic] Un-needed blank line.

```suggestion
```

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,508 @@
...
+    Public Sub B()
+        Dim y As String = stringArray.First()
+        Console.WriteLine(y)
+    End Sub
+End Class
+");
+        }
+
+        [Fact]
+        public async Task IdentifyConstArrays_MemberExtractionTest()
```
Evangelink, line 305: Don't know if it's better to update this test or add another one where there would be more than one conflict for the field name.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,132 @@
...
+                    {
+                        return;
+                    }
+
+                    if (argumentOperation is not null)
+                    {
+                        ITypeSymbol originalDefinition = argumentOperation.Parameter.Type.OriginalDefinition;
+
+                        // Can't be a ReadOnlySpan, as those are already optimized
+                        if (readonlySpanType is not null && originalDefinition.Equals(readonlySpanType))
```
Evangelink, line 101: You probably want to use `SymbolEqualityComparer` for the symbols equality.

```suggestion
                        if (SymbolEqualityComparer.Default.Equals(readonlySpanType, originalDefinition))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                    // Must be literal array
+                    if (arrayCreationOperation.Initializer.ElementValues.Any(x => x is not ILiteralOperation))
+                    {
+                        return;
+                    }
+
+                    Dictionary<string, string?> properties = new()
+                    {
+                        { "paramName", isDirectlyInsideLambda ? null : argumentOperation?.Parameter?.Name }
```
Evangelink, line 122: You are assuming that `Parameter` can be null here but not on line `98`, I am wondering if one place needs to be updated.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,132 @@
...
+                    };
+
+                    context.ReportDiagnostic(arrayCreationOperation.CreateDiagnostic(Rule, properties.ToImmutableDictionary()));
+                },
+                OperationKind.ArrayCreation,
+                OperationKind.Invocation);
```
Evangelink, line 128: It's definitely minor but I am wondering if code wouldn't be clearer with 2 calls to `RegisterOperationAction` (one for each operation kind) with the 2 parts calling the same method for the common code.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -246,6 +246,20 @@
...
+  </data>
+  <data name="AvoidConstArraysCodeFixTitle" xml:space="preserve">
+    <value>Extract to static readonly field</value>
+  </data>
+  <data name="AvoidConstArraysDescription" xml:space="preserve">
+    <value>Having a constant array passed as an argument is not ideally performant. Extract constant arrays as 'static readonly'.</value>
+    <comment>{Locked="static readonly"}</comment>
+  </data>
+  <data name="AvoidConstArraysMessage" xml:space="preserve">
+    <value>Make any constant arrays passed as arguments into 'static readonly' fields</value>
```
Evangelink, line 260: What about `Prefer 'static readonly' fields over local constant arrays arguments`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -246,6 +246,20 @@
...
     <value>Avoid unsealed attributes</value>
   </data>
+  <data name="AvoidConstArraysTitle" xml:space="preserve">
+    <value>Avoid constant arrays as arguments</value>
+  </data>
+  <data name="AvoidConstArraysCodeFixTitle" xml:space="preserve">
+    <value>Extract to static readonly field</value>
+  </data>
+  <data name="AvoidConstArraysDescription" xml:space="preserve">
+    <value>Having a constant array passed as an argument is not ideally performant. Extract constant arrays as 'static readonly'.</value>
```
Evangelink, line 256: What about `Constant arrays passed as arguments are not reused which implies a performance overhead. Consider extracting them into 'static readonly' fields to improve performance.`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        private static async Task<Document> ExtractConstArrayAsync(SyntaxNode node, SemanticModel model, DocumentEditor editor,
+            SyntaxGenerator generator, ImmutableDictionary<string, string> properties, CancellationToken cancellationToken)
+        {
+            IArrayCreationOperation arrayArgument = GetArrayCreationOperation(node, model, cancellationToken, out bool isInvoked);
+            IEnumerable<ISymbol> members = model.GetEnclosingSymbol(node.SpanStart, cancellationToken).ContainingType.GetMembers();
+
+            // Get a valid member name for the extracted constant
+            string newMemberName = GetExtractedMemberName(
+                members.Where(x => x is IFieldSymbol).Select(x => x.Name),
```
Evangelink, line 59: @Youssef1313 Do you recall if there was a real performance improvement when checking kind over `is XXX`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,170 @@
...
+                properties["paramName"] ?? GetMemberNameFromType(arrayArgument));
+
+            // Get method containing the symbol that is being diagnosed.
+            IOperation? methodContext = arrayArgument.GetAncestor<IMethodBodyOperation>(OperationKind.MethodBody);
+            methodContext ??= arrayArgument.GetAncestor<IBlockOperation>(OperationKind.Block); // VB methods have a different structure than CS methods
```
Evangelink, line 64:
```suggestion
            IOperation? methodContext = arrayArgument.GetAncestor<IMethodBodyOperation>(OperationKind.MethodBody)
                ?? arrayArgument.GetAncestor<IBlockOperation>(OperationKind.Block); // VB methods have a different structure than CS methods
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,170 @@
...
+            );
+
+            // Add any additional formatting
+            if (methodContext is not null)
+            {
+                newMember = newMember.FormatForExtraction(methodContext.Syntax);
+            }
+
+            // Insert the new extracted member before the first member
+            SyntaxNode firstMemberNode = await members.First().DeclaringSyntaxReferences.First()
```
Evangelink, line 82: I'm usually trying to play defensive and try to favor `FirstOrDefault` over `First` just in case we have an invalid tree.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (parameterName == "source" // for LINQ, "sourceArray" is clearer than "source"
+                || (memberNames.Contains(parameterName) && !hasCollectionEnding))
+            {
+                parameterName += "Array";
+            }
+
+            if (memberNames.Contains(parameterName))
+            {
+                int suffix = 0;
+                while (memberNames.Contains(parameterName + suffix))
```
Evangelink, line 134: Probably unlikely that it fails but maybe we should also put a condition on suffix.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,170 @@
...
+        private static string GetMemberNameFromType(IArrayCreationOperation arrayCreationOperation)
+        {
+#pragma warning disable CA1308 // Normalize strings to uppercase
+            return ((IArrayTypeSymbol)arrayCreationOperation.Type).ElementType.OriginalDefinition.Name.ToLowerInvariant() + "Array";
+#pragma warning restore CA1308 // Normalize strings to uppercase
+        }
+
+        private static Accessibility GetAccessibility(ISymbol? methodSymbol)
+        {
+            if (methodSymbol is not null && Enum.TryParse(methodSymbol.GetResultantVisibility().ToString(), out Accessibility accessibility))
```
Evangelink, line 153: Isn't it better/safer to switch on the result of `GetResultantVisibility` rather than relying on string equality?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
steveberdy: What do you mean?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
steveberdy: For some reason that actually doesn't work, since `??` must be applied to operands of the same type. Using `??=` works fine since it just assigns if null, and `IBlockOperation` can be assigned to an `IOperation`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(context =>
+            {
+                INamedTypeSymbol? readonlySpanType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemReadOnlySpan1);
+                INamedTypeSymbol? functionType = context.Compilation.GetOrCreateTypeByMetadataName("System.Func`2");
```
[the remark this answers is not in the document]
steveberdy: Yeah `Func` should definitely be there. I'll add it.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
steveberdy: That would result in some duplicate code that isn't necessary to split up. I'm actually cleaning it up and adding more comments to clarify the control flow. Whether the operation is an array creation or an invocation, what is important is that we have an array creation involved with some sort of argument, and then we report a diagnostic from that array creation given it meets other criteria. Tell me what you think.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
steveberdy: That's already in the test.

steveberdy:
@buyaa-n I've gotten this to a really good stage, and now I'm trying to apply the analyzer to dotnet repos. Do you know of a good way to do that? Is VS required or can I use vscode?

buyaa-n:
> @buyaa-n I've gotten this to a really good stage, and now I'm trying to apply the analyzer to dotnet repos. Do you know of a good way to do that? Is VS required or can I use vscode?

Great, thanks! The instructions are here: https://github.com/dotnet/roslyn-analyzers/blob/main/docs/NetCore_GettingStarted.md#testing-against-the-runtime-and-roslyn-analyzers-repo, please let me know if you got any questions with the doc

steveberdy:
I ran the build in runtime, and it stopped at the first warning. How do you recommend I run it to output all warnings to a text file like you did [here](https://github.com/dotnet/roslyn-analyzers/files/8384413/avoid.const.arrays.analyzer.findings.txt)?

stephentoub:
Add `-warnaserror 0` to the command

steveberdy:
When it ran `src\coreclr\build-runtime.cmd -x64 -debug -os windows` it said that I need Visual Studio 2019 or 2022 with C++ tools. Is there a way to build runtime without VS?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
Evangelink: Sorry I meant that there is no upper bound check for the int value of suffix. As I said, unlikely we would exceed int value.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                // If private or public, return private since public accessibility not wanted for fields by default
+                return methodSymbol.GetResultantVisibility() switch
+                {
+                    // Return internal if internal
+                    SymbolVisibility.Internal => Accessibility.Internal,
+                    _ => Accessibility.Private
+                };
+            }
+            return Accessibility.Private;
```
Evangelink, line 169: [minor] You can merge the if and the switch.

```suggestion
            // If private or public, return private since public accessibility not wanted for fields by default
            return methodSymbol?.GetResultantVisibility() switch
            {
                // Return internal if internal
                SymbolVisibility.Internal => Accessibility.Internal,
                _ => Accessibility.Private
            };
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
steveberdy: Should I make `suffix` a `uint` type?

Evangelink: No I think that's ok, if we exceed int limit that means we are in a really weird scenario so it's probably ok to fail.

steveberdy:
Can we check this over for merging again?

steveberdy:
Does someone want to take a look at this? Thanks

buyaa-n:
> Can we check this over for merging again?

Sorry for delay @steveberdy , the code fix of the analyzer is throwing exception in VS when try to run the fixer
```
System.InvalidOperationException : GetCurrentNode returned null with the following node: new[] { 1, 2, 3 } - line 281
   at Roslyn.Utilities.Contract.Fail(String message,Int32 lineNumber)
   at Microsoft.CodeAnalysis.Editing.SyntaxEditor.GetChangedRoot()
   at Microsoft.CodeAnalysis.Editing.DocumentEditor.GetChangedDocument()
   at Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer.ExtractConstArrayAsync(SyntaxNode root,SyntaxNode node,SemanticModel model,DocumentEditor editor,SyntaxGenerator generator,ImmutableDictionary`2 properties,CancellationToken cancellationToken)
   at async Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer.<>c__DisplayClass5_0.<RegisterCodeFixesAsync>b__0(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.<>c__DisplayClass11_0.<GetPreviewAsync>b__0(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Extensions.IExtensionManagerExtensions.PerformFunctionAsync[T](<Unknown Parameters>)
```
I am not sure how the unit tests passing, it is reproducible with a simplest code similar to the tests

If you are not sure what's going on @mavasani @Youssef1313 might give some suggestions.

steveberdy:
I don't know what that is. I haven't had any issues come up like that, even when I ran it on a part of the runtime repository.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,561 @@
...
+namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
+{
+    public class AvoidConstArraysTests
+    {
+        [Fact]
+        public async Task IdentifyConstArrays_ImplicitInitialization()
+        {
+            await VerifyCS.VerifyCodeFixAsync(@"
+using System;
+
```
buyaa-n, line 21: Seems adding a namespace fails the tests, please add a test case with a namespace and probably add another test with some other code/methods

steveberdy: Alrigth

buyaa-n:
> I don't know what that is. I haven't had any issues come up like that, even when I ran it on a part of the runtime repository.

Yeah there is something strange going on, what VS version are you using?  Mine is VS 2022 Preview

steveberdy:
I'm actually using VS code.

steveberdy:
I don't know if I'll have time to finish this analyzer and code fix. If someone wants to pick this up, I'd be open to that.

steveberdy:
@buyaa-n thank you so much for your patience regarding this PR. Could I get a review for this?

buyaa-n:
> @buyaa-n thank you so much for your patience regarding this PR. Could I get a review for this?

Sorry for the delay, I thought you fixed the fixer issue and tested it with runtime repo again, the fixer still throws when try to apply the codefix in VS:
```log
System.InvalidOperationException : GetCurrentNode returned null with the following node: new string[] { "close" } - line 288
   at Roslyn.Utilities.Contract.Fail(String message,Int32 lineNumber)
   at Microsoft.CodeAnalysis.Editing.SyntaxEditor.GetChangedRoot()
   at Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer.ExtractConstArrayAsync(SyntaxNode root,SyntaxNode node,SemanticModel model,DocumentEditor editor,SyntaxGenerator generator,ImmutableDictionary`2 properties,CancellationToken cancellationToken)
   at async Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer.<>c__DisplayClass5_0.<RegisterCodeFixesAsync>b__0(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.<>c__DisplayClass11_0.<GetPreviewAsync>b__0(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Extensions.IExtensionManagerExtensions.PerformFunctionAsync[T](<Unknown Parameters>)
```
I will try to look later this week

steveberdy:
The fixer works for bulk code fixes, but not for individual ones. I'll debug this, but if you have any ideas please let me know.

steveberdy:
I've identified the error and I've spent many hours trying to remedy it, but to no avail.
The analyzer works well in both testing and release environments, but the fixer only applies well in tests.

The problem is that the `GetCurrentNode` returns null for new nodes that are inserted, because they don't already exist on the syntax tree. The error is triggered by `editor.GetChangedDocument` once the `DocumentEditor` applies the replacement from the `editor.ReplaceNode` call in the snippet below.

https://github.com/dotnet/roslyn-analyzers/blob/fec9a2fba1a98ace148e17e4e09e7295a5e4c38e/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs#L94-L102

As I've said, I've spent hours debugging and trying other potential ways to insert the node without getting this error, but I can't get much success. Can any of you provide better insight? Is there something I'm overlooking?

Thanks in advance.

buyaa-n:
> The analyzer works well in both testing and release environments, but the fixer only applies well in tests.

Thank you for your effort and  sorry for delayed response @steveberdy, it sounded like some kind of race condition happening with editor while inserting/replacing the nodes. So, I moved the editor creation:  https://github.com/dotnet/roslyn-analyzers/blob/fec9a2fba1a98ace148e17e4e09e7295a5e4c38e/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs#L40
into the `ExtractConstArrayAsync` and the issue is fixed.

After the fixer issue is solved, further I evaluated the findings and found following issues:
1. False positives in static read only field/property initializations, also I think we should not exclude static constructor's initializations, repro:
  ```cs
  public class A
  {
      public static readonly A Connection = new A(new string[] { "close" }); // Exclude array initializations in readonly fields/properties
      private static readonly HashSet<string> s_acceptedDays = new HashSet<string>(
            new string[] { "monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday" },
            StringComparer.OrdinalIgnoreCase
        );

      public static List<string> Property { get; }
    
      public A(string[] arr) { }

      static A() // Exclude array initializations in static constructors
      {
          Property = GetValues(new string[] { "close" });
      }

      private static List<string> GetValues(string[] arr) => null;
  }
  ```

2. Another issue in fixer, the fixer throws  when try to apply suggestion for conditional expression:
  ```log
System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.
Parameter name: span
   at Microsoft.CodeAnalysis.SyntaxNode.FindNode(TextSpan span,Boolean findInsideTrivia,Boolean getInnermostNodeForTie)
   at async Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer.ExtractConstArrayAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.NetCore.Analyzers.Runtime.AvoidConstArraysFixer.<>c__DisplayClass5_0.<RegisterCodeFixesAsync>b__0(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.PreviewChangesSuggestedAction.CreateAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.GetPreviewChangesFlavorAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.CreateAllFlavorsAsync(<Unknown Parameters>)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.CodeAnalysis.Extensions.IExtensionManagerExtensions.PerformFunctionAsync[T](<Unknown Parameters>)
  ```
  The [code snippet](https://github.com/dotnet/runtime/blob/main/src/libraries/System.DirectoryServices.AccountManagement/src/System/DirectoryServices/AccountManagement/AD/ADStoreCtx_LoadStore.cs#L1080-L1081) looks like this:
  ```cs
      if (!de.Properties.Contains("nTSecurityDescriptor"))
          de.RefreshCache(new string[] { "nTSecurityDescriptor" });
  ```

3. NIT: The fixed value is misaligned in this case:
image
It would be great if it can be fixed. Thank you!

buyaa-n:
> The analyzer works well in both testing and release environments, but the fixer only applies well in tests.

Thanks @steveberdy I pushed a fix for this fixer issue that happening runtime and the issue that causing `System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.`. Also excluded analysis for static constructor and read-only properties. 

Now the analyzer is ready for the final review. @mavasani @Youssef1313 @Evangelink PTAL!

buyaa-n (approved):
LGTM

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        // See https://github.com/dotnet/roslyn/blob/main/docs/analyzers/FixAllProvider.md for more information on Fix All Providers
+        public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            Document document = context.Document;
+            SyntaxNode root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            SyntaxNode node = root.FindNode(context.Span);
+            SemanticModel model = await document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
```
Youssef1313, line 39: If possible, consider moving inside `ExtractConstArrayAsync`. The logic in `RegisterCodeFixesAsync` should be as minimal as possible.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            string title = MicrosoftNetCoreAnalyzersResources.AvoidConstArraysCodeFixTitle;
+            context.RegisterCodeFix(CodeAction.Create(
+                    title,
+                    async ct => await ExtractConstArrayAsync(document, root, node, model, context.Diagnostics.First().Properties, ct).ConfigureAwait(false),
+                    equivalenceKey: title),
```
Youssef1313, line 45:
```suggestion
                    equivalenceKey: nameof(MicrosoftNetCoreAnalyzersResources.AvoidConstArraysCodeFixTitle)),
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            Document document = context.Document;
+            SyntaxNode root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            SyntaxNode node = root.FindNode(context.Span);
+            SemanticModel model = await document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+
+            string title = MicrosoftNetCoreAnalyzersResources.AvoidConstArraysCodeFixTitle;
+            context.RegisterCodeFix(CodeAction.Create(
+                    title,
+                    async ct => await ExtractConstArrayAsync(document, root, node, model, context.Diagnostics.First().Properties, ct).ConfigureAwait(false),
```
Youssef1313, line 44:
```suggestion
                    async ct => await ExtractConstArrayAsync(document, root, node, model, context.Diagnostics[0].Properties, ct).ConfigureAwait(false),
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,153 @@
...
+                    {
+                        return;
+                    }
+
+                    if (context.Operation is IArrayCreationOperation arrayCreationOperation) // For arrays passed as arguments
+                    {
+                        argumentOperation = arrayCreationOperation.GetAncestor<IArgumentOperation>(OperationKind.Argument);
+
+                        // If no argument, return
+                        // If argument is passed as a params array but isn't itself an array, return
```
Youssef1313, line 60: I'm not sure why

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,153 @@
...
+                        {
+                            // This is an invocation that contains an array as an argument
+                            // This will get caught by the first case in another cycle
+                            return;
+                        }
+
+                        argumentOperation = invocationOperation.Arguments.FirstOrDefault();
+                        if (argumentOperation is not null)
+                        {
+                            if (argumentOperation.Children.First() is not IConversionOperation conversionOperation
```
Youssef1313, line 80: Is this the same?

```suggestion
                            if (argumentOperation.Value is not IConversionOperation conversionOperation
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,153 @@
...
+                    IArgumentOperation? argumentOperation;
+
+                    if (context.ContainingSymbol is IMethodSymbol method && method.MethodKind == MethodKind.StaticConstructor)
+                    {
+                        return;
+                    }
+
+                    if (context.Operation is IArrayCreationOperation arrayCreationOperation) // For arrays passed as arguments
+                    {
+                        argumentOperation = arrayCreationOperation.GetAncestor<IArgumentOperation>(OperationKind.Argument);
```
Youssef1313, line 57: Will `GetAncestor` go too far in niche cases similar to this?

```csharp
Console.WriteLine(() => { _ = new[] { "" }; });
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
```diff
@@ -0,0 +1,153 @@
...
+                        return;
+                    }
+
+                    // Must be literal array
+                    if (arrayCreationOperation.Initializer.ElementValues.Any(x => x is not ILiteralOperation))
+                    {
+                        return;
+                    }
+
+                    string? paramName = null;
```
Youssef1313, line 110: Move down near usage

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArraysTests.cs
```diff
@@ -0,0 +1,745 @@
...
+    public class A
+    {
+        public void B()
+        {
+            var x = new string[] { ""a"", ""b"" };
+            var y = x.Select(z => {|CA1861:new[] { ""c"" }|});
+        }
+    }
+}
+", @"
```
Youssef1313, line 277: Consider testing this as well:

```csharp
using System;
using System.Linq;
namespace Z
{
    public class A
    {
        public void B()
        {
            var x = new string[] { "a", "b" };
            var y = x.Select((z1, z2) => new[] { "c" });
        }
    }
}
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
buyaa-n: There was a bug for params array argument https://github.com/dotnet/roslyn-analyzers/pull/5383#issuecomment-1083663879

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.Fixer.cs
Youssef1313: We are currently not consistent. But having a constant (non-localizable) equivalence key was recommended by @sharwell.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/AvoidConstArrays.cs
Youssef1313: Maybe `argumentOperation = arrayCreationOperation.WalkUpConversion().Parent`?
Specific operation kinds can be skipped depending on what we find from the current tests.

buyaa-n: It's not always the 1st Parent after walk up conversion, gonna keep it as is

Issue #5362, opened 2021-08-10
LoggerMessageDefineAnalyzer throws OutOfRangeException when parameter is empty


pranavkm:
### Analyzer

LoggerMessageDefineAnalyzer

### Analyzer source

**SDK**:  6.0.100-rc.1.21410.3

### Describe the bug

Reproed on:https://github.com/dotnet/aspnetcore/blob/3ce1231d5c3900b7629c283d47c124dafb26903e/src/Components/Server/src/Circuits/ServerComponentDeserializer.cs#L229

```C#
private static readonly Action<ILogger, string, Exception> _invalidMarkerType =
                LoggerMessage.Define<string>(
                    LogLevel.Debug,
                    new EventId(4, "InvalidMarkerType"),
                    "Invalid component marker type '{}'.");
```

Stack trace:

```C#
System.IndexOutOfRangeException: Index was outside the bounds of the array.
at Microsoft.NetCore.Analyzers.Runtime.LoggerMessageDefineAnalyzer.AnalyzeFormatArgument(OperationAnalysisContext context, IOperation formatExpression, Int32 paramsCount, Boolean argsIsArray, Boolean usingLoggerExtensionsTypes, IMethodSymbol methodSymbol)
at Microsoft.NetCore.Analyzers.Runtime.LoggerMessageDefineAnalyzer.AnalyzeInvocation(OperationAnalysisContext context, INamedTypeSymbol loggerType, INamedTypeSymbol loggerExtensionsType, INamedTypeSymbol loggerMessageType)
at Microsoft.NetCore.Analyzers.Runtime.LoggerMessageDefineAnalyzer.<>c__DisplayClass27_0.<Initialize>b__1(OperationAnalysisContext context)
at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteOperationAction>b__63_0(ValueTuple`2 data)
at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info)
```

Fixing the parameter resolved the issue.

pranavkm:
FYI @maryamariyan

maryamariyan:
ok looking

maryamariyan:
fixed via #5384

PR #5384, opened 2021-08-17
Fixes exception on LoggerMessageAnalyzer for missing named placeholder

maryamariyan:
Fixes exception on LoggerMessageAnalyzer for missing named placeholder.

Fixes https://github.com/dotnet/roslyn-analyzers/issues/5362

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
@@ -223,7 +223,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 {
                     context.ReportDiagnostic(formatExpression.CreateDiagnostic(CA2253Rule));
                 }
-                else if (char.IsLower(valueName[0]))
+                else if (valueName.Length > 0 && char.IsLower(valueName[0]))
                 {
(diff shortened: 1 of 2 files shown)


maryamariyan:
cc @pranavkm

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
```diff
@@ -223,7 +223,7 @@ private void AnalyzeFormatArgument(OperationAnalysisContext context, IOperation
                 {
                     context.ReportDiagnostic(formatExpression.CreateDiagnostic(CA2253Rule));
                 }
-                else if (char.IsLower(valueName[0]))
+                else if (valueName.Length > 0 && char.IsLower(valueName[0]))
```
pranavkm, line 226: Is the expected to be never null? Would we do `!string.IsNullOrEmpty` instead?

pranavkm (approved):

buyaa-n (approved):
LGTM

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
maryamariyan: sure we could switch over to that

jmarolf (approved):

PR #5395, opened 2021-08-20
Support for detecting preview-ness in dependencies

pgovind:
The containing symbol information for dependencies didn't work the way I expected. If a symbol is not contained within a type, `symbol.ContainingSymbol` returns null. However the `symbol.ContainingAssembly` and `symbol.ContainingModule` are populated correctly, so we can use those instead.

Contributes to https://github.com/dotnet/designs/blob/main/accepted/2021/preview-features/preview-features.md

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
@@ -394,7 +394,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             {
                 if (SymbolIsAnnotatedAsPreview(baseInterfaceMember, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
                 {
-                    string baseInterfaceMemberName = baseInterfaceMember.ContainingType != null ? baseInterfaceMember.ContainingType.Name + "." + baseInterfaceMember.Name : baseInterfaceMember.Name;
+                    string baseInterfaceMemberName = baseInterfaceMember.ContainingSymbol != null ? baseInterfaceMember.ContainingSymbol.Name + "." + baseInterfaceMember.Name : baseInterfaceMember.Name;
                     context.ReportDiagnostic(propertyOrMethodSymbol.CreateDiagnostic(ImplementsPreviewMethodRule, propertyOrMethodSymbol.Name, baseInterfaceMemberName));
                 }
             }
@@ -404,7 +404,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                 ISymbol overridden = propertyOrMethodSymbol.GetOverriddenMember();
                 if (SymbolIsAnnotatedAsPreview(overridden, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
                 {
-                    string overriddenName = overridden.ContainingType != null ? overridden.ContainingType.Name + "." + overridden.Name : overridden.Name;
+                    string overriddenName = overridden.ContainingSymbol != null ? overridden.ContainingSymbol.Name + "." + overridden.Name : overridden.Name;
                     context.ReportDiagnostic(propertyOrMethodSymbol.CreateDiagnostic(OverridesPreviewMethodRule, propertyOrMethodSymbol.Name, overriddenName));
                 }
             }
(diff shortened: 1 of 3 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -626,6 +626,18 @@ private static bool SymbolIsAnnotatedAsPreview(ISymbol symbol, ConcurrentDiction
                     return true;
                 }
 
+                if (symbol.ContainingAssembly != null && symbol.ContainingAssembly.HasAttribute(previewFeatureAttribute))
```
pgovind, line 629: The docs here say that `symbol.ContainingAssembly` can be null if this symbol is part of multiple assemblies. I don't understand that case fully yet. Is this case common? If so, can someone give me a quick example of what this looks like?

jeffhandley (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -626,6 +626,18 @@ private static bool SymbolIsAnnotatedAsPreview(ISymbol symbol, ConcurrentDiction
...
+                    return true;
+                }
+
+                if (symbol.ContainingModule != null && symbol.ContainingModule.HasAttribute(previewFeatureAttribute))
+                {
+                    requiresPreviewFeaturesSymbols.GetOrAdd(symbol, true);
+                    return true;
+                }
+
                 INamedTypeSymbol? parent = symbol.ContainingType;
```
buyaa-n, line 641: Is there any reason to use `symbol.ContainingType` instead of `symbol.ContainingSymbol` here? If it was looping through all containing symbols it should have covered the `module` and `assembly`, therefore no need to check them separately above

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -626,6 +626,18 @@ private static bool SymbolIsAnnotatedAsPreview(ISymbol symbol, ConcurrentDiction
...
                 if (parent is INamespaceSymbol)
                 {
```
buyaa-n, line 643: If it needs to be `symbol.ContainingType` please remove this check, i don't think containing type would go up to namespace

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: > Is there any reason to use symbol.ContainingType instead of symbol.ContainingSymbol here?

Nope. They were giving the same information till today. 

> If it was looping through all containing symbols it should have covered the module and assembly (in addition to symbol.ContainingType )

Yup, you're right. I just tested locally and `ContainingSymbol` seems to containing information even from external dlls. Fixed now. Nice catch!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Changing to ContainingSymbol, so keeping this block

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -626,10 +626,10 @@ private static bool SymbolIsAnnotatedAsPreview(ISymbol symbol, ConcurrentDiction
                     return true;
                 }
 
-                INamedTypeSymbol? parent = symbol.ContainingType;
-                if (parent is INamespaceSymbol)
+                ISymbol? parent = symbol.ContainingSymbol;
```
pgovind, line 629: Just a note for myself. This analyzer uses `ContainingType` everywhere else. Test locally if everything can change to `ContainingSymbol`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Question is not important anymore since I removed the call to Containing Assembly

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Fixed now

buyaa-n (approved):
LGTM

PR #5396, opened 2021-08-20
Clarify instructions for running msbuild /t:Pack

jeffhandley:
It wasn't clear how to run this command or that `dotnet msbuild /t:Pack /v:m` could be used as an alternative command.


fowl2:
is `build.cmd -pack` equivalent, but without the PATH requirements?

PR #5397, opened 2021-08-20
Update ref for the unit tests

pgovind:
I need to be able to access `RequiresPreviewFeaturesAttribute.Message` (https://source.dot.net/#System.Private.CoreLib/RequiresPreviewFeaturesAttribute.cs,37). It was merged in RC1. This change didn't succeed for me locally, so I'm trying it in CI to see if it's a local environment issue.


pgovind:
@sharwell: Any chance you know what's going on here? Or better yet, it might be faster if you could just update it for me please?

pgovind:
sharwell pinged me elsewhere that the unit tests pull from Nuget where RC1/2 are not available yet. Going to close this PR. I'll work around it by defining a local in the unit tests

PR #5402, opened 2021-08-24
Better diagnostic location for types

pgovind:
I've implemented better diagnostic locations for types implementing preview interfaces. Instead of the diagnostic appearing on the type itself, it now appears on the interface in the interface list. I think this is a good starting point to get feedback first, and if I get sign off here, I can merge this PR and work on improving the diagnostic location for methods and other types with the same approach.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
@@ -15,8 +15,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
     /// <summary>
     /// Detect the use of [RequiresPreviewFeatures] in assemblies that have not opted into preview features
     /// </summary>
-    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
-    public class DetectPreviewFeatureAnalyzer : DiagnosticAnalyzer
+    public abstract class DetectPreviewFeatureAnalyzer : DiagnosticAnalyzer
     {
         internal const string RuleId = "CA2252";
         private static readonly LocalizableString s_localizableTitle = new LocalizableResourceString(nameof(MicrosoftNetCoreAnalyzersResources.DetectPreviewFeaturesTitle), MicrosoftNetCoreAnalyzersResources.ResourceManager, typeof(MicrosoftNetCoreAnalyzersResources));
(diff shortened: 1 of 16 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -133,6 +133,11 @@ public sealed class DetectPreviewFeatureAnalyzer : DiagnosticAnalyzer
             FieldOrEventIsPreviewTypeRule,
             StaticAbstractIsPreviewFeatureRule);
 
+        protected virtual SyntaxNode? GetPreviewInterfaceNodeForTypeImplementingPreviewInterface(ISymbol typeSymbol, ISymbol previewInterfaceSymbol)
```
pgovind, line 136: No VB implementation yet. It's not high priority right now

pgovind: Done.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Classes.cs
```diff
@@ -158,5 +158,167 @@ public abstract class AbClass
...
+        [InlineData("struct")]
+        public async Task TestPartialClassOrStructDeclarationEndOfList(string classOrStruct)
+        {
+            var csInput = @$" 
+using System.Runtime.Versioning; using System;
+namespace Preview_Feature_Scratch
+{{
+    public partial {classOrStruct} Zoo";
+
+            csInput += @": NonPreviewInterface, {|#0:IFoo|}
```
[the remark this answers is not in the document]
pgovind: > do you think it looks better?

Yup.  String interpolation uses `{}` and the diagnostic markers also use `{| |}`. Then I needed to change the code blocks using `{` to `{{` to construct `csInput`, so I decided to just split it at an appropriate place. Is this harder to read though? If it is, then I'll change it in the next PR

pgovind:
I addressed @buyaa-n's comments. Just waiting for sign off from @jmarolf or somebody else from @dotnet/roslyn-analysis

jmarolf (approved):

pgovind:
Thanks for the reviews @buyaa-n  and @jmarolf!!!

PR #5406, opened 2021-08-26
Reduce JIT size/time of analyzers

DustinCampbell:
So... here's a small change. :smile:

This pull request represents a a largely mechanical change that reduces the IL byte size and improves JIT time for the our analyzers' static constructors. Essentially, a pattern has propagated in the analyzers over the years that produces larger than necessary static constructors for each of our analyzers. The pattern is to create a static readonly field for each and every instance of `LocalizableResourceString` and pass the fields into the `DiagnosticDescriptor` constructor. In small doses, this pattern isn't a big deal, but over time it has ballooned and produced some very bloated static constructors. This negatively impacts IL size, JIT compile time, and also execution time.

The largest offending `.cctor` is `DiagnosticDescriptorCreationAnalyzer`, which clocks in at more than 3,000 IL bytes. That takes around 15 ms to JIT compile, which might not seem like a lot, but in aggregate it really adds up. After this change, that static constructor is cut in half down to around 1,500 bytes. There are three primary changes that achieve this:

1. `CreateLocalizableResourceString(...)` helpers have been added to handle all construction of `LocalizableResourceString` instance. These helpers avoid multiple `typeof(<SomeResourcesClass>)` expressions, removing tons of IL that looks like so:

```
ldtoken Microsoft.CodeAnalysis.Analyzers.CodeAnalysisDiagnosticsResources
call class [netstandard]System.Type [netstandard]System.Type::GetTypeFromHandle(valuetype [netstandard]System.RuntimeTypeHandle)
```

2. Static readonly fields are no longer declared for `LocalizableString` _unless_ it is referenced more than once. Singleton references are inlined directly into the `DiagnosticDescriptor` constructor calls. This avoids oodles unnecessary `stsfld` and `ldsfld` instructions to store and load the fields.

3. Several unnecessary array allocations have been removed when custom tags are passed into a `DiagnosticDescriptor`. Cached values for each case are already present in the code base - they just need to be used. That gets rid of a bunch of unneeded `newarr` instructions.

Finally, where possible I've changed each analyzer and fixer to use a getter-only auto property for `SupportedDiagnsotics` and `FixableDiagnosticIds` rather than expression-boded properties. With this tweak, we only create an array once for each analyzer, rather than every time these properties are accessed. This change is unrelated to static constructors, it was just an anti-pattern that has proliferated over the years.

Many apologies for the size of the PR. As I mentioned, it is mostly a mechanical change though I get that it'll be a pain to review.


Youssef1313 (commented):
⚠️ We have an [analyzer](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/DiagnosticDescriptorCreationAnalyzer.cs) that does some validations over the titles and it may be relying on this pattern. We should confirm the analyzer is still working with the new pattern before merging.

DustinCampbell:
@Youssef1313: Thanks! I'll take a look and augment the unit tests if needed.

Evangelink:
Sorry for the pretty late ping here but I am wondering if it wouldn't be a goood idea to backport this to the `2.9.x` branch? CC @dotpaul and @mavasani.

src/Microsoft.CodeAnalysis.Analyzers/CSharp/CSharpImmutableObjectMethodAnalyzer.cs
```diff
@@ -11,25 +11,22 @@
 
 namespace Microsoft.CodeAnalysis.CSharp.Analyzers
 {
+    using static CodeAnalysisDiagnosticsResources;
+
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     public class CSharpImmutableObjectMethodAnalyzer : DiagnosticAnalyzer
     {
-        // Each analyzer needs a public id to identify each DiagnosticDescriptor and subsequently fix diagnostics in CodeFixProvider.cs
```
Youssef1313, line 17: nit: This comment was unintentionally removed.

steveberdy:
Would it be better to have a global static using directive in just one place?
```csharp
global using static Microsoft.NetCore.Analyzers.MicrosoftNetCoreAnalyzersResources;
```
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/globalusingdirective#global-using-static-directives

sharwell:
> Would it be better to have a global ... using directive ... ?

No.

PR #5422, opened 2021-08-31
Allow PublicAPI analyzers to merge from multiple sources

### LINKED ISSUE #3187: API checker needs to support multi-targeted projects
Roslyn is [intending to change](https://github.com/dotnet/roslyn/issues/40766) many of the core projects, including those which have public APIs, to be multi-targeted. That potentially leads to an issue with the API checker as it will be asked to validate the public API surface area twice: once for netstandard2.0 and netcoreapp3.1.

That will lead to issues if the API differs between the different target frameworks. At the moment there is no intent to deviate here hence the double validation is likely to just be more work, not an actual blocker. The ability to deviate may become important in the future as we may want to expose new APIs / types across the netcoreapp3.1 build chain. At that point we will need to have a way to split up the API across different files based on TF.

twsouthwick:
This changes updates the public change analyzer to merge the data about what is a public API from any additional file that follows the pattern `PublicAPI.[Un]shipped.*txt`. This allows, for example, for additional TFMs to add specific APIs they expose in an additive way. By default, APIs are still added to just the `PublicAPI.[Un]shipped.txt` file, but will be updated in the alternate files if needed.

Fixes #3187


twsouthwick:
I've linked a PR for where I've built the PublicAPI docs around this change. You can see that the TFM specific ones are super small, and it also leaves it pretty obvious what the differences are.

twsouthwick:
@jmarolf I see you self-assigned this. Can you take a look? I'm readying a release of a library and would like to get it in to track changes for the next version if possible.

src/PublicApiAnalyzers/UnitTests/DeclarePublicApiAnalyzerTests.cs
```diff
@@ -1798,6 +1798,39 @@ public class C
...
+{
+    public string? {|RS0037:{|RS0016:NewField|}|}; // Newly added field, not in current public API.
+}
+";
+
+            var shippedText = @"";
+            var unshippedText1 = @"C
+C.C() -> void";
+            var unshippedText2 = @"";
+            var fixedUnshippedText1 = @"C
```
jmarolf, line 1816: I am not certain that we want to fix to just pick the first additional file to insert the changes into. I assume we would want to provider a fix for each file so the user could choose where to insert the new api definition.

twsouthwick:
@jmarolf I believe I've responded to all your feedback. Let me know if there's anything else I should do. Thanks!

twsouthwick:
@jmarolf when is this anticipated to be released? Is there a CI feed this will be pushed to before made publicly available? On the README the link sends me to a feed that hasn't been updated for a while.

jmarolf:
I need to update the readme. you can comsume these changes by creating a NuGet.config file that looks like this

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="dotnet7" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

nohwnd:
Thanks for the change @twsouthwick ! I used it for vstest, if anyone is looking for more examples. We are in a bit specific situation where we support multiple tfms, including netstandard1.0 which has significant differences from the other tfms, so I have 2 big base files. 1 that is common to all tfms, and 1 that is common to all tfms apart from netstandard1.0, and then tfm specific files. I then use msbuild condition to combine them, and it works perfectly, with very little duplication:

common shipped + common net shipped + net451 shipped = shipped api for net451 tfm
common shipped + netstandard1.0 shipped = api for netstandard tfm

I am also including code in the PR comments that helped me split my original files into the common files:

https://github.com/microsoft/vstest/pull/3165#pullrequestreview-803867024

PR #5434, opened 2021-09-01
adding required variable groups

jmarolf:


jmarolf:
Internal test build [here](https://dev.azure.com/dnceng/internal/_build/results?buildId=1335154)

PR #5454, opened 2021-09-07
Do not pass mutable value types by value

NewellClark:
Fix dotnet/runtime#33772.

As of right now, the analyzer reports all parameters that are typed as mutable value types and are not marked as `ref` or `out`. We also report all methods who's return type is a mutable value type that do not return by reference, unless the return type is a struct-enumerator type and the method's name is "GetEnumerator". 
We report all properties of a mutable value type that do not return by reference.

I ran the analyzer against dotnet/runtime and found multiple false-positives in [JsonElement.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonElement.cs) (`EnumerateArray()` and `EnumerateObject()`) and [ThrowHelper.Serialization.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/ThrowHelper.Serialization.cs).

There were false positive violations for both the return-values rule and the pass-by-value rule. All the return-value rule violations were factory methods. I think we should consider not reporting violations for method return values, as that would raise a false positive for any factory method, which seems like a fairly common scenario. 
In the meantime, I've added an exception in the analyzer for methods named "GetEnumerator" that return a struct-enumerator type, as otherwise we'd get a huge number of false positives throughout dotnet/runtime.

All of the pass-by-value violations were `in` parameters, and I was able to determine that the methods do not call any mutating members on the arguments in question. We may want to consider not flagging `in` parameters. Unlike passing by-value, which is the default, it seems likely that any parameter marked as `in` was done so with the intention that the method will not be modifying it, and therefore is likely safe.


src/NetAnalyzers/CSharp/GlobalSuppressions.cs
```diff
@@ -0,0 +1,5 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Style", "IDE0130:Namespace does not match folder structure", Justification = "<Pending>", Scope = "namespace", Target = "~N:Microsoft.NetCore.CSharp.Analyzers.Runtime")]
```
sharwell, line 5: ❗ This repository doesn't use global suppressions. The rule can be configured in the appropriate **.globalconfig** file.

sharwell:
I strongly believe that [RS0042 (Do not copy value)](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs) is superior to this proposed rule in all aspects.

sharwell:
This is a really complicated analysis scenario. I ran into a large number of issues working in this space in the past and I'm not sure a pull request review is the right place to try and cover them all.

src/NetAnalyzers/CSharp/GlobalSuppressions.cs
NewellClark: > This is a really complicated analysis scenario. I ran into a large number of issues working in this space in the past and I'm not sure a pull request review is the right place to try and cover them all.

Is there a slack or discord channel where we could discuss this further?

NewellClark:
> This is a really complicated analysis scenario. I ran into a large number of issues working in this space in the past and I'm not sure a pull request review is the right place to try and cover them all.

Is there a slack or discord channel where we could discuss this further?

src/NetAnalyzers/CSharp/GlobalSuppressions.cs
NewellClark: Fixed.

NewellClark:
> I strongly believe that [RS0042 (Do not copy value)](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Roslyn.Diagnostics.Analyzers/Core/AbstractDoNotCopyValue.cs) is superior to this proposed rule in all aspects.

@sharwell It looks like RS0042 does much more than what was specified [here](https://github.com/dotnet/runtime/issues/33772#issuecomment-854061368).

sharwell:
Yes, I'm still discussing this with the runtime team. It looks like some incorrect assumptions were made regarding `Utf8JsonReader`, and it's much easier to misuse than expected.

src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md
```diff
@@ -1620,6 +1620,30 @@ Number of parameters supplied in the logging message template do not match the n
 |CodeFix|False|
 ---
 
+## [CA2019](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2019): Do not pass mutable value types by value
```
tannergooding, line 1623: For both of these, what is the plan for handling and/or annotating value types which explicitly should be passed and returned by value?

For example, this is going to have tons of false positives in interop code. It's likewise not going to be taking SIMD (`Vector2`, `Vecto3`, or `Vector4`) or ABI (Unix can pass a struct in multiple registers sometimes) into account where passing by `ref` or `in` is generally inappropriate and may hurt perf or cause other regressions. This may likewise flag things like `Guid` incorrectly on older TFMs given that it wasn't updated to be `readonly` until newer releases (even though it was "functionally readonly" already).

NewellClark: As per [this post](https://github.com/dotnet/runtime/issues/33772#issuecomment-913721146), this analyzer only flags value types that one of the following:

- On a built-in list of problematic types (the list currently contains `SpinLock` and `Utf8JsonReader`)
- Added as problematic mutable value types via an .editorconfig option,
- Types that it determines to be struct-enumerator types (must be a nested value type who's name ends with "Enumerator").

PR #5468, opened 2021-09-12
Features/58591 warn on subsequent orderby calls

chucker:
This implements https://github.com/dotnet/runtime/issues/58591.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.Usage
+{
+    using static MicrosoftNetCoreAnalyzersResources;
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public class DoNotCallOrderByMultipleTimes : DiagnosticAnalyzer
```
Youssef1313, line 15: nit: sealed

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
```diff
@@ -0,0 +1,85 @@
...
+                    return;
+
+                if (invocationOperation.TargetMethod.Name is not "OrderBy" and not "OrderByDescending")
+                    return;
+
+                if (PreviousOrderByInvocation is null)
+                {
+                    PreviousOrderByInvocation = invocationOperation;
+                }
+                else if (PreviousOrderByInvocation != invocationOperation)
```
Youssef1313, line 78: Does the compiler make duplicate callbacks?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    context.ReportDiagnostic(PreviousOrderByInvocation.CreateDiagnostic(Rule));
+                }
```
Youssef1313, line 81: I'm feeling this will incorrectly trigger a diagnostic for:


```csharp
var x = MyCollection.OrderBy(c => c.X);
var y = MyOtherCollection.OrderBy(c => c.Y);
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public class DoNotCallOrderByMultipleTimes : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA2259";
+
+        internal static readonly DiagnosticDescriptor Rule = DiagnosticDescriptorHelper.Create(
+            RuleId,
+            CreateLocalizableResourceString(nameof(DoNotCallOrderByMultipleTimesTitle)),
+            CreateLocalizableResourceString(nameof(DoNotCallOrderByMultipleTimesDescription)),
```
Youssef1313, line 22: This is misleading. The parameter that follows title is a message, not description.

https://github.com/dotnet/roslyn-analyzers/blob/1595b90e6f3c387d684c04b3a570f87bce4e7580/src/Utilities/Compiler/DiagnosticDescriptorHelper.cs#L14-L17

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.Fixer.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        protected abstract Document ReplaceOrderByWithThenBy(Document document,
+                                                             SyntaxNode root,
+                                                             SyntaxNode node);
+
+        private class CodeAction : DocumentChangeAction
+        {
+            public CodeAction(Func<CancellationToken, Task<Document>> action)
+            : base(DoNotCallOrderByMultipleTimesTitle, action, DoNotCallOrderByMultipleTimes.RuleId)
```
Youssef1313, line 48: - Codefix title should be different than analyzer title.
- Use `nameof(DoNotCallOrderByMultipleTimesTitle)` as equivalence key.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
chucker: I'm not sure. In my testing, it looked like it may have.

(I'm not sure this approach is viable in any case; see other concerns.)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
chucker: Probably. My other concern is a false negative:

```csharp
var x = MyCollection.OrderBy(c => c.X);

var y = x.OrderBy(c => c.Y); // this should report the diagnostic
```

I tried (unsuccessfully, so far) to instead see if the type of `x` here is `IOrderedEnumerable`.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Usage/DoNotCallOrderByMultipleTimes.cs
chucker: I checked various analyzers and they seem to use a message that's identical to the title, so I'm not doing that, too. However, the `sarif` file apparently uses the title for the `shortDescription`, and the message for the `fullDescription`. That's a bit weird.

PR #5469, opened 2021-09-13
Exclude test projects from source-build

dseefeld:

### DIFF
--- a/src/PerformanceTests/Utilities/ReferenceAssemblies/ReferenceAssemblies.csproj
+++ b/src/PerformanceTests/Utilities/ReferenceAssemblies/ReferenceAssemblies.csproj
@@ -4,6 +4,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <IsPackable>false</IsPackable>
     <NonShipping>true</NonShipping>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/PerformanceTests/Utilities/ReferenceAssemblies/ReferenceAssemblies.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <IsPackable>false</IsPackable>
     <NonShipping>true</NonShipping>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
Evangelink, line 7: Shall it be defined as described by @sharwell in https://github.com/dotnet/arcade/issues/7132#issue-837193193?

dseefeld: I don't think so.  @sharwell is suggesting a change to arcade to make this easier for repos.

Youssef1313 (commented):
Why not in `Directory.Build.props` with a `'$(NonShipping) == 'true'` condition?

dseefeld:
> Why not in `Directory.Build.props` with a `'$(NonShipping)' == 'true'` condition?

As noted in the arcade comment above:

  > Some of the tools we use in the repository are needed for the build to complete but don't actually ship anywhere.

sharwell (approved):

Issue #5450, opened 2021-09-06
DetectPreviewFeatureAnalyzer crashes with System.ArgumentNullException


marek-safar:
### Analyzer

**Diagnostic ID**: CA2252

### Describe the bug

```
Analyzer 'Microsoft.NetCore.CSharp.Analyzers.Runtime.CSharpDetectPreviewFeatureAnalyzer' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null. (Parameter 'key')'. [/Users/xyz/git/linker/src/tlens/tlens.csproj]
                     Exception occurred with following context: (TaskId:107)
                     Compilation: tlens (TaskId:107)
                     IOperation: PropertyReference (TaskId:107)
                     SyntaxTree: /Users/xyz/git/linker/src/tlens/TLens/AssemlyReferenceResolver.cs (TaskId:107)
                     SyntaxNode: ReaderParameters [IdentifierNameSyntax]@[587..603) (19,3)-(19,19) (TaskId:107)
                     System.ArgumentNullException: Value cannot be null. (Parameter 'key') (TaskId:107)
                     at System.Collections.Concurrent.ConcurrentDictionary`2.ThrowKeyNullException() (TaskId:107)
                     at Microsoft.NetCore.Analyzers.Runtime.DetectPreviewFeatureAnalyzer.SymbolIsAnnotatedAsPreview(ISymbol symbol, ConcurrentDictionary`2 requiresPreviewFeaturesSymbols, INamedTypeSymbol previewFeatureAttribute) (TaskId:107)
                     at Microsoft.NetCore.Analyzers.Runtime.DetectPreviewFeatureAnalyzer.SymbolIsAnnotatedOrUsesPreviewTypes(ISymbol symbol, ConcurrentDictionary`2 requiresPreviewFeaturesSymbols, INamedTypeSymbol previewFeatureAttributeSymbol, ISymbol& referencedPreviewSymbol) (TaskId:107)
                     at Microsoft.NetCore.Analyzers.Runtime.DetectPreviewFeatureAnalyzer.OperationUsesPreviewFeatures(OperationAnalysisContext context, ConcurrentDictionary`2 requiresPreviewFeaturesSymbols, INamedTypeSymbol previewFeatureAttributeSymbol, ISymbol& referencedPreviewSymbol) (TaskId:107)
[...]

PR #5470, opened 2021-09-13
Fix crash when querying auto properties

pgovind:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5450

Currently, the analyzer assumes that a property.UsageInfo == Write implies that a `Set` method exists. This leads to a crash when dealing with auto properties where they are not backed by explicit fields. This PR fixes that.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
@@ -677,17 +677,17 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                     {
                         symbol = propertySymbol.SetMethod;
                     }
-
-                    if (symbol == null)
-                    {
-                        referencedPreviewSymbol = null;
-                        return false;
-                    }
                 }
 
                 if (operation is IEventAssignmentOperation eventAssignment && symbol is IEventSymbol eventSymbol)
                 {
                     symbol = eventAssignment.Adds ? eventSymbol.AddMethod : eventSymbol.RemoveMethod;
+                }
+
+                if (symbol == null)
+                {
+                    referencedPreviewSymbol = null;
+                    return false;
                 }
 
                 if (symbol is IMethodSymbol methodSymbol && methodSymbol.IsConstructor())
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
@@ -677,6 +677,12 @@ namespace Microsoft.NetCore.Analyzers.Runtime
                     {
                         symbol = propertySymbol.SetMethod;
                     }
+
+                    if (symbol == null)
+                    {
+                        referencedPreviewSymbol = null;
+                        return false;
+                    }
                 }
 
                 if (operation is IEventAssignmentOperation eventAssignment && symbol is IEventSymbol eventSymbol)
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -684,6 +684,12 @@ private bool OperationUsesPreviewFeatures(OperationAnalysisContext context, Conc
                     symbol = eventAssignment.Adds ? eventSymbol.AddMethod : eventSymbol.RemoveMethod;
                 }
 
+                if (symbol == null)
```
pgovind, line 687: On line 678 above, propertySymbol.SetMethod could be null. I'm just being defensive and adding the null check here.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Properties.cs
```diff
@@ -10,6 +10,30 @@ namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
 {
     public partial class DetectPreviewFeatureUnitTests
     {
+        [Fact]
+        public async Task TestPropertyGetterSetInConstructor()
+        {
+            var csInput = @"
+using System.Runtime.Versioning; using System;
```
Evangelink, line 17:
```suggestion
using System;
using System.Runtime.Versioning;
```

Evangelink: Not sure if you really need the `System.Runtime.Versioning` using.

Evangelink (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: I would not expect `SetMethod` to be null when `usageInfo.IsWrittenTo()` but `usageInfo == ValueUsageInfo.ReadWrite` might be, do not know for sure. 

I think it is better to set the symbol with `SetMethod` only if the getter was not available, i.e for lines 671-679:
```cs
        if (usageInfo.IsReadFrom())
        {
            symbol = propertySymbol.GetMethod;
        }
        else if (usageInfo.IsWrittenTo() || usageInfo == ValueUsageInfo.ReadWrite)
        {
            symbol = propertySymbol.SetMethod;
        }
```
Because property is expected to have a getter or setter (or both of course) no need to check for null after this

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Properties.cs
```diff
@@ -10,6 +10,30 @@ namespace Microsoft.NetCore.Analyzers.Runtime.UnitTests
...
+    {
+        public int Value { get; }
+        public Program()
+        {
+            Value = 1;
+        }
+
+        static void Main(string[] args)
+        {
+        }
```
buyaa-n, line 30: Doesn't seem to need it.
```suggestion
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: And I expect when there is a getter we should check/analyzer the getter

buyaa-n: > And I expect when there is a getter we should check/analyzer the getter

Ah i misunderstood, getter was not overwritten, please ignore the above comments

I wonder if we should continue the analysis with the original `symbol` (set the symbol with only if setter or getter is not null), which might be not needed as the property cannot be set from outside

buyaa-n (approved):
Left an NIT question, overall LGTM

PR #5471, opened 2021-09-13
Make CA2252 an error

pgovind:
This will round out the initial development of CA2252.


jmarolf (approved):

buyaa-n (approved):

jeffhandley (approved):

mavasani:
So, I see that the rule was already added to shipped file with Info severity here: https://github.com/dotnet/roslyn-analyzers/blob/release/6.0.1xx/src/NetAnalyzers/Core/AnalyzerReleases.Shipped.md

This change should have caused a build break as now the rule implementation has a different severity than the AnalyzerReleases file - the fact that did not happen indicates the analyzer ensuring the AnalyzerReleases.*.md files are in sync with the analyzer implementation is busted in our CI. I'll follow-up on that separately, that shouldn't block this PR though.

pgovind:
> This change should have caused a build break as now the rule implementation has a different severity than the AnalyzerReleases file - the fact that did not happen indicates the analyzer ensuring the AnalyzerReleases.*.md files are in sync with the analyzer implementation is busted in our CI. I'll follow-up on that separately, that shouldn't block this PR though.

Probably not related, just a data point: I did see a CI error that needed me to run msbuild /t:pack. It only changed the *.md/sarif files though

Issue #5415, opened 2021-08-27
CA2254 Should not apply when log message template references constant values


DamianEdwards:
The following code block receives info messages from CA2254 despite the interpolated values (the `nameof()` expression) being constant:
``` c#
private static IParameterBinder<TModel> LookupBinder(IServiceProvider services, ILogger logger)
{
    var binder = services.GetService<IParameterBinder<TModel>>();

    if (binder is not null)
    {
        logger.LogDebug($"{nameof(IParameterBinder<object>)}<{{0}}> resolved from DI container.", typeof(TModel).Name);

        return binder;
    }

    logger.LogDebug($"{nameof(IParameterBinder<object>)}<{{0}}> could not be resovled from DI container, using default JSON binder.", typeof(TModel).Name);

    return DefaultJsonParameterBinder.Instance;
}
```
![image](https://user-images.githubusercontent.com/249088/131198173-292d36cf-36a8-454c-83f4-433da9d86e78.png)

maryamariyan:
looking

Youssef1313:
I'm curious if the compiler version (not language version) you're using supports constant interpolated strings?

DamianEdwards:
I'm using whatever compiler version is in latest internal VS2022 Preview 4.0 builds, so pretty much as new as it gets.

Youssef1313:
Thanks for confirming @DamianEdwards 

@maryamariyan Small observations given a very quick look here:


https://github.com/dotnet/roslyn-analyzers/blob/7123df44e4503f5426290030d8187a3edb92bed7/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs#L250-L251

1. Why we're looking the Value if HasValue is false? I don't really get this part.
2. We should look at constant value for interpolation strings as well, not only literals. But this will work only on a compiler supporting constant interpolated strings. For older compiler versions, this will give false positives.

jaredpar:
@DamianEdwards Can you give us two bits of info:

1. What does `#error version` print when you add it to the code?
2. Are you using `net6.0`? 

Presumably yes to (2) but want to check

DamianEdwards:
@jaredpar 
`Error	CS8304	Compiler version: '4.0.0-4.21426.10 (6898d692)'. Language version: preview.`
Yes I'm using `net6.0`

jaredpar:
[...]

PR #5480, opened 2021-09-14
Skipping CA2254 diagnostic when message template references constant values

maryamariyan:
Skipping CA2254 diagnostic when message template references constant values

Fixes https://github.com/dotnet/roslyn-analyzers/issues/5415

/cc: @sharwell @jmarolf @333fred

TODO:

- [x] Add test

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
@@ -229,30 +229,9 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             switch (argumentExpression)
             {
-                case ILiteralOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
-                    return constantValue;
-                case IInterpolatedStringOperation interpolated:
-                    var text = "";
-                    foreach (var interpolatedStringContent in interpolated.Parts)
-                    {
-                        if (interpolatedStringContent is IInterpolatedStringTextOperation textSyntax)
-                        {
-                            text += textSyntax.Text;
-                        }
-                        else if (interpolatedStringContent is IInterpolationOperation { ConstantValue: { HasValue: true, Value: string constantValue } })
-                        {
-                            text += constantValue; // works for e.g. nameof
-                        }
-                        else
-                        {
-                            return null;
-                        }
-                    }
-                    return text;
-                case INameOfOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
+                case IOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
                     return constantValue;
                 case INameOfOperation:
-                    // return placeholder from here because actual value is not required for analysis and is hard to get
                     return "NAMEOF";
                 case IParenthesizedOperation parenthesized:
                     return TryGetFormatText(parenthesized.Operand);
@@ -267,11 +246,6 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
                     return null;
                 default:
-                    var constant = argumentExpression.ConstantValue;
-                    if (constant.HasValue && constant.Value is string constantString)
-                    {
-                        return constantString;
-                    }
                     return null;
             }
         }
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
@@ -231,10 +231,6 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             {
                 case IOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
                     return constantValue;
-                case INameOfOperation:
-                    return "NAMEOF";
-                case IParenthesizedOperation parenthesized:
-                    return TryGetFormatText(parenthesized.Operand);
                 case IBinaryOperation { OperatorKind: BinaryOperatorKind.Add } binary:
                     var leftText = TryGetFormatText(binary.LeftOperand);
                     var rightText = TryGetFormatText(binary.RightOperand);
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
@@ -229,7 +229,7 @@ namespace Microsoft.NetCore.Analyzers.Runtime
 
             switch (argumentExpression)
             {
-                case ILiteralOperation { ConstantValue: { HasValue: false, Value: string constantValue } }:
+                case ILiteralOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
                     return constantValue;
                 case IInterpolatedStringOperation interpolated:
                     var text = "";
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
```diff
@@ -257,6 +257,10 @@ private void AnalyzeFormatArgument(OperationAnalysisContext context, IOperation
                         {
                             text += textSyntax.Text;
                         }
+                        else if (interpolatedStringContent is IInterpolationOperation { ConstantValue: { HasValue: true, Value: string constantValue } })
+                        {
+                            text += constantValue; // works for e.g. nameof
```
maryamariyan, line 262: expected to work for nameof, does not seem to be supported yet

ericstj: Can we tag the folks delivering that change and link the issue so that it's clear we're blocked on it?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
```diff
@@ -247,7 +247,7 @@ private void AnalyzeFormatArgument(OperationAnalysisContext context, IOperation
 
             switch (argumentExpression)
             {
-                case ILiteralOperation { ConstantValue: { HasValue: false, Value: string constantValue } }:
+                case ILiteralOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
                     return constantValue;
                 case IInterpolatedStringOperation interpolated:
```
jmarolf, line 252: I believe we should just check if `IInterpolatedStringOperation.ConstantValue` is null. We should not need to do any of this looping through the parts

jmarolf: essentially, I don't understand why it can't just be this.

```csharp
case IInterpolatedStringOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
    return constantValue;
```

jmarolf:
@maryamariyan can you please add a test case for https://github.com/dotnet/roslyn-analyzers/issues/5415?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
333fred: (Also, as a note, I don't believe the `case INameOfOperation` when there's no constant value is possible to hit?)

333fred: Oh, and another note: you almost certainly don't want to be checking for `IParenthesizedOperation` unless you have found a specific test case in VB where the parentheses matter.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineTests.cs
```diff
@@ -27,6 +27,12 @@ public async Task CA2253IsProducedForNumericFormatArgument(string format)
             await TriggerCodeAsync(format);
         }
 
+        [Fact]
+        public async Task CA2254ShouldNotApplyForConstantValues()
+        {
+            await TriggerCodeAsync("{|CA1848:logger.LogDebug({|CA2254:$\"{nameof(System.Collections.Generic.IList<object>)}<{{0}}> could not be resovled from DI container, using default JSON binder.\"|}, typeof(string).Assembly)|};");
```
maryamariyan, line 33:
```suggestion
            await TriggerCodeAsync("{|CA1848:logger.LogDebug(:$\"{nameof(System.Collections.Generic.IList<object>)}<{{0}}> could not be resovled from DI container, using default JSON binder.\", typeof(string).Assembly)|};");
```

goal is to make test pass after applying this suggestion. Still with this fix it adds `CA2254` for this use case

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
maryamariyan: > essentially, I don't understand why it can't just be this.

replaced as such and seems like all tests pass this way. thanks!

maryamariyan: @333fred ok thanks, removed the two switch cases

maryamariyan:
PR ready for review again

jmarolf (approved):

jmarolf (approved):

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
```diff
@@ -247,29 +247,8 @@ private void AnalyzeFormatArgument(OperationAnalysisContext context, IOperation
...
-                        }
-                    }
-                    return text;
-                case INameOfOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
-                    return constantValue;
-                case INameOfOperation:
-                    // return placeholder from here because actual value is not required for analysis and is hard to get
-                    return "NAMEOF";
-                case IParenthesizedOperation parenthesized:
-                    return TryGetFormatText(parenthesized.Operand);
```
Youssef1313, line 272: Why the parenthesized operation was removed? (Note: this IOperation applies only for VB)

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
```diff
@@ -247,29 +247,8 @@ private void AnalyzeFormatArgument(OperationAnalysisContext context, IOperation
 
             switch (argumentExpression)
             {
-                case ILiteralOperation { ConstantValue: { HasValue: false, Value: string constantValue } }:
+                case IOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
```
Youssef1313, line 250: This will handle constant interpolated strings only for compiler versions that support it. I'm not sure whether this should support old versions or not.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/LoggerMessageDefineAnalyzer.cs
```diff
@@ -247,29 +247,8 @@ private void AnalyzeFormatArgument(OperationAnalysisContext context, IOperation
...
-                case INameOfOperation { ConstantValue: { HasValue: true, Value: string constantValue } }:
-                    return constantValue;
-                case INameOfOperation:
-                    // return placeholder from here because actual value is not required for analysis and is hard to get
-                    return "NAMEOF";
-                case IParenthesizedOperation parenthesized:
-                    return TryGetFormatText(parenthesized.Operand);
                 case IBinaryOperation { OperatorKind: BinaryOperatorKind.Add } binary:
                     var leftText = TryGetFormatText(binary.LeftOperand);
                     var rightText = TryGetFormatText(binary.RightOperand);
```
Youssef1313, line 254: What about user-defined operators (which may be poorly written and even produce "random" string regardless the operands)?

I'm also wondering if this can be hit? Won't the compiler correctly produce a constant value which is handled in the first case?

jmarolf:
@maryamariyan can you run `msbuild /t:pack` to update the markdown files?

Issue #5205, opened 2021-07-04
CA1416 does not recognise built-in guards for macOS-specific types


yaakov-h:
### Analyzer

**Diagnostic ID**: [CA1416](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1416): `Validate platform compatibility`

### Analyzer source

**SDK**: [Built-in CA analyzers in .NET 5 SDK or later](https://docs.microsoft.com/dotnet/fundamentals/productivity/code-analysis)

**Version**: SDK 6.0.100-preview.5.21302.13 (included in VS2022 Preview 1)

### Describe the bug

CA1416 triggers when I try to use a type annotated as supporting `"osx"` or `"macos"`, when I guard the calls with `RuntimeInformation.IsOSPlatform(OSPlatforms.OSX)`.

### Steps To Reproduce

Compile the following code example:

```csharp
using System.Runtime.InteropServices;
using System.Runtime.Versioning;

namespace DemoSupportedPlatformLib
{
    public abstract class PlatformSpecificFunction
    {
        public static PlatformSpecificFunction GetProvider()
        {
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
            {
                return new WindowsSpecificFunction();
            }
            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
            {
                return new OSXSpecificFunction();
            }
            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
            {
                return new LinuxSpecificFunction();
            }
            else
            {
                return new DefaultFunction();
            }
        }
    }

    [SupportedOSPlatform("windows")]
    class WindowsSpecificFunction : PlatformSpecificFunction
    {
    }

    [SupportedOSPlatform("macos")]
    class OSXSpecificFunction : PlatformSpecificFunction
    {
    }

    [SupportedOSPlatform("linux")]
    class LinuxSpecificFunction : PlatformSpecificFunction
    {
    }

    class DefaultFunction : PlatformSpecificFunction
    {
    }
}
```

### Expected behavior

No build warnings or errors.

### Actual behavior

> warning CA1416: This call site is reachable on: 'OSX'. 'OSXSpecificFunction' is only supported on: 'macOS/OSX'.

Youssef1313:
cc @buyaa-n

buyaa-n:
[...]

PR #5495, opened 2021-09-15
Fix RuntimeInformaiton.IsOSPlatform(OSPlatform.OSX) guard

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5205

Because `OSX` is the alias of `MacOS` we are replacing the `OSX` with `MacOS` while parsing the OS platform attributes

https://github.com/dotnet/roslyn-analyzers/blob/c5c6a7df6452524557baaa28c47da1d4828c1687/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs#L1815-L1816

There is no guard for OSX in `OperatingSystem` type and the guard `OperatingSystem.IsMacOS()` is working just fine with the above change, though we forgot about `RuntimeInformaiton.IsOSPlatform(OSPlatform.OSX)` can be used as a guard for `macOS/OSX`. This PR is fixing that gap.

Not sure if 6.0 is already closed, but it is good to have in 6.0 if possible @jmarolf @jeffhandley

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.Value.cs
@@ -121,7 +121,14 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                 if ((argumentValue is IPropertyReferenceOperation propertyReference) &&
                     propertyReference.Property.ContainingType.Equals(osPlatformType))
                 {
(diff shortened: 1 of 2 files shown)


mavasani (approved):

jeffhandley (approved):
LGTM. I defer to @jmarolf for whether or not he's already snapped into RC2. If not, then I support including this in the snap. If we've already snapped into RC2, then we would need to request tactics approval for either RC2 or GA.

jmarolf:
This is the last change for RC2

buyaa-n:
> This is the last change for RC2

Thanks @jmarolf

PR #5502, opened 2021-09-17
Support Url and Message parameters in RequiresPreviewFeaturesAttribute

pgovind:
https://github.com/dotnet/runtime/pull/56938 added the Message and URL properties to RequiresPreviewFeaturesAttribute. This PR updates the RequiresPreviewFeaturesAnalyzer to support the new properties in the message diagnostic.


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1720,7 +1720,7 @@
     <value>An assembly has to opt into preview features before using them.</value>
```
pgovind, line 1720: This PR is not ready for review yet. I need to clean it up a little bit,but I wanted to see how CI performs with the CultureInfo changes first.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -20,13 +22,24 @@ namespace Microsoft.NetCore.Analyzers.Runtime
     public abstract class DetectPreviewFeatureAnalyzer : DiagnosticAnalyzer
     {
         internal const string RuleId = "CA2252";
+        internal const string DefaultURL = "https://aka.ms/dotnet-warnings/preview-features";
         private static readonly LocalizableString s_localizableTitle = CreateLocalizableResourceString(nameof(DetectPreviewFeaturesTitle));
         private static readonly LocalizableString s_localizableDescription = CreateLocalizableResourceString(nameof(DetectPreviewFeaturesDescription));
+        internal static readonly LocalizableString s_detectPreviewFeaturesMessage = CreateLocalizableResourceString(nameof(DetectPreviewFeaturesMessage));
```
pgovind, line 28: There is an explicit conversion defined that does LocalizableString -> string that I use everywhere. I _assumed_ it'll work well with globalization, but if that's not the case, please let me know!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -225,33 +240,39 @@ public override void Initialize(AnalysisContext context)
...
+        private void ProcessFieldOrEventSymbolAttributes(SymbolAnalysisContext context,
+                                                         ISymbol symbol,
+                                                         ISymbol symbolType,
+                                                         ConcurrentDictionary<ISymbol, bool> requiresPreviewFeaturesSymbols,
+                                                         ConcurrentDictionary<ISymbol, ValueTuple<string?, string?>> previewSymbolsToMessageAndUrl,
+                                                         INamedTypeSymbol previewFeatureAttributeSymbol)
         {
             if (SymbolIsAnnotatedAsPreview(symbolType, requiresPreviewFeaturesSymbols, previewFeatureAttributeSymbol))
             {
+                string message = string.Format(CultureInfo.CurrentCulture, (string)s_fieldOrEventIsPreviewTypeMessage, symbol.Name, symbolType.Name);
```
pgovind, line 267: 1 example of where I'm explicitly converting LocalizableString -> string.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -767,6 +829,141 @@ private static ISymbol SymbolFromArrayCreationOperation(IArrayCreationOperation
...
+                    message = messageValue;
+                }
+            }
+
+            ImmutableArray<System.Collections.Generic.KeyValuePair<string, TypedConstant>> namedArguments = attribute.NamedArguments;
+            if (namedArguments.Length != 0)
+            {
+                foreach (System.Collections.Generic.KeyValuePair<string, TypedConstant> namedArgument in namedArguments)
+                {
+                    if (namedArgument.Key == "Message")
```
pgovind, line 852: I don't think this needs symbol level checks, but if there's a valid reason to do so, please let me know

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Classes.cs
```diff
@@ -50,7 +51,7 @@ interface IFoo
 }";
 
             var test = TestCS(csInput);
-            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.ImplementsPreviewInterfaceRule).WithLocation(0).WithArguments("IZoo", "IFoo"));
+            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.ImplementsPreviewInterfaceRule).WithLocation(0).WithArguments(string.Format(CultureInfo.CurrentCulture, (string)DetectPreviewFeatureAnalyzer.s_implementsPreviewInterfaceMessage, "IZoo", "IFoo"), string.Format(CultureInfo.CurrentCulture, (string)DetectPreviewFeatureAnalyzer.s_detectPreviewFeaturesUrl, DetectPreviewFeatureAnalyzer.DefaultURL)));
```
pgovind, line 54: These test changes are mechanical. Nothing to review here. I just used a regex find and replace to replace the simple arguments with the new `internal static` strings that are defined in the analyzer. There should be no visible differences to the end user in the diagnostic

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.Misc.cs
```diff
@@ -105,7 +106,88 @@ static void Main(string[] args)
 ";
 
             var test = TestCS(csInput);
-            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(0).WithArguments("DerivedException"));
+            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(0).WithArguments(string.Format(CultureInfo.CurrentCulture, (string)DetectPreviewFeatureAnalyzer.s_detectPreviewFeaturesMessage, "DerivedException"), string.Format(CultureInfo.CurrentCulture, (string)DetectPreviewFeatureAnalyzer.s_detectPreviewFeaturesUrl, DetectPreviewFeatureAnalyzer.DefaultURL)));
+            await test.RunAsync();
+        }
+
+        [Fact]
```
pgovind, line 113: The following unit tests are the ones that really test the new paths

src/Test.Utilities/AdditionalMetadataReferences.cs
```diff
@@ -96,7 +96,7 @@ public static class AdditionalMetadataReferences
                     "net6.0",
                     new PackageIdentity(
                         "Microsoft.NETCore.App.Ref",
-                        "6.0.0-preview.6.21352.12"),
+                        "6.0.0-rc.1.21451.13"),
```
pgovind, line 99: The new `Message` and `Url` properties on `RequiresPreviewFeatureAttribute` shipped in RC1

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
```diff
@@ -160,15 +160,17 @@ private static void AddValue(ArrayBuilder<PlatformMethodValue> infosBuilder, str
                     using var infosBuilder = ArrayBuilder<PlatformMethodValue>.GetInstance();
                     if (PlatformMethodValue.TryDecode(method, visitedArguments, DataFlowAnalysisContext.ValueContentAnalysisResult, _osPlatformType, infosBuilder))
                     {
+                        var version = EmptyVersion;
```
pgovind, line 163: The PlatformCompatAnalyzer fails in this PR without these changes from Buyaa. IIRC, this is because this PR is updating the unit tests to use the new RC1 sdk.

buyaa-n: Right, some tests failed because of the sdk update, which needed a test update, plus revealed this bug. Please let me know if a separate PR is needed for this fix cc @jeffhandley

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
```diff
@@ -160,15 +160,17 @@ private static void AddValue(ArrayBuilder<PlatformMethodValue> infosBuilder, str
...
                         {
                             var newValue = GlobalFlowStateAnalysisValueSet.Create(infosBuilder[i]);
                             value = i == 0 ? newValue : GlobalFlowStateAnalysis.GlobalFlowStateAnalysisValueSetDomain.Instance.Merge(value, newValue);
+                            version = infosBuilder[i].Version;
                         }
 
                         infosBuilder.Clear();
                         var attributes = method.GetAttributes();
-                        if (HasAnyGuardAttribute(attributes, out var mappedAttributes) && TryDecodeGuardAttributes(mappedAttributes, infosBuilder))
+                        if (HasAnyGuardAttribute(attributes, version, out var mappedAttributes) && TryDecodeGuardAttributes(mappedAttributes, infosBuilder))
```
buyaa-n, line 173: This method is only called for `OperatingSystem.IsXYZ()` methods, the guard attributes added into `OperatingSystem.IsXYZ()` methods will not have a version, instead its version should be set the same as the version of the `OperatingSystem.IsXYZVersionAtLeast(int,int,int,int)` parameter. Therefore added an overload for `HasAnyGuardAttribute(...) ` method which parses the attributes, the functionality is same as original method, just version is passed from  `OperatingSystem.IsXYZVersionAtLeast(int,int,int,int)` parameter

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.OperationVisitor.cs
jeffhandley: It's OK with me to keep it combined, assuming that makes the process simpler.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1720,7 +1720,7 @@
...
   </data>
   <data name="DetectPreviewFeaturesMessage" xml:space="preserve">
-    <value>Using '{0}' requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+    <value>Using '{0}' requires opting into preview features.</value>
```
buyaa-n, line 1723: Why not update the messages with required format? Like: `{2} Using '{0}' requires opting into preview features. See {1} for more information.` Maybe it should divided into to resource string:

 - if there is a message : `{2} Using '{0}' requires opting into preview features. See {1} for more information.`
 - if no message : `Using '{0}' requires opting into preview features. See {1} for more information.`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1732,10 +1732,10 @@
     <value>Converting an instance method to a static method may produce invalid code</value>
   </data>
   <data name="DerivesFromPreviewClassMessage" xml:space="preserve">
-    <value>'{0}' derives from preview class '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+    <value>'{0}' derives from preview class '{1}' and therefore needs to opt into preview features.</value>
   </data>
   <data name="ImplementsPreviewInterfaceMessage" xml:space="preserve">
-    <value>'{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+    <value>'{0}' implements the preview interface '{1}' and therefore needs to opt into preview features.</value>
```
buyaa-n, line 1738: This one could become:
 - '{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See {2} for more information.
 - {3} '{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See {2} for more information.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
buyaa-n: Hm, once i had a comment do not use manual string concatenation, each case should have its own resource string:
Is the resource string versions above could work for this:

- When no message: `'{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See {2} for more information.`
- When there is a message: `{3} '{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See {2} for more information.`

buyaa-n: Or we could just simply use for both `{3} '{0}' implements the preview interface '{1}' and therefore needs to opt into preview features. See {2} for more information.` so that when there is no message there will be one space before the message

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -152,6 +166,7 @@ public override void Initialize(AnalysisContext context)
                 }
 
                 ConcurrentDictionary<ISymbol, bool> requiresPreviewFeaturesSymbols = new();
+                ConcurrentDictionary<ISymbol, ValueTuple<string?, string?>> previewSymbolToMessageAndUrl = new();
```
buyaa-n, line 169: Why did not use Could not use the above dictionary? It seems to me now we have cache with similar items. I guess that is what you meant by `I could fold the new ConcurrentDictionary into the existing SymbolIsAnnotatedAsPreview call`,  for me that looks better, up to you

buyaa-n:
> if we wanted to be uber performant, I could fold the new ConcurrentDictionary into the existing `SymbolIsAnnotatedAsPreview` call. That would however mean changes to each caller and it would quickly mean touching every routine in the analyzer. Instead, I decided to just look for the `Url` and `Message` property once per symbol just before we report a diagnostic.

It seems to me you can use only one existing `ConcurrentDictionary` and keep all info (bool, url, message) together, by setting them once in `SymbolIsAnnotatedAsPreview`, doesn't seem need much change, but this is up to you. 

I have more stronger feeling for manual message concatenation, better to add placeholder for `message` and `url` for each resource string, this is not optional comment 😉

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Cool, I ended up doing the 2nd suggestion.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Classes.cs
pgovind: N/A anymore

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Working on that now.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Tasks/UseValueTaskCorrectlyTests.cs
```diff
@@ -123,7 +123,7 @@ Class C
                         Await New ValueTask()
                         Await Helpers.ReturnsValueTask()
                         Await New ValueTask(Of String)("", hello, "")
-                        Await New ValueTask(Of String)(Task.FromResult("", hello, ""))
+                        Await New ValueTask(Of String)(Task.FromResult(hello, "", ""))
```
pgovind, line 126: Revert

pgovind:
> It seems to me you can use only one existing ConcurrentDictionary and keep all info (bool, url, message) together, by setting them once in SymbolIsAnnotatedAsPreview, doesn't seem need much change, but this is up to you.

Yup, I fixed this now to use the existing ConcurrentDictionary.

> I have more stronger feeling for manual message concatenation, better to add placeholder for message and url for each resource string, this is not optional comment 😉

Changed this too. Now, there are 2 diagnostics defined: 1 for the regular case, and 1 for the case with the custom messages and URL

This is ready to review again

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1731,29 +1731,29 @@
   <data name="MakeMethodDeclaredOnImplementationTypeStaticMayProduceInvalidCode" xml:space="preserve">
     <value>Converting an instance method to a static method may produce invalid code</value>
   </data>
-  <data name="DerivesFromPreviewClassMessage" xml:space="preserve">
-    <value>'{0}' derives from preview class '{1}' and therefore needs to opt into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+  <data name="DerivesFromPreviewClassMessageWithCustomMessagePlaceholder" xml:space="preserve">
+    <value>'{3}' '{0}' derives from preview class '{1}' and therefore needs to opt into preview features. See '{2}' for more information.</value>
```
buyaa-n, line 1735: Just making sure, do you want the message added within a quote?

Not sure if there is a way to inform the localization team that the placeholder '{3}' is not included in the sentence and should not be moved elsewhere in the sentence. Not sure the `locked` parameter could work for this `<comment>{Locked="{3}"}</comment>`, there might be another parameter exist for this case

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -767,14 +878,160 @@ private static ISymbol SymbolFromArrayCreationOperation(IArrayCreationOperation
...
+                }
+            }
+
+            return message;
+        }
+
+        private static void ReportDiagnosticWithCustomMessageIfItExists(OperationAnalysisContext context,
+                                                                        IOperation operation,
+                                                                        ISymbol symbol,
+                                                                        ConcurrentDictionary<ISymbol, ValueTuple<bool, string?, string?>> requiresPreviewFeaturesSymbols,
```
buyaa-n, line 919: NIT: Using a named value tuple (`existing.url`) would look better than `existing.Item3`

```suggestion
                                                                        ConcurrentDictionary<ISymbol, (bool isPreview, string message, string url)> requiresPreviewFeaturesSymbols,
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -767,14 +878,160 @@ private static ISymbol SymbolFromArrayCreationOperation(IArrayCreationOperation
...
+                    context.ReportDiagnostic(operation.CreateDiagnostic(diagnosticDescriptorWithPlaceholdersForCustomMessage, diagnosticMessageArgument, url, customMessage));
+                }
+                else
+                {
+                    context.ReportDiagnostic(operation.CreateDiagnostic(diagnosticDescriptor, diagnosticMessageArgument, url));
+                }
+            }
+        }
+
+        private static void ReportDiagnosticWithCustomMessageIfItExists(SymbolAnalysisContext context,
```
buyaa-n, line 942: NIT: these overloads implementations are so similar, some parts might could be reused

buyaa-n (approved):
Left a few NIT comments, overall LGTM

pgovind:
Thanks for the stamp Buyaa. I'm just waiting for a review from one of the analyzer folks to address all the feedback in 1 go and merge.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -151,7 +236,7 @@ public override void Initialize(AnalysisContext context)
                     return;
                 }
 
-                ConcurrentDictionary<ISymbol, bool> requiresPreviewFeaturesSymbols = new();
+                ConcurrentDictionary<ISymbol, ValueTuple<bool, string?, string?>> requiresPreviewFeaturesSymbols = new();
```
jmarolf, line 239: Can we use the tuple syntax so these three elements have comprehensible names?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -767,14 +878,160 @@ private static ISymbol SymbolFromArrayCreationOperation(IArrayCreationOperation
...
-                if (symbol.HasAttribute(previewFeatureAttribute))
+                if (constructorArguments.First().Value is string messageValue)
                 {
-                    requiresPreviewFeaturesSymbols.GetOrAdd(symbol, true);
-                    return true;
+                    message = messageValue;
+                }
+            }
+
+            ImmutableArray<System.Collections.Generic.KeyValuePair<string, TypedConstant>> namedArguments = attribute.NamedArguments;
```
jmarolf, line 896: Can we add a using for `System.Collections.Generic`?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
jmarolf: yeah I agree

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
jmarolf: agree, just make all the diagnosticMessageArgument0..X optional parameters and check if they are null in a single methid instead of us having to maintain identical copies

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Dependencies.cs
```diff
@@ -68,9 +68,9 @@ public void ProgramMethod()
...
-            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(0).WithArguments("AMethod"));
-            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(1).WithArguments("Library"));
-            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(2).WithArguments("AProperty"));
+            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(0).WithArguments("AMethod", DetectPreviewFeatureAnalyzer.DefaultURL));
+            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(1).WithArguments("Library", DetectPreviewFeatureAnalyzer.DefaultURL));
+            test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(2).WithArguments("AProperty", DetectPreviewFeatureAnalyzer.DefaultURL));
```
jmarolf, line 73: Why are we mising diagnostic styles here?

either we don't need the `{||}` parse holes in `csCurrentAssemblyCode` or we don't need to manually specify the diagnostics here right?

jmarolf (changes requested):
Have a few things that need fixing and then we are good to go. I also don't remember why we specify the diagnostic locations in the test twice using two separate approaches

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
pgovind: Ah yea each method is just different enough that I made a new method for it :) Will clean it up in the final pass if there's an intuitive way to refactor the methods

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Dependencies.cs
pgovind: > Why are we mising diagnostic styles here?

What do diagnostic styles mean? I don't know that term :/

Hmm, not sure I completely understand, but the analyzer reports multiple diagnostics, so I need to use the `{| |}` syntax and add the diagnostics explicitly in the `ExpectedDiagnostics.Add` method right? Is there another way?

jmarolf: > but the analyzer reports multiple diagnostics

But the additional diagnostics have locations right? So you can do `{|{||}|}` to nest diagnostics whose locations overlap. I would prefer we use the brackets syntax as its easier to visually understand where the diagnostics are appearing.

pgovind: > But the additional diagnostics have locations right?

Wait maybe I wasn't being clear. I meant that DetectPreviewFeaturesAnalyzer can report ~17 different diagnostics, not that it's reporting the same diagnostic in multiple places. So the only option in the unit tests is to use **both** the `{| |}` syntax **and** the `ExpectedDiagnostics.Add` method IIRC

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureAnalyzer.cs
```diff
@@ -767,14 +879,160 @@ private static ISymbol SymbolFromArrayCreationOperation(IArrayCreationOperation
...
+#pragma warning restore CA1054,CA1055
         {
-            if (!requiresPreviewFeaturesSymbols.TryGetValue(symbol, out bool existing))
+            string? message = null;
+            url = null;
+            ImmutableArray<TypedConstant> constructorArguments = attribute.ConstructorArguments;
+            if (constructorArguments.Length != 0)
             {
-                if (symbol.HasAttribute(previewFeatureAttribute))
+                if (constructorArguments.First().Value is string messageValue)
```
Youssef1313, line 891: Named arguments can make this not the first argument.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1719,8 +1719,8 @@
   <data name="DetectPreviewFeaturesDescription" xml:space="preserve">
     <value>An assembly has to opt into preview features before using them.</value>
   </data>
-  <data name="DetectPreviewFeaturesMessage" xml:space="preserve">
-    <value>Using '{0}' requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information.</value>
+  <data name="DetectPreviewFeaturesMessageWithCustomMessagePlaceholder" xml:space="preserve">
+    <value>{2} Using '{0}' requires opting into preview features. See {1} for more information.</value>
```
buyaa-n, line 1723: The url here added without quote but for other messages enclosed in a quite, want to make it same for all, which message looks better?
`Using 'Foo' requires opting into preview features. See https://aka.ms/aspnet/kestrel/http3reqs for more information.` or 
`Using 'Foo' requires opting into preview features. See 'https://aka.ms/aspnet/kestrel/http3reqs' for more information.`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Dependencies.cs
buyaa-n: > So you can do {|{||}|} to nest diagnostics whose locations overlap.

@jmarolf  I might be missing something, but i don't see such overlaps in the diagnostics

jmarolf (approved):

jeffhandley (approved):
I pushed commits to remove the quotes around the Urls and to improve a little formatting/indentation.

PR #5506, opened 2021-09-19
Suspicious Cast from Char to Int32 Analyzer

NewellClark:
Fix [#47927](https://github.com/dotnet/runtime/issues/47927)
I noticed there wasn't a category for Correctness in DiagnosticCategoriesAndIdRanges.txt, so I went ahead and added one. However, when I did, I found that there are a couple analyzers that use this category, even though it wasn't present in the file, so I had to add an RS range of IDs as well. 

The analyzer reports a diagnostic on implicit char-to-int conversions on arguments to any `StringBuilder` constructor, as well as `string.Split(char, int, StringSplitOptions)` and `string.Split(string, int, StringSplitOptions)`. 

The fixer converts `string.Split` calls to either `string.Split(char[], StringSplitOptions)` or `string.Split(string[], StringSplitOptions)`, and converts `StringBuilder(int)` calls to `StringBuilder(string)` calls. Violations in other `StringBuilder` constructors are reported but not fixed.

I ran the analyzer against dotnet/runtime and found nothing. I tried to run the analyzer against roslyn-analyzers, but a bunch of other analyzers (not my analyzer) kept throwing exceptions. Maybe I'm running against roslyn-analyzers incorrectly?


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.cs
```diff
@@ -0,0 +1,156 @@
...
+namespace Microsoft.NetCore.Analyzers.Runtime
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+    public sealed class SuspiciousCastFromCharToIntAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string RuleId = "CA3200";
+
+        private static readonly LocalizableString s_localizableTitle = Resx.CreateLocalizableResourceString(nameof(Resx.SuspiciousCastFromCharToIntTitle));
+        private static readonly LocalizableString s_localizableMessage = Resx.CreateLocalizableResourceString(nameof(Resx.SuspiciousCastFromCharToIntMessage));
+        private static readonly LocalizableString s_localizableDescription = Resx.CreateLocalizableResourceString(nameof(Resx.SuspiciousCastFromCharToIntDescription));
```
Youssef1313, line 20: This pattern was changed in https://github.com/dotnet/roslyn-analyzers/pull/5406.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.cs
```diff
@@ -0,0 +1,156 @@
...
+                StringType = stringType;
+            }
+
+            public static bool TryGetSymbols(Compilation compilation, out RequiredSymbols result)
+            {
+                var charType = compilation.GetSpecialType(SpecialType.System_Char);
+                var int32Type = compilation.GetSpecialType(SpecialType.System_Int32);
+                var stringType = compilation.GetSpecialType(SpecialType.System_String);
+
+                if (charType is not null && int32Type is not null && stringType is not null)
```
Youssef1313, line 116: Only `GetTypeByMetadataName` can return null. `GetSpecialType` doesn't.

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
```diff
@@ -19,6 +19,7 @@ Naming: CA1700-CA1727
 Interoperability: CA1400-CA1419
 Maintainability: CA1500-CA1509
 Reliability: CA9998-CA9999, CA2000-CA2018
+Correctness: CA3200, RS1000-RS1999
```
Youssef1313, line 22: I don't think we need to introduce a new category. It's likely that the category is unintentionally wrong written in the runtime issue. cc @bartonjs

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.cs
```diff
@@ -0,0 +1,156 @@
...
+            return;
+
+            //  Local functions
+
+            bool IsImplicitCharToIntConversion(IArgumentOperation argument)
+            {
+                return argument.Value is IConversionOperation conversion &&
+                    conversion.IsImplicit &&
+                    conversion.Operand.Type.Equals(symbols.CharType, SymbolEqualityComparer.Default) &&
+                    argument.Parameter.Type.Equals(symbols.Int32Type, SymbolEqualityComparer.Default);
```
Youssef1313, line 94:
```suggestion
                    conversion.Operand.Type.SpecialType == SpecialType.System_Char &&
                    argument.Parameter.TypeSpecialType == SpecialType.System_Int32;
```

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
mavasani: Agreed, please don't add this category. Also RS* rules do not belong here.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToIntTests.cs
```diff
@@ -0,0 +1,225 @@
...
+public class Testopolis
+{{
+    public void M(string s, char c, int n)
+    {{
+        new StringBuilder({argumentList});
+    }}
+}}";
+            var diagnostics = Enumerable.Range(0, diagnosticsCount).Select(x => VerifyCS.Diagnostic(Rule).WithLocation(x)).ToArray();
+
+            return VerifyCS.VerifyAnalyzerAsync(source, diagnostics);
```
Youssef1313, line 196: Use VerifyCodeFixAsync even for non-fixable diagnostics. This ensures that we're not offering an "empty" codefix.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.Fixer.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var doc = context.Document;
+            var token = context.CancellationToken;
+            var root = await doc.GetSyntaxRootAsync(token).ConfigureAwait(false);
+            var model = await doc.GetSemanticModelAsync(token).ConfigureAwait(false);
+
+            if (model.GetOperation(root.FindNode(context.Span), token) is not IArgumentOperation argumentOperation)
+                return;
```
Youssef1313, line 32: Could this be parenthesized operation (for example)?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.Fixer.cs
```diff
@@ -0,0 +1,257 @@
...
+
+            var cache = CreateFixerCache(model.Compilation);
+
+            if (!cache.TryGetValue(targetMethod, out var fixer))
+                return;
+
+            var codeAction = CodeAction.Create(
+                MicrosoftNetCoreAnalyzersResources.SuspiciousCastFromCharToIntCodeFixTitle,
+                token => fixer.GetChangedDocumentAsync(new FixCharCastContext(context, argumentOperation.Parent, this, token)),
+                MicrosoftNetCoreAnalyzersResources.SuspiciousCastFromCharToIntCodeFixTitle);
```
Youssef1313, line 52:
```suggestion
                nameof(MicrosoftNetCoreAnalyzersResources.SuspiciousCastFromCharToIntCodeFixTitle));
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.cs
NewellClark: @Youssef1313 What about when building mscorelib? iirc I believe I was told in the past that `GetSpecialType` can very rarely return null.

src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
NewellClark: @Youssef1313 @mavasani Okay. Which category should I use? Reliability?

mavasani: Reliability sounds fine to me.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.Fixer.cs
NewellClark: @Youssef1313 No. `FindNode` always finds the outermost node, unless `getInnermostNodeForTie` is specified as `true`. If a parenthesized operation were passed as the argument, it would be the `Value` property of the argument operation. I will check locations that access the `Value` property; you're right that I may need some calls to `WalkDownParentheses` in those locations.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToInt.cs
NewellClark: Good to know.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/SuspiciousCastFromCharToIntTests.cs
NewellClark: Fixed.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpSuspiciousCastFromCharToInt.Fixer.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.NetCore.Analyzers.Runtime;
+
+namespace Microsoft.NetCore.CSharp.Analyzers.Runtime
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
```
Youssef1313, line 10: Should have `[Shared]` attribute

src/Utilities/Compiler/Extensions/IOperationExtensions.cs
```diff
@@ -57,6 +57,20 @@ internal static partial class IOperationExtensions
             return typeInfo.Type as INamedTypeSymbol;
         }
 
+        /// <summary>
+        /// Gets the arguments for the current <see cref="IInvocationOperation"/> ordered by their respective parameter's <see cref="IParameterSymbol.Ordinal"/> property.
+        /// </summary>
+        public static ImmutableArray<IArgumentOperation> GetArgumentsInParameterOrder(this IInvocationOperation invocation)
```
Youssef1313, line 63: There is already an extension for that.

https://github.com/dotnet/roslyn-analyzers/blob/01f38d805ae9254a5615614a7dd56e792c72f1ea/src/Utilities/Compiler/Extensions/IOperationExtensions.cs#L810-L823

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/Runtime/CSharpSuspiciousCastFromCharToInt.Fixer.cs
NewellClark: Fixed.

mavasani:
Ping @NewellClark

stephentoub:
@NewellClark, are you still working on this?

NewellClark:
@stephentoub My apologies. I completely forgot I was assigned to that. I will resume work on it immediately.

buyaa-n:
Ping @NewellClark are you still planning to work on this?

PR #5509, opened 2021-09-20
[release/7.0.1xx] Migrate to 1ES hosted pools

lpatalas:
We need to migrate pools to 1ES hosted pools. Only the yamls have to be changed. The supported functionality stays the same.

Tracking issue: https://github.com/dotnet/core-eng/issues/14276
/CC: @jonfortescue


lpatalas:
Branch no longer exists.

PR #5535, opened 2021-09-28
Update messages for symbol comparison diagnostic and codefix

### LINKED ISSUE #3427: Incorrect warning RS1024 Compare symbols correctly
**Version Used**: 3.5.0

**Steps to Reproduce**:
Compare two `ISymbol` through their `IEquatable<ISymbol>` interface
```csharp
ISymbol s1, s2;
// hovering indicates this is not Object.Equals(object) 
// but IEquatable<ISymbol>.Equals(ISymbol)
var eq = s1.Equals(s2); 
```

**Expected Behavior**:
No warning.

**Actual Behavior**:
Warning RS1024 _Compare symbols correctly_

ryzngard:
Fixes #3427

New messages: 

Title: `Symbols should be compared for equality`
Message: `Use 'SymbolEqualityComparer' when comparing symbols` 
Description: `Symbols should be compared for equality, not identity. Use an overload accepting an 'IEqualityComparer' and pass 'SymbolEqualityComparer'.`

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
@@ -341,13 +341,13 @@
     <value>Symbols should be compared for equality, not identity. Use an overload accepting an 'IEqualityComparer' and pass 'SymbolEqualityComparer'.</value>
   </data>
   <data name="CompareSymbolsCorrectlyMessage" xml:space="preserve">
-    <value>Compare symbols correctly</value>
+    <value>Use `SymbolEqualityComparer` when comparing symbols</value>
   </data>
   <data name="CompareSymbolsCorrectlyTitle" xml:space="preserve">
-    <value>Compare symbols correctly</value>
+    <value>Symbols should be compared for equality</value>
   </data>
   <data name="CompareSymbolsCorrectlyCodeFix" xml:space="preserve">
-    <value>Compare symbols correctly</value>
+    <value>Use a `SymbolEqualityComparer` for symbol comparison</value>
   </data>
   <data name="ConfigureGeneratedCodeAnalysisMessage" xml:space="preserve">
     <value>Configure generated code analysis</value>
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
@@ -341,13 +341,13 @@
     <value>Symbols should be compared for equality, not identity. Use an overload accepting an 'IEqualityComparer' and pass 'SymbolEqualityComparer'.</value>
   </data>
   <data name="CompareSymbolsCorrectlyMessage" xml:space="preserve">
-    <value>Use `SymbolEqualityComparer` when comparing symbols</value>
+    <value>Use 'SymbolEqualityComparer' when comparing symbols</value>
   </data>
   <data name="CompareSymbolsCorrectlyTitle" xml:space="preserve">
     <value>Symbols should be compared for equality</value>
   </data>
   <data name="CompareSymbolsCorrectlyCodeFix" xml:space="preserve">
-    <value>Use a `SymbolEqualityComparer` for symbol comparison</value>
+    <value>Use a 'SymbolEqualityComparer' for symbol comparison</value>
   </data>
   <data name="ConfigureGeneratedCodeAnalysisMessage" xml:space="preserve">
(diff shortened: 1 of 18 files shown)


mavasani (approved):

mavasani:
> One or more auto-generated documentation files were either edited manually, or not updated. Please revert changes made to the following files (if manually edited) and run `msbuild /t:pack` for each solution at the root of the repo to automatically update them:
      D:\workspace\_work\1\s\src\Microsoft.CodeAnalysis.Analyzers\Microsoft.CodeAnalysis.Analyzers.md
      D:\workspace\_work\1\s\src\Microsoft.CodeAnalysis.Analyzers\Microsoft.CodeAnalysis.Analyzers.sarif
      D:\workspace\_work\1\s\src\Microsoft.CodeAnalysis.Analyzers\RulesMissingDocumentation.md

@ryzngard You need to run `msbuild /t:pack` at the root of the repo and push the changes.

jmarolf (approved):

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -341,13 +341,13 @@
     <value>Symbols should be compared for equality, not identity. Use an overload accepting an 'IEqualityComparer' and pass 'SymbolEqualityComparer'.</value>
   </data>
   <data name="CompareSymbolsCorrectlyMessage" xml:space="preserve">
-    <value>Compare symbols correctly</value>
+    <value>Use 'SymbolEqualityComparer' when comparing symbols</value>
   </data>
   <data name="CompareSymbolsCorrectlyTitle" xml:space="preserve">
-    <value>Compare symbols correctly</value>
+    <value>Symbols should be compared for equality</value>
```
Youssef1313, line 347: If I understand correctly, this is misleading. A code violating this rule is already comparing for equality. e,g `symbolA.Equals(symbolB)`, and this is completely equivalent to `symbolA.Equals(symbolB, SymbolEqualityComparer.Default)`.

The thing is mostly about the "intent" when there are nullable differences.

src/Microsoft.CodeAnalysis.Analyzers/Core/CodeAnalysisDiagnosticsResources.resx
```diff
@@ -341,13 +341,13 @@
     <value>Symbols should be compared for equality, not identity. Use an overload accepting an 'IEqualityComparer' and pass 'SymbolEqualityComparer'.</value>
```
Youssef1313, line 341: The existing description (not touched in this PR) doesn't make much sense to me for the same reasoning in https://github.com/dotnet/roslyn-analyzers/pull/5535#r717964748.

ryzngard: Equality kind of stretches into inclusion/exclusion of nullable here. Shorthand to say "Should compare for equality" makes some sense to me, but I definitely see where the ", not identity" can be misleading.

Youssef1313: > Equality kind of stretches into inclusion/exclusion of nullable here


That's exactly it. this is what should be indicated by the diagnostic.

Probably:

- Title: `Make your symbol comparison intent explicit`
- Message: `Specify whether you want to include nullability when comparing symbols`
- Description: `When comparing symbols, explicitly specify a 'SymbolEqualityComparer' to indicate whether you want to include nullability in comparison`

(I know my wording isn't very good, but something along those lines?)

ryzngard: What about: 

* Title: `Symbol comparison should use 'SymbolEqualityComparer'`
* Message: `Use a 'SymbolEqualityComparer' to specify whether to include nullability when comparing symbols`
* Description: `When comparing symbols, it's good practice to specify whether symbols with different nullability should be considered equal. Use a 'SymbolEqualityComparer' to specify in this comparison whether  to include nullability as part of the equality check.`

ryzngard: That way the title tells you what we expect, the message and description both elaborate on the _why_

Youssef1313: Much better!

ryzngard: Thanks for the feedback :) I'll see if I can get that in this afternoon

PR #5540, opened 2021-09-28
Add milestone tracking automation (mutation disabled)

sharwell:
Implements the following rule, but does not enable the mutation at this point.

MicrosoftTeams-image

### DIFF
--- a/.github/workflows/milestone-tracking.yml
+++ b/.github/workflows/milestone-tracking.yml
@@ -1,0 +1,42 @@
+name: Milestone tracking
+on:
+  pull_request:
+    types: [closed]
+jobs:
+  add_milestone:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'dotnet/roslyn' && github.event.pull_request.merged_at != null && github.event.pull_request.milestone == null && github.event.pull_request.base.ref == 'main' }}
+    steps:
+    - name: Get milestone data
+      env:
+        ORGANIZATION: dotnet
+        REPOSITORY: roslyn-analyzers
+        MILESTONE_NAME: vNext
+      run: |
+        gh api graphql -f query='
+          query($org: String!, $repo: String!, $milestone: String!) {
+            repository(name: $repo, owner: $org) {
+              milestones(query: $milestone, first: 2) {
+                nodes {
+                  id
+                  title
+                }
+              }
+            }
+          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -f milestone="$MILESTONE_NAME" > milestone_data.json
+
+        cat milestone_data.json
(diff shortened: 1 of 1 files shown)


.github/workflows/milestone-tracking.yml
```diff
@@ -0,0 +1,41 @@
+name: Milestone tracking
+on:
+  pull_request:
+    types: [closed]
+jobs:
+  add_milestone:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'dotnet/roslyn' && github.event.pull_request.merged_at != null && github.event.pull_request.milestone == null && (github.event.pull_request.base.ref == 'main' || github.event.pull_request.base.ref == 'main-vs-deps') }}
```
jmarolf, line 8: we don't need the `main-vs-deps` statement for this repo

.github/workflows/milestone-tracking.yml
```diff
@@ -0,0 +1,41 @@
...
+          query($org: String!, $repo: String!, $milestone: String!) {
+            repository(name: $repo, owner: $org) {
+              milestones(query: $milestone, first: 2) {
+                nodes {
+                  id
+                  title
+                }
+              }
+            }
+          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -f milestone="$MILESTONE_NAME" > milestone_data.json
```
jmarolf, line 26: where is `MILESTONE_NAME` going to be defined?

sharwell: Line 14 above

.github/workflows/milestone-tracking.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Get milestone data
+      env:
+        ORGANIZATION: dotnet
+        REPOSITORY: roslyn-analyzers
+        MILESTONE_NAME: vNext
+      run: |
+        gh api graphql -f query='
+          query($org: String!, $repo: String!, $milestone: String!) {
+            repository(name: $repo, owner: $org) {
+              milestones(query: $milestone, first: 2) {
```
jmarolf, line 19: why `first: 2` shouldn't there only be one milestone with a matching name?

jmarolf: ah, its the first 2 nodes

jmarolf (approved):

.github/workflows/milestone-tracking.yml
sharwell: I wanted to use the equivalent of `.Single()` for validation, but couldn't find it. The query appears to be a substring match, and the docs aren't clear about how to be more specific.

sharwell: ➡️ Validation is now added

PR #5556, opened 2021-09-30
Enable auto-merge for OneLoc pull requests

sharwell:
Implements the following rule using GitHub Actions automation:

MicrosoftTeams-image (1)

### DIFF
--- a/.github/workflows/enable-auto-merge.yml
+++ b/.github/workflows/enable-auto-merge.yml
@@ -1,0 +1,29 @@
+name: Enable auto merge
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+permissions:
+  issues: write
+  pull-requests: write
+jobs:
+  add_milestone:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/enable-auto-merge.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+permissions:
+  issues: write
+  pull-requests: write
+jobs:
+  add_milestone:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'dotnet/roslyn-analyzers' && github.event.pull_request.author.login == 'dotnet-bot' && startsWith(github.event.pull_request.title, 'Localized file check-in') }}
```
jmarolf, line 11: The title will be something like "Localized file check-in by OneLocBuild Task: Build definition ID 830: Build ID 1208815". So we would want some way to say `if it starts with 'Localized file check-in'`

sharwell: I think you need to scroll right 😉

jmarolf: ah I see it :)

jmarolf (approved):

PR #5566, opened 2021-09-30
VB support for the virtual methods in DetectPreviewFeatureAnalyzer

pgovind:
Just leaving this PR up for now. This builds on top of https://github.com/dotnet/roslyn-analyzers/pull/5502. 

1. Once #5502 goes in, this PR needs to be rebased on the latest release/6.0.1xx. **Until then, please only look at the latest commit for changes specific to this PR**

What this PR does: It overrides all the virtual methods in DetectPreviewFeaturesAnalyzer **EXCEPT** `GetConstraintSyntaxNodeForTypeConstrainedByPreviewTypes`.

Pending work:
1. Implement `GetConstraintSyntaxNodeForTypeConstrainedByPreviewTypes` (can be it's own PR)
2. I've only added smoke tests to this PR. While all new code that's been added is being tested in the smoke tests, they are by no means exhaustive. One suggestion here is to create an issue to improve test coverage for VB, and mark it as up for grabs? Or, we could roll with just basic smoke tests for VB considering it's usage.

cc @jeffhandley cc @buyaa-n


src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicDetectPreviewFeatureAnalyzer.vb
```diff
@@ -11,23 +12,262 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
...
+                        End If
+                    End If
+                End If
+            Next
+
             Return Nothing
         End Function
 
         Protected Overrides Function GetConstraintSyntaxNodeForTypeConstrainedByPreviewTypes(typeOrMethodSymbol As ISymbol, previewInterfaceConstraintSymbol As ISymbol) As SyntaxNode
             Return Nothing
```
buyaa-n, line 171: I have tried to implement this but could not find a way to get implementation constraints from `ClassStatementSyntax` in VB, [filed an issue](https://github.com/dotnet/roslyn-analyzers/issues/5646) for that.

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicDetectPreviewFeatureAnalyzer.vb
```diff
@@ -11,23 +12,271 @@ Namespace Microsoft.NetCore.VisualBasic.Analyzers.Runtime
 
         Inherits DetectPreviewFeatureAnalyzer
 
+        Private Function IsSyntaxToken(identifier As SyntaxToken, previewInterfaceSymbol As ISymbol) As Boolean
+            Return identifier.ValueText.Equals(previewInterfaceSymbol.Name)
```
Youssef1313, line 16: Should be case insensitive?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Interfaces.cs
```diff
@@ -37,12 +40,38 @@ public interface IProgram
...
+
+        Namespace Preview_Feature_Scratch
+            Class Program
+                Implements IProgram
+
+                Private Shared Sub Main(ByVal args As String())
+                    Dim prog = New Program()
+                End Sub
+
+                Public Sub {|#0:MarkedMethodInInterface|}() Implements IProgram.MarkedMethodInInterface
```
jeffhandley, line 61: Could we apply the diagnostic like this instead?
```suggestion
                Public Sub MarkedMethodInInterface() Implements {|#0:Program.MarkedMethodInInterface|}
```

Or:
```suggestion
                Public Sub MarkedMethodInInterface() Implements IProgram.{|#0:MarkedMethodInInterface|}
```

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicDetectPreviewFeatureAnalyzer.vb
jeffhandley: Good catch! And we'll need tests that verify that VB is handled without case sensitivity.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Interfaces.cs
```diff
@@ -70,13 +99,39 @@ public interface IProgram
...
+        Imports System
+        Imports System.Runtime.Versioning
+        Module Preview_Feature_Scratch
+            Public Class Program
+                Implements {|#1:IProgram|}
+                Private Shared Sub Main(ByVal args As String())
+                    Dim prog = New Program()
+                End Sub
+
+                Public Sub {|#0:MarkedMethodInInterface|}() Implements IProgram.MarkedMethodInInterface
```
jeffhandley, line 119: Same as above (could we use the location of `IProgram.MarkedMethodInterface` or `MarkedMethodInterface` instead?).

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Properties.cs
```diff
@@ -85,6 +88,40 @@ public class Foo
             test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(5).WithArguments("_value", DetectPreviewFeatureAnalyzer.DefaultURL));
             test.ExpectedDiagnostics.Add(VerifyCS.Diagnostic(DetectPreviewFeatureAnalyzer.GeneralPreviewFeatureAttributeRule).WithLocation(6).WithArguments("_value", DetectPreviewFeatureAnalyzer.DefaultURL));
             await test.RunAsync();
+
+            var vbInput = @" 
+Imports System
+Imports System.Runtime.Versioning
+Imports System.Collections.Generic
+Module Preview_Feature_Scratch
+    Public Class {|#2:AFoo|}(Of T As Foo)
```
jeffhandley, line 97: Could/should this use the location of `Foo` instead of `AFoo`?

src/NetAnalyzers/VisualBasic/Microsoft.NetCore.Analyzers/Runtime/BasicDetectPreviewFeatureAnalyzer.vb
jeffhandley: Ah, I suspect this is the reason why `AFoo` was highlighted above instead of `Foo`!

jeffhandley (commented):
Thanks for catching the C# location issue while doing this. I left a couple comments but I think they may be related to the issue you filed.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureTests.Interfaces.cs
```diff
@@ -70,13 +98,48 @@ public interface IProgram
...
+                    End Get
+                End Property
+            End Class
+
+            <RequiresPreviewFeatures>
+            Public Interface Iprogram
+                Sub MarkedMethodInInterface()
+                ReadOnly Property Value() As String 
+            End Interface
+        End Module
```
jeffhandley, line 134: Please make the case-insensitivity test more apparently intentional. Ideally as a separate test that is named to represent this scenario. We could even have a C# test that shows that C# is treated with case sensitivity and then the VB test showing that it's case-insensitive. As it is now, someone in the future might think these were typos and "correct" them. :wink:

Issue #5565, opened 2021-09-30
Implement automation rule "Label community pull requests"


sharwell:
![MicrosoftTeams-image (4)](https://user-images.githubusercontent.com/1408396/135517746-56ca3a6d-4df9-4dcc-a711-03880984f0b5.png)

sharwell:
@Evangelink this rule should now be working. Can you submit a new PR to this repository as a test?

Youssef1313:
@sharwell It's working 🎉 

https://github.com/dotnet/roslyn-analyzers/pull/5584

PR #5567, opened 2021-09-30
Label community pull requests

sharwell:
See #5565

### DIFF
--- a/.github/workflows/community-pr.yml
+++ b/.github/workflows/community-pr.yml
@@ -1,0 +1,42 @@
+name: Label community PRs
+on:
+  pull_request_target:
+    types: [opened]
+permissions:
+  issues: write
+  pull-requests: write
+jobs:
+  add_milestone:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'dotnet/roslyn-analyzers' && github.event.pull_request.author_association != 'MEMBER' && github.event.pull_request.author_association != 'OWNER' }}
+    steps:
+    - name: Get label data
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        ORGANIZATION: dotnet
+        REPOSITORY: roslyn-analyzers
+        LABEL_NAME: Community
+      run: |
+        gh api graphql -f query='
+          query($org: String!, $repo: String!, $label: String!) {
+            repository(name: $repo, owner: $org) {
+              label(name: $label) {
+                id
+              }
+            }
+          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -f label="$LABEL_NAME" > label_data.json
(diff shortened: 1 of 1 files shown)


.github/workflows/community-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+on:
+  pull_request_target:
+    types: [opened]
+permissions:
+  issues: write
+  pull-requests: write
+jobs:
+  add_milestone:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'dotnet/roslyn-analyzers' && github.event.pull_request.author_association != 'MEMBER' && github.event.pull_request.author_association != 'OWNER' }}
```
sharwell, line 11: 📝 The V4 API (GraphQL) seems to be different from the V3 API for this field. The `author_assocation` field is shown in examples for the V3 API, but GitHub (in line with industry standards) has opted to leave the majority of API behaviors undocumented, so it's not clear what the possible values of this enumeration are or what their meaning is.

dibarbet: Is verifying the repo matches necessary?  Presumably since this is an action for roslyn-analyzers it should only be triggered by PRs to roslyn analyzers?  Or can this be triggered by forks

.github/workflows/community-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+    - name: Get label data
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        ORGANIZATION: dotnet
```
dibarbet, line 16: I think the repo and org can come from the `github` context if you don't want to hardcode it. 
e.g. ${{ github.repository }}
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

dibarbet (approved):

.github/workflows/community-pr.yml
sharwell: > Or can this be triggered by forks

I kept running into problems with microsoft/vs-extension-testing where actions ran in forks even though the documentation says they don't.

.github/workflows/community-pr.yml
dibarbet: I found this which seems to validate the author_association checks too https://docs.github.com/en/graphql/reference/enums#commentauthorassociation

sharwell: Those are the values returned by the V4 API. If you use GraphQL to get pull request #5569, you get `MEMBER` for the association:
image

However, if you use the V3 API to get the [same pull request](https://api.github.com/repos/dotnet/roslyn-analyzers/pulls/5569), it returns an `"author_association": "CONTRIBUTOR"` which has now broken this rule.

PR #5571, opened 2021-09-30
Update the flow analysis result evaluation to account the call site reachability

buyaa-n:
## Customer Impact

This fixes a customer-reported issue related to a combination of `SupportedOSPlatform` attributes along with flow analysis handling of `OperatingSystem` guard methods. With this edge case bug, we report a false-positive diagnostic on a call site, indicating that the call site is reachable on a version of the platform that differs from what is actually reachable.

Here is the specific scenario that was reported:

```cs

[SupportedOSPlatform("android23.0")]
[SupportedOSPlatform("ios13.0")]
[SupportedOSPlatform("windows8.1")]
[UnsupportedOSPlatform("MacCatalyst")]
class MyUsage
{
    public void M()
    {
        if (!OperatingSystem.IsWindows() || OperatingSystem.IsWindowsVersionAtLeast(10, 0, 10240))
        {
            // FALSE POSITIVE DIAGNOSTIC REPORTED
            // CA1416: This call site is reachable on: 'android' 23.0 and later, 'ios' 13.0 and later,
            // 'windows' 8.1 and later. 'MyType' is only supported on: 'windows' 10.0.10240 and later.
            var t = new MyType();
        }
    }
}
[SupportedOSPlatform("android23.0")]
[SupportedOSPlatform("ios13.0")]
[SupportedOSPlatform("windows10.0.10240")]
class MyType { }
```

It is caused by the logic we use for evaluating the flow analysis result, we suppress warnings with 2 steps:

 1. Call site attributes -> with this one the `[SupportedOSPlatform("android23.0")] [SupportedOSPlatform("ios13.0")]` attributes are suppressed (removed), therefore only unmatching `[SupportedOSPlatform("windows10.0.10240")]` will be left for MyType
 2. Check if the flow analysis result `!Windows() || Windows10OrLater()` is guarding the API call, from the OR conditionals the  `!Windows()` is not guarding the API for the remaining attribute `[SupportedOSPlatform("windows10.0.10240")]` therefore there is a warning

The issue is we are not accounting for the call site reachability when evaluating the flow analysis result. The fix here is now accounting for that in the flow analysis result evaluation.

## Testing

We have thorough unit test coverage for related scenarios, and new tests were added for these edge cases.

## Risk

Low. Minimal performance implications of adding additional code during flow analysis evaluation of which platforms/versions can reach the call site.

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzer.cs
@@ -485,9 +485,21 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
                                     }
                                     attribute.UnsupportedFirst = null;
                                 }
-                                else if (value.AnalysisValues.Contains(new PlatformMethodValue(info.PlatformName, EmptyVersion, false)))
+                                else
                                 {
-                                    csAttributes = SetCallSiteUnsupportedAttribute(csAttributes, info);
+                                    if (csAttributes != null &&
+                                        AllowList(attribute) &&
+                                        IsOnlySupportNeedsGuard(info.PlatformName, attributes, csAttributes))
+                                    {
+                                        attribute.SupportedFirst = null;
+                                        attribute.SupportedSecond = null;
+                                        attribute.UnsupportedSecond = null;
+                                        attribute.UnsupportedFirst = null;
+                                    }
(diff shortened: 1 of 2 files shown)


jmarolf (approved):

jeffhandley:
I will send an email to Tactics to request approval of this change for 6.0 GA.

terrajobst (approved):
Haven't looked at the implementation ('cause I lost my PhD in flow analysis) but the tests looked good :-)

jmarolf:
@jeffhandley this was approved so I am enabling auto-merge on this PR.

dotMorten:
Thanks everyone for fixing this so quick. My code base is gonna be a lot happier now 😁

PR #5581, opened 2021-10-01
Fix query and filter logic for labeling community PRs

sharwell:
See notes in the PR.

### DIFF
--- a/.github/workflows/community-pr.yml
+++ b/.github/workflows/community-pr.yml
@@ -28,8 +28,9 @@ jobs:
                 authorAssociation
               }
             }
-          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -f pull="$PULL_REQUEST_NUMBER" -f label="$LABEL_NAME" > data.json
+          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -F pull=$PULL_REQUEST_NUMBER -f label="$LABEL_NAME" > data.json
 
+        cat data.json
         echo 'LABEL_ID='$(jq -r '.data.repository.label.id' data.json) >> $GITHUB_ENV
         echo 'AUTHOR_ASSOCIATION='$(jq -r '.data.repository.pullRequest.authorAssociation' data.json) >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


.github/workflows/community-pr.yml
```diff
@@ -28,16 +28,17 @@ jobs:
                 authorAssociation
               }
             }
-          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -f pull="$PULL_REQUEST_NUMBER" -f label="$LABEL_NAME" > data.json
+          }' -f org=$ORGANIZATION -f repo=$REPOSITORY -F pull=$PULL_REQUEST_NUMBER -f label="$LABEL_NAME" > data.json
```
sharwell, line 31: This fixes the following error:

> Variable $pull of type Int! was provided invalid value

The GitHub CLI uses `-f` for "raw fields" which are always strings, and `-F` for typed fields.

.github/workflows/community-pr.yml
```diff
@@ -28,16 +28,17 @@ jobs:
...
 
+        cat data.json
```
sharwell, line 33: 📝 If the **Assign label** step is ever skipped, the log will not show the reason why it was skipped (which means you also cannot see the value of `AUTHOR_ASSOCIATION`). Logging this data will allow quick corrections if ever needed.

.github/workflows/community-pr.yml
```diff
@@ -28,16 +28,17 @@ jobs:
...
         echo 'LABEL_ID='$(jq -r '.data.repository.label.id' data.json) >> $GITHUB_ENV
         echo 'AUTHOR_ASSOCIATION='$(jq -r '.data.repository.pullRequest.authorAssociation' data.json) >> $GITHUB_ENV
 
     - name: Assign label
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         PULL_REQUEST_ID: ${{ github.event.pull_request.node_id }}
-      if: ${{ env.LABEL_ID != '' && env.LABEL_ID != 'null' && env.AUTHOR_ASSOCIATION == 'MEMBER' }}
+      if: ${{ env.LABEL_ID != '' && env.LABEL_ID != 'null' && env.AUTHOR_ASSOCIATION != 'MEMBER' }}
```
sharwell, line 41: 🤦‍♂️

mavasani (approved):

PR #5592, opened 2021-10-04
Update Loc strings for .NET Analyzers

jmarolf:
A few loc strings got missed. Will need to ask permission to merge these in tactics


jmarolf:
Approved by shiproom

PR #5618, opened 2021-10-11
Updates to build with source-build

dseefeld:
2 changes required to eliminate prebuilts when building with source-build:
1. ~~Update the version of Newtonsoft.Json to 13.0.1.  This is the version that is being built in source-build and the version that other .NET repos are referencing.~~
2. Exclude two projects from the source-build build.  The PerfDiff tool and some unit tests.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -65,7 +65,7 @@
     <MicrosoftNETCoreAppRefVersion>6.0.0-preview.4.21253.7</MicrosoftNETCoreAppRefVersion>
     <MicrosoftVisualBasicVersion>10.1.0</MicrosoftVisualBasicVersion>
     <MicrosoftVisualStudioCompositionVersion>16.1.8</MicrosoftVisualStudioCompositionVersion>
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -65,7 +65,7 @@
     <MicrosoftNETCoreAppRefVersion>6.0.0-preview.4.21253.7</MicrosoftNETCoreAppRefVersion>
     <MicrosoftVisualBasicVersion>10.1.0</MicrosoftVisualBasicVersion>
     <MicrosoftVisualStudioCompositionVersion>16.1.8</MicrosoftVisualStudioCompositionVersion>
-    <NewtonsoftJsonVersion>12.0.1</NewtonsoftJsonVersion>
+    <NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
```
sharwell, line 68: ❓  This is only referenced by **PerfDiff.csproj**. If that project is going to be excluded from source build anyway, why do we need to update this property?

dseefeld: Thanks.  I didn't realize that.  I reverted the newtonsoft.json change.

PR #5674, opened 2021-10-22
Switch to 1ES R&D pools on main

jonfortescue:
Per https://github.com/dotnet/core-eng/issues/14276.

cc/ @lpatalas @ulisesh

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,9 +35,9 @@ jobs:
       displayName: Build and Test
 
     - script: |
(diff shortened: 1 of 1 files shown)


jmarolf:
@jonfortescue can you add this to the pipeline: https://github.com/dotnet/roslyn-sdk/blob/9e0d174c165bccdc18f8231c4268b1a19f2be420/.vsts-pr.yaml#L52-L55

so we can get dumps of what is happening?

jonfortescue:
@jmarolf added.

jonfortescue:
@jmarolf does it look like we have the logs?

JoeRobich (approved):

PR #5716, opened 2021-11-11
No more noisy pragmas needed for 'new System.IO.MemoryStream'

### LINKED ISSUE #4147: CA2000 Dispose for MemoryStream
### Analyzer

**Diagnostic ID**: [CA2000](https://docs.microsoft.com/visualstudio/code-quality/ca2000)

### Describe the improvement

[MemoryStream](https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=netcore-3.1) doesn't hold anything that needs to be disposed.

```cs
var ms = new MemoryStream(myArray, 0, 10); // CA2000

using var msWithUsing = new MemoryStream(myArray, 0, 10); // Analyzer happy but makes redundant Dispose call.
```

The rule CA2000 should not warn on MemoryStream object.

### Solution

Add MemoryStream/UnmanagedMemoryStream to exclusion list:

```
dotnet_code_quality.CA2000.excluded_symbol_names = MemoryStream|UnmanagedMemoryStream
```

TimLovellSmith:
Fixes #4147.

We all never want to see this code again right!:
```
#pragma warning disable CA2000 // Dispose objects before losing scope
            var stream = new MemoryStream();
#pragma warning restore CA2000 // Dispose objects before losing scope
```

### DIFF
--- a/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
+++ b/src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
@@ -104,6 +104,13 @@ namespace Microsoft.CodeAnalysis.FlowAnalysis.DataFlow.DisposeAnalysis
                     return defaultValue;
                 }
 
+                // Special case: MemoryStream doesn't need to be disposed. Subclasses *might* need to be disposed, but that is the less common case, and the common case is a huge source of noisy warnings.
+                if (MemoryStreamNamedType != null &&
+                    instanceType.Equals(MemoryStreamNamedType))
+                {
+                    return defaultValue;
+                }
+
                 // Handle user option for additional excluded types
                 if (DataFlowAnalysisContext.IsConfiguredToSkipAnalysis(instanceType))
                 {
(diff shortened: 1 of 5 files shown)


mavasani (approved):

mavasani:
@TimLovellSmith Build needs to be fixed. You need to invoke the public API code fix on the newly added protected API to add a new entry in unshipped public API text file

src/Utilities/FlowAnalysis/FlowAnalysis/Analysis/DisposeAnalysis/DisposeAnalysis.DisposeDataFlowOperationVisitor.cs
```diff
@@ -104,6 +104,13 @@ protected override DisposeAbstractValue HandleInstanceCreation(IOperation creati
                     return defaultValue;
                 }
 
+                // Special case: MemoryStream doesn't need to be disposed. Subclasses *might* need to be disposed, but that is the less common case, and the common case is a huge source of noisy warnings.
+                if (MemoryStreamNamedType != null &&
+                    instanceType.Equals(MemoryStreamNamedType))
```
Youssef1313, line 109: Null check is unnecessary here. I believe `Equals` method checks and handles it correctly.

TimLovellSmith: @Youssef1313 I wondered about this, but was initially following the pattern of other null checks on named types in the code. Your feedback made me curious so I wondered why...

The main reason seems to be that the contract of GetOrCreateTypeByMetadataName does not guarantee it will not retrurn null.

So I propose not to fix this feedback..

Youssef1313: `GetOrCreateTypeByMetadataName` can indeed return null. The thing is that `instanceType.Equals(null)` will then return false. But since you're following existing code pattern, it's more appropriate for this to be a separate PR (unless the early null check is known to be faster than `instanceType.Equals(null)` and has a noticeable performance impact - but I doubt this is true).

TimLovellSmith: @Youssef1313 Sounds like the germs of some new code analysis rules in the making... 
"Do not bother checking for null before passing parameters to 'Equals', it usually will just return false anyway."
"Check parameters for null in Equals"

PR #5729, opened 2021-11-18
Replace rtf files by txt files

### LINKED ISSUE #5689: .rtf license files in nuget packages flagged as security risk
Per request from @jmarolf, I'm posting a linked issue here as well as [the original](https://github.com/dotnet/roslyn/issues/57385) on [dotnet/roslyn](https://github.com/dotnet/roslyn).

Original post:

_Based on a security bulletin issued by Microsoft, the enterprise context I work in adopted the (overly?) conservative strategy of blocking all .rtf files (and compressed files that contain them). This includes the Microsoft.CodeAnalysis and Microsoft.CodeAnalysis.Analyzers packages.
This naturally has led issues in developer experience, sometimes in ways that are not immediately apparent and/or take time to identify (especially for new team members).
I see other instances of license/third-part-notices files in this repo using plaintext (.txt) rather than .rtf. Would it be possible to consider (at least for nuget publishing) sticking to .txt or any format other than .rtf?_

Evangelink:
Fixes #5689


Evangelink:
@jmarolf @mavasani I went ahead with `.txt`, let me know if you prefer `.md`.

jmarolf:
@Evangelink thanks for this I think `txt` is the right thing here. I do not think we have any need for more complex formatting

tools/AnalyzerCodeGenerator/AnalyzerGenerator.Utilities/Utilities.cs
```diff
@@ -451,7 +451,7 @@ public static IEnumerable<string> GetCategories(IEnumerable<CheckData> checks)
                 "install.ps1",
                 "uninstall.ps1",
                 "packages.config",
-                "ThirdPartyNotices.rtf"
+                "ThirdPartyNoticestxt"
```
jmarolf, line 454: I think we are missing a period here

Youssef1313: @jmarolf Is this whole project actually used?

jmarolf: This does appear to be an old tool and I am not aware of the history here @mavasani do you remember what this was for?

jmarolf (approved):

tools/AnalyzerCodeGenerator/AnalyzerGenerator.Utilities/Utilities.cs
mavasani: Can’t recall the purpose of this tool, but certainly it is no longer used.

Evangelink: If you want to proceed with merging this PR, I will create another one where I remove the project. Shall it be the project only or the full AnalyzerGenerator solution?

Evangelink: See #5741

PR #5734, opened 2021-11-19
Use a property initializer to guarantee one-time array allocation

Youssef1313:
`ImmutableArray.Create` calls does an array allocation:

https://github.com/dotnet/runtime/blob/d9afc1ecd6f22d836e1d7665847917aa09f580e9/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs#L30-L53

With the previous code, `ImmutableArray.Create` will be called EVERY time `SupportedDiagnostics` is accessed. With an initializer, the call to `ImmutableArray.Create` will only be done once. See [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEDMACdAmbAYQFgAoAbzO2tx3QEYAGAbQF1sBle7AXgD5sMAKYB3XE2YBWVgG4qNLOJbsOBctgDmQgC4zsAX16DRSqbLL6gA===).

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
@@ -54,7 +54,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
 
         internal const string NonStaticMemberIsMethodKey = nameof(NonStaticMemberIsMethodKey);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(
             DynamicInterfaceCastableImplementationUnsupported,
             InterfaceMembersMissingImplementation,
             MembersDeclaredOnImplementationTypeMustBeStatic);
--- a/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
+++ b/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
@@ -153,7 +153,7 @@ namespace Analyzer.Utilities.Extensions
             {
                 IMethodSymbol m => m.Parameters,
                 IPropertySymbol p => p.Parameters,
-                _ => ImmutableArray<IParameterSymbol>.Empty
+                _ => ImmutableArray<IParameterSymbol>.Empty,
             };
         }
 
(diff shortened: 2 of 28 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
```diff
@@ -54,7 +54,7 @@ internal sealed class DynamicInterfaceCastableImplementationAnalyzer : Diagnosti
 
         internal const string NonStaticMemberIsMethodKey = nameof(NonStaticMemberIsMethodKey);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(
```
Evangelink, line 57: I think that @sharwell rejected these changes at some point saying that the perf gain was somehow null as we know the call is only done once (I might remember wrong though).

sharwell: IIRC it was submitted as part of a PR with something else, but yes, these calls only occur once in practice.

src/Utilities/Compiler/Extensions/ISymbolExtensions.cs
```diff
@@ -153,7 +153,7 @@ public static ImmutableArray<IParameterSymbol> GetParameters(this ISymbol? symbo
             {
                 IMethodSymbol m => m.Parameters,
                 IPropertySymbol p => p.Parameters,
-                _ => ImmutableArray.Create<IParameterSymbol>()
+                _ => ImmutableArray<IParameterSymbol>.Empty
```
sharwell, line 156: If you're changing this line, might as well fix the missing `,` at the same time

```suggestion
                _ => ImmutableArray<IParameterSymbol>.Empty,
```

sharwell (approved):

sharwell:
I'm approving on condition that @mavasani also approves

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DynamicInterfaceCastableImplementation.cs
Youssef1313: Side note: The PR doesn't only update `SupportedDiagnostics`, but also a couple of immutable arrays which may (or may not) be accessed multiple times.

mavasani (approved):

Issue #5169, opened 2021-06-25
Any reason not to have DiagnosticDescriptor readonly


Evangelink:
Hi there,

I was browsing the code base and realized that the `DiagnosticDescriptor` declared are `internal static` but not `readonly` while they are not expected to change. Is there a technical (perf?) reason to do so?

Youssef1313:
@Evangelink The repo isn't consistent on this matter, but the majority are not `readonly`. So I don't think there is a reason.

Evangelink:
If there are no technical consideration, it might be worth to fix it (and add an analyzer).

Youssef1313:
There is already IDE0044 for that.

Evangelink:
I forgot about it. That's weird because the issue is turned on but not raised on my VS (is it on yours?), it's caught by a different analyzer.

mavasani:
IDE0044 only works for private fields, not internal ones.

mavasani:
@Evangelink These should be converted to `private static readonly` fields

Youssef1313:
@mavasani Probably this is no longer relevant with changes from #5406 (we no longer have these fields)?

Evangelink:
You are right @Youssef1313. I am closing the ticket.

Youssef1313:
@Evangelink Sorry I was wrong :smile:

#5406 eliminated the LocalizableString fields, not the DiagnosticDescriptor fields. Found this while creating https://github.com/dotnet/roslyn-analyzers/pull/5734

PR #5735, opened 2021-11-19
Make DiagnosticDescriptors readonly

Youssef1313:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5169

### DIFF
--- a/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
+++ b/src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
@@ -14,7 +14,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
     [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
     public sealed class PreferIsKindAnalyzer : DiagnosticAnalyzer
     {
-        internal static DiagnosticDescriptor Rule = new(
+        internal static readonly DiagnosticDescriptor Rule = new(
             DiagnosticIds.PreferIsKindRuleId,
             CreateLocalizableResourceString(nameof(PreferIsKindTitle)),
             CreateLocalizableResourceString(nameof(PreferIsKindMessage)),
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/ProvidePublicParameterlessSafeHandleConstructor.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/ProvidePublicParameterlessSafeHandleConstructor.cs
@@ -15,7 +15,7 @@ namespace Microsoft.NetCore.Analyzers.InteropServices
     {
         internal const string RuleId = "CA1419";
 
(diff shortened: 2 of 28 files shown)


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.CodeAnalysis.Analyzers.MetaAnalyzers
     [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
     public sealed class PreferIsKindAnalyzer : DiagnosticAnalyzer
     {
-        internal static DiagnosticDescriptor Rule = new(
+        internal static readonly DiagnosticDescriptor Rule = new(
```
Evangelink, line 17: Have you made the changes manually? It might be better to enable a rule for such cases.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
```diff
@@ -33,7 +33,7 @@ public sealed class PreferAsSpanOverSubstring : DiagnosticAnalyzer
             isPortedFxCopRule: false,
             isDataflowRule: false);
 
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
```
Evangelink, line 36: Not related to this PR, you may want to move this over to your other PR.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
Youssef1313: @Evangelink Yeah these were manual changes :(

There is an analyzer, and it looks like it should be enabled as a warning. But I don't know why it's not working:

https://github.com/dotnet/roslyn-analyzers/blob/90089d9884b3feb90ef33cfe267dabd3c2854768/.editorconfig#L63

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferAsSpanOverSubstring.cs
Youssef1313: @Evangelink Indeed. I did `git stash apply` on the wrong branch 😄

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/PreferIsKindAnalyzer.cs
mavasani: @Youssef1313 That rule only works on private fields. There is no reliable way to flag internal fields without analyzing the entire solution.

mavasani (approved):

PR #5741, opened 2021-11-20
Remove unused project

Evangelink:
From https://github.com/dotnet/roslyn-analyzers/pull/5729#discussion_r753639824


Youssef1313 (approved):

Evangelink:
@Youssef1313 Sorry! I missed your review and force-pushed afterward...

Youssef1313 (approved):
Is the whole `tools` folder not used? We can fully delete it if that's the case.

mavasani:
@jmarolf do you know if anything inside tools folder is used? I don’t recall any usage.

Evangelink:
I have run a global search on the repo and I couldn't find any match so I'd say it seems safe to remove. I will wait for the green-light from @jmarolf

jmarolf (approved):
Thanks for removing this. I am not aware of any usage for this code. this is just more complexity for new contributors

PR #5745, opened 2021-11-22
CA1841: do not report on expression tree contexts

### LINKED ISSUE #5718: False positive for CA1841 in Entity Framework queries
### Analyzer

[CA1841](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1841) `Prefer Dictionary Contains methods`

### Analyzer source

[Microsoft.CodeAnalysis.NetAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers)

6.0.0 (Latest)

### Describe the bug

Entity Framework queries can make use of lists when building queries, e.g.:

``` csharp
var dictionary = new Dictionary<int, SomeOtherType>() { ... };

var data = await this.dataContext.Products.Where(p => dictionary.Keys.Contains(p.Id)).ToListAsync();
```

However it can't make use of ContainsKey because it doesn't translate well to SQL.

I understand that this is probably an edge case, but it would be nice if the rule didn't fire in this case.

### Steps To Reproduce

Use `Dictionary.Keys` in an Entity Framework query

### Expected behavior

Ideally this wouldn't trigger the code analysis rule.

### Actual behavior

The code analysis rule is triggered, meaning in codebases that have warnings as errors enabled the code has to be re-written, or the rule excluded.

Evangelink:
Fix #5718

I have added only one test, let me know if you want me to add the various tests I did on this PR #4522

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferDictionaryContainsMethods.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferDictionaryContainsMethods.cs
@@ -64,6 +64,8 @@ namespace Microsoft.NetCore.Analyzers.Runtime
             if (!compilation.TryGetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemCollectionsGenericIEnumerable1, out var ienumerableType))
                 return;
 
+            var linqExpressionType = compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqExpressionsExpression1);
+
             compilationContext.RegisterOperationAction(OnOperationAction, OperationKind.Invocation);
 
             return;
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferDictionaryContainsMethods.cs
```diff
@@ -86,19 +88,26 @@ void OnOperationAction(OperationAnalysisContext context)
                 //          - If the property's name is "Values", it must match the type substituted for TValue in the IDictionary`2 instance.
 
                 //  Once we have an applicable Contains method and an applicable property reference, we search for an applicable ContainsXXX
-                //  method on the IDictionary`2 receiver.
```
Evangelink, line 89: Arf sorry I forgot there was trailing conflicts on roslyn-analyzer... Let me fix that.

mavasani (approved):

PR #5746, opened 2021-11-22
CA1827 CA1829 CA1836: Do not report within expression tree

### LINKED ISSUE #5299: CA1829 Incorrectly Showing on EF Core 5.0 query
### Analyzer

**Diagnostic ID**: [CA1829](https://docs.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1829): `Use Length/Count property instead of Enumerable.Count method`

### Analyzer source

**Version**: [SDK 5.0.400-preview.21328.4](https://dotnet.microsoft.com/download/dotnet/5.0) (I think)

### Describe the bug

When upgrading one of our projects to use .Net 5 we are seeing an unrelated issue upgrading an EF Core query that used to use the `.Count` property as suggested. These queries now error and don't work so we have been updating to use `.Count()` instead (which fixes the issue with the EF Core query). After doing this we are now receiving `CA1829` on these queries. 

I was wondering if this is a false positive because it's suggesting that we should be using the `.Count` property but this causes the EF Core query to not work?

Happy to try and provide more information if this isn't clear.

Evangelink:
Fix #5299

### DIFF
--- a/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
+++ b/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
@@ -152,10 +152,12 @@ namespace Microsoft.NetCore.Analyzers.Performance
 
             ImmutableHashSet<ITypeSymbol> allowedTypesForCA1836 = allowedTypesBuilder.ToImmutable();
 
+            var linqExpressionType = context.Compilation.GetOrCreateTypeByMetadataName(WellKnownTypeNames.SystemLinqExpressionsExpression1);
+
             if (syncMethods.Count > 0 || asyncMethods.Count > 0)
             {
                 context.RegisterOperationAction(operationContext => AnalyzeInvocationOperation(
-                    operationContext, syncMethods.ToImmutable(), asyncMethods.ToImmutable(), allowedTypesForCA1836),
+                    operationContext, linqExpressionType, syncMethods.ToImmutable(), asyncMethods.ToImmutable(), allowedTypesForCA1836),
                     OperationKind.Invocation);
             }
 
@@ -177,6 +179,7 @@ namespace Microsoft.NetCore.Analyzers.Performance
 
         private static void AnalyzeInvocationOperation(
             OperationAnalysisContext context,
+            INamedTypeSymbol? linqExpressionType,
             ImmutableHashSet<IMethodSymbol> syncMethods,
             ImmutableHashSet<IMethodSymbol> asyncMethods,
             ImmutableHashSet<ITypeSymbol> allowedTypesForCA1836)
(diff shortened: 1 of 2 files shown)


src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -387,13 +395,14 @@ private static void ReportCA1836(OperationAnalysisContext context, string operat
...
-            bool shouldReplaceParent, bool isAsync, bool shouldNegateIsEmpty, bool hasPredicate, string methodName, string? operationKey,
+            bool shouldReplaceParent, bool isAsync, bool shouldNegateIsEmpty, bool hasPredicate, bool isWithinExpressionTree,
+            string methodName, string? operationKey,
             ImmutableHashSet<ITypeSymbol> allowedTypesForCA1836)
         {
             if (!shouldReplaceParent)
             {
                 // Parent expression doesn't met the criteria to be replaced; fallback on analysis for standalone call to Count().
-                if (!isAsync && !hasPredicate)
+                if (!isAsync && !hasPredicate && !isWithinExpressionTree)
```
Evangelink, line 405: This case is for CA1829.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -403,40 +412,39 @@ private static void ReportCA1836(OperationAnalysisContext context, string operat
                 bool shouldNegateAny = !shouldNegateIsEmpty;
                 ReportCA1828(context, shouldNegateAny, operationKey!, methodName, parent);
             }
-            else
+            else if (!isWithinExpressionTree)
```
Evangelink, line 415: This one for CA1827, CA1829 and CA1836.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs
```diff
@@ -403,40 +412,39 @@ private static void ReportCA1836(OperationAnalysisContext context, string operat
                 bool shouldNegateAny = !shouldNegateIsEmpty;
                 ReportCA1828(context, shouldNegateAny, operationKey!, methodName, parent);
```
Evangelink, line 413: In this case, we still want to report within expression tree (at least I think) because I don't expect `CountAsync` to do something specific in queryable contexts.

mavasani (commented):
@Evangelink Can we add unit tests?

Evangelink:
@mavasani PR is ready now.

Evangelink:
Ping @mavasani

mavasani (approved):

Evangelink:
Thank you @mavasani

PR #5754, opened 2021-12-03
Aggregate small fixes for IEnumerable Analyzer

Cosifne:
This PR is an aggregate PR for several small fixes & code clean up for the IEnumerable Analyzer.

Change 1:
Fix the VB cases.
VB will think the extension method as a instance method of the extension type. But C# think extension method is ordinary static method. So the fix is to check the Instances of the InvocationOperation for VB.
Also I add all the VB side unit tests.

Change 2:
Add a few newly added Linq method in .NET 6.

Change 3:
Add a special check for `AsEnumerable`. See the source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq/src/System/Linq/Enumerable.cs). 
Test case:
```
void Bar(IEnumerable<int> a)
{
    var b = [|a|].AsEnumerable();
    b.ElementAt(10);
    b.ElementAt(100);
}
```
because It only does a type cast, which doesn't return a new IEnumerable instance like `Select` `Where`.
So only `a` should be reported.

Change 4:
Some code simplifications.


src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
```diff
@@ -1674,12 +2880,265 @@ public class Bar
...
+            z.ToArray()
+        End Sub
+    End Class
+End Namespace";
+
+            await VerifyVB.VerifyAnalyzerAsync(vbCode);
+        }
+
+        [Fact]
+        public async Task TestCallExtensionMethodAsOrdinaryMethod()
```
Cosifne, line 2909: Starting from here is the 'real' new tests added in this PR.
All the previous change in this tests file are just adding the corresponding VB tests

Cosifne:
Tag @mavasani to review : )

mikadumont (approved):

mavasani (approved):
Looks great!

PR #5762, opened 2021-12-09
Merge feature/IEnumerableAnalyzer back to main

Cosifne:

PR #5749, opened 2021-11-24
Extend the key in Avoid Multiple Enumeration Analyzer

Cosifne:
Motivation of this PR: report case like this
```
var c = a.Concat([|b|].Select(i => i))
c.ElementAt(10);
[|b|].Last();
```

Previously in the AvoidMultipleEnumerations analyzer, the key of the tracked dictionary per block is the AbstractLocations of a Parameter/Local referencing operation, and the value of the dictionary is a set containing all the enumeration operations.

This PR contains mainly 2 changes:
1 Change the key of the dictionary, the key now could be
I. A Symbol
II. An Operation creates the IEnumerable types
III. A set, contains all the possible value sources of the IEnumerable Value.

2. Use a BFS to visit all the possible locations when we start to look into the creation of a local or parameter.
For example
```
void Bar(IEnumerable<int> c)
{
     var b = Enumerable.Range(1, 10);
     var a = b.Concat(c);
     a.ToArray();
}
```
`a.ToArray()` will record it enumerates `b.Concat(c)`, `Enumerable.Range(1, 10)`, `c`.

PR #5304, opened 2021-07-27
Add Analyzer detecting multiple enumerations

Cosifne:
Fix issue: https://github.com/dotnet/roslyn-analyzers/issues/3946

This PR adds the main DFA structure (by mostly reuse the GlobalFlowAnalysis), and the analyzer supports detecting ForEachLoop and most Linq methods.

There are several other things in my mind that could improve the result.
1. Use ValueContent/Copy/PointToAnalysis to have a better understanding of the IEnumreable<T> instance
2. Add a well-known attribute (or name) to denote enumeration
3. Inter-procedure analysis (if possible)

However, that's a lot of other changes. And I feel so far the analyzer is helpful to some extent. So I would like to do that in the following PR.

Issue #3946, opened 2020-08-03
Analyzer for multiple enumerations of Enumerables


mikadumont:
Direct customer ask copying here:

>Showing warnings for multiple enumerations of Enumerables, this actually helped us with a critical bug fix recently.  
![image](https://user-images.githubusercontent.com/46729679/89211367-902dcf80-d576-11ea-96c2-551727b3fe6a.png)

mavasani:
I was thinking about this analyzer over the weekend :-). I am going to give this a shot this week.

Evangelink:
Dupe of #3133. @mavasani which one do you prefer to close?

drewnoakes:
A challenging case for this analyzer:

```c#
public int M(IEnumerable<int> e)
{
    Requires.NotNull(e, nameof(e));

    return e.First();
}
```

How can you know that `Requires.NotNull` doesn't enumerate the object yet `Enumerable.First` does?

nschuessler:
This is a common case we see, somewhat related:

```csharp
class Something
{
     public void One()
     {
          List<string> l = new List<string>()
          {
                 "One", "Two", "Three"
           };

           this.Two(l);
     }

     private void Two(IEnumerable<string> l)
     {
            int count = l.Count();

           this.Three(l);
     }

     private void Three(IEnumerable<string> l)
     {
         foreach (string s in l)
         {
             Console.WriteLine(s);
         }
     }
}
```

Better to pass `IReadOnlyList<string>`  or `ICollection<string>` when passed between methods.

drewnoakes:
@nschuessler I would expect an analyzer to fire in the `Two` method from your example. What won't be clear to the analyzer however is whether the method that receives the enumerable actually enumerates it (i.e. it may just be validating it is not null).

mavasani:
@Evangelink We are considering this analyzer for .NET6 release, would you be interested in implementing it?

Evangelink:
I am a bit short in time lately so I'd say "yes but I can't really commit on ETA". So if anyone else wants to pick this, fine by me.

rui-ktei:
Google took me here. This is actually a much needed feature. In our team, we use Resharper which does underline "possible multiple enumeration" issue, but the problem is that people still ignore it sometimes, so I was searching a way to force this type of "suggestion" to be a compilation error to force devs to pay attention.

[...]

PR #5763, opened 2021-12-09
Cherry pick features/IEnumerable to release/6.0.2xx (Please don't checked in without approval from runtime team)

Cosifne:
features/IEnumerableAnalzyer is based on main.
So in order to get this targeting release/6.0.2.xx, I get this by running
git cherry-pick upstream/main..features/IEnumerableAnalyzer

(The diff commit could be found here https://github.com/dotnet/roslyn-analyzers/pull/5762)
All the merge commits are ignored

And there is another new commit https://github.com/dotnet/roslyn-analyzers/pull/5763/commits/1e8d4f65cc38c3bb3db9f70490945c3e951ee3a9 which fixes our config files


src/Utilities/Compiler/DiagnosticCategoryAndIdRanges.txt
```diff
@@ -12,7 +12,7 @@
 Design: CA2210, CA1000-CA1070
 Globalization: CA2101, CA1300-CA1310
 Mobility: CA1600-CA1601
-Performance: HA, CA1800-CA1849
+Performance: HA, CA1800-CA1849, CA1851
```
Cosifne, line 15: CA1850 is shipped in main, not in this branch.

Cosifne:
I guess @jmarolf should be the right person to review this

jmarolf (approved):

stephentoub:
This is merging the feature branch into a release branch? Was this analyzer approved?

Cosifne:
> This is merging the feature branch into a release branch? Was this analyzer approved?

@stephentoub 
Are you asking for the approved PRs for this analyzer?
https://github.com/dotnet/roslyn-analyzers/pull/5304
https://github.com/dotnet/roslyn-analyzers/pull/5749
https://github.com/dotnet/roslyn-analyzers/pull/5754 

This PR contains all the changes from these PRs. (with the target branch changed to release/6.0.2xx, so I cherry-picked all the commit to another branch)

I have an offline discussion with @jmarolf and he mentioned this branch is going to ship with VS 17.1, and we could ship it if this is disabled by default.

stephentoub:
> Are you asking for the approved PRs for this analyzer?

No, approving the existence of the analyzer.  We have a review process where new analyzers are proposed,  discussed,  approved/rejected, etc, separate from PRs.

Cosifne:
> > Are you asking for the approved PRs for this analyzer?
> 
> No, approving the existence of the analyzer. We have a review process where new analyzers are proposed, discussed, approved/rejected, etc, separate from PRs.
> 
> cc: @mavasani, @jeffhandley, @terrajobst

@stephentoub Do I need to attend any meetings? Or are we just going to discuss it using GitHub issues? (I think this is the analyzer tracking issue https://github.com/dotnet/roslyn-analyzers/issues/3946)

stephentoub:
> Do I need to attend any meetings? Or are we just going to discuss it using GitHub issues? (I think this is the analyzer tracking issue #3946)

I'll let cc: @mavasani, @jeffhandley, @terrajobst comment.  We may have some holes in our process here still :-) Thanks for your efforts and bearing with us.

mavasani:
We generally follow the below guidelines:
1. Any analyzer proposal related to .NET API usage always needs to have an issue opened and triaged in the runtime repo.
2. Analyzer proposal related to generic code quality improvements doesn’t need to go through runtime triage, unless requested by any Roslyn or runtime dev. However, it does need to be discussed at least with the Roslyn team before going ahead with the implementation.

This analyzer falls under the later bucket and was discussed within the Roslyn team and deemed to be a good performance analyzer. However, given that the analyzer has bunch of hard coded logic about which runtime APIs perform an enumeration or not, it seems reasonable to open an issue on runtime repo with the analyzer proposal and design and take it through triage to ensure we have a sound implementation.


Additionally, I agree that it seems a bit too aggressive to merge a new feature into release branch, especially given this is going to be disabled by default. We should let it go through runtime triage and follow regular code flow into .NET7, even if it means we can’t enable the analyzer in Dev17.1

Cosifne:
Thanks for Manish's comments. 
For everyone's reference, here are all the 'hard coded logic about which runtime APIs' in in this analyzer.
Basically they are used to indicate will a method enumerated the parameter or not.

https://github.com/dotnet/roslyn-analyzers/blob/927495b3992d289ba63be8c62d321954e53b8412/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerations/AvoidMultipleEnumerations.cs#L53

https://github.com/dotnet/roslyn-analyzers/blob/927495b3992d289ba63be8c62d321954e53b8412/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerations/AvoidMultipleEnumerations.cs#L87

https://github.com/dotnet/roslyn-analyzers/blob/927495b3992d289ba63be8c62d321954e53b8412/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerations/AvoidMultipleEnumerations.cs#L127

https://github.com/dotnet/roslyn-analyzers/blob/927495b3992d289ba63be8c62d321954e53b8412/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerations/AvoidMultipleEnumerations.cs#L42

Also there is another scenario that mentioned in the original PR comments https://github.com/dotnet/roslyn-analyzers/pull/5304#issuecomment-887909817
From the discussion this might be too aggressive. I can delete this part of code if needed.

jmarolf (changes requested):
waiting for the runtime team to give to go ahead

jeffhandley:
@buyaa-n -- can you extract from this an API proposal issue in dotnet/runtime to present for an API Review in January?

buyaa-n:
> @buyaa-n -- can you extract from this an API proposal issue in dotnet/runtime to present for an API Review in January?

From the email @mavasani sent about this PR, I thought @Cosifne is preparing a proposal and interested in participating in the API review. Anyway, sure I can help @Cosifne with creating an API proposal and get it through review.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerations/AvoidMultipleEnumerations.cs
```diff
@@ -0,0 +1,265 @@
...
+using Analyzer.Utilities;
+using Analyzer.Utilities.Extensions;
+using Analyzer.Utilities.PooledObjects;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeQuality.Analyzers.QualityGuidelines.AvoidMultipleEnumerations.FlowAnalysis;
+using static Microsoft.CodeQuality.Analyzers.MicrosoftCodeQualityAnalyzersResources;
+using static Microsoft.CodeQuality.Analyzers.QualityGuidelines.AvoidMultipleEnumerations.AvoidMultipleEnumerationsHelpers;
+
+namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines.AvoidMultipleEnumerations
```
buyaa-n, line 15: @mavasani is this the correct namespace for this analyzer? I thought we ship in `Microsoft.NetCore.Analyzers`

Cosifne: Might also be good to @jmarolf for the shipping question.
I think release/6.0.2xx will be inserted into VS, and also inserted into the SDK, is that correct?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
```diff
@@ -0,0 +1,3144 @@
...
+
+public class Bar
+{
+    public void Sub()
+    {
+        var i = Enumerable.Range(1, 10).ToArray();
+        foreach (var c in i.Select(p => p))
+        {
+        }
+        i.Count();
```
buyaa-n: I wonder why we are not warning here, i believe it is Linq (IEnumerable) method not Array

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
```diff
@@ -0,0 +1,3144 @@
...
+
+public class Bar
+{
+    public void Sub(IEnumerable<int> h, int x)
+    {
+        var i = Enumerable.Range(1, 10).ToArray();
+        if (x == 1)
+        {
+            i.Except([|h|]).ToList();
+            i.Intersect([|h|]).ToList();
```
buyaa-n: For me this 2 calls will be difficult to implement within a one enumeration, i think its better to not flag such scenario (whenever possible)

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
```diff
@@ -0,0 +1,3144 @@
...
+    public void Sub(int b, IEnumerable<int> j)
+    {
+        var i = Enumerable.Range(1, 10).ToArray();
+        i.First();
+        [|j|].SingleOrDefault();
+
+        if (b == 1)
+        {
+            i.First();
+            [|j|].First();
```
buyaa-n: Isn't it only return the `First` element in the sequesnce? I would not warn for the parameterless overload of `First()`

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
Cosifne: (Why the comment shown as outdated)
`i` is an array here, is that correct?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
Cosifne: Emm, interesting, so maybe for methods like `except` and 'intersect' should be treated differently. Let me see if I can find a full list of methods like this

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
Cosifne: That's interesting. Then do we also want to exclude `ElementAt(0)` ?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
```diff
@@ -0,0 +1,3144 @@
...
+End Namespace
+";
+            await VerifyVB.VerifyAnalyzerAsync(vbCode);
+        }
+
+        [Fact]
+        public async Task TestDelayExecutions()
+        {
+            // SkipLast and TakeLast are unavailable for NET472
+#if NETCOREAPP2_0_OR_GREATER
```
buyaa-n: Normally we specify the ReferenceAssemblies in the test settings instead
https://github.com/dotnet/roslyn-analyzers/blob/107d82a590a1c832f048d4f4bf1df61fd2fb3ec8/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/UseIntegralOrStringArgumentForIndexersTests.cs#L208

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
```diff
@@ -0,0 +1,3144 @@
...
+End Namespace
+";
+            await VerifyVB.VerifyAnalyzerAsync(vbCode);
+        }
+
+        [Fact]
+        public void TestNet6AddedMethod()
+        {
+            // newly added method in .Net 6, currently can't found by unit tests
+            Assert.Contains("MaxBy", AvoidMultipleEnumerations.s_enumeratedParametersLinqMethods);
```
buyaa-n: You can test APIs in .Net 6 https://github.com/dotnet/roslyn-analyzers/blob/107d82a590a1c832f048d4f4bf1df61fd2fb3ec8/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/DetectPreviewFeatureUnitTests.Misc.cs#L21

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
buyaa-n: > (Why the comment shown as outdated)

hm strange all of them shows outdated, not sure why

> i is an array here, is that correct?

Yes but `Array` doesn't have `Count()` method right?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
buyaa-n: Maybe, in case it can be excluded easily

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
Cosifne: Yes, my understanding it since `i` is an `Array` here, `Count()` won't call the GetEnuemrator() method to calculate the number of element, it will just fetch `Array.Count`, which is an O(1) operation. Is that correct?

buyaa-n: if you `ctrl+click` on the `Count` method of `i.Count();` above it will go to `IEnumerable.Count` extension method. As `Array` doesn't have `Count` method defined, and it implements `IEnumerable`, so when you call Array.Count it would call  `IEnumerable.Count` extension method

buyaa-n: Ah i see what you mean, even it calls `IEnumerable.Count` it would use ICollection.Cout(), never mind https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq/src/System/Linq/Count.cs#L18-L21

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/AvoidMultipleEnumerationsTests.cs
Cosifne: I have been thinking about this for some time when editing the code.
I think this might be two kinds of  'multiple enumerations'.
1. Multiple call to the `MoveNext()` mthod, like `i.ElementAt(10)`. Calling this over `IEnumerable<T>` is the same as doing an indexing operation over a linked list.

2. The source of `IEnuemrable<T>` itself might change.
```
void Bar()
{
    IEnumerable<int> result = FetchDataFromDatabase();
    var a = result.First();
    WriteSomeDataIntoDataBase();
    var b = result.First();
}
```
In this case, even it is just called `First`, I still feel we should flag it. Because `a` and `b` might be two different values. And `First` is really querying the DB twice.

My purpose solution here would be, for a given variable which is `IEnumerable<T>`, if it is assigned from `System.Linq.Range()` or `System.Line.Repeat()`, we can safely ignore the usage of `First`, `FirstOrDefault`, etc...
Ex:
```
void Bar()
{
     var c = Enumerable.Range(1, 10000);
     c.First();
     c.FirstOrDefault();
}
```
No diagnostic should be reported.

But if the variable is assigned from some unknown method call, we should consider `First`, `FirstOrDefault`, etc...

buyaa-n: > In this case, even it is just called First, I still feel we should flag it. Because a and b might be two different values.
> And First is really querying the DB twice.

Did not understood, for me `Because a and b might be two different values.` is the reason for not flagging it, becaues the values could be different its NEED to query twice

Cosifne: My point is this is a good case that people might misuse the Linq. (The first impression for this is that `a` should equal to `b`, but in fact because it queries DB twice, the value would be different)

buyaa-n: I am not sure how that could be a misuse of Linq, anyway i don't think we need to evaluate subsequent calls of `First()` or try to undestand the user expectations for those calls, because that doesn't seem to relate to the purpose of this analyzer (detecting multiple Enumerations), `First()` would never call `MoveNext()` more than once (by the point 1 you mentioned)

Cosifne: OK I agree we don't add this behavior in the first revision of this analyzer.
Probably we want to wait for some feedback to see if customers really want this. 
I do really feel this might lead to bugs in reality. but we can wait to see it

Cosifne: I am going to talk about this with our PM, because in the original post https://github.com/dotnet/roslyn-analyzers/issues/3946#issue-672955039
I think the post is showing 

> Multiple call to the MoveNext() mthod, like i.ElementAt(10). Calling this over IEnumerable<T> is the same as doing an indexing operation over a linked list.

But I want to double check if the customer may/may not want to 

> The source of IEnuemrable<T> itself might change.`

If needed I can bring that to the runtime design meeting again.

buyaa-n: I thnk you are mixing up my comment with other methods (like `ElementAt(10)`) My comment here is only about `First()` which would never make multiple call to the `MoveNext()`. For `ElementAt(10)` it would make multiple  `MoveNext()` therefore you should check/flag multiple calls

Cosifne: No, I understand your idea here. `First()` or `FirstOrDefault()` will never call 'MoveNext()' twice. And for method like `ElementAt()` this is a different thing.

My question is mainly regarding should the analyzer report diagnostic when calling `First` or `FirstOrDefault` (the second example I posted above) This is not put in the initial post, but I want to verify with the PM about this

buyaa-n: > but I want to verify with the PM about this

Sure thing

Cosifne: Based on the offline email discussion, I will mark this as resovle.
But thank you to raise this question. : )

Issue #5403, opened 2021-08-25
CA1416 doesn't support `Debug.Assert(bool, string? message)`


Bouke:
I have the following platform guarded code:

```c#
Debug.Assert(OperatingSystem.IsWindows(), "Only supported on windows");
System.ServiceProcess.ServiceBase.Run(service);
```

**Expected behaviour:**  
No diagnostic.

**Actual behaviour:**  
Diagnostic CA1416: "`Run` is only supported on 'windows'"

**Workaround:**  
Remove the message argument:

```c#
Debug.Assert(OperatingSystem.IsWindows());
System.ServiceProcess.ServiceBase.Run(service);
```

However this yields SA1405: "Debug.Assert should provide message text."

**Context:**  
Solution targeting `net5.0`, sdk is at version 5.0.400.

**Analysis:**  
I believe the following line to be relevant. While the method name `IsAnyAssertMethod` suggests that it matches any `Debug.Assert` signature, it only matches the first signature having 1 argument (`Debug.Assert(bool)`).

https://github.com/dotnet/roslyn-analyzers/blob/6b29c673cb8e31865216b64cd3b0a2aca281a32b/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs#L996-L997

https://github.com/dotnet/roslyn-analyzers/blob/6b29c673cb8e31865216b64cd3b0a2aca281a32b/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs#L993-L994

https://github.com/dotnet/roslyn-analyzers/blob/6b29c673cb8e31865216b64cd3b0a2aca281a32b/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs#L262-L263

https://github.com/dotnet/roslyn-analyzers/blob/6b29c673cb8e31865216b64cd3b0a2aca281a32b/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs#L336-L342

PR #5764, opened 2021-12-10
Support all Debug.Assert(...) overloads in flow analysis

buyaa-n:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/5403

Flow analysis doesn't support `Debug.Assert(bool, string? message)` or any other overload except `Debug.Assert(bool)`:
```cs
void M1()
{
    Debug.Assert(OperatingSystem.IsWindows(), "Only supported on windows");
    WindowsOnly(); // warns even not supposed to

    Debug.Assert(OperatingSystem.IsWindows());
    WindowsOnly(); // not warn
}

void WindwosOnly () { }
```
I believe all overloads of Debug.Assert(...) should be supported cc @mavasani

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
@@ -3523,6 +3523,55 @@ End Class";
         }
 
         [Fact]
+        public async Task GuardedWith_DebugAssertWithMessage()
+        {
+            var source = @"
+using System;
+using System.Diagnostics;
+using System.Runtime.Versioning;
+using System.Runtime.InteropServices;
+
+class Test
+{
+    void M1()
+    {
+        Debug.Assert(RuntimeInformation.IsOSPlatform(OSPlatform.Windows), ""{0} is only supported on windows"", nameof(M3));
+        M3();
+        [|M2()|];
+
+        Debug.Assert(OperatingSystem.IsOSPlatformVersionAtLeast(""Windows"", 10, 2), ""Only supported on windows"");
+        M2();
+        M3();
+    }
+
+    [SupportedOSPlatform(""Windows10.1.2.3"")]
+    void M2() { }
+
+    [SupportedOSPlatform(""Windows"")]
+    void M3() { }
+}";
+            await VerifyAnalyzerCSAsync(source);
+
+            var vbSource = @"
+Imports System.Diagnostics
+Imports System.Runtime.Versioning
+Imports System
+
+Class Test
+    Private Sub M1()
+        [|M2()|]
+        Debug.Assert(OperatingSystem.IsOSPlatformVersionAtLeast(""Windows"", 10, 2), ""Only supported on windows"")
+        M2()
+    End Sub
+
+    <SupportedOSPlatform(""Windows10.1.2.3"")>
+    Private Sub M2()
+    End Sub
+End Class";
+            await VerifyAnalyzerVBAsync(vbSource);
+        }
+
+        [Fact]
         public async Task GuardedWith_ResultSavedInLocalAsync()
         {
             var source = @"
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
@@ -3535,13 +3535,17 @@ class Test
 {
     void M1()
     {
-        Debug.Assert(RuntimeInformation.IsOSPlatform(OSPlatform.Windows), ""{0} is only supported on windows"", nameof(M3));
+        Debug.Assert(RuntimeInformation.IsOSPlatform(OSPlatform.Windows), ""Only supported on windows"", ""Detailed message"");
         M3();
         [|M2()|];
+        [|M4()|];
 
         Debug.Assert(OperatingSystem.IsOSPlatformVersionAtLeast(""Windows"", 10, 2), ""Only supported on windows"");
         M2();
         M3();
+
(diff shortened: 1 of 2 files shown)


src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -990,8 +990,11 @@ private bool IsContractCheckMethod(IMethodSymbol method)
             return false;
         }
 
-        private bool IsDebugAssertMethod(IMethodSymbol method)
-            => Equals(method, DebugAssertMethod);
+        private bool IsDebugAssertMethod(IMethodSymbol method) =>
+            DebugAssertMethod != null &&
```
buyaa-n, line 994: `DebugAssertMethod` property initialized with `Debug.Assert(bool)` method if exist

mavasani: Can you please add a doc comment and also rename the method to IsAnyDebugAssertMethod?

mavasani (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
```diff
@@ -3515,6 +3515,55 @@ Class Test
         M2()
     End Sub
 
+    <SupportedOSPlatform(""Windows10.1.2.3"")>
```
Evangelink, line 3518: That might be too much work, but it'd be nice to add tests for the various rules impacted by having this change.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -990,8 +990,11 @@ private bool IsContractCheckMethod(IMethodSymbol method)
             return false;
         }
 
-        private bool IsDebugAssertMethod(IMethodSymbol method)
-            => Equals(method, DebugAssertMethod);
+        private bool IsDebugAssertMethod(IMethodSymbol method) =>
```
Evangelink, line 993: Would it be worth to add some option to enable/disable this behavior? I absolutely understand the reason of this change but I think that depending on the context of the application we may want to be really sure of some "state". 
I am not sure if I am really clear but `Debug.Assert` gives some safety in `Debug` but after that in `Release` you no longer have any check which means you could end up in this invalid scenario.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -990,8 +990,11 @@ private bool IsContractCheckMethod(IMethodSymbol method)
...
+            DebugAssertMethod != null &&
+            method.ContainingSymbol.Equals(DebugAssertMethod.ContainingSymbol, SymbolEqualityComparer.Default) &&
+            method.Name == DebugAssertMethod.Name &&
+            method.ReturnType == DebugAssertMethod.ReturnType;
```
Youssef1313, line 997: SymbolEqualityComparer? I'm surprised `==` actually works.

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
buyaa-n: Sure

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/PlatformCompatibilityAnalyzerTests.GuardedCallsTests.cs
buyaa-n: > That might be too much work

That is what i thought 😛, ok, i will add some more

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
buyaa-n: Not sure I understood your point, but `Debug.Assert` is for asserting expected behavior in `Debug` mode and to catch a failure in a development phase, it should/could not assert `Release`

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
buyaa-n: Did not undertsood, for `SymbolEqualityComparer` I used `SymbolEqualityComparer.Equals` not `==`

Youssef1313: @buyaa-n I usee you used SymbolEqualityComparer for comparing `ContainingSymbol`s, but not for `ReturnType`s.

> method.ReturnType == DebugAssertMethod.ReturnType;

buyaa-n:
Perf failure unrelated, [filed an issue](https://github.com/dotnet/roslyn-analyzers/issues/5790), mergin the PR

src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DataFlowOperationVisitor.cs
```diff
@@ -990,11 +990,19 @@ private bool IsContractCheckMethod(IMethodSymbol method)
             return false;
         }
 
-        private bool IsDebugAssertMethod(IMethodSymbol method)
-            => Equals(method, DebugAssertMethod);
+        /// <summary>
+        /// Checks if the method is an overload of the <see cref="Debug.Assert(bool)"/> method.
+        /// </summary>
+        /// <param name="method">The IMethodSymbol to test.</param>
+        /// <returns>True if the method is an overlaod of the <see cref="Debug.Assert(bool)"/> method.</returns>
```
MartyIX, line 997: `/// <returns>True if the method is an overload of the <see cref="Debug.Assert(bool)"/> method.</returns>` :)

PR #5770, opened 2021-12-17
Analyzer: Use Null Suppression Correctly

steveberdy:
This analyzer and code fix is really about correcting over-usage of, and incorrect usage of, null suppression. It includes very obvious cases, yet these cases are never flagged, even by the compiler. Here are some examples of what I mean:

```csharp
string x = null!; // compiler doesn't flag this
int y = x.Length / 2; // but this throws NullReferenceException, because somebody lied to their compiler...
```

```csharp
string x = default!; // default should never be null suppressed
DateTime y = default!; // especially not here, DateTime is a value type
```

```csharp
string x = "Hello"!; // this is... dumb, but not discouraged by the compiler or anything else
```

This analyzer simply alerts of misusage of null suppression. It doesn't offer code fixes since there are quite a few complex cases.


Youssef1313:
There are cases where `null!` is okay. There is actually an open issue in dotnet/roslyn around unnecessary null suppressions https://github.com/dotnet/roslyn/issues/34714.

steveberdy:
Here is how I understand it, which is why it made sense to never allow `null!`. When you use null suppression, you're saying that although something is expressly nullable, the values it takes are guaranteed to never be null. When you use null forgiveness, on the other hand, you're saying that the value may be null, and if it is null, any properties you access from it will not throw a `NullReferenceException` but will rather return null or default. Correct me if I'm wrong.

stephentoub:
> When you use null suppression, you're saying that although something is expressly nullable, the values it takes are guaranteed to never be null

That's not the case.  You're saying "don't warn here", but there are many reasons that might be the case.  You can see a bunch of examples in dotnet/runtime, e.g. nulling out an array element upon removing an item from a collection,  setting fields to null during disposal, initializing a non-nullable field knowing a derived implementation will subsequently always set it to non-null, etc. `null!` and `default!` are allowed and necessary in various situations.

PR #5776, opened 2021-12-24
Handle 'CodeAnalysisTreatWarningsAsErrors' correctly in props/targets

### LINKED ISSUE #5775: Inconsistent severity level behavior.
**Version Used**: Microsoft.CodeAnalysis.NetAnalyzers v5.0.3

**Steps to Reproduce**:

1. .csproj
```
    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
    <TreatWarningsAsErrors></TreatWarningsAsErrors>
```
2. .editorconfig
```
dotnet_analyzer_diagnostic.severity = error
```
3. C# code
```
using var sha256 = new Rfc2898DeriveBytes(value, salt: CatalystTelemetrySalt, iterations: 4096);
var hashBytes = sha256.GetBytes(hashLengthBytes);
```

**Expected Behavior**:
This code should show two errors: CA5379 and CA5387

**Actual Behavior**:
CA5379 appears as an Error, but CA5387 appears as a Warning.

mavasani:
Fixes #5775

Current implementation handled the case where 'CodeAnalysisTreatWarningsAsErrors' is 'false' by appending all CA rules IDs to 'WarningsNotAsErrors' property. This PR ensures that we also handle the case where 'CodeAnalysisTreatWarningsAsErrors' is 'true' and 'TreatWarningsAsErrors' is not 'true' by appending the CA rule IDs to 'WarningsAsErrors' property


mavasani:
@jmarolf The performance failures seem unrelated

Youssef1313:
@sharwell FYI it looks like the bot added the "Community" label incorrectly.

mavasani:
@jmarolf Can you please review?

src/Tools/GenerateDocumentationAndConfigFiles/Program.cs
```diff
@@ -303,17 +303,17 @@ string getDisableNetAnalyzersImport()
...
                 return $@"
   <!-- 
-    This property group prevents the rule ids implemented in this package to be bumped to errors when
-    the 'CodeAnalysisTreatWarningsAsErrors' = 'false'.
+    This property group handles 'CodeAnalysisTreatWarningsAsErrors' for the CA rule ids implemented in this package.
   -->
   <PropertyGroup>
     <CodeAnalysisRuleIds>{allRuleIds}</CodeAnalysisRuleIds>
     <WarningsNotAsErrors Condition=""'$(CodeAnalysisTreatWarningsAsErrors)' == 'false'"">$(WarningsNotAsErrors);$(CodeAnalysisRuleIds)</WarningsNotAsErrors>
+    <WarningsAsErrors Condition=""'$(CodeAnalysisTreatWarningsAsErrors)' == 'true' and '$(TreatWarningsAsErrors)' != 'true'"">$(WarningsAsErrors);$(CodeAnalysisRuleIds)</WarningsAsErrors>
```
jmarolf, line 316: If `$(TreatWarningsAsErrors)` is not defined but `$(CodeAnalysisTreatWarningsAsErrors)` is set to truw this condition will occur is that correct?

jmarolf: I guess I need you to write out a TreatWarningsAsErrors.md file where you lay out the spec here. From my understanding we have:

- `CodeAnalysisTreatWarningsAsErrors`
- `TreatWarningsAsErrors`
- `WarningsAsErrors`
- `WarningsNotAsErrors`

Without a spec laying out what the expected behavior of all these is I do not know how to review this.

mavasani: @jmarolf Following are the precedence rules as per the values of these properties.

1. For non-CAxxxx rules
   1. `CodeAnalysisTreatWarningsAsErrors` has no relevance and is ignored.
   2. `TreatWarningsAsErrors`, if not set, defaults to false. If true, this property translates to `/warnaserror` command line switch.
   3. Compiler bumps all warnings to errors iff `TreatWarningsAsErrors` is true. Users can prevent escalation or force escalation of individual warnings to errors by appending the IDs to `WarningsNotAsErrors` or `WarningsAsErrors`, which just translate to `/warnaserror[+|-]:<%rule ids%> on the command line.

2. For CAxxxx rules:
   1. If `CodeAnalysisTreatWarningsAsErrors` and `TreatWarningsAsErrors` both are not set, no bulk settings to escalate or de-escalate warnings to errors is done.
   2. If `CodeAnalysisTreatWarningsAsErrors` is set, it overrides `TreatWarningsAsErrors` to determine if CA warnings are bulk escalated to errors or not.
   3. If `CodeAnalysisTreatWarningsAsErrors` is not set, it defaults to `TreatWarningsAsErrors`.
   4. If final value of `CodeAnalysisTreatWarningsAsErrors = false`, we append all CA rule IDs to `WarningsNotAsErrors` to ensure they are not escalated to errors. Users can still bump individual rule IDs to errors by editorconfig/ruleset entry, etc.
   5. If final value of `CodeAnalysisTreatWarningsAsErrors = true`, we append all CA rule IDs to `WarningsAsErrors` to ensure they are escalated to errors. This is the change done in this PR. I have tried to optimize it a bit more by avoiding this append if `TreatWarningsAsErrors` is also true, because then the compiler itself will take care of bumping all warnings to errors, and we don't need to pollute the command line with large number of CA rules IDs in a `/warnaserror+` switch. I expect this to be the most common case as well (`TreatWarningsAsErrors` is set by user to true, `CodeAnalysisTreatWarningsAsErrors` is never set and hence defaults to `true`), and I want to ensure we don't end up polluting the entire command line with CA rules IDs unless `TreatWarningsAsErrors` and `CodeAnalysisTreatWarningsAsErrors` have different settings.

jmarolf (approved):
Thanks so much for the explanation. Can you copy/paste that into a Readme.md file in `src/Tools/GenerateDocumentationAndConfigFiles`? Maybe there is a better place but I want to have this handy if I ever need to make changes here.

PR #5799, opened 2022-01-17
Implement 'Enum values should be monotonically increasing'

Youssef1313:
- Code fix not yet implemented. Will open an issue for follow up if I couldn't get it in this PR.
- VB not yet implemented. It should be trivial to do it.

Fixes https://github.com/dotnet/runtime/issues/63926


Youssef1313:
@RikkiGibson Would you like to take a look?

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/Maintainability/EnumValuesShouldBeMonotonicallyIncreasing.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        private void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            if (context.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Enum } enumSymbol)
+            {
+                return;
+            }
+
+            var members = enumSymbol.GetMembers();
+            BigInteger? previous = null;
```
Youssef1313, line 49: Not sure if using `BigInteger` is good here. It was the easiest to be able to handle all different possible enum underlying types.

RikkiGibson:
Thanks for taking on the implementation of this. 

At a glance it looks like this is not using an IOperation analyzer. Was there a specific reason for that?

Will review in more detail tomorrow (after holiday). Thanks again!

Youssef1313:
I couldn't find a suitable `OperationKind` to register the callback for.

stephentoub:
Shouldn't these be going through a triage before they're implemented in the. NET SDK?
@jeffhandley

Youssef1313:
@stephentoub In that case, can you move https://github.com/dotnet/roslyn-analyzers/issues/5783 to dotnet/runtime? Thanks!

Meanwhile, if it's going to take long for the analyzer to be approved, we can move this analyzer to `Roslyn.Diagnostics.Analyzers` package, which is intended to only be used by Roslyn.

jeffhandley:
I'm closing out this PR as it was inadvertently left open and has gotten quite stale. The backing issue hasn't been championed through API review.

PR #5810, opened 2022-01-24
RS1024: Add codefix for collection cases

Evangelink:


src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
```diff
@@ -58,6 +70,110 @@ private async Task<Document> ConvertToEqualsAsync(Document document, TextSpan so
...
+                    await CallOverloadWithEqualityComparerAsync(
+                        document, invocation.Syntax, invocation.TargetMethod, invocation.Arguments, IsExtensionMethodUsedAsSuch(invocation),
+                        (generator, args) => generator.InvocationExpression(GetExpression(invocation), args), iEqualityComparer, cancellationToken)
+                    .ConfigureAwait(false),
+
+                _ => document,
+            };
+
+            static bool IsExtensionMethodUsedAsSuch(IInvocationOperation invocation)
+                => invocation.IsExtensionMethodAndHasNoInstance()
```
Evangelink, line 104: I am a little surprise to see no difference on the `Instance` property when the extension method is used as extension or isn't. That's why I am checking that the 1st arg is implicit (which means it is used as extension) or not which means it's used directly.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
```diff
@@ -38,6 +41,15 @@ public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
                             equivalenceKey: nameof(CompareSymbolsCorrectlyFix)),
                         diagnostic);
                 }
+                else if (diagnostic.Descriptor == CompareSymbolsCorrectlyAnalyzer.CollectionRule)
+                {
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            CodeAnalysisDiagnosticsResources.CompareSymbolsCorrectlyCodeFix,
+                            cancellationToken => CallOverloadWithEqualityComparerAsync(context.Document, diagnostic.Location.SourceSpan, cancellationToken),
+                            equivalenceKey: nameof(CompareSymbolsCorrectlyFix)),
```
mavasani, line 50: Are we purposefully using the same equivalence key as for prior fix that converts to equals? Note that this will cause FixAll for a diagnostic triggered for one of these rules to also fix all instances of other rule. I think you want to probably use a different equivalence key?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
```diff
@@ -863,14 +863,26 @@ public async Task CollectionConstructorsKnownToRequireComparer_DiagnosticAsync()
...
+        [|new Dictionary<ISymbol, int>(42)|];
+        [|new Dictionary<ISymbol, int>(kvps)|];
+
         [|new HashSet<ISymbol>()|];
+        [|new HashSet<ISymbol>(42)|];
+        [|new HashSet<ISymbol>(symbols)|];
+
         [|new ConcurrentDictionary<ISymbol, int>()|];
+        [|new ConcurrentDictionary<ISymbol, int>(kvps)|];
+        [|new ConcurrentDictionary<ISymbol, int>(1, 42)|];
```
mavasani, line 885: Can we add unit tests where existing invocations and object creations use named arguments? Also do you know if we have existing tests where code already passes a comparer and hence no diagnostic is raised?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
```diff
@@ -883,6 +895,46 @@ public void MethodWithoutDiagnostics()
...
+    public void MethodWithoutDiagnostics()
+    {
+        new Dictionary<int, ISymbol>();
+        new HashSet<string>();
+        new ConcurrentDictionary<int, ISymbol>();
+    }
+}",
+                        SymbolEqualityComparerStubCSharp,
+                    }
+
```
mavasani, line 936: Nit
```suggestion
```

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
Evangelink: I think I didn't think it through enough but somehow I don't see why people willing to fix `Equals` would not want to fix the collection at the same time. If there was a way to differentiate the rule I would think splitting the code-fix would make sense but it's not the case.

I'll let you have the final word, so please let me know your preference.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
Evangelink: @mavasani Is there a bug in roslyn or is it the intended behavior?

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
Evangelink: I have added one test with named argument and one test with passing ` SymbolEqualityComparer.Default`.

src/Microsoft.CodeAnalysis.Analyzers/UnitTests/MetaAnalyzers/CompareSymbolsCorrectlyTests.cs
```diff
@@ -858,60 +858,170 @@ public async Task CollectionConstructorsKnownToRequireComparer_DiagnosticAsync()
             {
                 TestState =
                 {
+                    ReferenceAssemblies = CreateNetCoreReferenceAssemblies(),
```
Evangelink, line 861: Looks like something changed, I was able to run the test without that before.

Evangelink:
Ping @mavasani

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
mavasani: @Evangelink This is by design. IOperation is meant to reflect semantics, and the difference you mention is purely syntactic.

src/Microsoft.CodeAnalysis.Analyzers/Core/MetaAnalyzers/Fixers/CompareSymbolsCorrectlyFix.cs
mavasani: Fine with keeping this as-is. If we get complaints, we can split.

mavasani: Fine with keeping this as-is. If we get complaints, we can split.

mavasani (approved):

PR #5812, opened 2022-01-24
CA1810: Report only if field belongs to ctor's type

### LINKED ISSUE #5784: false-positive for CA1810
### Analyzer

**Diagnostic ID**: [CA1810](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1810): `Initialize reference type static fields inline`

### Analyzer source

**Version**: [SDK 6.0.100](https://dotnet.microsoft.com/download/dotnet/6.0)

### Describe the bug

False Positive. Static constructor sets static class fields.
(https://github.com/dotnet/msbuild/blob/797fd829a374a880f63fd4eea4ea6990404c48f4/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs#L39-L49)

### Steps To Reproduce
Run the analyzer on the above mentioned code.

## Additional context
Using the analyzer on msbuild the related issue is [#7179](https://github.com/dotnet/msbuild/pull/7179)

Evangelink:
Fix #5784

### DIFF
--- a/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
@@ -370,6 +370,41 @@ public class C2
             }.RunAsync();
         }
 
+        [Fact, WorkItem(5784, "https://github.com/dotnet/roslyn-analyzers/issues/5784")]
+        public async Task CA1810_InitializeMemberOfAnotherClass()
+        {
+            await VerifyCS.VerifyAnalyzerAsync(@"
+using System;
+
+public class C1
+{
+    internal static int Size = 0;
+}
+
+public class C2
+{
+    static C2()
+    {
+        C1.Size = 42;
+    }
+}
+");
+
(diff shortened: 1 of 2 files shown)


Youssef1313 (approved):

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/Runtime/InitializeStaticFieldsInlineTests.cs
```diff
@@ -370,6 +370,41 @@ static C2()
...
+public class C1
+{
+    internal static int Size = 0;
+}
+
+public class C2
+{
+    static C2()
+    {
+        C1.Size = 42;
```
mavasani, line 388: What would be the result if the static ctor initializes both static fields of another type and also static fields of the containing type? Would it match our expected result in that case?

sharwell (changes requested):
Haven't reviewed the implementation, but need to verify this is truly a false positive. CA1810 is a performance analyzer that I don't believe is enabled by default, so we'll want to make sure the analyzer is fully preserving the original intent after this change.

Evangelink:
@sharwell Please let me know if you think that's better to close PR.

sharwell:
@Evangelink yep, the analyzer is behaving by design here

PR #5813, opened 2022-01-24
CA1805: Fix FP with parameterless ctor on value types

### LINKED ISSUE #5750: False positive CA1805 when using parameterless value type ctor (C# 10)
### Analyzer

**Diagnostic ID**: [CA1805](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/CA1805): `Do not initialize unnecessarily.`

### Analyzer source

**Version**: 6.0.0 (Latest)

### Describe the bug

False positive when using parameterless value type ctor (C# 10).

### Steps To Reproduce
```C#
struct S
{
    public static readonly S Value = new(); //false positive CA1805

    public S() => throw null;
}
```

Beware that ctor does not need to be explicitly stated. In following example parameterless ctor is generated implicitly and CA1805 keeps to be false positive.

```C#
struct S
{
    public static readonly S Value = new(); //false positive CA1805

    private List<int> list = new();
}
```

Evangelink:
Fixes #5750

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -44,7 +44,7 @@
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
     <MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>3.7.0</MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>
-    <MicrosoftCodeAnalysisVersionForTests>4.0.1</MicrosoftCodeAnalysisVersionForTests>
+    <MicrosoftCodeAnalysisVersionForTests>4.2.0-1.22070.1</MicrosoftCodeAnalysisVersionForTests>
     <DogfoodAnalyzersVersion>3.3.3</DogfoodAnalyzersVersion>
     <DogfoodNetAnalyzersVersion>5.0.4-preview1.21126.5</DogfoodNetAnalyzersVersion>
     <MicrosoftCodeAnalysisBannedApiAnalyzersVersion>$(DogfoodAnalyzersVersion)</MicrosoftCodeAnalysisBannedApiAnalyzersVersion>
--- a/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
+++ b/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
@@ -46,7 +46,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                     !field.IsConst &&
                     init.Value != null &&
                     field.GetAttributes().IsEmpty && // in case of attributes that impact nullability analysis
-                    UsesKnownDefaultValue(init.Value, field.Type))
+                    UsesKnownDefaultValue(init.Value, field.Type, field.ContainingType))
                 {
                     context.ReportDiagnostic(init.CreateDiagnostic(DefaultRule, field.Name));
                 }
@@ -59,14 +59,14 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
                 if (prop != null &&
                     init.Value != null &&
                     prop.GetAttributes().IsEmpty && // in case of attributes that impact nullability analysis
-                    UsesKnownDefaultValue(init.Value, prop.Type))
+                    UsesKnownDefaultValue(init.Value, prop.Type, prop.ContainingType))
                 {
                     context.ReportDiagnostic(init.CreateDiagnostic(DefaultRule, prop.Name));
                 }
             }, OperationKind.PropertyInitializer);
         }
 
-        private bool UsesKnownDefaultValue(IOperation value, ITypeSymbol type)
+        private bool UsesKnownDefaultValue(IOperation value, ITypeSymbol type, INamedTypeSymbol containingType)
         {
             // Skip through built-in conversions
             while (value is IConversionOperation conversion && !conversion.Conversion.IsUserDefined)
@@ -83,7 +83,7 @@ namespace Microsoft.CodeQuality.Analyzers.QualityGuidelines
 
             // If this is default(T) or new ValueType(), it's the default.
             if (value is IDefaultValueOperation ||
-                (type.IsValueType && value is IObjectCreationOperation oco && oco.Arguments.IsEmpty && oco.Initializer is null))
+                (type.IsValueType && value is IObjectCreationOperation oco && oco.Arguments.IsEmpty && oco.Initializer is null && !SymbolEqualityComparer.Default.Equals(oco.Type, containingType)))
             {
                 return !IsNullSuppressed(value);
             }
--- a/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarilyTests.cs
+++ b/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarilyTests.cs
@@ -2,6 +2,7 @@
 
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Testing;
+using Test.Utilities;
 using Xunit;
 using VerifyCS = Test.Utilities.CSharpCodeFixVerifier<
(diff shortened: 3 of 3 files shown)


mavasani:
@stephentoub for review

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -46,7 +46,7 @@ public sealed override void Initialize(AnalysisContext context)
                     !field.IsConst &&
                     init.Value != null &&
                     field.GetAttributes().IsEmpty && // in case of attributes that impact nullability analysis
-                    UsesKnownDefaultValue(init.Value, field.Type))
+                    UsesKnownDefaultValue(init.Value, field.Type, field.ContainingType))
```
stephentoub, line 49: Why is this focusing on the field's containing type? This doesn't look right. In C# 10, any struct can define a parameterless ctor, the issue isn't just when the field is of a type it's a part of.

sharwell: 📝 See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3432 for how a similar situation was handled, in case we want to do the same here.

Evangelink: @stephentoub I didn't quite understand when the change was actually meaningful (@sharwell comment on StyleCop ticket helped me).

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarilyTests.cs
```diff
@@ -311,6 +312,65 @@ public class Test
                     .WithLocation(3));
         }
 
+        [Fact, WorkItem(5750, "https://github.com/dotnet/roslyn-analyzers/issues/5750")]
+        public async Task CA1805_ParameterlessValueTypeCtorAsync()
```
stephentoub, line 316: Nit: the naming of this doesn't match the pattern used by the rest of the tests.

stephentoub (approved):
Thanks.

eng/Versions.props
```diff
@@ -44,7 +44,7 @@
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
     <MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>3.7.0</MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>
-    <MicrosoftCodeAnalysisVersionForTests>4.0.1</MicrosoftCodeAnalysisVersionForTests>
+    <MicrosoftCodeAnalysisVersionForTests>4.2.0-1.22070.1</MicrosoftCodeAnalysisVersionForTests>
```
sharwell, line 47: ❓   Was this necessary? It looks like parameterless struct constructors were merged to 4.0.1, in which case we would not need to update the test support as part of this pull request.

sharwell (changes requested):
Waiting for question to be answered

eng/Versions.props
Evangelink: When I tried to reproduce the issue, I had to bump the version but maybe in between something was merged. I will try to revert and will let you know (probably tonight, I am pretty full today).

Evangelink: @sharwell So it wasn't required!

sharwell (approved):

Youssef1313 (commented):
Can you add few tests for https://github.com/dotnet/roslyn-analyzers/issues/5887?


```csharp
public record struct MyRecord()
{
    public bool SomeBool { get; set; } = false;
}
```

----

```csharp
public record struct MyRecord
{
    public MyRecord() { }
    public bool SomeBool { get; set; } = false;
}
```

Youssef1313 (commented):
There is no test for a `struct` without explicit constructor. This now produces CS8983 error anyway, so we may not care, but it didn't use to produce an error before https://github.com/dotnet/roslyn/commit/3e4cc4cd39422b519a379ca1bb71e860687f77d0 (even before that commit I'm not sure why issue states it being a false positive)

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -83,7 +83,7 @@ private bool UsesKnownDefaultValue(IOperation value, ITypeSymbol type)
 
             // If this is default(T) or new ValueType(), it's the default.
             if (value is IDefaultValueOperation ||
-                (type.IsValueType && value is IObjectCreationOperation oco && oco.Arguments.IsEmpty && oco.Initializer is null))
+                (type.IsValueType && value is IObjectCreationOperation oco && oco.Arguments.IsEmpty && oco.Initializer is null && oco.Constructor.IsImplicitlyDeclared))
```
Youssef1313, line 86: Per https://github.com/dotnet/roslyn-analyzers/issues/5887#issuecomment-1044617402, the diagnostic should never be reported for structs.

stephentoub: My comment was suggesting we shouldn't report it when initializing instance members on structs.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarilyTests.cs
```diff
@@ -311,6 +312,101 @@ public class Test
...
+        {
+            await new VerifyCS.Test
+            {
+                TestCode = @"
+using System;
+
+struct S
+{
+    public static readonly S Value = new();
+    public static readonly TimeSpan Time = new();
```
Evangelink, line 326: With my initial change we were reporting here but now that we exclude all instance members of struct we no longer report. Let me know if you would prefer to have a few hardcoded types for which we know we can still "simplify" initialization.

Evangelink:
@stephentoub @Youssef1313 ready for review. Thanks!

stephentoub (approved):
Thanks.

src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs
```diff
@@ -58,6 +59,7 @@ public sealed override void Initialize(AnalysisContext context)
                 IPropertySymbol? prop = init.InitializedProperties.FirstOrDefault();
                 if (prop != null &&
                     init.Value != null &&
+                    prop.ContainingType?.TypeKind != TypeKind.Struct && // Do not report on instance members of struct
```
Youssef1313, line 62: The comment mentions *instance* members, but the check is for structs in general, including static and instance members. Anything I'm missing? @stephentoub @Evangelink

stephentoub: Ideally the rule would still flag statics.

Evangelink: Damn, sorry I missed that. Will look into it tonight (CET time).

Evangelink: It's now fixed.

Evangelink:
@stephentoub @Youssef1313 PR should be finally ready.

ManickaP:
I'm hitting this problem in my PR, so it would be great if we could merge this. @Evangelink is the PR ready to be merged? Has it become stale?

Evangelink:
@ManickaP I will merge main to have the CI issues fixed but as far as I remember it was ready to be merged.

PR #5822, opened 2022-01-28
Implement DisableRuntimeMarshalling analyzer and code fixer

jkoritzinsky:
Fixes dotnet/runtime#63714


jkoritzinsky:
I meant to open this as a draft PR since it's not totally ready yet. My mistake.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,256 @@
...
+        public const string CanConvertToDisabledMarshallingEquivalentKey = nameof(CanConvertToDisabledMarshallingEquivalentKey);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(
+            FeatureUnsupportedWhenRuntimeMarshallingDisabled,
+            MethodUsesRuntimeMarshallingEvenWhenMarshallingDisabled);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution(); // TODO: Make the auto-layout cache multithreading friendly.
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```
Youssef1313, line 58: In general we ignore generated code for non-security analyzers.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,256 @@
...
+                    marshalMethods.AddRange(marshalType.GetMembers("PtrToStructure"));
+                    _marshalMethods = marshalMethods.ToImmutable();
+                }
+            }
+
+            public void AnalyzeMethodCall(OperationAnalysisContext context)
+            {
+                IInvocationOperation invocation = (IInvocationOperation)context.Operation;
+
+                if (invocation.TargetMethod.IsGenericMethod ? _marshalMethods.Contains(invocation.TargetMethod.ConstructedFrom) : _marshalMethods.Contains(invocation.TargetMethod))
```
Youssef1313, line 106: I think we can always use `invocation.TargetMethod.ConstructedFrom`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,256 @@
...
+                    return;
+                }
+
+                AnalyzeMethodSignature(context.ReportDiagnostic, functionPointerInvocation.GetFunctionPointerSignature(), ImmutableArray.Create(functionPointerInvocation.WrappedOperation.Syntax.GetLocation()));
+            }
+
+            public void AnalyzeType(SymbolAnalysisContext context)
+            {
+                INamedTypeSymbol type = (INamedTypeSymbol)context.Symbol;
+                if (type.TypeKind != TypeKind.Delegate || (_unmanagedFunctionPointerAttribute is not null && !type.HasAttribute(_unmanagedFunctionPointerAttribute)))
```
Youssef1313, line 143: Consider not registering the symbol action if `_unmanagedFunctionPointerAttribute` is null. This will safe few unnecessary callbacks.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -0,0 +1,1072 @@
...
+        where U : unmanaged
+    {
+        object obj = default(ValueType);
+        int instanceSize = {|CA1421:Marshal.SizeOf(obj)|};
+        int size = {|CA1421:Marshal.SizeOf(typeof(ValueType))|};
+        int sizeGeneric = {|CA1421:Marshal.SizeOf<ValueType>()|};
+        int sizePassedInType = {|CA1421:Marshal.SizeOf(type)|};
+        int sizePassedInGeneric = {|CA1421:Marshal.SizeOf<T>()|};
+        int sizePassedInGenericUnmanaged = {|CA1421:Marshal.SizeOf<U>()|};
+    }
```
Youssef1313, line 853: This tests whether we'll crash on accessing the first argument or not.

>                     "SizeOf" => invocation.TargetMethod.IsGenericMethod || invocation.Arguments[0].Value is ITypeOfOperation { TypeOperand.IsUnmanagedType: true },

```suggestion
        int nonGenericNoParameter = Marshal.SizeOf();
    }
```

Consider also adding the test for `StructureToPtr`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1834,4 +1834,46 @@
...
+    <value>Types with '[StructLayout(LayoutKind.Auto)]'</value>
+  </data>
+  <data name="ByRefParameters" xml:space="preserve">
+    <value>By-ref parameters</value>
+  </data>
+  <data name="FeatureUnsupportedWhenRuntimeMarshallingDisabledDescription" xml:space="preserve">
+    <value>Using features that require runtime marshalling when runtime marshalling is disabled will result in runtime exceptions.</value>
+  </data>
+  <data name="FeatureUnsupportedWhenRuntimeMarshallingDisabledMessage" xml:space="preserve">
+    <value>{0} requires runtime marshalling to be enabled</value>
```
Youssef1313, line 1847:
```suggestion
    <value>'{0}' requires runtime marshalling to be enabled</value>
```

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,119 @@
...
+                    if (operation.TargetMethod.TypeArguments[0].IsUnmanagedType)
+                    {
+                        addUnsafeToEnclosingMethod = true;
+                        editor.ReplaceNode(syntax, SyntaxFactory.SizeOfExpression(GetGenericTypeParameterSyntax(syntax)));
+                    }
+                }
+                else if (operation.Arguments[0].Value is ITypeOfOperation { TypeOperand.IsUnmanagedType: true } typeOf)
+                {
+                    addUnsafeToEnclosingMethod = true;
+                    editor.ReplaceNode(syntax, SyntaxFactory.SizeOfExpression(GetTypeOfTypeSyntax((TypeOfExpressionSyntax)typeOf.Syntax)));
```
Youssef1313, line 75: Consider handling trivia.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                editor.SetModifiers(enclosingMethod, editor.Generator.GetModifiers(enclosingMethod).WithIsUnsafe(true));
+            }
+
+            return editor.GetChangedDocument();
+
+            static TypeSyntax GetGenericTypeParameterSyntax(InvocationExpressionSyntax syntax)
+            {
+                var memberAccess = (MemberAccessExpressionSyntax)syntax.Expression;
+                var genericName = (GenericNameSyntax)memberAccess.Name;
```
Youssef1313, line 99: Is it possible the expression bind against the generic method, but the `<..>` is omitted from syntax and it's implicitly inferred, in which case the cast could fail?

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (addUnsafeToEnclosingMethod)
+            {
+                editor.SetModifiers(enclosingMethod, editor.Generator.GetModifiers(enclosingMethod).WithIsUnsafe(true));
+            }
+
+            return editor.GetChangedDocument();
+
+            static TypeSyntax GetGenericTypeParameterSyntax(InvocationExpressionSyntax syntax)
+            {
+                var memberAccess = (MemberAccessExpressionSyntax)syntax.Expression;
```
Youssef1313, line 98: I think this cast can fail:

```csharp
using static System.Runtime.InteropServices.Marshal;

public class C
{
    public void M()
    {
        SizeOf(...);
    }
}
```

mavasani:
> I meant to open this as a draft PR since it's not totally ready yet. My mistake.

@jkoritzinsky Is the PR ready for review after your latest commits? If not, can you please convert it to a draft? Thanks.

jkoritzinsky:
@mavasani this is ready for review.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,286 @@
...
+            {
+                if (operation.TargetMethod.IsGenericMethod)
+                {
+                    if (operation.TargetMethod.TypeArguments[0].IsUnmanagedType)
+                    {
+                        editor.ReplaceNode(syntax, SyntaxFactory.SizeOfExpression((TypeSyntax)editor.Generator.TypeExpression(operation.TargetMethod.TypeArguments[0])));
+                        return true;
+                    }
+                }
+                else if (operation.Arguments[0].Value is ITypeOfOperation { TypeOperand.IsUnmanagedType: true } typeOf)
```
Youssef1313, line 184: This doesn't feel safe for error scenarios.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,283 @@
...
+                        invocation.TargetMethod.ToDisplayString()));
+                }
+            }
+
+            private static bool CanTransformToDisabledMarshallingEquivalent(IInvocationOperation invocation)
+            {
+                return invocation.TargetMethod.Name switch
+                {
+                    "OffsetOf" => false,
+                    "SizeOf" => invocation.TargetMethod.IsGenericMethod || invocation.Arguments[0].Value is ITypeOfOperation { TypeOperand.IsUnmanagedType: true },
```
Youssef1313, line 131: I don't think accessing the first element is safe. The invocation may not have a parameter (error scenario).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,283 @@
...
+                    "StructureToPtr" => invocation.Arguments[0].Value.Type.IsUnmanagedType,
```
Youssef1313, line 132: Same comment here

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,283 @@
...
+                finally
+                {
+                    _layoutCacheLock.ExitReadLock();
+                }
+
+                if (_structLayoutAttribute is not null)
+                {
+                    foreach (var attr in type.GetAttributes())
+                    {
+                        if (attr.AttributeClass.OriginalDefinition.Equals(_structLayoutAttribute))
```
Youssef1313, line 243: `OriginalDefinition` doesn't seem necessary.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,283 @@
...
+                }
+                finally
+                {
+                    _layoutCacheLock.ExitReadLock();
+                }
+
+                if (_structLayoutAttribute is not null)
+                {
+                    foreach (var attr in type.GetAttributes())
+                    {
```
Youssef1313, line 242: Consider using `HasAttribute` extension:

https://github.com/dotnet/roslyn-analyzers/blob/60eb13c44c366205e5a4dec195404477009c41de/src/Utilities/Compiler/Extensions/ISymbolExtensions.cs#L628-L631

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,283 @@
...
+                        if (attr.AttributeClass.OriginalDefinition.Equals(_structLayoutAttribute))
+                        {
+                            var argument = attr.ConstructorArguments[0];
```
Youssef1313, line 245: This doesn't feel safe in error scenarios.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,283 @@
...
+                    }
+                }
+
+                foreach (var member in type.GetMembers())
+                {
+                    if (member is IFieldSymbol { IsStatic: false, Type.IsValueType: true } valueTypeField)
+                    {
+                        return TypeIsAutoLayoutOrContainsAutoLayout(valueTypeField.Type);
+                    }
+                }
```
Youssef1313, line 277: Is it expected we return on the first non-static value type field without looking into the remaining fields?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -0,0 +1,1072 @@
...
+        {
+            return VerifyCSCodeFixAsync(source, source, allowUnsafeBlocks);
+        }
+
+        private static async Task VerifyCSCodeFixAsync(string source, string codeFix, bool allowUnsafeBlocks = false)
+        {
+            var test = new VerifyCS.Test
+            {
+                LanguageVersion = LanguageVersion.CSharp9,
+                ReferenceAssemblies = new ReferenceAssemblies("net7.0", new PackageIdentity("Microsoft.NETCore.App.Ref", "7.0.0-preview.1.22075.6"), Path.Combine("ref", "net7.0"))
```
Youssef1313, line 1055: @sharwell Can this be added in the test library?

sharwell: Yes

sharwell: Though I'm hesitant to do so before 7.0.0-preview.1 is available on NuGet...

jkoritzinsky: Yeah I don't think we should update the testing SDK until we know for sure exactly which package version will be Preview 1. This just happened to be the newest Preview 1 package when I started working on this PR.

jkoritzinsky:
@dotnet/roslyn-analysis can someone take a look at reviewing this? It's been sitting for 2 weeks waiting for review

eng/Versions.props
```diff
@@ -45,7 +45,7 @@
     <!-- Roslyn -->
     <MicrosoftCodeAnalysisVersion>3.3.1</MicrosoftCodeAnalysisVersion>
     <MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>3.7.0</MicrosoftCodeAnalysisForRoslynDiagnosticsAnalyzersVersion>
-    <MicrosoftCodeAnalysisVersionForTests>4.0.1</MicrosoftCodeAnalysisVersionForTests>
+    <MicrosoftCodeAnalysisVersionForTests>4.1.0-3.22075.3</MicrosoftCodeAnalysisVersionForTests>
```
mavasani, line 48: 4.1.0 is now released.
```suggestion
    <MicrosoftCodeAnalysisVersionForTests>4.1.0</MicrosoftCodeAnalysisVersionForTests>
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+                {
+                    bool canTransformToDisabledMarshallingEquivalent = CanTransformToDisabledMarshallingEquivalent(invocation);
+                    context.ReportDiagnostic(invocation.CreateDiagnostic(
+                        MethodUsesRuntimeMarshallingEvenWhenMarshallingDisabled,
+                        ImmutableDictionary.Create<string, string?>().Add(CanConvertToDisabledMarshallingEquivalentKey, canTransformToDisabledMarshallingEquivalent ? "true" : null),
+                        invocation.TargetMethod.ToDisplayString()));
+                }
+            }
+
+            private static bool CanTransformToDisabledMarshallingEquivalent(IInvocationOperation invocation)
```
mavasani, line 126: Make it a local function?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+                {
+                    context.ReportDiagnostic(method.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabled, SetLastErrorTrue));
+                }
+
+                if (!method.MethodImplementationFlags().HasFlag(System.Reflection.MethodImplAttributes.PreserveSig))
+                {
+                    context.ReportDiagnostic(method.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabled, HResultSwapping));
+                }
+
+                if (_lcidConversionAttribute is not null && method.GetAttributes(_lcidConversionAttribute).Any())
```
mavasani, line 182:
```suggestion
                if (method.HasAttribute(_lcidConversionAttribute))
```

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+                {
+                    var paramLocation = locationsOverride.IsDefaultOrEmpty ? param.Locations : locationsOverride;
+                    if (param.RefKind != RefKind.None)
+                    {
+                        reportDiagnostic(paramLocation.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabled, ByRefParameters));
+                    }
+                    AnalyzeSignatureType(paramLocation, param.Type);
+                }
+
+                void AnalyzeSignatureType(ImmutableArray<Location> locations, ITypeSymbol type)
```
mavasani, line 208: consider making this static and passing in `reportDiagnostic`

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        private class PerCompilationAnalyzer
+        {
+            private readonly INamedTypeSymbol? _unmanagedFunctionPointerAttribute;
+            private readonly INamedTypeSymbol? _structLayoutAttribute;
+            private readonly INamedTypeSymbol? _lcidConversionAttribute;
+            private readonly ImmutableArray<ISymbol> _marshalMethods;
+            private readonly ReaderWriterLockSlim _layoutCacheLock = new();
+            private readonly Dictionary<ITypeSymbol, bool> _isAutoLayoutOrContainsAutoLayoutCache = new();
```
mavasani, line 80: We normally use a `ConcurrentDictionary` in this repo, but this is fine if you prefer dictionary guarded with a lock. Is there a specific reason for using it here or just your preference? try/catch/reader writer locks do make the code look bit more verbose

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+                                finally
+                                {
+                                    _layoutCacheLock.ExitWriteLock();
+                                }
+                            }
+                        }
+                    }
+                }
+
+                foreach (var member in type.GetMembers())
```
mavasani, line 269: Should we also be writing to the cache for true/false return values from here on in this method? It seems the cache is only ever populated with true values for structs with StructLayoutAttribute with auto layout kind, and we would end up repeatedly executing the code from here on in this method for structs without auto layout StructLayoutAttribute.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+                void AnalyzeSignatureType(ImmutableArray<Location> locations, ITypeSymbol type)
+                {
+                    if (!type.IsUnmanagedType)
+                    {
+                        reportDiagnostic(locations.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabled, ManagedParameterOrReturnTypes));
+                    }
+
+                    if (type.IsValueType && TypeIsAutoLayoutOrContainsAutoLayout(type))
+                    {
+                        reportDiagnostic(locations.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabled, ByRefParameters));
```
mavasani, line 217: So, use of localized strings as string format arguments for diagnostic messages is not recommended. This can turn out to generate incorrect localized strings for various locales. Instead, you should use different localized strings for each of these combinations of `FeatureUnsupportedWhenRuntimeMarshallingDisabled` diagnostics. This would mean you need to create different diagnostic descriptors for each of these district messages, all of them using the same diagnostic ID, but just a different message, which is fine, we do it at bunch of places in the repo. For example, see this for quite a descriptor explosion: https://github.com/dotnet/roslyn-analyzers/blob/577deb8e3365755b7d047c4b411d4a86f852be59/src/Text.Analyzers/Core/IdentifiersShouldBeSpelledCorrectly.cs#L53-L211

We are forced to take this approach to ensure proper localized strings for diagnostic messages, no other alternatives really work well here.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1834,4 +1834,46 @@
...
+    <value>HResult-swapping</value>
+  </data>
+  <data name="ManagedParameterOrReturnTypes" xml:space="preserve">
+    <value>Managed parameter or return types</value>
+  </data>
+  <data name="MethodUsesRuntimeMarshallingEvenWhenMarshallingDisabledDescription" xml:space="preserve">
+    <value>This method uses runtime marshalling even when runtime marshalling is disabled, which can cause unexpected behavior differences at runtime due to different expectations of a type's native layout.</value>
+  </data>
+  <data name="MethodUsesRuntimeMarshallingEvenWhenMarshallingDisabledMessage" xml:space="preserve">
+    <value>'{0}' uses runtime marshalling even when 'System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute' is applied. Use features like 'sizeof' and pointers directly to ensure accurate results</value>
```
mavasani, line 1862: Message string with multiple sentences should end with a period. We do have an RSxxxx analyzer to catch these - I know it was busted for a while, but I thought we fixed the analyzer. @Youssef1313 @Evangelink do you recall the current state for this resx validation analyzer?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1834,4 +1834,46 @@
...
+  <data name="SetLastErrorTrue" xml:space="preserve">
+    <value>Setting SetLastError to 'true'</value>
+  </data>
+  <data name="UsingLCIDConversionAttribute" xml:space="preserve">
+    <value>Using 'LCIDConversionAttribute'</value>
+  </data>
+  <data name="VarargPInvokes" xml:space="preserve">
+    <value>Varadic P/Invoke signatures</value>
+  </data>
+  <data name="UseDisabledMarshallingEquivalent" xml:space="preserve">
```
mavasani, line 1876: We generally try to suffix resx string used for code action titles with `CodeFix` or `CodeFixTitle` for clarity. Also, is it possible to use a smaller string content for the title as this will cause the lightbulb to be a lot wider?

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
```diff
@@ -362,10 +362,10 @@ class C2 { }
 }";
 
             var expectedMetricsText = @"
-Assembly: (Lines: 7, ExecutableLines: 0, MntIndex: 100, CycCxty: 2, CoupledTypes: {N1.C1, System.Attribute}, DepthInherit: 2)
-   N1: (Lines: 7, ExecutableLines: 0, MntIndex: 100, CycCxty: 2, CoupledTypes: {N1.C1, System.Attribute}, DepthInherit: 2)
+Assembly: (Lines: 7, ExecutableLines: 2, MntIndex: 100, CycCxty: 2, CoupledTypes: {N1.C1, System.Attribute}, DepthInherit: 2)
```
mavasani, line 365: Interesting - so I presume this change is coming from consuming a new Microsoft.CodeAnalysis package referenced in tests, and is unrelated to this PR. We seem to be counting IOperation blocks for attributes as having 2 executable lines of code (basically 2 explicit operations in the operation block).

@333fred @Youssef1313 did we already change the operation tree for attributes in 4.1.0 version of MS.CA? The code for computing executable lines for an operation block for code metrics is here: https://github.com/dotnet/roslyn-analyzers/blob/086d1cc888ca0d8c96b94444935f488fcaf3fe98/src/Utilities/Compiler/CodeMetrics/ComputationalComplexityMetrics.cs#L87-L355

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -0,0 +1,1072 @@
...
+            string source = @"
+using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+
+[assembly:DisableRuntimeMarshalling]
+
+class Foo
+{
+    [DllImport(""Foo"", SetLastError = true)]
+    public static extern void {|CA1420:Bar|}();
```
mavasani, line 28: Based on the suggested change in https://github.com/dotnet/roslyn-analyzers/pull/5822/files#r817552866, writing tests for CA1420 would become more tricky when you have the analyzer reporting multiple descriptors with the same diagnostic ID but different messages. @sharwell do you have an example showing the recommended format for writing the interpolation arg for such tests? Or will the correct format continue to work and test library will just pick the first descriptor to verify only the reported diagnostic ID?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -0,0 +1,1072 @@
...
+        IntPtr offsetGeneric = {|CA1421:Marshal.OffsetOf<ValueType>(""field"")|};
+    }
+}
+
+struct ValueType
+{
+    int field;
+}
+";
+            await VerifyCSAnalyzerAsync(source);
```
mavasani, line 829: Given CA1421 has a code fix, we should `VerifyCSCodeFixAsync` here, even if the scenario has no applicable code fix and we are forced to use `VerifyCSCodeFixAsync(source, source);`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -0,0 +1,1072 @@
...
+                LanguageVersion = LanguageVersion.CSharp9,
+                ReferenceAssemblies = new ReferenceAssemblies("net7.0", new PackageIdentity("Microsoft.NETCore.App.Ref", "7.0.0-preview.1.22075.6"), Path.Combine("ref", "net7.0"))
+                    .WithNuGetConfigFilePath(Path.Combine(Path.GetDirectoryName(typeof(DisableRuntimeMarshallingTests).Assembly.Location), "NuGet.config")),
+                TestCode = source,
+                FixedCode = codeFix,
+                SolutionTransforms =
+                {
+                    (solution, projectId) => solution.WithProjectCompilationOptions(projectId, (solution.GetProject(projectId).CompilationOptions as CSharpCompilationOptions)?.WithAllowUnsafe(allowUnsafeBlocks))
+                },
+                NumberOfFixAllIterations = -2
```
mavasani, line 1063: What does -2 mean for number of fixall iterations? @sharwell?

src/Utilities/Compiler/Lightup/IFunctionPointerInvocationOperationWrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+    using System;
+    using System.Collections.Immutable;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq.Expressions;
+    using System.Reflection;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Operations;
+
+    [SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "Not a comparable instance.")]
+    internal readonly struct IFunctionPointerInvocationOperationWrapper : IOperationWrapper
```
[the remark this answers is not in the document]
mavasani: I am presuming function pointer operations shipped in 4.1.0 or earlier. If so, I am wondering if we can bump the version for the entire repo instead of doing it just for tests: https://github.com/dotnet/roslyn-analyzers/blob/4b0b87381bd345d9e52c9d915dc6d3442cf84f5b/eng/Versions.props#L46

@jmarolf Is .NET7 guaranteed to ship with Roslyn version greater than or equals 4.1.0?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,282 @@
...
+using Analyzer.Utilities.Lightup;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices
+{
+    using static MicrosoftNetCoreAnalyzersResources;
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```
mavasani, line 21: The analyzer supports VB, but we don't have any VB tests in the PR. We either need to add VB tests for the analyzer OR remove VB support for the analyzer (either move it out to a follow-up change or completely remove it, either is fine).

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
jkoritzinsky: ConcurrentDictionary has a lot of built-in overhead, so I’m a little weary of using it by default without benchmarking (I’ve been burned by perf issues from using it).

stephentoub: @jkoritzinsky, can you elaborate on what overhead you're concerned about? RWLS generally has way more overhead.

jkoritzinsky: When I was working on C#/WinRT, I defaulted to using ConcurrentDictionary when I needed a multithreaded dictionary, and it’s mechanisms of allocating caches across multiple threads to ensure efficiency has caused no end of perf headaches due to how many ConcurrentDictionary instances are used there. If this isn’t a concern, I can switch this back to ConcurrentDictionary.

stephentoub: I don't believe it should be a concern until you measure it to be a concern (and if perf is the driving reason, you also need to measure the overhead of RWLS, which has way higher overhead on every read and write).  But if it's really a concern, you can set the concurrencyLevel to 1 in the ctor.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
jkoritzinsky: The definition of `VerifyCSAnalyzerAsync` in this file calls `VerifyCSCodeFixAsync(source, source)` on line 1047.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,286 @@
...
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.NetCore.Analyzers.InteropServices
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
+    public sealed class CSharpDisableRuntimeMarshallingFixer : CodeFixProvider
```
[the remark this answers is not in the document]
jkoritzinsky: All of the code-fixes provide fixes that are not applicable to VB code. They all require C# `unsafe` language features that are not available in VB.

src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/Maintainability/CodeMetricsComputationTests.cs
333fred: It's not merged yet.

src/Utilities/Compiler/Lightup/IFunctionPointerInvocationOperationWrapper.cs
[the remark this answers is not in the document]
333fred: > I am presuming function pointer operations shipped in 4.1.0.

Yes, that was VS 17.1/4.1.0: https://github.com/dotnet/roslyn/pull/57191.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        public override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(DisableRuntimeMarshallingAnalyzer.MethodUsesRuntimeMarshallingEvenWhenMarshallingDisabledId);
+
+        public sealed override FixAllProvider GetFixAllProvider()
+        {
+            return CustomFixAllProvider.Instance;
+        }
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
```
mavasani, line 113: Nit: Can we move the overrides to the top of the file and the nested classes to separate files?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/MicrosoftNetCoreAnalyzersResources.resx
```diff
@@ -1834,4 +1834,46 @@
   <data name="UsesPreviewTypeParameterMessage" xml:space="preserve">
     <value>'{0}' uses the preview type '{1}' and needs to opt into preview features. See {2} for more information.</value>
   </data>
+  <data name="ByRefParameters" xml:space="preserve">
+    <value>By-ref parameters</value>
+  </data>
```
mavasani, line 1839: Unused?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -277,6 +487,45 @@ struct ValueType2
...
         }
 
+        [Fact]
+        public async Task VB_PInvokeWithUnmanagedValueTypeParameter_WithAutoLayoutField_Emits_Diagnostic()
+        {
+            string source = @"
+Imports System.Runtime.CompilerServices
+Imports System.Runtime.InteropServices
+<assembly:DisableRuntimeMarshalling>
+Public Class Foo
```
mavasani, line 497: I just realized that policheck will flag `Foo`. Can we rename all the `Foo` and `foo` strings in this file to something else?

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -1042,6 +1321,38 @@ class ClassType
             await VerifyCSCodeFixAsync(source, source, allowUnsafeBlocks: false);
         }
 
+
+
```
mavasani, line 1325: Nit: Unnecessary new lines

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -1060,6 +1371,12 @@ private static async Task VerifyCSCodeFixAsync(string source, string codeFix, bo
                 {
                     (solution, projectId) => solution.WithProjectCompilationOptions(projectId, (solution.GetProject(projectId).CompilationOptions as CSharpCompilationOptions)?.WithAllowUnsafe(allowUnsafeBlocks))
                 },
+                // Because we can't always fix all cases of our diagnostic,
+                // the test infrastructure can run up to 2 iterations of our fix-all code fix.
+                // The first run fixes all the fixable diagnostics.
+                // Since there are still some (unfixable) diagnostics, the test infrastructure decides to run the fix-all provider again.
+                // The second run doesn't do anything, since the remaining diagnostics are unfixable.
+                // Setting NumberOfFixAllIterations to -2 specifies that the fix-all provider can be run up to 2 times as part of a test run.
```
mavasani, line 1379: Thank you for the comment!

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -169,22 +241,22 @@ public void AnalyzeMethod(SymbolAnalysisContext context)
...
                 }
 
                 if (!method.MethodImplementationFlags().HasFlag(System.Reflection.MethodImplAttributes.PreserveSig))
                 {
-                    context.ReportDiagnostic(method.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabled, HResultSwapping));
+                    context.ReportDiagnostic(method.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabledHResultSwapping));
                 }
 
-                if (_lcidConversionAttribute is not null && method.GetAttributes(_lcidConversionAttribute).Any())
+                if (_lcidConversionAttribute is not null && method.HasAttribute(_lcidConversionAttribute))
```
mavasani, line 252: `_lcidConversionAttribute is not null` is redundant, `HasAttribute` performs the null check.

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -1382,6 +1382,7 @@ private static async Task VerifyCSCodeFixAsync(string source, string codeFix, bo
 
             // Verify that there are some instances of the diagnostic that we can't fix.
             test.FixedState.MarkupHandling = MarkupMode.Allow;
+            test.MarkupOptions = MarkupOptions.UseFirstDescriptor;
```
mavasani, line 1385: This is good to know!

mavasani (approved):
Overall LGTM. Have added minor comments to address. Feel free to merge after addressing/responding to them. Thanks.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
```diff
@@ -0,0 +1,296 @@
...
+
+            protected override string CodeActionTitle => MicrosoftNetCoreAnalyzersResources.UseDisabledMarshallingEquivalentCodeFix;
+
+            protected override async Task<SyntaxNode> FixAllInDocumentAsync(FixAllContext fixAllContext, Document document, ImmutableArray<Diagnostic> diagnostics)
+            {
+                if (document.Project.CompilationOptions is CSharpCompilationOptions { AllowUnsafe: false })
+                {
+                    // We can't code fix if unsafe code isn't allowed.
+                    return await document.GetSyntaxRootAsync(fixAllContext.CancellationToken);
+                }
```
Youssef1313, line 36: Is this redundant? This was already checked in `RegisterCodeFixesAsync`

src/NetAnalyzers/UnitTests/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshallingTests.cs
```diff
@@ -0,0 +1,1404 @@
...
+using System.Runtime.CompilerServices;
+
+[assembly:DisableRuntimeMarshalling]
+
+class Foo
+{
+    [DllImport(""Foo"")]
+    [LCIDConversion(0)]
+    public static extern void {|CA1420:Bar|}();
+}
```
Youssef1313, line 67: Consider testing a local function and validate if the behavior for it is correct.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,354 @@
...
+
+                if (dllImportData.SetLastError)
+                {
+                    context.ReportDiagnostic(method.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabledSetLastErrorTrue));
+                }
+
+                if (!method.MethodImplementationFlags().HasFlag(System.Reflection.MethodImplAttributes.PreserveSig))
+                {
+                    context.ReportDiagnostic(method.CreateDiagnostic(FeatureUnsupportedWhenRuntimeMarshallingDisabledHResultSwapping));
+                }
```
Youssef1313, line 250: Is there a test for this?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,354 @@
...
+
+            public void AnalyzeMethod(SymbolAnalysisContext context)
+            {
+                IMethodSymbol method = (IMethodSymbol)context.Symbol;
+
+                DllImportData? dllImportData = method.GetDllImportData();
+                if (dllImportData is null)
+                {
+                    return;
+                }
```
Youssef1313, line 240: Why is the return? There are other checks that don't depend on `dllImportData` below?

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
```diff
@@ -0,0 +1,354 @@
...
+
+                AnalyzeMethodSignature(context.ReportDiagnostic, type.DelegateInvokeMethod);
+            }
+
+            public void AnalyzeMethod(SymbolAnalysisContext context)
+            {
+                IMethodSymbol method = (IMethodSymbol)context.Symbol;
+
+                DllImportData? dllImportData = method.GetDllImportData();
```
Youssef1313, line 236: I'm surprised this didn't cause a crash due to:

https://github.com/dotnet/roslyn/blob/7e2966e180eef601cb0fc4a029d4f2093d22b639/src/Compilers/CSharp/Portable/Symbols/FunctionPointers/FunctionPointerMethodSymbol.cs#L829

Also surprised on roslyn's side there is an exception thrown in this code path instead of simply returning null. cc @333fred

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
jkoritzinsky: This analyzer doesn't apply to any non-DllImported methods.

src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/InteropServices/DisableRuntimeMarshalling.cs
jkoritzinsky: We don't go down this path for function pointers. We go from `AnalyzeFunctionPointerCall` to `AnalyzeMethodSignature` directly.

Youssef1313: `FunctionPointerMethodSymbol` is a *method* symbol. Shouldn't we get callbacks for it from `                context.RegisterSymbolAction(AnalyzeMethod, SymbolKind.Method);`? then we meet `method.GetDllImportData()` unconditionally. I know this is likely not happening, otherwise a test would have crashed. But I don't yet see the reason

jkoritzinsky: I think we don't get callbacks for it for the same reason that we don't get callbacks for `SymbolKind.Method` for local functions (which I just added code to handle in my most recent commit). It seems that symbols that are introduced due to an `IOperation` are exposed through that operation in `RegisterOperationAction` instead of through `RegisterSymbolAction`.

Youssef1313: Good to know! Thanks!

Youssef1313 (commented):
Do we want to report a diagnostic for properties?

e.g:

```csharp
public static extern string P { [DllImport("abc")] get; }
```

I don't know if there will already be a diagnostic or not.

jkoritzinsky:
I didn't know that was even valid...

Yes, we should add the diagnostic there too.

Youssef1313:
@jkoritzinsky Consider also:

```
using System;

public class C
{
    [DllImport("abc")]
    public static extern int operator +(C a, C b);
    
    [method: DllImport("abc")]
    public extern static event Action G;
}
```

Note: those edge cases are being stolen from Roslyn tests 😄

jkoritzinsky:
This is so much sadness right now 😢 

The way the event test case compiles down also makes me sad, but I guess I should add tests to validate behavior.

src/NetAnalyzers/CSharp/Microsoft.NetCore.Analyzers/InteropServices/CSharpDisableRuntimeMarshalling.Fixer.cs
jkoritzinsky: I tried turning this into an assert and it tripped on a few tests, so I guess it's not redundant.

jkoritzinsky:
@mavasani I don't have permissions to merge, so can you hit the button for me?

